PR #2613, opened 2021-12-17
Smry trans

berland:
**Issue**
Resolves #my_issue

**Approach**
_Short description of the approach_

## Pre review checklist

- [ ] Added appropriate labels

PR #2685, opened 2022-01-11
Fix SPE1 yaml config for smry_keys

### LINKED ISSUE #2683: Fix recordtree transmitter for _storage
When transmitting recordtree to storage, we obtain a wrong url with `userdata` in the end, which needs to fixed, eg.:
```
/ensembles/f450a2f3-3c19-4a8b-9330-1402b366d4a5/records/summary?realization_index=0/userdata
```

### LINKED ISSUE #2684: SPE1 demo is currently broken
**Describe the bug**
The spe1 demo silently fails on the main branch (see https://github.com/equinor/ert/runs/4772647822?check_suite_focus=true)

**To Reproduce**
1. `pip install ert`
2. `ert3 service start storage`
3. `ert3 init --example spe1`
4. `cd spe1`
5.  `ert3 run evaluation`

This results in: 
```
Error while loading stages configuration data:
5 validation errors for StagesConfig
__root__ -> 0 -> output -> keys
  extra fields not permitted (type=value_error.extra)
__root__ -> 0 -> function
  field required (type=value_error.missing)
__root__ -> 0 -> script
  extra fields not permitted (type=value_error.extra)
__root__ -> 0 -> transportable_commands
  extra fields not permitted (type=value_error.extra)
__root__ -> 0 -> output -> keys
  extra fields not permitted (type=value_error.extra)
```

**Expected behavior**
The above commands should run the evaluation. If there is a problem with the configuration the script should return a non-zero exit code (and thus fail the Github Action pipeline):

**Enviromment**
 - OS: RHEL7 and Ubuntu 20.04
 - ERT Release: main
 - Python version: 3.6 and 3.8

berland:
**Issue**
Resolves #2684 
Resolves #2683 

**Approach**
Following up syntax change in #2613 

## Pre review checklist
.


ert/storage/_storage.py
```diff
@@ -85,17 +85,18 @@ async def _transmit_recordtree(
         storage_url = self._uri[: self._uri.rfind("/")]
         for record_path in record.flat_record_dict:
             record_name = record_path.split("/")[-1]
-            transmitter = StorageRecordTransmitter(record_name, storage_url)
+            transmitter = StorageRecordTransmitter(
+                record_name, storage_url, iens=self._real_id
```
berland, line 89: `iens` is a mnemonic used for legacy reasons I guess. Probably another PR to change that to something more precise, like "`real_id`"

ert/storage/_storage.py
```diff
@@ -85,17 +85,18 @@ async def _transmit_recordtree(
...
         await self._transmit_blob_record(
             ert.data.BlobRecord(data=json.dumps(data).encode("utf-8"))
         )
-        if self._real_id:
-            url = f"{self._uri}/userdata?realization_index=0"
-        else:
-            url = f"{self._uri}/userdata?"
-        await _put_to_server_async(url, {}, json={"record_type": record._record_type})
+        if self._real_id == 0:
+            await add_record_metadata(
```
berland, line 97: Should there be a comment here about why we treat realization 0 differentl? What about ensembles where realization 0 crashes, or an evaluation run where only a said interval of realizations are run?

jondequinor: Yes, definitely.

xjules: This'll change in the near future (see https://github.com/equinor/ert-storage/issues/195). Wrt. record metadata the current consensus is it to annotate `real_id==0` only (see `add_record_metadata`).

ert/storage/_storage.py
xjules: Fully agree there...

ert/storage/_storage.py
xjules: Added comment ...

xjules (approved):
I think it looks good now!

sondreso (approved):
Remember to squash before merging

PR #2692, opened 2022-01-11
Use lowercase letters when specifying psb resources

sondreso:
**Issue**

Using uppercase letters causes problem on PSB clusters in Azure.

## Pre review checklist


markusdregi (approved):
🌩️

PR #2702, opened 2022-01-12
Improve error messages in ert3 config validation

berland:
* Expose which .ert directory must be deleted for `ert3 init`
* Expose explicitly inconsistencies in stage vs ensemble records

## Pre review checklist

### DIFF
--- a/ert3/config/_experiment_run_config.py
+++ b/ert3/config/_experiment_run_config.py
@@ -194,6 +194,18 @@ class ExperimentRunConfig:
             input.record for input in self._ensemble_config.input
         )
         if ensemble_input_names != stage_input_names:
+            msg: str = ""
+
+            missing_in_ensemble = set(stage_input_names) - set(ensemble_input_names)
+            if missing_in_ensemble:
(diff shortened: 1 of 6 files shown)


ert3/config/_experiment_run_config.py
```diff
@@ -194,6 +194,18 @@ def _validate_stage(self) -> None:
...
+
+            missing_in_ensemble = set(stage_input_names) - set(ensemble_input_names)
+            if missing_in_ensemble:
+                msg += (
+                    f"\nMissing record names in ensemble input: {missing_in_ensemble}."
+                )
+
+            missing_in_stage = set(ensemble_input_names) - set(stage_input_names)
+            if missing_in_stage:
+                msg += f"\nMissing record names in stage input: {missing_in_stage}."
```
sondreso, line 207: I think it can be debated whether this means that the stage is missing an input or the ensemble has an input to much, but I think it's ok to leave it as is.

sondreso (approved):
LGTM!

Issue #1429, opened 2021-02-18
Gather resources in <workspace>/bin fails


markusdregi:
**Describe the bug**
When setting up the SPE1 example I wanted to gather all the scripts in `<workspace>/steps/bin` or something similar. However this fails with

```
(env) glados:polynomial mfane$ rm -rf .ert; ert3 init; ert3 run evaluation
An exception occurred while starting the ensemble evaluation
Traceback (most recent call last):
  File "/Users/mfane/Development/ert/ert_shared/ensemble_evaluator/prefect_ensemble/prefect_ensemble.py", line 288, in _evaluate
    self.run_flow(ee_id, dispatch_url, input_files)
  File "/Users/mfane/Development/ert/ert_shared/ensemble_evaluator/prefect_ensemble/prefect_ensemble.py", line 318, in run_flow
    flow = self.get_flow(ee_id, dispatch_url, input_files, realization_range)
  File "/Users/mfane/Development/ert/ert_shared/ensemble_evaluator/prefect_ensemble/prefect_ensemble.py", line 226, in get_flow
    + self.store_resources(step["resources"]),
  File "/Users/mfane/Development/ert/ert_shared/ensemble_evaluator/prefect_ensemble/prefect_ensemble.py", line 326, in store_resources
    stored_resources = [storage.store(res) for res in resources]
  File "/Users/mfane/Development/ert/ert_shared/ensemble_evaluator/prefect_ensemble/prefect_ensemble.py", line 326, in <listcomp>
    stored_resources = [storage.store(res) for res in resources]
  File "/Users/mfane/Development/ert/ert_shared/ensemble_evaluator/prefect_ensemble/storage_driver.py", line 27, in store
    shutil.copyfile(self._run_path / local_name, storage_uri)
  File "/usr/local/opt/python@3.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/shutil.py", line 121, in copyfile
    with open(dst, 'wb') as fdst:
FileNotFoundError: [Errno 2] No such file or directory: '/Users/mfane/Development/ert/examples/polynomial/.ert/tmp/evaluation/.my_storage/global/bin/poly.py'
Task exception was never retrieved
```

**To Reproduce**
1. `pip install ert`
2. `cd examples/polynomial; mkdir bin; mv poly.py bin/`
3. Replace `poly.py` with `bin/poly.py` in `stages.yml`
4. `ert3 init; ert3 run evaluation`

**Expected behavior**
[...]

PR #2703, opened 2022-01-12
Snake ert3

### LINKED ISSUE #2676: Port snake_oil to ert3
Port the snake_oil ert example to ert3.

berland:
**Issue**
Resolves #2676 

**Approach**
Mimic spe1 and polynomial ert3 examples.

## Pre review checklist


ert/data/record/_transmitter.py
```diff
@@ -192,6 +192,7 @@ async def transmit_record(self, record: Record) -> None:
         elif isinstance(record, BlobRecord):
             uri = await self._transmit_blob_record(record)
         elif isinstance(record, (NumericalRecordTree, BlobRecordTree)):
+            print(type(self))
```
berland, line 195: rm

ert3_examples/snake_oil/stages.yml
```diff
@@ -0,0 +1,55 @@
...
+
+  output:
+    -
+      record: summary
+      location: SNAKE_OIL_FIELD.SMSPEC
+      smry_keys: ["FOPT", "FOPR"]
+
+  transportable_commands:
+    - name: render
+      # These must be in root directory of ert workspace
```
markusdregi, line 45: This is documented here: https://github.com/equinor/ert/issues/1429

markusdregi:
I do indeed see the value of including the snake oil case as a light weight workspace example emulating a reservoar simulator! Hence, I think this is great 👏🏻 

A discussion I think is worth having is whether we should continue the ERT3 example with links into the ERT2 test example or just copy the data once and then let them diverge if it happens. I'm not entirely sure what the upside is of keeping them in sync. The negative part is that it does indeed make the example a bit harder to navigate...

berland:
Reverted the mime-types in ensemble.yml, following closing of #2722

ert3_examples/snake_oil/stages.yml
```diff
@@ -0,0 +1,55 @@
...
+      location: op2_offset.json
+    - record: bpr_555_persistence
+      location: bpr_555_persistence.json
+    - record: bpr_138_persistence
+      location: bpr_138_persistence.json
+
+    -
+      # Fixed seed file, templating does not work
+      record: seedtxt
+      is_directory: true  # Also works for files
```
berland, line 29: Do a double check to verify that this is really needed.

berland:
> A discussion I think is worth having is whether we should continue the ERT3 example with links into the ERT2 test example or just copy the data once and then let them diverge if it happens. I'm not entirely sure what the upside is of keeping them in sync. The negative part is that it does indeed make the example a bit harder to navigate...

I will copy the snaky simulator and tune it for the ERT3 application (e.g. remove time_map functionality)

ert3_examples/snake_oil/stages.yml
berland: Not needed in this particular case. To be removed.

.github/workflows/run_examples_snake_oil.yml
```diff
@@ -0,0 +1,69 @@
...
+ pull_request:
+
+jobs:
+  run-ert3-snakeoil:
+    timeout-minutes: 20
+    services:
+      postgres:
+        image: postgres:10.8
+        env:
+          POSTGRES_USER: postgres
```
xjules, line 17: Why do we need these env variables?

ert3_examples/polynomial/experiments/evaluation/ensemble.yml
```diff
@@ -4,6 +4,7 @@ input:
   -
     source: stochastic.coefficients
     record: coefficients
+    mime: application/json
```
xjules, line 7: ?

ert3_examples/snake_oil/README
```diff
@@ -0,0 +1,22 @@
+Introduction
+============
+
+Snake oil is a synthetic reservoir simulator, that creates
```
xjules, line 4: You can remove the comma.

ert3_examples/snake_oil/run_demo
```diff
@@ -0,0 +1,36 @@
...
+ert3 run evaluation
+ert3 export evaluation
+# Assert that we have FOPT in the output
+grep FOPT experiments/evaluation/data.json >/dev/null
+
+
+# DOE (design of experiment)
+design2params 0 resources/designmatrix.xlsx designsheet defaults  # [semeio]
+python designmatrixtxt2json.py  # [hack]
+ert3 record load BPR_138_PERSISTENCE resources/designed_bpr_138_persistence.json
```
xjules, line 19: I think we should provide Transformation that loads either multiple `NumericalRecord` / files (matching a certain filer eg. designed_*) or a single file with multiple records within into a single `RecordTree`...

.github/workflows/run_examples_snake_oil.yml
berland: This is a copy from the analogue polynomial github action setup. But running snake_oil also through postgres is probably not necessary, so it can be skipped for snake_oil.

ert3_examples/polynomial/experiments/evaluation/ensemble.yml
berland: Hmm, this should not have gotten in there.

ert3_examples/snake_oil/run_demo
```diff
@@ -0,0 +1,36 @@
...
+ert3 record load OP1_OCTAVES resources/designed_op1_octaves.json
+ert3 record load OP1_OFFSET resources/designed_op1_offset.json
+ert3 record load OP1_PERSISTENCE resources/designed_op1_persistence.json
+ert3 record load OP2_DIVERGENCE_SCALE resources/designed_op2_divergence_scale.json
+ert3 record load OP2_OCTAVES resources/designed_op2_octaves.json
+ert3 record load OP2_OFFSET resources/designed_op2_offset.json
+ert3 record load OP2_PERSISTENCE resources/designed_op2_persistence.json
+ert3 run doe
+ert3 export doe
+grep FOPT experiments/doe/data.json >/dev/null
```
xjules, line 31: I would then include test for FOPR too...

ert3_examples/snake_oil/run_demo
```diff
@@ -0,0 +1,36 @@
...
+
+# Let ERT draw and run a stochastic ensemble:
+ert3 run evaluation
+ert3 export evaluation
+# Assert that we have FOPT in the output
+grep FOPT experiments/evaluation/data.json >/dev/null
+
+
+# DOE (design of experiment)
+design2params 0 resources/designmatrix.xlsx designsheet defaults  # [semeio]
```
xjules, line 17: ok, so this actually handles the `xlrd` issue. Just confirming that it needs to be there :)

berland: Necessary for producing the json files yes :)

ert3_examples/snake_oil/run_demo
berland: Fixed.

ert3_examples/snake_oil/run_demo
berland: Agree, but it is a bigger scope than this PR.

xjules (approved):
Very good job Haavard! :rocket:

PR #2704, opened 2022-01-12
add poly_template and a test that might do something

mortalisk:


tests/performace_tests/performace_utils.py
```diff
@@ -0,0 +1,178 @@
+import numpy
```
berland, line 1: Directory and filename typo: performace -> performance

tests/performace_tests/test_poly_example_performace.py
```diff
@@ -0,0 +1,204 @@
+import os
+
+import json
+import pytest
+
+from performace_utils import make_poly_template, dark_storage_app
```
berland, line 6: ditto typo performace

mortalisk:
[poly.tar.gz](https://github.com/equinor/ert/files/8131904/poly.tar.gz)

Here is the folder with the project the test was running at the time, with log files, and storage, etc.

mortalisk:
Closing since I made a new PR in #3051

PR #2706, opened 2022-01-12
Pretty print json for human readable files

### LINKED ISSUE #318: Prettyprint status.json (whitespace)
I suggest pretty-printing the dumped `status.json` with whitespace, similar to how `jobs.json` is dumped.  
```
 $ wc status.json jobs.json 
    0   717  7852 status.json
 1288  3764 43923 jobs.json
```

berland:
**Issue**
Resolves #318 once more..

**Approach**
Add `indent=4` to occurences of `json.dump()` whenever it writes to a file that a human might read.

~~There are some occurences of `indent=1` already in the code base, not sure if consistency with that is important, or if we should try to be consistent with how `black` would format the same dictionaries in source code~~

Changed existing `indent=1` into `indent=4`.

Considered to add `sort_keys=True` also, but skipped it, might confuse.

Occurences of `json.dump()` in tests are not touched.

## Pre review checklist

### DIFF
--- a/ert/serialization/_serializer.py
+++ b/ert/serialization/_serializer.py
@@ -33,6 +33,7 @@ class _json_serializer(Serializer):
     def encode(
         self, obj: Any, *args: Any, indent: Optional[int] = 4, **kwargs: Any
     ) -> str:
+        # pylint: disable=arguments-differ
         return json.dumps(obj, *args, indent=indent, **kwargs)
 
     def decode(self, series: str, *args: Any, **kwargs: Any) -> Any:
--- a/ert/serialization/_serializer.py
+++ b/ert/serialization/_serializer.py
@@ -1,7 +1,7 @@
 import json
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import Any, Union
+from typing import Any, Optional, Union
 
 import aiofiles
 import yaml
@@ -30,10 +30,10 @@ class Serializer(ABC):
 
 
 class _json_serializer(Serializer):
-    def encode(self, obj: Any, *args: Any, **kwargs: Any) -> str:
-        if "indent" not in kwargs:
-            kwargs["indent"] = 4
-        return json.dumps(obj, *args, **kwargs)
+    def encode(
+        self, obj: Any, *args: Any, indent: Optional[int] = 4, **kwargs: Any
+    ) -> str:
+        return json.dumps(obj, *args, indent=indent, **kwargs)
 
     def decode(self, series: str, *args: Any, **kwargs: Any) -> Any:
         return json.loads(series, *args, **kwargs)
--- a/ert/serialization/_serializer.py
+++ b/ert/serialization/_serializer.py
@@ -31,6 +31,8 @@ class Serializer(ABC):
 
 class _json_serializer(Serializer):
     def encode(self, obj: Any, *args: Any, **kwargs: Any) -> str:
+        if "indent" not in kwargs:
+            kwargs["indent"] = 4
         return json.dumps(obj, *args, **kwargs)
 
     def decode(self, series: str, *args: Any, **kwargs: Any) -> Any:
@@ -40,7 +42,7 @@ class _json_serializer(Serializer):
         self, obj: Any, path: Union[str, Path], *args: Any, **kwargs: Any
     ) -> None:
         async with aiofiles.open(path, mode="wt", encoding="utf-8") as filehandle:
-            await filehandle.write(json.dumps(obj))
+            await filehandle.write(json.dumps(obj, indent=4))
 
     async def decode_from_path(
         self, path: Union[str, Path], *args: Any, **kwargs: Any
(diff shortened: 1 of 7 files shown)


markusdregi:
> There are some occurences of indent=1 already in the code base, not sure if consistency with that
is important

I would say that is not important and that you should feel free to bump those to `4` as well in this PR...

job_runner/reporting/file.py
```diff
@@ -198,4 +198,4 @@ def _dump_ok_file(self):
 
     def _dump_status_json(self):
         with open(self.STATUS_json, "w") as fp:
-            json.dump(self.status_dict, fp, indent=1)
+            json.dump(self.status_dict, fp, indent=4)
```
jondequinor, line 201: There's a home made JSON parser somewhere. Just make sure that this file either is not read by that parser, or if it does, that the parser can deal with this indentation. 😅

markusdregi: I was hoping for the test to detect that 😁

berland: 🙏🏻

markusdregi: I'm not even sure the C-layer reads this file anymore... 🤔 It definitively checks whether it exists (and removes it upon starting a forward model) to deduce running state, but I'm not sure it actually loads the file. It appears that this is all done in the Python layer... If so, there is no worries related to parsing it :)

ert/serialization/_serializer.py
```diff
@@ -31,6 +31,8 @@ async def decode_from_path(
 
 class _json_serializer(Serializer):
     def encode(self, obj: Any, *args: Any, **kwargs: Any) -> str:
+        if "indent" not in kwargs:
```
markusdregi, line 34: Is it possible to move this into the function declaration as an optional argument and then afterwards passed explicitly to `dumps`. It should yield the same behaviour and maybe look a bit less wacky :)

tests/ert_tests/ert3/data/test_serializer.py
```diff
@@ -16,14 +13,16 @@
...
     assert json_serializer.decode(obj_json) == obj
 
     # Test that keyword arguments are passed through
     assert json_serializer.decode(json_serializer.encode(obj, indent=True)) == obj
-    assert json_serializer.encode(obj, separators=(",", ":")) == obj_json.replace(
-        " ", ""
-    )
+
+    # Most compact json serialization:
+    assert json_serializer.encode(
```
markusdregi, line 23: An alternative method here would be to just drop all whitespaces both from `obj_json` and the result of `json_serializer`. It might be easier to read in the future and it removes the need to ever toggle `indent`.

ert/serialization/_serializer.py
berland: Yes, that worked! Looks better now.

tests/ert_tests/ert3/data/test_serializer.py
berland: I am not sure I got your question? There are multiple ways to reach a compact json serialization, one way is through the options to `encode()`. This test is basically testing that `separators` is passed on while we are overriding `indent`.

ert/serialization/_serializer.py
berland: Nope, `pylint` complained:
```
************* Module ert.serialization._serializer
ert/serialization/_serializer.py:33:4: W0221: Number of parameters was 4 in 'Serializer.encode' and is now 5 in overridden '_json_serializer.encode' method (arguments-differ)

```

So either go wacky, add pyllint exception, or add `indent` as a argument to all `encode()` implementations.

jondequinor:
An argument against this, is that raw JSON will use more space (a lot of ` ` and `\n`). Some of these indented JSON files will surely be large, and they will touch "hot" cloud storage, which is expensive.

Wouldn't `cat some.json|jsonpp` take care of the human readable part of this PR?

berland:
That is correct. `jq . < somefile.json | less` works on-prem, but is not as user-friendly as `less somefile.json`. The extra whitespace shuffled around might be acceptable until ert3 is mature enough for few users needing to debug in the terminal. 

Avoiding whitespace for speed reasons is possibly premature optimization.

jondequinor:
> That is correct. `jq . < somefile.json | less` works on-prem, but is not as user-friendly as `less somefile.json`. The extra whitespace shuffled around might be acceptable until ert3 is mature enough for few users needing to debug in the terminal.
> 

Fair enough.

> Avoiding whitespace for speed reasons is possibly premature optimization.

Well, we were avoiding whitespace _before_ this PR, but I get your point.

berland:
Setting `indent=1` as default may be a compromise though. `indent=4` does not improve readability by much compared to 1.

jondequinor:
No, the debugging argument is good, and 4 spaces is the sweet spot.

tests/ert_tests/ert3/data/test_serializer.py
markusdregi: Sorry for the late reply. I retract my point :)

markusdregi (approved):
There are good arguments both ways, but this is also easy to change at any point based on new data. I suggest we merge this as is now :)

PR #2681, opened 2022-01-10
Use `ies::config::config` class for configuration also of std enkf module

joakim-hove:
Another small step in the direction of unifying the std_enkf module and the ies module

~~This is on top of: https://github.com/equinor/ert/pull/2621~~

PR #2709, opened 2022-01-13
Use std variant

joakim-hove:
All the SVD routines can specify the number of singular values to retain in two different ways - either as an integer specifying the number of singular values directly, or as a floating point number specifying the amount variation explained.

With this PR this is implemented with `std::variant<double, int>`

On top of #2681

### DIFF
--- a/libres/lib/analysis/enkf_linalg.cpp
+++ b/libres/lib/analysis/enkf_linalg.cpp
@@ -92,43 +92,38 @@ static int enkf_linalg_num_significant(int num_singular_values,
     return num_significant;
 }
 
-int enkf_linalg_svdS(const matrix_type *S, double truncation, int ncomp,
+int enkf_linalg_svdS(const matrix_type *S,
+                     const std::variant<double, int> &truncation,
                      dgesvd_vector_enum store_V0T, double *inv_sig0,
                      matrix_type *U0, matrix_type *V0T) {
 
     double *sig0 = inv_sig0;
     int num_significant = 0;
 
-    if (((truncation > 0) && (ncomp < 0)) ||
-        ((truncation < 0) && (ncomp > 0))) {
-        int num_singular_values =
-            std::min(matrix_get_rows(S), matrix_get_columns(S));
-        {
-            matrix_type *workS = matrix_alloc_copy(S);
-            matrix_dgesvd(DGESVD_MIN_RETURN, store_V0T, workS, sig0, U0, V0T);
-            matrix_free(workS);
-        }
+    int num_singular_values =
+        std::min(matrix_get_rows(S), matrix_get_columns(S));
+    {
+        matrix_type *workS = matrix_alloc_copy(S);
+        matrix_dgesvd(DGESVD_MIN_RETURN, store_V0T, workS, sig0, U0, V0T);
+        matrix_free(workS);
+    }
 
-        if (ncomp > 0)
-            num_significant = ncomp;
-        else
-            num_significant = enkf_linalg_num_significant(num_singular_values,
-                                                          sig0, truncation);
+    if (std::holds_alternative<int>(truncation))
+        num_significant = std::get<int>(truncation);
+    else
+        num_significant = enkf_linalg_num_significant(
+            num_singular_values, sig0, std::get<double>(truncation));
 
-        {
-            int i;
-            /* Inverting the significant singular values */
-            for (i = 0; i < num_significant; i++)
-                inv_sig0[i] = 1.0 / sig0[i];
+    {
+        int i;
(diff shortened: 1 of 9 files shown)


joakim-hove:
Test ert please!

libres/lib/analysis/enkf_linalg.cpp
```diff
@@ -93,60 +93,55 @@ void enkf_linalg_genX2(matrix_type *X2, const matrix_type *S,
 /*This function is similar to enkf_linalg_svdS but it returns the eigen values without its inverse and also give the matrices truncated U VT and Sig0*/
 
 // Trunc.SVD(S)  = U0 * Sig0 * V0T
-int enkf_linalg_svd_truncation(const matrix_type *S, double truncation,
-                               int ncomp, dgesvd_vector_enum store_V0T,
-                               double *sig0, matrix_type *U0,
-                               matrix_type *V0T) {
+int enkf_linalg_svd_truncation(const matrix_type *S,
```
dafeda, line 96: We removed this function in the following commit:

https://github.com/equinor/ert/commit/ba43cfb51327e4a227b580c0cfb9753e7ad63662

libres/lib/analysis/enkf_linalg.cpp
```diff
@@ -93,60 +93,55 @@ void enkf_linalg_genX2(matrix_type *X2, const matrix_type *S,
...
+                               const std::variant<double, int> &truncation,
+                               dgesvd_vector_enum store_V0T, double *sig0,
+                               matrix_type *U0, matrix_type *V0T) {
 
     int num_significant = -1;
     int nrows = matrix_get_rows(S);
     int ncolumns = matrix_get_columns(S);
 
-    if (((truncation > 0) && (ncomp < 0)) ||
```
dafeda, line 105: Are we comfortable removing this check?

dafeda:
What is the main reason of replacing two input parameters with one `std::variant<double, int>`?

libres/lib/analysis/enkf_linalg.cpp
joakim-hove: I'll rebase

libres/lib/analysis/enkf_linalg.cpp
joakim-hove: Yes - the point about using `std::variant<>` is to enforce this from the compiler

dafeda: I see, bit I don't think there is anything stopping us from passing in a negative truncation.
Perhaps this is a non-issue?

dafeda:
> > What is the main reason of replacing two input parameters with one std::variant<double, int>?
> 
> You can specify the number of singular values to retain in two _mutually exclusive_ ways:
> 
> 1. Using a floating point truncation - and include singular values up to that fraction of variation explained.
> 2. Specify an integer number of singular values.
> 
> Before this PR this was handled by using
> 
> ```
> double truncation;
> int num_singular;
> ```
> 
> and then a system of using invalid (i.e. `-1`) values for the inactive alternative. When using `std::variant<double,int>`the compiler enforces for us that exactly one of the options is active.

That makes good sense, thanks.

That description would also make a great commit message.

libres/lib/analysis/enkf_linalg.cpp
joakim-hove: > I see, bit I don't think there is anything stopping us from passing in a negative truncation.

No - in principle I agree with that, but that should caught when the truncation is set.

dafeda: I agree.

libres/lib/include/ert/analysis/enkf_linalg.hpp
```diff
@@ -22,7 +19,13 @@ void enkf_linalg_init_sqrtX(matrix_type *X5, const matrix_type *S,
 void enkf_linalg_Cee(matrix_type *B, int nrens, const matrix_type *R,
                      const matrix_type *U0, const double *inv_sig0);
 
-int enkf_linalg_svdS(const matrix_type *S, double truncation, int ncomp,
+int enkf_linalg_svd_truncation(const matrix_type *S,
```
dafeda, line 22: Thing this can be removed.

dafeda (approved):
Please remember to squash commits, and do add a description in the commit message (in addition to the usual one-liner).

LGTM 👍

libres/lib/include/ert/analysis/enkf_linalg.hpp
joakim-hove: OK

joakim-hove:
test ert please

PR #2710, opened 2022-01-13
Replace ResLog with Python logging

pinkwah:
We've recently implemented Python logging in C++ via pybind. ResLog was the previous attempt at this.

This PR deprecates `LOG_LEVEL` and `LOG_FILE` ERT config keys as well.

TODO:
- [x] Deprecate `LOG_LEVEL`
- [x] Deprecate `LOG_FILE`
- [x] Generalise `logging.hpp` testing
- [x] Fix old_tests to use `ILogger` rather than `reslog`.


ManInFez:
This looks like a solid piece of work :+1: Would you not say that LOG keywords are not deprecated but ignored or not supported?

dafeda (approved):
Real nice 👍

PR #2712, opened 2022-01-13
Remove deprecated (py38) loop parameter

berland:
**Issue**
Resolves deprecation warnings
`...../ert/ert_shared/ensemble_evaluator/evaluator.py:49: DeprecationWarning: The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.
`  

**Approach**
Removed the argument at selected places in the source, and looking for test failures.

## Pre review checklist

### DIFF
--- a/ert_shared/ensemble_evaluator/evaluator.py
+++ b/ert_shared/ensemble_evaluator/evaluator.py
@@ -46,9 +46,7 @@ class EnsembleEvaluator:
         self._done = self._loop.create_future()
 
         self._clients: Set[WebSocketServerProtocol] = set()
-        self._dispatchers_connected: asyncio.Queue[None] = asyncio.Queue(
-            loop=self._loop
-        )
+        self._dispatchers_connected: asyncio.Queue[None] = asyncio.Queue()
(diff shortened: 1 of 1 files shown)


ert_shared/ensemble_evaluator/evaluator.py
```diff
@@ -46,9 +46,7 @@ def __init__(self, ensemble, config, iter_, ee_id: str = "0"):
         self._done = self._loop.create_future()
 
         self._clients: Set[WebSocketServerProtocol] = set()
-        self._dispatchers_connected: asyncio.Queue[None] = asyncio.Queue(
-            loop=self._loop
-        )
+        self._dispatchers_connected: asyncio.Queue[None] = asyncio.Queue()
```
sondreso, line 49: The problem here is that when you remove the loop here it uses the current event loop at the time when the EnsembleEvaluator is instantiated, however we want this to be the event loop from the thread that is running websocket.  So we might need to create all of these things inside that thread instead

sondreso:
Superseded by https://github.com/equinor/ert/pull/2834

PR #2722, opened 2022-01-17
Explicit mime type to avoid warning

berland:
**Issue**
Resolves warnings on mime inconsistencies when running ert3, example:

`Warning: Conflicting ensemble mime 'application/octet-stream' and stage mime 'application/json' for input 'coefficients'.`

**Approach**
Explicitly set the mime type in each experiments `ensemble.yml`. There is a hunch that this is a workaround, so maybe not something we want to leak into this file.

## Pre review checklist


sondreso (dismissed):
LGTM!

markusdregi:
With #2764 in mind I would suggest we abandon this PR...

PR #2723, opened 2022-01-17
Use ert port range for pbs driver

sondreso:
**Issue**
Use ert ports also for the obs driver

## Pre review checklist


markusdregi (approved):
I just tested this in Azure and it works smoothly 👌🏻

PR #2732, opened 2022-01-18
Make ert3 log to console when realisations complete

markusdregi:
**Approach**
Make `ert3.evaluator` print a message whenever an event with a completing realisation occurs.

The implementation is crude in at least two ways:
- it allows `ert3.evaluator` to print directly to console instead of propagating this as an event to an entity that knows how to present this and
- it inspects the event with no tooling or support code.

## Pre review checklist


markusdregi:
Question is: is this bearable until the monitoring milestone is initiated?

sondreso (approved):
LGTM!

PR #2737, opened 2022-01-19
Add RES_LIB_SUBMODULE to pybind11 code

pinkwah:
This is a macro that:
1. Creates a submodule (if needed)
2. Calls the init function with the module object

Registration of submodules can then be done implicitly by using
`RES_LIB_SUBMODULE` in a `.cpp` file without having to add it to
`lib/python/init.cpp`.

---

My goal was to use pybind11's [custom type casting](https://pybind11.readthedocs.io/en/stable/advanced/cast/custom.html) functionality to remove the entirety of `py_conversion::*` and the wrapper classes, but alas it wasn't possible with the current state of our codebase. Pybind11 uses C++'s `typeid`, which cannot be used with forward-declared types.

### DIFF
--- a/libres/lib/enkf/config_keys.cpp
+++ b/libres/lib/enkf/config_keys.cpp
@@ -2,7 +2,7 @@
 #include <ert/enkf/config_keys.hpp>
 #include <pybind11/pytypes.h>
 
-void init_config_keys(py::module_ m) {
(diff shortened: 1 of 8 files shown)


pinkwah:
Jenkins test this please

oyvindeide (approved):
Looks like a good solution, assuming the CI failure is a flaky test this looks good to me 👍

PR #2740, opened 2022-01-19
Allow inline shell script in stages

berland:
**Issue**
Resolves the need for making and transporting a wrapper script in order for more fancy commands in `stages.yml`, like e.g.

```yaml
script:
  - sh -c "cd rms/model; rms -project foo"
```

**Approach**
Change `str.split()` to `shlex.split()` which will not split inside quotations.

## Pre review checklist

### DIFF
--- a/tests/ert_tests/ert3/evaluator/test_evaluator.py
+++ b/tests/ert_tests/ert3/evaluator/test_evaluator.py
@@ -4,8 +4,8 @@ import pytest
 
 import ert
 import ert3
-from ert_shared.ensemble_evaluator.ensemble.builder import create_step_builder
 from ert_shared.async_utils import get_event_loop
+from ert_shared.ensemble_evaluator.ensemble.builder import create_step_builder
 
 TEST_PARAMETRIZATION = [
     ([(0, 0, 0)], [[0] * 10]),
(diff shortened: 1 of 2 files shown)


markusdregi:
This can easily be overused :) Also, it makes the entire script a command during monitoring. But I don't see why we should not support it...

However, this introduces support for something new and somewhat complex behaviour - although the change in code is small. That warrants a test that would be failing prior to this commit :)

berland:
There is an associated issue with this solution, the `PATH` manipulation done by https://github.com/equinor/ert/blob/main/ert_shared/ensemble_evaluator/entity/unix_step.py#L35-L37 is not inherited in the inlined script.

jondequinor:
Using your example, why shouldn't I declaratively be able to do this using an rms stage? 

Or, to put it another way, shouldn't an ambition of ert3, in terms of forward models, be to become more than a thin layer around (for the time being) Prefect?

berland:
The ERT2 forward model for RMS takes care of that itself yes (via a default argument that sets cwd), so for the case of RMS only, this should not be needed.

berland:
> There is an associated issue with this solution, the `PATH` manipulation done by https://github.com/equinor/ert/blob/main/ert_shared/ensemble_evaluator/entity/unix_step.py#L35-L37 is not inherited in the inlined script.

Should be fixable,  as  this works fine:
```
import subprocess

env = {"FOO": "BAR"}

subprocess.run(["sh", "-c", "echo $FOO"], universal_newlines=True, env=env)

```

markusdregi:
While I don't necessarily think that the RMS example here is a good reason for this PR, I think the unix step itself should present itself as a useful, yet bare-bone shell command possibility for implementing steps. The fact that you can monitor the list of commands if you write them one by one is neat functionality that makes migration from an ERT2 forward model into an ERT3 forward model much more pleasant.

I just used this PR to be able to run the line:

```bash
sh -c "echo '[0, 1, 2, 3]' > some_numbers.json"
```

which also is not business critical. But I also don't think ERT3 should stop me from doing it.

tests/ert_tests/ert3/evaluator/test_evaluator.py
```diff
@@ -159,3 +159,33 @@ def test_evaluator_function(
...
+        output=[],
+        transportable_commands=tuple(),
+    )
+    step_builder = create_step_builder().set_name("inline_script_test").set_type("unix")
+
+    ensemble = ert3.evaluator.build_ensemble(
+        stage=step, driver="local", ensemble_size=1, step_builder=step_builder
+    )
+
+    assert ensemble.get_reals()[0].get_steps()[0].get_jobs()[0].get_args() == tuple(
```
markusdregi, line 189: Perhaps we should verify that `name` also is set correctly? :)

tests/ert_tests/ert3/evaluator/test_evaluator.py
```diff
@@ -159,3 +159,33 @@ def test_evaluator_function(
...
     assert expected == evaluation_records
+
+
+@pytest.mark.parametrize(
+    "commandline, parsed_cmd_args",
+    [
+        ("echo inline", ["inline"]),
+        ("bash -c 'echo inline'", ["-c", "echo inline"]),
+        ('bash -c "echo inline"', ["-c", "echo inline"]),
+        ("bash -c echo inline", ["-c", "echo", "inline"]),  # (user error)
```
markusdregi, line 170: Perhaps what I used it for could serve as an even more complex test case? :)

```suggestion
        ("bash -c echo inline", ["-c", "echo", "inline"]),  # (user error)
        ("sh -c" + `"` + "echo '[0, 1, 2, 3]' > some_numbers.json" + `"`, [ "-c", "echo '[0, 1, 2, 3]' > some_numbers.json"]),
```

markusdregi (commented):
This looks good! I appreciate the test 😎 

Please address the comments and I'll approve 🚢

markusdregi (approved):
Looking good 😎

PR #2754, opened 2022-01-21
Describe SIMULATION_JOB

### LINKED ISSUE #2111: Documentation - SIMULATION_JOB
[SIMULATION_JOB](https://ert.readthedocs.io/en/latest/reference/configuration/keywords.html#simulation-job) is in the [List of Keywords](https://ert.readthedocs.io/en/latest/reference/configuration/keywords.html) defined as:

"Experimental alternative to FORWARD_MODEL."

In [reference/configuration/forward_model](https://ert.readthedocs.io/en/latest/reference/configuration/forward_model.html#the-simulation-job-alternative) we write that:

"The difference between FORWARD_MODEL and SIMULATION_JOB is in how commandline arguments are passed to the final executable."

I've got two questions.

1. Is it still experimental?

2. When and why should I use one over the other?

dafeda:
**Issue**
Resolves #2111

## Pre review checklist

### DIFF
--- a/libres/lib/enkf/config_keys.cpp
+++ b/libres/lib/enkf/config_keys.cpp
@@ -101,6 +101,10 @@ RES_LIB_SUBMODULE("config_keys", m) {
     cls.attr("SEED") = "SEED";
     cls.attr("SETENV") = SETENV_KEY;
     cls.attr("SIMULATION") = "SIMULATION";
+    // SIMULATION_JOB is a lightweight version of FORWARD_MODEL that allows passing
+    // raw command line arguments to executable.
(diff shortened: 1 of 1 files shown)


markusdregi (approved):
You are the best to evaluate whether this would have helped you when you started looking into this :)

The documentation is correct, relevant and in a decent file so I have no problems merging this...

dafeda:
> You are the best to evaluate whether this would have helped you when you started looking into this :)

I think it would have been helpful as the first thing I did was search for `SIMULATION_JOB` in my IDE.

PR #2755, opened 2022-01-21
Implement ERT3 test run functionality

markusdregi:
**Issue**
To make it reasonable to set up a unix step this PR aims at implementing support for running a single realisation in a locally available directory.

I would like to support something along the lines of:

```bash
ert3 run evaluation --local-test-run
```

That would only work if:
- the experiment is an evaluation and
- all input records are fetched from `parameters.yml`.

The reason being is that any extension beyond this would involve a more complicated implementation and the above would already provide quite some help.

The consequence of using the flag `--local-test-run` would be that the unix step is executed in `<workspace>/local-test-run` with the ensemble size being set to `1` and queue system being forced to `local`.

**Approach**
Let the `UnixStep` be able to take the run path as an argument and utilise it. Do the above changes to the configuration if `--local-test-run` is enabled.

**Future work**
Beyond this PR it would be nice to support:
- input records from storage,
- specifying which realisation to run.

The above should be persisted in an issue before this PR is merged.

## Pre review checklist


ert_shared/ensemble_evaluator/ensemble/builder.py
```diff
@@ -687,19 +691,35 @@ def build(self):
...
+        else:
+            return _Step(
+                stage.get_id(),
+                stage.get_inputs(),
+                stage.get_outputs(),
+                jobs,
+                stage.get_name(),
+                source,
+            )
+            # raise AssertionError(f"Unexpected step type: {self._type}")
```
markusdregi, line 722: The fact that this makes tests fail and that replicating the logic from prior to this PR where a plain `_Step` is returned if it is neither `function` nor `unix` makes me a bit worried... Is this intentional behaviour?

ert3/console/_console.py
```diff
@@ -210,16 +216,51 @@ def _init(args: Any) -> None:
...
+        sys.exit(
+            "Local test runs are only supported for evaluation experiments, was:"
+            str(experiment_run_config.experiment_config.type) 
+        )
+
+    raw_ensemble_config = original_experiment_run_config.ensemble_config.dict()
+    raw_ensemble_config["size"] = 1
+    raw_ensemble_config["forward_model"]["driver"] = "local"
+
+    for input_ in raw_ensemble_config["input"]:
```
markusdregi, line 230: This is a consequence of #2764

markusdregi: Question is: is it an acceptable hack until the issue is picked up?

ert3/engine/_run.py
```diff
@@ -165,6 +166,10 @@ def run(
         .set_type("function" if isinstance(stage, ert3.config.Function) else "unix")
     )
 
+    if local_test_run:
+        local_runpath = Path(workspace._path) / "local-test-run"
```
markusdregi, line 170: This is misusing the workspace. A way forward could be to extend the workspace such that I could ask for an export folder and then use it 🤷🏻

markusdregi:
I would like a discussion around this PR before I continue. Debugging is definitively a challenge when running without shared scratch disks and this is a first, but big step in aiding at debugging without compromising the change in mentality induced by the changed infrastructure.

I believe the implementation should work and you can test it on examples using this PR and the `--local-test-run` option.

The PR lacks tests completely, which obviously needs to be addressed if this POC is to be converted into something that can be considered for merging. In addition, there is a question of who should be responsible for the concept of a local test run. Currently the `console` translates what it means into the configuration and `engine` translates it into the specific steps, which is a bit peculiar. But to do it differently we would need to decide a specific entity that should hold this. One solution would be to introduce the concept of a local test run in the configuration and then potentially also accept it as a console argument. This would make it possible for the engine to once again build the ensemble entirely from configuration 🤷🏻 

Besides this, there are a few other discussions lifted in various comments. Input appreciated 🙏🏻

markusdregi:
Jenkins: Retest this please

markusdregi:
Jenkins: Retest this please

sondreso:
> all input records are fetched from `parameters.yml`.

Is the reason for this limitation that its difficult to handle records that have a size set by the original ensemble size?

ert_shared/ensemble_evaluator/ensemble/builder.py
sondreso: Is there a log from this test run, or was it made locally?

markusdregi:
> Is the reason for this limitation that its difficult to handle records that have a size set by the original ensemble size?

Yes, they are the only input data for which size is allocated dynamically depending on the ensemble size. Data stored in storage already have a fixed size and currently the `engine` will complain if the size of the ensemble and the size of the `RecordEnsemble` are of different size. I'm sceptical of removing this logic without adding some logic for mapping data to the different realisations or filtering active/inactive realisations. Just silently grabbing the first `x` records and hope they where input to the ensemble of size `x` is not a good solution in my head. But, doing the mapping or filtering is a slightly bigger endeavour and I would hope that the features of the PR would be useful only supporting stochastic parameters and resources :)

ert_shared/ensemble_evaluator/ensemble/builder.py
markusdregi: It was reproduced both locally and in CI. I'll bring it into a separate PR 👍🏻

sondreso:
I agree, but it's also limiting the usefulness of this quite a bit :grimacing:  If I understand this correctly, there would be no problem using uniform records from storage? Could we not approach this in a optimistic way, where we try to run it, and rather catch the Exception that the engine throws and then print the current error message? That would at least allow a few more input types like resources and storage (where we have a uniform record)

markusdregi:
Ah, resources should be allowed. Apparently that commit was only on my laptop 🙈 I'll patch it ASAP. `stochastic.` and `resources.` should work always. Then, as you say we could occasionally also support records from storage. The tradeoff is that the error message of attempting and then forwarding the message of that exception might become a bit more unreadable depending on where the responsibility for handling the local test run logic should be. If we decide it to be in the engine we could even forward a private argument that reduces the record ensemble to only be one element inside the engine and hence always work. But, that would not be possible if we want to put the responsibility elsewhere... So, it is a tradeoff between functionality, technical dept and scope of the work - as so many time 😂

sondreso:
Ahh, well if resources are allowed I think this is a perfectly fine approach, and we can leave the storage limitation for later :sweat_smile: Also I created this: https://github.com/equinor/ert/issues/2794

sondreso:
I think this is a valuable addition. If the PR is rebased and the test fixed, I will start reviewing it :medal_sports:

ert3/console/_console.py
```diff
@@ -36,6 +36,13 @@ def _build_init_argparser(subparsers: Any) -> None:
 def _build_run_argparser(subparsers: Any) -> None:
     run_parser = subparsers.add_parser("run", help="Run experiment")
     run_parser.add_argument("experiment_name", help="Name of the experiment")
+    run_parser.add_argument(
+        "--local-test-run",
+        default=False,
+        action="store_true",
+        help="Will evaluate a single realization locally in the "
+        "folder `<workspace root>/local-test-run`.",
```
sondreso, line 44: I think we should mention the numbers that will be added to this path as well here

ert3/console/_console.py
```diff
@@ -232,17 +239,65 @@ def _init(args: Any) -> None:
...
+            "Local test runs are only supported for evaluation experiments, was:\n"
+            f"{experiment_run_config.experiment_config.type}"
+        )
+
+    for input_record in experiment_run_config.ensemble_config.input:
+        if not (
+            input_record.source.startswith("stochastic.")
+            or input_record.source.startswith("resources.")
+        ):
+            sys.exit(
```
sondreso, line 256: Think we should raise an more specific ErtError here (and the same above), and let the calling code handle the exiting. Possibly a new error type is needed. That would also allow for easier testing of these special cases, which would be nice to have included :slightly_smiling_face:

ert_shared/ensemble_evaluator/ensemble/builder.py
```diff
@@ -704,6 +708,7 @@ def build(self):
                 jobs,
                 stage.get_name(),
                 source,
+                run_path=self._run_path,
```
sondreso, line 711: Can we avoid forwarding the runpath to the function step, since you should not assume that you have a file system available when running it?

ert3/console/_console.py
berland: Since this relates to #3016 I suggest NotImplementedError.

ert3/console/_console.py
```diff
@@ -232,17 +239,65 @@ def _init(args: Any) -> None:
...
+            workspace,
+            args.experiment_name,
+            local_test_run=args.local_test_run,
+        )
     elif experiment_run_config.experiment_config.type == "sensitivity":
         ert3.engine.run_sensitivity_analysis(
             experiment_run_config, workspace, args.experiment_name
         )
+    else:
+        raise AssertionError(
```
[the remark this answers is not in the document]
berland: This is apparently already part of the validation, so this `else:` can just be removed.

ert_shared/ensemble_evaluator/ensemble/builder.py
berland: (?)  The run-path is only forwared to `_UnixStep`, not to `_FunctionStep`.

sondreso: Ahh, Github confused me in the way it showed the diff, my bad!

sondreso (approved):
Very nice!

berland:
Issue with the current PR:

```
$ ert3 run evaluation --local-test-run
Warning: Conflicting ensemble mime 'application/octet-stream' and stage mime 'application/json' for input 'coefficients'.
Realization 0 completed successfully (1/1)
$ ert3 run evaluation
Experiment 'evaluation' has been carried out already.
To re-run, first clean the experiment: ert3 clean evaluation
```

An experiment should not be considered "carried out" after a local test run.

berland:
test please

berland:
> An experiment should not be considered "carried out" after a local test run.

This will be filed as a bug after merging this PR.

ert3/workspace/_workspace.py
```diff
@@ -227,6 +227,34 @@ def _validate_resources(self, ensemble_config: ert3.config.EnsembleConfig) -> No
...
+
+        Args:
+            basename: This will be part of returned directory name, with
+                an index appended.
+            runid: An id that will always be included in the suggested runpath
+
+        Returns:
+            Non-existing path
+        """
+        suggestion = Path(self._path.absolute() / (basename + f"-{run_id}"))
```
xjules, line 246: Why not just `basename + f"-str(uuid.uuid4())"`, instead of checking every +1 ?

berland: I think 6 digit hex looks less scary for users compared to `uuid4()`. This is used to make directories that users should `cd` into.

jondequinor:
```python
>>> uuid.uuid4().hex
'fd1d4c1f25b944749b11480f36622dfe'
```

jondequinor: What about we print out a user manual for inspecting the output? That manual would include 

```bash
$ cd local-test-run-fd1d4c1f25b944749b11480f36622dfe

# You are now free to look around and inspect files…
```

jondequinor: Use `uuid` just because its more conventional for unique resources.

berland: ok!

tests/ert_tests/ert3/console/integration/test_cli.py
```diff
@@ -1,16 +1,18 @@
...
+import os
 import pathlib
 import sys
-import os
 from unittest.mock import MagicMock, patch
 
 import pytest
 import yaml
+from ert_utils import chdir
+from sklearn.metrics import jaccard_score
```
sondreso, line 11: Probably not needed? :sweat_smile:

sondreso (approved):
LGTM! Very nice work on picking this up and getting it over the finish line! :checkered_flag:

Issue #2650, opened 2022-01-04
Investigate the use of threadpool loading data


oysteoh:
Issue:

Using threadpool to parallelize loading of realizations with Summary data
For record requests for Summary data, approx 25-40% of time is spent in the function EnsemblePlotData:
https://github.com/equinor/ert/blob/main/res/enkf/export/summary_collector.py#L61

which in turn calls enkf_plot_data_load in libres
https://github.com/equinor/ert/blob/main/libres/lib/enkf/enkf_plot_data.cpp#L108

When loading the data a thread pool is used to load realizations in parallel, but if this code is removed and replaced by serial loading of realizations then the total time is greatly reduced by 90%. This then leads to a total reduction in record request time by between 25-50%

BjarneHerland:
Using drogon-data provided by @TerryHannant  and a C++ driver I can confirm the essence of this issue (which for reference originates from https://github.com/equinor/webviz-ert/issues/197).

The point of using data from drogon is to have enough summary-data to measure load-times properly - I'm not aware of a dataset in our test-toolbox which contains this. The particular dataset contains 142 summary-keys over 553 (active / relevant) realisations, causing the test to load 79520 elements. Not sure why number of elements is `142*560` and not `142*553` - it sums size of all returned `enkf_plot_data_type` - but it indicates the amount of data loaded.

In order to remove (potentially) unnecessary overhead of the legacy thread/threadpool it was replaced with C++ futures from `<async>`. These can launch in three modes: `std::launch::async` schedules jobs immediately (multiple threads), `std::launch::deferred` runs jobs on-demand (essentially loads on the calling thread when looping to collect results), and combination of both modes which leaves scheduling to the system.

The numbers below are from a small number of benchmark-runs for each case, approximated to something close to the average. Note that they are subject to several factors, most notably the speed and load on my laptop when benchmarking, but also the effect of caching ERT-files in memory (see the two footnotes).

Average load-time (ms) per key (approximated over a couple of runs)
```
                                   Native       Python    Overhead
Threadpool, 4 threads (original) :  41(*)       245       204
std::async, launch::deferred     :   6          173       167
std::async, launch::async        : 112          290       178          
std::async, system decides       : 110(**)      290       180
No concurrency                   :   5          171       166

(*) Also saw ~101ms at some point, but could not reproduce this
(**) Also saw ~35ms at some point, but could not reproduce this
```

[...]

PR #2757, opened 2022-01-23
Replaced reuse_addr with something more insisting

### LINKED ISSUE #2728: Remove reuse_addr parameter from port_handler
After some investigation (Thanks to @BjarneHerland and @mortalisk ) it seems like the `reuse_addr=true` in the `port_handler.py ` is the cause of the flakiness in the `ert3/engine/integration/test_engine` tests.

When asked for a socket the port handler returns a socket which the storage server is not able to use causing an exception. If we set `reuse_addr=false` the port handler seems to only return usable sockets. 

It might be a good idea to revisit the code where this parameter is used and examine the consequences of removing it.

BjarneHerland:
**Issue**
Resolves #2728

**Approach**
Renamed ™reuse_addr" to "will_close_then_reopen_socket" to make users of this flag conscious of why it is there.

Added documentation with reasoning and cleaned up some rather clunky logic in find_available_port(). Cleaned up unnecessary use of the flag in tests and code.

## Pre review checklist


BjarneHerland:
In particular I would appreciate input on the approach and docs.

frode-aarstad:
Looks good. The comment is informative and helps explain the different issues with sockets in the implementation. 
I guess its not easy to create a test testing the fix since its mostly seen on clusters.

ert_shared/ensemble_evaluator/ensemble/prefect.py
```diff
@@ -57,14 +57,15 @@ async def _eq_submit_job(self, script_filename):
...
 def _get_executor(custom_port_range, name="local"):
-    _, port, _ = find_available_port(custom_range=custom_port_range, reuse_addr=True)
     if name == "local":
         cluster_kwargs = {
             "silence_logs": "debug",
-            "scheduler_options": {"port": port},
         }
         return LocalDaskExecutor(**cluster_kwargs)
     elif name == "lsf":
+        _, port, _ = find_available_port(
```
oysteoh, line 66: Is this needed @sondreso ? From what i remember when debugging the port issues - the executor running lsf and pbs should not need a port? And - available port on the node running this code is not equivalent to an available port on a node running on the cluster?  Or am i just thinking and remembering wrong here?

ert_shared/ensemble_evaluator/ensemble/prefect.py
```diff
@@ -57,14 +57,15 @@ async def _eq_submit_job(self, script_filename):
 
 
 def _get_executor(custom_port_range, name="local"):
-    _, port, _ = find_available_port(custom_range=custom_port_range, reuse_addr=True)
     if name == "local":
         cluster_kwargs = {
             "silence_logs": "debug",
```
oysteoh, line 62: what is the reason of the removal of the port when running locally?

sondreso: It does (I think), but when running locally we are not restricted bu infrastructure firewall :slightly_smiling_face:

BjarneHerland: Ahh - ok. So it's free to pick whatever port id wants then, which it does when we do not provide a port-number.

BjarneHerland:
Following up on @frode-aarstad 's challenge 😄  I started to write tests for this and discovered quite a lot of funky behaviour.

Hopefully these tests with docs adequately describes current behaviour and provides users of this module some insight into what to expect...

frode-aarstad:
> Following up on @frode-aarstad 's challenge 😄 I started to write tests for this and discovered quite a lot of funky behaviour.
> 
> Hopefully these tests with docs adequately describes current behaviour and provides users of this module some insight into what to expect...

Looks very good

ert_shared/ensemble_evaluator/ensemble/prefect.py
```diff
@@ -57,14 +57,15 @@ async def _eq_submit_job(self, script_filename):
...
-            "scheduler_options": {"port": port},
         }
         return LocalDaskExecutor(**cluster_kwargs)
     elif name == "lsf":
+        _, port, _ = find_available_port(
+            custom_range=custom_port_range, will_close_then_reopen_socket=true
```
oysteoh, line 67: true -> True

ert_shared/ensemble_evaluator/ensemble/prefect.py
```diff
@@ -82,6 +83,9 @@ def _get_executor(custom_port_range, name="local"):
             debug=True,
         )
     elif name == "pbs":
+        _, port, _ = find_available_port(
+            custom_range=custom_port_range, will_close_then_reopen_socket=true
```
oysteoh, line 87: true -> True

ert_shared/ensemble_evaluator/ensemble/prefect.py
BjarneHerland: Hehe... :}

oysteoh:
Looks very good - also good with more tests verifying the behavior! 👏

oysteoh (dismissed):

joakim-hove:
@BjarneHerland : I tried to find an issue/PR with questions of thread-pool; could not find now. Easiest to reach me on email joakim.hove AT opm-op.com Might be a bit tight today. Cool if you post a link to the issue/PR here. Thank you.

BjarneHerland:
test ert please

BjarneHerland:
test ctest please

oysteoh:
> @BjarneHerland : I tried to find an issue/PR with questions of thread-pool; could not find now. Easiest to reach me on email joakim.hove AT opm-op.com Might be a bit tight today. Cool if you post a link to the issue/PR here. Thank you.

@joakim-hove the issue you are looking for is located here https://github.com/equinor/ert/issues/2650

BjarneHerland:
test this please

BjarneHerland:
test ert please

BjarneHerland:
@oysteoh , @sondreso  mind taking a look? I hope to merge in the (narrow) window between green tests and need for rebase...  😄

oysteoh (approved):
Very good! Nice with good documented testing verifying the behavior of this! 🚀

ert_shared/ensemble_evaluator/ensemble/prefect.py
```diff
@@ -57,14 +57,13 @@ async def _eq_submit_job(self, script_filename):
...
 def _get_executor(custom_port_range, name="local"):
-    _, port, _ = find_available_port(custom_range=custom_port_range, reuse_addr=True)
     if name == "local":
         cluster_kwargs = {
             "silence_logs": "debug",
-            "scheduler_options": {"port": port},
         }
         return LocalDaskExecutor(**cluster_kwargs)
     elif name == "lsf":
+        _, port, _ = find_available_port(custom_range=custom_port_range)
```
sondreso, line 66: I thought the reason for having the `will_close_then_reopen_socket`, was that it was needed here?

sondreso (commented):
LGTM! Only had a small question

ert_shared/ensemble_evaluator/ensemble/prefect.py
BjarneHerland: I thought so too initially but while writing the test-suite I figured out that the desired behaviour can be obtained without the flag. But keep reading please...

Setting the flag corresponds to pattern in the test named `test_reuse_passive_close_ok_ok`, i.e. obtain the initial port/socket with flag=true, then close socket without using it. Observe that we get the same behaviour when **not** setting the flag, demonstrated in `test_def_passive_live_nok_nok_close_ok_ok` (read its executive summary).

However, when writing this commnent, I realize that both tests explicitly close the socket while ERT relies on that to happen automatically. According to [Python documentation](https://docs.python.org/3.8/library/socket.html#socket-objects) :

> Sockets are automatically closed when they are garbage-collected, but it is
> recommended to close() them explicitly, or to use a with statement around them.

Hence, this code

` _, port, _ = find_available_port(custom_range=custom_port_range)
`

relies on GC kicking in **before** Dask tries to bind the port, which probably not happens from time to time. Moreover, according to the test `test_reuse_passive_live_linux_nok_ok`, if the socket is not closed (GCd in our case) Dask will fail to bind the port on Linux (unless it sets SO_REUSEADDR, which it doesn't). IMO this is a likely source of flakiness.

So I'll probably just replace ERT code with

```
 _, port, s = find_available_port(custom_range=custom_port_range)
s.close()
```

following the pattern in `test_def_passive_live_nok_nok_close_ok_ok` and getting the behaviour we want by (deterministically) closing the socket.

_**Note that nothing above reserves the port until it is bound by Dask, which is an entirely different issue. Any other process may hijack the port between us closing the socket and Dask re-binding the port. Fixing this requires Dask to accept an already bound socket, a port-range in which to find the port, or alternatively a socket-factory or similar.**_

Thoughts?

frode-aarstad: Regarding gc. Might be worthwhile making a test(s) that explicitly invokes the gc to verify the desired behavior.

sondreso: > following the pattern in test_def_passive_live_nok_nok_close_ok_ok and getting the behaviour we want by (deterministically) closing the socket.

Will this allow us to rebind the socket immediately? :thinking:

BjarneHerland: > Will this allow us to rebind to the same port immediately? 🤔

AFICS, yes. Please refer to testcode (pointing out any mistakes, of course  😄  )

BjarneHerland: > Regarding gc. Might be worthwhile making a test(s) that explicitly invokes the gc to verify the desired behavior.

I'm inclined to trust Python on this, but I'll give it a shot.

BjarneHerland: Mmmm...  that was an interesting exercise... 😬 

Looks like GC is not the crucial part but rather refcounting. I added `test_gc_closes_socket` which may or may not provide insight or value. Didn't figure out how to mock close() because that method belongs to the superclass `_socket.socket` which is a built-in or extension type, apparently.

sondreso: Alright, then I am fine with merging this!

BjarneHerland: Since I don't have the overview of use of Prefect - could you point out which test(s) are most likely to be affected by this change?

BjarneHerland:
If no other questions or concerns, I'll just merge this one. Would be interesting, though, to identify one or two tests crucial in this context.

PR #2772, opened 2022-01-25
Refactor row scaling class

pinkwah:
Use pybind11 rather than cwrap for the Python and remove the C functions
while refactoring row scaling C++ a bit.

### DIFF
--- a/libres/lib/analysis/update.cpp
+++ b/libres/lib/analysis/update.cpp
@@ -341,7 +341,7 @@ void save_parameters(enkf_fs_type *target_fs,
 load a set of parameters from a enkf_fs_type storage into a set of
 matrices with the corresponding row-scaling object.
 */
-std::vector<std::pair<matrix_type *, const row_scaling_type *>>
+std::vector<std::pair<matrix_type *, std::shared_ptr<RowScaling>>>
 load_row_scaling_parameters(enkf_fs_type *target_fs,
                             ensemble_config_type *ensemble_config,
                             int_vector_type *iens_active_index, int last_step,
@@ -350,7 +350,8 @@ load_row_scaling_parameters(enkf_fs_type *target_fs,
 
     int matrix_start_size = 250000;
 
-    std::vector<std::pair<matrix_type *, const row_scaling_type *>> parameters;
+    std::vector<std::pair<matrix_type *, std::shared_ptr<RowScaling>>>
+        parameters;
 
     const auto &scaled_keys = ministep->scaled_keys();
     if (scaled_keys.size() > 0) {
@@ -374,13 +375,10 @@ load_row_scaling_parameters(enkf_fs_type *target_fs,
                                    column, active_list, A);
                 }
             }
-            const row_scaling_type *row_scaling =
-                ministep->get_row_scaling(key);
+            auto row_scaling = ministep->get_row_scaling(key);
 
             matrix_shrink_header(A, row_scaling->size(), matrix_get_columns(A));
-            parameters.push_back(
-                std::pair<matrix_type *, const row_scaling_type *>(
-                    matrix_alloc_copy(A), row_scaling));
+            parameters.emplace_back(matrix_alloc_copy(A), row_scaling);
         }
         matrix_free(A);
     }
@@ -394,8 +392,8 @@ Store a set of row-scaled parameters into a enkf_fs_type storage
 void save_row_scaling_parameters(
     enkf_fs_type *target_fs, ensemble_config_type *ensemble_config,
     int_vector_type *iens_active_index, const local_ministep_type *ministep,
-    std::vector<std::pair<matrix_type *, const row_scaling_type *>>
-        row_scaling_list) {
+    const std::vector<std::pair<matrix_type *, std::shared_ptr<RowScaling>>>
+        &row_scaling_list) {
 
     assert(row_scaling_list.size() > 0);
     const auto &scaled_keys = ministep->scaled_keys();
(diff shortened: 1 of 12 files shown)


libres/lib/include/ert/enkf/row_scaling.hpp
```diff
@@ -1,67 +1,21 @@
...
-  (at your option) any later version.
-
-  ERT is distributed in the hope that it will be useful, but WITHOUT ANY
-  WARRANTY; without even the implied warranty of MERCHANTABILITY or
-  FITNESS FOR A PARTICULAR PURPOSE.
-
-  See the GNU General Public License at <http://www.gnu.org/licenses/gpl.html>
-  for more details.
-*/
+#pragma once
```
dafeda, line 1: Most of our code uses header guards.
Should we use pragma once instead?
My knowledge of the difference between the two is based entirely on the first paragraph of pragma onces' wiki page!

libres/lib/include/ert/enkf/row_scaling.hpp
```diff
@@ -1,67 +1,21 @@
...
+#include <ert/res_util/matrix.hpp>
+#include <vector>
+#include <memory>
 
-#ifndef ROW_SCALING_H
-#define ROW_SCALING_H
+class RowScaling : public std::enable_shared_from_this<RowScaling> {
+    size_t m_resolution = 1000;
```
dafeda, line 7: Do we know what `m_resolution` is and why it's 1000?

libres/lib/include/ert/enkf/row_scaling.hpp
pinkwah: Yeah I think it's the less confusing of the two. Consider if you're not careful and have two `.hpp` with the same name you might end up with the same header guard.

It'll be in writing at some point.

dafeda: From wiki:

"On the other hand, #pragma once is not necessarily available in all compilers and its implementation is tricky and might not always be reliable."

Is this something to worry about?

libres/lib/include/ert/enkf/row_scaling.hpp
pinkwah: @joakim-hove

libres/lib/analysis/update.cpp
```diff
@@ -341,7 +341,7 @@ void save_parameters(enkf_fs_type *target_fs,
 load a set of parameters from a enkf_fs_type storage into a set of
 matrices with the corresponding row-scaling object.
 */
-std::vector<std::pair<matrix_type *, const row_scaling_type *>>
+std::vector<std::pair<matrix_type *, const RowScaling *>>
```
dafeda, line 344: Should we use shared pointers here instead of `const RowScaling *`?

libres/lib/enkf/row_scaling.cpp
```diff
@@ -15,19 +15,18 @@
    See the GNU General Public License at <http://www.gnu.org/licenses/gpl.html>
    for more details.
 */
+#include "ert/python.hpp"
```
dafeda, line 18: Consider moving to where the other stuff from `ert/` is included, and also consider using `<ert/python.hpp>`.

libres/lib/enkf/row_scaling.cpp
```diff
@@ -183,54 +182,67 @@ void row_scaling::multiply(matrix_type *A, const matrix_type *X0) const {
...
+    # Load scaling vector from grdecl file; typically created with
+    # geomodelling software.
+    with open("scaling.grdecl") as fh:
+        kw_global = EclKW.read_grdecl(fh, "SCALING")
 
-int row_scaling_get_size(const row_scaling_type *row_scaling) {
-    return row_scaling->size();
-}
+    # Create a ecl_kw copy with only the active elements.
+    kw_active = grid.compressed_kw_copy(kw_global)
```
dafeda, line 222: What is `grid` in this example?

libres/old_tests/enkf/test_row_scaling.cpp
```diff
@@ -28,22 +28,18 @@
...
 
-    row_scaling_iset(row_scaling, 9, 0.25);
-    test_assert_double_equal(row_scaling_iget(row_scaling, 9), 0.25);
-
-    row_scaling_free(row_scaling);
+    row_scaling.assign(9, 0.25);
+    test_assert_double_equal(row_scaling[9], 0.25);
 }
 
 static void scaleX(matrix_type *X, const matrix_type *X0, double alpha) {
```
dafeda, line 45: Any reason to re-implement `scaleX` in the test?
It is already implemented in `row_scaling.cpp`.

libres/lib/include/ert/enkf/row_scaling.hpp
pinkwah: It is available in all mainstream compilers and thus is both already implemented and reliable.

This is in reference to any of the hundreds of random half-baked C++ compilers that chip manufacturers make for their platforms.

pinkwah: As in, the portability concerns that the C++ standard has, the compiler and we have are 3 entirely different beasts.

The C++ standard has to consider every single possible architecture, real or imagined (eg. there is hardware where allocating memory is not a thing, therefore large parts of the C++ standard library are designed to work without allocating memory. Another example is `NULL` not being the integer value `0` on some architectures, which is one of the reasons that C++11 introduced `nullptr`).

The compiler has to support a bajillion different operating systems and CPU architectures, including really outdated ones (eg. modern GCC can produce assembly compatible with the Intel 80368, first manufactured in 1985).

We only have the following: `(Linux + macOS) x (x86_64 + aarch64)`.

libres/lib/enkf/row_scaling.cpp
pinkwah: Probably `EclGrid`: https://github.com/equinor/ecl/blob/main/python/ecl/grid/ecl_grid.py

libres/old_tests/enkf/test_row_scaling.cpp
pinkwah: Nope, but it's outside the scope of this PR

libres/lib/include/ert/enkf/row_scaling.hpp
dafeda: Thanks for the nice write-up!

dafeda (approved):
LGTM 👍 

Nice if we can get some info about the `m_resolution = 1000`,  but I think we can merge this and then add that in later.

PR #2779, opened 2022-01-26
Improve error handling around ert3 initialization

### LINKED ISSUE #2760: Running ert3 experiments gives error message about __ensemble_records__
**Describe the bug**
ert3 run evaluation sometimes gives error message

`Experiment /ert/polynomial.__ensemble_records__ does not exist`

**To Reproduce**
TBD

**Expected behavior**
TBD

**Enviromment**
 - OS: RHEL7
 - ERT Release: main
 - Python version: 3.6
 - Remote/HPC execution involved: yes and no (Related ti Azure testing)

verveerpj:
**Issue**
Resolves #2760

**Approach**
- Add error handling for the case that `ert3 init` fails to connect to the storage.
- Add error handling for the case that ert3 tries to use a storage that not has been initialized properly.

## Pre review checklist

### DIFF
--- a/ert/storage/_storage.py
+++ b/ert/storage/_storage.py
@@ -544,6 +544,17 @@ def init(*, workspace_name: str) -> None:
         )
 
 
+def assert_storage_initialized(workspace_name: str) -> None:
+    response = _get_from_server(path="experiments")
+    experiment_names = {exp["name"]: exp["ensemble_ids"] for exp in response.json()}
+
+    for special_key in _SPECIAL_KEYS:
+        if f"{workspace_name}.{special_key}" not in experiment_names:
+            raise ert.exceptions.StorageError(
+                "Storage is not initialized properly. The workspace needs to be reinitialized"
+            )
+
+
 def init_experiment(
     *,
     experiment_name: str,
--- a/ert3/console/_console.py
+++ b/ert3/console/_console.py
@@ -187,7 +187,6 @@ def _init(args: Any) -> None:
 
     if args.example is None:
         workspace = ert3.workspace.initialize(pathlib.Path.cwd())
-        ert.storage.init(workspace_name=workspace.name)
     else:
         example_name = args.example
         pkg_examples_path = _get_ert3_examples_path()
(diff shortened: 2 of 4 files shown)


ert/storage/_storage.py
```diff
@@ -544,6 +544,15 @@ def init(*, workspace_name: str) -> None:
         )
 
 
+def assert_storage_initialized(workspace_name: str) -> None:
+    response = _get_from_server(path="experiments")
+    experiment_names = {exp["name"]: exp["ensemble_ids"] for exp in response.json()}
+
+    for special_key in _SPECIAL_KEYS:
+        if f"{workspace_name}.{special_key}" not in experiment_names:
+            raise ert.exceptions.StorageError("Storage has not been initialized")
```
sondreso, line 553: Can we give a hint as to how the user can solve this? Maybe something like: "Storage has not been initialized with this workspace. To fix this, the workspace need to be initialized again"

ert3/console/_console.py
```diff
@@ -220,12 +219,20 @@ def _init(args: Any) -> None:
             )
 
         workspace = ert3.workspace.initialize(wd_example_path)
+
+    try:
         ert.storage.init(workspace_name=workspace.name)
+    except TimeoutError as err:
+        workspace.delete()
+        raise ert.exceptions.StorageError(
+            "Failed to contact storage. Is it initialized?"
```
sondreso, line 228:
```suggestion
            "Failed to contact storage. Is it running?"
```

sondreso (commented):
Looks good! Only two minor comments

ert/storage/_storage.py
verveerpj: I have update the message. For this particular case, where `ert3 init` failed, the user should not see it anymore. If he/she does see it, then `ert3 init` was fine, but something else is broken with the storage and it needs to be re-initialized. That is currently not possible, since that requires re-running `ert3 init`, which will fail since the `.ert` directory will be present. The solution is currently to remove the `.ert` directory and re-run `ert3 init`. So that needs some thought at some point.

sondreso (approved):
Nice work!

PR #2780, opened 2022-01-26
Backport pins

sondreso:
**Issue**
Backport pinning of dependencies 

## Pre review checklist


sondreso:
Mypy fails because this is not back ported: https://github.com/equinor/ert/commit/6ca5c9892ec36a73a74bb6e5066cd2711eccd452

I dont think we should backport this feature (I have also pinned the ert-storage dependency before this was introduced), and vote for using admin privileges to get it merged. Rest of the release pipeline is not dependent on mypy passing.

markusdregi (approved):
Looks good :)

PR #2783, opened 2022-01-26
Stop using a Client instance over long time spans

### LINKED ISSUE #2756: Long-running unix step will kill fail due to websocket ping timeout
**Describe the bug**
Long-running unix steps will likely make the `ee_client` used to send messages before and after tasks run—die. It will appear as though sending something failed abnormally. What likely happened is that pings from the client to the evaluator server were not sent, and the evaluator server summarily closed the connection. We don't see this exception since we do not capture WS logs.

**To Reproduce**
Run a long-running task in a unix step

**Expected behavior**
`run_until_complete` should ideally be called once, really early in `run` such that all subsequent subroutines are `async`.

**Enviromment**
 - OS: RHEL7
 - ERT/Komodo Release: main
 - Python version: any
 - Remote/HPC execution involved: n/a

jondequinor:
**Issue**
Resolves #2756 

**Approach**
If Client's loop doesn't run in <20 second intervals, the WS connection
will time out and #2756 will happen for unix steps running longer than
20 seconds. Hence we make all Client usage single-use.

## Pre review checklist

### DIFF
--- a/ert_shared/ensemble_evaluator/entity/unix_step.py
+++ b/ert_shared/ensemble_evaluator/entity/unix_step.py
@@ -1,5 +1,5 @@
 import asyncio
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 from pathlib import Path
 import subprocess
 import tempfile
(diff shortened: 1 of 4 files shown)


ert_shared/ensemble_evaluator/entity/unix_step.py
```diff
@@ -48,27 +48,36 @@ def run_job(self, client: Client, job: Any, run_path: Path):
...
-                ev_data={ids.ERROR_MSG: cmd_exec.stderr},
-            )
+            with Client(
+                prefect.context.url, prefect.context.token, prefect.context.cert
+            ) as client:
+                client.send_event(
+                    ev_type=ids.EVTYPE_FM_JOB_FAILURE,
+                    ev_source=job.get_source(self._ee_id),
+                    ev_data={ids.ERROR_MSG: cmd_exec.stderr},
+                )
```
sondreso, line 58: We could wrap this in a send_event function that takes the type, source and data, since we reuse this piece of code a bit now. But I am also fine with leaving it as is

sondreso (approved):
LGTM! Only a small comment that I am fine with ignoring

sondreso (approved):
LGTM!

PR #2807, opened 2022-01-28
Do not build vanilla step on unexpected step type

markusdregi:
During my work in #2755 I discovered that the step builder will build a vanilla step if the type is neither `unix` nor `function`. I think it should fail instead...

## Pre review checklist

### DIFF
--- a/ert_shared/ensemble_evaluator/ensemble/builder.py
+++ b/ert_shared/ensemble_evaluator/ensemble/builder.py
@@ -687,19 +687,26 @@ class _StepBuilder(_StageBuilder):
                 self._job_name,
                 self._run_arg,
             )
-        cls = _Step
         if self._type == "function":
-            cls = _FunctionStep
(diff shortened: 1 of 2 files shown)


ert_shared/ensemble_evaluator/ensemble/builder.py
```diff
@@ -687,19 +687,26 @@ def build(self):
...
+            return _UnixStep(
+                stage.get_id(),
+                stage.get_inputs(),
+                stage.get_outputs(),
+                jobs,
+                stage.get_name(),
+                source,
+            )
+        else:
+            raise AssertionError("Unexpected type while building step: {self._type}")
```
jondequinor, line 709: I don't think assertionerror is what we want here. Value or type error instead?

jondequinor (approved):
Lgtm! And I totally agree with the premise.

PR #2812, opened 2022-01-31
Black 22.1.0

markusdregi:
Continuation of #2811 where all files are reformatted with `black 22.1.0`.

## Pre review checklist


sondreso:
You need to reformat all python files in the repository, easiest is just to run `black .` from the root.

markusdregi:
Pinning black for now instead as this bump makes it impossible to reformat the code in a 3.6 environment. The consequences of that is not a Monday morning debate 😁 

Pinning in: #2811

PR #2815, opened 2022-01-31
Add upper limit for ERT Storage dependency

pinkwah:
We need to allow breaking changes to happen in ERT Storage.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -79,7 +79,7 @@ setup(
         "deprecation",
         "dnspython >= 2",
         "ecl >= 2.12.0",
-        "ert-storage >= 0.3.7",
+        "ert-storage >= 0.3.7, < 0.4",
         "fastapi==0.70.1",
         "graphene",
         "graphlib_backport; python_version < '3.9'",


markusdregi:
I'm a bit sceptical towards this 🤔 Could you explain the "need to allow breaking changes"?

pinkwah:
ERT Storage is currently pre-stable API which means that breaking changes should be expected. Because ERT depends on the _newest_ version of ERT Storage, it makes it very difficult to make any large changes. Currently, the only possible workflow is to:

1. Deprecate current unwanted API
2. Introduce new API
3. Update ERT to use it
4. Update Webviz-ERT to use it
4. Remove deprecated API

In a lot of cases, deprecating the current API is a lot more work than writing new API. That is, a simple rename leads to writing and testing functions that serve only as steps between API versions so as to not impede ERT development.

Furthermore, now that dark storage is part of ERT, releasing a new version of ERT Storage that changes the API is any way will suddenly make ERT CI fail for everyone due to the tests in https://github.com/equinor/ert/blob/main/tests/ert_tests/dark_storage/test_api_compatibility.py .

I think we've put ourselves in a situation where it's impossible to develop ERT Storage for no reason other than bureaucracy and blind adherence to running the latest version of every package.

This change would make it possible to:
1. Replace old API with new one immediately
2. Create a new PR in ERT at your own leisure, bumping the ERT Storage to the next minor version (while major is 0)
3. Continue appreciating life

markusdregi:
> I think we've put ourselves in a situation where it's impossible to develop ERT Storage for no reason other than bureaucracy and blind adherence to running the latest version of every package.

While I do to some extent agree with your consequence analysis I think you're off on the reasoning 🤷🏻 

To me there are a multitude of interesting components here, the ERT data API, the ert-storage implementation of the API, the dark storage implementation of the API, ERT2 core, ERT3 core and webviz-ert. The first is the definition of the API, the next two are implementations of the API and the last three are consumers of the API.

We'll have to accept that the API will be in development for quite some time (read: indefinitely, but probably at slower pace in the future). And to do so successfully we depend upon;
- being able to introduce changes in the API without too large cascading effects and
- being able to introduce changes in the API and cascade that effect to the consumer motivating that change.

The first allows you to introduce changes in the API and implement the backend without adopting all the consumers simultaneously. The second allows you to start consume the wished for change in the API. Both of these are important for development speed. And the standard solution for this is to version the API. Until we do this, we will be in some development pain. As long as we believe that pain is worth the saved work in developing the API according to industry standards; we do so. But we will have to version it at some point and development pain is a decent indicator for when this should be addressed 🤷🏻

If we start pinning the versions of ert-storage I think we move further away from how the development workflow should be (deprecation is closer to versioning the API). It will require persistence to always ensure that the breaking changes are ASAP introduced in both implementations and all three consumers. Because until that happens, others are blocked to introduce and propagate their changes (unless they also make the adoption to the other breaking changes).

This discussion also becomes a bit special from an ert-storage perspective because currently the API resides within ert-storage, together with the ert-storage implementation. But really, the API itself has grown beyond that local ownership and must be considered a shared definition between all the 5 implementations and consumers.

We can temporarily go down the road this PR heads for, but I think we should discuss it in the team first. Because it will require additional efforts to always adopt changes in following PR's as quickly as possible in the `ert` and `webviz-ert` repository.

jondequinor:
Some fortune cookie wisdom from me: CI is supposed to break. Releases and Komodo are our exports where things are expected to work …

pinkwah:
@markusdregi: It seems to me that you are fundamentally misunderstanding the purpose of `ert-storage` being pre-stable and assuming that it must come together with the expectations one would have from a stable API. Deprecating parts of the API comes with a non-trivial development burden that is larger than the development burden of the feature itself. Having a pre-stable API version was supposed to help development speed but it seems like we've disregarded this in favour of assuming that version 0 is actually a secret version 1.

A part of your job is to help us minimise scope creep and make sure people don't impede each other's work, and this is the opposite of that.

@jondequinor: You want a CI that will break if anything goes wrong, not a CI that breaks even though you haven't done anything wrong. The latter is my biggest complaint when I come to work.

Each release of `ert-storage` passes its CI and therefore as a project it is consistent and good. However, with unpinned `ert-storage` in `ert`, a tested `ert` release that is supposed to be stable is not guaranteed to work. If we publish a new `ert-storage`, `pip install ert` will be broken until we fix it in `ert`. That is, by _not_ pinning, we are publishing non-functional software onto PyPI.

Komodo stable is going to be stable by virtue of us not releasing until it works. Until the hypothetical situation exists where our projects are in disrepair, it's a non-issue.

markusdregi:
> A person's freedom ends where another man's freedom begins.

... and all that...

I don't find this discussion to be constructive in written text anymore. Lets have a chat on video instead 👍🏻

PR #2820, opened 2022-02-01
Retrieve responses for _export via the dedicated endpoint

### LINKED ISSUE #2715: Crash when ert3 is exporting a blob record with tar-contents
**Describe the bug**
Crash when exporting a record of directory type (tar-blob).

```
  File "pandas/_libs/parsers.pyx", line 1917, in pandas._libs.parsers.raise_parser_error
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 3256: invalid start byte
```

**To Reproduce**
1. Add an output record to `stages.yml`:
```
  output:
    -   
      record: runpathdir
      location: .
      is_directory: true
```

that intends to store the entire filesystem at the time of a finished run. 

2. `ert3 run evaluation`
3. `ert3 export evaluation`

**Expected behavior**

No crash. Potentially the blob record  dumped to `experients/evaluation/runpathdir.tar` but that is possibly a feature request.

**Screenshots**

**Enviromment**
 - OS: Ubuntu 21.10
 - ERT release: 2.32.+++
 - Python 3.9

**Additional context**
The blob record in storage can be `curl`ed down and extracted using `tar`.

xjules:
**Issue**
Resolves #2715 

**Approach**
Instead of using `ensemble.output` as a source to response names when exporting experiment we can use directly the dedicated endpoint (`ens_id/responses`) which filters out automatically only `NumericalRecord` types records.

## Pre review checklist

### DIFF
--- a/ert/storage/_storage.py
+++ b/ert/storage/_storage.py
@@ -710,11 +710,18 @@ def get_experiment_responses(*, experiment_name: str) -> Iterable[str]:
         )
 
     ensemble_id = experiment["ensemble_ids"][0]  # currently just one ens per exp
-    response = _get_from_server(f"ensembles/{ensemble_id}")
+    response = _get_from_server(f"ensembles/{ensemble_id}/responses")
 
     if response.status_code != 200:
         raise ert.exceptions.StorageError(response.text)
-    return list(response.json()["response_names"])
+
+    # The ensemble responses are sent in the following form:
+    #     {
+    #         "polynomial_output": {"id": id, "name": name, "userdata": {}}
+    #     }
+    # therefore we extract only the keys
+
+    return list(response.json().keys())
 
 
 def delete_experiment(*, experiment_name: str) -> None:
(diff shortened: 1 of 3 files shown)


ert/storage/_storage.py
```diff
@@ -710,11 +710,11 @@ def get_experiment_responses(*, experiment_name: str) -> Iterable[str]:
         )
 
     ensemble_id = experiment["ensemble_ids"][0]  # currently just one ens per exp
-    response = _get_from_server(f"ensembles/{ensemble_id}")
+    response = _get_from_server(f"ensembles/{ensemble_id}/responses")
 
     if response.status_code != 200:
         raise ert.exceptions.StorageError(response.text)
-    return list(response.json()["response_names"])
+    return list(response.json().keys())
```
sondreso, line 717: What are the values here that we are just discarding?

sondreso (commented):
I think we should include an export test where we have both numerical and blob records, to show/test the new behavior

ert/storage/_storage.py
xjules: I can enrich it via a comment ....

sondreso: That is very nice! I was just wondering and making sure that we didn't download the actual data unnecessary

sondreso (approved):
LGTM

PR #2828, opened 2022-02-02
Introduce nested namespace ies::data

joakim-hove:
Initial step on: https://github.com/equinor/ert/pull/2826

The first commit is a breaking change in the user config - the settings
```
ANALYSIS_SET_VAR  STD_ENKF USE_EE  xxxx
ANALYSIS_SET_VAR  STD_ENKF USE_GE  xxxx
```
will now be ignored with a warning. Personally my hunch is that this is not used at all. The same functionality can be achieved with
```
ANALYSIS_SET_VAR STD_ENKF INVERSION  EXACT | SUBSPACE_EXACT_R | SUBSPACE_EE_R | SUBSPACE_RE
```

The class `ies::data_struct` will become *the* state holding datastructure for analysis. In this PR it is moved into a nested namespace for clearer code.

The two commits are unrelated, and the PR can be split in two.


dafeda (approved):
LGTM 👍

PR #2829, opened 2022-02-02
Remove function pointers

joakim-hove:
On top of: #2828 

With this PR the `analysis_table` structure with function pointers is completely removed. I consider this ready for review and merge. I tried to make a sensible commit history - but every N'th commit is e.g. a `clang-format` or similar clean up; I can make an effort to beautify the git history.

Although I consider this a unit ready for review and merging I do not consider the refactoring process as a whole as complete. Topics which are still on my list (which I or someone else can do):

1. Use proper C++ for the `ies::` structures.
2. In the current PR  the `analysis_module` class is retained, but most of the interesting functions are one-liners calling directly down to a corresponding `ies::` function. The reason for doing it like this is twofold: 
    a) The user configuration in the config file is unchanged.
    b) The `update.cpp` implementation can (for now) be unchanged (this is not problematic - just avoiding some churn).
But it might be desirable to also rip out `analysis_module.cpp` and address points a) and b) above.
3. The current `analysis_module` which can call either `ies::initX()` and do Ensemble Smoother, or `ies::updateA()` and do (parts) of an iterated smoother - now has some half-baked policy (i.e. how should we run) in itself - that should be brought further up to a run model.
4. I would like to move the `ies::data::data_type` instance from `analysis_module`to the local_set.
5. The Python / C++ interface has not received any love.
6. ....
 
Update: I am happy to split, reorganize commits or other related tasks to simplify the review process.


dafeda:
Great work!
@pinkwah and I have read through and agree. 👍 

One small suggestion is to rename the enums to `ENSEMBLE_SMOOTHER` and `ITERATED_ENSEMBLE_SMOOTHER` for readability purposes. 

```python
class AnalysisModeEnum(BaseCEnum):
    TYPE_NAME = "analysis_mode_enum"
    ES = None
    ITERABLE = None
```

dafeda:
I think this should be a single commit, so consider squashing before merging.

joakim-hove:
> One small suggestion is to rename the enums to ENSEMBLE_SMOOTHER and ITERATED_ENSEMBLE_SMOOTHER for readability purposes.

Done

dafeda (approved):
LGTM 👍

PR #2830, opened 2022-02-02
Free internal resources in ies::data_free()

joakim-hove:
This drive by commit fixes a bad resource leak.


joakim-hove:
The Equinor test failure????

pinkwah (approved):

PR #2832, opened 2022-02-03
Increase worker memory requirements on PBS

sondreso:
**Issue**
Related to #2831

**Approach**
Increase memory on compute node to handle forward models with RMS and Eclipse

## Pre review checklist


lars-petter-hauge (approved):
Tested with a relatively small case, until a more flexible solution is in place this should be sufficient

sondreso:
Jenkins test this please

PR #2836, opened 2022-02-03
[Feature branch] API adoption

markusdregi:
This is a feature branch for the work to adopt all current solutions providing data to the end user to do so via the data API.

The reason this is kept in a feature branch for now is the current performance issues related to the API.


DanSava:
To me, this seems to revert the revert in main once merged

sondreso:
Can we close this PR, and open when we are ready to merge it?

PR #2848, opened 2022-02-04
Speed up snake_oil_diff

mortalisk:
By using EclSum.numpy_vector() the speed of fetching data is
greatly improved, which makes snake_oil run a lot faster.


sondreso (approved):
LGTM!

PR #2850, opened 2022-02-04
Add TabularDataTransformation

### LINKED ISSUE #2847: Transformation that converts design matrix into records
When working on doe (design of experiment) it is currently quite cumbersome to read the parameters from the design matrix file (eg. by means of an external job) and then create parameter json representation (one by one) in order to load them later on explicitly as records. Therefore it would be meaningful to have a transformation that reads such a design matrix csv file and does the parameter records automatically.

xjules:
**Issue**
Resolves #2847 

**Approach**
Add a new transformation with the goal to create `RecordCollection(records=Tuple[NumericalRecord, ...])` by parsing a give design matrix file. 
Here, we introduce a new transformation parameter `columns` that specifies which columns from the csv /design matrix file are to be exported as a `NumericalRecord`.

## Pre review checklist


ert3_examples/snake_oil/stages.yml
```diff
@@ -49,3 +49,49 @@
...
+        - BPR_555_PERSISTENCE
+        - OP1_DIVERGENCE_SCALE
+        - OP1_OCTAVES
+        - OP1_OFFSET
+        - OP1_PERSISTENCE
+        - OP2_DIVERGENCE_SCALE
+        - OP2_OCTAVES
+        - OP2_OFFSET
+        - OP2_OCTAVES
+      location: resources/designmatrix.txt
```
berland, line 70: Let the filename have a `csv`-extension?

ert/data/record/_transformation.py
```diff
@@ -295,3 +296,31 @@ def _sync_eclsum_transform_output(
...
+        if not columns:
+            raise ValueError("columns must be non-empty")
+        self._columns = columns
+
+    async def transform_output_sequence(
+        self,
+        mime: str,
+        location: Path,
+    ) -> Tuple[Record, ...]:
+        dframe = pd.read_csv(location, sep=" ")
```
berland, line 319: Ah, it is not CSV files we are parsing, with `sep=" "`. I think we should parse CSV files as in `pd.read_csv()` without any options, and rather apply a hack upfront to deal with (convert to csv) the space-separated-values file from `semeio`.

xjules: Agree, I've used sep=" " as I was using the file yielded by the `semeio/design2params` job.

ert3_examples/snake_oil/experiments/doe/ensemble.yml
```diff
@@ -2,41 +2,25 @@ size: 10
...
+        - OP2_OFFSET
+        - OP2_OCTAVES
+    record: params
   -
     source: resources.seedtxt
     record: seedtxt
   -
-    source: resources.snake_oil_txt_tmpl
-    record: snake_oil_params_txt_tmpl
+    source: resources.snake_oil_txt_tmpl_params
```
berland, line 22: No need to change filename, just replace `snake_oil_txt_tmpl`

ert3_examples/snake_oil/experiments/doe/ensemble.yml
```diff
@@ -46,4 +30,4 @@ output:
 
 forward_model:
   driver: local
-  stage: run_simulator
+  stage: run_simulator_dm
```
berland, line 33: I think we should be bold in such a PR and replace the existing stage and thereby say that this is the way to do it by only providing one stage.

ert/data/__init__.py
```diff
@@ -20,6 +20,7 @@
     RecordTransformation,
     RecordTreeTransformation,
     TarRecordTransformation,
+    TabularData2NumericalRecord,
```
berland, line 23: `TabularDataTransformation`?

ert/data/__init__.py
```diff
@@ -50,6 +51,7 @@
     "FileRecordTransformation",
     "TarRecordTransformation",
     "ExecutableRecordTransformation",
+    "TabularData2NumericalRecord",
```
berland, line 54: (I would use the opportunity to sort this list)

ert/data/record/_record.py
```diff
@@ -496,8 +503,9 @@ async def load_collection_from_file(
             length=length,
             collection_type=RecordCollectionType.UNIFORM,
         )
-    return RecordCollection(
-        records=await ert.data.FileRecordTransformation().transform_output_sequence(
-            mime, file_path
+    else:
```
berland, line 506: Is this triggering a pylint warning ala uneccessary else after return?

ert3_examples/snake_oil/experiments/doe/ensemble.yml
xjules: Right, I thought that this could show different approaches but you might be right.

ert/data/record/_record.py
xjules: This needs to be fine-tuned since `xlsx` gets `mime-type=application/json` :shrug:
When using it directly in python:
```
In [2]: mimetypes.guess_type("resources/designmatrix.xlsx")
Out[2]: ('application/vnd.openxmlformats-officedocument.spreadsheetml.sheet', None)
```
So it sort-of belongs to the else branch.

xjules: The issue is that we still depend on having a file serializer (which currently support only json and yaml) :
```python
guess = mimetypes.guess_type(str(values.get(basis, "")))[0]
    if guess:
        if not ert.serialization.has_serializer(guess):
            return DEFAULT_RECORD_MIME_TYPE
```
Are we to remove serializers and replacing them with transformation?

jondequinor: The answer here is we're keeping the serializers, but they've been demoted to an implementation detail in the serializationtransformation…

sondreso:
Should we close this PR, and rather pick it up again after we have landed the concepts for input grouping and transformation of ensemble wide inputs? @xjules

xjules:
> Should we close this PR, and rather pick it up again after we have landed the concepts for input grouping and transformation of ensemble wide inputs? @xjules

Agree.

PR #2853, opened 2022-02-07
Cleanup redundant parameters

sregales-TNO:
**Approach**
Remove and replace redundant parameters.

### DIFF
--- a/libres/lib/include/ert/enkf/enkf_main.hpp
+++ b/libres/lib/include/ert/enkf/enkf_main.hpp
@@ -79,7 +79,7 @@ bool enkf_main_smoother_update(enkf_main_type *enkf_main,
 void enkf_main_create_run_path(enkf_main_type *enkf_main,
                                const ert_run_context_type *run_context);
 
-enkf_main_type *enkf_main_alloc(const res_config_type *, bool, bool);
+enkf_main_type *enkf_main_alloc(const res_config_type *, bool = false);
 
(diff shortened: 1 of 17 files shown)


libres/lib/include/ert/enkf/enkf_main.hpp
```diff
@@ -79,7 +79,7 @@ bool enkf_main_smoother_update(enkf_main_type *enkf_main,
 void enkf_main_create_run_path(enkf_main_type *enkf_main,
                                const ert_run_context_type *run_context);
 
-enkf_main_type *enkf_main_alloc(const res_config_type *, bool, bool);
+enkf_main_type *enkf_main_alloc(const res_config_type *, bool = false);
```
dafeda, line 82: Do we want to set default to false both in header and cpp?

dafeda, line 82: Do we want to set default to false both in header and cpp?

sregales-TNO: I have realize that most calls to the function always have it as false. So I thought to keep it false until explicitly told to be true. I know that is always better to be explicit than implicit, but what is the value of the if almost all calls are `false`

sregales-TNO: plus is not set as defualt on the exposed endpoint on python

dafeda (approved):
LGTM 👍

PR #2854, opened 2022-02-07
Add types and perform clean-up of BaseRunModel

dafeda:


sondreso:
Think this looks good, any reason why it is still a draft?

dafeda:
> Think this looks good, any reason why it is still a draft?

Had a chat with @oyvindeide before putting in review, good to go now.

oyvindeide (approved):
LGTM

PR #2855, opened 2022-02-07
Pass ensemble size when creating ies::ies_data

joakim-hove:
**Approach**
The `ies::ies_data` had some internal state which was mandatory, but maintained by using a `xxx_set()` call late in the process, the lifetime was schematically like this:

```C++

auto * data = ies::data::alloc( void );
...
...
for (;;) {
    ies::data::set_xxx(data, arg);
    ...
    ies::data::use_xxx(data);
}
```
with this PR the ensemble size is passed to the construction function, thereby removing a need for `ies::data::set_xxx()` calls in the `ENSEMBLE_SMOOTHER` codepath. This again should be a step towards a proper C++ beautification of `struct ies::data::data_type`.

**Update**: This PR inadvertently stumbled upon undefined behavior in a test. An attempt to "solve" that here: https://github.com/equinor/ert/pull/2881


libres/lib/analysis/ies/ies_data.cpp
```diff
@@ -40,18 +41,20 @@ struct ies::data::data_struct {
...
+    data->ens_size = ens_size;
     data->ens_mask = NULL;
     data->obs_mask0 = NULL;
     data->obs_mask = NULL;
-    data->W = NULL;
     data->A0 = NULL;
     data->E = NULL;
     data->converged = false;
     data->config = ies::config::alloc(ies_mode);
+    data->W = matrix_alloc(data->ens_size, data->ens_size);
```
dafeda, line 56: This PR looks good.
I think it makes sense to allocate `W` here and it would be nice if the final commit message mentions it.

dafeda (approved):
LGTM 👍

libres/lib/analysis/ies/ies_data.cpp
joakim-hove: Well spotted - that minor change was actually of some significance; I will try to pull it out in a separate commit.

PR #2857, opened 2022-02-07
Analysis refactor clean up

joakim-hove:
**Approach**
Two independent commits:
1. For some time now the `std_enkf_initx()` routine has internally called the `ies::initX()` routine. With #2829 all calls to `std_enkf_initx()`were removed, and with this PR the (remains of) the implementation itself is removed.
2. When analysis was invoked through function pointers the state information was passed as `void *` - that is no longer necessary and with the second commit we make sure to use proper `ies::data::data_type *` instead of `void *`.


dafeda (approved):
Very nice 👍

PR #2862, opened 2022-02-08
Format change: make sure eeach initializer is on separate line

joakim-hove:
A minor change to `.clang-format` to ensure that initializer lists are not packed:

**Current master**
```C++
    local_ministep_type(const char *name, analysis_module_type *analysis_module)
        : name(strdup(name)), analysis_module(analysis_module),
          obs_data(nullptr), active_size(hash_alloc()) {
```

**This PR:**
```C++
    local_ministep_type(const char *name, analysis_module_type *analysis_module)
        : name(strdup(name))
        , analysis_module(analysis_module)
        , obs_data(nullptr)
        , active_size(hash_alloc()) {
```


joakim-hove:
Cool if this can be reviewed & merged - or closed ....

dafeda (approved):
I like this, but good if @pinkwah takes a look as well.

pinkwah:
While I agree that this change makes the code look better, we don't want to open the pandora's box of starting changing `.clang-format` defaults. That is an endless discussion that is not productive.

In most cases that are provided in the PR, a better change would've been to move initialisation to the class definition. Ie:

```cpp
struct local_ministep_type {
    obs_data_type *obs_data = nullptr;
    std::unordered_map<...> active_size;  // implicit ctor
};
```
or into the constructor body.

Closing.

PR #2863, opened 2022-02-08
Use proper C++ for ies classes

joakim-hove:
Use proper C++ for `ies::data::Data` and `ies::config::Config`

### DIFF
--- a/libres/lib/analysis/analysis_module.cpp
+++ b/libres/lib/analysis/analysis_module.cpp
@@ -150,19 +150,18 @@ void analysis_module_init_update(analysis_module_type *module,
 static bool analysis_module_set_int(analysis_module_type *module,
                                     const char *flag, int value) {
 
-    auto *ies_config = ies::data::get_config(module->module_data);
+    auto &ies_config = ies::data::get_config(module->module_data);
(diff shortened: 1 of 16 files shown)


libres/lib/analysis/analysis_module.cpp
```diff
@@ -42,7 +42,7 @@ auto logger = ert::get_logger("analysis");
 
 struct analysis_module_struct {
     UTIL_TYPE_ID_DECLARATION;
-    ies::data::data_type *module_data;
+    std::unique_ptr<ies::data::Data> module_data;
```
pinkwah, line 45: Why `unique_ptr` and not just?
```suggestion
    ies::data::Data module_data;
```

joakim-hove: Good question - I'll update

pinkwah: No. `unique_ptr` is not a trivially-constructible object so it cannot be used in conjunction with `malloc` (undefined behaviour). Please change the alloc function to use `new`.

pinkwah (approved):
:tada: :rocket: :full_moon:

PR #2877, opened 2022-02-09
Move enkf_main_smoother_update into python

### LINKED ISSUE #2878: Move enkf_main_smoother_update into python

ManInFez:
**Issue**
Resolves #2878 

**Approach**

## Pre review checklist

### DIFF
--- a/libres/lib/analysis/update.cpp
+++ b/libres/lib/analysis/update.cpp
@@ -446,7 +446,7 @@ void assert_size_equal(int ens_size, const bool_vector_type *ens_mask) {
             ", mask_size:" + std::to_string(bool_vector_size(ens_mask)));
 }
 
-// Opens and returns a log file.  A subroutine of enkf_main_smoother_update.
+// Opens and returns a log file.
 static FILE *log_step_list(const char *log_path,
                            const std::vector<int> &step_list) {
     std::string log_file;
--- a/libres/lib/analysis/update.cpp
+++ b/libres/lib/analysis/update.cpp
@@ -20,6 +20,7 @@
 #include <ert/analysis/ies/ies_data.hpp>
 #include <ert/analysis/ies/ies.hpp>
 #include <ert/analysis/analysis_module.hpp>
+#include <ert/python.hpp>
 
 static auto logger = ert::get_logger("analysis.update");
 
@@ -445,6 +446,24 @@ void assert_size_equal(int ens_size, const bool_vector_type *ens_mask) {
             ", mask_size:" + std::to_string(bool_vector_size(ens_mask)));
 }
 
+// Opens and returns a log file.  A subroutine of enkf_main_smoother_update.
+static FILE *log_step_list(const char *log_path,
+                           const std::vector<int> &step_list) {
+    std::string log_file;
+    if (step_list.size() == 1)
+        log_file = fmt::format("{}{}{:04d}", log_path, UTIL_PATH_SEP_CHAR,
+                               step_list.front());
+    else
+        log_file = util_alloc_sprintf("{}{}{:04d}-{:04d}", log_path,
+                                      UTIL_PATH_SEP_CHAR, step_list.front(),
+                                      step_list.back());
+    FILE *log_stream = fopen(log_file.data(), "w");
+    if (log_stream == nullptr)
+        throw std::runtime_error(fmt::format(
+            "Error opening '{}' for writing: {}", log_file, strerror(errno)));
+    return log_stream;
+}
+
 bool smoother_update(std::vector<int> step_list,
                      const local_updatestep_type *updatestep,
                      int total_ens_size, enkf_obs_type *obs,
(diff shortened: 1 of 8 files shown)


libres/lib/analysis/update.cpp
```diff
@@ -763,7 +783,28 @@ bool smoother_update(std::vector<int> step_list,
     int_vector_free(ens_active_list);
     meas_data_free(meas_data);
     bool_vector_free(ens_mask);
-
+    fclose(log_stream);
     return true;
 }
 } // namespace analysis
+
+static bool smoother_update(std::vector<int> step_list, py::object updatestep,
```
pinkwah, line 791: I neglected to mention this, but ideally all vectors should be made `const&` in function parameters. It sucks that it's more verbose, but it's good form.

```suggestion
static bool smoother_update(const std::vector<int>& step_list, py::object updatestep,
```

Reason:
```c++
// C++ uses value semantics by default
std::vector<int> a = big_vector();
std::vector<int> b = a;  /* Make a copy of 'a' and put it into 'b'. Potentially very expensive */

// But we need to prefer reference semantics
std::vector<int> a = big_vector();
std::vector<int>& b = a;  /* 'b' is semantically 'a'. No copy is made. */
```

```py
# Python is pass-by-reference and doesn't have value semantics.
# This is more like pointers in C or `object` type in Java.
a = big_vector()
b = a  # a refers to the same object as b
assert b is a  # => true
```

C++ references are a bit more funky since you cannot distinguish between a reference and the object it refers to. This makes it possible for the compiler to implement references as pointers or just rename `b` to `a` since it may have additional knowledge. Eg. taking the address of a reference `&b` will give you the address of the original object. So `&b == &a`. Even though `b` could've been implemented as a pointer in the machine code.

Anyway. The rule of thumb is that you do `const&` for everything, except anything that is made cheap to copy: all primitives (`int`, `double`, `T *`), `std::string_view` (`const std::string&` is still the correct choice), function pointers, and `py::handle`/`py::object`.

ManInFez: I agree

ManInFez:
Test ert please!

ManInFez:
Test ctest please!

libres/lib/analysis/update.cpp
```diff
@@ -616,14 +617,33 @@ void assert_size_equal(int ens_size, const bool_vector_type *ens_mask) {
             ", mask_size:" + std::to_string(bool_vector_size(ens_mask)));
 }
 
+// Opens and returns a log file.  A subroutine of enkf_main_smoother_update.
```
dafeda, line 620: I think we can remove the part "A subroutine of enkf_main_smoother_update."

dafeda:
Is the plan to remove `enkf_main_smoother_update` before this is merged?

ManInFez:
> Is the plan to remove `enkf_main_smoother_update` before this is merged?

it is still referenced in a workflow job, so I am thinking I will leave it.

dafeda:
> > Is the plan to remove `enkf_main_smoother_update` before this is merged?
> 
> it is still referenced in a workflow job, so I am thinking I will leave it.

Should we write that in the commit message, and create issues to get rid of it?

ManInFez:
> > > Is the plan to remove `enkf_main_smoother_update` before this is merged?
> > 
> > 
> > it is still referenced in a workflow job, so I am thinking I will leave it.
> 
> Should we write that in the commit message, and create issues to get rid of it?

There exist an issue to look over all workflows internal to ERT  https://github.com/equinor/ert/issues/1880

libres/lib/analysis/update.cpp
```diff
@@ -445,14 +446,33 @@ void assert_size_equal(int ens_size, const bool_vector_type *ens_mask) {
...
+        log_file = util_alloc_sprintf("%s%c%04d", log_path, UTIL_PATH_SEP_CHAR,
+                                      step_list.front());
+    else
+        log_file =
+            util_alloc_sprintf("%s%c%04d-%04d", log_path, UTIL_PATH_SEP_CHAR,
+                               step_list.front(), step_list.back());
+    FILE *log_stream = util_fopen(log_file, "w");
+
+    free(log_file);
+    return log_stream;
```
pinkwah, line 463: Since we're touching this already...

```suggestion
    std::string log_file;
    if (step_list.size() == 1)
        log_file = fmt::format("{}{}{:04d}", log_path, UTIL_PATH_SEP_CHAR,
                                      step_list.front());
    else
        log_file =
            util_alloc_sprintf("{}{}{:04d}-{:04d}", log_path, UTIL_PATH_SEP_CHAR,
                               step_list.front(), step_list.back());
    FILE *log_stream = fopen(log_file.data(), "w");
    if (log_stream == nullptr)
        throw std::runtime_error(fmt::format("Error opening '{}' for writing: {}", log_file, strerror(errno)));
    return log_stream;
```

(Might have to include `cerrno` and `fmt/format.h`)

res/enkf/es_update.py
```diff
@@ -1,44 +1,63 @@
...
         target_fs = run_context.get_target_fs()
-        return self._smoother_update(data_fs, target_fs)
+
+        time_map = source_fs.getTimeMap()
+
+        last_step = time_map.getLastStep()
+        if last_step < 0:
+            last_step = self.ert.getModelConfig().get_last_history_restart()
+
+        step_list = [i for i in range(last_step + 1)]
```
pinkwah, line 37:
```suggestion
        step_list = list(range(last_step + 1))
```

I miss the old Python 2 days where `range` would get you a list and not a generator. :cry:

pinkwah (approved):
lgtm otherwise

PR #2883, opened 2022-02-09
Remove usage of thread_pool in job_queue

### LINKED ISSUE #2840: Investigate and clean up threading in job_queue.cpp
The file libres/lib/job_queue/job_queue.cpp runs a 1-thread (!) thread_pool from a detached pthread

https://github.com/equinor/ert/blob/d6cd336e67a9070ea48a7f062e6fda3690e5ef72/libres/lib/job_queue/job_queue.cpp#L891-L903

Not convinced about the value of this. Investigate and possibly clean up...

BjarneHerland:
Resolves #2840

**Approach**
The threadpool served two purposes in this module: 1) make handlers run asynchronously i.e. not on the job_queue -thread and 2) it limited the number of concurrently running handlers by the size of the pool. Both these are taken care of using standard C++. Also handling a subtle race-condition present in original code.

## Pre review checklist

### DIFF
--- a/libres/lib/job_queue/job_queue.cpp
+++ b/libres/lib/job_queue/job_queue.cpp
@@ -21,6 +21,9 @@
 #endif
 
 #include <algorithm>
+#include <vector>
+#include <future>
+#include <chrono>
 #include <filesystem>
 
 #include <string.h>
@@ -31,7 +34,6 @@
 
 #include <ert/util/util.hpp>
 #include <ert/res_util/arg_pack.hpp>
-#include <ert/res_util/thread_pool.hpp>
 #include <ert/res_util/res_portability.hpp>
 #include <ert/logging.hpp>
 
@@ -234,7 +236,9 @@ struct job_queue_struct {
     unsigned long usleep_time; /* The sleep time before checking for updates. */
     pthread_mutex_t
         run_mutex; /* This mutex is used to ensure that ONLY one thread is executing the job_queue_run_jobs(). */
-    thread_pool_type *work_pool;
+
+    /* This holds future results of currently running callbacks */
+    std::vector<std::future<void>> active_callbacks;
 };
 
 /*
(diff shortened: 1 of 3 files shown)


libres/lib/job_queue/job_queue.cpp
```diff
@@ -538,81 +554,69 @@ static bool job_queue_check_node_status_files(const job_queue_type *job_queue,
...
-        if (OK)
-            OK = job_queue_node_run_DONE_callback(node);
+    // There is a small timeslot in which status may change after we decide to
+    // run this handler, and before we get the readlock above. Handle it...
+    if (JOB_QUEUE_DONE != job_queue_node_get_status(node)) {
+        logger->warning("Job {}: expected status {} got {}",
+                        job_queue_node_get_name(node), JOB_QUEUE_DONE,
+                        job_queue_node_get_status(node));
+        return;
+    }
```
jondequinor, line 570: If `run_handlers` locks for reading, we can't mutate the status here… So the hope is the `job_queue_loop` will re-run this handler?

So if this is fine, we should log `info`, if it's not we should crash.

jondequinor: Note that there is a second TOCTOU here 😂 which means you could get a log message indicating _nothing_ went wrong.

jondequinor (approved):
This looks good, and works well with some rudimentary testing.

+ for raii

libres/lib/job_queue/job_queue.cpp
BjarneHerland: I probably don't grasp 100% what you mean, but AFAICS this module has never held any writelock when changing status of a node, of which the robustness is another discussion. This method is the outmost scope under our control in the thread, and grabbing the readlock (using `JobListReadLock`) prevents anyone from mutating the list (FWIW) like the original code did.

The difference is that this code "asserts" that status is as expected. As you pointed out in a deleted comment, there is in fact a possible race-condition between the check and the printout which becomes possible because we don't protect status-changes with a lock, as commented above. Should be avoided now.

Returning without mutating state and leave this to other code seems reasonable here. Logging with info-level is fine.

BjarneHerland:
test this please

PR #2834, opened 2022-02-03
Python 3.10

sondreso:
**Issue**
Resolves #2555 

**Approach**
Add Python 3.10 to Github Action workflows, update ecl dependency and fix usage of deprecated keyword

## Pre review checklist

- [x] Added appropriate labels

Adding labels helps the maintainers when writing release notes, see sections and the
corresponding labels here: https://github.com/equinor/ert/blob/main/.github/release.yml

PR #2889, opened 2022-02-11
Remove deprecated parameter 'loop' in Queues 

sondreso:
**Issue**
Python 3.10 have deprecated the loop keyword when creating asyncio
queues. Instead it will use the currently running loop.

**Approach**
Remove the loop parameter. However, in the evaluator, this will bind to the wrong
loop, so we delay the creation of the queue until it is actually needed and we are
in the context of the "correct" thread. This definitely shows that our usage of
asyncio and threads is not optimal, but this is a bigger design discussion that
needs to happen.

Originally part of #2834 

## Pre review checklist


xjules (approved):
LGTM! :rocket:

PR #2892, opened 2022-02-11
Split ies

joakim-hove:
The `ies::data::Data` class has had a `ies::config::Config` member, with this PR that members is pulled out of the `ies::data::Data` class - and calling scope must manage two instances separately. The motivation for this change is twofold:

1.  Better separation of concerns - the `ies::config::Config` class contains configuration settings for how the analyis should be performed - provided by the user (or defaulted). The `ies::data::Data` class contains state managed by the iterative analysis - this class does not hold any user serviceable parts.
2. Right now we have exactly one instance of the `ies::config::Config`class and one instance of the `ies::data::Data` class - in the future the ownership of the `ies::data::Data` instance should be moved to one of the `local_analysis` datastructures - then we will have multiple `ies::data::Data` instances. This will eventually allow the combined use of iterated update modules and local updates. This is the main motivation for this PR.


ManInFez (approved):
I agree. Separation of config and data is :+1:

PR #2897, opened 2022-02-14
Reorder smoother update responsibility

ManInFez:
**Issue**
Resolves #my_issue

**Approach**

## Pre review checklist


ManInFez:
Test ert please!

ManInFez:
Test ert please!

ManInFez:
Test ert please!

ManInFez:
test ert please!

ManInFez:
Test ert please!

libres/lib/analysis/update.cpp
```diff
@@ -67,6 +66,46 @@ typedef struct {
         *node_info; /* mutable: For the serialization of one node */
 } serialize_info_type;
 
+class update_data_type {
```
dafeda, line 69: Considering adding docstring.
Something along the lines of "Container for all matrices needed for an update step"

libres/lib/analysis/update.cpp
```diff
@@ -67,6 +66,46 @@ typedef struct {
...
+            A_with_rowscaling_in,
+        const bool_vector_type *obs_mask_in)
+        : obs_mask(obs_mask_in) {
+        S = S_in;
+        E = E_in;
+        D = D_in;
+        R = R_in;
+        A = A_in;
+        A_with_rowscaling = A_with_rowscaling_in;
+        valid = true;
```
dafeda, line 85: What is valid?

joakim-hove: It is the entire `update_data_type` instance which is valid - or not. The `valid` value is based on whether we have any remaining observations - if the number of observations is zero (and hence no update should be performed), we have `valid = false`. A better name would maybe be `have_observations`

libres/lib/analysis/update.cpp
```diff
@@ -354,7 +382,8 @@ void run_analysis_update_with_rowscaling(
 
     ert::utils::Benchmark benchmark(logger,
                                     "run_analysis_update_with_rowscaling");
-    assert(parameters.size() > 0);
+    if (parameters.size() == 0)
+        return;
```
dafeda, line 386: Should this throw an error or warning of sorts?

libres/lib/analysis/update.cpp
```diff
@@ -446,96 +482,123 @@ void assert_size_equal(int ens_size, const bool_vector_type *ens_mask) {
...
-bool smoother_update(std::vector<int> step_list,
-                     const local_updatestep_type *updatestep,
+update_data_type make_update_data(enkf_fs_type *source_fs,
+                                  enkf_fs_type *target_fs, enkf_obs_type *obs,
+                                  ensemble_config_type *ensemble_config,
+                                  const analysis_config_type *analysis_config,
+                                  bool_vector_type *ens_mask,
+                                  local_ministep_type *ministep,
+                                  rng_type *shared_rng, FILE *log_stream) {
+    /*
```
dafeda, line 503: Is this docstring outdated?

libres/lib/analysis/update.cpp
joakim-hove: I would say no - this is a totally legitimate situation; one could move to the check to calling scope and if `parameters.size() == 0` not call the function at all.

libres/lib/analysis/update.cpp
```diff
@@ -446,96 +482,123 @@ void assert_size_equal(int ens_size, const bool_vector_type *ens_mask) {
...
+    assert_matrix_size(R, "R", active_obs_size, active_obs_size);
+    const bool_vector_type *obs_mask = obs_data_get_active_mask(obs_data);
+    obs_data_scale(obs_data, S, E, D, R, nullptr);
+
+    auto row_scaling_parameters = load_row_scaling_parameters(
+        target_fs, ensemble_config, iens_active_index, active_ens_size,
+        ministep);
+
+    local_ministep_add_obs_data(ministep,
+                                obs_data); // This is not correct conceptually
```
dafeda, line 565: What is not correct conceptually?

libres/lib/analysis/update.cpp
dafeda: From what you write I would suggest `has_observations` and a short descriptive comment.

libres/lib/analysis/update.cpp
dafeda: If we move the check to the calling scope, would it then not make sense to throw an error if `run_analysis_update_with_rowscaling` is called with parameter.size() == 0?

libres/lib/analysis/update.cpp
ManInFez: ministep should only hold the configuration objects, like selected observations and parameters for the update step, not the actual data.

libres/lib/analysis/update.cpp
ManInFez: valid whether is has the data for doing an update step

libres/lib/analysis/update.cpp
dafeda: That's a better comment I think.

libres/lib/analysis/update.cpp
joakim-hove: No - the early return here is just a convenience, it is perfectly legitimate to run through this function with `paramaters = []` it is just a minor waste.

libres/lib/analysis/update.cpp
joakim-hove: To me it looks like this is (partly ?) addressed in #2923

libres/lib/analysis/update.cpp
dafeda: Alright, I don't like it but can definitely live with it.

libres/lib/analysis/update.cpp
ManInFez: I have added it but it does not show up in the diff :shrug:

dafeda (approved):
Good stuff 👍

PR #2903, opened 2022-02-15
Move and rename evaluator tracker to ert/ert

### LINKED ISSUE #2894: Move cli monitoring from `ert/ert_shared` into `ert/ert`
* Add type hinting
* Refactor ert2 pull this code from a new location
* Move/refactor existing tests.

Blunde1:
**Issue**
Resolves #2894 

**Approach**
- Make early PR to allow discussions on approach
- Move definition of `EvaluatorTracker()` to ert/ert/tracker
- Discuss if this is enough to move, or if more of the tracker module in ert/ert_shared/status/tracker should be moved (i.e., the legacy tracker, factory and utils)
    - Discussion yields: Inspect if legacy and thus factory as extension may be avoided
- Create type-hints
- Simplify and refactor `EvaluatorTracker()` for usage in ert3
- Make tests pass

### DIFF
--- a/ert/tracker/evaluator_tracker.py
+++ b/ert/tracker/evaluator_tracker.py
@@ -46,7 +46,6 @@ class EvaluatorTracker:
         next_ensemble_evaluator_wait_time: int = 5,
     ):
         self._model = model
-
         self._monitor_host = host
         self._monitor_port = port
         self._token = token
@@ -54,15 +53,12 @@ class EvaluatorTracker:
         self._protocol = "ws" if cert is None else "wss"
         self._monitor_url = f"{self._protocol}://{host}:{port}"
         self._next_ensemble_evaluator_wait_time = next_ensemble_evaluator_wait_time
-
         self._work_queue: Queue = Queue()
-
         self._drainer_thread = threading.Thread(
             target=self._drain_monitor,
             name="DrainerThread",
         )
         self._drainer_thread.start()
-
         self._iter_snapshot: Dict[int, Snapshot] = {}
 
     def _drain_monitor(self) -> None:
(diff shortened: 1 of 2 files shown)


ert/tracker/evaluator_tracker.py
```diff
@@ -123,7 +125,9 @@ def _drain_monitor(self):
         drainer_logger.debug("tasks complete")
         self._model.teardown_context()
 
-    def track(self):
+    def track(
+        self,
+    ) -> Iterator[Union[FullSnapshotEvent, SnapshotUpdateEvent, EndEvent]]:
```
jondequinor, line 130: This is actually a generator, which should have the hint structure `Generator[yield_type, send_type, return_type]` so then you get

```python
[Generator[Union[FullSnapshotEvent, SnapshotUpdateEvent, EndEvent], None, None]]
```

See https://docs.python.org/3/library/typing.html#typing.Generator)

ert/tracker/evaluator_tracker.py
```diff
@@ -36,13 +39,12 @@ class EvaluatorTracker:
...
-        host,
-        port,
-        general_interval,
-        detailed_interval,
-        token=None,
-        cert=None,
-        next_ensemble_evaluator_wait_time=5,
+        host: str,
+        port: str,
+        general_interval: int,
```
jondequinor, line 44: Should be removed, it's probably no longer useful.

sondreso:
Was confused why Github did not show syntax highlighting, but missing `.py` in the name of the file :sweat_smile:

Blunde1:
@xjules could you take a look if this will work with your PR?

jondequinor (changes requested):
This seems like a copy operation… Do we want two almost identical versions of this tracker? 🤔

Blunde1:
> This seems like a copy operation… Do we want two almost identical versions of this tracker? 🤔

@jondequinor Trying to make it work with @xjules' PR https://github.com/equinor/ert/pull/2944, the choice of currently having one for ert2 and one for ert3 is related to this.
It is unmarked as a draft-PR so that it can be tried out by @xjules and then iteratively improving as needed.
Tried rewriting with major changes, i.e. write an implementation that was scaled to current ert3, but it was not beautiful and due to this it made more sense to do revert, take a step back, and do this closely with #2944

PR #2912, opened 2022-02-15
Add observations to responses query

frode-aarstad:
**Issue**
Resolves equinor/webviz-ert#214

**Approach**
Add observations to the response response

### DIFF
--- a/ert_shared/dark_storage/endpoints/records.py
+++ b/ert_shared/dark_storage/endpoints/records.py
@@ -248,9 +248,28 @@ async def get_record_data(
 def get_ensemble_responses(
     *, res: LibresFacade = Depends(get_res), ensemble_id: UUID
 ) -> Mapping[str, js.RecordOut]:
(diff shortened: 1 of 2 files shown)


jondequinor (commented):
But what if you're only interested in the list?

oysteoh:
> But what if you're only interested in the list?

Guess we could hide the observation loading behind a flag in the api? But is it worth it? Datasets we have been looking at shows that the amount of observations is quite limited, maybe like ~5 on a vector with ~1000 entries.

oysteoh (approved):
I'll say approve, but with the possibility of @jondequinor to argue we should improve the api with this feature. From my point of view - currently we have one consumer, and that consumer needs the observations to be present. YAGNI tells me we should wait with implementing a feature until it is actually needed, but i'm open for discussing this 😸

frode-aarstad:
Jenkins test this please

PR #2914, opened 2022-02-15
make tests ert-storage 0.3.9 compatible

ManInFez:
ert3 does not use get_parameters. Making tests work for current ert-storage api


ManInFez:
fixed by https://github.com/equinor/ert/pull/2917/files

DanSava:
Sorry @ManInFez  missed this PR

PR #2929, opened 2022-02-17
Remove unused argumeent analysis_module_type in ministep

joakim-hove:
This extra module was never used


oyvindeide (approved):
Thanks for cleaning 👍 There is a typo in the commit text, other than that it looks good

PR #2933, opened 2022-02-17
Move logging obs and meas data used in analysis to after loaded

ManInFez:
**Issue**
Resolves https://github.com/equinor/semeio/issues/402

### DIFF
--- a/libres/lib/analysis/update.cpp
+++ b/libres/lib/analysis/update.cpp
@@ -1,7 +1,8 @@
 #include <vector>
 #include <string>
 #include <assert.h>
-#include <optional>
+#include <fmt/format.h>
+#include <cerrno>
 
 #include "ert/analysis/update.hpp"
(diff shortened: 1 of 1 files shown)


libres/lib/analysis/update.cpp
```diff
@@ -1,7 +1,8 @@
 #include <vector>
 #include <string>
 #include <assert.h>
-#include <optional>
+#include <fmt/format.h>
```
xjules, line 4: Why these includes?

ManInFez: because of https://github.com/equinor/ert/blob/6df137c28285ea9f0d495a207939cc636a8e6cf1/libres/lib/analysis/update.cpp#L493

The compiler decided to just ignore the missing include :shrug:

oyvindeide (approved):
🚀

PR #2943, opened 2022-02-18
Change one_at_the_time() to one_at_a_time()

### LINKED ISSUE #2940: Inconsistent one-at-the-time versus one-at-a-time
This is a typo. Most of ert calls this "one-at-a-time", but we have "one-at-the-time" in a few places. This should be synchronized and corrected. Beware it will break the API.

Example: 
https://github.com/equinor/ert/blob/f3d78fe105f87a7910b5b9ad9846d96960b51ed6/ert3/algorithms/_sensitivity.py#L51

berland:
Probably a typo, "one at a time" is the common phrase used
in the ERT repo.

API break in ert3.

**Issue**
Resolves #2940 

**Approach**
Search and replace

## Pre review checklist

### DIFF
--- a/ert3/algorithms/__init__.py
+++ b/ert3/algorithms/__init__.py
@@ -1,3 +1,3 @@
-from ert3.algorithms._sensitivity import one_at_the_time, fast_sample, fast_analyze
+from ert3.algorithms._sensitivity import one_at_a_time, fast_sample, fast_analyze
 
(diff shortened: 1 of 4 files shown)


jondequinor:
How is this an API break? Isn't the ERT3 API the cli for the time being?

berland:
> Isn't the ERT3 API the cli for the time being?

If so, this is not an "API break" ..

PR #2944, opened 2022-02-18
CLI monitoring for ert3

### LINKED ISSUE #2895: Create tracker events in ert3
* Create a dummy `BaseRunModel`, see API below
* Add `create_tracker()` in ert3 console

The `BaseRunModel`-like object needs to implement the following API:

- `isFinished() -> bool`: a method indicating whether or not the experiment is done. Since there's nothing beyond evaluating ensembles, this is `true` as soon as prefect finishes and the results are stored.
- `teardown_context() -> None`: a noop for ert3
- `hasRunFailed() -> bool`: a method which returns `true` if something failed
- `getFailMessage() -> Optional[str]`: a method which may return an error message in the case of failure
- `getPhaseName() -> str`: Set to `Evaluating...`?
- `currentPhase() -> int`: Set to `0`.
- `phaseCount() -> int`: Set to `1`.
- `isIndeterminate() -> bool`: whether or not we cannot determine progress. Always `false` for the time being.

xjules:
**Issue**
Resolves #2895 

**Approach**

* Make a dummy run model
* Make use of `cli/Monitor` to handle cli

## Pre review checklist

### DIFF
--- a/ert3/evaluator/_evaluator.py
+++ b/ert3/evaluator/_evaluator.py
@@ -1,26 +1,95 @@
 import pickle
-from typing import Dict
+import time
+from typing import Dict, Optional, Union, Any
+from concurrent import futures
+import sys
 
 import ert
 from ert_shared.ensemble_evaluator.config import EvaluatorServerConfig
+from ert_shared.models.base_run_model import BaseRunModel
+from ert_shared.status.tracker.evaluator import EvaluatorTracker
+from ert_shared.cli.monitor import Monitor
 from ert_shared.ensemble_evaluator.ensemble.base import Ensemble
-from ert_shared.ensemble_evaluator.entity.identifiers import EVTYPE_EE_TERMINATED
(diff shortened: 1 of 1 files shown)


ert3/evaluator/_evaluator.py
```diff
@@ -91,7 +78,7 @@ def isIndeterminate(self) -> bool:
         # whether or not we cannot determine progress. Always false for the time being.
         return False
 
-    def setIndeterminate(self, indeterminate: Union[bool, None]):
+    def setIndeterminate(self, indeterminate: Union[bool, None]) -> None:
```
jondequinor, line 84: I think this could safely be `raise NotImplementedError`. IIRC the `BaseRunModel` itself calls this API during updates. We don't have that yet, so let's just make that clear in code.

xjules:
Closing this (cont. #2960)

PR #2945, opened 2022-02-18
Add timeout message when starting services

TerryHannant:
**Issue**
When starting ert api there is a timeout associated with starting services. If the timeout is exceeded  the ert process is shut down without any indication of the cause. This causes confusion when determining the cause of the shutdown

**Approach**
Print simple message. There may be a need to have a more robust handling of this situation in future


ert_shared/services/_base_service.py
```diff
@@ -188,6 +188,9 @@ def _read_conn_info(self, proc: Popen) -> Optional[str]:
 
             # Timeout reached, exit with a failure
             if ready == ([], [], []):
+                print(
+                    f"Service {self._service_name} startup exceeded defined timeout {self._timeout}s, initiating shutdown "
```
berland, line 192: Intentional space at the end of the string?

TerryHannant: Nope!, Fixed

PR #2951, opened 2022-02-21
Fix removing duplicates, added test to verify, improved performance

### LINKED ISSUE #2899: Investigate duplicate check in summary collector
Preliminary profiling of retrieving Summary vectors via api found that  a check for duplicates performed at

https://github.com/equinor/ert/blob/main/ert_shared/libres_facade.py#L163

Can take longer to perform than reading data from dark storage
For example 200 realizations x 30 000 on RGS node:

- reading data from dark storage  - 2.0s
- checking for duplicates  - 2.7s

Thus this check is a heavy overhead when retrieving data via API.

Investigation needs to be done to see if the duplication check is needed or if it can be removed.

BjarneHerland:
**Issue**
Resolves #2899 and adds test for removing duplicates.


ert_shared/libres_facade.py
```diff
@@ -1,5 +1,8 @@
 import os
+import time
+import logging
 from pandas import DataFrame
+import numpy as np
```
DanSava, line 5: Is this import needed? Also, `time` seems to be imported whiteout usage.

ert_shared/libres_facade.py
```diff
@@ -1,5 +1,8 @@
 import os
+import time
```
sondreso, line 2: Not used?

ert_shared/libres_facade.py
BjarneHerland: Thx...  left over from testing

ert_shared/libres_facade.py
BjarneHerland: See above

tests/ert_tests/shared/test_libres_facade_summarydata.py
```diff
@@ -0,0 +1,82 @@
...
+    df = _orig_loader(*args, **kwargs)
+    # Append copy of last date to each realization
+    idx = pd.MultiIndex.from_tuples(
+        [(i, df.loc[i].iloc[-1].name) for i in df.index.levels[0]]
+    )
+    df_new = pd.DataFrame(0, index=idx, columns=df.columns)
+    return pd.concat([df, df_new]).sort_index()
+
+
+class LibresFacadeSummarydataTest(TestCase):
```
DanSava, line 28: Does this need to be a separate test or can be integrated into existing test files that use LibresFacade like `test_integration_data.py`

tests/ert_tests/shared/test_libres_facade_summarydata.py
```diff
@@ -0,0 +1,82 @@
...
+    # TODO: same as in test_libres_facade.py - perhaps move into utils?
```
DanSava, line 29: TODO still hanging around, do you still need to to move to utils or was just forgotten here ?

tests/ert_tests/shared/test_libres_facade_summarydata.py
```diff
@@ -0,0 +1,82 @@
...
+
+        rc = ResConfig(user_config_file=config_file)
+        rc.convertToCReference(None)
+        ert = EnKFMain(rc)
+        facade = LibresFacade(ert)
+        return facade
+
+    @pytest.fixture(autouse=True)
+    def inject_fixtures(self, caplog):
+        self._caplog = caplog
```
DanSava, line 41: This is a bit strange to me, the way you are using caplog. Why not just use it for the tests that require it? 
as an example check out `test_plugin_manager.py:test_workflows_merge_duplicate`

DanSava (approved):
Sorry for the wait time, had some small comments regarding the test but overall looks good. Adress them if you think appropriate and merge when ready.

tests/ert_tests/shared/test_libres_facade_summarydata.py
BjarneHerland: Good point...  I'll merge them and get rid of the TODO below as well.

tests/ert_tests/shared/test_libres_facade_summarydata.py
BjarneHerland: I added explanations in code

PR #2956, opened 2022-02-22
Improve failure behaviour from ert client to storage

### LINKED ISSUE #2953: `ert3 init` leaves directory `.ert` behind when initialization fails
**Describe the bug**
If an ert3 workspace is attempted initialized, but connection with the storage server fails, the directory `.ert` is still created and rerunning `ert3 init` (after storage is started) will not work until the directory is manually removed.

**To Reproduce**
1. terminal A: `ert3 init`. Results possibly in `RuntimeError: Server started, but none of the URLs provided worked` when it has tried using URL from an outdated `storage_server.json`.
2. terminal B: `ert3 service start storage` (and ensure the wrong `storage_server.json` is overwritten or overridden)
3. terminal A: `ert3 init`. Results in `Already inside an ERT workspace, found /work/projects/ert/ert3_examples/polynomial/.ert`

**Expected behavior**
Successful initialization in step 3.

berland:
**Issue**
Resolves #2953 

**Approach**
The console already has functionality for removing the `.ert` directory, but only if it catches a TimeoutError. The storage initialization raises a RuntimeError. I chose to change the exception raised instead of catching one  more exception.

https://github.com/equinor/ert/blob/main/ert3/console/_console.py#L223-L229

## Pre review checklist

### DIFF
--- a/ert/storage/_storage.py
+++ b/ert/storage/_storage.py
@@ -535,7 +535,9 @@ def init(*, workspace_name: str) -> None:
 
     for special_key in _SPECIAL_KEYS:
         if f"{workspace_name}.{special_key}" in experiment_names:
-            raise ValueError("Storage already initialized")
+            raise RuntimeError(
+                f"Workspace {workspace_name} already registered in storage"
+            )
         _init_experiment(
             experiment_name=f"{workspace_name}.{special_key}",
             parameters={},
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -8,8 +8,9 @@ sphinx-argparse
 sphinx_rtd_theme
 sphinx-autoapi
 pytest-asyncio
-requests
+pytest-mock
 pytest-timeout
+requests
 jsonpath_ng
 sphinxcontrib.datatemplates
 decorator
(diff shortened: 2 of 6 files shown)


berland:
Bonus commit added that handles a similar situation when the workspace already exists in storage.

sondreso (approved):
LGTM!

ert/storage/_storage.py
```diff
@@ -535,7 +535,9 @@ def init(*, workspace_name: str) -> None:
 
     for special_key in _SPECIAL_KEYS:
         if f"{workspace_name}.{special_key}" in experiment_names:
-            raise ValueError("Storage already initialized")
+            raise ValueError(
```
xjules, line 538: Just a small detail here. Shouldn't this be a different error: eg. `NameError`?

jondequinor: `NameError` is very specific, maybe `RuntimeError` is a better match?

xjules: `RuntimeError` is another option and is indeed more general.

berland:
Tests added. Is it ok to expose the ".ert" implementation detail in tests?

sondreso (approved):
I think the usage of ".ert" in the test is ok in this scenario.

PR #2957, opened 2022-02-22
Remove unused function local_obsdata_node_copy_active_list()

joakim-hove:
Remove unused function; small part of #2906


oyvindeide (approved):
Good catch! :+1:

PR #2965, opened 2022-02-23
Remove legacy tracker

### LINKED ISSUE #1323: Rip out the legacy legacy consumption
Legacy tracker would be really good to kill.

**Why?**
During the work of creating the evaluator and the event based interaction between the evaluation and the monitoring the existing mechanism of communicating directly via shared memory was kept. First, it was kept as default and later as a fallback mechanism via `--disable-ensemble-evaluator`. It is time to remove the option of direct memory communication to lower the complexity of the implementation.

**Definition of Done**
The option `--disable-ensemble-evaluator` is no longer supported and the code to support it is removed.

sondreso:
**Issue**
Resolves #1323

**Approach**
- Remove option for disabling ensemble evaluator
- Move evaluator config out of simulation args

## Pre review checklist


sondreso:
Jenkins test this please

ert_shared/models/base_run_model.py
```diff
@@ -138,11 +140,9 @@ def teardown_context(self) -> None:
 
     @job_queue(None)
     def killAllSimulations(self) -> None:
```
lars-petter-hauge, line 142: Can this not be completely removed? I don't find any subclasses that implement this function.. And the `LegacyTracker` that uses it is removed in this PR.

ert_shared/models/ensemble_experiment.py
```diff
@@ -26,18 +25,9 @@ def runSimulations__(self, arguments, run_msg):
...
-            ee_config = arguments["ee_config"]
-            num_successful_realizations = self.run_ensemble_evaluator(
-                run_context, ee_config
-            )
-        else:
-            self._job_queue = self._queue_config.create_job_queue()
-            num_successful_realizations = (
-                self.ert()
-                .getEnkfSimulationRunner()
-                .runEnsembleExperiment(self._job_queue, run_context)
```
lars-petter-hauge, line 39: As we are removing the usage of `SimulationRunner` to a large degree here, we should investigate if we can further remove the implementation as well, or if it is in use anywhere.

ert_shared/cli/main.py
```diff
@@ -61,17 +61,16 @@ def run_cli(args):
             )
             raise ErtCliError(msg)
 
-        ee_config = EvaluatorServerConfig(custom_port_range=args.port_range)
-        argument.update({"ee_config": ee_config})
+        evaluator_config = EvaluatorServerConfig(custom_port_range=args.port_range)
```
lars-petter-hauge, line 64: I agree that it should be renamed, and that `evaluator_config` is better than `ee_config`, but I wouldn't guess that `evaluator_config` contained only connection information, such as token, cert, protocol etc.. Should we rename it further towards server? I would think `evaluator_config` contained information on how to evaluate the ensemble, and things like `algorithm` would be included in it.

lars-petter-hauge:
I had some minor comments on the PR, but otherwise I agree with the changes done. Cheers for removing the `LegacyTracker`

ert_shared/cli/main.py
jondequinor: `evaluator_server_config` then ? :)

ert_shared/models/base_run_model.py
jondequinor: Indeed. This basically calls `self._job_queue.kill_all_jobs()`, which the legacy ensemble also calls.

ert_shared/models/ensemble_experiment.py
jondequinor: Everest depends on _some_ code in the sim. runner, IIRC.

jondequinor (approved):
ive-been-looking-forward-to-this-dooku

jondequinor:
Storage and ERT3 failures—I suggest merge.

sondreso:
Jenkins test this please

sondreso:
I forgot to squash that last commit...

PR #2979, opened 2022-02-25
Add an optional has_observations flag to the record

### LINKED ISSUE #2970: Add a flag to responses in dark storage if it has observations
`GET "/ensembles/{ensemble_id}/records/{name}",`

Due to the slow reading of observations adding a list of all observations for a response made everything too slow. 

Instead we only add a flag indicating if the response has observations.

frode-aarstad:
**Issue**
Resolves #2970


oysteoh (approved):
Seems to work and give performance as it was before the commit introducing the observations in the response!

frode-aarstad:
Jenkins test this please

PR #2987, opened 2022-02-28
Add timeout message when starting services

TerryHannant:
**Issue**
When starting ert api there is a timeout associated with starting services. If the timeout is exceeded the ert process is shut down without any indication of the cause. This causes confusion when determining the cause of the shutdown

**Approach**
Print simple message. There may be a need to have a more robust handling of this situation in future


frode-aarstad (approved):
LGTM

PR #3000, opened 2022-02-28
Remove dependency on BLAS and LAPACK

pinkwah:
The functionality that previously required BLAS and LAPACK have been
replaced with Eigen3.

Eigen3 may compile down to code that calls BLAS or LAPACK procedures if
enable, but that is something that we should consider doing only after
major refactoring is done and we have a need for it.


dafeda (approved):
So nice 👍

PR #3022, opened 2022-03-04
Fix dependecy issues

sondreso:
**Issue**
Ert is not compatible with ert-storage<3.11. Also the install order in the coverage workflow is slightly unusual, and could cause problems.

**Approach**
- Pin ert-storage>=0.3.11
- Flip install order in coverage workflow

## Pre review checklist

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -32,8 +32,8 @@ jobs:
 
     - name: Install with dependencies
       run: |
-        pip install -r dev-requirements.txt
         pip install .
+        pip install -r dev-requirements.txt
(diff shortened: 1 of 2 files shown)


jondequinor (approved):

.github/workflows/coverage.yml
```diff
@@ -32,8 +32,8 @@ jobs:
 
     - name: Install with dependencies
       run: |
-        pip install -r dev-requirements.txt
         pip install .
+        pip install -r dev-requirements.txt
```
jondequinor, line 36: Makes sense!

PR #3026, opened 2022-03-04
Use the key_manager from enkf_main

### LINKED ISSUE #2973: Dark Storage - observations_for_obs_keys performance
If observation data is  loaded for all summary responses by repeatedly calling  observations_for_obs_keys there is a large performance cost, In some cases approx 280 seconds, which makes it unusable as part of HTTP API

### Initial investigation:
Profiling shows that 99% of the time of calls to observations_for_obs_keys is used in the call to loadObservationData 

https://github.com/equinor/ert/blob/324d702235a69961d6396afffa8e7e4eac4a6dab/res/enkf/export/summary_observation_collector.py#L18

within this function majority of time is spent in SummaryObservationCollector.getAllObservationKeys

https://github.com/equinor/ert/blob/324d702235a69961d6396afffa8e7e4eac4a6dab/res/enkf/export/summary_observation_collector.py#L31

which in turn calls  summaryKeysWithObservations

https://github.com/equinor/ert/blob/324d702235a69961d6396afffa8e7e4eac4a6dab/res/enkf/key_manager.py#L50

At this point the code loops over all SummaryKeys loading each related node, even though observation for a single summary key is being loaded

https://github.com/equinor/ert/blob/324d702235a69961d6396afffa8e7e4eac4a6dab/res/enkf/key_manager.py#L56-L58
```
for key in self.summaryKeys()
    if len(self.ensembleConfig().getNode(key).getObservationKeys()) > 0
``` 
So thus if loadObservationData  is called once per summary response then in turns reloads each summary vector resulting in O(N**2) problem
         
Since a new KeyManager instance is used for each call the caching of results at instance level is not used

frode-aarstad:
**Issue**
Resolves #2973

### DIFF
--- a/res/enkf/export/gen_kw_collector.py
+++ b/res/enkf/export/gen_kw_collector.py
@@ -3,7 +3,6 @@ from ecl.util.util import BoolVector
 from pandas import DataFrame
 from res.enkf import EnKFMain
 from res.enkf.enums import RealizationStateEnum
-from res.enkf.key_manager import KeyManager
 
 
 class GenKwCollector:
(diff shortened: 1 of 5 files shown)


oysteoh (approved):
Seems like a big improvement reusing the keymanager present in enkf_main instead of instantiate new ones in all of the collectors. 🚀

PR #3027, opened 2022-03-04
clang-format: Remove version check

pinkwah:
This commit also speeds up `clang-format` when installed from PyPI. It
detects whether the `clang-format` executable is a Python wrapper and
finds where the actual binary is location.


dafeda (approved):
LGTM 👍

PR #3041, opened 2022-03-07
Add jenkins script running flaky tests repeatedly

lars-petter-hauge:
**Issue**
As part of working with flaky-tests we want to run those marked as flaky daily repeatedly and verify that fixes does improve robustness.

Will need to go through the list of flaky tests and verify that the ones we need are included.


BjarneHerland (approved):
LGTM

PR #3045, opened 2022-03-08
Introduce read-only info class derived from EvaluatorServerConfig

### LINKED ISSUE #3044: EvaluatorTracker should utilize EvaluatorServerConfig
The signature of `EvaluatorTracker` has too many arguments and too many many attributes for pylint. It is always called using an `EvaluatorServerConfig` in code.

**Proposed solution:**
Refactor `EvaluatorTracker` to utilize `EvaluatorServerConfig`

Blunde1:
**Issue**
Resolves #3044 

**Approach**
~~Utilize `EvaluatorServerConfig` and make sure references are okay~~
- [x] Create a new read-only information class derived from `EvaluatorServerConfig`.
- [x] Employ the new class (called `EvaluatorConnectionInfo`)

## Pre review checklist

### DIFF
--- a/ert_shared/status/tracker/evaluator.py
+++ b/ert_shared/status/tracker/evaluator.py
@@ -15,6 +15,7 @@ from aiohttp import ClientError
 from websockets.exceptions import ConnectionClosedError
 
 from ert_shared.models.base_run_model import BaseRunModel
+from ert_shared.ensemble_evaluator.config import EvaluatorConnectionInfo
 import ert_shared.ensemble_evaluator.entity.identifiers as ids
 from ert_shared.ensemble_evaluator.entity.snapshot import PartialSnapshot, Snapshot
 from ert_shared.ensemble_evaluator.monitor import create as create_ee_monitor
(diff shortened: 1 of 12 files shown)


ert_shared/status/tracker/evaluator.py
```diff
@@ -215,11 +204,11 @@ def request_termination(self):
             return
 
         with create_ee_monitor(
```
xjules, line 206: Good approach! Can this create function take directly `ee_config` as parameter too?

Blunde1: Yes, implemented it.
The same is likely true for `SyncWebsocketDuplexer`. Looking into this now.

sondreso:
The `EvaluatorServerConfig` has much more functionality that consumers besides the server itself should not use. I don't think it's the best option to spread this config around more in the code than it already is. If we want something that can encapsulate this info that is always passed around together, I think we should create a new class, that is only the data. I think there has already been an attempt at this with the `EvaluatorServerConfigInfo` that is in the same file as the `EvaluatorServerConfig`, but it is maybe to specifically made for the dispatchers and not the clients. So if we are to change this, I would suggest creating a class with only the necessary data (read-only), that can serve the need for connection onfo for both the dispatchers and the clients.

Blunde1:
@sondreso like this: https://github.com/equinor/ert/pull/3045/commits/cafea9acce6577ae7911eaebdbefaa9b8f86d293h ?
Then use `get_connection_info()` from the `ee_config` where `ee_config` is now passed?

xjules:
> The `EvaluatorServerConfig` has much more functionality that consumers besides the server itself should not use. I don't think it's the best option to spread this config around more in the code than it already is. If we want something that can encapsulate this info that is always passed around together, I think we should create a new class, that is only the data. I think there has already been an attempt at this with the `EvaluatorServerConfigInfo` that is in the same file as the `EvaluatorServerConfig`, but it is maybe to specifically made for the dispatchers and not the clients. So if we are to change this, I would suggest creating a class with only the necessary data (read-only), that can serve the need for connection info for both the dispatchers and the clients. (Maybe a proper name could be `EvaluatorConnectionInfo` ?)

I've seen the  `EvaluatorServerConfigInfo` class, which is made for dispatchers as you said. If we are not use exploit the info class then what about to use a simple dict_cfg as we discussed yesterday @Blunde1 @sondreso ?

ert_shared/status/tracker/evaluator.py
```diff
@@ -36,30 +37,18 @@ class EvaluatorTracker:
     def __init__(
         self,
         model: BaseRunModel,
-        host,
-        port,
-        token=None,
-        cert=None,
+        ee_config: EvaluatorServerConfig,
```
xjules, line 40: `EvaluatorConnectionInfo` ?

Blunde1: Yes! Will get to refactor everything now. But this is exactly it. : ) 👍

xjules (approved):
Good job Berent! Once tests are green 🚀

PR #3048, opened 2022-03-08
WIP: Remove ActiveList

pinkwah:
This commit removes the ActiveList class but keeps the functionality. In
essense, ActiveList is a optional set of indices that are 'active'. If
this optional is unset, then we consider _all_ indices to be active.
Because of the simplicity, using a dedicated class for this data is an
unreasonable code maintenance burden.

Resolves: #2485


pinkwah:
Closing due to age. `ActiveList` can be grouped into two use-cases:
1. Specifying which observation is active in a list of observations, or `None` for all active.
2. When scaling, it is the mask that specifies which values to scale, or `None` to scale all.

In the latter, the name "active" makes little sense, since it doesn't control which of the values are 'active'.

PR #3059, opened 2022-03-08
Make port range larger in test

mortalisk:
Test was not working when many workers running, as no ports were available.
https://ci.equinor.com/scout/job/komodo_testing/job/ert-custom-script-pr/9340/console

### DIFF
--- a/ert3/evaluator/_evaluator.py
+++ b/ert3/evaluator/_evaluator.py
@@ -1,5 +1,5 @@
 import pickle
-from typing import Dict
+from typing import Dict, Optional
 
 import ert
 from ert_shared.ensemble_evaluator.config import EvaluatorServerConfig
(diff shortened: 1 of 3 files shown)


ert3/evaluator/_evaluator.py
```diff
@@ -50,8 +50,9 @@ def _run(
 
 def evaluate(
     ensemble: Ensemble,
+    custom_port_range: Optional[range] = None,
```
xjules, line 53: Are we to expose `port_range` as parameter here?

mortalisk: I dunno. Any reason why not?

xjules: You might be right though as I can't see any other way to deliver the port range there.

jondequinor: I agree with both of you.

mortalisk: :monkey: :monkey: :monkey: :monkey: :question:

xjules (approved):
🚀

PR #3063, opened 2022-03-08
Wrap unused_tcp_port fixture to mock localhost

### LINKED ISSUE #2778: NoPortsInRangeException failure on Jenkins
I received the below exception in CI on Jenkins. It is unrelated to my work and both the test and the code appears untouched for months. Hence, I expect it to either have been flaky for some time or be affected by other changes.

**Stacktrace**
```
_______________________ test_find_available_port_forced ________________________

unused_tcp_port = 57880

    def test_find_available_port_forced(unused_tcp_port):
        custom_range = range(unused_tcp_port, unused_tcp_port)
        host, port, sock = port_handler.find_available_port(
>           custom_range=custom_range, reuse_addr=True
        )

/tmp/jenkins-scout-f_scout_ci/workspace/komodo_testing/ert-custom-script-pr@2/test-root/tests/ert_tests/shared/test_port_handler.py:23: 

custom_host = None, custom_range = range(57880, 57880), reuse_addr = True

    def find_available_port(
        custom_host: Optional[str] = None,
        custom_range: Optional[range] = None,
        reuse_addr: bool = False,
    ) -> Tuple[str, int, socket.socket]:
        current_host = custom_host if custom_host is not None else _get_ip_address()
        current_range = (
            custom_range if custom_range is not None else range(51820, 51840 + 1)
        )
        if current_range.start == current_range.stop:
            try:
                return (
                    current_host,
                    current_range.start,
                    _bind_socket(
                        host=current_host, port=current_range.start, reuse_addr=reuse_addr
                    ),
                )
            except PortAlreadyInUseException:
                pass
        else:
            attempts = 0
            while attempts < current_range.stop - current_range.start:
                try:
                    attempts += 1
                    num = random.randrange(current_range.start, current_range.stop)
                    return (
                        current_host,
                        num,
                        _bind_socket(host=current_host, port=num, reuse_addr=reuse_addr),
                    )
                except PortAlreadyInUseException:
                    continue
    
>       raise NoPortsInRangeException(f"No available ports in predefined {current_range}.")
E       ert_shared.port_handler.NoPortsInRangeException: No available ports in predefined range(57880, 57880).

/tmp/jenkins-scout-f_scout_ci/workspace/komodo_testing/ert-custom-script-pr@2/test-kenv/root/lib64/python3.6/site-packages/ert_shared/port_handler.py:52: NoPortsInRangeException
```

**Link to job**
https://ci.equinor.com/scout/job/komodo_testing/job/ert-custom-script-pr/6142/console

lars-petter-hauge:
**Issue**
Resolves #2778 

**Approach**
Wrap unused_tcp_port fixture in order to set host to localhost.

ran `pytest --count=30000 -x tests/ert_tests/shared/test_port_handler.py::test_find_available_port_forced` locally, passed (used to fail after 1000-4000 runs)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -59,3 +59,24 @@ def pytest_runtest_setup(item):
 
 def pytest_configure(config):
     config.addinivalue_line("markers", "equinor_test")
+
+
+@pytest.fixture()
+def unused_tcp_port(unused_tcp_port, monkeypatch):
+    # The unused_tcp_port checks against localhost when providing an unused port:
+    # https://github.com/pytest-dev/pytest-asyncio/blob/9246f5825f589ff01f1c67620ce34bdc416c5af3/pytest_asyncio/plugin.py#L490
+    # In find_available_port, we use our public ip address if a current_host is
+    # not provided (as given in _get_ip_address). Using the public ip address
+    # is fine in production, but for unit testing paired with unused_tcp_port
+    # we have to specify localhost. We wrap unused_tcp_port and mock
+    # the `_get_ip_address` to only provide localhost. The original unused_tcp_port
+    # remains untouched
+    import ert_shared.port_handler
+
+    def mocked_get_ip_address():
+        return "127.0.0.1"
+
+    monkeypatch.setattr(
+        ert_shared.port_handler, "_get_ip_address", mocked_get_ip_address
+    )
+    return unused_tcp_port


tests/conftest.py
```diff
@@ -45,3 +45,26 @@ def copy_case(path, config_file):
...
+    # is fine in production, but for unit testing paired with unused_tcp_port
+    # we have to specify localhost. We wrap unused_tcp_port and mock
+    # the `_get_ip_address` to only provide localhost.
+    import ert_shared.port_handler
+    import socket
+
+    def mocked_get_ip_adress():
+        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        s.connect(("0.0.0.0", 80))
+        return s.getsockname()[0]
```
jondequinor, line 65:
```suggestion
    def mocked_get_ip_adress():
        return "127.0.0.1"
```

tests/conftest.py
```diff
@@ -45,3 +45,26 @@ def copy_case(path, config_file):
 
     with tmpdir.as_cwd():
         yield copy_case
+
+
+@pytest.fixture()
+def unused_tcp_port(unused_tcp_port, monkeypatch):
```
jondequinor, line 51: Are we sure we want to overwrite this fixture (coming from pytest-asyncio, right?) for all of ERT?

Shouldn't localhost be a value that can be set _somewhere_ and ultimately reach `custom_host` of `find_available_port`?

tests/conftest.py
```diff
@@ -45,3 +45,26 @@ def copy_case(path, config_file):
...
+    # https://github.com/pytest-dev/pytest-asyncio/blob/9246f5825f589ff01f1c67620ce34bdc416c5af3/pytest_asyncio/plugin.py#L490
+    # In find_available_port, we use our public ip address if a current_host is
+    # not provided (as given in _get_ip_address). Using the public ip address
+    # is fine in production, but for unit testing paired with unused_tcp_port
+    # we have to specify localhost. We wrap unused_tcp_port and mock
+    # the `_get_ip_address` to only provide localhost.
+    import ert_shared.port_handler
+    import socket
+
+    def mocked_get_ip_adress():
```
berland, line 62: adress -> address

tests/conftest.py
lars-petter-hauge: Nono, that's just making it too complicated...

tests/conftest.py
lars-petter-hauge: The concern here was that although it is easy for the low level tests to add `hostname`, the same may not be true for e.g. integration tests. Expanding the api exclusively to cater for running the tests locally seemed unnecessary. I'll admit I did not go through all the tests that uses `unused_tcp_port`, but I can.

Are there any obvious scenarios where we would _want_ the public ip address out of `_get_ip_address` in a test? (I do understand that even with a "no" to that question, it's still debatable to wrap the fixture for the entire ERT test setup..)

> (coming from pytest-asyncio, right?)

Correct

BjarneHerland: > Are there any obvious scenarios where we would _want_ the public ip address out of `_get_ip_address` in a test?

From a different angle: Are there scenarios where a test would use `unused_tcp_port` and fail if the system binds it to localhost, i.e. is it likely to hurt us to wrap this fixture?

The obvious alternative is to introduce our own fixture, e.g. `ert_unused_tcp_port`, but then 1) lots of existing tests must be patched, and 2) we risk having new tests use the original `unused_tcp_port` and miss this fix.

lars-petter-hauge: > 2) we risk having new tests use the original unused_tcp_port and miss this fix.

This is valid point. It wasn't obvious what the issue was based on the error when a test failed. It could be that test documentation on how we should run tests that opens sockets must be setup would improve this (perhaps make it searchable by adding typical error message)

tests/conftest.py
jondequinor: I think also `0.0.0.0` will bind to all possible interfaces, so they're not equivalent.

tests/conftest.py
jondequinor: > Are there any obvious scenarios where we would want the public ip address out of _get_ip_address in a test?

Some test will test our secure websocket/http communication, and I'm not entirely sure if you can have e.g. websockets accept certs made for localhost? If you do, then no—we probably want localhost everywhere.

> Just for reference, the unused_tcp_port is used in 16 tests from what I could see

Did you follow fixtures using this fixture? The search has to be recursive.

> I would say it's reasonable that an EvaluatorServerConfig should accept a host.

I agree.

lars-petter-hauge: > Did you follow fixtures using this fixture? The search has to be recursive.

My search was not exhaustive it seems, there were about 30 more tests that used `unused_tcp_port`. A single fixture used it, but in combination with `localhost`

BjarneHerland: > > Are there any obvious scenarios where we would want the public ip address out of _get_ip_address in a test?
> 
> Some test will test our secure websocket/http communication, and I'm not entirely sure if you can have e.g. websockets accept certs made for localhost? If you do, then no—we probably want localhost everywhere.

Mmmm...  could you point to any of these tests so we could check?

It should be the original (unchanged) "fixture-object" from `pytest-asyncio` which is given to the test-method. Our wrapper implements a side-effect - perhaps we can just parametrize our wrapper to avoid this side-effect if it is not desired?

jondequinor: > Mmmm... could you point to any of these tests so we could check?

I think it's more or less the default. But a good compromise would be one integration test with secure comms, and all others without.

lars-petter-hauge:
Superseded by https://github.com/equinor/ert/pull/3096

PR #3067, opened 2022-03-09
Increasing default timeout from 20s to 120s in BaseService.

BjarneHerland:
Also properly log startup-failures due to timeout + extend testcase to verify this.

### DIFF
--- a/ert_shared/services/_base_service.py
+++ b/ert_shared/services/_base_service.py
@@ -29,7 +29,6 @@ from typing import (
 if TYPE_CHECKING:
     from inspect import Traceback
 
-logger = getLogger("ert_shared.storage")
 T = TypeVar("T", bound="BaseService")
 ConnInfo = Union[Mapping[str, Any], Exception, None]
 
@@ -116,7 +115,7 @@ class _Proc(threading.Thread):
         global SERVICE_NAMES
         SERVICE_NAMES.add(self._service_name)
 
-        self._proc = Popen(
+        self._childproc = Popen(
             self._exec_args,
             pass_fds=(fd_write,),
             env=env,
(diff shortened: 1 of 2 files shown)


ert_shared/services/_base_service.py
```diff
@@ -333,6 +329,9 @@ def wait_until_ready(self, timeout: Optional[float] = None) -> bool:
             return not (
                 self._conn_info is None or isinstance(self._conn_info, Exception)
             )
+
+        if isinstance(self._conn_info, TimeoutError):
+            self.logger.critical(f"startup exceeded defined timeout {self._timeout}s")
```
lars-petter-hauge, line 334: reminder: should this not be the `timeout` passed as the argument, instead of `self._timeout`?

ert_shared/services/_base_service.py
```diff
@@ -206,10 +202,10 @@ def _do_shutdown(self) -> None:
             return
         try:
             self._proc.terminate()
-            self._proc.wait(10)
+            self._proc.wait(10)  # Give it 10s to shut down first..
         except TimeoutExpired:
-            self._proc.kill()
-            self._proc.wait()
+            self._proc.kill()  # ... then kick it harder...
+            self._proc.wait(self._timeout)  # ... and wait a little longer
```
lars-petter-hauge, line 208: So the way this was setup before was that it would simply hang on `wait` if the process was not killed?  And now we could end up returning without process being completely killed? Should we then check the return value of `wait` (which will be false in the given scenario of timeout reached)? (not sure what else to do at this stage though.... Perhaps at least some logging would be beneficial)

ert_shared/services/_base_service.py
BjarneHerland: Good point! This class don't really need `self._timeout`

ert_shared/services/_base_service.py
BjarneHerland: My understanding of existing code is that it waits 10s after trying terminate(), then tries kill() and waits forever. I don't think this is a big problem, but it feels a little safer to allow a timeout also in the second wait, and I'll definitely add logging

BjarneHerland:
The main motivation for this has been achieved by #3076 and #3078. Closing this for now.

PR #3081, opened 2022-03-14
Make one ies::data::Data instance for each ministep

joakim-hove:
This PR is not complete - an might never be. The main purpose of this PR is to illustrate that we should - in one way or another - create one `ies::data::Data` instance per ministep and pass that to the update routine. The interesting change is in `es_update`routine in the last commit.


ertomatic:

PR #3085, opened 2022-03-14
Remove UTIL_TYPE_ID and everything related to it

pinkwah:
The usage of the following macros has been removed:
- `UTIL_TYPE_ID_INIT`
- `UTIL_TYPD_ID_DECLARATION`
- `UTIL_IS_INSTANCE_HEADER`
- `UTIL_IS_INSTANCE_FUNCTION`
- `UTIL_SAFE_CAST_HEADER`
- `UTIL_SAFE_CAST_HEADER_CONST`
- `UTIL_SAFE_CAST_FUNCTION`
- `UTIL_SAFE_CAST_FUNCTION_CONST`

The invocations of `_safe_cast` and `_safe_cast_const` have been
replaced with an equivalent `reinterpret_cast<T>` "call", and the
invocations for `_is_instance` are assumed to be `true` in all instances
and removed.

This system is both technically unsafe and superfluous due to the
existence of C++ functionality that achieves the same thing. It is
unsafe because each function assumes that a `void*` can be turned into a
`int*` for type-identification. This may not always be the case, where
either the `void*` is pointing to empty memory (eg. by `malloc(0)`) or
the first four bytes may not be correct aligned to `int`'s alignment,
potentially causing load errors (on CPUs that are not supported by ERT,
but still). Other systems work around this by having a dedicated type
which does have the necessary properties, like eg. Python's `PyObject*`,
but we didn't do this. As it stands currently, it's more appropriate
long-term to get rid of this mechanism entirely rather than fixing it.

Resolves: #2625

### DIFF
--- a/libres/lib/config/config_content.cpp
+++ b/libres/lib/config/config_content.cpp
@@ -30,10 +30,8 @@
 #include <ert/config/config_path_stack.hpp>
 #include <ert/config/config_content.hpp>
 
-#define CONFIG_CONTENT_TYPE_ID 6612520
 
 struct config_content_struct {
-    UTIL_TYPE_ID_DECLARATION;
     std::set<std::string> *
         parsed_files; /* A set of config files whcih have been parsed - to protect against circular includes. */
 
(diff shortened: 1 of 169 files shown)


pinkwah:
Closing due to staleness.

PR #3089, opened 2022-03-15
Avoid branching

BjarneHerland:
Probably handled well by a proper optimizing compiler but code just feels...  clunky.


jondequinor (approved):
😆

PR #2132, opened 2021-10-04
Experiment server

sondreso:
**Issue**
Resolves #1250

**Approach**
1. [x] Refactor out the tracking of an evaluation.
2. [x] Allow the evaluator to track multiple evaluations
3. [x] Add the concept of an experiment with its own status, containing multiple evaluations.
4. [x] Make the run models setup, start and finish the experiment at the correct times.
5. [x] Make experiment endpoint for clients (decide on "out-of-band" communication for experiments and evaluations vs all status in one channel)
6. [ ] Make evaluator communication session
7. [ ] Cleanup

PR #3108, opened 2022-03-16
Refactor RunDialog to depend less on RunModel

### LINKED ISSUE #3024: Make GUI independent of RunModel
After the evaluator was introduced, the GUI now gets most of the information about the currently running experiment from the events that it receives (from the EvaluatorTracker). However, there is still some information that is accessed directly from the run model in the RunDialog ([run_dialog.py](https://github.com/equinor/ert/blob/main/ert_gui/simulation/run_dialog.py)). This is mostly related to getting the mask of active realizations and tiem onformation about the entire experiment. We should not depend on getting this information directly from the run model, and instead also pass the necessary information via events.

xjules:
**Issue**
Resolves #3024 

**Approach**
- [x] Refactor `RunDialog` by migrating simulation running info into `BaseRunModel`
- [x] Move more processing logic into RunModel
~~- [ ] Make `EvaluatorTracker` as a proxy for `BaseRunModel`~~ this might be affected by #2132 so left it for now

## Pre review checklist

### DIFF
--- a/ert_shared/models/base_run_model.py
+++ b/ert_shared/models/base_run_model.py
@@ -1,6 +1,6 @@
 import time
 import uuid
-from typing import Optional, List, Union
+from typing import Optional, List, Union, Dict, Any
 import asyncio
 
 from ecl.util.util import BoolVector
@@ -30,11 +30,19 @@ class ErtRunError(Exception):
 
 
 class BaseRunModel:
-    def __init__(self, ert: EnKFMain, queue_config: QueueConfig, phase_count: int = 1):
+    def __init__(
+        self,
+        simulation_arguments: Dict[str, Any],
+        ert: EnKFMain,
+        queue_config: QueueConfig,
+        phase_count: int = 1,
+    ):
         """
 
         Parameters
         ----------
+        simulation_arguments : Parameters for running the simulation,
+            eg. activate realizations, analysis module
         queue_config : QueueConfig
         phase_count : Optional[int], optional
             Number of data assimilation cycles / iterations an experiment will have, by default 1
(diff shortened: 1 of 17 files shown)


jondequinor (commented):
Did you figure out the segfault?

xjules:
> Did you figure out the segfault?

Yes, I did. Rebasing commits one by one did the trick 😄

ert_shared/models/base_run_model.py
```diff
@@ -74,21 +83,59 @@ def reset(self) -> None:
...
+
+    def has_failed_realizations(self) -> bool:
+        completed = self.completed_realizations_mask
+        initial = self.initial_realizations_mask
+        for (index, successful) in enumerate(completed):
+            if initial[index] and not successful:
+                return True
+        return False
+
+    def create_mask_from_failed_realizations(self) -> BoolVector:
```
jondequinor, line 104:
```suggestion
    def _create_mask_from_failed_realizations(self) -> BoolVector:
```

ert_shared/models/base_run_model.py
```diff
@@ -74,21 +83,59 @@ def reset(self) -> None:
...
+        :return: Type BoolVector
+        """
+        completed = self.completed_realizations_mask
+        initial = self.initial_realizations_mask
+        inverted_mask = BoolVector(default_value=False)
+        for (index, successful) in enumerate(completed):
+            inverted_mask[index] = initial[index] and not successful
+        return inverted_mask
+
+    def count_successful_realizations(self) -> int:
```
jondequinor, line 116:
```suggestion
    def _count_successful_realizations(self) -> int:
```

jondequinor (approved):
This is solid stuff. Really good refactor. Had some minor comments to rein in the BRM API.

Tested out all run models in the GUI, seems to work well! Also tried restart on a failed mda run.

👏 👏

sondreso:
Could you shorten the commit message to be < 50 characters :slightly_smiling_face:

PR #3112, opened 2022-03-17
Automatically increase softlimit for max open files while running tests

BjarneHerland:
Far too often, the testsuite fails because default `ulimit -n` in the shell is too low. This patch uses an autouse-fixture and the Python resource-API to temporarily bump this value to max for the test-session.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import os
+import resource
 import pkg_resources
 import shutil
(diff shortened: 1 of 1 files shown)


BjarneHerland:
Jenkins test this please

BjarneHerland:
Jenkins test this please

jondequinor (approved):
Good stuff.

PR #3115, opened 2022-03-17
WIP: Refactor `block_fs.cpp`

pinkwah:
This PR does a whole bunch of stuffs, including:
- Removing mostly unused code
- Making variables into constants. (Eg, `block_size` is always 64 in production code, so remove the ability to configure it to anything else)
- **Removed `.index` file** _[Behaviour makes no sense when multiple processes read the same block-fs. Eg, dark storage + ert. If the file isn't found then the old implementations will "rebuild" the index file, so this does not break anything]_
- **No buffered data writes** _[Old behaviour was only to sync every 10 writes. We now sync for every write]_

**Notes**:
I have discovered [`liblockfile`](https://linux.die.net/man/1/lockfile-create) which is supposedly atomic on NFS also. It is usually not installed by default nor is it available via Conan. Perhaps we use it for PyPI builds and CI, and use homemade lockfiles if the library isn't found?


pinkwah:
Closing in favour of https://github.com/equinor/ert/pull/3185 . Other parts of this PR will be used to create other PRs.

PR #3119, opened 2022-03-18
Enhancing test to verify whether the actual jobs finished or not

BjarneHerland:
Let the "jobs" touch a file upon normal exit, use existence of this file as robust indication of whether jobs succeded or not.


lars-petter-hauge (approved):
Agreed!

Issue #2824, opened 2022-02-01
FAILED ert_tests/dark_storage/test_http_endpoints.py::test_get_response


jondequinor:
**Describe the bug**
```
FAILED ert_tests/dark_storage/test_http_endpoints.py::test_get_response - ass...
FAILED ert_tests/dark_storage/test_http_endpoints.py::test_misfit_endpoint - ...

=================================== FAILURES ===================================
______________________________ test_get_response _______________________________

poly_example_tmp_dir = None
dark_storage_client = <starlette.testclient.TestClient object at 0x7f0466356fa0>

    def test_get_response(poly_example_tmp_dir, dark_storage_client):
        resp: Response = dark_storage_client.get("/experiments")
        experiment_json = resp.json()
    
        ensemble_id1 = experiment_json[0]["ensemble_ids"][0]
        ensemble_id2 = experiment_json[0]["ensemble_ids"][1]
    
        resp: Response = dark_storage_client.get(f"/ensembles/{ensemble_id1}")
        ensemble_json = resp.json()
        assert ensemble_json["response_names"][0] == "POLY_RES@0"
    
        resp: Response = dark_storage_client.get(f"/ensembles/{ensemble_id2}")
        ensemble_json2 = resp.json()
        assert ensemble_json2["response_names"][0] == "POLY_RES@0"
    
        resp: Response = dark_storage_client.get(
            f"/ensembles/{ensemble_id1}/responses/POLY_RES@0/data"
        )
        stream = io.BytesIO(resp.content)
        response_df1 = pd.read_csv(stream, index_col=0, float_precision="round_trip")
    
        resp: Response = dark_storage_client.get(
            f"/ensembles/{ensemble_id2}/responses/POLY_RES@0/data"
        )
        stream = io.BytesIO(resp.content)
        response_df2 = pd.read_csv(stream, index_col=0, float_precision="round_trip")
    
>       assert len(response_df1.columns) == 10
E       assert 0 == 10
E         +0
E         -10

/home/runner/work/ert/ert/tests/ert_tests/dark_storage/test_http_endpoints.py:84: AssertionError
_____________________________ test_misfit_endpoint _____________________________

poly_example_tmp_dir = None
dark_storage_client = <starlette.testclient.TestClient object at 0x7f04663fa1f0>

    def test_misfit_endpoint(poly_example_tmp_dir, dark_storage_client):
        resp: Response = dark_storage_client.get("/experiments")
        experiment_json = resp.json()
[...]

PR #3132, opened 2022-03-21
Add additional information on failure in test_http_response

lars-petter-hauge:
The result in https://github.com/equinor/ert/issues/2824 is difficult to reproduce.

A hypothesis for the failures is that there could be remnants of previously ran tests in memory which interferes with this test. This PR changes variable names to non-default values and adds some additional output on failures. Also disproving our hypothesis would be helpful.

### DIFF
--- a/tests/ert_tests/dark_storage/conftest.py
+++ b/tests/ert_tests/dark_storage/conftest.py
@@ -26,15 +26,16 @@ def poly_example_tmp_dir_shared(
         os.path.join(source_root, "test-data", "local", "poly_example"),
         poly_dir,
     )
-
     with poly_dir.as_cwd():
         parser = ArgumentParser(prog="test_main")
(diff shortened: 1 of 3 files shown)


lars-petter-hauge:
```
pytest --count 500 dark_storage/test_http_endpoints.py::test_get_response --force-disable-session

dark_storage/test_http_endpoints.py .................................... [  7%]
........................................................................ [ 21%]
........................................................................ [ 36%]
........................................................................ [ 50%]
........................................................................ [ 64%]
........................................................................ [ 79%]
........................................................................ [ 93%]
................................                                         [100%]


================= 500 passed, 1 warning in 9207.86s (2:33:27) ==================
```

Not easily reproduced at least.

BjarneHerland (approved):
Nit: should probably comment why we check for len==3, but good to go!

PR #3145, opened 2022-03-23
Ies from python

ManInFez:
Cleaning a bit includes in analysis

Binding ies functions to python

## Pre review checklist


oyvindeide (approved):
LGTM! Though could you add some more text in the last commit instead of the title of the previous commits? Could you also update the PR text to be a bit more descriptive?

PR #3156, opened 2022-03-24
Add support for log-uniform distribution

### LINKED ISSUE #2925: Add support for log-uniform distribution in ert3
Make it possible to model parameters with a log-uniform distribution in ert3, similar to LOGUNIF in ert2.

Make it possible to write:

```
- name: my_fault_multiplier
  type: stochastic
  distribution:
    type: loguniform
    input:
      lower_bound: 0.00001
      upper_bound: 0.1
```

**Additional context**
General distribution support issue:  #1458

verveerpj:
**Issue**
Resolves #2925

**Approach**
Duplicate mostly the code for the uniform distribution, translating upper and lower bound to the shape parameters of the loguniform distribution, taking into account that the lower bound must be > 0.

## Pre review checklist

### DIFF
--- a/ert3/config/_parameters_config.py
+++ b/ert3/config/_parameters_config.py
@@ -61,7 +61,7 @@ class _UniformInput(_ParametersConfig):
     upper_bound: float
 
     @root_validator
-    def _ensure_lower_upper(cls, values):  # type: ignore
+    def _ensure_lower_upper(cls, values: Dict[str, Any]) -> Dict[str, Any]:
         low = values.get("lower_bound")
         up = values.get("upper_bound")
 
@@ -76,11 +76,32 @@ class _UniformInput(_ParametersConfig):
         )
 
 
+class _LogUniformInput(_ParametersConfig):
+    lower_bound: float
+    upper_bound: float
+
+    @root_validator
+    def _ensure_lower_upper(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        low = values.get("lower_bound")
+        up = values.get("upper_bound")
+
+        if low is None or up is None:
+            return values
+
+        if up > low > 0:
+            return values
+
+        raise ValueError(
+            f"Expected lower_bound ({low}) to be > 0, "
+            "and smaller than upper_bound ({up})"
+        )
+
+
 class _DiscreteInput(_ParametersConfig):
     values: List[float]
 
     @root_validator
-    def _ensure_proper_values(cls, values):  # type: ignore
+    def _ensure_proper_values(cls, values: Dict[str, Any]) -> Dict[str, Any]:
         value_list = values.get("values")
 
         if value_list is not None and len(value_list) > 0:
@@ -101,6 +122,11 @@ class _GaussianDistribution(_ParametersConfig):
 class _UniformDistribution(_ParametersConfig):
     type: Literal["uniform"]
     input: _UniformInput
+
+
+class _LogUniformDistribution(_ParametersConfig):
+    type: Literal["loguniform"]
+    input: _LogUniformInput
 
 
 class _DiscreteDistribution(_ParametersConfig):
@@ -152,6 +178,7 @@ class _ParameterConfig(_ParametersConfig):
         _DiscreteDistribution,
         _GaussianDistribution,
         _UniformDistribution,
+        _LogUniformDistribution,
     ]
     variables: Optional[_VariablesConfig] = None
     size: Optional[int] = None
@@ -200,6 +227,16 @@ class _ParameterConfig(_ParametersConfig):
             assert dist_config.input.upper_bound is not None
 
             return ert3.stats.Uniform(
+                dist_config.input.lower_bound,
+                dist_config.input.upper_bound,
+                index=index,
+                size=size,
+            )
+        elif dist_config.type == "loguniform":
+            assert dist_config.input.lower_bound is not None
+            assert dist_config.input.upper_bound is not None
+
+            return ert3.stats.LogUniform(
                 dist_config.input.lower_bound,
                 dist_config.input.upper_bound,
                 index=index,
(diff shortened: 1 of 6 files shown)


ertomatic:

sondreso:
Jenkins add to whitelist

ert3/config/_parameters_config.py
```diff
@@ -76,6 +76,27 @@ def _ensure_lower_upper(cls, values):  # type: ignore
         )
 
 
+class _LogUniformInput(_ParametersConfig):
+    lower_bound: float
+    upper_bound: float
+
+    @root_validator
+    def _ensure_lower_upper(cls, values):  # type: ignore
```
Blunde1, line 84: Could we use
`def _ensure_lower_upper(cls, values: Dict[str, Any]) -> Dict[str, Any]:`
?

ert3/config/_parameters_config.py
```diff
@@ -205,6 +232,16 @@ def as_distribution(self) -> ert3.stats.Distribution:
                 index=index,
                 size=size,
             )
+        elif dist_config.type == "loguniform":
+            assert dist_config.input.lower_bound is not None
+            assert dist_config.input.upper_bound is not None
```
Blunde1, line 237: Linking this to the comment above. And again I see that this check is done for all distribution types, so I guess there is a reason to it. But still I am curious why this is not done in the `_<distribution_name>Input` instead, instead of accepting `None`?

Blunde1: These two comments might come outside of the PR review, but placing them here while I have them.

ert3/stats/_stats.py
```diff
@@ -163,6 +163,50 @@ def type(self) -> str:
...
+
+        def rvs(size: int) -> np.ndarray:  # type: ignore
+            return np.array(
+                scipy.stats.loguniform.rvs(
+                    a=self._lower_bound, b=self._upper_bound, size=self._size
+                )
+            )
+
+        def ppf(x: np.ndarray) -> np.ndarray:  # type: ignore
+            return np.array(
```
Blunde1, line 186: Is it an option to use `npt.NDArray` from `numpy.typing as npt` instead of `# type: ignore`?
https://numpy.org/doc/stable/reference/typing.html#module-numpy.typing

Blunde1 (commented):
Great job @verveerpj!

I had a few comments relevant to the PR only with typing, and two comments that are outside the scope of the PR but that I asked to understand the module / pydantic-way-implementation of the _parameters_config.py module better.

I have left to look at the tests, which I will do in a separate review coming soon. 🙂

ert3/config/_parameters_config.py
```diff
@@ -76,6 +76,27 @@ def _ensure_lower_upper(cls, values):  # type: ignore
...
+    lower_bound: float
+    upper_bound: float
+
+    @root_validator
+    def _ensure_lower_upper(cls, values):  # type: ignore
+        low = values.get("lower_bound")
+        up = values.get("upper_bound")
+
+        if low is None or up is None:
+            return values
```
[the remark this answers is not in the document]
verveerpj: I did not notice that since I basically the code, but I think this because a pydantic root validator is used. If you do that you have to be prepared that some values are None, due to the occurrence of an validation error before this validator is called. For instance, if you provide a value that cannot be parsed to a float, that causes a validation error before this root validator is called. The root validator will still be called, with a None for the offending value.

ert3/config/_parameters_config.py
verveerpj: yes, we should. Will fix in all other classes too.

ert3/config/_parameters_config.py
verveerpj: It seems to me that these asserts do not really need to be there. I don't think those values can be `None`.

Blunde1:
Another thing I am curious about is how the LogUniform will work with the _sensitivity.py module.
Specifically, will this break?
https://github.com/equinor/ert/blob/420f42d7d7bb3e0fe283a200e32758e7e5b224c5/ert3/algorithms/_sensitivity.py#L85-L100
We are importing algorithms from SALib, and from what I see they only have Gaussian, Uniform and constant possibilities?
https://salib.readthedocs.io/en/latest/basics.html#another-example

ert3/stats/_stats.py
verveerpj: Yes, but Python 3.6 can only use an older version of numpy that does not support that. That would still be ok, if we make strings out of those types, which is not very elegant. It seems we have not done that anywhere in ERT yet, maybe something to revisit when we have 3.8 as the minimum.

verveerpj:
> Another thing I am curious about is how the LogUniform will work with the _sensitivity.py module. Specifically, will this break?
> 
> https://github.com/equinor/ert/blob/420f42d7d7bb3e0fe283a200e32758e7e5b224c5/ert3/algorithms/_sensitivity.py#L85-L100
> 
> 
> We are importing algorithms from SALib, and from what I see they only have Gaussian, Uniform and constant possibilities?
> https://salib.readthedocs.io/en/latest/basics.html#another-example

I have no idea, but the code in ERT specifically only supports gaussian and uniform in the sensitivity analysis. We should probably leave that to another PR if somebody argues it should also support loguniform.

ert3/config/_parameters_config.py
Blunde1: But perhaps should be moved in a separate PR?

verveerpj: The asserts may be overly cautious, but don't hurt, let's just leave them.

Blunde1 (approved):
This LGTM! Great job @verveerpj !
I really like the tests in `test_parameters_config`. I kept wishing to test for mathematically wrong inputs, and of course this was tested in the last test here. 🙂 
The ones in `test_engine` and `test_distributions` seemed very thorough and good as well.

PR #3161, opened 2022-03-25
Pin pylint to version <2.13.0

sondreso:
**Issue**
Version 2.13.0 introduces changes that makes our testing pipelines fail.

**Approach**
This commits pins the version until a fix is implemented.

See also follow up issue: https://github.com/equinor/ert/issues/3162

## Pre review checklist


BjarneHerland (approved):
Go!

PR #3163, opened 2022-03-28
Fix to circumvent pylint bug

### LINKED ISSUE #3162: Unpin pylint
In https://github.com/equinor/ert/pull/3161 pylint was pinned, due to incompatibility with version 2.13.0. We should fix the errors and lift the restriction on the allowed versions again.

verveerpj:
**Issue**
Resolves #3162

**Approach**
Replace a `...` with `Ellipsis`. It is a bug in `pylint`, but the fix is still appropriate.

## Pre review checklist

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,7 +5,7 @@ decorator
 flake8>=4.0.0
 jsonpath_ng
 mock; python_version <= '3.7' # missing AsyncMock added in py38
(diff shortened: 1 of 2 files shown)


Blunde1 (approved):
Given @jondequinor comment in `ert-dev` I guess this is okay?
LGTM at least. 👍

PR #3166, opened 2022-03-30
Update readme after first setup

### LINKED ISSUE #3338: Update README with setup instructions

valentin-krasontovitsch:
**Issue**
Troubles when first trying to install ert for development and run complete test suite.

**Approach**
Document all necessary steps and implement any changes to get the tests to run and pass.

## Pre review checklist

- [X] Added appropriate labels

Resolves #3338

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,10 +14,10 @@
 [![License: GPL v3](https://img.shields.io/badge/License-GPLv3-blue.svg)](https://www.gnu.org/licenses/gpl-3.0)
 [![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black)
 
-ERT - Ensemble based Reservoir Tool - is designed for running 
+ERT - Ensemble based Reservoir Tool - is designed for running
 ensembles of dynamical models such as reservoir models,
 in order to do sensitivity analysis and data assimilation.
-ERT supports data assimilation using the Ensemble Smoother (ES), 
+ERT supports data assimilation using the Ensemble Smoother (ES),
 Ensemble Smoother with Multiple Data Assimilation (ES-MDA) and
 Iterative Ensemble Smoother (IES).
 
@@ -61,7 +61,23 @@ the easiest to work with and is likely what most developers will work with.
 
 ### Developing Python
 
-To start developing the Python code, install ERT in editable mode into a virtualenv:
+You might first want to make sure that some system level packages are installed
+before attempting setup:
+
+```
+- pip
+- python include headers
+- (python) venv
+- (python) setuptools
+- (python) wheel
+```
+
+It is left as an exercise to the reader to figure out how to install these on
+their respective system.
+
+To start developing the Python code, we suggest installing ERT in editable mode
+into a [virtual environment](https://docs.python.org/3/library/venv.html) to
+isolate the install (substitute the appropriate way of sourcing venv for your shell):
 
 ```sh
 # Create and enable a virtualenv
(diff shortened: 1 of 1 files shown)


ertomatic:

dev-requirements.txt
```diff
@@ -7,6 +7,7 @@ jsonpath_ng
 mock; python_version <= '3.7' # missing AsyncMock added in py38
 pylint
 pytest-asyncio
+pytest-benchmark
```
berland, line 10: Should this one be in this PR?

README.md
```diff
@@ -55,20 +55,50 @@ Documentation for ert is located at [https://ert.readthedocs.io/en/latest/](http
...
 *ERT* is Python and C software. To start developing, install it in editable
-mode:
+mode. We suggest using [virtual
+environment](https://docs.python.org/3/library/venv.html) to isolate the
+install (substitute the appropriate way of sourcing venv for your shell):
 
 ```
 $ git clone https://github.com/equinor/ert
 $ cd ert
+$ python -m venv venv
```
berland, line 65: (I prefer not to have my venv inside the ert directory, as it will be wiped if I do a `git clean`)

README.md
```diff
@@ -55,20 +55,55 @@ Documentation for ert is located at [https://ert.readthedocs.io/en/latest/](http
...
 ```
 
+As a simple test of your `ert` installation, you may try to run one of the
+examples, for instance:
+
+```
+$ cd test-data/local/poly_example
+$ # for non-gui trial run
+$ ert ensemble_experiment poly.ert
+$ # for gui trial run
```
jondequinor, line 93: There's also `test_run` which is smaller and faster.

README.md
valentin-krasontovitsch: Sorry, total venv newb (tried to avoid dependency management in python in the past, as we only used python for one single project) - could you kindly make a suggestion for a well-suited location that could be used in the readme?

README.md
valentin-krasontovitsch: could elaborate on this please? i don't see `test_run` as an example in `test-data/local`,  so I'm not sure what you're suggesting should be run precisely

dev-requirements.txt
valentin-krasontovitsch: `pytest-benchmark` was something I needed to install for the tests to run - so I figured I'd add it to the requirements. it might be that I was wrong though or misunderstoof, do you think it's not necessary?

README.md
jondequinor: `ert test_run` is a subset of `ert ensemble_experiment` in that it only runs _one_ realization. So it's faster.

The `ensemble_experiment` will run an ensemble of 100 realizations, which is a bit heavy =)

valentin-krasontovitsch: ah thanks i see now! gonna change that accordingly, and rebase to main, i think @pinkwah made some changes here in the meantime, too

README.md
```diff
@@ -59,7 +59,17 @@ the easiest to work with and is likely what most developers will work with.
...
 ### Developing Python
 
-To start developing the Python code, install ERT in editable mode into a virtualenv:
+You might first want to make sure that some system level packages are installed
+before attempting setup:
+
+```
+apt-get install -y python3-pip python3-dev python3-venv \
+  python3-setuptools python3-wheel
+```
```
jondequinor, line 68: I think you should rather list the packages, or specify that this applies to debian/ubuntu. If you go for the latter, you probably need the same for RH and Mac as well?

valentin-krasontovitsch: oof ja very good point. i could setup a virtual machine to test for RH (i suppose?!), but don't know of any way to test for mac. is it enough if i just list the packages and specify that this information concerns a debian-based setup? and leave it up to RH / mac users to do the same for RH / mac, respectively (if it's even necessary)?

jondequinor (approved):
One small nitpick. Other than that, thank you!

README.md
jondequinor: Why not say that you need

- pip
- python include headers
- venv
- setuptools
- wheel

and leave it as an exercise to the reader?

valentin-krasontovitsch: that sounds fair : )

PR #3167, opened 2022-03-30
Gui monitoring

### LINKED ISSUE #2896: Use the GUI monitoring in ert3 (Use feature flag to enable)
stdout/stderr is not part of this issue.

Note that it is the `TrackerWorker` that consumes the `Tracker`, it is not to be consumed in the GUI code directly.

It's set up like this, in `run_dialog.py`:

```python
        tracker = create_tracker(
            self._run_model,
            ee_con_info=evaluator_server_config.get_connection_info(),
        )

        worker = TrackerWorker(tracker)
```

xjules:
**Issue**
Resolves #2896 

**Approach**
- Use RunDialog from ert2
- Introduce a new `job_index` attribute into the job builder
- Each Snapshot realization will get its own list of steps and each step its jobs
- Tested on MacOS 12.3.1 and all examples worked with `--gui` option.

## Pre review checklist

- [x] write tests


berland:
I have tried this and noted two things:
* The "Kill simulation" seems to work (i.e. something is killed), but the experience in the terminal window is a little rough (multiple tracebacks).
* When showing "Details", the order of jobs in the table seems random, but they execute in the correct order.

berland:
Would it make sense (and be possible) to open up the GUI earlier? It takes some time (posting data to storage) before the GUI pops up (testing on snake_oil).

xjules:
> I have tried this and noted two things:
> 
> * The "Kill simulation" seems to work (i.e. something is killed), but the experience in the terminal window is a little rough (multiple tracebacks).
> * When showing "Details", the order of jobs in the table seems random, but they execute in the correct order.

Yes, the `pyrsisten.pmap` does not preserve the order of jobs - trying to find alternative to it. `Kill simulation` is not supported yet.

> Would it make sense (and be possible) to open up the GUI earlier? It takes some time (posting data to storage) before the GUI pops up (testing on snake_oil).

Good suggestion (as it annoys me too), will check it out.

xjules:
> > I have tried this and noted two things:
> > 
> > * The "Kill simulation" seems to work (i.e. something is killed), but the experience in the terminal window is a little rough (multiple tracebacks).
> > * When showing "Details", the order of jobs in the table seems random, but they execute in the correct order.
> 
> Yes, the `pyrsisten.pmap` does not preserve the order of jobs - trying to find alternative to it. `Kill simulation` is not supported yet.

- Due to the `pyrsistent.pmap` issue I need to revert it back to use `job_index`.
- `Kill simulations` features was decided to be disabled and thus hiding the button for now.

xjules:
Test ERT please!

ert/ensemble_evaluator/builder/_job.py
```diff
@@ -35,42 +36,48 @@ class _UnixJob(_BaseJob):
...
     ) -> None:
-        super().__init__(id_, name, "")  # no step_source needed for legacy (pt.)
+        super().__init__(id_, index, name, "")  # no step_source needed for legacy (pt.)
         self.ext_job = ext_job
 
 
 class _BaseJobBuilder:
     def __init__(self) -> None:
         self._id: Optional[str] = None
+        self._index: Optional[str] = None
```
jondequinor, line 80: Why did  you choose string? Just curious

ert/ensemble_evaluator/builder/_job.py
```diff
@@ -107,14 +118,16 @@ def set_args(self, args: Tuple[str, ...]) -> "_JobBuilder":
     def build(self) -> Union[_FunctionJob, _UnixJob]:
         if self._id is None:
             self._id = str(uuid.uuid4())
+        if self._index is None:
+            self._index = "0"
```
jondequinor, line 122: I think we should require this? Fail if not set?

jondequinor (commented):
Looks very good! A few comments.

ert/ensemble_evaluator/builder/_job.py
xjules: Good point. As it being converted to json and back I thought better to have it directly in `str`. But if you prefer int, I agree to change it.

ert_gui/simulation/run_dialog.py
```diff
@@ -302,6 +310,57 @@ def run():
...
+                        if event.data.get("status") == ENSEMBLE_STATE_FAILED:
+                            self._run_model._failed = True
+                            self._run_model._fail_message = "Ensemble evaluation failed"
+                            raise RuntimeError("Ensemble evaluation failed")
+                    if event["type"] == EVTYPE_EE_TERMINATED and isinstance(
+                        event.data, bytes
+                    ):
+                        result = pickle.loads(event.data)
+                        self._run_model.setPhase(1, "Simulations completed.")
+            return result
```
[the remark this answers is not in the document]
xjules: Agree, but I have bigger plans with this function, but as you said let's focus on this later.

ert/ensemble_evaluator/builder/_job.py
xjules: It was failing when haven't specified it though; ie. failing anyway. I have a question: why `self._id` can be None also?

jondequinor: I think it's just tailored for Prefect, so that Prefect can get IDs generated… No other reasons IIRC.

.github/workflows/typing.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         python -m pip install -r types-requirements.txt
     - name: Run mypy
       run: |
-        mypy --disallow-untyped-defs --ignore-missing-imports -p ert_shared.models
+        mypy --disallow-untyped-defs --follow-imports=skip --ignore-missing-imports -p ert_shared.models
```
berland, line 36: Is this a mypy relaxation that is necessary in order to take in `ert_gui` ? But do we need it both here and in  `.mypy-strict.ini`?

ert/ensemble_evaluator/snapshot.py
```diff
@@ -15,6 +15,7 @@
     get_job_id,
     get_real_id,
     get_step_id,
+    get_job_index,
```
berland, line 18: (these imports can be sorted alphabetically - or it will be done automatically when we at one point are able to use `isort`)

ert_gui/simulation/run_dialog.py
```diff
@@ -302,6 +310,57 @@ def run():
         worker_thread.started.connect(worker.consume_and_emit)
         self._worker_thread.start()
 
+    def startSimulationErt3(self, ensemble_evaluator):
```
berland, line 313: should we add typing to new functions like this? (easening future typing of everything)

berland:
> * The "Kill simulation" seems to work (i.e. something is killed), but the experience in the terminal window is a little rough (multiple tracebacks).
> * When showing "Details", the order of jobs in the table seems random, but they execute in the correct order.

These are now resolved.

ert_gui/simulation/run_dialog.py
xjules: Now it's added

.github/workflows/typing.yml
xjules: Unfortunately we need it now in both as ert_gui is leaking into `ert3` through `RunDialog`. We could create two .ini files specifying modules to ignore?

ert3/evaluator/_evaluator.py
```diff
@@ -17,10 +17,6 @@
     EVTYPE_EE_TERMINATED,
 )
```
jondequinor, line 19: Any reason for removing this?

xjules: I don't see it actually. What was removed?

ert3/evaluator/_evaluator.py
```diff
@@ -17,10 +17,6 @@
...
 
-preload_module_name = "ert.ensemble_evaluator.builder._prefect_forkserver_preload"
-loader = importlib.util.find_spec(preload_module_name)
-if not loader:
-    raise ModuleNotFoundError(f"No module named {preload_module_name}")
 from ert_gui.simulation.run_dialog import run_monitoring_ert3
```
jondequinor: @xjules this was removed, sorry for the bad reference in the previous comment.

xjules:
Tested on MacOS 12.3.1 and all examples worked with `--gui` option.

ert/ensemble_evaluator/builder/_job.py
xjules: Now it's mandatory

ert/ensemble_evaluator/builder/_job.py
```diff
@@ -159,8 +176,10 @@ def set_ext_job(self, ext_job: ExtJob) -> "_LegacyJobBuilder":
     def build(self) -> _LegacyJob:
         if self._id is None:
             self._id = str(uuid.uuid4())
+        if self._index is None:
+            self._index = "0"
```
jondequinor, line 180: I thought you meant to make this mandatory?

berland (approved):
One happy reviewer ⭐

jondequinor (approved):
- Wrap the commit body at 72 characters (other than that, great commit!)
- As the GUI code stands now, I don't see how else to do the QT import. I don't think the import itself is the problem, rather I think it is the way the current code is laid out. But that can be refactored.
- Attaching `index` to `source` of the CloudEvent is an inaccuracy I think, as the spec says that "the 'source' is the context in which the occurrence happened". The job should be _data_ on the job itself, but this is largely semantics so feel free to ignore.
- I think we should not have a default `"0"` index on these jobs, especially since the index is used to communicate order. Seems this should be explicitly set and a required value.


Other than that, great job wrapping your head around this and implementing it. This issue pays back important technical debt, and it looks good when you run it 🎉

ert/ensemble_evaluator/builder/_job.py
xjules: That's what the reviews are for - forgot on that one. Will change this one too!

xjules: > Attaching index to source of the CloudEvent is an inaccuracy I think, as the spec says that "the 'source' is the context in which the occurrence happened". The job should be data on the job itself, but this is largely semantics so feel free to ignore.

There is still time to fix it. I just thought that since `index` was on the same semantic level as `id` therefore it should be in the source. But you might be right that the context should be identifiable uniquely and thus index is providing only noise into it ...

jondequinor: > There is still time to fix it. I just thought that since `index` was on the same semantic level as `id` therefore it should be in the source. But you might be right that the context should be identifiable uniquely and thus index is providing only noise into it ...

I can go either way with this one, since it is down to semantics. I can create an issue and let's discuss it there?

PR #3170, opened 2022-03-30
Api adoption

DanSava:
**Issue**
Rebase on main


DanSava:
Will create new feature branch

PR #3171, opened 2022-03-30
Fix AnalysisIterConfig._repr_() and add test

### LINKED ISSUE #2559: repr(AnalysisIterConfig()) is broken
calling `repr` on `AnalysisIterConfig` fails with `TypeError('%x format: an integer is required, not method') raised in repr()` due to self._adress being a method and not a property:

https://github.com/equinor/ert/blob/3f84b979b985376aade9203a6b977b9b541d8c14/res/enkf/analysis_iter_config.py#L110-L118

Should be 
```python
        return ret % (its, rets, fmt, cfs, self._address())
```

BjarneHerland:
Resolves #2559

### DIFF
--- a/tests/libres_tests/res/enkf/test_analysis_iter_config.py
+++ b/tests/libres_tests/res/enkf/test_analysis_iter_config.py
@@ -23,6 +23,7 @@ from res.enkf import AnalysisIterConfig
 class AnalysisIterConfigTest(ResTest):
     def test_set_analysis_iter_config(self):
         c = AnalysisIterConfig()
+        self.assertTrue(repr(c).startswith("AnalysisIterConfig"))
 
         self.assertFalse(c.caseFormatSet())
         c.setCaseFormat("case%d")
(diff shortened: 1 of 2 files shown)


tests/libres_tests/res/enkf/test_analysis_iter_config.py
```diff
@@ -23,6 +23,7 @@
 class AnalysisIterConfigTest(ResTest):
     def test_set_analysis_iter_config(self):
         c = AnalysisIterConfig()
+        self.assertTrue(str(c).startswith("AnalysisIterConfig"))
```
sondreso, line 26: Maybe test repr directly? str(obj) falls back to __repr__ when __str__ is not implemented
```suggestion
        self.assertTrue(repr(c).startswith("AnalysisIterConfig"))
```

BjarneHerland: 👍

sondreso (approved):
LGTM!

PR #3185, opened 2022-03-31
Remove .index files from block_fs

pinkwah:
This commit removes the `.index` files that are created by BlockFS when the
object is destructed. This has caused flakiness in our Python code, where the
object is destructed after the testing suite changes directories. This leads to
a `util_abort` since the directory we're trying to write to no longer exists.
Instead of making failure during creation of this file silent, or otherwise
working around it, we've decided to remove the functionality altogether.

This is mostly due to how, in garbage collected languages like Python, it's
inadvisable to have program-crashing side-effects that happen on object deletion
(via the `__del__` instance method). And it's in general surprising to the
programmer that a file is created when the object is deleted.

Furthermore, the functionality of it doesn't allow read-only clients to read
up-to-date data. The reason is that the `.index` file represents the entirety of
the saved data. Data might have been written to the `.data_0` file, but it will
not be seen by other clients until the `.index` file is written at the end of
operation. This could be a benefit, but the benefit of allowing clients to poll
this file while ERT is running and writing data (eg. while waiting for all
forward models to complete) is greater.

It should be noted that the `.data_0` already contains all of the data that is
found in `.index`, where the latter is meant to be a "quick access" structure.
As we see in the performance section, just loading index data from `.data_0` is
not that much slower.

Forward-compatibility is kept. If an older version of ERT doesn't find the
`.index` file, it resorts to loading the index from the `.data_0` file. As such,
this commit removes the `.index` if present, in case the user wants to open the
storage with an older ERT.

---

# Performance

The following data sets were used to benchmark. Each data set is made by
creating a new `block_fs` file with `#Blocks` blocks (aka 'nodes' aka 'files').
Each block is named `TEST.%d` where `%d` is a counter. The size for each block
is given by sampling from a discrete uniform distribution with a default seed.

| Dataset  | #Blocks | Min block size | Max block size | Approximate data file size |
|----------|---------|----------------|----------------|----------------------------|
| `small`  | 100k    | 10 B           | 20 B           | 6.2 MB                     |
| `medium` | 10k     | 100 B          | 1 KB           | 5.9 MB                     |
| `large`  | 1k      | 1 MB           | 10 MB          | 5.2 GB                     |

The tests were ran once with `.index` and once without `.index` using the Catch
2 benchmarking suite, which runs 100 iterations in total.

## Personal development machine
This was ran on my personal machine.
- **OS**: Fedora 35
- **CPU**: Intel i5 4690K (factory clock, performance governor)
- **RAM**: 24GB 1333MHz DDR3
- **Disk**: Samsung EVO 850 m.2 SSD

The following were the loading times:
| Dataset  | with `.index` (mean, std dev) | without `.index` (mean, std dev) | Ratio (with / without) |
|----------|-------------------------------|----------------------------------|------------------------|
| `small`  | 86.818 ms, 1.4040 ms          | 208.98 ms, 1.5176 ms             | 0.415                  |
| `medium` | 9.2418 ms, 382.18 us          | 18.486 ms, 781.89 us             | 0.500                  |
| `large`  | 1.1754 ms, 738.74 us          | 2.4453 ms, 124.33 us             | 0.481                  |

## Equinor RGS node
This was ran on a Stavanger RGS node. Each test was ran on a fresh RGS node to
simulate cold open as best as was possible.
- **OS**: RHEL 7.9
- **CPU**: Intel Xeon E5-2667 v3 (VMWare virtualisation)
- **RAM**: 120GB
- **Disk**: `/private/zom` NFS disk

The following were the loading times:
| Dataset  | with `.index` (mean, std, max) | without `.index` (mean, std, max) | Ratio (with / without) |
|----------|--------------------------------|-----------------------------------|------------------------|
| `small`  | 148.68 ms, 19.421 ms           | 304.81 ms, 27.823 ms              | 0.488                  |
| `medium` | 26.575 ms, 4.3172 ms           | 30.352 ms, 10.441 ms              | 0.876                  |
| `large`  | 6.9251 ms, 155.39 us           | 5.025 ms, 394.29 us               | 1.38                   |

The `large` sample is seemingly faster without `.index`. This is likely due to
the relatively low number of iterations and high variance. That is, for few and
large blocks, the performance is about the same.

## Conclusion
The speed during load (calling `block_fs_mount`) is about half with this PR.
However, the rest of the code functions the same. The load happens once opening
a 'case'/'ensemble'. So, for running ES MDA with 4 iterations, the slowdown will
occur a grand total of 4 times. The same applies to QtPlotter / webviz-ert. I
reckon we can resolve the slowdowns in the future by using a proper /
centralised database for storing data in the future. (Although note that HTTP
could potentially be even slower)

Also, consider that the worst case scenario is where there are a lot of files.
The size matters little, as large portions of the data file are simply skipped
in the no-`.index` case. Having 100k blocks in a single case is unrealistic in
practice, since each of these have to be entered manually by the user into their
ERT config. Afaik, the design matrix isn't actually internalised in EnKF FS, but
I might be wrong.

## Appendix
Code for the benchmark

<details><summary>`libres/create_bfs.cpp`</summary>
This creates the block fs. Ran only once.

```c++
#include <filesystem>
#include <fstream>
#include <random>

#include <fmt/format.h>
#include <ert/res_util/block_fs.hpp>

using namespace std::string_literals;
namespace fs = std::filesystem;

void create(const std::string &base_name, const int nblocks,
            const size_t min_size, const size_t max_size) {
    fmt::print("Create '{}', nblocks={}, min_size={}, max_size={}\n", base_name,
               nblocks, min_size, max_size);

    std::error_code ec;
    fs::remove(base_name + ".mnt"s, ec);
    fs::remove(base_name + ".data_0"s, ec);
    fs::remove(base_name + ".index"s, ec);

    auto bfs =
        block_fs_mount((base_name + ".mnt"s).c_str(), 64 /* block_size */,
                       1.0 /* fragmentation_limit */, 10 /* fsync_interval */,
                       false /* read_only */, false /* use_lockfile */);

    std::ifstream urandom{"/dev/urandom"};
    std::vector<char> urandom_data;
    std::default_random_engine random_engine{};

    std::uniform_int_distribution size_dist{min_size, max_size};
    for (int i{}; i < nblocks; ++i) {
        auto fname = fmt::format("TEST.{}", i);
        size_t fsize = size_dist(random_engine);
        urandom_data.resize(fsize);
        urandom.read(urandom_data.data(), urandom_data.size());
        block_fs_fwrite_file(bfs, fname.c_str(), urandom_data.data(),
                             urandom_data.size());
    }

    block_fs_close(bfs, false /* unlink_empty */);
}

int main() {
    create("small", /* nblocks */ 100'000, /* min_size */ 10,
           /* max_size */ 20);
    create("medium", /* nblocks */ 10'000, /* min_size */ 100,
           /* max_size */ 1000);
    create("large", /* nblocks */ 1000, /* min_size */ 1'000'000,
           /* max_size */ 10'000'000);
}
```
</details>

<details><summary>`libres/load_bfs.cpp`</summary>
The actual benchmark.

```c++
#include <catch2/catch.hpp>
#include <ert/res_util/block_fs.hpp>

TEST_CASE("BlockFs") {
    BENCHMARK("open and close") {
        std::string base_name = GENERATE("small", "medium", "large");

        auto bfs = block_fs_mount(
            (base_name + ".mnt").c_str(), 64 /* block_size */, 1.0 /* fragmentation_limit */,
            10 /* fsync_interval */, false /* read_only */,
            false /* use_lockfile */);
        block_fs_close(bfs, false /* unlink_empty */);
    };
}
```
</details>

<details><summary>`libres/CMakeLists.txt`</summary>
Diff required to have CMake compile the new files.

```diff
@@ -97,6 +97,7 @@ conan_cmake_run(
   # Options
   OPTIONS
   catch2:with_main=True
+  catch2:with_benchmark=True
   ${conan_opts}
   # Force cppstd to be the same as this CMakeLists.txt's
   SETTINGS
@@ -150,3 +151,11 @@ enable_testing()
 add_subdirectory(lib)
 add_subdirectory(old_tests)
 add_subdirectory(tests)
+
+add_executable(create_bfs create_bfs.cpp)
+target_link_libraries(create_bfs res)
+
+find_package(Catch2 REQUIRED)
+include(Catch)
+add_executable(load_bfs load_bfs.cpp)
+target_link_libraries(load_bfs res Catch2::Catch2WithMain)
```
</details>

### DIFF
--- a/libres/lib/res_util/block_fs.cpp
+++ b/libres/lib/res_util/block_fs.cpp
@@ -37,8 +37,6 @@ namespace fs = std::filesystem;
 
 #define MOUNT_MAP_MAGIC_INT 8861290
 #define BLOCK_FS_TYPE_ID 7100652
-#define INDEX_MAGIC_INT 1213775
-#define INDEX_FORMAT_VERSION 1
 
 /*
   During mounting a significant part of the time is spent on filling
@@ -123,7 +121,6 @@ struct block_fs_struct {
 
     char *data_file;
     char *lock_file;
-    char *index_file;
 
     int data_fd;
     FILE *data_stream;
@@ -150,6 +147,13 @@ struct block_fs_struct {
                                             fragmentation_limit == 0.0 : Rotate when one byte is wasted. */
     bool data_owner;
     int fsync_interval; /* 0: never  n: every nth iteration. */
+
+    fs::path index_file() const {
+        fs::path mount_file{this->mount_file};
+        fs::path path = mount_file.parent_path();
+        std::string base_name = mount_file.stem();
+        return path / (base_name + ".index");
+    }
 };
 
 static void block_fs_rotate__(block_fs_type *block_fs);
@@ -523,14 +527,11 @@ static void block_fs_set_filenames(block_fs_type *block_fs) {
 
     free(block_fs->data_file);
     free(block_fs->lock_file);
-    free(block_fs->index_file);
 
     block_fs->data_file =
         util_alloc_filename(block_fs->path, block_fs->base_name, data_ext);
     block_fs->lock_file =
         util_alloc_filename(block_fs->path, block_fs->base_name, lock_ext);
-    block_fs->index_file =
-        util_alloc_filename(block_fs->path, block_fs->base_name, index_ext);
 
     free(data_ext);
     free(lock_ext);
@@ -582,7 +583,6 @@ static block_fs_type *block_fs_alloc_empty(const char *mount_file,
     }
     block_fs->data_file = NULL;
     block_fs->lock_file = NULL;
-    block_fs->index_file = NULL;
     block_fs_reinit(block_fs);
 
     {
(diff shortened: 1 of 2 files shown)


BjarneHerland (approved):
Looks solid! 👍

PR #3186, opened 2022-03-31
Modify ModelConfig.enspath to hold an absolute file-path

BjarneHerland:
Filenames in `block_fs_driver` are derived from this field, which currently is relative to the directory in which the config-file resides. If a process changes directory after instantiating `ModelConfig`, existing `block_fs_driver`-instances refer to wrong filenames on disk.

1. The `block_fs_driver` dtor currently modifies / remove files which may (currently) happen on Python GC, i.e. outside of our control

2. Our Python-based tests run in separate tmp-directories with no guarantee of GC between

Combining 1 and 2 above with relative paths opens up for flakiness in the tests. By using absolute paths we avoid this.

### DIFF
--- a/libres/lib/enkf/model_config.cpp
+++ b/libres/lib/enkf/model_config.cpp
@@ -253,7 +253,7 @@ void model_config_set_max_internal_submit(model_config_type *model_config,
     model_config->max_internal_submit = max_resample;
 }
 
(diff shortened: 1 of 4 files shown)


BjarneHerland:
test flaky please

BjarneHerland:
@pinkwah any other comments?

PR #3189, opened 2022-04-01
Have repeated flaky tests run new python instance

lars-petter-hauge:
**Approach**
Garbage cleaning and session controlled objects in tests could reduce the usage of `pytest-repeat`. Instead we completely run a new python instance on each try. 

## Pre review checklist

### DIFF
--- a/ci/jenkins/testkomodo-repeat-flaky.sh
+++ b/ci/jenkins/testkomodo-repeat-flaky.sh
@@ -22,29 +22,50 @@ install_test_dependencies () {
     pip install pytest-repeat
 }
 
+run_libres_ert_tests(){
+    # If a developer would like to test a specific function, remove these
+    # lines and include the specific tests one would like to run.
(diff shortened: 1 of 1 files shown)


lars-petter-hauge:
Tested that the script runs as expected on internal CI

BjarneHerland (changes requested):
Looks good but could we comment in run_libres_ert_tests() why certain tests are masked out? Ideally, there should also be some comment/placeholder for how/where I tailor this script to my own PR in case I want to stress-test some particular test.

lars-petter-hauge:
I used the same test setup that we are already running - not that that is any excuse. There are no comments on those tests.

I'll do some checking and see if we can add the reason as to why they should be ignored (and if it is because of flakyness we can simply remove it, as we are in fact looking for it).

I can add a comment on how one would swap out for a specific test

BjarneHerland (approved):
👍

lars-petter-hauge:
test ert please

PR #3190, opened 2022-04-01
Drop Title Case In Ert Gui

berland:
Title Case were inconsistently in use in various GUI elements. This commit
removes All Of Them, except Ensemble Reservoir Toolbox.


sondreso (approved):
> _Consistency is the foundation of virtue._

- Francis Bacon

PR #3193, opened 2022-04-03
Change default inversion to IES_INVERSION_EXACT

### LINKED ISSUE #3195: Change default inversion to `IES_INVERSION_EXACT`
As long as we are using a diagonal observation error matrix (R in code), we should use `IES_INVERSION_EXACT` as it is more efficient, and does not require truncation of eigenvalue spectrum.

The current default is IES_INVERSION_SUBSPACE_EXACT_R.

dafeda:
As long as we are using diagonal observation error matrices R,
we should use exact inversion because it does not entail truncation
of the eigenvalue spectrum, and is faster.

**Issue**
Resolves #3195

## Pre review checklist


ManInFez (approved):
ship it, but fix HM tutorial also!

PR #3194, opened 2022-04-03
Make `block_fs` append-only + other refactorings

pinkwah:
Block FS has support for data deletion, as well as data replacement. The
deletion part via the function `enkf_fs_unlink` was removed a while ago
since the functionality was unused. It is also possible for data to be
replaced. I don't believe this functionality is actually used, but it's
been difficult to prove it due to the amount of locations from which the
`block_fs` writing function can be accessed (including from Python).

The places that are of interest, namely writing of parameters to
PARAMETER and response data to RESPONSE after a successful ERT run are
written once and only once. That is, by the time ERT knows about FOPR
from a forward-model, it has already internalised all of the data, and
thus won't need to rewrite the data. Essentially, as far as I can tell,
ERT doesn't make use of this "partial" data writing functionality.
However, due to the breadth of sources from which the writing
function can be called, I can't say this for certain.

However, by making `block_fs` append-only, we make it easy to discover
new data as it's being written. It makes it possible to poll the
file (by eg. dark storage) and if the file has been updated, one only
has to look at the end of the file to find out what's new. This could
potentially make it easier to update visualisations before all forward
models complete.

Compatibility with older ERTs are kept: Should I be incorrect in my
assumption that no data is being rewritten, storages generated by older
ERTs could have gaps in the data file marked with `NODE_FREE`, which is
safe to ignore. If this append-only variant has data that is rewritten,
it'll appear twice in the file as `NODE_IN_USE`. Older ERT would read
both, but due to the current one appearing last, that'll be the one that
ERT will internalise in its dictionary.

### DIFF
--- a/libres/lib/res_util/block_fs.cpp
+++ b/libres/lib/res_util/block_fs.cpp
@@ -30,7 +30,7 @@
 
 #include <fmt/ostream.h>
 
-#include <ert/util/buffer.hpp>
+#include <ert/python.hpp>
 #include <ert/util/hash.hpp>
 #include <ert/util/vector.hpp>
 
@@ -61,7 +61,6 @@ namespace fs = std::filesystem;
 */
 
 #define NODE_IN_USE_BYTE 85 /* Binary(85)  =  01010101 */
-#define NODE_FREE_BYTE 170  /* Binary(170) =  10101010 */
 #define WRITE_START__ 77162
 
 static const int NODE_END_TAG =
@@ -72,30 +71,11 @@ static const int NODE_WRITE_ACTIVE_END = 776512;
 typedef enum {
     NODE_IN_USE =
         1431655765, /* NODE_IN_USE_BYTE * ( 1 + 256 + 256**2 + 256**3) => Binary 01010101010101010101010101010101 */
-    NODE_FREE =
-        -1431655766, /* NODE_FREE_BYTE   * ( 1 + 256 + 256**2 + 256**3) => Binary 10101010101010101010101010101010 */
     NODE_WRITE_ACTIVE = WRITE_START__, /* This */
     NODE_INVALID = 13 /* This should __never__ be written to disk */
 } node_status_type;
 
-/*
-   The free_node_struct is used to implement a doubly linked list of
-   free nodes; i.e. holes in the file which are available for other use.
-*/
 typedef struct file_node_struct file_node_type;
-typedef struct free_node_struct free_node_type;
-
-struct free_node_struct {
-    free_node_type *next;
-    free_node_type *prev;
-    file_node_type *file_node;
-};
-
-/*
-  Datastructure representing one 'block' in the datafile. The block
-  can either refer to a file (status == NODE_IN_USE) or to an empty
-  slot in the datafile (status == NODE_FREE).
-*/
 
 struct file_node_struct {
     long int
@@ -104,14 +84,8 @@ struct file_node_struct {
     int node_size; /* The size in bytes of this node - must be >= data_size. NEVER Changed. */
     int data_size; /* The size of the data stored in this node - in addition the node might need to store header information. */
     node_status_type
-        status; /* This should be: NODE_IN_USE | NODE_FREE; in addition the disk can have NODE_WRITE_ACTIVE for incomplete writes. */
+        status; /* This should be: NODE_IN_USE; in addition the disk can have NODE_WRITE_ACTIVE for incomplete writes. */
 };
-
-/*
-   data_size   : manipulated in block_fs_fwrite__() and block_fs_insert_free_node().
-   status      : manipulated in block_fs_fwrite__() and block_fs_unlink_file__();
-   data_offset : manipulated in block_fs_fwrite__() and block_fs_insert_free_node().
-*/
 
 struct block_fs_struct {
     UTIL_TYPE_ID_DECLARATION;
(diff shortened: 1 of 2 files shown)


libres/lib/res_util/block_fs.cpp
```diff
@@ -130,8 +133,7 @@ struct block_fs_struct {
     int block_size;          /* The size of blocks in bytes. */
     int lock_fd; /* The file descriptor for the lock_file. Set to -1 if we do not have write access. */
 
-    pthread_mutex_t io_lock;  /* Lock held during fread of the data file. */
-    pthread_rwlock_t rw_lock; /* Read-write lock during all access to the fs. */
+    std::mutex mutex;
```
BjarneHerland, line 106: Note that a simple mutex is not able to distinguish between write-access and read-access like `pthread_mutex_t` does.

You might want to use a [`shared_mutex`](https://en.cppreference.com/w/cpp/thread/shared_mutex)
and use `lock_shared` for read-access.

OTOH, if you feel this is unnecessary granularity, feel free to argue the fact.

libres/tests/enkf/test_enkf_fs.cpp
```diff
@@ -81,7 +80,7 @@ TEST_CASE("block_fs", "[enkf_fs]") {
                 REQUIRE(!block_fs_has_file(bfs, "BAR"));
             }
 
-            AND_THEN("data can be read from the same instance") {
+            THEN("data can be read from the same instance") {
```
BjarneHerland, line 82: Why remove dependency? Wouldn't read always fail if the exists-test fails?

libres/tests/enkf/test_enkf_fs.cpp
```diff
@@ -93,14 +92,22 @@ TEST_CASE("block_fs", "[enkf_fs]") {
...
+                            "tried to write to read only filesystem"));
+                }
 
                 THEN("data exists") {
                     REQUIRE(block_fs_has_file(bfs, "FOO"));
                     REQUIRE(!block_fs_has_file(bfs, "BAR"));
                 }
 
-                AND_THEN("data can be read") {
+                THEN("data can be read") {
```
BjarneHerland, line 109: Like above

libres/lib/res_util/block_fs.cpp
pinkwah: I feel this is unnecessary granularity, yes.

`block_fs_has_file` is used only by `ert::block_fs_driver::has_node`, which is only used by `enkf_fs_has_node`, which is used by two tests. But I reckon this is reasonable functionality to keep.

In all other instances it's actually touching the file handle, and you need exclusive access to do this.

libres/lib/res_util/block_fs.cpp
```diff
@@ -107,12 +93,6 @@ struct file_node_struct {
         status; /* This should be: NODE_IN_USE | NODE_FREE; in addition the disk can have NODE_WRITE_ACTIVE for incomplete writes. */
```
BjarneHerland, line 93: NODE_FREE mentioned

libres/lib/res_util/block_fs.cpp
```diff
@@ -549,9 +416,9 @@ static void block_fs_fseek_node_data(block_fs_type *block_fs,
 
 /*
    This function will read through the datafile seeking for one of the
-   identifiers: NODE_IN_USE | NODE_FREE. If one of the valid status
-   identifiers is found the stream is repositioned at the beginning of
-   the valid node, so the calling scope can continue with a
+   identifiers: NODE_IN_USE. If one of the valid status identifiers is found the
```
BjarneHerland, line 419: Rephrase "for one of the..."

libres/lib/res_util/block_fs.cpp
```diff
@@ -564,7 +431,7 @@ static bool block_fs_fseek_valid_node(block_fs_type *block_fs) {
     int status;
     while (true) {
         if (fread(&byte, sizeof byte, 1, block_fs->data_stream) == 1) {
-            if (byte == NODE_IN_USE_BYTE || byte == NODE_FREE_BYTE) {
+            if (byte == NODE_IN_USE_BYTE) {
```
BjarneHerland, line 434: Can we be a little smarter (+ make the code a little clearer) here...  reading first a single byte, then perhaps an integer seems a little over the top?

libres/lib/res_util/block_fs.cpp
```diff
@@ -725,9 +536,6 @@ static void block_fs_build_index(block_fs_type *block_fs,
                         block_fs_insert_index_node(block_fs, filename,
                                                    file_node);
                         break;
-                    case (NODE_FREE):
```
BjarneHerland, line 809: Only two possibilities - replace switch with if?

libres/lib/res_util/block_fs.cpp
pinkwah: Not sure what you mean. Are you suggesting we don't read the byte but just attempt to read the int to begin with?

libres/lib/res_util/block_fs.cpp
BjarneHerland: In other words, when existing code asks for read-only access (`block_fs->io_lock`) it already holds the `rw_lock` and thus has exclusive access. That's fair enough.

libres/lib/res_util/block_fs.cpp
BjarneHerland: I assume this file can be addressed byte-wise, hence we cannot just read and check the full int. But instead of reading one byte, then fseek back one byte and read the int, we might read some reasonable-sized block into memory, scan it for `NODE_IN_USE` and fseek() to the right pos if required. Would probably be quite a lot faster, but perhaps more suitable for a different PR...

BjarneHerland (changes requested):
In general LGTM. But would it be possible to add info-level logging to collect information from real usage (aka telemetry) whether anyone actually rewrite blocks? In addition, is it possible to add a test to back up this statement from commit-msg:

`Older ERT would read both, but due to the current one appearing last, that'll be the one that ERT will internalise in its dictionary.`

pinkwah:
I have updated this PR:
- Rebased and kept only the last commit which we have yet to merge
- Added test which checks that overwriting works.
- Added logging statement

BjarneHerland (approved):

PR #3201, opened 2022-04-04
Refactor `StateMap` into C++ & pybind11

pinkwah:
WIP on the way to turning `EnkfFs` into a C++ & pybind11 class. The next mini step is to convert `TimeMap`, which should be enough to convert the rest of `EnkfFs` into something more manageable. The end goal is to get rid of this ref/deref stuff and use `std::shared_ptr` with a well-behaved C++ binding. The other end goal is to allow for a better `EnkfFs` API. Basically, reduce all of the complexity down to `storage.load_parameters("FOPR")`[^1] which returns a `np.ndarray`.

[^1]: I think `BlockStorage` is a more descriptive name than `EnkfFs`, considering that the latter is neither a filesystem but a less advanced DOOM (1993) WAD, and also there is no ensemble kalman filter anymore so the prefix makes little sense.

### DIFF
--- a/src/libres/lib/analysis/update.cpp
+++ b/src/libres/lib/analysis/update.cpp
@@ -108,8 +108,7 @@ void deserialize_node(enkf_fs_type *fs,
     // deserialize the matrix into the node (and writes it to the fs)
     enkf_node_deserialize(node, fs, node_id, active_list, A, row_offset,
                           column);
-    state_map_update_undefined(enkf_fs_get_state_map(fs), iens,
-                               STATE_INITIALIZED);
+    enkf_fs_get_state_map(fs).update_undefined(iens, STATE_INITIALIZED);
     enkf_node_free(node);
 }
 
@@ -268,10 +267,9 @@ void copy_parameters(enkf_fs_type *source_fs, enkf_fs_type *target_fs,
             enkf_node_free(data_node);
         }
 
-        state_map_type *target_state_map = enkf_fs_get_state_map(target_fs);
-        state_map_set_from_inverted_mask(target_state_map, ens_mask,
-                                         STATE_PARENT_FAILURE);
-        state_map_set_from_mask(target_state_map, ens_mask, STATE_INITIALIZED);
+        auto &target_state_map = enkf_fs_get_state_map(target_fs);
+        target_state_map.set_from_inverted_mask(ens_mask, STATE_PARENT_FAILURE);
+        target_state_map.set_from_mask(ens_mask, STATE_INITIALIZED);
         enkf_fs_fsync(target_fs);
     }
 }
--- a/src/libres/lib/enkf/enkf_fs.cpp
+++ b/src/libres/lib/enkf/enkf_fs.cpp
@@ -16,6 +16,7 @@
    for more details.
 */
 
+#include "ert/python.hpp"
 #include <filesystem>
 #include <memory>
 #include <string>
@@ -153,7 +154,7 @@ struct enkf_fs_struct {
     /** Whether this filesystem has been mounted read-only. */
     bool read_only;
     time_map_type *time_map;
-    state_map_type *state_map;
+    std::shared_ptr<StateMap> state_map;
     summary_key_set_type *summary_key_set;
     /* The variables below here are for storing arbitrary files within the
      * enkf_fs storage directory, but not as serialized enkf_nodes. */
@@ -211,7 +212,7 @@ enkf_fs_type *enkf_fs_alloc_empty(const char *mount_point, bool read_only) {
     enkf_fs_type *fs = new enkf_fs_type;
     UTIL_TYPE_ID_INIT(fs, ENKF_FS_TYPE_ID);
     fs->time_map = time_map_alloc();
-    fs->state_map = state_map_alloc();
+    fs->state_map = std::make_shared<StateMap>();
     fs->summary_key_set = summary_key_set_alloc();
     fs->misfit_ensemble = misfit_ensemble_alloc();
     fs->read_only = true;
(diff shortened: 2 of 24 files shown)


pinkwah:
Jenkins test this please

libres/lib/enkf/state_map.cpp
```diff
@@ -136,140 +84,141 @@ bool state_map_legal_transition(realisation_state_enum state1,
...
+    m_assert_writable();
+    std::lock_guard lock{m_mutex};
+
+    if (index < 0)
+        index += m_state.size();
+    if (index < 0)
+        throw std::out_of_range(
+            fmt::format("index out of range: {} < 0", index));
+    if (index >= m_state.size())
+        m_state.resize(index + 1, STATE_UNDEFINED);
```
oyvindeide, line 103: I know this is the old behavior, but it might makes sense to at least write an issue with future changes, this being one of the key ones, it is quite confusing behavior.

res/enkf/state_map.py
```diff
@@ -1,137 +1,54 @@
...
+    """
+    Will create a list of all realisations with state equal to state_value.
+
+    @type state_value: RealizationStateEnum
+    @rtype: ecl.util.IntVector
+    """
+    return [index for index, value in enumerate(self) if value == state_value]
+
+
+def createMask(self, state_value):
```
ManInFez, line 32: The old behavior returned a list of bools not a "BoolVector"

libres/lib/enkf/state_map.cpp
pinkwah: https://github.com/equinor/ert/issues/3493

res/enkf/state_map.py
```diff
@@ -1,137 +1,54 @@
...
+def __iter__(self):
+    index = 0
+    size = len(self)
+
+    while index < size:
+        yield self[index]
+        index += 1
+
+
+def realizationList(self, state_value):
```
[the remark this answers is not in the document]
pinkwah: ditto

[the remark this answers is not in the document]
ManInFez: could we keep the typing

res/enkf/state_map.py
ManInFez: could we keep the typing

libres/lib/enkf/enkf_fs.cpp
```diff
@@ -432,7 +434,11 @@ static void enkf_fs_fread_cases_config(enkf_fs_type *fs) {
 
 static void enkf_fs_fread_state_map(enkf_fs_type *fs) {
     char *filename = enkf_fs_alloc_case_filename(fs, STATE_MAP_FILE);
-    state_map_fread(fs->state_map, filename);
+    try {
+        fs->state_map->read(filename);
+    } catch (const ert::read_error &) {
+        /* Exception is ignored. StateMap is reset */
```
oyvindeide, line 440: Wouldnt it be more readable to but the resetting here instead of where the exception is thrown?

res/enkf/state_map.py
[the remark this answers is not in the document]
pinkwah: oki

res/enkf/state_map.py
pinkwah: oki

libres/lib/python/init.cpp
```diff
@@ -22,6 +23,18 @@ ert::detail::Submodule::Submodule(const char *path, init_type &init)
 }
 
 PYBIND11_MODULE(_lib, m) {
+    py::register_exception_translator([](std::exception_ptr p) {
+        if (!p)
+            return;
+
+        try {
+            std::rethrow_exception(p);
+        } catch (const std::ios_base::failure &e) {
```
oyvindeide, line 32: Is this functionality tested on the python side?

libres/lib/enkf/enkf_fs.cpp
```diff
@@ -344,7 +345,11 @@ static void enkf_fs_fread_time_map(enkf_fs_type *fs) {
 
 static void enkf_fs_fsync_state_map(enkf_fs_type *fs) {
     char *filename = enkf_fs_alloc_case_filename(fs, STATE_MAP_FILE);
-    state_map_fwrite(fs->state_map, filename);
+    try {
+        fs->state_map->write(filename);
+    } catch (std::ios_base::failure &) {
+        // Write errors are ignored
```
eivindjahren, line 351: Maybe log? Could be interesting to see on our board.

libres/lib/enkf/enkf_fs.cpp
```diff
@@ -356,7 +361,11 @@ static void enkf_fs_fsync_summary_key_set(enkf_fs_type *fs) {
 
 static void enkf_fs_fread_state_map(enkf_fs_type *fs) {
     char *filename = enkf_fs_alloc_case_filename(fs, STATE_MAP_FILE);
-    state_map_fread(fs->state_map, filename);
+    try {
+        fs->state_map->read(filename);
+    } catch (const std::ios_base::failure &) {
+        /* Read error is ignored. StateMap is reset */
```
eivindjahren, line 367: Same comment about logging

libres/lib/enkf/enkf_main.cpp
```diff
@@ -646,6 +643,9 @@ void init_active_run(const res_config_type *res_config,
 }
 } // namespace enkf_main
 
+#include "./enkf_main_ensemble.cpp"
+#include "./enkf_main_manage_fs.cpp"
+
```
eivindjahren, line 648: What is the reason for the reorder here?

libres/lib/include/ert/enkf/state_map.hpp
```diff
@@ -1,62 +1,47 @@
...
+    int size() const;
+    bool operator==(const StateMap &) const;
+
+    bool is_readonly() const { return m_read_only; };
+
+    void set(int index, realisation_state_enum new_state);
+    realisation_state_enum get(int index) const;
+    void update_matching(size_t index, int state_mask,
+                         realisation_state_enum new_state);
+    void update_undefined(size_t index, realisation_state_enum new_state) {
```
eivindjahren, line 30: What the methods of StateMap does is not completely obvious to me. Can we have some doxygen documentation?

PR #3210, opened 2022-04-05
dlopen libres with RTLD_LOCAL

pinkwah:
Using RTLD_GLOBAL adds all of the symbols within libres into the global symbol table of the running process. This can interfere with other libraries that rely on the same dependencies that we do (eg. pybind11) but use a different version. The symbols conflict and you might get a segmentation fault or worse.


pinkwah:
jenkins test this please

pinkwah:
jenkins test this please

libres/lib/job_queue/workflow_job.cpp
```diff
@@ -241,33 +304,46 @@ static void workflow_job_iset_argtype_string(workflow_job_type *workflow_job,
...
-                else
-                    fprintf(stderr, "Failed to load symbol:%s Error:%s \n",
-                            workflow_job->function, dlerror());
-            } else {
-                if (workflow_job->module != NULL)
-                    fprintf(stderr, "Failed to load module:%s Error:%s \n",
-                            workflow_job->module, dlerror());
+
+            if (!workflow_job->lib_handle) {
+                fprintf(stderr, "Failed to load symbol:%s Error:%s \n",
```
BjarneHerland, line 325: Perhaps "Failed to load module" since that is what failed?

PR #3211, opened 2022-04-05
Fix scaling of ESMDA weights

### LINKED ISSUE #3239: ESMDA scaling is wrong
Sum of reciprocals of scaled weights must sum to one, which they do not in the current implementation.

dafeda:
ESMDA requires that inverses of scaled weights sum to 1.0,
which is not always the case with the current implementation.

**Issue**
Resolves #3239

**Approach**

```python
import numpy as np
from math import sqrt

def current(weights):
    length = sqrt(sum((1.0 / x) * (1.0 / x) for x in weights))
    return np.array([x * length for x in weights])

def new(weights):
    length = sum(1.0 / x for x in weights)
    return np.array([x * length for x in weights])

weights = [1, 1]
print(current(weights))
print(np.sqrt(new(weights)))

# >> [1.4142135623730951, 1.4142135623730951]
# >> [1.41421356 1.41421356]

weights = [4, 2, 1]
print(current(weights))
print(np.sqrt(new(weights)))

# >> [4.58257569495584, 2.29128784747792, 1.14564392373896]
# >> [2.64575131 1.87082869 1.32287566]
```

## Pre review checklist

### DIFF
--- a/tests/ert_tests/gui/test_multiple_data_assimilation.py
+++ b/tests/ert_tests/gui/test_multiple_data_assimilation.py
@@ -1,66 +1,33 @@
-from ert_utils import ErtTest
+from multiprocessing.sharedctypes import Value
+import numpy as np
+import pytest
 
 from ert_shared.models import MultipleDataAssimilation as mda
 
 
-class MDAWeightsTest(ErtTest):
-    def test_weights(self):
+def test_normalized_weights():
+    weights = mda.normalizeWeights([1])
+    assert weights == [1.0]
 
-        weights = mda.parseWeights("2, 2, 2, 2")
-        print(weights)
-        self.assertAlmostEqualList([2, 2, 2, 2], weights)
+    weights = mda.normalizeWeights([1, 1])
+    assert weights == [2.0, 2.0]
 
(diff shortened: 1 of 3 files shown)


ManInFez:
test ert please!

tests/ert_tests/gui/test_multiple_data_assimilation.py
```diff
@@ -1,8 +1,21 @@
...
+    assert weights == [1.0]
+
+    weights = mda.normalizeWeights([1, 1])
+    assert weights == [2.0, 2.0]
+
+    weights = np.array(mda.normalizeWeights([8, 4, 2, 1]))
+    assert np.reciprocal(weights).sum() == 1.0
+
+
 class MDAWeightsTest(ErtTest):
```
ManInFez, line 19: could you maybe just convert this to a regular pytest and remove the need for using ErtTest?

dafeda: Done 👍

ManInFez (approved):
Good catch!

PR #3230, opened 2022-04-07
Remove flag for using aa_projection in IES and ES

### LINKED ISSUE #3198: Remove `use_aa_projection`
A^+A projection should be done when `state_size <= (ens_size - 1)` **and** the model is non-linear.
This means that `use_aa_projection` is a flag allowing users to say whether a model is linear or not.
This is not necessary.

```cpp
    if (A) {
        const int state_size = matrix_get_rows(A);
        if (use_aa_projection && (state_size <= (ens_size - 1))) {
            ies::linalg_compute_AA_projection(A, Y);
        }
    }
```

ManInFez:
AA projection should only be used when you have less parameters than
realizations and you have a linear model. We are now removing the
user config option to toggle this for those cases.

**Issue**
Resolves #3198 

## Pre review checklist


dafeda:
I think we should write that we will still do AA projection when there are less parameters than realizations, but that we will assume a non-linear model.

dafeda (approved):
Looks good 👍

PR #3245, opened 2022-04-08
Avoid DeprecationWarning from qtbot

BjarneHerland:
Avoids this warning:

`DeprecationWarning: waitForWindowShown is deprecated, as the underlying Qt method was obsoleted in Qt 5.0 and removed in Qt 6.0. Its name is imprecise and the pytest-qt wrapper does not raise TimeoutError if the window wasn't shown. Please use the qtbot.waitExposed context manager instead.`


oyvindeide (commented):
Looks good, just one minor comment

jondequinor (approved):
LGTM

PR #3251, opened 2022-04-11
Avoid BoolVector in Python code

### LINKED ISSUE #3121: Port handling of active realization masks to python
In the ERT2 gui (and cli), there is a possibility for selecting which realizations should be active and not.

The user facing portion of this is strings like "1-4,10-99" (a rangestring) that needs to be converted into a list of booleans (a mask), up to a (predefined) length. 

From a mask to a rangestring, the code is in:
https://github.com/equinor/ert/blob/main/ert_gui/ertwidgets/models/activerealizationsmodel.py#L6-L32

The reverse direction is handled by:
https://github.com/equinor/ecl/blob/5e16a6e8d9bdbd34a632e9ad0133b7babd1d0d02/lib/util/string_util.cpp#L73-L119
and https://github.com/equinor/ecl/blob/main/python/ecl/util/util/bool_vector.py

#2794 needs to support the same functionality, but does not want to drag along the `ecl` dependency, this functionality should thus be ported to Python and put somewhere accessible for both ERT2 and ERT3, and ERT2 should be refactored to use the ported code.

### LINKED ISSUE #3141: Avoid using bool_vector in simulation arguments

berland:
**Issue**
Resolves #3121
Resolves #3141

**Approach**
Replace usage of `BoolVector` with `List[bool]`except for the final call to wrapped C-code. Use new Python object `ActiveRange` for interpreting range strings.

## Pre review checklist

### DIFF
--- a/ert_gui/ertwidgets/models/activerealizationsmodel.py
+++ b/ert_gui/ertwidgets/models/activerealizationsmodel.py
@@ -1,35 +1,9 @@
+from typing import List
+
 from ecl.util.util import BoolVector
+from ert.ensemble_evaluator.activerange import ActiveRange, mask_to_rangestring
 from ert_gui.ertwidgets.models.valuemodel import ValueModel
 from ert_shared.libres_facade import LibresFacade
-
-
-def mask_to_rangestring(mask):
-    """Convert a mask (ordered collection of booleans) into a range string
-
-    For instance, `0 1 0 1 1 1` would be converted to `1, 3-5`
-    """
-    ranges = []
-
-    def store_range(begin, end):
-        if end - begin == 1:
-            ranges.append("{}".format(begin))
-        else:
-            ranges.append("{}-{}".format(begin, end - 1))
-
-    start = None
-    for i, is_active in enumerate(mask):
-        if is_active:
-            if start is None:  # begin tracking a range
-                start = i
-            assert start is not None
-        else:
-            if start is not None:  # store the range and stop tracking
-                store_range(start, i)
-                start = None
-            assert start is None
-    if start is not None:  # complete the last range if any
-        store_range(start, len(mask))
-    return ",".join(ranges)
 
 
 class ActiveRealizationsModel(ValueModel):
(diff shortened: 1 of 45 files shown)


ert_shared/libres_facade.py
```diff
@@ -83,7 +84,7 @@ def run_path(self):
         return self._enkf_main.getModelConfig().getRunpathAsString()
 
     def load_from_forward_model(
-        self, case: str, realisations: BoolVector, iteration: int
+        self, case: str, realisations: List[bool], iteration: int
```
xjules, line 87: I see that we use both spelling for `realization` and `realisation`. Maybe we should unify them at same point.

ert_shared/models/multiple_data_assimilation.py
```diff
@@ -254,11 +254,10 @@ def create_context(
         if initialize_mask_from_arguments:
             mask = self._simulation_arguments["active_realizations"]
         else:
-            state = (
+            mask = sim_fs.getStateMap().createMask(
```
xjules, line 257: You've removed intermediate `state` here, but not in EnsembleSmoother i.e. `ert_shared/models/ensemble_smoother.py` :)

ert_shared/libres_facade.py
berland: Yes, that would be an issue of its own.

ert_shared/models/multiple_data_assimilation.py
berland: fixed!

xjules (approved):
Good job! LGTM :rocket:

ert_gui/ertwidgets/models/activerealizationsmodel.py
```diff
@@ -57,14 +31,7 @@ def getDefaultValue(self):
...
-
-        mask = BoolVector(default_value=False, initial_size=count)
-        if not mask.updateActiveMask(self.getValue()):
-            raise ValueError('Error while parsing range string "%s"!' % self.getValue())
-
-        if len(mask) != count:
-            raise ValueError("Mask size changed %d != %d!" % (count, len(mask)))
-
-        return mask
+    def getActiveRealizationsMask(self) -> List[bool]:
```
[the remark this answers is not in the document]
berland: That would incur a slight API break?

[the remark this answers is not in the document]
sondreso: I think that is ok, these modules are not considered to be part of the public API of ERT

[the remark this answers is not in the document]
berland: Skipping, as it belongs to a future PR doing all function names in this class

berland:
jenkins test pleas

PR #3231, opened 2022-04-07
Replace deprecated `..index.is_all_dates` in plottery

oysteoh:
**Issue**
`..index.is_all_dates` is deprecated and needs to be replaced. 

**Approach**
Extracted functionality making the check and verified the behavior with test.

## Pre review checklist

- [x] Added appropriate release note label
- [x] PR title captures the intent of the changes, and is fitting for release notes.
- [x] Commit history is consistent and clean, in line with the [contribution guidelines](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md).

Adding labels helps the maintainers when writing release notes. This is the [list of release note labels](https://github.com/equinor/ert/labels?q=release-notes).

PR #3265, opened 2022-04-19
Replace deprecated `..index.is_all_dates` in plottery

oysteoh:
This PR adjusting a PR merged too quickly ( #3231 ). The test is
now removed as it is just testing pandas functionality. The util
method is not needed and the call to `..data.index.inferred_type`
is done directly where it should be used.


ert_gui/plottery/plots/ensemble.py
```diff
@@ -25,7 +24,7 @@ def plot(self, figure, plot_context, case_to_data_map, observation_data):
             data = data.T
 
             if not data.empty:
-                if not index_is_datetime(data):
+                if not data.index.inferred_type == "datetime64":
```
BjarneHerland, line 27: `!=`  perhaps?

ert_gui/plottery/plots/statistics.py
```diff
@@ -27,7 +26,7 @@ def plot(self, figure, plot_context, case_to_data_map, _observation_data):
         for case, data in case_to_data_map.items():
             data = data.T
             if not data.empty:
-                if not index_is_datetime(data):
+                if not data.index.inferred_type == "datetime64":
```
BjarneHerland, line 29: like above

BjarneHerland (approved):
A few nits but all-in-all good stuff  😄

ert_gui/plottery/plots/ensemble.py
oysteoh: YES! absolutely... :see_no_evil:

PR #3284, opened 2022-04-21
Enable `SortInclude` in clang-format configuration

pinkwah:
Resolves: #2932


oyvindeide (commented):
Changes look good, but can you either merge the commits or reorder them? Currently the test will not pass on all commits as the requirement comes before the formatting.

oyvindeide (approved):

PR #3286, opened 2022-04-21
Remove `setup_requires` from `setup.py`

pinkwah:
Resolves: #668


oyvindeide (approved):
LGTM! Just not that this: https://github.com/equinor/ert/pull/3285 does some of the same, though it is a bit different.

PR #3301, opened 2022-04-26
Remove the current parameter group concept

### LINKED ISSUE #3295: Revise the "parameter group" concept in ert3
After #2934 has been merged, the "parameter group" concept makes less sense than before, wherever it is mentioned. 

One should go through the output of `grep -r -i "parameter group"` and update the error messages and the documentation and make them more fitting for the current implementation. Roughly speaking, parameter groups should now be called parameters (but a parameter can be both a list-thing and a dict-thing, and if it is a dict-thing, a single parameter can contain many "variables").

berland:
What was previously a parameter group in ert3 is now just a parameter,
but it may contain multiple values (list-like or dict-like).

The command line API is altered, in the positional argument name to `ert3 record sample`.

**Issue**
Resolves #3295 

**Approach**
`grep -r group` and similar, rewritten wherever applicable.

## Pre review checklist

### DIFF
--- a/ert3/config/_parameters_config.py
+++ b/ert3/config/_parameters_config.py
@@ -148,7 +148,7 @@ class _VariablesConfig(_ParametersConfig):
             return variables
 
         raise ValueError(
-            "Parameter group cannot have empty variable list.\n"
+            "A parameter cannot have an empty variable list.\n"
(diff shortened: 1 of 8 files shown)


sondreso (approved):
LGTM!

Issue #3299, opened 2022-04-25
Only plot one historical vectors across all realizations if they are equal


oysteoh:
Pyqt plotter is currently adding the historical vector from the refcase and plot both the summaryvector and - if any - a historical vector. The summaryvector should be removed, ref https://github.com/equinor/ert/issues/3298 - and the historical vector should be added to all realizations, ref https://github.com/equinor/ert/issues/3297. The historical vectors are usually equal, but could be different in rare cases. Removing duplicates should (?) give better performance in the plotting.

oysteoh:
The functionality around plotting of historical vectors either from refcase or summary-data for each realization is not changed by the removal of the refcase in the plotter, and I don't see any reason to change this now as the performance have not been reported to be an issue before.

oysteoh:
> The functionality around plotting of historical vectors either from refcase or summary-data for each realization is not changed by the removal of the refcase in the plotter, and I don't see any reason to change this now as the performance have not been reported to be an issue before.

wrong assumption, the plotter just picks the historical vector from the first case if any - or use refcase. So - this needs to be changed into plotting all historical vectors from the summary data IF they are not the same.

PR #3305, opened 2022-04-26
Remove refcase in plotter

oysteoh:
**Approach**
Seems like retrieval of historical vectors either by refcase or from the realization itself is already supported in the code. In this PR i have removed everything related to the refcase in the plotter. In addition i have moved the retrieval from using libres_facade directly to utilize our already made endpoint for getting data from keys. What is missing as i see it is to generate historical vectors for all cases if they are NOT equal, but there is another issue for that - #3299 . 

## Pre review checklist


DanSava (commented):
Looks good :+1: , one question should `refcase_data` from `ert_shared/libres_facade.py` be also removed. Seems it is linked to  `history_data` but it might not really need to be. Also used in two other tests:
`test_case_refcase_data`  & test_case_refcase_data_missing_key both in `tests/ert_tests/shared/test_libres_facade.py`

oysteoh:
> Looks good +1 , one question should `refcase_data` from `ert_shared/libres_facade.py` be also removed. Seems it is linked to `history_data` but it might not really need to be. Also used in two other tests: `test_case_refcase_data` & test_case_refcase_data_missing_key both in `tests/ert_tests/shared/test_libres_facade.py`

I will have a look, might be :+1:

oysteoh:
> Looks good +1 , one question should `refcase_data` from `ert_shared/libres_facade.py` be also removed. Seems it is linked to `history_data` but it might not really need to be. Also used in two other tests: `test_case_refcase_data` & test_case_refcase_data_missing_key both in `tests/ert_tests/shared/test_libres_facade.py`

`has_refcase` i can remove since it is not used, but `refcase_data` i'll leave as-is. The point here is that users using the refcase should have the history vector from the refcase plotted. If a refcase or a history-vector from the refcase does not exist - we should get the history-vector from the summary-data.

Did a minor refactoring changing the variable-name `key` -> `history_key` when assigned to make it more clear - i did not like the reuse of the argument to `history_data` function.

ert_gui/tools/plot/plot_api.py
```diff
@@ -209,12 +205,13 @@ def _add_index_range(self, data):
...
-        return self._facade.refcase_data(key)
-
     def history_data(self, key, case=None):
         """Returns a pandas DataFrame with the data points for the history for a given data key, if any.
         The row index is the index/date and the column index is the key."""
-        return self._facade.history_data(key, case)
+
+        if ":" in key:
+            head, tail = key.split(":", 2)
+            history_key = "{}H:{}".format(head, tail)
```
berland, line 214: Pylint will complain here if you don't use `f-strings`.

oysteoh: might be - just moved old code from one function to another here - haven't made this myself. I will have a look at it :)

frode-aarstad:
Did a small comparison between stable and this branch. 
image

frode-aarstad (approved):
LGTM

PR #3311, opened 2022-04-28
Display error messages in ert GUI if dark storage request fails

### LINKED ISSUE #3255: User feedback for dark storage request failures ERT QT gui

DanSava:
**Issue**
Resolves #3255 

## Pre review checklist

### DIFF
--- a/ert_gui/tools/plot/plot_api.py
+++ b/ert_gui/tools/plot/plot_api.py
@@ -7,6 +7,7 @@ from typing import List
 from ert_shared.services import Storage
 from pandas.errors import ParserError
 from itertools import combinations as combi
+from json.decoder import JSONDecodeError
 
 logger = logging.getLogger(__name__)
 
@@ -30,12 +31,14 @@ class PlotApi(object):
         with Storage.session() as client:
             try:
                 response = client.get("/experiments", timeout=self._timeout)
+                self._check_response(response)
                 experiments = response.json()
                 experiment = experiments[0]
                 for ensemble_id in experiment["ensemble_ids"]:
                     response = client.get(
                         f"/ensembles/{ensemble_id}", timeout=self._timeout
                     )
+                    self._check_response(response)
                     response_json = response.json()
                     case_name = response_json["userdata"]["name"]
                     self._all_cases.append(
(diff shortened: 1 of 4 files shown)


ert_gui/tools/plot/plot_api.py
```diff
@@ -174,27 +161,22 @@ def observations_for_key(self, case_name, key):
         case = self._get_case(case_name)
 
         with Storage.session() as client:
+            response = client.get(f"/ensembles/{case['id']}/records/{key}/observations")
+            self._check_response(response)
+
+            obs = response.json()[0]
```
frode-aarstad, line 167: Could this throw an exception? Perhaps put in try

ert_gui/tools/plot/plot_window.py
```diff
@@ -91,13 +120,25 @@ def currentPlotChanged(self):
...
                 self._updateCustomizer(plot_widget)
                 cases = self._case_selection_widget.getPlotCaseNames()
-                case_to_data_map = {
-                    case: self._api.data_for_key(case, key) for case in cases
-                }
+                case_to_data_map = {}
+                for case in cases:
+                    try:
+                        case_to_data_map[case] = self._api.data_for_key(case, key)
+                    except RequestError as e:
```
frode-aarstad, line 127: Perhaps add a TimeoutError here aswell. I guess all api calls could result in a timeout

ert_gui/tools/plot/plot_window.py
```diff
@@ -91,13 +120,25 @@ def currentPlotChanged(self):
...
+                        )
+
+                observations = None
                 if key_def["observations"] and cases:
-                    observations = self._api.observations_for_key(cases[0], key)
-                else:
-                    observations = None
+                    try:
+                        observations = self._api.observations_for_key(cases[0], key)
+                    except RequestError as e:
```
frode-aarstad, line 137: As above

oysteoh:
Running this on a RHEL 7 machine - python38 - hitting a request timeout - this is the result. I guess there should be some kind of message in the box?

error_msg

ert_gui/tools/plot/plot_api.py
DanSava: Yes, it could, but if there is an exception thrown it means the observation response schema was changed and that error would be caught by the API compatibility tests

DanSava: Thought a bit more and realized that I should not blindly trust in our very good testing suite and if we ever get in the situation where a schema change snicks through without notice might not be nice to crash the application. Will add a try catch.

DanSava:
Yes, there should be a message in the box :astonished: not sure why there is none

oysteoh:
> Yes, there should be a message in the box astonished not sure why there is none. ~Can you test it again I added some more instances of catching timeout errors as @frode-aarstad suggested~ Different timeout error is caught by my changes so it should not affect the behaviour.

Yes, still the same.. Doesnt seem to be able to draw the content of the message before there is an abort which tears down the whole GUI. I managed to get another error-msg which showed up correctly. Looks like this..
failed_to_load_keys

Is the `Show this message again` something we can remove? Cannot see the reason for having it :man_shrugging:

frode-aarstad:
Perhaps the QErrorMessage() needs a parent, ie QErrorMessage(parent)

oysteoh:
> This error appears when there is storage_server.json lingering in the run folder after a previous failed run.

yes - the error was not related to your changes - sorry for not making that clear! Just wanted to show that i managed to get an error msg which actually had information and looked ok instead of the black one :+1:

DanSava:
Seems `QErrorMessage` really likes the checkbox `Show this message again` so I replaced it with a critical `QMessageBox`.

oysteoh:
Now it is working as expected i would say. Maybe we should add something to the error msgs telling our users how to recover from the failure ( timeout - maybe wait a bit more? http errors like 400/500 - something wrong with the data - please report back to us? etc.. )  Just one generic message and an Ok button could end up being annoying and our users have no clue what they should do .

DanSava:
> Just one generic message and an Ok button could end up being annoying and our users have no clue what they should do .

It should not be just one generic message for every error, the message the user will see is coming from the underlying error message so it should not be very generic.

oysteoh:
> > Just one generic message and an Ok button could end up being annoying and our users have no clue what they should do .
> 
> It should not be just one generic message for every error, the message the user will see is coming from the underlying error message so it should not be very generic.

Agree to that, but what i meant is that we maybe could filter a bit more on what have happened and propose a solution / what to do. I.e, "Request failed - timeout error" and an OK button does not tell the user much. And i guess even worse when you get a 500 error and a - for the users - cryptic error message. What are they supposed to do?

DanSava:
> And I guess even worse when you get a 500 error and a - for the users - cryptic error message. What are they supposed to do?

I guess for this case a user could take a screenshot and send it to ert-user slack channel.

But I think I get your meaning and it is possible to customize the error messages and possibly suggest some actions to the user but it will be hard not to have general messages for some of the errors, i.e when getting a 500 in which case if we want to just notify the user with a server error message and as a suggestion and what could be the suggested action for this instance?

ert_gui/tools/plot/plot_api.py
```diff
@@ -46,38 +49,40 @@ def _get_all_cases(self) -> List[dict]:
...
-            except (httpx.RequestError, IndexError) as exc:
+            except IndexError as exc:
                 logging.exception(exc)
                 raise exc
 
+    @staticmethod
+    def _check_response(response: requests.Response):
+        if response.status_code != httpx.codes.OK:
+            raise httpx.RequestError(
+                f" Please repost this error and try restarting the application."
```
oysteoh, line 60: Meaning `report` i guess? :thinking:

oysteoh (changes requested):
This looks good to me now! Just one small comment - i guess there is a spellingmistake?

DanSava:
> This looks good to me now! Just one small comment - i guess there is a spellingmistake?

Yup just a butter finger error

oysteoh (approved):

PR #3320, opened 2022-04-29
Merge main into Api adoption

mortalisk:


oysteoh:
This was a quite big PR! :sweat_smile: I see most of it at least does not relate to files where we have done anything so my assumptions is that most of it is ok. Rest of it seems ok from my point of view, cannot see anything special that is not correct. But we should also have some more eyes on it before merging. @DanSava ?

ert_gui/gert_main.py
```diff
@@ -62,20 +64,18 @@ def run_gui(args):
...
     args.config = os.path.basename(args.config)
     ert = EnKFMain(res_config, strict=True, verbose=args.verbose)
-    with Storage.connect_or_start_server(
-        res_config=os.path.basename(args.config)
-    ) as storage:
+    with Storage.connect_or_start_server(res_config=os.path.basename(args.config)):
         # window reference must be kept until app.exec returns
         notifier = ErtNotifier(args.config)
-        window = _start_window(ert, notifier, args)
+        _start_window(ert, notifier, args)
```
DanSava, line 70: not sure if this is OK given the comment `# window reference must be kept until app.exec returns` I am sure it is fine but if it is not then we will get some very strange errors that are hard to debug, so I would keep 
```
window = _start_window(ert, notifier, args) # noqa
```

berland: This is a mixup with the flake8 commit, where this change has been reverted. Should go away when rebased on main.

berland: https://github.com/equinor/ert/commit/b06303d5f2011495da1e70f7da972cc4ad0e1998

DanSava (commented):
Had one small comment but otherwise, it looks good!

ert_gui/gert_main.py
berland: Not needed at all, remove if you wish.

DanSava: ok, I thought it was some obscure new flake8 thing :smile:

DanSava (approved):
:+1:  looks good!

PR #3324, opened 2022-04-29
Limit calls to job_queue_node_get_status()

BjarneHerland:
Each access to self.status in `JobQueueNode` results in a call from Python down to native `job_queue_node_get_status()`.

Apart from the fact that this is suboptimal in terms of performance, it also opens theoretical race-conditions in `_job_monitor()`. This patch carefully caches status, updating the cached status where it is "safe" and may very well improve the situation we see in #3113, although this is hard to prove.

An example of possible race-condition is between lines 172 and 175 here

https://github.com/equinor/ert/blob/287753054871c5f8697a607b25a29d582a16cfdc/res/job_queue/job_queue_node.py#L172-L176

Note that `self.is_running()` in line 172 may do up to 4 calls to C++ and line 175 will do 1 call. When executing line 175 it is assumed that status is in `[PENDING, SUBMITTED, RUNNING, UNKNOWN]` but this might have changed independently in C++ between Python executing lines 172 and 175. (This particular race do not have the most critical consequences but is simple to illustrate.) The state-machines will (in theory) limit these problems but in this code it is easy to prevent by just caching (remembering) the status.

## Pre review checklist

### DIFF
--- a/res/job_queue/job_queue_node.py
+++ b/res/job_queue/job_queue_node.py
@@ -138,12 +138,13 @@ class JobQueueNode(BaseCClass):
     def run_exit_callback(self):
         return self.exit_callback_function(self.callback_arguments)
 
-    def is_running(self):
-        return (
-            self.status == JobStatusType.JOB_QUEUE_PENDING
-            or self.status == JobStatusType.JOB_QUEUE_SUBMITTED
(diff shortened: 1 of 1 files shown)


jondequinor (approved):
LGTM! Good catch.

PR #3336, opened 2022-05-03
Fix historical vectors

### LINKED ISSUE #3299: Only plot one historical vectors across all realizations if they are equal

frode-aarstad:
**Issue**
Resolves #3299 

Fix the displaying of historical data in the QT plot. 
Merge columns with equal history data

### DIFF
--- a/ert_gui/tools/plot/plot_api.py
+++ b/ert_gui/tools/plot/plot_api.py
@@ -1,11 +1,12 @@
-import pandas as pd
 import logging
 import requests
 import io
 import httpx
+import pandas as pd
 from typing import List
 from ert_shared.services import Storage
 from pandas.errors import ParserError
+from itertools import combinations as combi
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


ert_gui/tools/plot/plot_api.py
```diff
@@ -208,20 +209,31 @@ def _add_index_range(self, data):
...
-        index is the key."""
-        return self._facade.refcase_data(key)
-
-    def history_data(self, key, case=None):
+    def history_data(self, key, case=None) -> pd.DataFrame:
         """Returns a pandas DataFrame with the data points for the history for a
         given data key, if any.  The row index is the index/date and the column
         index is the key."""
 
+        if str(key).endswith("H") or "H:" in str(key):
```
oysteoh, line 217: Should there be an explanation on this  piece of code? From my knowledge i think i understand the reason you added it - that being the cases where we actually have keys ending with `H`  - it is not an historical vector? The "loose" concept here is quite fragile, and i support adding this check...

frode-aarstad: Yes, if you select an historical vector this should be picked up here. I can add a comment for this line

ert_gui/tools/plot/plot_api.py
```diff
@@ -224,4 +219,18 @@ def history_data(self, key, case=None):
...
+
+        df = self.data_for_key(case, history_key)
+
+        if not df.empty:
+            df = df.T
+            # Drop columns with equal data
+            duplicate_cols = [
+                cc[1] for cc in combi(df.columns, r=2) if (df[cc[0]] == df[cc[1]]).all()
+            ]
+            duplicate_cols = [
```
oysteoh, line 231: isn't this a reassignment of the 228 duplicate_cols..? I struggle a bit to understand how this should work :thinking:

frode-aarstad: fixed

oysteoh (approved):

PR #3339, opened 2022-05-03
Simplify return type to reflect function behaviour

### LINKED ISSUE #3323: Sensible return-value from job_queue_node_update_status_simple()
The function `job_queue_node_update_status_simple()` seems to always return `false` and might be changed to a void-function.  (Probably left-over from cut'n'paste from `job_queue_node_update_status()`)

valentin-krasontovitsch:
Closes #3323

**Issue**
Resolves #3323

**Approach**
Changed return type from `bool` to `void` as suggested, changed return statements accordingly, removed unused return variable.

## Pre review checklist

- [ ] Added appropriate release note label **not sure what to choose here**

### DIFF
--- a/libres/lib/job_queue/job_node.cpp
+++ b/libres/lib/job_queue/job_node.cpp
@@ -557,20 +557,17 @@ cleanup:
     return status_change;
 }
 
-bool job_queue_node_update_status_simple(job_queue_node_type *node,
-                                         queue_driver_type *driver) {
-    bool status_change = false;
+job_status_type job_queue_node_refresh_status(job_queue_node_type *node,
+                                              queue_driver_type *driver) {
     pthread_mutex_lock(&node->data_mutex);
-    job_status_type current_status;
+    job_status_type current_status = job_queue_node_get_status(node);
     bool confirmed;
 
     if (!node->job_data) {
         job_queue_node_update_timestamp(node);
         pthread_mutex_unlock(&node->data_mutex);
-        return status_change;
+        return current_status;
     }
-
-    current_status = job_queue_node_get_status(node);
 
     confirmed = job_queue_node_status_update_confirmed_running__(node);
 
@@ -592,9 +589,10 @@ bool job_queue_node_update_status_simple(job_queue_node_type *node,
         job_status_type new_status =
             queue_driver_get_status(driver, node->job_data);
         job_queue_node_set_status(node, new_status);
+        current_status = job_queue_node_get_status(node);
     }
     pthread_mutex_unlock(&node->data_mutex);
-    return status_change;
+    return current_status;
 }
 
 bool job_queue_node_status_transition(job_queue_node_type *node,
--- a/res/job_queue/job_queue_node.py
+++ b/res/job_queue/job_queue_node.py
@@ -38,8 +38,9 @@ class JobQueueNode(BaseCClass):
     _get_status = ResPrototype(
         "job_status_type_enum job_queue_node_get_status(job_queue_node)"
     )
-    _update_status = ResPrototype(
-        "bool job_queue_node_update_status_simple(job_queue_node, driver)"
+
+    _refresh_status = ResPrototype(
+        "job_status_type_enum job_queue_node_refresh_status(job_queue_node, driver)"
     )
     _set_status = ResPrototype(
         "void job_queue_node_set_status(job_queue_node, job_status_type_enum)"
@@ -114,6 +115,9 @@ class JobQueueNode(BaseCClass):
     def submit_attempt(self):
         return self._get_submit_attempt()
 
+    def refresh_status(self, driver):
+        return self._refresh_status(driver)
+
     @property
     def status(self):
         return self._get_status()
@@ -168,8 +172,7 @@ class JobQueueNode(BaseCClass):
         if submit_status is not JobSubmitStatusType.SUBMIT_OK:
             self._set_status(JobStatusType.JOB_QUEUE_DONE)
 
-        self.update_status(driver)
-        current_status = self.status
+        current_status = self.refresh_status(driver)
 
         while self.is_running(current_status):
             if (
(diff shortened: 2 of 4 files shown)


sondreso:
Jenkins add to whitelist

libres/lib/job_queue/job_node.cpp
```diff
@@ -594,7 +593,7 @@ bool job_queue_node_update_status_simple(job_queue_node_type *node,
         job_queue_node_set_status(node, new_status);
     }
     pthread_mutex_unlock(&node->data_mutex);
-    return status_change;
+    return;
```
DanSava, line 596: Is this `return` statement still required?

valentin-krasontovitsch: ah thanks, very good point, no, it is not : )

valentin-krasontovitsch: just updated the branch (force-pushed the change of removing that line)

valentin-krasontovitsch:
going to rework this PR according to [bjarne's comment about changing functionality of function in question](https://github.com/equinor/ert/issues/3323#issuecomment-1117296365)

res/job_queue/job_queue_node.py
```diff
@@ -178,7 +174,6 @@ def _job_monitor(self, driver, pool_sema, max_submit):
             ):
                 self._start_time = time.time()
             time.sleep(1)
-            self.update_status(driver)
             if self._should_be_killed():
                 self._kill(driver)
```
valentin-krasontovitsch, line 178: it seems to me that the call to `update_status` in 181 does not need to replaced by anything, as `_should_be_killed` inspect `thread_status` (and `runtime`) and does not depend on `status`. I might be wrong though...

BjarneHerland: I would suggest to use the refactored call here

https://github.com/equinor/ert/blob/175a265d68020b650eb0d46fd64c6929a10f8217/res/job_queue/job_queue_node.py#L119

and drop the internal method `_get_status()` together with its binding to C++.

valentin-krasontovitsch:
of course, the tests... : D
will do, thank you!

valentin-krasontovitsch:
OK so looking at the failing tests and the code, there's a wee bit of a problem.
Right now, I have replaced the `status` method (property) of the `JobQueueNode` class with a call to the new refresh method - but that method needs a driver to be passed in!

there aren't many contexts where that function is called, and it seems like a driver is present in those. e.g. in the `JobQueueNode`'s `is_running` method, we can just change the signature and pass a driver. also in the `JobQueueManager` class, in its methods `isJob[someState]` - the queue seems to have a driver which i assume can be used here.

yet it feels a bit off, to call a method that is supposed to be a property, and pass a driver. but that's perhaps just my feeling, and doesn't have to say much : )

what do you think, @BjarneHerland?

valentin-krasontovitsch:
ok sorry it's a force push, since i've rebased, but just on main - the commits that comprise this work or unaltered

res/job_queue/job_queue_node.py
```diff
@@ -38,8 +38,9 @@ class JobQueueNode(BaseCClass):
     _get_status = ResPrototype(
```
BjarneHerland, line 38: 1) this can be removed

res/job_queue/job_queue_node.py
```diff
@@ -114,6 +115,9 @@ def timed_out(self):
     def submit_attempt(self):
         return self._get_submit_attempt()
 
+    def refresh_status(self, driver):
```
BjarneHerland, line 118: 2) this is not necessary

BjarneHerland, line 118: 2) this is not necessary

res/job_queue/job_queue_node.py
```diff
@@ -114,6 +115,9 @@ def timed_out(self):
...
+        return self._refresh_status(driver)
+
     @property
     def status(self):
         return self._get_status()
```
BjarneHerland, line 123: 3) tis can use self._refresh_status

BjarneHerland, line 123: 3) tis can use self._refresh_status

res/job_queue/job_queue_node.py
```diff
@@ -168,8 +172,7 @@ def _job_monitor(self, driver, pool_sema, max_submit):
         if submit_status is not JobSubmitStatusType.SUBMIT_OK:
             self._set_status(JobStatusType.JOB_QUEUE_DONE)
 
-        self.update_status(driver)
-        current_status = self.status
```
BjarneHerland, line 172: 5) leave this as `current_status = self.status` assuming (3) above

BjarneHerland, line 172: 5) leave this as `current_status = self.status` assuming (3) above

res/job_queue/job_queue_node.py
```diff
@@ -197,7 +199,7 @@ def _job_monitor(self, driver, pool_sema, max_submit):
                     with self._mutex:
                         self._timed_out = True
 
-            current_status = self.status
+            current_status = self.refresh_status(driver)
```
BjarneHerland, line 202: 6) like (5)

BjarneHerland, line 202: 6) like (5)

BjarneHerland (changes requested):
See comments

valentin-krasontovitsch:
ah, but Bjarne, what you're suggesting is exactly what I tried before, and it didn't quite work out, see my discussion of the problem above... but if you have some suggestions to solve the problems I was facing, I'm all ears

BjarneHerland:
> ah, but Bjarne, what you're suggesting is exactly what I tried before, and it didn't quite work out, see my discussion of the problem above... but if you have some suggestions to solve the problems I was facing, I'm all ears

You are right - sorry. I didn't notice that `driver` never becomes a property of the object. Figuring out why this is (and potentially fix it) is beyond the scope of this issue, and may not be worth it at all in fact. Your patch is fine! 😄

BjarneHerland (approved):
Nice!

valentin-krasontovitsch:
great, thank you!

PR #3352, opened 2022-05-04
Remove deprecated UMASK from test data

berland:
There are some further remnants of UMASK in test data
that will have to be removed later.

Cherry-picked from #2921


sondreso (dismissed):
LGTM

PR #3354, opened 2022-05-05
Use ISO date format in tests

berland:
Removes a couple of thousands outputted lines of DeprecationWarning regarding ISO-format when running tests.

Leaving one line of deprecated date format in test code, to be removed whenever the legacy date format support is purged.


verveerpj (approved):
LGTM!

PR #3358, opened 2022-05-06
Set timeout=120sec on our http requests

oysteoh:
Testing with a large case the requests are quite fast and usually within maximum a couple of seconds. But - sometimes requests take a long time, especially switching cases. Specifying a 120 sec timeout seems to hold for all of what i've tested (and ran on LSF).


frode-aarstad (approved):
Looks good. Guess we can monitor this and change if this causes any problems later

PR #3375, opened 2022-05-12
Compare fail

### LINKED ISSUE #3280: Add baseline for comparison for performance tests
When performance tests in tests/performace are run, pytest-benchmark can generate reports. We should store a copy of this for the current state, so that we can fail if the performance drops significantly when running the complete set of jobs in Jenkins. This should be stored in the repo, so that we have a defined baseline to compare against, and must take action if changing it. If possible the test might fail also if performance improves significantly, with a helpful message that we should update the baseline, so that we do not forget to do that.

We could use these files in github actions as well, but that does not run the complete data-set, so it is not so essential.

This work has been started for the Jenkins job, only with files stored on shared disks in stavanger.

mortalisk:
**Issue**
Resolves #3280 

See https://ci.equinor.com/komodo/job/ert-performance/ for jenkins job


oysteoh (approved):
LGTM

PR #3382, opened 2022-05-12
Shorten list of forward models in main GUI

berland:
Repeated forward models are written like

  MAKE_DIRECTORY x32

instead of listing the string MAKE_DIRECTORY 32 times, and flooding the GUI real estate.
Order is kept.

**Issue**
Resolve a flooded GUI display element.

**Approach**
Run-length encoding of string lists

## Pre review checklist


ert_gui/ertwidgets/summarypanel.py
```diff
@@ -103,3 +108,18 @@ def addColumn(self, text):
...
         self.layout.addLayout(layout)
+
+
+def _runlength_encode_list(strings: List[str]) -> List[Tuple[str, int]]:
+    """Runlength encode a list of strings.
+
+    Returns a list of tuples, first element is the string, and the second
+    element is the count of consecutive occurences of the string at the current
+    position."""
+    string_counts: List[Tuple[str, int]] = []
```
xjules, line 119: Out of curiosity, why not to use `Dict[str, int]`? i.e., `If string in string_counts: string_counts[string]+=1 else string_counts[string]=0` ?

berland:
I have tried getting the multipler value to the left of the forward model name, but not successful due to QT only supporting a subset of CSS: https://doc.qt.io/qt-5/richtext-html-subset.html

In HTML, it would be possible to do:
`<div style="text-indent: 5px"><span style="width: 3em; display: inline-block">32x </span>SNAKE_OIL</div>
`
and it would render nice, but this does not work through QT. A workaround would be to use a html table, but I am unable to find the CSS that creates equal vertical distance between every line.

xjules (approved):
LGTM! :rocket:

PR #3384, opened 2022-05-12
Remove unnecessary class EnKFState, bind used functions with pybind11

### LINKED ISSUE #3371: Remove cwrap-bindings in EnKFState and bind callbacks using Pybind11
The class `res/enkf/EnKFState` seems to be used only as a bridge to (static) functions in C++. I can not find any code using actual instances of this class.

Thus, we should be able to remove the current implementation (and the two factory-methods in `_RealEnKFMain`) because the implementation is mostly to handle lifecycle of instances. The callbacks can then be bound with Pybind11, which allows exceptions to flow from C++ to Python for handling in upper layers.

BjarneHerland:
Resolves #3371.

EnKFState was never instantiated and only served as an entrypoint for two functions in C++, which are now bound with pybind11.

## Pre review checklist
- [x] Commit history is consistent and clean, in line with the [contribution guidelines]


oyvindeide (approved):
LGTM!

PR #3403, opened 2022-05-19
Utilize `fmt` arg when invoke plot_date to avoid UserWarning

oysteoh:
Plot date provides a default `fmt` argument with a value `o`. This
is in conflict with our keyword-args specifying marker and line-style.
There is printed warnings in the terminal the keyword args is taking
precedense over the provided `fmt` because they are conflicting.
By utilizing the `fmt` keyword to provide marker and line-style,
this will solve the issue.


oysteoh:
Jenkins: Test this please!

oysteoh:
Guess the failing check now is due to changes in the api of our feature-branch and could be ignored.

DanSava (approved):
Looks good :+1:

PR #3405, opened 2022-05-20
Faster state transitions

### LINKED ISSUE #3401: Pydantic statemachine (snapshot) might halt aioloop
For models that are sufficiently large and fast, i.e. 200 realisations and 30+ make/copy directory jobs early in the forward model, some batches in the evaluator event batcher may consist of ~12k events. On Rgs nodes, this takes more than 10 seconds to process, which blocks the aioloop for the entire duration. 

Blocking the evaluator aioloop is consistent with what is seen in one asset's logs: cluster web socket connections dying for no apparent reason, the evaluator's health check timing out (with a 10 seconds timeout) and intermittent traffic for clients connected to the server though the cluster is computing as full speed (thus traffic should be non-intermittent).

Pydantic is used by the statemachine to type the different constituents (jobs, realisations, etc) and it could be the source of the sluggishness. Alternatively, the recursive merge that is used to update the SM may also contribute.

After some investigation I found that `recursive_update` is the culprit.

jondequinor:
**Issue**
Resolves #3401

**Approach**
- divide work into smaller batches, meaning blocking the asyncio loop for a shorter amount of time
- use processpool to do the heavy snapshot update

## Pre review checklist

### DIFF
--- a/ert/ensemble_evaluator/builder/_ensemble.py
+++ b/ert/ensemble_evaluator/builder/_ensemble.py
@@ -54,15 +54,12 @@ class _Ensemble:
     def snapshot(self) -> Snapshot:
         return self._snapshot
 
-    def update_snapshot(self, events: List[CloudEvent]) -> PartialSnapshot:
-        snapshot_mutate_event = PartialSnapshot(self._snapshot)
-        for event in events:
-            snapshot_mutate_event.from_cloudevent(event)
-        self._snapshot.merge_event(snapshot_mutate_event)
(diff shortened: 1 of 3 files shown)


jondequinor:
Seems to have capped asyncio loop blockage at ~4 seconds, compared to ~20 seconds before the fix. This may be OK.

BjarneHerland:
Fixed by https://github.com/equinor/ert/pull/3408

PR #3408, opened 2022-05-20
Limit size of single batch in order to reduce max heartbeat-interval

### LINKED ISSUE #3401: Pydantic statemachine (snapshot) might halt aioloop

BjarneHerland:
**Issue**
Resolves #3401

**Approach**
If the EnsembleEvaluator attempts to process large chunks of events in one batch it might block the asyncio-loop long enough to create problems for the other activities in the server. This patch splits up the events in chunks which get processed separately.

## Pre review checklist

- [x] Commit history is consistent and clean, in line with the [contribution guidelines]


jondequinor (approved):
LGTM! 👏

PR #3419, opened 2022-05-23
Separate benchmarks into different runs

mortalisk:
Also adds code to share the generated data from the template_config

After merging, remember to update jenkins-job to point to main again.


oysteoh (approved):
Looks reasonable and in line with what we have discussed :+1:

PR #3449, opened 2022-05-30
Use pd.concat instead of append

### LINKED ISSUE #3448: FutureWarning: The frame.append method is deprecated
Plotting POLY_RES@0 of the poly example gives this:

loader.py:244: FutureWarning: The frame.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead.
  pd.DataFrame(

dafeda:
**Issue**
Resolves #3448

**Approach**

## Pre review checklist

### DIFF
--- a/ert_data/loader.py
+++ b/ert_data/loader.py
@@ -240,11 +240,11 @@ def _load_general_obs(facade, observation_keys, case_name):
             index_list = [node.getIndex(nr) for nr in range(len(node))]
             index = _create_multi_index(index_list, index_list)
 
-            data.append(
-                pd.DataFrame(
-                    [node.get_data_points()], columns=index, index=["OBS"]
-                ).append(pd.DataFrame([node.get_std()], columns=index, index=["STD"]))
+            df_obs = pd.DataFrame(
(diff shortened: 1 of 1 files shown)


ert_data/loader.py
```diff
@@ -241,10 +241,16 @@ def _load_general_obs(facade, observation_keys, case_name):
...
-                    [node.get_data_points()], columns=index, index=["OBS"]
-                ).append(pd.DataFrame([node.get_std()], columns=index, index=["STD"]))
+                pd.concat(
+                    [
+                        pd.DataFrame(
+                            [node.get_data_points()], columns=index, index=["OBS"]
+                        ),
+                        pd.DataFrame([node.get_std()], columns=index, index=["STD"]),
+                    ]
+                )
```
berland, line 251: Since there already is a `pd.concat` on line 254, would it work to just append each of these frames to `data`?

dafeda: Good point, and yes I think we can.
I've also removed the `observations` list as it seems redundant.

dafeda: Tried making bigger changes, but tests are failing and I don't think it's worth it spending too much time on this.

PR #3450, opened 2022-05-30
Release GIL before waiting for (async) futures in C++

BjarneHerland:
C++ futures may be executed in separate threads. If the code which waits for such futures holds the GIL, e.g. it was invoked from Python via Pybind11, this prevents code running in the futures from aquiring the GIL, causing deadlocks if code in future tries to use our logging-mechanism.

### DIFF
--- a/libres/lib/enkf/enkf_main.cpp
+++ b/libres/lib/enkf/enkf_main.cpp
@@ -519,6 +519,13 @@ int enkf_main_load_from_run_context(enkf_main_type *enkf_main,
     Semafoor concurrently_executing_threads(100);
     std::vector<std::tuple<int, std::future<int>>> futures;
 
+    // If this function is called via pybind11 we need to release
+    // the GIL here because this function may spin up several
+    // threads which also may need the GIL (e.g. for logging)
+    PyThreadState *state = nullptr;
+    if (PyGILState_Check() == 1)
+        state = PyEval_SaveThread();
+
     for (int iens = 0; iens < ens_size; ++iens) {
         if (bool_vector_iget(iactive, iens)) {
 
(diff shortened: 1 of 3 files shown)


BjarneHerland:
Jenkins: test ert please

BjarneHerland:
Jenkins: test libres please

BjarneHerland:
Jenkins: test ert please

libres/lib/enkf/enkf_main.cpp
```diff
@@ -519,6 +519,13 @@ int enkf_main_load_from_run_context(enkf_main_type *enkf_main,
     Semafoor concurrently_executing_threads(100);
     std::vector<std::tuple<int, std::future<int>>> futures;
 
+    // If this function is called via pybind11 we need to release
+    // the GIL here because this function may spin up several
+    // threads which also may need the GIL (e.g. for logging)
+    PyThreadState *state = nullptr;
```
pinkwah, line 525: Using the pybind11 `py::scoped_gil_release` here should be safer than using the C API due to RAII (`logger->warning` may throw an exception via Python, for example).

BjarneHerland: My first thought also, but it blows up due to some tests calling these functions directly from C++ (i.e. not holding the GIL, then trying to release it). Moreover, the `scoped_gil_release` is (as name indicates) indeed scoped, so it cannot be used conditionally with an if-check like the current patch uses.

BjarneHerland: @pinkwah further comments?

pinkwah (approved):

PR #3461, opened 2022-05-31
Set file dialog to reasonable width and height - simplified approach

### LINKED ISSUE #997: [GUI] Sizing of stderr/stdout file-viewer
Suggest to auto-size, at least horizontally the file viewer for stdout/stderr files. If not auto-size, set default to 80 columns.

image

Auto-sizing can also be relevant vertically.

valentin-krasontovitsch:
We set the file dialog for viewing stdout and stderr of simulations to a
standard width of 80, and a height of 1/3 of the screen.

**Issue**
Closes #997

**Approach**
See commit message above

## Pre review checklist

### DIFF
--- a/ert_gui/tools/file/file_dialog.py
+++ b/ert_gui/tools/file/file_dialog.py
@@ -1,4 +1,5 @@
-from qtpy.QtCore import QThread, Slot, Qt
+from math import floor
+from qtpy.QtCore import QThread, Slot, Qt, QSize
 from qtpy.QtWidgets import (
     QDialog,
     QMessageBox,
@@ -59,10 +60,23 @@ class FileDialog(QDialog):
         self._thread.quit()
         self._thread.wait()
 
-    def _init_layout(self):
-        self.setMinimumWidth(600)
-        self.setMinimumHeight(400)
+    def _calculate_font_based_width(self):
+        font_metrics = self._view.fontMetrics()
+        desired_width_in_characters = 120
+        extra_bit_of_margin_space = 2
+        extra_space_for_vertical_scroll_bar = 5
+        return (
+            desired_width_in_characters
+            + extra_bit_of_margin_space
+            + extra_space_for_vertical_scroll_bar
+        ) * font_metrics.averageCharWidth()
 
+    def _calculate_screen_size_based_height(self):
+        screen_height = QApplication.primaryScreen().geometry().height()
+        max_ratio_of_screen = 1.0 / 3.0
+        return floor(screen_height * max_ratio_of_screen)
+
+    def _init_layout(self):
         dialog_buttons = QDialogButtonBox(QDialogButtonBox.Ok)
         dialog_buttons.accepted.connect(self.accept)
 
@@ -129,3 +143,10 @@ class FileDialog(QDialog):
         if self._follow_mode:
             self._view.moveCursor(QTextCursor.End)
         self._view.appendPlainText(text)
+        self.adjustSize()
+
+    def sizeHint(self) -> QSize:
+        return QSize(
+            self._calculate_font_based_width(),
+            self._calculate_screen_size_based_height(),
+        )
--- a/tests/ert_tests/gui/tools/file/stdout-long-and-extra-wide
+++ b/tests/ert_tests/gui/tools/file/stdout-long-and-extra-wide
@@ -1,0 +1,146 @@
+BLOOD & THUNDER PROPHET: ...And the bezan shall be huge and black, and the eyes thereof red with the blood of living creatures, and the whore of Babylon shall ride forth on a three-headed serpent, and throughout the lands, there'll be a great rubbing of parts. Yeeah...
+
+FALSE PROPHET: ...For the demon shall bear a nine-bladed sword. Nine-bladed!  Not two or five or seven, but nine, which he will wield on all wretched sinners, sinners just like you, sir, there, and the horns shall be on the head, with which he will...
+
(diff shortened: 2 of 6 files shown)


ert_gui/tools/file/file_dialog.py
```diff
@@ -129,3 +144,7 @@ def _append_text(self, text: str) -> None:
         if self._follow_mode:
             self._view.moveCursor(QTextCursor.End)
         self._view.appendPlainText(text)
+        self.adjustSize()
+
+    def sizeHint(self) -> QSize:
```
berland, line 149: does this have to be camelCased?

valentin-krasontovitsch: it looks like that's the convention of the python qt stuff

ert_gui/tools/file/file_dialog.py
```diff
@@ -1,4 +1,5 @@
-from qtpy.QtCore import QThread, Slot, Qt
+from qtpy.QtCore import QThread, Slot, Qt, QSize
+from math import floor
```
berland, line 2: For style reasons, the `math` import should be outside (ahead) of the `qtpy` imports. I fancy `isort`, which would fix this (but it might fail in case of circular imports so we cannot auto-apply it to all files)

ert_gui/tools/file/file_dialog.py
```diff
@@ -59,10 +60,24 @@ def _stop_thread(self):
         self._thread.quit()
         self._thread.wait()
 
-    def _init_layout(self):
-        self.setMinimumWidth(600)
-        self.setMinimumHeight(400)
+    def _calculate_desired_width(self):
```
berland, line 63: In the last line of this function you describe what is "desired" by a temporary variable. I would suggest renaming the function to `_calculated_font_based_width` and then a direct `return` at the end (no temp variable called `font_based_with`)

ert_gui/tools/file/file_dialog.py
hnformentin: I think this is because PEP8 style conventions:

> Imports should be grouped in the following order:
> 
> Standard library imports.
> Related third party imports.
> Local application/library specific imports.
> You should put a blank line between each group of imports.
> 

https://peps.python.org/pep-0008/#imports

tests/ert_tests/gui/tools/file/stdout-long-and-extra-wide
```diff
@@ -0,0 +1,49 @@
...
+this thy hand grenade, that with it thou mayst blow thine enemies to tiny bits,
+in thy mercy.' And the Lord did grin. And the people did feast upon the lambs,
+and sloths, and carp, and anchovies, and orangutans, and breakfast cereals, and
+fruit bats, and large chulapas. And the Lord spake, saying, 'First shalt thou
+take out the Holy Pin. Then shalt thou count to three, no more, no less. Three
+shall be the number thou shalt count, and the number of the counting shall be
+three. Four shalt thou not count, neither count thou two, excepting that thou
+then proceed to three. Five is right out. Once the number three, being the
+third number, be reached, then lobbest thou thy Holy Hand Grenade of Antioch
+towards thy foe, who, being naughty in My sight, shall snuff it.'
```
hnformentin, line 49: Hei, the PR seems large than it actually is for me. The three files `stdout-ust-right`, `stdout-long` and `std-long-and-extra-wide` have the same content, just repeated to increase the size. Do you think it would have a way to avoid this repetition?

hnformentin:
It seems I came late to the party 👯 sorry

ert_gui/tools/file/file_dialog.py
valentin-krasontovitsch: done, similar renaming also applied to corresponding height function.

ert_gui/tools/file/file_dialog.py
valentin-krasontovitsch: thank you! done.

tests/ert_tests/gui/tools/file/stdout-long-and-extra-wide
valentin-krasontovitsch: Since we only care about the length / formatting of the files, and not about the content, and don't need to maintain the content, I thought this would be a pythonic way of solving the need for text files. introducing randomness into tests where it is not needed by the specs invites unnecessary brittleness, in my opinion...

I'll change the content of the files, to make it DRYer. it'll still seem like a long PR, but as the lengthy part is just dummy text files that don't require maintenance, I personally think that's fine. Please let me know if you strongly disagree.

Also, are there any policies / regulations / metrics regarding the number of (changed / new) lines in a PR at SCOUT or equinor that I'm not aware of? Wondering, since I got several comments on the length of this PR.

hnformentin: Good point to discuss in the standup. I don't see clearly why the use of `hypothesis` would introduce problematic randomness, but do not have a strong opinion. In regards to policies, regulations and metrics, I don't know of any besides the common understanding that a PR should not be longer than it needs to be, and it should solve only one issue. But again, good to check what others think about it 🎉

valentin-krasontovitsch: you're right, i don't think it would introduce anything dangerous or brittle in this case! BUT after lengthy discussion in the standup today, i shall merge it as is, as it's already approved.

PR #3466, opened 2022-06-01
Handle dying batcher

### LINKED ISSUE #3197: ensemble evaluator swallows exceptions
**Describe the bug**
Imagine you introduce a bug in some code used by the dispatcher/batcher in the ensemble evaluator

```
def _match_token(token: str, source: str) -> str:
    f_pattern = _regexp_pattern.format(token=token)
    match = re.search(f_pattern, source)
    raise RuntimeError("nope")
    return match if match is None else match.group()  # type: ignore
```

You won't actually see this error, it will be silently swallowed because the `asyncio.Task` that sees that exception, is never awaited (normally it is, but after nominal completion).

**Expected behavior**
Errors must not be swallowed, they must be propagated so that the application fails

BjarneHerland:
**Issue**
Resolves #3197

### DIFF
--- a/ert_shared/ensemble_evaluator/dispatch.py
+++ b/ert_shared/ensemble_evaluator/dispatch.py
@@ -3,28 +3,40 @@ import asyncio
 import time
 
 from collections import defaultdict, OrderedDict
-from typing import Optional
 
 from ert.ensemble_evaluator import identifiers
-
-from ert.ensemble_evaluator.state import (
-    ENSEMBLE_STATE_CANCELLED,
-    ENSEMBLE_STATE_FAILED,
-    ENSEMBLE_STATE_STOPPED,
-)
 
 logger = logging.getLogger(__name__)
 
 
-class Batcher:
-    def __init__(self, timeout, max_batch=1000, loop=None):
+class BatchingDispatcher:
+    def __init__(self, loop, timeout, max_batch=1000):
         self._timeout = timeout
         self._max_batch = max_batch
+
+        self._LOOKUP_MAP = defaultdict(list)
         self._running = True
         self._buffer = []
 
         # Schedule task
         self._task = asyncio.ensure_future(self._job(), loop=loop)
+        self._task.add_done_callback(self._done_callback)
+
+    def _done_callback(self, *args):
+        try:
+            failure = self._task.exception()
+            if failure is not None:
+                logger.warning(f"exception in batcher: {failure}")
+            else:
+                logger.debug("batcher finished normally")
+                return
+        except asyncio.CancelledError as ex:
+            logger.warning(f"batcher was cancelled: {ex}")
+
+        # call any registered handlers for FAILED. since we don't have
+        # an event, pass empty list and let handler decide how to proceed
+        funcs = self._LOOKUP_MAP[identifiers.EVTYPE_ENSEMBLE_FAILED]
+        asyncio.gather(*[f([]) for f, _ in funcs])
 
     async def _work(self):
         t0 = time.time()
@@ -43,9 +55,6 @@ class Batcher:
         for f, events in function_to_events_map.items():
             await f(events)
 
-    def put(self, f, event):
-        self._buffer.append((f, event))
-
     async def _job(self):
         while self._running:
             await asyncio.sleep(self._timeout)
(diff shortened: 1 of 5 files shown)


ert_shared/ensemble_evaluator/dispatch.py
```diff
@@ -3,28 +3,39 @@
...
+    def _done_callback(self, *args):
+        try:
+            failure = self._task.exception()
+            if failure is not None:
+                logger.warning(f"exception in batcher: {failure}")
+            else:
+                logger.debug("batcher finished normally")
+                return
+        except asyncio.CancelledError as ex:
+            failure = ex
```
jondequinor, line 34: Unnecessary. Leftover from when you used `failure` in the outmost scope?

ert_shared/ensemble_evaluator/dispatch.py
```diff
@@ -3,28 +3,39 @@
...
-    ENSEMBLE_STATE_FAILED,
-    ENSEMBLE_STATE_STOPPED,
-)
-
 logger = logging.getLogger(__name__)
 
 
-class Batcher:
-    def __init__(self, timeout, max_batch=1000, loop=None):
+class BatchingDispatcher:
```
jondequinor, line 12: :1st_place_medal:

ert_shared/ensemble_evaluator/dispatch.py
```diff
@@ -3,28 +3,39 @@
...
+                logger.warning(f"exception in batcher: {failure}")
+            else:
+                logger.debug("batcher finished normally")
+                return
+        except asyncio.CancelledError as ex:
+            failure = ex
+            logger.warning(f"batcher was cancelled: {failure}")
+
+        funcs = self._LOOKUP_MAP[identifiers.EVTYPE_ENSEMBLE_FAILED]
+        asyncio.gather(*[f([]) for f, _ in funcs])
```
jondequinor, line 38: Maybe add a comment here, explaining that the handler will have special behavior in case of an empty list of events?

ert_shared/ensemble_evaluator/evaluator.py
```diff
@@ -162,7 +208,15 @@ async def handle_dispatch(self, websocket, path):
...
                     continue
-                await self._dispatcher.handle_event(event)
+                try:
+                    await self._dispatcher.handle_event(event)
+                except BaseException:
+                    logger.warning(
+                        "cannot handle event - closing connection to dispatcher"
+                    )
+                    websocket.close()
+                    return
```
jondequinor, line 218:
```suggestion
                except BaseException as ex:
                    logger.exception(
                        "cannot handle event - closing connection to dispatcher"
                    )
                    websocket.close(code=1011, reason=f"cannot handle event: {ex}")
                    return
```
Maybe omit `ex` from `reason`, but give one.

ert_shared/ensemble_evaluator/evaluator.py
```diff
@@ -195,24 +249,24 @@ async def evaluator_server(self, done):
...
+            logger.debug(
+                f"Got done signal. {self._dispatchers_connected.qsize()} dispatchers to disconnect..."  # noqa: E501
+            )
+            try:  # Wait for dispatchers to disconnect
+                await asyncio.wait_for(self._dispatchers_connected.join(), timeout=20)
             except asyncio.TimeoutError:
                 logger.debug("Timed out waiting for dispatchers to disconnect")
-            await self._batcher.join()
+            logger.debug("Joining batcher...")
+            await self._dispatcher.join()
```
jondequinor, line 261:
```suggestion
            try:
                await asyncio.wait_for(self._dispatcher.join(), timeout=20)
            except asyncio.TimeoutError:
                logger.debug(…)
```
for consistency?

ert_shared/ensemble_evaluator/evaluator.py
```diff
@@ -195,24 +249,24 @@ async def evaluator_server(self, done):
...
 
             terminated_attrs = {}
             terminated_data = None
             if self._result:
                 terminated_attrs["datacontenttype"] = "application/octet-stream"
                 terminated_data = cloudpickle.dumps(self._result)
-            message = self._create_cloud_event(
+            logger.debug("Notifying clients...")
```
jondequinor, line 268:
```suggestion
            logger.debug("Sending termination to clients...")
```

jondequinor (commented):
Some minor comments.

Overall I think this is very good and goes above and beyond what I expected—which was logging. But if the batcher now dies, clients would be notified, which I think is great.

ert_shared/ensemble_evaluator/dispatch.py
BjarneHerland: Not really. `asyncio.Task.exception()` raises `CancellerError` if the task has been cancelled (i.e. `set_exception()` was not called). AFAICS this code covers all possibilities in a done-handler?

ert_shared/ensemble_evaluator/dispatch.py
BjarneHerland: Will do

ert_shared/ensemble_evaluator/evaluator.py
BjarneHerland: Good catch! Will do.

ert_shared/ensemble_evaluator/dispatch.py
jondequinor: I was referring to `failure = ex`

BjarneHerland: Ahh... yes. Thx!

ert_shared/ensemble_evaluator/evaluator.py
BjarneHerland: yup

ert_shared/ensemble_evaluator/evaluator.py
BjarneHerland: fair enough

jondequinor (approved):

PR #3469, opened 2022-06-02
Add libres apidocs to fmudocs

eivindjahren:
Generate libres api documentation using doxygen, convert to rst with breathe and include it in fmu-docs.

The intent behind this change is to enable easier discoverability of documentation in the libres code. 
The general idea is to introduce doxygen markup to the c/cpp comments so that it is possible to cross-reference,
and use the breathe/sphinx pipeline to add that documentation to fmu-docs. 

The current approach does not include the entire apidoc by default, only those explicitly added to `docs/rst/manual/libres_apidoc`. This is to avoid poor signal to noise ratio as large parts of the code
has no documentation. It is possible to add every piece of code to the doc via either `breathe-apidoc`
or the `doxygenindex` directive in the future.

documentation_example

## Pre review checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -238,6 +238,9 @@ jobs:
 
     - name: Test docs
       run: |
+        sudo apt-get install -y doxygen
+        doxygen docs/rst/manual/doxyfile
+        breathe-apidoc -g file docs/rst/manual/doxygen/xml -o docs/rst/manual/libres_apidoc/
         sphinx-build -n -v -E -W ./docs/rst/manual ./tmp/ert_docs
 
   tests-libres:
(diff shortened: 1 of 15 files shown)


dafeda:
I think this looks real neat, but probably a good idea if @pinkwah takes a peek.

pinkwah:
@dafeda lgtm

eivindjahren:
@dafeda @pinkwah Then I'll go back and have a look at this again. The readthedocs builder will have to be updated and fmu-docs will similarly have to update its docs building.

eivindjahren:
The current state of this is that breathe generates rst with a lot of warnings which we turn into errors in order to verify that the docs are built without missing references etc. Its not at all trivial to fix all of these issues.

PR #3482, opened 2022-06-08
Add experimental feature flag to webviz ert

frode-aarstad:
**Issue**
Resolves https://github.com/equinor/webviz-ert/issues/249

### DIFF
--- a/ert_shared/services/_webviz_ert_main.py
+++ b/ert_shared/services/_webviz_ert_main.py
@@ -6,7 +6,8 @@ import logging
 import tempfile
 import yaml
 import pathlib
-from typing import Any, Optional
+import argparse
+from typing import Any, Optional, Dict
 from webviz_ert.assets import WEBVIZ_CONFIG
 
 
@@ -26,7 +27,10 @@ def handle_exit(
 
 
 def create_config(
-    project_identifier: Optional[str], config_file: pathlib.Path, temp_config: Any
+    project_identifier: Optional[str],
+    config_file: pathlib.Path,
+    temp_config: Any,
+    experimental_mode: bool,
 ) -> None:
     with open(config_file, "r") as f:
         config_dict = yaml.safe_load(f)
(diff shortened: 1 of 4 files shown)


ert_shared/services/_webviz_ert_main.py
```diff
@@ -73,4 +96,8 @@ def run_webviz_ert():
 
 
 if __name__ == "__main__":
-    run_webviz_ert()
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--experimental_mode", type=str, default="False")
+    args = parser.parse_args()
+
+    run_webviz_ert(args.experimental_mode.lower() == "true")
```
DanSava, line 103: This is a bit strange to me, it is not equivalent to:

```python
    parser = argparse.ArgumentParser()
    parser.add_argument("--experimental_mode", action="store_true")
    args = parser.parse_args()

    run_webviz_ert(args.experimental_mode)
```
Also, I would also add:

```python
    parser.add_argument("--experimental_mode", "-e",  action="store_true")
```

Sorry if I am missing the obvious reason why we are required to convert the command line argument to a string and then back to a boolean.

ert_shared/services/_webviz_ert_main.py
```diff
@@ -23,15 +24,34 @@ def handle_exit(*args: Any) -> None:  # pylint: disable=unused-argument)
...
         config_dict = yaml.safe_load(f)
         for page in config_dict["pages"]:
             for element in page["content"]:
                 for key in element:
                     element[key] = {"project_identifier": project_identifier}
-    output_str = yaml.dump(config_dict)
+
+    new_config_dict = config_dict
+
+    def filter_experimental(e, b: bool) -> bool:
```
DanSava, line 41: Is it possible to make `e` and `b` a bit more explicit something like `page` and `experimental_mode` 

also if `b` gets a type (`b: bool`) maybe also e should get a type just for consistency (`e: Dict[str, Any]`)  I am also ok with no variable types for these inner functions

ert_shared/services/_webviz_ert_main.py
```diff
@@ -23,15 +24,34 @@ def handle_exit(*args: Any) -> None:  # pylint: disable=unused-argument)
...
+        if "experimental" in e:
+            if e["experimental"]:
+                return b
+            else:
+                return True
+        else:
+            return True
```
DanSava, line 48: This might be equivalent with:
``` python 
if e.get("experiment"):
   return b
return True 
```

DanSava (approved):
Looks good had just some minor pedantic comments

PR #3483, opened 2022-06-08
Reverting an earlier attempt to optimize creation

BjarneHerland:
Should resolve the issue seen in https://ci.equinor.com/komodo/job/x-komodo-test/1544/

**Approach**
Create the asyncio.Queue under the right asyncio-loop.


BjarneHerland:
Jenkins: test ert please

PR #3509, opened 2022-06-13
Let qstat fail a couple of times before giving up

### LINKED ISSUE #405: Torque driver unable to get job status
Running Ert (cli) on the Azure cluster using the torque driver
Right after starting, a message like this is spit out for every job:```** Warning: failed to get job status for job:218 from file:/tmp/enkf-qstat-5956-365084```  
I do have write access to the `/tmp` folder, but the file is not created (regular users might have write access to that folder, but that's another issue).

The jobs actually are started and complete correctly, but Ert hangs, most likely because it's unable to detect that the jobs are completed

berland:
qstat is used to ask the queue server for torque for the
status of individual jobs. With ERTs ping rate at once pr sec
pr realization, we need to accept intermittent failures and retry.

**Issue**
Resolves #405 

**Approach**
Retry up to 3 times with exponential sleep. If qstat/PBS is really down it will give failure state to job_queue after 14 seconds. 

Manual testing towards Azure cluster and PBS with 160 realizations has never needed more than one retry.

## Pre review checklist

### DIFF
--- a/libres/lib/job_queue/torque_driver.cpp
+++ b/libres/lib/job_queue/torque_driver.cpp
@@ -521,8 +521,21 @@ torque_driver_get_qstat_status(torque_driver_type *driver,
         const char **argv = (const char **)util_calloc(1, sizeof *argv);
         argv[0] = jobnr_char;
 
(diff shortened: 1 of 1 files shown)


berland:
Superseded by #3537

PR #3540, opened 2022-06-17
Select pytest asyncio_mode=auto

berland:
Not specifying this option yields the legacy mode which triggers a DeprecationWarning

We have the option of selecting between `strict` and `auto`. The documentation says that we should use 
`strict` "if multiple async frameworks should be combined in the same test suite", which we probably are not doing (?)

https://github.com/pytest-dev/pytest-asyncio#auto-mode

## Pre review checklist


xjules (approved):
LGTM!

PR #3537, opened 2022-06-17
Retry qsub and qstat in case of failures

berland:
**Issue**
Resolves #405 

**Approach**
Retry qstat and qsub commands in case of (intermittent) failures.

~Based on (blocked by) #3518  and #3490~

## Pre review checklist

- [x] Added appropriate release note label
- [x] PR title captures the intent of the changes, and is fitting for release notes.
- [x] Commit history is consistent and clean, in line with the [contribution guidelines](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md).

Adding labels helps the maintainers when writing release notes. This is the [list of release note labels](https://github.com/equinor/ert/labels?q=release-notes).

PR #3553, opened 2022-06-21
Add proxy script for qstat

### LINKED ISSUE #3541: Update Torque driver to handle thousands of realizations
See #405 and #3537

berland:
**Issue**
Resolves #3541 

**Approach**
This shell script is significantly faster to run compared to the
real qstat ("backend") as it only greps in a file on /tmp.
The file on /tmp acts as a cache, and will be updated whenever
it gets too old (~seconds).
    
The queue manager in ERT polls the queing system (that is qstat)
once pr. second for every realization. This can replicate a DDoS on
the cluster queue server, which we mitigate by using this proxy.

Blocked by #3537 

## Pre review checklist


berland:
Note: Fails on macos due to diffent syntax for the `stat` command.

jondequinor (dismissed):
Why did you choose bash? flock (`fcntl`), stat, cat should all be supported by python stdlib.

Unless there's a really good reason to choose bash, we should choose the portable python route.

For example, try building a komodo distribution on macos. It will fail mostly because it assumes `rename`, `rmdir` to come from `coreutils`, which on macos they do not.

berland:
`sh` was chosen as I speculate it is significantly cheaper/faster to spawn compared to Python. As this is command will be called every second for every realization with the current queue manager, startup time can be critical. The file locking used by `flock` is in any case not portable to Mac.

xjules:
Just to recap our discussion: add one more test to check that `proxyfile` gets new content when timeout happens. Otherwise it's a solid piece of work! Nice @berland

berland:
Rebase pending reviewer @xjules

xjules (approved):
Nice @berland ! Very good job! :rocket:

PR #3558, opened 2022-06-22
Exp server from gui

lars-petter-hauge:
Some testing from workshop. Note that this will freeze gui while experiment is running.


lars-petter-hauge:
Closing - just a poc that didn't really provide a good way of implementing this anyway.

PR #3562, opened 2022-06-23
Remove fixing of nodes

eivindjahren:
The fixing of nodes would replace a broken block with a block marked as
invalid. That fixed block would have its size set so that it ends at thenext valid node.

The idea of doing so was that the block could be reused. However, commit
https://github.com/equinor/ert/commit/64aacd49c18513750d41431129b019f8612b0cd9 made it so that invalid nodes are never reused.

The fixing of nodes are always done after the index is built, meaning
"fixing" has the following effect

1) NODE_WRITE_ACTIVE is changed to NODE_INVALID
2) And end tag might be added where it is missing before a valid block.

The next time we build the index, the error messages are different when
fixing is applied, however, if a if fixed block is encountered
(`NODE_INVALID`) then it performas a `block_fs_fseek_valid_node` on and
asked to be fixed.

Therefore, "fixing" has little effect, and so we simply do not perform
it. This means `NODE_INVALID` should never be written to the file, but
we keep it for backwards compatability.

## Pre review checklist

### DIFF
--- a/libres/lib/res_util/block_fs.cpp
+++ b/libres/lib/res_util/block_fs.cpp
@@ -59,6 +59,8 @@ typedef enum {
     /** NODE_IN_USE_BYTE * ( 1 + 256 + 256**2 + 256**3) => Binary 01010101010101010101010101010101 */
     NODE_IN_USE = 1431655765,
     NODE_WRITE_ACTIVE = WRITE_START__,
+    /** NODE_INVALID is no longer written to the file, but its kept for
+     * backwards compatability */
     NODE_INVALID = 13
 } node_status_type;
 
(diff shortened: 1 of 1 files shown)


eivindjahren:
Jenkins: Retest this please

libres/lib/res_util/block_fs.cpp
```diff
@@ -471,16 +429,19 @@ block_fs_type *block_fs_mount(const fs::path &mount_file, int fsync_interval,
     if (!fs::exists(mount_file))
         /* This is a brand new filesystem - create the mount map first. */
         block_fs_fwrite_mount_info(mount_file);
+<<<<<<< HEAD
```
dafeda, line 432: Remnants of a rebase.

PR #3568, opened 2022-06-23
POC for using Protobuf and gRPC

BjarneHerland:
This consists of three standalone Python-executables: `server.py`, `client.py` and `dispatcher.py`. Together they implement and explore building-blocks for replacing the existing EnsembleEvaluator, as well as moving to an "experiment-server".

`server.py` implements the server-functionality with "endpoints" for dispatchers and clients, and a preliminary state-machine on the server-side.

`client.py` implements a client monitoring the state of an experiment, getting periodic updates to the experiment. Monitoring any other level in the object-hierarchy is trivial to do from this client but should be verified.

`dispatcher.py` implements a Job pushing state-updates to the server. The updates are encoded sequentially by the method `play()`

An important part of this PR is the description of the datamodel and service, encoded in `experimentserver.proto`. The datamodel is close to the existing one in the ERT codebase, although not 100%, but should be ok for a POC. The service-definition mirrors the current `EnsembleEvaluator`.

NOTE: If changing `experimentserver.proto` see README.txt for how to re-generate necessary Python-files


xjules:
[#protobuf](https://github.com/equinor/ert/pull/3733) is an ideological continuation of this POC.

PR #3570, opened 2022-06-24
Make dependency on file location in Block explicit

eivindjahren:
A small refactoring to make dependency of the `Block::write_data()` byte offset explicit.

The `Block::write_data()` function depended on being called with the `data_stream` being at the location
it should write, which was unlike any of the other `write_*` functions. The `write_header()` function would
guarantee that this was the case. This ment the order of operations in `write` was very much significant although
it may not look like it.

The `write_*` functions follow a general pattern which is not uncommon. They `fseek()` in the `stream` to where they should write and afterwords `stream` is at the end of where we wrote. Another common pattern is to not enforce the location, and simply write wherever the stream is currently pointing, but the current design makes that infeasible for all of the functions, yet `write_data()` followed that convention. Instead of having this surprising inconsistency, the `write_header()` function now supplies the data offset which is passed to `write_data()`, making the dependency explicit.

All the `write_*` functions now have the following behavior

1. move to some predetermined location in the stream
2. Write your thing there
3. After finishing the stream location is after where it wrote.

Points 2. and 3. is the same behavior as `fwrite`. Point 1. is a reasonable convention.

## Pre review checklist

### DIFF
--- a/libres/lib/res_util/block_fs.cpp
+++ b/libres/lib/res_util/block_fs.cpp
@@ -170,9 +170,10 @@ public:
 
     void write(const char *filename, FILE *data_stream, const void *ptr) {
         write_active_markers(data_stream);
-        write_header(filename, data_stream);
-        if (status == NODE_IN_USE)
-            write_data(data_stream, ptr);
+        long data_offset = write_header(filename, data_stream);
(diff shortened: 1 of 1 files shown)


dafeda (approved):
Nice 👍

PR #3605, opened 2022-07-07
Send fewer but larger dataChanged signals

jondequinor:
**Approach**
For some assets, models are extremely busy in the early
parts of the forward model. Many (up to 30) jobs for probably
all realizations are part of a partial snapshot. dataChanged
should not be signalled for each and every one of these changes,
but rather be signalled as a larger selection.

**Issue**
Resolves no issue, but this probably affects the GUI very negatively for the aforementioned assets.

## Pre review checklist

### DIFF
--- a/src/ert/gui/model/snapshot.py
+++ b/src/ert/gui/model/snapshot.py
@@ -165,6 +165,10 @@ class SnapshotModel(QAbstractItemModel):
             )
             stack.callback(self.dataChanged.emit, iter_index, iter_index_bottom_right)
 
+            # A list of changed realizations, so that we can send _one_
+            # dataChange signal for all realizations at the end
+            reals_changed: List[int] = []
+
             for real_id in iter_node.data[SORTED_REALIZATION_IDS]:
                 real = partial.data()[ids.REALS].get(real_id)
                 if not real:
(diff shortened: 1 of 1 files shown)


jondequinor:
test this please

Blunde1:
@berland this looks fine to me except for missing tests. But how to test the GUI? Any suggestions?

berland:
It would be interesting to mock an ert config that had `x*100` of forward models (a `SLEEP()` might suffice..) and `x*100` realizations and see if the GUI behaves differently with and without this PR?

Blunde1:
**Setup:** 
- added 99 jobs in `poly.ert` to a total of 100
- landed on active 100 realizations (i.e. _not_ 10000) due to the time-consuming aspect of it all

**Test without this PR**
The GUI takes a second or two to start (loading) but is responsive (albeit slightly lagging) when playing around in the details window (dragging to view different realizations and jobs, etc.).
image


**Test with this PR**
Everything runs and feels okay, perhaps slightly better than in the comparative run. The GUI is still responsive when playing around during run, but with the same small lags perhaps.
image

Blunde1:
@berland comments on the test-setup?

If you are okay with the test-setup, we can perhaps go to review?

berland:
When testing the poly case with 30 realizations, the number of emitted dataChanged goes from 121 to 31 during an ensemble experiment with this PR.

ert_gui/model/snapshot.py
```diff
@@ -238,6 +238,24 @@ def _add_partial_snapshot(self, partial: PartialSnapshot, iter_: int):
...
+                        stack.callback(
+                            self.dataChanged.emit, job_top_left, job_bottom_right
+                        )
+
+            if reals_changed:
+                real_top_left = self.index(min(reals_changed), 0, iter_index)
+                real_bottom_right = self.index(
+                    max(reals_changed), self.columnCount(iter_index) - 1, iter_index
+                )
+                stack.callback(self.dataChanged.emit, real_top_left, real_bottom_right)
```
berland, line 257: There is a tradeoff being made here. If say only two realizations are changed, the first and the last one in a large ensemble, this `emit` will target every realization with this PR instead of only the two in question.

berland: The tradeoff is ok. Realizations are dominantly consecutive, and the cost of updating redundant indices in QT is probably small.

berland (approved):

PR #3624, opened 2022-07-14
config_parser: Make paths absolute

pinkwah:


libres/lib/config/config_content.cpp
```diff
@@ -70,7 +70,7 @@ config_content_type *config_content_alloc(const char *filename) {
     content->config_file = util_alloc_string_copy(filename);
     content->abs_path = util_alloc_abs_path(filename);
     content->config_path = util_split_alloc_dirname(content->abs_path);
-    content->invoke_path = config_root_path_alloc(NULL);
```
pinkwah, line 73: when the first argument of `config_root_path_alloc` is `NULL`, it uses `getcwd`. This value is never not-`NULL` except for the `config_root_path` tests. So replacing with `fs::current_path()` is fine.

oysteoh (approved):

PR #3626, opened 2022-07-24
Add annotation of errors

eivindjahren:
Adds annotation to PRs to catch linting errors in review. A simple mypy fix in `queue.py` and unused variable in `ies.cpp` is added in order to see the system working as only files with changes in the PR are annotated.

## Pre review checklist

### DIFF
--- a/.github/clang-tidy-matcher.json
+++ b/.github/clang-tidy-matcher.json
@@ -1,0 +1,32 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "clang-tidy-error",
+      "severity": "error",
+      "pattern": [
+        {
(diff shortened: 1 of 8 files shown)


dafeda (approved):
If this turns out to be helpful, we should consider not running pylint from the style-workflow.

eivindjahren:
> If this turns out to be helpful, we should consider not running pylint from the style-workflow.

Agreed, that was my goal. Currently, that workflow is not giving contributors enough feedback as most linting issues are disabled. This approach makes it visible to everyone when a nearby linting issue could be resolved so they are hopefully resolved and new issues not added.

PR #3662, opened 2022-08-01
Fix duplicate missing package data in setup.py

eivindjahren:
Fixes duplicate ert3_examples field in package data


dafeda (approved):
Yes please 👍

PR #3665, opened 2022-08-01
Refactor ert/analysis import structure

eivindjahren:
**Issue**
Resolves circular dependencies due to `ert/analysis`. Also resolves some typing errors as mypy uses strict=True inside `ert/analysis`.

This breaks imports in semeio that we will need to resolve. This is done in https://github.com/equinor/semeio/pull/443.

**Approach**
Avoid circular import due to import `ert.analysis` from `res.enkf.es_update` which is imported from `ert_shared` which is imported from `ert` by moving `res.enkf.es_update` to `ert.analysis`. This has the added benefit of making `ert.analysis.ies` private to `ert.analysis`.

## Pre review checklist


dafeda (approved):
👍

oyvindeide (approved):

Issue #3639, opened 2022-07-27
Very flaky test: test_experiment_server


eivindjahren:
Experiment server is failing quite consistently with:
```
=================================== FAILURES ===================================
__________________ test_experiment_server_ensemble_experiment __________________

tmpdir = local('/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/test_experiment_server_ensembl0')
source_root = PosixPath('/Users/runner/work/ert/ert')
capsys = <_pytest.capture.CaptureFixture object at 0x139fbba60>

    @pytest.mark.skipif(sys.version_info < (3, 7), reason="requires python3.7 or higher")
    @pytest.mark.integration_test
    @pytest.mark.timeout(20)
    def test_experiment_server_ensemble_experiment(tmpdir, source_root, capsys):
        shutil.copytree(
            os.path.join(source_root, "test-data", "local", "poly_example"),
            os.path.join(str(tmpdir), "poly_example"),
        )
    
        with tmpdir.as_cwd():
            parser = ArgumentParser(prog="test_main")
            parsed = ert_parser(
                parser,
                [
                    ENSEMBLE_EXPERIMENT_MODE,
                    "poly_example/poly.ert",
                    "--port-range",
                    "1024-65535",
                    "--enable-experiment-server",
                    "--realizations",
                    "0-4",
                ],
            )
    
            FeatureToggling.update_from_args(parsed)
    
>           run_cli(parsed)

ert_tests/cli/test_integration_cli.py:264: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../../hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/ert_shared/cli/main.py:56: in run_cli
    asyncio.run(
../../../../hostedtoolcache/Python/3.8.13/x64/lib/python3.8/asyncio/runners.py:44: in run
    return loop.run_until_complete(main)
../../../../hostedtoolcache/Python/3.8.13/x64/lib/python3.8/asyncio/base_events.py:603: in run_until_complete
    self.run_forever()
../../../../hostedtoolcache/Python/3.8.13/x64/lib/python3.8/asyncio/base_events.py:570: in run_forever
    self._run_once()
[...]

PR #3682, opened 2022-08-02
Ensure communication pipe file descriptor is closed

berland:
**Issue**
Resolves  ~#3639 and~ #3677 

**Approach**
Close it.

## Pre review checklist

### DIFF
--- a/src/ert_shared/services/_base_service.py
+++ b/src/ert_shared/services/_base_service.py
@@ -220,6 +220,7 @@ class _Proc(threading.Thread):
         file = Path(f"{self._service_name}_server.json")
         if file.exists():
             file.unlink()
+        os.close(self._comm_pipe.fileno())
 
     @property
     def logger(self) -> Logger:
--- a/tests/ert_tests/services/test_storage_service.py
+++ b/tests/ert_tests/services/test_storage_service.py
@@ -6,7 +6,6 @@ from ert_shared.services import Storage, _storage_main
 from ert_shared import port_handler
 
 
-@pytest.mark.skip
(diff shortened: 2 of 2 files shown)


lars-petter-hauge (approved):
Thanks for fixing!

xjules:
We (@lars-petter-hauge) still need to confirm that #3639 is caused by this since currently it looks more like having a huge discrepancy in running time for the actual test function; ranging from `7secs` to like `28secs`. This means just increasing timeout should fix it.

lars-petter-hauge:
Yes, we can remove resolving #3639. We ran main without timeout on the test failing in said issue and then it didn't fail on 100 runs. See also https://github.com/equinor/ert/issues/3639#issuecomment-1202702874 for timings observed during running.

berland:
Tried asserting file descriptors are closed by wrapping the test with the context manager:

```
class AssertNoFileDescriptorLeaks:
    def __init__(self):
        self.fd_dir: Path = Path("/proc/") / str(os.getpid()) / "fd"

    def __enter__(self):
        self.fd_count_at_init = len(list(self.fd_dir.glob("*")))
        return self

    def __exit__(self, type, value, traceback):
        fd_count_now = len(list(self.fd_dir.glob("*")))
        print(f"fd count before: {self.fd_count_at_init}, now: {fd_count_now}\n")
        assert fd_count_now == self.fd_count_at_init
```

but this reveals no change with or without the commit in this PR that closes `fd_read`.

berland:
Keep closed until further investigations finds that this is really an issue.

PR #3692, opened 2022-08-03
Move docs/rst/manual/ to just docs/

pinkwah:
Additionally, removes unused documentation (latex) and changes author to SCOUT rather than Joakim Hove specifically.


sondreso (approved):
Very nice!

Issue #3595, opened 2022-07-04
Reorganization of module structure


eivindjahren:
The long term idea is that all the ert code resides in `src/ert`. In general, the module structure should support rapid development by avoiding circular imports and shotgun surgery. To do so one needs to have modules that have single responsibilities, depend (import) in the direction of stable abstract modules, and supports adding new features by adding code, not changing it. That will be an ongoing issue for the lifetime of the project, but we need to start somewhere. These following first steps are suggested (Open new issue for each of the following):

- [x] #3596
- [x] `ert3` is moved to `src/ert/ert3`
- [x] `ert3_examples` is moved to `src/ert/ert3/examples`
- [x] `ert_data` is moved to `src/ert/data`
- [x] `ert_logging` is moved to `src/ert/logging`
- [x] `ert_shared` is moved to `src/ert/shared`
- [x] `job_runner` is moved to `src/ert/job_runner`.
- [x] `libres` is moved to `src/clib`
- [x] `res` is moved to `src/ert/_c_wrappers`

sondreso:
I think this is a very good idea, and a good ordering here is probably to go from top to bottom. See also https://github.com/equinor/ert/discussions/2227

eivindjahren:
All the things are moved, we just need to deal with the fallout.

PR #3696, opened 2022-08-04
Move out res.test.synthesizer to separate package oil_reservoir_synthesizer

eivindjahren:
**Issue**
More setup for #3595

**Approach**
`res.test.synthesizer` is used by more projects than ert. However, it isn't really part of what ert does. Therefore this functionality is split into its own package. This code then benefits from being independent, which should be good as it:

1) Only depends on python standard library
2) Only exposes one class.

In order to not install this code with ert, snake_oil is removed from ert3_examples. When this functionality is needed later we can add more examples through a plugin architecture.

## Pre review checklist


eivindjahren:
Jenkins is failing due to an unrelated breakage in semeio (rename of `ErtRunContext` to `RunContext`).

oyvindeide (approved):

berland (approved):

PR #3717, opened 2022-08-08
Remove ERT splash + contextmanage GUILogHandler

pinkwah:
**Remove ERT Splash Window**
The vast majority of ERT's initialisation time comes from Python and all
of the imports that are required by ERT. By the time the splash screen
is displayed this work has already been done.

Additionally, we delay ERT's startup by an additional two seconds to show
the splash screen. This is just annoying. (Changelog: "Improved ERT GUI
startup times!")

**Add contextmanager for GUILogHandler**
Without this, anything logged after Qt exits will still attempt to
log to the GUILogHandler, causing an exception. This commit uses a
contextmanager to cleanly remove the handler after exit.


dafeda (approved):
Real nice 👍

PR #3724, opened 2022-08-09
Use id for ensemble instead of ensemble evaluator

### LINKED ISSUE #3723: Ensemble should have id instead of ensemble evaluator
Currently each job, step and realization has an id, while the ensemble does not. Instead the ensemble evaluator has an id (`ee_id`), which for instance is propagated to the dispatchers on the cluster side and used in the source when sending cloud events. Now also an experiment is granted an id.

We want to be more consistent and thus should refactor ensemble / ensemble evaluator to be in line with the rest of the structure.

lars-petter-hauge:
Instead of each ensemble evaluators having an ee_id, we specify an id
for each individual ensemble. This makes more sense when referring to an
entity (such as realization, job, and now also ensemble). Will be more
consistent with an experiment having an id as well.

As with an experiment, the id is not created on initialization but
rather set after creation. This means the caller will be responsible for
creating the id instead of the object itself.

**Issue**
Resolves #3723

**Approach**
Refactor `EnsembleEvaluator` and `_Ensemble` and move the `id` to the `Ensemble`. We use the same approach as for `Experiment` and have the caller of the ensemble to create and set the `id`. 

## Pre review checklist

### DIFF
--- a/src/ert/ensemble_evaluator/builder/_ensemble.py
+++ b/src/ert/ensemble_evaluator/builder/_ensemble.py
@@ -27,25 +27,30 @@ logger = logging.getLogger(__name__)
 
 class _Ensemble:
     def __init__(
-        self, reals: Sequence[_Realization], metadata: Mapping[str, Any]
+        self, reals: Sequence[_Realization], metadata: Mapping[str, Any], id_: str
     ) -> None:
         self.reals = reals
         self.metadata = metadata
         self._snapshot = self._create_snapshot()
         self.status = self._snapshot.status
         self._status_tracker = EnsembleStateTracker(self._snapshot.status)
+        self._id: str = id_
 
     def __repr__(self) -> str:
         return f"Ensemble with {len(self.reals)} members"
 
-    def evaluate(self, config: "EvaluatorServerConfig", ee_id: str) -> None:
+    def evaluate(self, config: "EvaluatorServerConfig") -> None:
         pass
 
-    async def evaluate_async(self, config: "EvaluatorServerConfig", ee_id: str) -> None:
+    async def evaluate_async(self, config: "EvaluatorServerConfig") -> None:
         pass
 
     def cancel(self) -> None:
         pass
+
+    @property
+    def id_(self) -> str:
+        return self._id
 
     @property
     def cancellable(self) -> bool:
--- a/src/ert/ensemble_evaluator/builder/_ensemble_builder.py
+++ b/src/ert/ensemble_evaluator/builder/_ensemble_builder.py
@@ -40,6 +40,7 @@ class _EnsembleBuilder:  # pylint: disable=too-many-instance-attributes
         self._max_retries = 0
         self._retry_delay = 5
         self._executor: str = "local"
+        self._id: Optional[str] = None
 
     def set_forward_model(
         self, forward_model: _RealizationBuilder
@@ -100,6 +101,10 @@ class _EnsembleBuilder:  # pylint: disable=too-many-instance-attributes
 
     def set_executor(self, executor: str) -> "_EnsembleBuilder":
         self._executor = executor
+        return self
+
+    def set_id(self, id_: str) -> "_EnsembleBuilder":
+        self._id = id_
         return self
 
     @staticmethod
@@ -187,6 +192,9 @@ class _EnsembleBuilder:  # pylint: disable=too-many-instance-attributes
         if not (self._reals or self._forward_model):
             raise ValueError("Either forward model or realizations needs to be set")
 
+        if self._id is None:
+            raise ValueError("ID must be set prior to building")
+
         real_builders: List[_RealizationBuilder] = []
         if self._forward_model:
             # duplicate the original forward model into realizations
@@ -205,7 +213,9 @@ class _EnsembleBuilder:  # pylint: disable=too-many-instance-attributes
         reals = [builder.build() for builder in real_builders]
 
         if self._legacy_dependencies:
-            return _LegacyEnsemble(reals, self._metadata, *self._legacy_dependencies)
+            return _LegacyEnsemble(
+                reals, self._metadata, *self._legacy_dependencies, id_=self._id
+            )
         else:
             return PrefectEnsemble(
                 reals=reals,
@@ -216,4 +226,5 @@ class _EnsembleBuilder:  # pylint: disable=too-many-instance-attributes
                 executor=self._executor,
                 retry_delay=self._retry_delay,
                 custom_port_range=self._custom_port_range,
+                id_=self._id,
             )
--- a/src/ert/ensemble_evaluator/builder/_legacy.py
+++ b/src/ert/ensemble_evaluator/builder/_legacy.py
@@ -37,14 +37,15 @@ experiment_logger = logging.getLogger("ert.experiment_server")
 
 
 class _LegacyEnsemble(_Ensemble):
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments
         self,
         reals: List[_Realization],
         metadata: Dict[str, Any],
         queue_config: QueueConfig,
         analysis_config: AnalysisConfig,
+        id_: str,
     ) -> None:
-        super().__init__(reals, metadata)
+        super().__init__(reals, metadata, id_)
         if not queue_config:
             raise ValueError(f"{self} needs queue_config")
         if not analysis_config:
@@ -52,7 +53,6 @@ class _LegacyEnsemble(_Ensemble):
         self._job_queue = queue_config.create_job_queue()
         self._analysis_config = analysis_config
         self._config: Optional["EvaluatorServerConfig"] = None
-        self._ee_id: Optional[str] = None
         self._output_bus: "asyncio.Queue[CloudEvent]" = asyncio.Queue()
 
     def setup_timeout_callback(
@@ -65,7 +65,7 @@ class _LegacyEnsemble(_Ensemble):
             timeout_cloudevent = CloudEvent(
                 {
                     "type": identifiers.EVTYPE_FM_STEP_TIMEOUT,
-                    "source": f"/ert/ee/{self._ee_id}/real/{run_args.iens}/step/0",
+                    "source": f"/ert/ensemble/{self.id_}/real/{run_args.iens}/step/0",
                     "id": str(uuid.uuid1()),
                 }
             )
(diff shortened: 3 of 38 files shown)


src/ert/ensemble_evaluator/builder/_legacy.py
```diff
@@ -172,7 +169,7 @@ async def _evaluate_inner(
         result = CloudEvent(
             {
                 "type": identifiers.EVTYPE_ENSEMBLE_STOPPED,
-                "source": f"/ert/ee/{self._ee_id}/ensemble",
```
lars-petter-hauge, line 175: Have yet to make sure that this event is handled properly when the `ensemble` suffix is removed.

src/ert/ensemble_evaluator/builder/_ensemble.py
```diff
@@ -34,14 +34,27 @@ def __init__(
         self._snapshot = self._create_snapshot()
         self.status = self._snapshot.status
         self._status_tracker = EnsembleStateTracker(self._snapshot.status)
+        self._id: Optional[str] = None
+
+    @property
+    def id_(self) -> str:
+        if not self._id:
+            raise RuntimeError(f"{self} does not have an ID")
+        return self._id
```
lars-petter-hauge, line 43: Not sure if I like the property name. Although I get we can't use `id` (built-in), I don't think this is very readable, as it looks a bit too much like accessing a private property..

src/ert_shared/models/base_run_model.py
```diff
@@ -302,13 +302,14 @@ def run_ensemble_evaluator(
             self.ert().resConfig(),
         ).build()
 
+        ensemble.id = str(uuid.uuid1()).split("-", maxsplit=1)[0]
```
lars-petter-hauge, line 305: Would be better if this can be set as part of the builder I believe.

src/ert/ensemble_evaluator/builder/_ensemble.py
xjules: I think for the sake of consistency (`.id_` is used for all the sub-ensemble components) we can keep `id_`.

src/ert/ensemble_evaluator/builder/_prefect.py
```diff
@@ -271,7 +269,7 @@ def _evaluate(self) -> None:
                 token=self._ee_con_info.token,
                 cert=self._ee_con_info.cert,
             ):
-                self.run_flow(self._ee_id)
+                self.run_flow(self.id_)
```
xjules, line 272: why to send `self.id_` as a parameter; as it is the attribute?

src/ert_shared/ensemble_evaluator/evaluator.py
```diff
@@ -312,7 +311,7 @@ def _run_server(self, loop):
 
     def run(self) -> ee_monitor._Monitor:
         self._ws_thread.start()
-        self._ensemble.evaluate(self._config, self._ee_id)
+        self._ensemble.evaluate(self._config, self.ensemble.id_)
```
xjules, line 314: is `self.ensemble.id_` a necessary parameter?

src/ert_shared/models/base_run_model.py
xjules: I agree. It shouldn't leak into `base_run_model`.

tests/ert_tests/ensemble_evaluator/ensemble_evaluator_utils.py
```diff
@@ -208,25 +208,25 @@ def with_failure(self):
 
 class AutorunTestEnsemble(TestEnsemble):
     # pylint: disable=arguments-differ
-    def _evaluate(self, client_url, dispatch_url, ee_id):
-        super()._evaluate(dispatch_url, ee_id)
+    def _evaluate(self, client_url, dispatch_url):
+        super()._evaluate(dispatch_url, self.ensemble.id_)
```
xjules, line 212: Same redundant usage of `self.ensemble.id_`.

lars-petter-hauge:
Draft PR, there are tests not passing and places I haven't updated yet.

src/ert_shared/ensemble_evaluator/evaluator.py
lars-petter-hauge: removed in update

lars-petter-hauge:
There are still some submodules that need to be fixed, but there's progress!

src/ert/ensemble_evaluator/builder/_legacy.py
lars-petter-hauge: Did not find any reasons for this suffix to change the behaviour anywhere

lars-petter-hauge:
Unsure why typechecking fails on things I haven't been near... But I guess I'll fix it..

(`src/ert/ert3/evaluator/_evaluator.py:129: error: Call to untyped function "EnsembleEvaluator" in typed context`)

src/ert/ensemble_evaluator/builder/_ensemble_builder.py
```diff
@@ -204,10 +209,11 @@ def build(self) -> _Ensemble:
 
         reals = [builder.build() for builder in real_builders]
 
+        ens: _Ensemble
         if self._legacy_dependencies:
-            return _LegacyEnsemble(reals, self._metadata, *self._legacy_dependencies)
+            ens = _LegacyEnsemble(reals, self._metadata, *self._legacy_dependencies)
```
xjules, line 214: `.set_id(self._id)`?

src/ert/ensemble_evaluator/builder/_ensemble_builder.py
```diff
@@ -217,3 +223,6 @@ def build(self) -> _Ensemble:
                 retry_delay=self._retry_delay,
                 custom_port_range=self._custom_port_range,
             )
+        if self._id is not None:
```
xjules, line 226: Can you just use the same as before, but using the `_LegacyEnsemble(reals, self._metadata, *self._legacy_dependencies).set_id(self.id_)` in both returns?

lars-petter-hauge: The default value of `id` is `None`. I was uncertain if there should be a default `uuid` set instead?

As the implementation of setting an `id` enforces the caller to only set it once, I wondered if it would be too strict to force them to include it in the builder. This way the `id` can be included in the builder or be added post build.

I am very open to suggestions here.. Should we fail if the id is not set in the builder?

src/ert_shared/models/base_run_model.py
lars-petter-hauge: I think the result of the discussion https://github.com/equinor/ert/pull/3724#discussion_r947617303 will affect how this should look. Let's resolve that first.

src/ert/ensemble_evaluator/builder/_ensemble_builder.py
xjules: > Should we fail if the id is not set in the builder?` 

I think the `id` should be set only in the builder, what I mean is why to set it post build? Therefore my opinion would be like "Yes, we should fail." :)

lars-petter-hauge: So then it should not have a default either, and fail if the caller has not explicitly set the id on the builder before calling `build`?

xjules: Currently it's set when constructing `EnsembleEvaluator` https://github.com/equinor/ert/blob/1764672345ceddcf75b688b097ee5f5572384ffc/src/ert/shared/models/base_run_model.py#L315, we might just inject it as a parameter for `_Ensemble` too or what do you think?

src/ert/shared/models/base_run_model.py
```diff
@@ -302,13 +302,14 @@ def run_ensemble_evaluator(
             self.ert().resConfig(),
         ).build()
 
+        ensemble.id_ = str(uuid.uuid1()).split("-", maxsplit=1)[0]
```
xjules, line 305: This looks a bit out of place and referring to our discussion can it be shipped as a parameter in the constructor?

lars-petter-hauge: hence this comment: https://github.com/equinor/ert/pull/3724#discussion_r947745474 ;)

src/ert/ensemble_evaluator/builder/_ensemble_builder.py
lars-petter-hauge: Am I understanding correctly if I suggest the following?

```diff
diff --git a/src/ert/shared/models/base_run_model.py b/src/ert/shared/models/base_run_model.py
index a59d715d3..2cd67a132 100644
--- a/src/ert/shared/models/base_run_model.py
+++ b/src/ert/shared/models/base_run_model.py
@@ -294,13 +294,17 @@ class BaseRunModel:
     def run_ensemble_evaluator(
         self, run_context: RunContext, ee_config: EvaluatorServerConfig
     ) -> int:
-        ensemble = EnsembleBuilder.from_legacy(
+        ensemble = (EnsembleBuilder.from_legacy(
             run_context,
             self.get_forward_model(),
             self._queue_config,
             self.ert().analysisConfig(),
             self.ert().resConfig(),
-        ).build()
+        )
+        .self.set_id(str(uuid.uuid1()).split("-", maxsplit=1)[0])
+        .build()
+        )
+
 
         self.ert().initRun(run_context)
 
@@ -308,7 +312,6 @@ class BaseRunModel:
             ensemble,
             ee_config,
             run_context.iteration,
-            ee_id=str(uuid.uuid1()).split("-", maxsplit=1)[0],
         ).run_and_get_successful_realizations()
 
         self.deactivate_failed_jobs(run_context)
```

xjules: Hm, that's also an option I think it's fine :)

lars-petter-hauge: Alright, and given:

>  Therefore my opinion would be like "Yes, we should fail." :)

There will then be raised an exception if anyone tries to execute `build` without setting the id first.

lars-petter-hauge: > Furthermore, being consistent with how the Experiment protocol is used in e.g. BaseRunModel, I do not include the id as part of the __init__. Instead the id is set after an Ensemble is created (but this will also be done within the builder).

So, I decided to disagree with myself.. I don't like the pattern of creating an instance of an object and then remembering to always set the property afterwards. I think the `id` should be required on init instead. The builder will work as proposed above though!

I also think we should change this in any `Experiment` classes.

src/ert/shared/models/base_run_model.py
lars-petter-hauge: Alright, out of place should be fixed!

src/ert/ensemble_evaluator/builder/_ensemble_builder.py
lars-petter-hauge: I've included the changes and pushed it now, please let me know if you agree on how it looks :)

xjules (approved):
Very good job! Just squash those two commits and ready to :rocket:

lars-petter-hauge:
Rebased and incorporated all changes into the protobuf PR as well. Commits squashed!

PR #3736, opened 2022-08-10
Resolve circular imports

eivindjahren:
Solves some circular imports in preparation for resolving #3595 

This is done by moving `simulators` into `ert` and moving the constant file names used by simulators and `job_queue` into `ert.constant_filenames`. `LibresFacade` is also moved from `ert` to `ert_shared` which avoids importing `job_queue` when importing `LibresFacade`. Because the ert package is checked with strict mypy, this means adding a lot of more typing.

This hasn't completely removed problems with circular imports which can still happen when you go from `ert` -> `ert_shared` -> `res.job_queue` -> `ert.constant_filenames`. However, this is much less likely to happen, and by restructuring `res.job_queue`, `ert_shared.hook_implementations`, or `ert_shared.plugins` we might be able to get rid of it all together. Another possibility is changing `ert.constant_filenames` to a config file which is loaded on import rather than imported.

## Pre review checklist


dafeda (approved):
🚀

PR #3755, opened 2022-08-12
Change dates written by ERT to ISO-8601

berland:
* The magic string `<DATE>` now writes 2022-01-01 instead of 01/01/2022. This is a breaking change.
* Document the magic string DATE.
* Change other occurences of old date formats in mostly error messages

**Issue**
Resolves missing ISO-8601 fixes.

**Approach**
Change to ISO-8601

## Pre review checklist

### DIFF
--- a/src/libres/lib/enkf/enkf_main.cpp
+++ b/src/libres/lib/enkf/enkf_main.cpp
@@ -639,8 +639,8 @@ static void update_case_log(enkf_main_type *enkf_main, const char *case_path) {
             util_set_datetime_values_utc(now, &second, &minute, &hour, &day,
                                          &month, &year);
 
-            fprintf(stream, "TIME:%02d/%02d/%4d-%02d.%02d.%02d\n", day, month,
-                    year, hour, minute, second);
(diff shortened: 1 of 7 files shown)


src/libres/lib/enkf/enkf_main.cpp
```diff
@@ -639,8 +639,8 @@ static void update_case_log(enkf_main_type *enkf_main, const char *case_path) {
             util_set_datetime_values_utc(now, &second, &minute, &hour, &day,
                                          &month, &year);
 
-            fprintf(stream, "TIME:%02d/%02d/%4d-%02d.%02d.%02d\n", day, month,
-                    year, hour, minute, second);
+            fprintf(stream, "TIME:%04d-%02d-%02dT%02d:%02d:%02d\n", year, month,
```
xjules, line 642: There is also `std::put_time`, which can be used to produce the correct isoformat. But not necessary I guess.

berland: I don't think there is a test for this so maybe better to be conservative (and aim for Python reimplementation?)

xjules (approved):
Nice! :rocket:

PR #3762, opened 2022-08-15
Add <except.hpp> for convenient C++ exceptions

pinkwah:
The only difference is that these exceptions have an additional
constructor allowing fmtlib format strings. It makes throwing code
slightly shorter and nicer.


eivindjahren (approved):
pre-approved!

PR #3637, opened 2022-07-26
Experiment server API for all RunModels

xjules:
**Issue**
Resolves #3413 

**Approach**
Based on Jonase's 655cdc8797aea6a2fdd592b95936cca3291ac5e4 draft, which implements API for the most of the runmodels.

## Pre review checklist

- [x] Added appropriate release note label
- [x] PR title captures the intent of the changes, and is fitting for release notes.
- [ ] Commit history is consistent and clean, in line with the [contribution guidelines](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md).

Adding labels helps the maintainers when writing release notes. This is the [list of release note labels](https://github.com/equinor/ert/labels?q=release-notes).

PR #3438, opened 2022-05-25
Introduce basic experiment server

jondequinor:
**Issue**
Resolves #3414 

**Approach**
- Add feature flag: `--enable-experiment-server`
- Allow execution of an _Ensemble Experiment_ via the experiment server
- Add basic functionality, like printing whether or not the experiment
  was successful.

The experiment server is created by instantiating `ExperimentServer`. This starts a websockets server that listens for communication from `job_dispatch.py` (the remote, distributed workers). An experiment is created per usual, using `create_model`, but it is owned by the `ExperimentServer`. This experiment implements the `Experiment` protocol (see [pep-0544](https://peps.python.org/pep-0544/). We immediately run the experiment, which will initiate three event "flows":

1. the `EnsembleExperiment.run()` implementation will at different stages "dispatch" CloudEvents that describes what happened to the experiment, e.g. the event `com.equinor.ert.experiment.started` and `com.equinor.ert.experiment.hook_ended`.
2. During the evaluation of an ensemble, the ensemble puts CloudEvents on a "public facing" `asyncio.Queue` that the experiment then dispatches. These are events like `com.equinor.ert.ensemble.started` and `com.equinor.ert.ensemble.failed`, but also events emitted by the job queue, like `com.equinor.ert.forward_model_step.waiting` and `com.equinor.ert.forward_model_step.timeout`.
3. The queue runs the `job_dispatch.py` program on the cluster (or locally), which will send its events to the experiment server. These events are, based on the source, dispatched to the correct experiment. The events include `com.equinor.ert.forward_model_job.start` and `com.equinor.ert.forward_model_job.running`.

At the moment, the dispatch method of the experiment simply counts the successful realizations and stores them in the _state machine_.

When the experiment is done, the server shuts down.

## Pre review checklist

- [x] Added appropriate release note label
- [x] PR title captures the intent of the changes, and is fitting for release notes.
- [x] Commit history is consistent and clean, in line with the [contribution guidelines](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md).

Adding labels helps the maintainers when writing release notes. This is the [list of release note labels](https://github.com/equinor/ert/labels?q=release-notes).

PR #3768, opened 2022-08-16
Use Experiment server in all run models when feature is enabled

### LINKED ISSUE #3413: Implement experiment interface in all run models
The current interface is 
```python
class Experiment(Protocol):
    """The experiment protocol which all experiments must implement."""

    async def run(self, evaluator_server_config: "EvaluatorServerConfig") -> None:
        """Run the experiment to completion."""
        pass

    async def dispatch(self, event: CloudEvent, iter_: int) -> None:
        """Pass an event for iter_ to the experiment. The experiment will internalize
        the event and update its state."""
        pass

    # TODO: this is preliminary, see https://github.com/equinor/ert/issues/3407
    async def successful_realizations(self, iter_: int) -> int:
        pass

    @property
    def id_(self) -> str:
        """The id of the experiment."""
        pass

    @id_.setter
    def id_(self, value: str) -> None:
        """Set the id of the experiment. It should not be possible to set this
        more than once."""
        pass

```

berland:
**Issue**
Resolves #3413

**Approach**
`runSimulations()` in each run model has been ported to an async function alongside that uses the experiment server.

## Pre review checklist

### DIFF
--- a/src/ert/shared/models/ensemble_experiment.py
+++ b/src/ert/shared/models/ensemble_experiment.py
@@ -1,10 +1,7 @@
 import asyncio
 import concurrent
 import logging
-import uuid
 from typing import Any, Dict
-
-from cloudevents.http import CloudEvent
 
 from ert._c_wrappers.enkf import RunContext
 from ert._c_wrappers.enkf.enkf_main import EnKFMain, QueueConfig
@@ -29,24 +26,16 @@ class EnsembleExperiment(BaseRunModel):
 
     async def run(self, evaluator_server_config: EvaluatorServerConfig) -> None:
 
-        # Send EXPERIMENT_STARTED
         experiment_logger.debug("starting ensemble experiment")
-        await self.dispatch(
-            CloudEvent(
-                {
-                    "type": identifiers.EVTYPE_EXPERIMENT_STARTED,
-                    "source": f"/ert/experiment/{self.id_}",
-                    "id": str(uuid.uuid1()),
-                }
-            ),
-            0,
+        await self._dispatch_ee(
+            identifiers.EVTYPE_EXPERIMENT_STARTED,
+            iteration=0,
         )
 
         loop = asyncio.get_running_loop()
         with concurrent.futures.ThreadPoolExecutor() as executor:
             run_context = await loop.run_in_executor(executor, self.create_context)
 
-            # Create runpaths
             experiment_logger.debug("creating runpaths")
             await loop.run_in_executor(
                 executor,
@@ -60,11 +49,10 @@ class EnsembleExperiment(BaseRunModel):
                 HookRuntime.PRE_SIMULATION, run_context.iteration, loop, executor
             )
 
-            # Evaluate
             experiment_logger.debug("evaluating")
             await self._evaluate(run_context, evaluator_server_config)
 
-            num_successful_realizations = self._state_machine.successful_realizations(
+            num_successful_realizations = await self.successful_realizations(
                 run_context.iteration
             )
 
@@ -75,19 +63,12 @@ class EnsembleExperiment(BaseRunModel):
                 self.checkHaveSufficientRealizations(num_successful_realizations)
             except ErtRunError as e:
 
-                # Send EXPERIMENT_FAILED
-                await self.dispatch(
-                    CloudEvent(
-                        {
-                            "type": identifiers.EVTYPE_EXPERIMENT_FAILED,
-                            "source": f"/ert/experiment/{self.id_}",
-                            "id": str(uuid.uuid1()),
-                        },
-                        {
-                            "error": str(e),
-                        },
-                    ),
-                    run_context.iteration,
+                await self._dispatch_ee(
+                    identifiers.EVTYPE_EXPERIMENT_FAILED,
+                    data={
+                        "error": str(e),
+                    },
+                    iteration=run_context.iteration,
                 )
                 return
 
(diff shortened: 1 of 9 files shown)


xjules:
You can check this one too: https://github.com/equinor/ert/pull/3637 and close it afterwards.

src/ert/shared/models/ensemble_smoother.py
```diff
@@ -24,6 +34,178 @@ def setAnalysisModule(self, module_name: str) -> None:
...
+            rerun_context,
+        )
+
+        await self._run_hook(
+            HookRuntime.PRE_SIMULATION, rerun_context.iteration, loop, threadpool
+        )
+
+        # Push ensemble, parameters, observations to new storage
+        ensemble_id = await loop.run_in_executor(
+            threadpool, self._post_ensemble_data, update_id
```
berland, line 176: `_post_ensemble_data` calls `self.setPhaseName()` which should maybe invoke a call to the experiment server?

berland: That will happen later when the experiment server supports it.

src/ert/shared/models/ensemble_experiment.py
```diff
@@ -59,11 +57,10 @@ async def run(self, evaluator_server_config: EvaluatorServerConfig) -> None:
                 HookRuntime.PRE_SIMULATION, run_context.iteration, loop, executor
             )
 
-            # Evaluate
             experiment_logger.debug("evaluating")
             await self._evaluate(run_context, evaluator_server_config)
 
-            num_successful_realizations = self._state_machine.successful_realizations(
+            num_successful_realizations = await self.successful_realizations(
```
xjules, line 63: Agree, makes more sense to use the defined member :+1

src/ert/shared/models/ensemble_smoother.py
```diff
@@ -24,6 +34,175 @@ def setAnalysisModule(self, module_name: str) -> None:
...
+                {
+                    "type": identifiers.EVTYPE_EXPERIMENT_STARTED,
+                    "source": f"/ert/experiment/{self.id_}",
+                    "id": str(uuid.uuid1()),
+                }
+            ),
+            prior_context.iteration,
+        )
+
+        self._checkMinimumActiveRealizations(prior_context)
```
xjules, line 55: `loop.run_in_executor(` ? Not sure where should be a threshold for saying this function is "atomic" does not need to be async.

src/ert/shared/models/ensemble_smoother.py
```diff
@@ -24,6 +34,175 @@ def setAnalysisModule(self, module_name: str) -> None:
...
+                ),
+                prior_context.iteration,
+            )
+            raise ErtRunError(
+                f"Analysis of simulation failed with the following error: {e}"
+            ) from e
+
+        experiment_logger.debug("update complete")
+
+        await self.dispatch(
```
xjules, line 132: We might consider creating a helper class for `dispatch`, which automatically inject the correct source string.

src/ert/shared/models/ensemble_smoother.py
```diff
@@ -24,6 +34,175 @@ def setAnalysisModule(self, module_name: str) -> None:
...
+            experiment_logger.exception("analysis failed")
+            await self.dispatch(
+                CloudEvent(
+                    {
+                        "type": identifiers.EVTYPE_EXPERIMENT_FAILED,
+                        "source": f"/ert/experiment/{self.id_}",
+                        "id": str(uuid.uuid1()),
+                    },
+                    {
+                        "error": str(e),
```
xjules, line 121: Wondering where this `error` field is actually used?

src/ert/shared/models/ensemble_smoother.py
berland: Tracing through the code, this seems to only involve trivial getters and an integer comparison. My gut feeling is then not to  
yield through await since there might be some overhead with our switching between sync and async context? (this is partly speculation)

xjules: You are right there actually! For reference: https://stackoverflow.com/questions/54813458/asyncio-performance-with-synchronous-code

src/ert/shared/models/ensemble_smoother.py
berland: Good idea, implemented. Unsure of the name I chose,   "_dispatch_ee"

src/ert/shared/models/ensemble_smoother.py
berland: Not sure, but I guess it does make sense to pass the error information on. The same data argument is used for hook names as far as I can see, maybe not picked up (yet!) either(?)

xjules: I see. I wonder mainly whether this kind of general "data" field should be present for Protobuf objects (which of-course it can't, but we can supplement some `error` fields at least) ...
Anyway, for this I understand you would like to keep regardless it's being consumed or not?

berland:
todo: Write test function for `_dispatch_ee`

xjules:
> todo: Write test function for `_dispatch_ee`

Considering `_dispatch_ee` again.... I think the code looks much more compact. Additional, since it only affects the evaluator than I would keep it without the test function since this call can be considered "atomic"; ie. essentially one liner ...

lars-petter-hauge:
Thanks for picking up on the continuation of https://github.com/equinor/ert/pull/3438. With this implementation though we are adding a fair bit of duplicated code for the `run` function, from what I can see because:
1) we want it to be `async`
2) we want the function called from `experiment_server` to also send events.

Could we not use the samme pattern that we already implement for `evaluate`?

There we have both `def evaluate` and `async def evaluate_async`, in which they both call `async def _evaluate_inner`. The synchronous version has to call the async version with `get_event_loop.run_until_complete`. 

I would really like us to avoid duplicating the `runSimulations` functionality, as it will be a real hassle to keep both up to date (we will likely fail..).

xjules:
> Thanks for picking up on the continuation of #3438. With this implementation though we are adding a fair bit of duplicated code for the `run` function, from what I can see because:
> 
> 1. we want it to be `async`
> 2. we want the function called from `experiment_server` to also send events.
> 
> Could we not use the samme pattern that we already implement for `evaluate`?
> 
> There we have both `def evaluate` and `async def evaluate_async`, in which they both call `async def _evaluate_inner`. The synchronous version has to call the async version with `get_event_loop.run_until_complete`.
> 
> I would really like us to avoid duplicating the `runSimulations` functionality, as it will be a real hassle to keep both up to date (we will likely fail..).

As discussed on our standup, we will do the "merging" process on one by one basis. Therefore, this can be merged as is.

xjules (approved):
Nice and LGTM! I linked to the corresponding issue to this PR.

PR #3793, opened 2022-08-22
Propagate ensemble id for source when building

lars-petter-hauge:
Ensemble id has been included as a property and should be included as
part of source and propagated through realizations, steps and jobs.

## Pre review checklist


src/ert/ensemble_evaluator/builder/_template.py
```diff
@@ -1,4 +1,4 @@
-_SOURCE_TEMPLATE_BASE = "/ert/ensemble/{ens_id}/"
```
lars-petter-hauge, line 1: Note the removal of `/` here (could perhaps be a commit of its own?). Whenever `source` was asked for, the returned value would be `/ert/ensemble/some_id//real...`, but I believe the only place we used the source for any logic was when we are asking for the ensemble id, hence the rest didn't matter much

berland (approved):
:+1:

PR #3794, opened 2022-08-22
Allow c++ as compiler in build script

valentin-krasontovitsch:
**Issue**
Did not open an issue for this one. should i? i can describe the problem though - working on ubuntu 22.04, for some reason the suggested compilers in the build script are not enough - my system is using `c++`.

**Approach**
i'm adding `c++` as an accepted compiler

## Pre review checklist

HELP! I have no idea how I should label this - is it an improvement? an enhancement? only developers gain from this... does it need a release notes label? would that be misc?

### DIFF
--- a/script/build
+++ b/script/build
@@ -41,7 +41,7 @@ def check_cxx_version() -> None:
     """
     Check that the user's compiler is compatible with ERT.
     """
-    cxx = shutil.which(os.environ.get("CXX", "c++"))
+    cxx = Path(shutil.which(os.environ.get("CXX", "c++"))).resolve()
     print_info(f"Using C++ compiler: {cxx}")
 
     version_line = subprocess.check_output([cxx, "--version"]).splitlines()[0].decode()
@@ -51,7 +51,7 @@ def check_cxx_version() -> None:
         sys.exit(1)
     version = tuple(int(version_match[x + 1]) for x in range(3))
 
-    if "GCC" in version_line:
+    if "GCC" in version_line or "g++" in version_line:
         min_version = MIN_GCC_VERSION
     elif "clang" in version_line:
         min_version = MIN_CLANG_VERSION


pinkwah (changes requested):
`c++` is a symlink to either GCC or Clang, and is used to determine your system's default compiler. It is not a compiler in itself (ie. there's no compiler called just `c++`)

Run `c++ --version`. Likely what's happening is that GCC is being written in lowercase `gcc` and the fix is to change line 54 to
```py
if "gcc" in version_line.lower():
```

valentin-krasontovitsch:
if only t'were so, Zohar. i checked the output of the version call as well, after reading what is actually causing the error. that don't help, the output literally says `c++`.

here's some digging:

```
~ ─> ><}}*> c++ --version
c++ (Ubuntu 11.2.0-19ubuntu1) 11.2.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

~ ─> ><}}*> ls -l (which c++)
lrwxrwxrwx 1 root root 21 May  3 14:10 /usr/bin/c++ -> /etc/alternatives/c++
~ ─> ><}}*> ls -l /etc/alternatives/c++
lrwxrwxrwx 1 root root 12 May  3 14:10 /etc/alternatives/c++ -> /usr/bin/g++
~ ─> ><}}*> ls -l /usr/bin/g++
lrwxrwxrwx 1 root root 6 Aug  5  2021 /usr/bin/g++ -> g++-11
~ ─> ><}}*> ls -l /usr/bin/g++-11
lrwxrwxrwx 1 root root 23 Mar 24 15:22 /usr/bin/g++-11 -> x86_64-linux-gnu-g++-11
~ ─> ><}}*> ls -l /usr/bin/x86_64-linux-gnu-g++-11
-rwxr-xr-x 1 root root 862976 Mar 24 15:22 /usr/bin/x86_64-linux-gnu-g++-11
```

i'll be happy to consider other alternative changes though that help in this situation : )

pinkwah:
Ubuntu's GCC is just using the name of the symlink that you called it from. That is ridiculously annoying. But it doesn't seem to be doing that for Clang:
```
root@efa4f3e9c8fc:/# c++ --version
Ubuntu clang version 14.0.0-1ubuntu1
```

Perhaps we should check whether the user is on Debian (it has the same behaviour) or Ubuntu and assume that the versions are okay. Python 3.10 features the function `from platform import freedesktop_os_release` but you can do the manually read the same file:
```py
def check_cxx_version() -> None:
    try:
        with open("/etc/os-release") as f:
            os_release = {
                words[0]: words[1].strip() for words in (line.split("=") for line in f)
            }
        os_id = os_release.get("ID_LIKE", os_release["ID"])
        if "debian" in os_id:
            # Debian and its forks (Ubuntu et al) mess with the version string;
            # assume that the user has a supported compiler
            return
    except IOError:
        # File doesn't exist (maybe macOS or unusal flavour of Linux), falling
        # back to normal check
        pass

    # Do the current `c++ --version` detection
```

This should do the trick on Debian, Ubuntu and Linux Mint, and probably most other Debian-based distros. The rest (RedHat/Fedora, Arch Linux) all are sane and print `GCC` in the version string.

valentin-krasontovitsch:
I'm sorry for being nitpicky / difficult, but it's not a symlink though - I have two different binaries that `c++` and `gcc` use, namely `/usr/bin/x86_64-linux-gnu-g++-11` and `/usr/bin/x86_64-linux-gnu-gcc-11`, they come from different packages, and it looks like there is an actual difference between the two, according to [a stackoverflow answer](https://stackoverflow.com/questions/172587/what-is-the-difference-between-g-and-gcc)

this neither seems insane, nor does it seem to call for a hack solution - but then again, I have not worked with C in a very long time, and have never worked with c++, so this is just a very blue-eyed impression i have.

I also see the following in gcc(1) (on my system), not sure how that helps in this discussion though :sweat_smile: :

> The usual way to run GCC is to run the executable called gcc, or machine-gcc when cross-compiling, or machine-gcc-version to run a specific version of GCC.  When you compile C++ programs, you should invoke GCC as g++ instead.

anyways maybe we shouldn't spend a lot of time arguing about how to interpret / treat a version string in a helper build script - if the script fails, it fails... i'm gonna use the solution you suggested.

pinkwah:
1. GCC is an acronym standing for the "GNU Compiler Collection".
2. `gcc` is the GNU C compiler, and is generically referred to as `cc`. 
3. `g++` is the GNU C++ compiler, and is generically referred to as `c++`.

You have found that `c++` points to `/usr/bin/x86_64-linux-gnu-g++-11` (The GNU C++ compiler version 11 that was built for x86_64 Linux), which is correct. Your `cc` should point to `/usr/bin/x86_64-linux-gnu-gcc-11`.

We are using C++ and not C, so we only care about what `c++` points to.

`c++` is not a compiler implementation but a symlink to a specific implementation, be it `g++`, `clang++`, [Intel C++ compiler](https://en.wikipedia.org/wiki/Intel_C%2B%2B_Compiler), [AMD Optimizing C++ compiler](https://en.wikipedia.org/wiki/AMD_Optimizing_C/C%2B%2B_Compiler), [EDG C++ compiler](https://www.edg.com/c), IBM XLC++, Oracle C++, NVIDIA HPC C++, and probably countless others. Debian allows the system default (`/usr/bin/c++`) to be changed using `update-alternatives`.

valentin-krasontovitsch:
got it! i know finally also see what you meant in the beginning, with `Ubuntu's GCC is just using the name of the symlink that you called it from.` - when I call `g++ --version`, which at the end points to the same binary as `c++`, it prints `g++` in the version string.

that is indeed funny.

but we could use this, and just check in the version string for whatever command we used to check the version - or is it too brittle to rely on that behaviour?

script/build
```diff
@@ -41,6 +41,21 @@ def check_cxx_version() -> None:
     """
     Check that the user's compiler is compatible with ERT.
     """
+    # we treat silly debian first
```
pinkwah, line 44: Rewrite to fit the language style of the rest of the codebase.

pinkwah:
We could try to discover where `shutil.which("c++")` points to and accept `g++` as a valid alias GCC. This is perhaps cleaner and less edge-casey.
```py
import shutil
from pathlib import Path


path_to_cxx = Path(shutil.which("c++")).resolve()
```

Note that on my system (Fedora) `/usr/bin/c++` is not a symlink but an executable.

valentin-krasontovitsch:
> We could try to discover where shutil.which("c++") points to

and then use that for calling with `--version` and checking in the version string?

> Note that on my system ...

so the snippet you posted will resolve to `/usr/bin/c++` - is that a problem?

I still don't understand though, this seems to make it (just a tad bit more, but still unnecessarily?) complicated, why not just use `c++` and accept that as an alias? it's already the default value for calling the compiler in the code, if the env var `CXX` is not set. is it because the binary is called g++, so that's closer to the truth?
in which sense is this cleaner? which edge cases are you thinking of?

script/build
valentin-krasontovitsch: it is done.

pinkwah:
Because `c++` is a generic name for any compiler. We might not support it (eg. Intel C++ Compiler) or we might not support the version of a supported compiler. We cannot use the generic name.

It's fine that it resolves to `/usr/bin/c++` because it contains `GCC` in the version string. So I reckon the two-step resolve then `--version` check should suffice in all cases.

valentin-krasontovitsch:
I see! OK, thank you.

valentin-krasontovitsch:
just pushed again, i hope this is what you had in mind : )

OBS: I'm using the `GCC` min version if we find `g++`

pinkwah:
Sorry for being nitpicky but:
```py
if "GCC" in version_line or "g++" in version_line:
    ...
```

It's sort of ugly to check for gcc, then clang, then gcc again.

valentin-krasontovitsch:
no no, please, I'm obviously lost in details that I know nothing about here - please be nitpicky, we should name things correctly.

but what you're saying is: merge the first and third if? yes, that's exactly what you're saying : )

gonna fix

pinkwah (approved):
🍊

valentin-krasontovitsch:
github action failed first, with

```
FAILED libres_tests/integration/test_parameter_sample_types.py::test_initialize_random_seed[We initialize twice without setting the random seed either time so expect the result to be different both times]
```

which apparently is flakiness - we did not expect the test to fail so soon. problem was reported to the powers that be, not sure if i should or they will do something about it

PR #3795, opened 2022-08-22
Make unparsed status files transparent to C layer

### LINKED ISSUE #2741: Clean-up status parsing logic in the C-implementation of the queue
While running the forward model the job dispatchers writes two files related to the status of the forward model (`STATUS` and `status.json`).

It was noticed during https://github.com/equinor/ert/pull/2706 that the C-layer of the queue no longer parses any of these files. It uses the presence and time stamp of these files to make conclusions about the state of the forward model, but that is in no way transparent from the implementation.

Hence, the content of the status files are written by the job dispatcher in Python and read only by the Forward Model logic implemented in Python and utilised by the ERT client.

I suggest that we consider making this clearer in the implementation, starting by getting an overview of the remaining usage of the status files in the C-layer. If possible, we could stop letting the C-layer have a notion of these files at all and instead use a `.<some-fitting-name>` file to use for the existence and time step verification that is currently done in the remaining implementation.

**Suggested definition of done**
The C-layer no longer have a notion of the status files of the forward model.

valentin-krasontovitsch:
**Issue**

The files `STATUS` and `status.json` which contain various information
about job status are not actually parsed by the C layer. Only the queue
system checks the files for existence, and checks their timestamp.

Closes #2741

**Approach**

We introduce a new file, `.c-queue-status`, which is used instead by the
C layer.

Simultaneously, on the python side, we touch the file whenever
one of the other files get created / written to, and remove it when the
old files are removed.

## Pre review checklist


pinkwah:
Are we sure this file isn't accessed by users directly? Ie. they have their own scripts that depend on this file existing.

valentin-krasontovitsch:
oh the original `STATUS` and `status.json` file, and their contents, are not changed! this PR just introduces a new file. and makes the C layer read that new file.

does that help?

this really is just a draft, and Sondre wanted to look at this with me, but he don't have time until Wednesday, so since I have you here, Zohar, what do you think in general? do these changes make sense? do they reflect what the original poster of the issue had in mind?

pinkwah:
My knee-jerk reaction is to avoid having new files in a user-readable area (the runpath has become an unofficial public API so we should be weary), but I have no strong opinions. I'd probably want to communicate whatever data this is using EE, but I don't really know what you're doing so I can't say anything for sure.

valentin-krasontovitsch:
> I don't really know what you're doing

have you read the issue (not my wall of text), and the code and commit message, and don't know what i'm doing? if that is case, i'm sorry, and i'm not communicating well enough :thinking: 

EE = eletrical engineering?

valentin-krasontovitsch:
closing after discussion with @sondreso, cf. comment in #2741

PR #3729, opened 2022-08-09
Create an experiment id via uuid

xjules:
**Issue**
Resolves https://github.com/equinor/ert/issues/3720

**Approach**
- [x] Adds an experiment id, via uuid, when adding a new experiment to the registry.
- [ ] Propagated correctly to the job_runner

## Pre review checklist

- [x] Added appropriate release note label
- [x] PR title captures the intent of the changes, and is fitting for release notes.
- [x] Commit history is consistent and clean, in line with the [contribution guidelines](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md).

Adding labels helps the maintainers when writing release notes. This is the [list of release note labels](https://github.com/equinor/ert/labels?q=release-notes).

PR #3811, opened 2022-08-24
Generate experiment_id and propagate to communication channels

### LINKED ISSUE #3720: Create an experiment_id within the Experiment Server and propagate to the Ensemble Evaluator

lars-petter-hauge:
**Issue**
Resolves #3720

**Approach**
Generate experiment id, attach to experiment type objects and propagate id as required.

Continued work from: https://github.com/equinor/ert/pull/3729
Depends on: https://github.com/equinor/ert/pull/3793

Edit:
I initially approached this by changing the `source` content in each of `ensemble`, `realization` etc, but found that adding this information which would not be used in any way (as the `snapshot` model does not take into account the experiment aspect), I decided to revert. Now the experiment id is included in the ensemble and only added to the `jobs.json` file, as the dispatcher side uses protobuf if `experiment_server` is used.

As we include protobuf messaging into more locations (e.g. jobqueue dispatch), the experiment_id will need to be included there as well, given the construct of our protobuf message.

## Pre review checklist

### DIFF
--- a/src/ert/ensemble_evaluator/builder/_legacy.py
+++ b/src/ert/ensemble_evaluator/builder/_legacy.py
@@ -131,17 +131,23 @@ class _LegacyEnsemble(_Ensemble):
         finally:
             get_event_loop().close()
 
-    async def evaluate_async(self, config: "EvaluatorServerConfig") -> None:
+    async def evaluate_async(
+        self,
+        config: "EvaluatorServerConfig",
+        experiment_id: str,
+    ) -> None:
         self._config = config
         await self._evaluate_inner(
             cloudevent_unary_send=self.queue_cloudevent,
             output_bus=self.output_bus,
+            experiment_id=experiment_id,
         )
 
     async def _evaluate_inner(
         self,
         cloudevent_unary_send: Callable[[CloudEvent], Awaitable[None]],
         output_bus: Optional["asyncio.Queue[CloudEvent]"] = None,
+        experiment_id: Optional[str] = None,
     ) -> None:
         """
         This (inner) coroutine does the actual work of evaluating the ensemble. It
@@ -213,13 +219,15 @@ class _LegacyEnsemble(_Ensemble):
             # Tell queue to pass info to the jobs-file
             # NOTE: This touches files on disk...
             sema = threading.BoundedSemaphore(value=CONCURRENT_INTERNALIZATION)
-            if output_bus:  # when running experiment server
+            if (
+                output_bus and experiment_id is not None
+            ):  # when running experiment server
                 self._job_queue.add_dispatch_information_to_jobs_file(
                     ens_id=self.id_,
                     dispatch_url=self._config.dispatch_uri,
                     cert=self._config.cert,
                     token=self._config.token,
-                    experiment_id=self.id_,
+                    experiment_id=experiment_id,
                 )
                 # Finally, run the queue-loop until it finishes or raises
                 await self._job_queue.execute_queue_comms_via_bus(
--- a/src/ert/experiment_server/_registry.py
+++ b/src/ert/experiment_server/_registry.py
@@ -15,7 +15,7 @@ class _Registry:
     def __init__(self) -> None:
         self._experiments: Dict[str, Experiment] = {}
 
-    def add_experiment(self, experiment: Experiment) -> str:
+    def add_experiment(self, experiment: Experiment) -> None:
         """Add an experiment to the registry.
 
         An id is generated here, but it should share (or receive) this ID from
@@ -23,9 +23,7 @@ class _Registry:
 
         [1] https://github.com/equinor/ert/issues/3437#issue-1247962008
         """
-        experiment_id = str(len(self._experiments.keys()))
-        self._experiments[experiment_id] = experiment
-        return experiment_id
+        self._experiments[experiment.id_] = experiment
 
     @property
     def all_experiments(self) -> List[Experiment]:
--- a/src/ert/experiment_server/_server.py
+++ b/src/ert/experiment_server/_server.py
@@ -127,9 +127,10 @@ class ExperimentServer:
         ensemble_size: int,
         current_case_name: str,
         args: Any,
-        factory: Callable[[EnKFMain, int, str, Any], Experiment],
+        factory: Callable[[EnKFMain, int, str, Any, str], Experiment],
+        experiment_id: str,
     ) -> str:
-        """add_legacy_experiment(self, ert, ensemble_size: int,current_case_name: str, factory: Callable[[Any, int, str, Any], Experiment])
+        """add_legacy_experiment(self, ert, ensemble_size: int,current_case_name: str, factory: Callable[[Any, int, str, Any, str], Experiment], experiment_id: str)
 
         The ert parameter, as well as the first input parameter in the factory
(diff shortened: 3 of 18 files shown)


src/ert/shared/cli/main.py
```diff
@@ -48,6 +49,7 @@ def run_cli(args):
         return
 
     evaluator_server_config = EvaluatorServerConfig(custom_port_range=args.port_range)
+    exp_id = str(uuid.uuid1()).split("-", maxsplit=1)[0]
```
lars-petter-hauge, line 52: I don't really like this trimming of uuid, we should reconsider how to assign uuids/what to use (see also https://stackoverflow.com/a/56398787)

lars-petter-hauge: Made it the full `uuid`

src/ert/ert3/evaluator/_builder.py
```diff
@@ -156,6 +156,7 @@ def build_ensemble(  # pylint: disable=too-many-arguments
         .set_max_retries(0)
         .set_executor(driver)
         .set_id(ens_id)
+        .set_exp_id("0")
```
berland, line 159: is this work in progress, or should it always be zero for "ert3"?

src/ert/experiment_server/_registry.py
```diff
@@ -15,17 +15,15 @@ class _Registry:
     def __init__(self) -> None:
         self._experiments: Dict[str, Experiment] = {}
 
-    def add_experiment(self, experiment: Experiment) -> str:
+    def add_experiment(self, experiment: Experiment) -> None:
         """Add an experiment to the registry.
 
         An id is generated here, but it should share (or receive) this ID from
```
berland, line 21: This comment is now outdated.

src/ert/shared/cli/model_factory.py
```diff
@@ -8,21 +8,23 @@
 from ert.shared.models.single_test_run import SingleTestRun
 
 
-def create_model(ert, ensemble_size, current_case_name, args):
+def create_model(ert, ensemble_size, current_case_name, args, id_):
     # Setup model
     if args.mode == "test_run":
         model = _setup_single_test_run(ert)
```
berland, line 14: Is it intentional to not associate test-runs with the experiment server?

berland: I see, it is defaulted to "0" in the    `SingleTestRun.__init__`

berland: This leads to a crash when rebased on top of `experiment_api`, and then tested with `test_run` and with experiment-server applied. Seemingly resolved by passing the exp_id to `_setup_single_test_run`.

src/ert/ert3/evaluator/_builder.py
lars-petter-hauge, line 159: set to "experiment_id" instead?

lars-petter-hauge: Had not planned for ert3 yet to be honest. Removed either way, will not be part of `EnsembleBuilder`

lars-petter-hauge: fixed

lars-petter-hauge:
The jenkins test failing is:

```
def test_initialize_random_seed(tmpdir, caplog, check_random_seed, expectation):
...
E                   Failed: DID NOT RAISE <class 'AssertionError'>
```

src/ert/shared/cli/model_factory.py
lars-petter-hauge: Yeh, getting a `KeyError` when asking the `experiment_server` to run experiment with given id, as it ignores the id created in the `async_run`. I've made it propagate, not sure if I like how it's done though, as the `SingleTestRun` doesn't include the `queue_config` - hence instantiating the object suddenly differs from the rest.. (the `id` is the third argument instead of the fourth...).

Thanks for picking up on it though!

src/ert/experiment_server/_server.py
```diff
@@ -141,12 +142,13 @@ def add_legacy_experiment(  # pylint: disable=too-many-arguments
             ensemble_size,
             current_case_name,
             args,
+            experiment_id,
```
xjules, line 145: Why does it need to be set  and not within the constructor (or later on via the builder as we discussed) ?

lars-petter-hauge: I'm not sure if I have strong preferences on this, so we can change it. But what we have today is builder for `job`, `step`, `stage`, `realization` and `ensemble`, but _not_ `experiment`.

We did discuss adding a builder for the experiment as well, with the counter argument being that we don't know up front what the ensemble will look like (successive ensembles are affected by the previous ones, with deactivated realizations). While when we create ensemble, we know for sure how the entire hierarchy will look like. Once a builder builds, the returned object should be completely populated, which we can't do for experiments as we only know what the first ensemble looks like. At the end the suggestions was to add an `EnsembleBuilder` instance to the `Experiment` object, with the at the time known information, which will be used to create ensembles. I believe we agreed to wait with this though.

For all the classes down the hierarchy today, they all accept an `id` in their `init`, so I suppose I did it this way to align with the rest. The builders all have a `set_id`, to be used when populating the instances themselves. Some builders accept that `id` is not set, and will then create them when trying to build. None of the classes create the `id` in their constructor.

In the end we have also discussed the point that the `id` could be the same as the one used in `storage`, but this is also implementing for the future, so not really a solid argument.

src/ert/ensemble_evaluator/builder/_legacy.py
```diff
@@ -219,7 +225,7 @@ async def _evaluate_inner(
                     dispatch_url=self._config.dispatch_uri,
                     cert=self._config.cert,
                     token=self._config.token,
-                    experiment_id=self.id_,
+                    experiment_id=experiment_id,
```
xjules, line 228: Shall we then require that also `experiment_id is not None`?:
```python
if output_bus and experiment_id:  # when running experiment server
        self._job_queue.add_dispatch_information_to_jobs_file(
            ens_id=self.id_,
            dispatch_url=self._config.dispatch_uri,
            cert=self._config.cert,
            token=self._config.token,
            experiment_id=experiment_id,
        )
```

tests/ert_tests/cli/test_model_factory.py
```diff
@@ -66,7 +66,7 @@ def test_init_iteration_number(self):
             ert = work_area.getErt()
             args = Namespace(iter_num=10, realizations=None)
             model = model_factory._setup_ensemble_experiment(
-                ert, args, ert.getEnsembleSize()
+                ert, args, ert.getEnsembleSize(), "0"
```
xjules, line 69: As it might be a bit unclear what about using "experiment_id" or "exp_id" instead of "0"?

lars-petter-hauge: Sure, fixed!

src/ert/ensemble_evaluator/builder/_legacy.py
lars-petter-hauge: Yepp, can do! and Done!

xjules (approved):
Very good job @lars-petter-hauge ! Just rebase and good to :rocket:

PR #3837, opened 2022-08-30
Increase readability of rng-config test

### LINKED ISSUE #2556: The RNGConfig test incorrectly documents config_dict
`tests/libres_tests/res/enkf/test_rng_config.py` uses a structure for config dicts which is not correct, but does not fail due to fudging.

valentin-krasontovitsch:
**Issue**
Closes #2556 

**Approach**
We introduce a new variable representing the dict passed to rng config to increase clarity of the test in question.

## Pre review checklist

### DIFF
--- a/tests/libres_tests/res/enkf/test_rng_config.py
+++ b/tests/libres_tests/res/enkf/test_rng_config.py
@@ -38,7 +38,7 @@ class RNGConfigTest(ResTest):
             },
         }
 
-    def test_dict_constructor(self):
+    def test_compare_config_content_and_dict_constructor(self):
         config = self.create_base_config()
 
         case_directory = self.createTestPath("local/simple_config")
@@ -46,11 +46,19 @@ class RNGConfigTest(ResTest):
             work_area.copy_directory(case_directory)
 
             random_seed = "abcdefghijklmnop"
-            config["SIMULATION"]["SEED"] = {ConfigKeys.RANDOM_SEED: random_seed}
+            rng_config_dict = {ConfigKeys.RANDOM_SEED: random_seed}
+            config["SIMULATION"]["SEED"] = rng_config_dict
 
-            rng_config = RNGConfig(config_dict=config["SIMULATION"]["SEED"])
-            res_config = ResConfig(config=config)
-            self.assertEqual(rng_config, res_config.rng_config)
+            # we instantiate an `RNGConfig` class using a config dict
+            rng_config_from_dict = RNGConfig(config_dict=rng_config_dict)
+            # NB: We are using the `config` argument, which essentially
+            # passes a config content directly to the init function of the
+            # `ResConfig` class
+            res_config_from_config_content = ResConfig(config=config)
+
+            self.assertEqual(
+                rng_config_from_dict, res_config_from_config_content.rng_config
+            )
 
     def test_random_seed(self):
         config = self.create_base_config()


tests/libres_tests/res/enkf/test_rng_config.py
```diff
@@ -46,11 +46,13 @@ def test_dict_constructor(self):
             work_area.copy_directory(case_directory)
 
             random_seed = "abcdefghijklmnop"
-            config["SIMULATION"]["SEED"] = {ConfigKeys.RANDOM_SEED: random_seed}
+            simple_config = {ConfigKeys.RANDOM_SEED: random_seed}
+            config["SIMULATION"]["SEED"] = simple_config
 
-            rng_config = RNGConfig(config_dict=config["SIMULATION"]["SEED"])
+            rng_config_from_dict = RNGConfig(config_dict=simple_config)
             res_config = ResConfig(config=config)
```
eivindjahren, line 53: I seem to be permanently confused by this. There still seems to be fudging involved, because ResConfig does not unpack `config["SIMULATION"]["SEED"]` before passing it to `RNGConfig`:
https://github.com/equinor/ert/blob/84b3351bb364be8781963cdddb18db1ba3e326dc/src/ert/_c_wrappers/enkf/res_config.py#L256

valentin-krasontovitsch: so there's two different things being considered here, I think - the test creates an `RNG` config from a config dict using the `RNCConfig` class directly.

in `ert/src/ert/_c_wrappers/enkf/res_config.py`, we are looking at the `ResConfig` class, and in particular at its `_alloc_from_dict` method.

so the test currently compares the RNG part of
- a res config created from `ResConfig`'s `__init__` with the `config` argument given, which in turn calls `_create_from_content`
- an rng config created from `RNGConfig`'s `__init__` with the `config_dict` argument given

I suppose what one would like to compare is ... I actually don't know XD

using the `ResConfig` class in both instances, once with `config`, once with `config_dict`? or using the RNGConfig in both instances?

do we just want to say 'hey, we know this part of the code is wrong, let's write a regression test to capture it', or do we want a test with some other objective, that also happens to capture this fudginess?

valentin-krasontovitsch: > The test could be

```
            config = {ConfigKeys.RANDOM_SEED: random_seed}
            res_config = ResConfig(config=config)
```
isn't that against the intention? `ResConfig` should get a more complicated structure, no? or do we not want it to expect to find the rng config under `SIMULATION`?

eivindjahren: Ah now I see, we are using `ResConfig(config=config)` and not `ResConfig(config_dict=...)`. This stuff is so confusing. Could you add some small code comment to this?

eivindjahren: There really should not be three ways of instantiating `ResConfig` but such is life.

valentin-krasontovitsch: > Could you add some small code comment to this?

in which spot? in the test? what should the comment say?

and I'm at least still lost with respect to the questions I posed - what kind of test(s) do we want here?

> three ways of instantiating ResConfig

You mean from config file, from config content, and from config dict? seems like a good array of options : ) I'm sure we'll be able to refactor it into something sensible and reasonable during this milestone!

eivindjahren: I think just a comment in the tests for now. Eventually, only the config_dict path should be taken, but in a nicer way.

eivindjahren (approved):

Issue #3708, opened 2022-08-05
Include 'es-mda' method in ert/docs/advanced folder examples


ajikmr:
Hello.
Examples in the folder  'ert/docs/advanced' show 'ensemble_smoother' and 'iterative_smoother' methods. It will be good to add 'es-mda' to the same for clarity and comparison purpose.

Also, importing 'ies' does not work. Instead '_ies' works.

Thanks.

dafeda:
Solved by creating ES-MDA example in `iterative_ensemble_smoother`-package: https://github.com/equinor/iterative_ensemble_smoother/tree/main/docs/source

PR #3844, opened 2022-08-31
Move out the iterative ensemble smoother

eivindjahren:
Adresses problem raised in #3708 where `_ies` is made a private module in ert. The public code is now part of the `iterative_ensemble_smoother` package. 

Currently, this is waiting for version 0.0.3 of iterative_ensemble_smoother.


eivindjahren:
Testing on jenkins is currently failing because `iterative_ensemble_smoother` is not in bleeding. It will be as of midnight.

dafeda:
A few comments (as discussed on Slack)

Ensemble Smoother is identical to Iterative Ensemble Smoother with step-length of 1.0, which means that `IterativeEnsembleSmoother.update_step(step_length=1)` is sufficient.

Row scaling or any type of localization should not be part of the package.

Personally, I prefer using variablen names as defined in the paper instead of coming up new names. For example `S` instead of `response_matrix`.

I would have preferred the tests to already have been moved before we start using this in production.

A link to the paper that defines the algorithm should be included in the readme.

eivindjahren:
> A few comments (as discussed on Slack)
> 
> Ensemble Smoother is identical to Iterative Ensemble Smoother with step-length of 1.0, which means that `IterativeEnsembleSmoother.update_step(step_length=1)` is sufficient.
> 
> Row scaling or any type of localization should not be part of the package.
> 
> Personally, I prefer using variable names as defined in the paper instead of coming up new names. For example `S` instead of `response_matrix`. In this specific case, I would not know what to name `S`. Perhaps sensitivity matrix?
> 
> I would have preferred the tests to already have been moved before we start using this in production. Mostly because this kind of work tends to get delayed...
> 
> A link to the paper that defines the algorithm should be included in the readme.

I have renamed the variable to `sensitivity_matrix` as suggested in https://github.com/equinor/iterative_ensemble_smoother/pull/22. Also renamed `parameter_matrix` to `centered_anomaly_matrix` as requested. The reason for these names is that they are part of the public API so you get problems of name collisions and the like. The name is still S everywhere, and discussed in the documentation.

As for row_scaling, it is described as an experimental feature of `iterative_ensemble_smoother`, which is a compromise that reduces complexity a lot. It means we can reduce the API surface to just python.

dafeda:
Come to think of it, I think `response_matrix` is the correct term, but that it should be `Y` instead of `S`. Would have to check the paper to be certain though.

eivindjahren:
We have now added back the c-tests as python tests. As these test the code being run, rather than a different c code-path, I consider this an improvement.

The tests should pass as soon as we get the 0.0.3 release out. I believe we are then ready for review.

eivindjahren:
https://github.com/equinor/iterative_ensemble_smoother/issues/26

eivindjahren:
https://github.com/equinor/iterative_ensemble_smoother/issues/27

setup.py
```diff
@@ -135,6 +135,7 @@ def package_files(directory):
         "ert-storage >= 0.3.16",
         "fastapi",
         "graphlib_backport; python_version < '3.9'",
+        "iterative_ensemble_smoother>=0.0.2",
```
oysteoh, line 138: 0.0.3?

eivindjahren: Yes

eivindjahren: Done

dafeda:
> equinor/iterative_ensemble_smoother#27

Should we this before merging or do you want to wait?

eivindjahren:
jenkins, test this please

oysteoh:
Jenkins: Test this please!

eivindjahren:
> > [equinor/iterative_ensemble_smoother#27](https://github.com/equinor/iterative_ensemble_smoother/issues/27)
> 
> Should we this before merging or do you want to wait?

I think it would be good to integrate first and simplify after. Simplifying now might confound potential problems with the existing solution.

dafeda (approved):
Great work 👍

PR #3846, opened 2022-09-01
Test update with INIT_FILES and FORWARD_INIT

dafeda:
## Pre review checklist


test-data/local/externally_sampled_params/config.ert
```diff
@@ -0,0 +1,17 @@
...
+QUEUE_OPTION LOCAL MAX_RUNNING 50
+
+RUNPATH out/realization-%d/iter-%d
+
+OBS_CONFIG observations
+TIME_MAP time_map
+
+NUM_REALIZATIONS 10
+
+GEN_KW COEFFS coeff.tmpl params.json coeff_priors FORWARD_INIT:True INIT_FILES:externally_sampled_params0
```
joakim-hove, line 13: I *think* the `INIT_FILES:/path/to/init/files%d` needs to contain a `%d` placeholder in order to ensure that the different realizations pick up different init files.

dafeda: My reading of [the docs](https://ert.readthedocs.io/en/latest/reference/configuration/data_types.html?highlight=INIT_FILES#initialization-with-forward-init) is that "%d" is not needed when `FORWARD_INIT=True` which means that the forward model is what does the sampling.
Is this perhaps wrong?

joakim-hove: I am sorry for the misleading comment - you had it right. 

My only remaining comment would be to recommend removing the literal `0` from the INIT_FILES name - that fooled me into thinking it represented a (fixed) realization number.

dafeda: Thanks, I've pushed a fix.

oyvindeide (approved):

PR #3849, opened 2022-09-01
Close stdin/stdout/stderr files when done

berland:
Opted for not using a context manager (ExitStack)
to avoid having to indent all lines in the function

Plus bonus pylinting

**Issue**
Resolves unclosed sdin/stdout/stdin files opened by the `Job` object.

**Approach**
Close at the end.

## Pre review checklist

### DIFF
--- a/src/ert/job_runner/job.py
+++ b/src/ert/job_runner/job.py
@@ -22,6 +22,7 @@ class Job:
         self.std_out = job_data.get("stdout")
 
     def run(self):
+        # pylint: disable=consider-using-with
         start_message = Start(self)
 
         errors = self._check_job_files()
@@ -44,7 +45,7 @@ class Job:
             arg_list += self.job_data["argList"]
 
         if self.job_data.get("stdin"):
-            stdin = open(self.job_data.get("stdin"))
+            stdin = open(self.job_data.get("stdin"), encoding="utf-8")
         else:
             stdin = None
 
@@ -53,7 +54,7 @@ class Job:
                 os.path.dirname(os.path.abspath(self.std_err)),
                 exist_ok=True,
             )
-            stderr = open(self.std_err, "w")
+            stderr = open(self.std_err, "w", encoding="utf-8")
         else:
             stderr = None
 
@@ -62,7 +63,7 @@ class Job:
                 os.path.dirname(os.path.abspath(self.std_out)),
                 exist_ok=True,
             )
-            stdout = open(self.std_out, "w")
+            stdout = open(self.std_out, "w", encoding="utf-8")
         else:
             stdout = None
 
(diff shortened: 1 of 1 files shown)


hnformentin:
Looks good!
(1) Do you think it would make sense to have a test for this?
(2) There is a linting opportunity "Use a single if-statement instead of nested if-statements"  in line 155. Would that make sense to address in this issue?

berland:
> Looks good! (1) Do you think it would make sense to have a test for this? (2) There is a linting opportunity "Use a single if-statement instead of nested if-statements" in line 155. Would that make sense to address in this issue?

(1): I tried writing a test for this, using `psutil.Process.open_files()` but it turns out that I cannot make this test fail. The reason for  this is that the stdout/stderr handles in the `Job.run()` code goes out of scope and the filehandles are closed by the garbage collector in CPython. So in this case, it is not a problem that the files are not closed, it is more of a recommendation to always close the files.

(2): Fixed!

src/ert/job_runner/job.py
```diff
@@ -22,6 +22,7 @@ def __init__(self, job_data, index, sleep_interval=1):
         self.std_out = job_data.get("stdout")
 
     def run(self):
+        # pylint: disable=consider-using-with
```
hnformentin, line 25: What do you think about commenting `# stdin, stout and stderr are closed on the bottom of the function`?

hnformentin: I just got this idea from your other PR, but it is a minor thing not needed if you prefer so.

hnformentin:
> > Looks good! (1) Do you think it would make sense to have a test for this? (2) There is a linting opportunity "Use a single if-statement instead of nested if-statements" in line 155. Would that make sense to address in this issue?
> 
> (1): I tried writing a test for this, using `psutil.Process.open_files()` but it turns out that I cannot make this test fail. The reason for this is that the stdout/stderr handles in the `Job.run()` code goes out of scope and the filehandles are closed by the garbage collector in CPython. So in this case, it is not a problem that the files are not closed, it is more of a recommendation to always close the files.
> 
> (2): Fixed!

Nice! It was a good opportunity for me to learn about garbage collector in Python. Thanks for the detailed answer.

hnformentin (changes requested):
LGTM except for the test failing. Can you please take a look in it?

berland:
jenkins test this please

berland:
The failing test looked like Jenkins flakyness, seems ok now.

hnformentin:
Python coverage / Python Coverage (ert) (pull_request) Cancelled after 30m.
Will restart tests.

jenkins test this please

hnformentin:
jenkins test this please

hnformentin (approved):

berland:
Changing release note label to maintenance since it is not really a bug in CPython as of now.

berland:
Oh no, forgot to squash!

PR #3864, opened 2022-09-06
Test size of covariance instead of variance

dafeda:
It is a mistake to expect that the variance of each
parameter will be reduced by the update step.

## Pre review checklist

### DIFF
--- a/tests/libres_tests/res/enkf/test_es_update.py
+++ b/tests/libres_tests/res/enkf/test_es_update.py
@@ -2,6 +2,7 @@ import sys
 from argparse import ArgumentParser
 from pathlib import Path
 
+import numpy as np
 import pytest
 from iterative_ensemble_smoother import IterativeEnsembleSmoother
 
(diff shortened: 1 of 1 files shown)


tests/libres_tests/res/enkf/test_es_update.py
```diff
@@ -205,9 +206,13 @@ def test_that_posterior_has_lower_variance_than_prior_with_init_files(copy_case)
     df_default = GenKwCollector.loadAllGenKwData(ert, "default")
     df_target = GenKwCollector.loadAllGenKwData(ert, "target")
 
-    assert df_default["COEFFS:COEFF_A"].var() > df_target["COEFFS:COEFF_A"].var()
-    assert df_default["COEFFS:COEFF_B"].var() > df_target["COEFFS:COEFF_B"].var()
-    assert df_default["COEFFS:COEFF_C"].var() > df_target["COEFFS:COEFF_C"].var()
+    # We expect that ERT's update step lowers the "size" of the covariance
+    # of the parameters, where "size" can be measured by the determinant.
```
Blunde1, line 210: How about using the concept of "generalized variance" here? https://en.wikipedia.org/wiki/Generalized_variance
E.g.
"We expect that ERT's update step lowers the generalized variance for the parameters"

tests/libres_tests/res/enkf/test_es_update.py
```diff
@@ -205,9 +206,13 @@ def test_that_posterior_has_lower_variance_than_prior_with_init_files(copy_case)
...
+    assert (
+        0
+        < np.linalg.det(df_target.cov().to_numpy())
+        < np.linalg.det(df_default.cov().to_numpy())
+    )
```
Blunde1, line 215: 👍 
This definitely holds for the population covariances that the ensemble/sample is drawn from. Because `0<det(I-KH)<1`. So it also definitely holds when ensemble size tends to infinity.
It should also definitely hold for finite ensemble sizes when `df_target` and `df_default` are the same sample, i.e. when `df_target` is only an update of the sample `df_default`, and not an entirely new sample from the population posterior.

dafeda: Thanks!

tests/libres_tests/res/enkf/test_es_update.py
dafeda: Much better, thanks.

Blunde1 (approved):
LGTM, cool!

PR #3887, opened 2022-09-13
Consolidate config file and dict paths for substitution config

### LINKED ISSUE #3802: SubstConfig has different default when passed with config_dict and filname
```
>>> from ert._c_wrappers.enkf import SubstConfig
>>> filename = "..." # name of config file
>>> config_dict = {...} # Config dictionary with same values as the config file
>>> SubstConfig(filename)
SubstConfig([(<DATE>, 2022-08-23, The current date.), (<NUM_CPU>, 1, Supplied by the user in the configuration file.) ...
>>> SubstConfig(config_dict=config_dict)
SubstConfig([(<DATE>, 2022-08-23, The current date.), (<NUM_CPU>, 0, The number of CPU used for one forward model.) ...
``

### PR DESCRIPTION
**Pull request by valentin-krasontovitsch**
**Issue**
Resolves #3802

**Approach**
Small changes here and there to make the acceptance test in question pass.

# I recommend reading commit by commit, it will make the review easierer (I suppose)

## Pre review checklist

### DIFF
--- a/src/ert/_c_wrappers/enkf/subst_config.py
+++ b/src/ert/_c_wrappers/enkf/subst_config.py
@@ -93,8 +93,14 @@ class SubstConfig(BaseCClass):
                 "The name of a file with a list of run directories.",
             )
 
+            if ConfigKeys.NUM_CPU in config_dict:
+                subst_list.addItem(
+                    "<NUM_CPU>",
+                    str(config_dict[ConfigKeys.NUM_CPU]),
+                    "The number of CPU used for one forward model.",
+                )
             # Read num_cpu from Eclipse DATA_FILE
-            if ConfigKeys.DATA_FILE in config_dict:
+            elif ConfigKeys.DATA_FILE in config_dict:
                 file_path = os.path.join(
                     config_directory, config_dict[ConfigKeys.DATA_FILE]
                 )
--- a/tests/libres_tests/res/enkf/test_res_config.py
+++ b/tests/libres_tests/res/enkf/test_res_config.py
@@ -337,7 +337,9 @@ def test_extensive_config(setup_case):
 
 
 def test_res_config_dict_constructor(setup_case):
-    _ = setup_case("local/snake_oil_structure", "ert/model/user_config.ert")
+    config_file_name = "user_config.ert"
+    relative_config_path = f"ert/model/{config_file_name}"
+    _ = setup_case("local/snake_oil_structure", relative_config_path)
     # create script file
     script_file = "script.sh"
     with open(script_file, "w") as f:
@@ -346,12 +348,8 @@ def test_res_config_dict_constructor(setup_case):
     st = os.stat(script_file)
     os.chmod(script_file, stat.S_IEXEC | st.st_mode)
 
-    # add a missing entries to config file
-    with open("user_config.ert", "a+") as ert_file:
-        ert_file.write("JOB_SCRIPT ../../../script.sh\n")
-
     # split config_file to path and filename
-    cfg_path, cfg_file = os.path.split(os.path.realpath("ert/model/user_config.ert"))
+    absolute_config_dir, _ = os.path.split(os.path.realpath(relative_config_path))
 
     config_data_new = {
         ConfigKeys.ALPHA_KEY: 3,
@@ -366,7 +364,7 @@ def test_res_config_dict_constructor(setup_case):
         ConfigKeys.RUNPATH: "<SCRATCH>/<USER>/<CASE_DIR>/realization-%d/iter-%d",
         ConfigKeys.NUM_REALIZATIONS: 10,  # model
         ConfigKeys.MAX_RUNTIME: 23400,
-        ConfigKeys.JOB_SCRIPT: "../../../script.sh",
+        ConfigKeys.JOB_SCRIPT: f"../../{script_file}",
         ConfigKeys.QUEUE_SYSTEM: QueueDriverEnum.LSF_DRIVER,
         ConfigKeys.USER_MODE: True,
         ConfigKeys.MAX_SUBMIT: 13,
@@ -501,11 +499,16 @@ def test_res_config_dict_constructor(setup_case):
                     config_data_new[ConfigKeys.DEFINE_KEY].get(define_key),
                 )
 
-    # change dir to actual location of cfg_file
-    os.chdir(cfg_path)
+    # change dir to actual location of config file
+    os.chdir(absolute_config_dir)
+
+    # add missing entries to config file
+    with open(config_file_name, "a+") as ert_file:
+        ert_file.write(f"JOB_SCRIPT ../../{script_file}\n")
+        ert_file.write("NUM_CPU 0\n")
 
     # load res_file
-    res_config_file = ResConfig(user_config_file=cfg_file)
+    res_config_file = ResConfig(user_config_file=config_file_name)
 
     # get site_config location
     ert_share_path = os.path.dirname(res_config_file.site_config.getLocation())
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Review by eivindjahren — CHANGES_REQUESTED** [review]
Looks good! Just need to finish making the tests pass and rebase the "Fix formatting" commit :)

File: src/clib/lib/enkf/queue_config.cpp [hunk 1]
```diff
@@ -55,7 +55,7 @@ static queue_config_type *queue_config_alloc_empty() {
     queue_config->driver_type = NULL_DRIVER;
     queue_config->user_mode = false;
     queue_config->max_submit = 2; // Default value
-    queue_config->num_cpu = 0;
```
eivindjahren, line 58: I think this change would be the correct if this code was following normal conventions. However, I think something non-standard is going on. num_cpu=0 is clearly not correct, but perhaps this is signalling some sort of "don't force the number of cpus" behavior. I think we will have to dig a bit to see what is going on before we make a change here.

eivindjahren:
I believe we can merge everything except bb97391f222a90ed019136cddc80ddff0f9335c5 which we will have to investigate further before we do anything.

src/clib/lib/enkf/queue_config.cpp
valentin-krasontovitsch: i'm gonna try to investigate a bit then! thanks for pointing it out.

valentin-krasontovitsch: found the first hint in `src/ert/ensemble_evaluator/builder/_ensemble_builder.py`, in the function `from_legacy`, lines 124 ff:

```
    num_cpu = res_config.queue_config.num_cpu
    if num_cpu == 0:
        num_cpu = res_config.ecl_config.num_cpu
```

valentin-krasontovitsch: and we have the same pattern in `src/ert/_c_wrappers/job_queue/queue.py`, in `add_job_from_run_arg`:

```
    num_cpu = res_config.queue_config.num_cpu
    if num_cpu == 0:
        num_cpu = res_config.ecl_config.num_cpu
```

what do you think about the solution I proposed above, @eivindjahren ?

valentin-krasontovitsch: that seem to be all instances of usages of `queue_config`'s `num_cpu`

eivindjahren: It seems like this code is signaling that 0 means use `res_config.ecl_config.num_cpu`. I think we should clean that up in a separate PR. Isn't it possible to just recreate how this number is instantiated in this PR and then clean this particular behavior up separately?

valentin-krasontovitsch: i'll see if i can get away in this PR without those changes. i think something made me introduce them, but perhaps it was just confusion ; )

valentin-krasontovitsch: I have removed that commit, and rather pushed it with changes as discussed here in #3919

valentin-krasontovitsch: oh and i fixed the test that was failing and triggered me to do the changes in question to work without those changes - basically just needed to make sure that the same value for `NUM_CPU` was present both in config file and config dict (the former was not the case)

valentin-krasontovitsch:
@eivindjahren i think i have resolved the problematic commit (removed it and fixed the failing test), and all checks have passed (yeahy), so I would appreciate a hopefully final review - you may focus on the last four commits, if you wish, which can also be rebased to make things more compact

eivindjahren (approved):
squash  ef7b95ebf54a39e230c422bf23d45e03a6881842 and ef7b95ebf54a39e230c422bf23d45e03a6881842 and I think we are ready to go!

eivindjahren (changes requested):
I believe your refactoring of the test might be doing something incorrectly. Other than that squash  ef7b95ebf54a39e230c422bf23d45e03a6881842 and ef7b95ebf54a39e230c422bf23d45e03a6881842 and I think we are ready to go!

tests/libres_tests/res/enkf/test_res_config.py
```diff
@@ -505,7 +505,6 @@ def test_res_config_dict_constructor(setup_case):
 
     # add missing entries to config file
     with open(config_file_name, "a+") as ert_file:
-        ert_file.write("JOB_SCRIPT ../../../script.sh\n")
```
eivindjahren: I believe what is going on here is that your refactoring of the test left cwd in a different place once you gave it to `ResConfig` so that `../../../script.sh` is no longer pointing at the scrip.

valentin-krasontovitsch: hmm it doesn't seem like the refactoring changed the path for the chdir call to me, but of course i may be wrong. i'm gonna investigate!

eivindjahren:
So I am thinking we should have "Write whole config dict for tests" and "Consolidate config file and dict paths for substitution config" in the same commit and we drop ["Refactoring in test"](https://github.com/equinor/ert/pull/3887/commits/0b7dccfd2d4b0783516312bd7fc43bb233ad1c96) and 
"Remove breaking line in test". Then we can merge.

src/ert/_c_wrappers/enkf/res_config.py
```diff
@@ -685,3 +685,23 @@ def __eq__(self, other):
...
+        return (
+            "{\n"
+            f"SubstConfig: {self.subst_config},\n"
+            f"SiteConfig: {self.site_config},\n"
+            f"RNGConfig: {self.rng_config},\n"
+            f"AnalysisConfig: {self.analysis_config},\n"
+            f"ErtWorkflowList: {self.ert_workflow_list},\n"
+            f"HookManager: {self.hook_manager},\n"
+            f"ErtTemplates: {self.ert_templates},\n"
+            f"ECLConfig: {self.ecl_config},\n"
```
eivindjahren, line 702: `ECLConfig` is misspelled here. Also `RNGConfig` is gone on master so your branch is too far behind.

valentin-krasontovitsch: fixed

src/ert/_c_wrappers/enkf/subst_config.py
```diff
@@ -93,8 +93,14 @@ def __init__(self, config_content=None, config_dict=None):
                 "The name of a file with a list of run directories.",
             )
 
+            if ConfigKeys.NUM_CPU in config_dict:
+                subst_list.addItem(
```
eivindjahren, line 97: I think this will add the item twice when `os.path.isfile(file_path) and os.access(file_path, os.R_OK)` (line 108). That might become problematic down the line.

tests/libres_tests/res/enkf/test_res_config.py
```diff
@@ -349,6 +349,7 @@ def test_res_config_dict_constructor(setup_case):
     # add a missing entries to config file
     with open("user_config.ert", "a+") as ert_file:
         ert_file.write("JOB_SCRIPT ../../../script.sh\n")
+        ert_file.write("NUM_CPU 0\n")
```
eivindjahren: This entry is optional. Why do we need to add it?

valentin-krasontovitsch: because in the config dict, it's set to zero, and the config file path sets that value to 1 if it's not given. so it's to make the test pass.

src/ert/_c_wrappers/enkf/subst_config.py
valentin-krasontovitsch: i think the logic is

- if the NUM_CPU is given in the config dict, use that
- if it's not given in the config dict, try to find it in the DATA FILE

so i'm not sure i follow how it will be added twice?

tests/libres_tests/res/enkf/test_res_config.py
eivindjahren: Don't we want the config_dict path to also set it to 1 then?

valentin-krasontovitsch:
OK ready for review of "delete the umask" - there's two commits, the first sets a default umask in the job runner, which actually uses a umask. I chose (18)_10 = (022)_8 (base 10 and base 8 notation), based on some comments / that value seemingly being a sort of default.

BUT we could

- remove that settings of the umask altogether (consequences? why is it necessary? security concerns? should we just let it be whatever it is on the given system?)
- use a different default value

i welcome any sort of opinion / arguments on the matter.

eivindjahren (approved):
Looks good!

PR #3890, opened 2022-09-14
Move storage_service.json to enspath  and propagate the ensepath when starting the webviz_ert service

### LINKED ISSUE #3873: Move `storage_server.json` and `.webviz` to enspath
**Describe the bug**
There seems to be a severe bug (or a breaking change?) related to ERT plotting, that occurs when opening two ERT instances. The second instance that is opened seems to read data from the storage of the first instance. This happens despite the two cases having seperate/unique ENSPATH values.Example of storage/ENSPATH structure in the case I tested, (the ENSPATH's have a common level (output/storage/, with a unique foldername at end)
storage
├── 01_drogon_ahm
├── 01_drogon_ahm_noseis
├── 01_drogon_ahm_seis
├── 01_drogon_design
├── drogon_exercise_32
└── reports 

This has worked fine before the Aug release of komodo (2022.08.02-py38) (edited) 

**Expected behaviour**
We should count for a setup as described with the use of enspath's. The example with drogon visualize this perfectly. There is one folder with a lot of config files. Running `ert gui 01_drogon_ahm.ert` here will create one `storage_server.json` file in this folder. If you at the same time wants to visualize `01_drogon_ahm_noseis` and runs `ert gui 01_drogon_ahm_noseis.ert` it opens the gui but connects to the `api` already running and by that also visualize from this which of course is very wrong. A proposed solution would be to place the `storage_server.json` ( and also `.webviz` i think? )  in the folder `enspath` specifies. If any other ideas of how to solve this occurs feel free to discuss it !

DanSava:
**Issue**
Resolves #3873 

**Approach**

## Pre review checklist

### DIFF
--- a/src/ert/shared/services/_base_service.py
+++ b/src/ert/shared/services/_base_service.py
@@ -33,12 +33,12 @@ T = TypeVar("T", bound="BaseService")
 ConnInfo = Union[Mapping[str, Any], Exception, None]
 
 
-SERVICE_NAMES: Set[str] = set()
+SERVICE_CONF_PATHS: Set[str] = set()
 
 
 def cleanup_service_files(signum: int, frame: Optional[FrameType]) -> None:
-    for service_name in SERVICE_NAMES:
-        file = Path(f"{service_name}_server.json")
+    for file_path in SERVICE_CONF_PATHS:
+        file = Path(file_path)
         if file.exists():
             file.unlink()
     raise OSError(f"Signal {signum} received.")
@@ -94,6 +94,7 @@ class _Proc(threading.Thread):
         exec_args: Sequence[str],
         timeout: int,
         set_conn_info: Callable[[ConnInfo], None],
+        project: Path,
     ):
         super().__init__()
 
@@ -103,6 +104,7 @@ class _Proc(threading.Thread):
         self._exec_args = exec_args
         self._timeout = timeout
         self._set_conn_info = set_conn_info
+        self._service_config_path = Path(project) / f"{self._service_name}_server.json"
 
         self._assert_server_not_running()
 
@@ -112,7 +114,7 @@ class _Proc(threading.Thread):
         env = os.environ.copy()
         env["ERT_COMM_FD"] = str(fd_write)
 
-        SERVICE_NAMES.add(self._service_name)
+        SERVICE_CONF_PATHS.add(str(self._service_config_path))
 
         self._childproc = Popen(
             self._exec_args,
@@ -162,17 +164,17 @@ class _Proc(threading.Thread):
         we have waited long enough before sys.exit
         """
         for i in range(3):
-            if (Path.cwd() / f"{self._service_name}_server.json").exists():
+            if self._service_config_path.exists():
                 print(
                     f"{self._service_name}_server.json is "
                     f"present on this location. Retry {i}"
                 )
                 time.sleep(1)
 
-        if (Path.cwd() / f"{self._service_name}_server.json").exists():
+        if self._service_config_path.exists():
             print(
                 f"A file called {self._service_name}_server.json is present from this "
-                "location. This indicates there is already a ert instance running. "
+                f"location. This indicates there is already a ert instance running. "
                 "If you are certain that is not the case, try to delete the file "
                 "and try again."
             )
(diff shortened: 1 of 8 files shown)


DanSava:
Jenkins test this, please!

src/ert/shared/services/storage_service.py
```diff
@@ -41,6 +43,16 @@ def fetch_auth(self) -> Tuple[str, Any]:
...
+    # @classmethod
+    # def start_service(cls, *args: Any, **kwargs: Any):
+    #     try:
+    #         service = cls.connect(timeout=0, project=kwargs.get("project"))
+    #         # Attempt to fetch the service url
+    #         _ = service.fetch_url()
+    #     except TimeoutError:
+    #         return cls.start_server(*args, **kwargs)
+    #     return _Context(service)
+
```
valentin-krasontovitsch, line 55: were these commented code lines left by accident?

src/ert/shared/services/_base_service.py
```diff
@@ -103,6 +104,7 @@ def __init__(
         self._exec_args = exec_args
         self._timeout = timeout
         self._set_conn_info = set_conn_info
+        self._service_config_path = Path(project) / f"{self._service_name}_server.json"
```
valentin-krasontovitsch, line 107: is it strictly speaking necessary to call `Path` on `project`? according to the function definition of `__init__`, `project` is already of type `Path`

src/ert/shared/services/_base_service.py
```diff
@@ -162,17 +164,17 @@ def _assert_server_not_running(self) -> None:
...
                     f"present on this location. Retry {i}"
                 )
                 time.sleep(1)
 
-        if (Path.cwd() / f"{self._service_name}_server.json").exists():
+        if self._service_config_path.exists():
             print(
                 f"A file called {self._service_name}_server.json is present from this "
-                "location. This indicates there is already a ert instance running. "
+                f"location. This indicates there is already a ert instance running. "
```
valentin-krasontovitsch, line 177: no f necessary in this string / line : )

src/ert/shared/services/_base_service.py
```diff
@@ -312,12 +312,9 @@ def connect(
...
         while t < timeout:
-            search_path = Path(path)
-            while search_path != Path("/"):
-                if (search_path / name).exists():
-                    with (search_path / name).open() as f:
-                        return cls([], conn_info=json.load(f))
-                search_path = search_path.parent
+            if (path / name).exists():
+                with (path / name).open() as f:
+                    return cls([], conn_info=json.load(f), project=path)
```
valentin-krasontovitsch, line 318: we don't need to search upwards anymore? is that because we have more confidence about where we should find the file? or was the search strictly speaking not really necessary before, either?

src/ert/shared/services/webviz_ert_service.py
```diff
@@ -1,20 +1,19 @@
...
+        if kwargs.get("verbose", False):
             exec_args.append("--verbose")
-        if title:
-            exec_args.append("--title")
-            exec_args.append(title)
-        super().__init__(exec_args)
+        exec_args.extend(["--title", str(kwargs.get("title"))])
+        exec_args.extend(["--project_identifier", str(kwargs.get("project"))])
+
+        super().__init__(exec_args, project=kwargs.get("project"))
```
valentin-krasontovitsch, line 19: just to be a bit nitpicky - `kwargs.get("project")` can be refactored for DRYness :nerd_face:

valentin-krasontovitsch (approved):
Besides the questions / comments, this loogs good : )

src/ert/shared/services/storage_service.py
DanSava: Yes, I forgot to stash the changes before committing  :+1:

src/ert/shared/services/_base_service.py
DanSava: I don't think it was really necessary before also very prone to issues.

src/ert/shared/services/_base_service.py
DanSava: This is python and types in the function definitions are just hints, so it falls on the developer to make sure that what you are passing to the function is of a specified type, otherwise, bad things happen :). 

I can remove the explicit conversion to Path. Still, I need to make sure all the initialization of the BaseService is with an actual Path object and not a string, and what I get in the end is Path conversion in multiple other places, so I will change the expected type to string and only have the Path conversion here.

oysteoh:
Looks very good and seems to solve the issue perfectly :tada:

DanSava:
Jenkins test this, please!

src/ert/shared/services/_base_service.py
valentin-krasontovitsch: gotcha!

PR #3892, opened 2022-09-14
Remove UMASK config option

### LINKED ISSUE #2553: Passing config dict without umask to SiteConfig fails
The following test fails with `SiteConfig` attempting to pass `umask=None` to `_alloc_full`.

```python
import os
from res.enkf import ResConfig, SiteConfig

def touch(filename):
    with open(filename, "w") as fh:
        fh.write(" ")

def test_stuff(tmpdir):
    touch(tmpdir + "/rpfile")
    touch(tmpdir + "/datafile")
    os.mkdir(tmpdir + "/license")
    SiteConfig(
        config_dict={
            "CONFIG_DIRECTORY": str(tmpdir),
            "DATA_FILE": "datafile",
            "LICENSE_PATH": "license",
            "RES_CONFIG_FILE": "test.ert",
            "RUNPATH_FILE": "rpfile",
        }
    )
```

valentin-krasontovitsch:
Closes #2553

**Approach**
We simply ~supply a default value (`0022`) for the config dict path in case no `UMASK` is given~ remove the UMASK.

## Pre review checklist


valentin-krasontovitsch:
the fail in jenkins seems to be the same issue that has been seen elsewhere, and is already being discussed on slack, see this [internal slack link](https://equinor.slack.com/archives/CKC3H3W74/p1663138350517919)

tests/test_config_parsing/test_site_config.py
```diff
@@ -38,3 +38,13 @@ def test_site_config_defaults_umask(tmpdir):
...
         }
     )
+
+
+@pytest.mark.usefixtures("use_tmpdir")
+@given(config_dicts())
+def test_site_config_dict_without_umask_works(config_dict):
+    config_dict.pop(ConfigKeys.UMASK)
+    try:
+        SiteConfig(config_dict=config_dict)
```
oysteoh, line 48: should this test also verify the default value is set  and used?

valentin-krasontovitsch: that's a good question. I think it might make sense. Gonna add that.

valentin-krasontovitsch: it is done.

valentin-krasontovitsch:
@oysteoh test is updated, checks have passed!

oysteoh:
Looks good, just a final question. Where is the source of truth to the default value? Is this a default specified only once, or is this value supplied different places? If more than one - should we have a constant for the default value in case it is changed one place we are sure it is reflected everywhere? I guess we in not too distant future will have only one way of initializing the config so i'm not sure it's worth the effort.

src/ert/_c_wrappers/enkf/site_config.py
```diff
@@ -132,7 +132,7 @@ def __init__(self, user_config_file=None, config_content=None, config_dict=None)
             )
 
             env_var_list.convertToCReference(None)
-            umask = config_dict.get(ConfigKeys.UMASK)
+            umask = config_dict.get(ConfigKeys.UMASK, 0o022)
```
DanSava, line 135: what is `0o022` this exactly when I use this value as a default it results in a Resconfig with a SiteConfig with umask with value 18... and when I initialize a simple config from file the default value of the unmak is 2.

Then why not use config_dict.get(ConfigKeys.UMASK, 2) ?

valentin-krasontovitsch:
OK i thought the source was right there, but i must have found it during debugging - it's in `src/clib/lib/enkf/site_config.cpp` line 350, I quote:

> Set the umask for all file creation. The default equinor value is 0022, i.e. write access is removed from group and others.

valentin-krasontovitsch:
about being used several places, ah, i'm afraid you got me - i did not bother to check :disappointed:

i'm gonna take another look. of course we should be consistent.

src/ert/_c_wrappers/enkf/site_config.py
valentin-krasontovitsch: Hmm very good issue you're raising - so first of, `0o022` is (python) base 8 int notation, meaning 2*1 + 2*8 = 18.

the inconsistency with initiating a ResConfig from config file and getting a 2 is of course something that should be fixed, I will look into it.

the reason for using octal notation, I guess, is just convention / what some people might be used to in terms of linux file permissions. for those used to that notation, this way of writing it is easy to parse, while an `18` in base 10 notation is more difficult to interpret.

valentin-krasontovitsch:
OK after discussing a bit during standup, we've decided that I'm gonna try and remove UMASK from the code base - it's been deprecated for years, so this potentially breaking (config files) change should be OK.

valentin-krasontovitsch:
OK ready for review of "delete the umask" - there's two commits, the first sets a default umask in the job runner, which actually uses a umask. I chose (18)_10 = (022)_8 (base 10 and base 8 notation), based on some comments / that value seemingly being a sort of default.

BUT we could

- remove that settings of the umask altogether (consequences? why is it necessary? security concerns? should we just let it be whatever it is on the given system?)
- use a different default value
- 
i welcome any sort of opinion / arguments on the matter.

eivindjahren (commented):
Looks good! I will double check with @oyvindeide that we are ready for this.

oysteoh:
One question - you mention the reason for choosing `0022` is based on a comment. Is there no places where this value is actually used in code executed?

valentin-krasontovitsch:
another rebase with conflict, config dict generator had its structure changed, and i had to remove a reference to umask in that new structure

valentin-krasontovitsch:
> Is there no places where this value is actually used in code executed?

so as far as i can tell exactly where i'm using that value, the only left appearance of the string `umask` in the code base, is the place where it's used - the job runner sets the umask and then starts its work, as far as i understand

valentin-krasontovitsch:
as discussed in meeting, I have now upadted the value and the commit message to reflect that we are using the de facto default of `002`

mortalisk:
Darn, I also did this. Too many caffeine suppositories, I guess. Well, if you want to compare with mine to see if it's the same, here: https://github.com/equinor/ert/pull/3939

valentin-krasontovitsch:
test has been added ^^

oysteoh:
Looked through the logs we have, and UMASK is quite often set - and to different values... 😬  Removing it and specify the value ourselves will probably make a lot of user configurations to not work as intended anymore ( i think just removing the possibility of UMASK will not crash ERT  - just ignore the key? )

valentin-krasontovitsch:
I'm sorry, I'm not quite following, so I'm gonna try to paraphrase to see if I understood correctly:

- users are still using UMASK, and setting it to different values
- this PR would make ert not work as intended for those users whose config files specify a UMASK value different from the existing and used here default value
- removing the UMASK config option, as done in the current state of this PR, will not break ert as software, ert will still run
- you propose that we ignore the config key UMASK

I believe ignoring an unknown key is default behaviour for the config parser, right? so are you actually suggesting any changes to this PR?

I guess in general, if we make this change reality (removing UMASK), we should communicate this change to the users, to make them aware of the potential consequences, using perhaps the "executive summary" option as discussed in the last ... was it a retrospective?

oysteoh:
Been discussing this a bit now, and we should just remove it as done in this PR. BUT - @oyvindeide and @berland claims it should not be necessary to set the UMASK at all - it should just use the OS default - we should not care. Testing this PR in different environments and check which permissions are set is probably a good thing - just verify that a sane value is set.

valentin-krasontovitsch:
aha! so that means

- i should remove the line in job runner that sets the UMASK
- i can remove the test which checks what UMASK is used / set (as we cannot predict that value)
- someone (I?) should test in different environments what UMASK is used if we don't set UMASK

which environments do you suggest? is an RGS node the most / only important one? should i also try on my own ubuntu setup? should I ask somebody to test on a mac?

in that case I will remove the request for your review, @oyvindeide

oysteoh:
> which environments do you suggest? is an RGS node the most / only important one? should i also try on my own ubuntu setup? should I ask somebody to test on a mac?

Yes! :+1:  For RGS - at least log in to the Stavanger and Bergen clusters and verify the behavior at those locations.

valentin-krasontovitsch:
did the code changes as requested.

checking different environments:
- linux (ubuntu 22.04): ran under home dir, umask 0002 (same as default in ert, no write access for world)
- mac: TBD
- RGS node Bergen: TBD
- RGS node Stavanger: TBD

valentin-krasontovitsch:
ok i'm not able to use the RGS stuff right now, and in order to keep this civil and the language clean, i will not comment about my feelings towards RGS through connectit on an unmanaged laptop right now.

i have a creeping suspicion though that the default in these situations (sane *nix setup) is umask 0002, and we should be good.

maybe try again tomorrow, calling it a day for now.

valentin-krasontovitsch:
requesting final review after rebase to current main and verification of default UMASKs here and there (see list [above](https://github.com/equinor/ert/pull/3892#issuecomment-1257978991))

@eivindjahren @DanSava

mortalisk (approved):

valentin-krasontovitsch:
thank you @mortalisk :tada:

PR #3909, opened 2022-09-19
Pin protobuf and grpcio-tools

xjules:
**Issue**
Test for ert_documentation fails due to we need to have the older version of protobuf package.
```sh
 _EMPTY = _descriptor.Descriptor(
/prog/res/komodo/bleeding-py38-rhel7/root/lib64/python3.8/site-packages/google/protobuf/descriptor.py:313: in __new__
    _message.Message._CheckCalledFromGeneratedFile()
E   TypeError: Descriptors cannot not be created directly.
E   If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
E   If you cannot immediately regenerate your protos, some other possible workarounds are:
E    1. Downgrade the protobuf package to 3.20.x or lower.
E    2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).
```

**Approach**

## Pre review checklist

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,7 +3,7 @@ cmake-format
 decorator
 ecl_data_io
 flake8>=4.0.0
-grpcio-tools
+grpcio-tools==1.41.0
 isort
 jsonpath_ng
 jupytext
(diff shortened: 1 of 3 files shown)


lars-petter-hauge (approved):

lars-petter-hauge:
I think we should at the same time investigate how we can remove the pin as well

xjules:
> I think we should at the same time investigate how we can remove the pin as well

Yes, I think we should. Although, I guess this falls onto resinisght / rips.

PR #3919, opened 2022-09-20
Simplify EclConfig

### LINKED ISSUE #3918: redesign EclConfig
This started out as an investigation into `NUM_CPU`, but we figured that it might be nice to try and clean up `EclConfig`.

Arises from work on issue #3802 as can be seen in discussion in PR #3887, to be precise this [conversation thread in said PR](https://github.com/equinor/ert/pull/3887#discussion_r971671276).

at the moment, the default # of CPUs as set in some (perhaps all) paths of inititating for instance a queue config is `0`, which seems non-sensical.

fixing this requires a modification elsewhere in the code: in two spots, the queue config's # of CPUs being zero is used as a check.

valentin-krasontovitsch:
**Issue**
Closes #3918

**Approach**
Confer [discussion in other PR](https://github.com/equinor/ert/pull/3887#discussion_r971671276)

## Pre review checklist


valentin-krasontovitsch:
one of the github action test runs failed with 429 responses from github (too many requests) O_o

eivindjahren:
I believe this might introduce changed behavior as 0is used to indicate that NUM_CPU is read from somewhere else. I think we should first create some tests that fail if num_cpu is not handled correctly when not set in the config.

valentin-krasontovitsch:
based on the discussion we had i think we take care of all cases where `NUM_CPU=0` is used to signal something. do you agree that all occurences of the above are covered in the changes in this PR? 

tests sound like a good idea - what do you mean by "if num_cpu is not handled correctly when not set in the config" ? let's say num_cpu is not set in the config - what do we consider to be incorrect handling of that fact?

eivindjahren:
What I think would be reasonable behavior is:

* `res_config.num_cpu = None` when NUM_CPU is not set in the config.
* There is a function `get_num_cpu_from_ecl() -> Optional[int]` which returns the number of CPUs to use set in the eclipse configuration file (None if there is no such config or no such setting in that config)
`EnkfMain` then arbitrates between this sources with
```
def get_num_cpu(self):
  if self.resConfig().num_cpu is not None:
    return self.resConfig().num_cpu
  if self.get_num_cpu_from_ecl() is not None:
    return self.get_num_cpu_from_ecl()
  return 1
```

valentin-krasontovitsch:
well that is conceptionally definitely very nice : )

i can try to make that work!

I guess i'm still kind of timid, and want to change as little as possible / necessary, and don't think much about "what should this actually look like", so thanks for the nudge!

valentin-krasontovitsch:
just to make sure i understand - this would remove NUM_CPU from queue config, and move it up to ResConfig? Or does a call of `self.resConfig().num_cpu` actually under the hood look at queue config?

src/clib/lib/enkf/enkf_state.cpp
```diff
@@ -32,15 +32,13 @@ void enkf_state_initialize(rng_type *rng, enkf_fs_type *fs,
...
 
-ecl_sum_type *load_ecl_sum(const ecl_config_type *ecl_config,
-                           const char *run_path, const char *eclbase) {
+ecl_sum_type *load_ecl_sum(const char *run_path, const char *eclbase) {
     ecl_sum_type *summary = NULL;
 
-    const bool fmt_file = ecl_config_get_formatted(ecl_config);
     char *header_file = ecl_util_alloc_exfilename(
-        run_path, eclbase, ECL_SUMMARY_HEADER_FILE, fmt_file, -1);
+        run_path, eclbase, ECL_SUMMARY_HEADER_FILE, DEFAULT_FORMATTED, -1);
```
eivindjahren, line 39: So I would expect here that `ecl_config_get_formatted` and `ecl_config_type->io_config` would be deleted in this commit.

eivindjahren: Ah, now I understand, the entire file is deleted later, nvm.

src/clib/lib/enkf/ecl_config.cpp
```diff
@@ -1,483 +0,0 @@
...
-     actualt value is internalized as a job_name in the model_config.
-   */
-
-    if (config_content_has_item(config, ECLBASE_KEY)) {
-        ui_return_type *ui_return = ecl_config_validate_eclbase(
-            ecl_config, config_content_iget(config, ECLBASE_KEY, 0, 0));
-        if (ui_return_get_status(ui_return) == UI_RETURN_OK)
-            ecl_config->have_eclbase = true;
-        else
-            util_abort("%s: failed to set eclbase format. Error:%s\n", __func__,
```
eivindjahren, line 290: Abort is bad, but we should probably have similar error messages raised.

src/clib/lib/enkf/ecl_config.cpp
```diff
@@ -1,483 +0,0 @@
...
-static void handle_has_data_file_key(ecl_config_type *ecl_config,
-                                     const config_content_type *config) {
-    const char *data_file =
-        config_content_get_value_as_abspath(config, DATA_FILE_KEY);
-    ui_return_type *ui_return =
-        ecl_config_validate_data_file(ecl_config, data_file);
-    if (ui_return_get_status(ui_return) == UI_RETURN_OK)
-        ecl_config_set_data_file(ecl_config, data_file);
-    else
-        util_abort("%s: problem setting ECLIPSE data file (%s)\n", __func__,
```
eivindjahren, line 305: Abort is bad, but we should probably have similar error messages raised.  It would be good to add a test that checks that good error messages are given (ie. `pytest.raises(ValueError, match="{good message}"):`)

src/clib/lib/enkf/ecl_config.cpp
```diff
@@ -1,483 +0,0 @@
...
-static void handle_has_grid_key(ecl_config_type *ecl_config,
-                                const config_content_type *config) {
-    const char *grid_file =
-        config_content_get_value_as_abspath(config, GRID_KEY);
-
-    ui_return_type *ui_return = ecl_config_validate_grid(ecl_config, grid_file);
-    if (ui_return_get_status(ui_return) == UI_RETURN_OK)
-        ecl_config_set_grid(ecl_config, grid_file);
-    else
-        util_abort("%s: failed to set grid file:%s  Error:%s \n", __func__,
```
eivindjahren, line 319: Abort is bad, but we should probably have similar error messages raised. I see that you check whether the file has the correct type. It would be good to add a test that checks that good error messages are given (ie. `pytest.raises(ValueError, match="{good message}"):`)

src/clib/lib/enkf/ecl_config.cpp
```diff
@@ -1,483 +0,0 @@
...
-    ui_return_free(ui_return);
-}
-
-static void handle_has_refcase_key(ecl_config_type *ecl_config,
-                                   const config_content_type *config) {
-    const char *refcase_path =
-        config_content_get_value_as_abspath(config, REFCASE_KEY);
-
-    if (!ecl_config_load_refcase(ecl_config, refcase_path))
-        fprintf(stderr, "** Warning: loading refcase:%s failed \n",
```
eivindjahren, line 331: We should probably have similar warning message printed, and test that we get the expected warning: `pytest.warns(UserWarning, match="..."):`.

src/clib/lib/enkf/ecl_config.cpp
```diff
@@ -1,483 +0,0 @@
...
-
-    if (config_content_has_item(config, GRID_KEY))
-        handle_has_grid_key(ecl_config, config);
-
-    if (config_content_has_item(config, REFCASE_KEY))
-        handle_has_refcase_key(ecl_config, config);
-
-    if (ecl_config->can_restart)
-        fprintf(
-            stderr,
```
eivindjahren, line 363: We should probably have similar warning message printed, and test that we get the expected warning: pytest.warns(UserWarning, match="..."):.

src/clib/lib/enkf/ecl_config.cpp
```diff
@@ -1,483 +0,0 @@
...
-const char *ecl_config_get_depth_unit(const ecl_config_type *ecl_config) {
-    switch (ecl_config->unit_system) {
-    case ECL_METRIC_UNITS:
-        return "M";
-    case ECL_FIELD_UNITS:
-        return "FT";
-    case ECL_LAB_UNITS:
-        return "CM";
-    default:
-        util_abort("%s: unit system enum value:%d not recognized \n", __func__,
```
eivindjahren, line 464: Abort is bad, but we should probably have similar error messages raised. It might be that this error message is already good, but we should add a test for such a message (ie. pytest.raises(ValueError, match="{good message}"):)

src/clib/lib/enkf/ecl_config.cpp
```diff
@@ -1,483 +0,0 @@
...
-const char *ecl_config_get_pressure_unit(const ecl_config_type *ecl_config) {
-    switch (ecl_config->unit_system) {
-    case ECL_METRIC_UNITS:
-        return "BARSA";
-    case ECL_FIELD_UNITS:
-        return "PSIA";
-    case ECL_LAB_UNITS:
-        return "ATMA";
-    default:
-        util_abort("%s: unit system enum value:%d not recognized \n", __func__,
```
eivindjahren, line 479: Abort is bad, but we should probably have similar error messages raised. It might be that this error message is already good, but we should add a test for such a message (ie. pytest.raises(ValueError, match="{good message}"):)

eivindjahren (changes requested):

tests/libres_tests/res/enkf/test_ecl_config.py
```diff
@@ -1,93 +1,59 @@
...
-
+def test_ecl_config_default_num_cpu_value(source_root):
     dfile = str(source_root / "test-data/eclipse/SPE1.DATA")
-    ui = ec.validateDataFile(dfile)
-    assert ui
-    ec.setDataFile(dfile)
-    assert dfile == ec.getDataFile()
+    ec = EclConfig(data_file=dfile)
+    assert dfile == ec._data_file
+    assert 1 == ec.num_cpu
```
oysteoh, line 13: Use 'ec.num_cpu == 1' instead of '1 == ec.num_cpu' (Yoda-conditions)

valentin-krasontovitsch:
@eivindjahren there's only three things left that might need to raise warnings or exceptions. it's gonna be difficult to insert error checking for those, as they basically just call python or in one case a c method from the ecl repo. with this simplification, we are at the mercy of those functions doing something meaningful.

also just got an idea, i could wrap it all in a try and except thingy, hoping that the functions raise something and then we can work with that. i'm gonna check out a bit how the functions are written with respect to error handling.

if you have any good ideas about this, please let me know!

regarding validation of like is a refcase file, grid file, data file given, i skipped that, as i'm not sure what we want to have, and where to put it - cf. also #3985

src/ert/_c_wrappers/enkf/ecl_config.py
```diff
@@ -1,207 +1,96 @@
...
-        """
-        Has ECLIPSE been configured?"
-        """
-        return self._active()
-
-    def getLastHistoryRestart(self):
-        return self._get_last_history_restart()
+        return cls(**ecl_config_args)
+
+    def _load_grid(grid_file: str) -> Optional[EclGrid]:
```
oysteoh, line 43: i might not understand this fully, but the assignment of `self._grid_file` in the initialization and then the later `EclConfig._load_grid(grid_file)` .... Why do we need to add the `grid_file` as a parameter? Could the `_load_grid(..)` be called without an argument and inside the function just use the already assigned `self._grid_file`?

src/ert/_c_wrappers/enkf/ecl_config.py
```diff
@@ -1,207 +1,96 @@
...
+        data_file: Optional[str] = None,
+        grid_file: Optional[str] = None,
+        refcase_file: Optional[str] = None,
+    ):
+        self._data_file = data_file
+        self._grid_file = grid_file
+        self.grid = EclConfig._load_grid(grid_file)
+
+        self._refcase_file = refcase_file
+        self.refcase = EclConfig._load_refcase(refcase_file)
```
oysteoh, line 24: Argument 1 to "_load_refcase" of "EclConfig" has incompatible type "Optional[str]"; expected "str"

src/ert/_c_wrappers/enkf/ecl_config.py
oysteoh, line 43: `def _load_grid(self, grid_file: str) -> Optional[EclGrid]`

src/ert/_c_wrappers/enkf/ecl_config.py
```diff
@@ -1,207 +1,96 @@
...
+            return None
+        ecl_grid_file_types = [
+            EclFileEnum.ECL_GRID_FILE,
+            EclFileEnum.ECL_EGRID_FILE,
+        ]
+        if not get_file_type(grid_file) in ecl_grid_file_types:
+            raise ValueError(f"grid file {grid_file} does not have expected type")
+        return EclGrid.load_from_file(grid_file)
+
+    def _load_refcase(refcase_file: str) -> EclSum:
```
oysteoh, line 54: `self` 😄

src/ert/_c_wrappers/enkf/ecl_config.py
valentin-krasontovitsch: sorry, that should have been a class method (to make it clear that it can be moved outta there)

tests/libres_tests/res/enkf/test_ecl_config.py
valentin-krasontovitsch: ah but isn't it assert expected is equal to actual? or is it always exactly the other way around? :thinking:

src/clib/lib/enkf/ecl_config.cpp
valentin-krasontovitsch: code no longer used

src/ert/_c_wrappers/enkf/ecl_config.py
```diff
@@ -1,207 +1,96 @@
...
+class EclConfig:
+    def __init__(
+        self,
+        data_file: Optional[str] = None,
+        grid_file: Optional[str] = None,
+        refcase_file: Optional[str] = None,
+    ):
+        self._data_file = data_file
+        self._grid_file = grid_file
+        self.grid = EclConfig._load_grid(grid_file)
```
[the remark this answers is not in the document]
valentin-krasontovitsch: i should really check these annotations myself :sweat_smile: 
gonna check out what went wrong here!

src/ert/_c_wrappers/enkf/ecl_config.py
valentin-krasontovitsch: ja indeed my intention here is that i had in the back of my head that there should be as little logic in the subconfig objects as possible, or at least all parsing should happen in the res config, or something along those lines.

so i wanted to make it clear that this function can basically stand on its own, I guess it doesn't even need to be defined as part of the class, can just be used by it. i'm gonna try to make it so.

valentin-krasontovitsch: never mind, making it a stand alone function, see above

src/ert/_c_wrappers/enkf/ecl_config.py
valentin-krasontovitsch: indeed : )

valentin-krasontovitsch: gonna take same approach as above

src/ert/_c_wrappers/enkf/ecl_config.py
valentin-krasontovitsch: that sounds strangely reasonable, and like it doesn't even need deep understanding of object oriented stuff :sweat_smile:

ok so simple instance methods, that don't get the files as arguments, but grab them from the instance, got it!

src/clib/lib/enkf/ecl_config.cpp
valentin-krasontovitsch: eclbase is no longer used, so we can scrap the warning, too!

src/clib/lib/enkf/ecl_config.cpp
valentin-krasontovitsch: gonna check out the python function now used for refcase loading for its error handling and see what can be done there, as per other comment

src/clib/lib/enkf/ecl_config.cpp
valentin-krasontovitsch: validate data file just checks for its existence, so also here i will reference the validation of files, which is one of the problems that we took and put all the way over there: #3985

tests/libres_tests/res/enkf/test_ecl_config.py
valentin-krasontovitsch: anyways it is fixed now. the real truth will remain a mystery of the cosmos to me, as i seemingly cannot remember either way.

src/ert/_c_wrappers/enkf/ecl_config.py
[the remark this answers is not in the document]
valentin-krasontovitsch: yes, yes, of course, i see it now - the file can be None, that's silly of me. it is fixed now!

src/ert/_c_wrappers/enkf/ecl_config.py
valentin-krasontovitsch: also fixed!

valentin-krasontovitsch:
so checking out loading the refcase, which used the `__init__` function of the `EclSum` class found in the `ecl` repo, there are exceptions raised in python. no technical details (as in strings explaining anything) are passed up from the C code called under the hood in `ecl`, but we get a message a la `Failed to create summary instance from argument:[INSERT FILENAME]` of type `IOError`

i am welcome to hear suggestions on how we can improve on this.

in the meantime, i'm gonna try to write a test or two, to tickle out these error messages, as suggested by Eivind!

valentin-krasontovitsch:
oh, two questions:

- do we want to catch the errors and make them warnings? as this happens at the moment during init of config, we might want to say "whatever if you don't manage to load this refile, just warn about it, and move in with life i.e. execution of code"? or do we want to fail with an exception? i guess for the former we should also have a plan about what happens downstream in the program, i.e. make sure that we catch such a situtation (no refcase where one is expected).
- where do i put a test of this? it's not really config parsing, it's testing logic in the config object. do i create a new folder `config` in `tests/ert_tests`?

valentin-krasontovitsch:
okay i think i addressed everything - the things i fixed or think we don't need to fix i resolved, and the only thing left to discuss i think is where to test and how to handle undesired events when loading stuff (see comment above).

I've pushed the (non trivial / minute) changes in extra commits, for your convenience, so you can easily review them.

Please also tell me how you feel about the handling of config content as a dict (and the simplification needed there to extract values that are packed as listlists) - see own commit d903f1c4e55b7fc2d69db968d1227f1a6a63aa36

tests/test_config_parsing/test_ecl_config.py
```diff
@@ -17,4 +17,59 @@ def test_ecl_config_dict_creates_equal_config(config_dict):
...
+    grid_file = "BRICKWALL"
+    grid_file_content = """
+_|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|
+___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|__
+_|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|
+___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|__
+_|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|
+___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|__
+_|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|
+___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|___|__
```
eivindjahren, line 55: He he, funny but perhaps just say 4 lines of this? Also, maybe a clear indication that this is not what a grid file looks like for those not initiated into dark magic.

tests/test_config_parsing/test_ecl_config.py
```diff
@@ -17,4 +17,59 @@ def test_ecl_config_dict_creates_equal_config(config_dict):
...
+
+"""
+    with open(grid_file, "w+", encoding="utf-8") as grid_file_handler:
+        grid_file_handler.write(grid_file_content)
+    try:
+        EclConfig(grid_file=grid_file)
```
eivindjahren, line 71: You can use

```python
with pytest.raises(ValueError, match=grid_file):
  _ = EclConfig(grid_file=grid_file)
```

match can also contain other parts of the error message.

tests/test_config_parsing/test_ecl_config.py
```diff
@@ -17,4 +17,59 @@ def test_ecl_config_dict_creates_equal_config(config_dict):
...
+\\______   \\_   _____/\\_   _____/\\_   ___ \\   /  _  \\  /   _____/\\_   _____/
+ |       _/|    __)_  |    __)  /    \\  \\/  /  /_\\  \\ \\_____  \\  |    __)_
+ |    |   \\|        \\ |     \\   \\     \\____/    |    \\/        \\ |        \\
+ |____|_  /_______  / \\___  /    \\______  /\\____|__  /_______  //_______  /
+        \\/        \\/      \\/            \\/         \\/        \\/         \\/
+"""
+    with open(refcase_file, "w+", encoding="utf-8") as refcase_file_handler:
+        refcase_file_handler.write(refcase_file_content)
+    try:
+        EclConfig(refcase_file=refcase_file)
```
eivindjahren, line 37: Also here we can use
```python
with pytest.raises(ValueError, match=refcase_file):
  _ = EclConfig(refase_file=refcase_file)
```

match can also contain other parts that should be in the error message.

src/ert/_c_wrappers/enkf/ecl_config.py
```diff
@@ -1,207 +1,91 @@
...
+        if not self._data_file:
+            return None
+        return get_num_cpu_from_data_file(self._data_file)
+
+    def _load_grid(self) -> Optional[EclGrid]:
+        ecl_grid_file_types = [
+            EclFileEnum.ECL_GRID_FILE,
+            EclFileEnum.ECL_EGRID_FILE,
+        ]
+        if not get_file_type(self._grid_file) in ecl_grid_file_types:
```
eivindjahren, line 53: :robot:

eivindjahren (changes requested):
Just some small comments here, I think the implementation is ready, but it would be nice to use pytest.raises in the tests.

tests/test_config_parsing/test_ecl_config.py
valentin-krasontovitsch: pytest.raises, gotcha - i'm learning every day! and will add a comment : )

oysteoh (commented):
i think from my part it looks allright now ! but i'll leave it up to @eivindjahren  to approve the changes he have requested 👍

tests/test_config_parsing/test_ecl_config.py
valentin-krasontovitsch: will do, thanks : )

tests/test_config_parsing/test_ecl_config.py
valentin-krasontovitsch: sure thing!

src/ert/_c_wrappers/enkf/ecl_config.py
valentin-krasontovitsch: not sure about this one - my hunch is, it should `X not in LIST` rather thatn `not X in LIST`?

tests/test_config_parsing/test_ecl_config.py
valentin-krasontovitsch: done!

tests/test_config_parsing/test_ecl_config.py
valentin-krasontovitsch: done!

tests/test_config_parsing/test_ecl_config.py
valentin-krasontovitsch: done!

src/ert/_c_wrappers/enkf/ecl_config.py
valentin-krasontovitsch: if that's it, then it's done

eivindjahren (approved):
Just needs to fix style and all should be good :rocket:

src/ert/_c_wrappers/enkf/ecl_config.py
valentin-krasontovitsch: that was it!

src/clib/lib/enkf/enkf_state.cpp
```diff
@@ -302,6 +300,8 @@ static std::pair<fw_load_status, std::string> enkf_state_internalize_results(
 
     enkf_fs_type *sim_fs = run_arg_get_sim_fs(run_arg);
     int last_report = time_map_get_last_step(enkf_fs_get_time_map(sim_fs));
+    // TODO investigate usage of last history: is last_report < 0 possible? is
```
dafeda, line 303: I think it's better to create an issue than adding TODO's to code as they are quite hidden.

valentin-krasontovitsch: ARGH thank you for catching that! that's a personal note that should never have been merged :grimacing:

valentin-krasontovitsch: gonna make a PR that removes this immediately.

valentin-krasontovitsch: fixed in https://github.com/equinor/ert/pull/4037

PR #3935, opened 2022-09-23
Make test_site_config_dict_same_as_from_file pass 

### LINKED ISSUE #3801: Creating SiteConfig from dict and from file results in very different objects
When creating `SiteConfig` from a file name. The job list will be populated, while if you create from dict, the list will not be populated. This will have to be reconsiled in some way. Perhaps the best way to handle it is that both code paths will have to explicitly add the jobs.

Current behavior:
```python
>>> from ert._c_wrappers.enkf import SiteConfig
>>> filename = "..." # name of config file
>>> config_dict = {...} # Config dictionary with same values as the config file
>>> SiteConfig(filename).getAvailableJobs()
ExtJoblist(size=28, joblist=['DELETE_FILE: 94668020356560',...
>>> SiteConfig(config_dict=config_dict).getAvailableJobs()
ExtJoblist(size=0, joblist=[]) ...
```

mortalisk:
**Issue**
Resolves #3801

### DIFF
--- a/src/clib/lib/enkf/site_config.cpp
+++ b/src/clib/lib/enkf/site_config.cpp
@@ -145,7 +145,7 @@ static void site_config_load_config(site_config_type *site_config) {
 /**
  * NOTE: The queue config is not loaded until the site_config_alloc_load_user.
  */
-static site_config_type *site_config_alloc_default() {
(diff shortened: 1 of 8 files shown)


DanSava (approved):
LGTM :+1:

PR #3939, opened 2022-09-23
Remove umaks

mortalisk:


valentin-krasontovitsch:
a quick diff against my branch shows that

- I had to rebase onto main : )
- you have missed a few spots, some with the literal `umask` string, some more hidden, like in `src/ert/_c_wrappers/enkf/site_config.py`, the `ResPrototype` call of `site_config_alloc_full` shouldn't have an `int` in the argument list anymore
- you should run `clang-format` on `src/clib/lib/include/ert/job_queue/forward_model.hpp`
- the `docs` (directory) can also be purged of umask

and finally the most important difference (all of the above is just little things) - you removed the line

```
os.umask(int("002", 8))
```

from the job runner. We discussed this a bit yesterday, and it seems like we have to be careful about which file permissions the files created by ert get - i will spare you the details for now. we went for keeping that default value of `002`, but i would love to hear your thoughts on why you chose to remove it : )

mortalisk:
Regarding the removing of `os.umask(int("002", 8))`: I just assumed the program would have the correct defaults, and the shared file system would also default to something. I have no idea, I just did it fast and furious, and thought a review would discuss it.

valentin-krasontovitsch:
that sounds very reasonable : )

PR #2603, opened 2021-12-15
real->realizations, iter_ -> iter-

berland:
**Issue**
Resolves #2605 

**Approach**
Search and replace, with manual selection where to fix. Limited to ert configuration files and example directories, and some changes to RST-documentation.

There is a random mix of realisation vs realization in the source code, as in log messages and function naming. This has not been touched. 

Adjustments to some tests were needed.

## Pre review checklist

- [x] Added appropriate labels

PR #3952, opened 2022-09-27
Initialize res config form minimal dictionary 

### LINKED ISSUE #2554: Minimal ResConfig via config dict fails with segfault
While the following works:
```python
import os
from res.enkf import ResConfig, SiteConfig

def touch(filename):
    with open(filename, "w") as fh:
        fh.write(" ")

def test_works(tmpdir):
    touch(tmpdir + "/rpfile")
    touch(tmpdir + "/datafile")
    os.mkdir(tmpdir + "/license")
    with open(tmpdir + "/test.ert", "w") as fh:
        fh.write(
            """
JOBNAME  Job%d
NUM_REALIZATIONS  1
RUNPATH_FILE rpfile
DATA_FILE datafile
LICENSE_PATH license
"""
        )
    ResConfig(str(tmpdir + "/test.ert"))
```

The corresponding config_dict segfaults:
```python
def test_fails(tmpdir):
    touch(tmpdir + "/rpfile")
    touch(tmpdir + "/datafile")
    os.mkdir(tmpdir + "/license")
    ResConfig(str(tmpdir + "/test.ert")).site_config == ResConfig(
        config_dict={
            "CONFIG_DIRECTORY": str(tmpdir),
            "DATA_FILE": "datafile",
            "LICENSE_PATH": "license",
            "RES_CONFIG_FILE": "test.ert",
            "RUNPATH_FILE": "rpfile",
            "UMASK": 2,
        }
    )
```

DanSava:
**Issue**
Resolves #2554

## Pre review checklist

### DIFF
--- a/src/ert/_c_wrappers/enkf/model_config.py
+++ b/src/ert/_c_wrappers/enkf/model_config.py
@@ -130,7 +130,7 @@ class ModelConfig(BaseCClass):
             num_realizations = config_dict.get(ConfigKeys.NUM_REALIZATIONS)
 
             # RUNPATH_KEY
-            run_path = config_dict.get(ConfigKeys.RUNPATH)
+            run_path = config_dict.get(ConfigKeys.RUNPATH, "simulations/realization%d")
             if run_path is not None:
                 run_path = os.path.realpath(run_path)
 
@@ -140,7 +140,7 @@ class ModelConfig(BaseCClass):
                 data_root = os.path.realpath(data_root_from_config)
 
             # ENSPATH_KEY
-            ens_path = config_dict.get(ConfigKeys.ENSPATH)
+            ens_path = config_dict.get(ConfigKeys.ENSPATH, "storage")
             if ens_path is not None:
                 ens_path = os.path.realpath(ens_path)
 
--- a/docs/reference/configuration/keywords.rst
+++ b/docs/reference/configuration/keywords.rst
@@ -67,7 +67,7 @@ Keyword name                                                            Required
 :ref:`NUM_REALIZATIONS <num_realizations>`                              YES                                                                     Set the number of reservoir realizations to use
 :ref:`OBS_CONFIG <obs_config>`                                          NO                                                                      File specifying observations with uncertainties
 :ref:`QUEUE_OPTION <queue_option>`                                      NO                                                                      Set options for an ERT queue system
-:ref:`QUEUE_SYSTEM <queue_system>`                                      NO                                                                      System used for running simulation jobs
+:ref:`QUEUE_SYSTEM <queue_system>`                                      NO                                      LOCAL_DRIVER                                System used for running simulation jobs
 :ref:`REFCASE <refcase>`                                                NO                                                                      Reference case used for observations and plotting (See HISTORY_SOURCE and SUMMARY)
 :ref:`RERUN_START  <rerun_start>`                                       NO                                      0                               Deprecated
 :ref:`RESULT_PATH  <result_path>`                                       NO                                      results/step_%d                 Define where ERT should store results
--- a/src/ert/_c_wrappers/enkf/queue_config.py
+++ b/src/ert/_c_wrappers/enkf/queue_config.py
@@ -14,6 +14,7 @@
 #  See the GNU General Public License at <http://www.gnu.org/licenses/gpl.html>
 #  for more details.
 
+import shutil
 from typing import Optional
 
 from cwrap import BaseCClass
@@ -22,6 +23,7 @@ from ert._c_wrappers import ResPrototype
 from ert._c_wrappers.config import ConfigContent
 from ert._c_wrappers.enkf.config_keys import ConfigKeys
 from ert._c_wrappers.job_queue import Driver, JobQueue
+from ert._c_wrappers.job_queue.driver import LOCAL_DRIVER
 
 
 class QueueConfig(BaseCClass):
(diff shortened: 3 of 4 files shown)


valentin-krasontovitsch:
so something i feel is funny - if i comment out `touch("rpfile")`, the test still runs and passes, no problem
if i comment out `touch("datafile")` on the other hand, i get a segfault :thinking:
the `mkdir("license")` also seems to be unnecessary

i know it goes beyond the original issue of just making the test pass, but maybe we can use this occasion to investigate more what's going on with these config keys exactly? what do you think, Dan?

valentin-krasontovitsch:
after checking a bit more, this seems to be just an unhappy coincidence that only affects our tests as pytest tries to be helpful and display the objects involved in its output, calling `__repr__` on our objects. if one executes the contents of the test as a script, with commented out data file, one gets the expected error message:

```
Could not find ECLIPSE data file: /tmp/hufffff/datafile
```

I'm gonna make a separate PR for fixing those pesky seg faults... maybe it'll help (somewhat)

still it is strange that license and prdata don't need to exist. that means that those lines in the test are superfluous... not sure what the right way of handling this is.

DanSava:
> so something i feel is funny - if i comment out `touch("rpfile")`, the test still runs and passes, no problem if i comment out `touch("datafile")` on the other hand, i get a segfault thinking the `mkdir("license")` also seems to be unnecessary
> 
> i know it goes beyond the original issue of just making the test pass, but maybe we can use this occasion to investigate more what's going on with these config keys exactly? what do you think, Dan?

Nothing too crazy is going on with these config keys:
The reason we can remove `touch("rpfile")` without affecting the tests is because if this rpfile (run path file) does not exist ERT will create it for us so it might be that adding it to the test is a bit unneeded.

Regarding the license folder that one is used when setting up the `SiteConfig` object and am sure once #3801 is resolved this will be properly handled, and by that I mean check that the folder exists and if we have the key in the config dictionary.

So yes there are other keys that are not really needed in order to initialize a ResConfig object, and it is true that the test does not represent the **minimal** config dictionary needed, I will update the tests accordingly

valentin-krasontovitsch:
I see, thanks for explaining! i even got it to work with an empty dict - while that from a programmer's point of view might be the minimal config dict, i guess from an ert user's point of view, it might not make much sense...

valentin-krasontovitsch:
could you comment shortly why you're adding the default values in `model_config.py`? is that the default values one gets when initializing from file without setting those options?

and i guess the same holds for `queue_config.py`?

DanSava:
> could you comment shortly why you're adding the default values in `model_config.py`? is that the default values one gets when initializing from file without setting those options?

Yes 

> and i guess the same holds for `queue_config.py`? 

Yes

DanSava:
> I see, thanks for explaining! i even got it to work with an empty dict - while that from a programmer's point of view might be the minimal config dict, 

I see you merged your UMASK removal PR so yes I guess we can now initialize a res config with an empty dictionary :partying_face: 

>i guess from an ert user's point of view, it might not make much sense...

I think this is exactly what we want, to test, that we get reasonable defaults when we don't pass all the keys (or any of them)

valentin-krasontovitsch:
oh another question, maybe rather technical - i just found the ENSPATH default value "storage" already defined in `src/clib/lib/include/ert/enkf/enkf_defaults.hpp`. is there any way that we can recycle the value from there, so we don't have it defined in two spots?
or is this not really necessary to think about at the moment, as we're hoping to get rid of (most parts of?) the C code for the config anyways?

DanSava:
> ok cool - very good : )
> 
> can we then also add the values to our docs, and trigger the `Required` to No where applicable?

Good observation but, I am not sure this is applicable [the fmu docs listing all config keys](https://fmu-docs.equinor.com/docs/ert/reference/configuration/keywords.html) is referring to keys in the config file and not keys in the dictionary used to initialized a ResConfig object. The first is user facing the second is not really. 

> i'm looking at [the fmu docs listing all config keys](https://fmu-docs.equinor.com/docs/ert/reference/configuration/keywords.html) - is this automagically generated from something in the ert repo?

I think so yes

> the following key - default value pairs are missing, as far as i can see:
> 
> * `NUM_REALIZATIONS` : 1 (should have required: no)

Not really it should still be 1,  it is required in the ert config file

> * `QUEUE_SYSTEM` : `LOCAL_DRIVER`

Yes this probably should be added

> the others are either already correct or will be removed by eivind in the very near future.

Maybe all should be dealt with as part of an issue that focuses with updating these config file keys.

DanSava:
Jenkins test this, please!

valentin-krasontovitsch:
the origin for these default values is what happens when you init from config file without setting the keys in question, and that is user facing. these defaults are based on the de facto defaults of ert - we should update the docs to reflect this. also, we want the config file and the dict to behave identically, right?
i think updating the docs for parts of the software that we touch whenever we touch them is an integral part of the development process - otherwise, it will always be an afterthought. but that's just me... also, of course, if these docs are not generated based on something in ert, my point here is a bit moot : )
i'm gonna ask in slack if anybody knows what those docs are based on, and create an issue if it's outside of ert

DanSava:
> apparently the file comes directly from `docs/reference/configuration/keywords.rst` - if you update it as requested, I promise bacon : )
> 
> besides two points:
> 
> * you probably didn't see my [comment about the constant shared between C and python code](https://github.com/equinor/ert/pull/3952#issuecomment-1262190182)

I saw your comment but I thought it was very strange what you were asking and I just thought that you will rethink and realize you are asking to use a constant defined in a C header file inside python. I cannot think of an easy straightforward way of doing that without adding some crazy hacky code to the codebase which I really don't want to do, there is enough of that already. As you already mentioned we will be removing the config C code once we completely move the config parsing to python so we won't have things defined in multiple places.   

> * > Not really it should still be 1, it is required in the ert config file
> 
> If it is required in the config file, it should be required in the config dict as well, no? 

Good point :+1:

src/ert/_c_wrappers/enkf/model_config.py
```diff
@@ -130,7 +130,7 @@ def __init__(
             num_realizations = config_dict.get(ConfigKeys.NUM_REALIZATIONS)
 
             # RUNPATH_KEY
-            run_path = config_dict.get(ConfigKeys.RUNPATH)
+            run_path = config_dict.get(ConfigKeys.RUNPATH, "simulations/realization%d")
```
berland, line 133: Suggest to stick to the FMU standard, which would include a dash between `realization` and `%d`.

DanSava: Thanks for the suggestion but when looking at `src/clib/lib/include/ert/enkf/enkf_defaults.hpp` the default run path is as follows:
```
#define DEFAULT_RUNPATH "simulations/realization%d"
```
So I am not sure the FMU standard is used in this case, but maybe I am confused

berland: Fair point, your line 133 should probably be consistent with what's in `enkf_defaults.hpp`. I thought I had changed it all in https://github.com/equinor/ert/pull/2603 but apparently not.

valentin-krasontovitsch (approved):
very nice!

DanSava:
Jenkins test this, please!

PR #3981, opened 2022-10-04
Mute external DEBUG messages

berland:
This removed DEBUG logs from some packages that are not as critical to the ERT application.

In particular, for Drogon, `subscript.fmuobs.parsers` emit 200-300 lines of DEBUG messages when parsing the observations.

**Issue**
Resolves overloading users and debuggers with not-so-interesting log messages.

**Approach**
Adjust the log-level on a per package level away from the default `DEBUG`. This does not scale perfectly, as more packages introduced by users through configuration can still spur similar loads of debug messages.

## Pre review checklist

### DIFF
--- a/src/ert/logging/logger.conf
+++ b/src/ert/logging/logger.conf
@@ -71,14 +71,12 @@ loggers:
     level: DEBUG
     handlers: [eefile]
     propagate: no
-  websockets.server:
-    level: ERROR
-    handlers: [eefile]
-    propagate: yes
   ert.experiment_server:
     level: DEBUG
     handlers: [experiment_server_file]
     propagate: yes
+  h5py:
+    level: INFO
   matplotlib:
     level: INFO
   res:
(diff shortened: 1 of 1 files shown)


src/ert/logging/logger.conf
```diff
@@ -87,6 +85,14 @@ loggers:
   res.config:
     level: WARNING
     handlers: [stderr, file]
+  shapely:
+    level: INFO
+  subscript:
+    level: INFO
```
hnformentin, line 91: - How did you choose the packages to be included here (h5py, shapely and subscript)?  Was it based on the logs from Drogon?
- Is there a default option if we don't specify the level for a given package?

berland: These package names are taken from an observation of what fills up the terminal window (or the ert log files) when running Drogon or similar. 

(the default is DEBUG, which is slightly unfortunate for some external packages that might be very chatty at this level, notably `subscript.fmuobs.parsers` in the case of Drogon.)

hnformentin (approved):
LGTM!

PR #3982, opened 2022-10-04
Make default ensemble path consistent

### LINKED ISSUE #3925: Inconsistent  behavior when setting  ENSPATH in a config file versus not setting it
The following test should **SUCCEED** but it **FAILS** 

``` python
def test_default_resconfig_ens_path(tmpdir):
    from ert._c_wrappers.enkf import ResConfig

    with tmpdir.as_cwd():
        config_file = "test.ert"
        with open(config_file, "w") as f:
            f.write("""
NUM_REALIZATIONS  1
            """)
        res_config = ResConfig(config_file)
        # By default, the ensemble path is set to 'storage'
        default_ens_path = res_config.model_config.getEnspath()

        with open(config_file, "a") as f:
            f.write("""
ENSPATH storage            
            """)

        # Once we set the ENSPATH in the config file the behaviour
        # of getEnspath changes, it should not!
        res_config = ResConfig(config_file)
        ens_path = res_config.model_config.getEnspath()

        assert default_ens_path == ens_path
```

DanSava:
**Issue**
Resolves #3925

**Approach**
Get the same ensemble path if the user adds `storage` as the ENSPATH in the config file or not.

## Pre review checklist

### DIFF
--- a/src/clib/lib/enkf/model_config.cpp
+++ b/src/clib/lib/enkf/model_config.cpp
@@ -39,6 +39,8 @@
 #include <ert/enkf/model_config.hpp>
 #include <ert/enkf/site_config.hpp>
 
+namespace fs = std::filesystem;
+
 static auto logger = ert::get_logger("enkf");
 
 #define MODEL_CONFIG_TYPE_ID 661053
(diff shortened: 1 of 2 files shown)


src/clib/lib/enkf/model_config.cpp
```diff
@@ -444,6 +446,9 @@ void model_config_init(model_config_type *model_config,
         model_config_set_enspath(
             model_config,
             config_content_get_value_as_abspath(config, ENSPATH_KEY));
+    else
+        model_config_set_enspath(model_config,
+                                 fs::absolute(DEFAULT_ENSPATH).c_str());
```
valentin-krasontovitsch, line 451: doesn't it make more sense to add the `fs::absolute` in line 262, in `model_config_alloc_empty`? that seems to be where the inconsitent value `storage` is written.

valentin-krasontovitsch (changes requested):
LGTM besides the location of adding the absolute path

src/clib/lib/enkf/model_config.cpp
DanSava: Good point, initial solution required access to the `config_content` object until I realized I don't really need it :)

valentin-krasontovitsch (approved):

valentin-krasontovitsch:
:ship: :it:

PR #3999, opened 2022-10-05
Setup log configuration file for job_dispatch logger

### LINKED ISSUE #3998: Job_dispatch logger is not assigned a handler
When running `job_dispatch` we have multiple log statements, however we have not setup a streamhandler for the logger, hence it will by default be sent to stdout (`void` in the case of ert..).

Add a filehandler for the logger on a per realization basis.

lars-petter-hauge:
By default python logging will go to stdout, which for calling job_dispatch will go into the void. We specify a logfile that currently we want placed into runpath (that is, one job_runner log file per realization).

Considering that we want to remove the abstraction of a runpath for the user we might want to reconsider where these logs eventually are placed.

**Issue**
Resolves #3998

**Approach**
Add logging configuration.

## Pre review checklist


xjules:
> Getting error on type checking, but cannot recreate locally (updated mypy and friends).
> 
> ```
> Run mypy src/ert --config-file .mypy-strict.ini
> [7](https://github.com/equinor/ert/actions/runs/3189642920/jobs/5203687522#step:7:8)
> src/ert/logging/__init__.py:12: error: Function is missing a type annotation for one or more arguments
> ```
> 
> It's missing type annotations on `*args` and `*kwargs`, which I'm not certain what should be. Wondered if perhaps `*kwargs` could be `dict[str, Any]`. Funny thing is, my IDE already states this without type annotation, and if I add type annotation it says `dict[str, dict[str, Any]]`, so apparently vscode already has some form of detection on positional and keywords arguments.

When I check our code we tend to use `Any` for both (which you did).

src/ert/logging/job_runner_log.conf
```diff
@@ -0,0 +1,20 @@
+version: 1
+disable_existing_loggers: false
```
lars-petter-hauge, line 2: The failing test in libres tried to assert the content of `caplog`. When running the test individually everything passed, however, if the `test_job_dispatch.py` was run in advance, later tests failed. This was observed even with a trivial test function that shouldn't do anything at all:

`pytest tests/libres_tests/job_runner/test_job_dispatch.py::test_no_jobs_json_file tests/libres_tests/res/enkf/test_load_forward_model.py`

The problem with the failing test in libres was that when we are setting the logging config, the existing loggers are disabled. This is not a problem when running ert as it is expected to (in its own process), but in the tests we run the function from `job_dispatch.py` individually, and then disabling the already configured loggers..

https://docs.python.org/3/library/logging.config.html#logging.config.fileConfig

> disable_existing_loggers – If specified as False, loggers which exist when this call is made are left enabled. 
> The default is True because this enables old behaviour in a backward-compatible way.

tests/libres_tests/job_runner/test_job_dispatch.py
```diff
@@ -269,10 +270,12 @@ def test_job_dispatch_kills_itself_after_unsuccessful_job(unused_tcp_port):
     host = "localhost"
     port = unused_tcp_port
     jobs_json = json.dumps({"ens_id": "_id_", "dispatch_url": f"ws://localhost:{port}"})
+    with open(JOBS_FILE, "w") as fh:
+        fh.write(jobs_json)
```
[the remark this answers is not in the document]
lars-petter-hauge: irrelevant

src/ert/logging/job_runner_log.conf
lars-petter-hauge: No longer relevant after refactor. Interesting find though..

xjules (approved):
:rocket:

PR #4001, opened 2022-10-05
Make sure that processing reporter.event_queue does not yield ConnectionError

xjules:
**Issue**
Make sure that processing reporter.event_queue does not yield `ConnectionError`.

**Approach**
We don't yield ConnectionError, but instead keep processing next events in the queue. If the `Finish` event is received, we setup the timer (60 seconds), which when reached will automatically stop the publisher thread and exit. 

## Pre review checklist

### DIFF
--- a/src/_ert_job_runner/reporting/event.py
+++ b/src/_ert_job_runner/reporting/event.py
@@ -1,3 +1,4 @@
+import datetime
 import logging
 import queue
 import threading
@@ -6,6 +7,7 @@ from typing import Any, Dict
 
 from cloudevents.conversion import to_json
 from cloudevents.http import CloudEvent
+from websockets.exceptions import ConnectionClosedError
 
 from _ert_job_runner.client import Client
 from _ert_job_runner.reporting.base import Reporter
@@ -53,15 +55,34 @@ class Event(Reporter):
         self._step_id = None
         self._event_queue = queue.Queue()
         self._event_publisher_thread = threading.Thread(target=self._publish_event)
+        self._sentinel = object()  # notifying the queue's ended
+        self._timeout_timestamp = None
+        self._timestamp_lock = threading.Lock()
+        # seconds to timeout the reporter the thread after Finish() was received
+        self._reporter_timeout = 60
 
     def _publish_event(self):
         logger.debug("Publishing event.")
-        with Client(self._evaluator_url, self._token, self._cert) as client:
+        with Client(
+            url=self._evaluator_url,
+            token=self._token,
+            cert=self._cert,
+        ) as client:
             while True:
+                with self._timestamp_lock:
+                    if (
+                        self._timeout_timestamp is not None
+                        and datetime.datetime.now() > self._timeout_timestamp
+                    ):
+                        self._timeout_timestamp = None
+                        break
                 event = self._event_queue.get()
-                if event is None:
-                    return
-                client.send(to_json(event).decode())
+                if event is self._sentinel:
+                    break
+                try:
+                    client.send(to_json(event).decode())
+                except (ConnectionError, ConnectionClosedError):
+                    pass
 
     def report(self, msg):
         self._statemachine.transition(msg)
@@ -140,5 +161,9 @@ class Event(Reporter):
             )
 
     def _finished_handler(self, msg):
-        self._event_queue.put(None)
+        self._event_queue.put(self._sentinel)
+        with self._timestamp_lock:
+            self._timeout_timestamp = datetime.datetime.now() + datetime.timedelta(
+                seconds=self._reporter_timeout
+            )
         self._event_publisher_thread.join()
--- a/tests/unit_tests/c_wrappers/job_runner/test_event_reporter.py
+++ b/tests/unit_tests/c_wrappers/job_runner/test_event_reporter.py
@@ -1,7 +1,10 @@
 import json
 import os
+import time
+from unittest.mock import patch
 
 import pytest
+from websockets.exceptions import ConnectionClosedError
 
 from _ert_job_runner.job import Job
 from _ert_job_runner.reporting import Event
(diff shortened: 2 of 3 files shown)


src/ert/job_runner/reporting/event.py
```diff
@@ -140,5 +141,6 @@ def _job_handler(self, msg: Message):
             )
 
     def _finished_handler(self, msg):
+        self._event_queue.join()
```
lars-petter-hauge, line 144: Could I come with a suggestion? (inspired by https://stackoverflow.com/questions/38484535/join-for-possible-empty-queue-in-python). Add the `None` object prior to `join` and also do a `task_done` when event is `None`. In this scenario we are doing a `join`, then adding more to the `queue`, which seems counter-intuitive.

(the stack overflow also had a point on using an `Object()` as a sentinel instead of instead of `None`, which prevents the chance of a fake positive)

lars-petter-hauge:
What do you think?: `def _publish_event(self) -> def _event_publisher(self)`

As this is a function that is designed to continuously publish events.

lars-petter-hauge:
Thanks for the PR, just had some small comments/questions :)

lars-petter-hauge:
Could you please rephrase the commit message a bit? ("but" -> "by"?, should also be less than 80 chars or something, otherwise it's wrapped..)

src/ert/job_runner/reporting/event.py
```diff
@@ -53,15 +53,20 @@ def __init__(self, evaluator_url, token=None, cert_path=None):
...
                 event = self._event_queue.get()
-                if event is None:
+                if event is self._sentinel:
+                    self._event_queue.task_done()
                     return
-                client.send(to_json(event).decode())
+                try:
+                    client.send(to_json(event).decode())
+                finally:
+                    self._event_queue.task_done()
```
lars-petter-hauge, line 69: We haven't completely decided on the behaviour of sending events... But some comments:

Here the exception will be thrown after the `finally`, hence for any remaining events in the queue, they will not be completed with `task_done`. This was an issue before you had the `finally` anyway, so nice that we look over it. Could be tested by running the `job_dispatch` without setting up any server. (should be practical to make as a test as well, just make sure that we don't have to wait for the entire backoff time before it fails. Possible to set number of retries and time length on the `Client`).

Please do a check on the expectation from ert if a job never sends its finished event, will the jobqueuemanager fail on `MAX_RUNTIME` for a single job?

xjules: We've abandoned marking queue task_done since we might never finish up the queue processing. Current implementation just looses events if no connection's been made and such automatically closes, when the `Finish` event was received.

xjules:
> What do you think?: `def _publish_event(self) -> def _event_publisher(self)`
> 
> As this is a function that is designed to continuously publish events. Not a biggie, just a small suggestion..

We might do the change in another PR as it touches all reporters

src/_ert_job_runner/reporting/event.py
```diff
@@ -53,15 +55,32 @@ def __init__(self, evaluator_url, token=None, cert_path=None):
...
-        with Client(self._evaluator_url, self._token, self._cert) as client:
+        with Client(
+            url=self._evaluator_url,
+            token=self._token,
+            cert=self._cert,
+        ) as client:
             while True:
+                with self._timestamp_lock:
+                    if self._stop_timestamp is not None:
+                        if datetime.datetime.now() > self._stop_timestamp:
```
berland, line 74: Good annotate-python-linting suggestion to use a single if-statement.

src/_ert_job_runner/reporting/event.py
```diff
@@ -53,15 +55,32 @@ def __init__(self, evaluator_url, token=None, cert_path=None):
...
+                            self._stop_timestamp = None
+                            break
                 event = self._event_queue.get()
-                if event is None:
-                    return
-                client.send(to_json(event).decode())
+                try:
+                    if event is self._sentinel:
```
berland, line 79: I suppose this `if` can be moved out (i.e. up) of the `try` block.

src/_ert_job_runner/reporting/event.py
```diff
@@ -53,15 +55,32 @@ def __init__(self, evaluator_url, token=None, cert_path=None):
         self._step_id = None
         self._event_queue = queue.Queue()
         self._event_publisher_thread = threading.Thread(target=self._publish_event)
+        self._sentinel = object()  # notifying the queue's ended
+        self._stop_timestamp = None
```
berland, line 59: `.replace("stop", "timeout")` ?

tests/unit_tests/c_wrappers/job_runner/test_event_reporter.py
```diff
@@ -153,3 +156,57 @@ def test_report_inconsistent_events(unused_tcp_port):
             match=r"Illegal transition None -> \(MessageType<Finish>,\)",
         ):
             reporter.report(Finish())
+
+
+def test_report_with_failed_reporter_but_finished_jobs(unused_tcp_port):
+    mock_send_retry_time = 2
```
berland, line 162: Add a comment that what we test is that the reporter is not hanging forever (assuming the ert code does a similiar `join`) when the `send` deliberately fails.

tests/unit_tests/c_wrappers/job_runner/test_event_reporter.py
```diff
@@ -153,3 +156,57 @@ def test_report_inconsistent_events(unused_tcp_port):
...
+        # set _stop_timestamp to None only when timer stopped
+        assert reporter._stop_timestamp is None
+    assert len(lines) == 0, "expected 0 Job running messages"
+
+
+def test_report_with_reconnected_reporter_but_finished_jobs(unused_tcp_port):
+    mock_send_retry_time = 3
+
+    def mock_send(msg):
+        time.sleep(mock_send_retry_time)
```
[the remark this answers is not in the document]
xjules: Yes, it is to "simulate" reconnect efforts; ie. therefore the sleep time.

tests/unit_tests/c_wrappers/job_runner/test_event_reporter.py
```diff
@@ -153,3 +156,57 @@ def test_report_inconsistent_events(unused_tcp_port):
             match=r"Illegal transition None -> \(MessageType<Finish>,\)",
         ):
             reporter.report(Finish())
+
+
+def test_report_with_failed_reporter_but_finished_jobs(unused_tcp_port):
+    mock_send_retry_time = 2
+
+    def mock_send(msg):
+        time.sleep(mock_send_retry_time)
```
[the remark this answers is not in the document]
xjules: Yes, it will hang due to calling `join()`, since the thread has finished during the processing of events. Just in case I added a test if the thread is still alive.

berland (approved):
Lets go!

sondreso:
Should this be backported to version 4.1?

PR #4019, opened 2022-10-07
Change default runpath

### LINKED ISSUE #3600: Change default runpath
The current default runpath is "simulations/realization%d" and is defined here:

https://github.com/equinor/ert/blob/5021243874589b99502cc8295d27c627adf87e64/libres/lib/include/ert/enkf/enkf_defaults.hpp#L99

This default has a good chance of having to be overriden since it isn't different per iteration. Suggest changing it to `"simulations/realization<IENS>/iteration<ITER>"` which also does away with the anonymous "%d" formatting.

eivindjahren:
**Issue**
Resolves #3600

**Approach**

The default runpath did not create separate directories for each iteration, which create surprising behavior if e.g. realizations fail.

It is still possible to keep the behavior of the previous default runpath by explicitly setting it in the config file ie.

    RUNPATH simulations/realization%d

## Pre review checklist

### DIFF
--- a/src/clib/lib/include/ert/enkf/enkf_defaults.hpp
+++ b/src/clib/lib/include/ert/enkf/enkf_defaults.hpp
@@ -96,7 +96,7 @@
 #define DEFAULT_ITER_RETRY_COUNT 4
 
 /* Default directories. */
-#define DEFAULT_RUNPATH "simulations/realization%d"
+#define DEFAULT_RUNPATH "simulations/realization-<IENS>/iter-<ITER>"
 #define DEFAULT_ENSPATH "storage"
 
 #define SUMMARY_KEY_JOIN_STRING ":"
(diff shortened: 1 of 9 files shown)


src/clib/lib/include/ert/enkf/enkf_defaults.hpp
```diff
@@ -96,7 +96,7 @@
 #define DEFAULT_ITER_RETRY_COUNT 4
 
 /* Default directories. */
-#define DEFAULT_RUNPATH "simulations/realization%d"
+#define DEFAULT_RUNPATH "simulations/realization<IENS>/iteration<ITER>"
```
berland, line 99: The FMU standard would be `"simulations/realization-<IENS>/iter-<ITER>"`. I suggest we change to that to avoid confusion when we have the opportunity.

DanSava (commented):
There is a mention of `"simulations/realization%d"` on ln 115 in `src/ert/_c_wrappers/enkf/model_config.py` that might also need to be updated

tests/libres_tests/res/enkf/config/test_gen_kw_config.py
```diff
@@ -185,5 +185,5 @@ def test_gen_kw_is_log_or_not(tmpdir, distribution, expect_log, parameters_regex
         ert.createRunPath(ert.create_ensemble_experiment_run_context(0))
         assert re.match(
             parameters_regex,
-            Path("simulations/realization0/parameters.txt").read_text(),
+            Path("simulations/realization-0/iteration-0/parameters.txt").read_text(),
```
berland, line 188: one left?

berland (approved):
👍🏻

PR #4051, opened 2022-10-12
Remove libecl-style "type-safety"

pinkwah:
Old `clib` C classes provide `_safe_cast`, `_safe_cast_const` and `_is_instance` functions that take a `void *` as input and cast it to the correct function, or `util_abort` if the types are incorrect. This works because old-style structs start with a `int __type_id` (`UTIL_TYPE_ID_DECLARATION` macro), and this number is the magic `_TYPE_ID` for that type. Eg. `enkf_fs_type`'s TYPE_ID is `1089763` -- a unique arbitrary number chosen when the type was first created. If the `__type_id` matches the given TYPE_ID for a type, then the `_safe_cast` functions pass. 

There are multiple issues with this:
1. There is no guarantee that the `void *` object starts with `int __type_id`. Running `enkf_fs_safe_cast("foo")` where the type is `const char *` is undefined behaviour. So while the code may compile for a lot of pointers, it might not be a safe operation (ie. "safe cast" isn't actually safe).
2. C++ is a strongly-typed language. Unlike C, there are no implicit pointer casts (except _to_ `void *`. Notably this is why you need to cast `malloc()` in C++ but not C). We should use the type system correctly and resolve any type issues at compile-time rather than run-time. "Safe" casting consumes CPU cycles when the correct solution would not.
3. libecl's types use type-erasure to make things behave in C. For example, `vector_type` cannot know which type it contains, so it must rely on `void *` to do the job. Therefore, a lot of interaction with libecl's API is using `void *`. We are moving away from this approach, instead relying on C++ templated classes which do preserve the type.


src/clib/lib/include/ert/enkf/summary_key_set.hpp
```diff
@@ -1,8 +1,6 @@
 #ifndef ERT_SUMMARY_KEY_SET_H
 #define ERT_SUMMARY_KEY_SET_H
 
-#include <ert/util/type_macros.h>
-
 #include <ert/enkf/enkf_types.hpp>
```
eivindjahren, line 4: aparently stringlist_type was transiently imported from type_macros here.

eivindjahren (approved):
Definitely better! The diff is so big I this is going to be war :boom:, good luck with your deployment soldier :vulcan_salute: !

eivindjahren:
`old_tests/res_util/test_ert_util_ui_return.cpp` seems to not find `ui_return.hpp`, and after looking into it, that test neither compiled nor ran.

Issue #4048, opened 2022-10-12
Simplify EnsembleConfig


valentin-krasontovitsch:

valentin-krasontovitsch:
to summarize a bit what has been done so far - we moved grid and refcase (files) from ecl config to ensemble config, where it's actually used.

in the course of that work, we saw that we did not have new config parser style tests for ensemble config, and started working on creating those, in particular extending the config dict generator to create options and option arguments for ensemble config.

we covered some keywords, but not all. the keywords that were covered, are

- `GEN_DATA`
- `FIELD`

the uncovered ones that I can see are

- `GEN_KW`
- `SURFACE`
- `SUMMARY`
- `SCHEDULE_PREDICTION_FILE`

with their respective arguments, which can be found (possibly partly) in the docs, and in the config dict code path of ensemble config, and of course in the C code. so the config dict generator should be extended to cover the remaining keywords.

what also happened was that for the covered keywords, we investigated which of the options were actually used, and discovered that some were unused, and removed them (and code associated to them).

finally, one might also look at possibly existing old tests to draw inspiration on what the options and arguments and their respective possible values are (we have not done this).

and last but not least, no work has been done yet to port C code to python, besides the mentioned removal of unsed code paths.

valentin-krasontovitsch:
instead of TODOs in the code, I'm gonna note two things here. In config dict generator, we should

- also cover the key `ConfigKeys.GEN_KW_TAG_FORMAT`
- we should investigate if `REPORT_STEPS` as an argument to `GEN_DATA` should be sorted (it's a list of ints, somehow used for reporting perhaps, and supposed to match something, cf. #4072 )

valentin-krasontovitsch:
gonna reopen this, as we want to have another go at lifting moar parts of ensemble config to python

PR #4057, opened 2022-10-13
Remove no longer used min std

valentin-krasontovitsch:
According to my investigation, this was only used for printing things - and Eivind discovered that the printing in question was not used, and conveniently removed it in 0adcb91c9841110a7e999ded930847404a466def

So we remove min std as well, as it's literally not used anymore at all.

Related to #4048

## Pre review checklist


src/clib/lib/include/ert/enkf/enkf_config_node.hpp
```diff
@@ -126,10 +118,10 @@ bool enkf_config_node_internalize(const enkf_config_node_type *node,
 /*
     The enkf_node_free() function declaration is in the enkf_config_node.h header,
     because the enkf_config_node needs to know how to free the min_std node.
+    That's not really the reason, because min_std doesn't exist anymore, but
```
DanSava, line 121: This comment is a bit unhelpful. But I don't really have a better suggestion than to move the function in the `enfk_node.h` file, which I am not sure it should be done as part of this PR. So I guess leave it as it is :man_shrugging:

DanSava (approved):
Very nice! :+1:

src/clib/lib/include/ert/enkf/enkf_config_node.hpp
valentin-krasontovitsch: thanks for looking at / into it!

PR #4067, opened 2022-10-14
Job runner yields Running event only when memory consumption has significant change

xjules:
Job runner yields `Running` event only when memory consumption has significant change. If there is a small change job runner yields `RunningNoMemChange` instead.

## Pre review checklist

### DIFF
--- a/src/_ert_job_runner/cli.py
+++ b/src/_ert_job_runner/cli.py
@@ -7,7 +7,7 @@ import sys
 import typing
 
 from _ert_job_runner import reporting
-from _ert_job_runner.reporting.message import Finish
+from _ert_job_runner.reporting.message import Finish, RunningNoMemChange
 from _ert_job_runner.runner import JobRunner
 
 JOBS_FILE = "jobs.json"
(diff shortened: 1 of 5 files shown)


lars-petter-hauge:
Thanks for the PR, will reduce traffic towards the server!

src/ert/job_runner/job.py
```diff
@@ -108,10 +114,13 @@ def run(self):
...
-            if memory > max_memory_usage:
-                max_memory_usage = memory
+            max_memory_usage = max(memory, max_memory_usage)
 
-            yield Running(self, max_memory_usage, memory)
+            if abs(memory - last_memory_usage) >= self.MEMORY_INCREASE_THRESHOLD:
+                last_memory_usage = memory
+                yield Running(self, max_memory_usage, memory)
+            else:
+                yield None
```
lars-petter-hauge, line 123: Why does it need to yield None? Can it not just continue?

xjules: There are at least two reasons:
1. We sort of keep it alive (while yielding control to the primary `for loop` in `cli.py`)
2. Makes it much easier to test since we get to know that there was no change as it yields None. Otherwise, it becomes a blocker.

lars-petter-hauge (approved):
Alright, let's get this in

PR #4074, opened 2022-10-14
Add pip packages to PATH for macos-latest CI job

valentin-krasontovitsch:
conan install on macos latest with system pip and root without pyvenv or anything does not land in PATH, which leads to trouble

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ['ubuntu-latest', 'macos-10.15']
+        os: ['ubuntu-latest', 'macos-latest']
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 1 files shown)


src/ert/_c_wrappers/enkf/hook_workflow.py
```diff
@@ -36,3 +36,6 @@ def __ne__(self, other):
 
     def __repr__(self):
         return f"HookWorkflow({self.getWorkflow().src_file}, {self.getRunMode()})"
+
+    def free(self):
```
andreas-el, line 40: I've gone over the other implementations and also came to the conclusion that only these three are missing free() override. As far as i understand, using pass is also the way to go here, since we have no underlying content to actually free/release. :+1:

valentin-krasontovitsch: right. one of them was an actual free, the hook workflow, but that failed a bunch of tests with ghastly "you're trying to break the universe" kind of errors, so i removed it ^^ i should have read the code more carefully before just adding that one, i guess :sweat_smile:

valentin-krasontovitsch:
jenkins, retest this please

andreas-el (approved):
LGTM :suspect:

PR #4076, opened 2022-10-14
Move ert.job_runner to _ert_job_runner

eivindjahren:
Importing ert has significant overhead. Therefore we need to make sure job_runner.py does not do that. By avoiding this, the overhead of running job_runner.py goes from 2 seconds to 0.1 seconds. This seems to make the tests run 30 % faster as a consequence because it spawns job_dispatch.py many times.


mortalisk:
fix ya style

mortalisk:
Looks beautiful. The only concern is that of @sondreso regarding having two top level packages. Perhaps he would like to chime in here?

sondreso:
My concern was that there was some workaround required for having multiple top-level packages with skbuild, but if this is OK I don't have any more concerns :sweat_smile:

mortalisk (approved):

PR #4056, opened 2022-10-13
Refactor num cpu

valentin-krasontovitsch:
**Issue**
Closes #4005

## Pre review checklist

- [x] Added appropriate release note label
- [x] PR title captures the intent of the changes, and is fitting for release notes.
- [ ] Commit history is consistent and clean, in line with the [contribution guidelines](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md).

Adding labels helps the maintainers when writing release notes. This is the [list of release note labels](https://github.com/equinor/ert/labels?q=release-notes).

PR #4100, opened 2022-10-19
Move Eclipse Grid and Refcase to EnsembleConfig

### LINKED ISSUE #4035: Remove EclConfig
It seems like we might be able to remove ecl config altogether, after all the dust has settled.

At the time of writing, it holds three data:

- num cpu
- grid (and its file)
- refcase (and its file)

the files are only used to load grid and refcase.
there is a PR ready for review right now which removes num cpu from ecl config.
investigating uses of grid and refcase shows that, besides tests:

ecl config's grid is
- passed to ensemble config in res config's init,
- used to construct observations in enkfmain, and
- mentioned in a comment about row scaling (which i sadly don't really understand)

while ecl config's refcase is
- also used to construct observations in enkfmain, and
- attached to the time map in storage if defined, and
- used to extract data in libres facade
- passed on to both ensemble and model config in res config's init

my idea is

- to have the grid live only in ensemble config, and passed from there to construct observations in enkf main, and
- to have refcase reside in ensemble config~~res config~~, and passed down from there to ~~ensemble and~~ model config _as a boolean, as per its use described in a comment below_.

~~but for the second point, i shall investigate usage of refcase in both ensemble and model config, and see if any of them can actually do without, and then adjust the plan accordingly.~~

andreas-el:
Resolves #4035 

Moved files and filenames for grid and refcase from Ecl config to EnsembleConfig.
Copies existing tests that cover grid and refcase contents.
Adds repr and eq functionality for comparisons.

Once this passes, and refactor num_cpu https://github.com/equinor/ert/pull/4056/, we can consider removing ecl_config altogether.

## Pre review checklist

### DIFF
--- a/tests/test_config_parsing/config_dict_generator.py
+++ b/tests/test_config_parsing/config_dict_generator.py
@@ -7,6 +7,7 @@ import hypothesis.strategies as st
 from hypothesis import assume
 
 from ert._c_wrappers.enkf import ConfigKeys
+from ert._c_wrappers.enkf.enums import GenDataFileType
 from ert._c_wrappers.job_queue import QueueDriverEnum
 from ert._c_wrappers.util.enums import MessageLevelEnum
 
@@ -30,16 +31,37 @@ def define_keys(draw):
 
 
 def touch(filename):
-    with open(filename, "w") as fh:
+    with open(file=filename, mode="w", encoding="utf-8") as fh:
         fh.write(" ")
 
 
 file_names = words
+format_file_names = st.builds(lambda file_name: file_name + "-%d", file_names)
 
 
 @st.composite
 def directory_names(draw):
     return "dir" + draw(words)
+
+
+transforms = st.sampled_from(
+    [
+        "DENORMALIZE_PERMX",
+        "NORMALIZE_PORO",
+        "LN",
+        "LOG10",
+        "EXP0",
+        "LOG",
+        "EXP",
+        "TRUNC_POW10",
+        "LN0",
+        "DENORMALIZE_PORO",
+        "NORMALIZE_PERMZ",
+        "NORMALIZE_PERMX",
+        "DENORMALIZE_PERMZ",
+        "POW10",
+    ]
+)
 
 
 small_floats = st.floats(min_value=1.0, max_value=10.0, allow_nan=False)
@@ -74,6 +96,7 @@ def queue_name(queue_driver):
         return "TORQUE"
     if queue_driver == QueueDriverEnum.SLURM_DRIVER:
         return "SLURM"
+    raise ValueError(f"unexpected driver `{queue_driver}`")
 
 
 def valid_queue_options(queue_system):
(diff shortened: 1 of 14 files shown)


tests/test_config_parsing/config_dict_generator.py
```diff
@@ -310,15 +367,79 @@ def config_dicts(draw):
     return config_dict
 
 
-def to_config_file(filename, config_dict):
-    with open(filename, "w+") as config:
+def to_config_file(filename, config_dict):  # pylint: disable=too-many-branches
```
eivindjahren, line 370: This function seems to perhaps have outlived being just one if/elif chain. My solution here would be to use a separate structure which we turn into this ConfigKeys. This we could keep in the tests for now but eventually turn into a intermediate representation for the parser.

What I am thinking is to have a 

```python
@dataclass
def Field:
   name: str = ""
   var_type: str = ""
   ...
   
   def to_file_contents(self) -> str:
     return " ".join ( ["FIELD", ...]
```

so that `to_config_file` becomes

```python
 to_config_file(file_name, config_ast):
     with open(file_name) as f:
         for keyword in config_ast:
             f.write(keyword.to_file_contents())
 ```

File: tests/unit_tests/c_wrappers/res/enkf/test_ensemble_config.py [hunk 2]
```diff
@@ -84,5 +84,4 @@ def test_ensemble_config_constructor(setup_case):
                 }
             ],
         },
-        grid=res_config.ensemble_config.grid,
```
eivindjahren: This change seems to be in the wrong commit

eivindjahren (changes requested):
Need to remove WIP from commit message and squash the 3 commits that remove ecl_config, add grid and refcase to ensemble config and changes eq/repr of ensemble_config.

tests/unit_tests/c_wrappers/res/enkf/test_ensemble_config.py
andreas-el: A slip up! Thanks! :wrench:

tests/test_config_parsing/config_dict_generator.py
andreas-el: Thanks.
- Removed WIP from commit message
- Squashed mentioned commits
- Split lint commit and fixup's where applicable

tests/test_config_parsing/config_dict_generator.py
```diff
@@ -30,18 +31,39 @@ def define_keys(draw):
...
 file_names = words
+format_file_names = st.builds(lambda file_name: file_name + "-%d", file_names)
 
 
 @st.composite
 def directory_names(draw):
     return "dir" + draw(words)
 
 
+functions = st.sampled_from(
```
eivindjahren, line 47: Perhaps call these `transforms` instead of `functions` as `functions` sounds like you are making python functions.

eivindjahren:
Sorry, only one more small comment, then I think we are ready to go :)

tests/test_config_parsing/config_dict_generator.py
andreas-el: I have no objections to this. :+1:

eivindjahren (approved):
:rocket:

andreas-el:
Jenkins test this please!

valentin-krasontovitsch (approved):
looks good!

andreas-el:
Jenkins test this please!

andreas-el:
Jenkins retest this please!

andreas-el:
Jenkins retest this please!

PR #4113, opened 2022-10-21
Show restart button when all realisations fail

### LINKED ISSUE #4112: Restart button not showing when all realisations fail
**Describe the bug**
Restart button not showing when all realisations fail

**To reproduce**
1. `pip install ert`
2. Add sys.exit(1) to _evaluate() in poly_eval.py in poly_example. This will cause all realisations to fail.
3. Run ensemble experiment
4. Check that no restart button is visible in run dialog

**Expected behaviour**
Restart button should be visible when all realisations fail.

**Screenshots**
Screenshot 2022-10-21 at 10 00 10

**Environment**
 - OS: macOS
 - ERT/Komodo release: 2022.09.03
 - Python version 3.8
 - Remote/HPC execution involved: [yes|no] no

dafeda:
**Issue**
Resolves #4112

## Pre review checklist

### DIFF
--- a/src/ert/shared/models/base_run_model.py
+++ b/src/ert/shared/models/base_run_model.py
@@ -138,12 +138,16 @@ class BaseRunModel:
         Creates a list of bools representing the failed realizations,
         i.e., a realization that has failed is assigned a True value.
         """
-        return [
-            initial and not completed
-            for initial, completed in zip(
-                self._initial_realizations_mask, self._completed_realizations_mask
-            )
-        ]
+        if self._completed_realizations_mask:
+            return [
+                initial and not completed
+                for initial, completed in zip(
+                    self._initial_realizations_mask, self._completed_realizations_mask
+                )
+            ]
+        else:
+            # If all realisations fail
+            return [True] * len(self._initial_realizations_mask)
 
     def _count_successful_realizations(self) -> int:
         """
--- a/tests/unit_tests/models/test_base_run_model.py
+++ b/tests/unit_tests/models/test_base_run_model.py
@@ -68,6 +68,7 @@ def test_active_realizations(initials, expected):
         ([False, False], [False, False], False, [False, False]),
         ([False, False], [True, True], False, [False, False]),
         ([True, True], [False, True], True, [True, False]),
+        ([False, False], [], True, [True, True]),
     ],
 )
 def test_failed_realizations(initials, completed, any_failed, failures):


src/ert/shared/models/base_run_model.py
```diff
@@ -138,12 +138,16 @@ def _create_mask_from_failed_realizations(self) -> List[bool]:
         Creates a list of bools representing the failed realizations,
         i.e., a realization that has failed is assigned a True value.
         """
-        return [
-            initial and not completed
-            for initial, completed in zip(
-                self._initial_realizations_mask, self._completed_realizations_mask
-            )
-        ]
+        if len(self._completed_realizations_mask) == 0:
```
berland, line 141: Can we change this to `if self._completed_realizations_mask is None` instead, and then have this attribute `None` from object initialization and also set to `None` instead of `[]` in the case of an `ErtRunError` (which I speculate is happening here?)

berland: If possible (meaning that the ensemble length is always known) we can also change it to always have the correct length (but it is maybe needed to allow `None` at init?)

dafeda: > Can we change this to `if self._completed_realizations_mask is None` instead, and then have this attribute `None` from object initialization and also set to `None` instead of `[]` in the case of an `ErtRunError` (which I speculate is happening here?)

We would then also have to change the input argument to BaseRunModel to Optional[List[bool]] = None right?
Not sure we gain that much from using None instead of [] here.

berland: Right. A zero-length ensemble is always an error in Ert (?) so `[]` is "free" to be used equivalent to `None`. It might be cleaner still to use `None` as `[]` could be thought of as ok if one believes zero-length ensembles is a thing.

dafeda: So, I tried running poly.ert with 0 realisations and when trying to run an ensemble experiment get this crash:

ValueError: Wrong range syntax 0--1

😬

berland: 👍🏻

dafeda: Using None would mean that we also have to change `_count_successful_realizations` I think, since it's using `.count`.
Think we'll just leave as is and use [].

berland: that is ok!

berland: (but pylint frowns upon `if len(list) == 0`, do `if list` instead.)

dafeda: Pylint does not like bare len(list), explicit checks like len(list)==0 is fine according to pep8.
This was considered a bug in pylint:
https://github.com/PyCQA/pylint/pull/2815

berland: ah, but pep8 still says that `if not list` is good in this situation?

dafeda: I'm fine with either as long as the bug is fixed.
I've pushed a version that uses `if list`.

berland (approved):
💯

PR #4117, opened 2022-10-21
Remove unused qsub_emulators

berland:
This is better solved in test_job_queue_manager_torque.py

Drive-by cleanup.


berland:
qsub is used after all. Will reconsider.

PR #4131, opened 2022-10-26
[WIP] Purge ecl config

### LINKED ISSUE #4035: Remove EclConfig

andreas-el:
**Issue**
Resolves #4035 

**Approach**

## Pre review checklist


src/ert/_c_wrappers/enkf/row_scaling.py
```diff
@@ -25,7 +25,7 @@ def assign(self, target_size, func):
         >>> ens_config = ert.ensembleConfig()
         >>> config_node = ens_config["PORO"]
         >>> field_config = config.node.getFieldModelConfig()
-        >>> grid = ert.eclConfig().grid
+        >>> grid = ert.ensembleConfig().grid
```
valentin-krasontovitsch, line 28: haha good on you for fixing this one - i still don't know / understand what it does ^^

tests/unit_tests/c_wrappers/res/enkf/test_res_config.py
```diff
@@ -259,24 +259,17 @@ def test_extensive_config(setup_case):
...
+        == Path(ensemble_config._grid_file).resolve()
     )
 
-    ensemble_config = res_config.ensemble_config
-    assert set(
-        snake_oil_structure_config["SUMMARY"]
-        + snake_oil_structure_config["GEN_KW"]
-        + snake_oil_structure_config["GEN_DATA"]
-    ) == set(ensemble_config.alloc_keylist())
-
```
valentin-krasontovitsch, line 279: how come this block was removed? sorry, it's not immediately clear to me

valentin-krasontovitsch:
i have not gotten much information about the deeper meaning behind the names of the relase labels, besides that every PR should have one (otherwise they get the default `misc`). also, it doesn't seem to be very enforced to put a label, at least in my experience. in this case, i would agree and also put maintenance.

i don't think you missed anything! after all, moft of the clean up was already done in previous PRs, so this should be relatively pain free : )

besides cleaning up the history, i think you're good to go ^^

valentin-krasontovitsch (approved):
very good! delete all the code!

tests/unit_tests/c_wrappers/res/enkf/test_res_config.py
andreas-el: I thought this was introduced when rebasing on main. After some investigation, i see this is not the case.
I've reverted this snippet since this is not my intended change.

src/ert/_c_wrappers/enkf/row_scaling.py
andreas-el: This is pseudo only, but it could be even more accurate by accessing the ensembleConfig via ens_config instead. Thanks for raising awareness here.

valentin-krasontovitsch: oh OK i really just find that comment block puzzling / amusing, since I don't understand it, but good that my silliness actually led to improvement : )

andreas-el:
Jenkins retest this please!

Issue #4073, opened 2022-10-14
Remove remaining AnalysisConfig C implementation


DanSava:
Blocked by #3987 

**Issue**

Some logic still keeps track of the `AnalysisModules` that can be lifted from the `AnalysisConfig` C object into python. Once that is done we should be able to remove altogether the `AnalysisConfig` C object.

PR #4133, opened 2022-10-26
Move C implementation for ies_config analysis_config and analysis_module in python

DanSava:
**Issue**
Resolves https://github.com/equinor/ert/issues/4073

## Pre review checklist


DanSava:
I had to close PR #4103 because it was not getting updated when pushing on the forked branch

src/ert/_c_wrappers/analysis/analysis_module.py
```diff
@@ -1,162 +1,230 @@
...
+        if var_name in self.SPECIAL_KEYS:
+            self.handle_special_key_set(var_name, value)
+        elif var_name in self._variables:
+            var = self._variables[var_name]
+            try:
+                new_value = var["type"](value)
+                if new_value > var["max"]:
+                    var["value"] = var["max"]
+                    logger.warning(
+                        f"New value {new_value} for key"
```
oysteoh, line 163: maybe also in these messages state what is currently used as min/max thresholds?

src/ert/_c_wrappers/enkf/analysis_config.py
```diff
@@ -61,42 +42,46 @@ def __init__(
...
+        analysis_copy = analysis_copy or []
+        es_module = AnalysisModule.ens_smoother_module()
+        ies_module = AnalysisModule.iterated_ens_smoother_module()
+        self._modules: Dict[str, AnalysisModule] = {
+            AnalysisMode.ENSEMBLE_SMOOTHER: es_module,
+            AnalysisMode.ITERATED_ENSEMBLE_SMOOTHER: ies_module,
+        }
+        self._active_module = analysis_select or AnalysisMode.ENSEMBLE_SMOOTHER
+        # copy modules
+        for element in analysis_copy:
```
oysteoh, line 55: have you considered instead of creating an comment explaining the functionality to extract a private func `_copy_modules()` ? Not saying it is better, but sometimes this could be beneficial keeping the init more clean...

src/ert/_c_wrappers/enkf/analysis_config.py
```diff
@@ -61,42 +42,46 @@ def __init__(
...
-                    var_name = set_var[ConfigKeys.VAR_NAME]
-                    value = set_var[ConfigKeys.VALUE]
-
-                module = self._get_module(module_name)
-                module._set_var(var_name, str(value))
-
-            if self._analysis_select:
-                self._select_module(self._analysis_select)
+        # set var list
+        for set_var in analysis_set_var:
```
oysteoh, line 75: if previous comment is a good idea - maybe also here call a `_set_var_list()`  function?

src/ert/_c_wrappers/analysis/analysis_module.py
```diff
@@ -1,162 +1,230 @@
...
+        variables: Dict[str, "VariableInfo"],
+        iterable: bool,
+    ):
+        self.mode = mode
+        self.name = name
+        self.iterable = iterable
+        self._variables = variables
+
+    @classmethod
+    def ens_smoother_module(cls, name: AnalysisMode = AnalysisMode.ENSEMBLE_SMOOTHER):
```
eivindjahren, line 100: The name parameter seems to have incorrect type and default value here. Did you mean just `name:str` without a default value?

src/ert/_c_wrappers/analysis/analysis_module.py
```diff
@@ -1,162 +1,230 @@
...
+        return cls(
+            mode=AnalysisMode.ENSEMBLE_SMOOTHER,
+            name=name,
+            variables=get_mode_variables(AnalysisMode.ENSEMBLE_SMOOTHER),
+            iterable=False,
+        )
+
+    @classmethod
+    def iterated_ens_smoother_module(
+        cls, name: AnalysisMode = AnalysisMode.ITERATED_ENSEMBLE_SMOOTHER
```
eivindjahren, line 110: name seems to have incorrect type here, did you mean to have `str:name` without a default value?

src/ert/_c_wrappers/analysis/analysis_module.py
```diff
@@ -1,162 +1,230 @@
...
+                self.set_var("IES_INVERSION", inversion_str_map[value])
+            else:
+                logger.warning(
+                    f"Unknown value {value} used for INVERSION key"
+                    f"supported values are {list(inversion_str_map.keys())}"
+                )
+        elif var_name in self.TRUNC_ALTERNATE_KEYS:
+            self.set_var("ENKF_TRUNCATION", value)
+
+    def set_var(self, var_name: str, value: Union[float, int, bool, str]):
```
eivindjahren, line 153: It is possible to type this more succinctly with the overload

```python
from typing import overload

int_var_names = [v["name"] for v in ies_variables.values() if v["type"] == int]
float_var_names = [v["name"] for v in ies_variables.values() if v["type"] == float]

@overload
def set_var(self, var_name: Literal[int_var_names], value: int): ...
@overload
def set_var(self, var_name: Literal[float_var_names], value: float): ...
def set_var(self, var_name, value):
    if var_name in self.DEPRECATED_KEYS:
        logger.warning(...
```

eivindjahren (changes requested):
Generally amazing improvement! I had a few small comments that should be easy to fix.

src/ert/_c_wrappers/analysis/analysis_module.py
```diff
@@ -1,162 +1,230 @@
...
+        raise KeyError(f"Variable {name} not found in module {self.name}")
+
+    def variable_value_dict(self) -> Dict[str, Union[float, int]]:
+        return {name: var["value"] for name, var in self._variables.items()}
+
+    def handle_special_key_set(self, var_name, value):
+        if var_name in self.DEPRECATED_KEYS:
+            logger.warning(
+                f"{var_name} settings have been removed"
+                f"use the INVERSION setting instead"
```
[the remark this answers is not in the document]
valentin-krasontovitsch: I might be wrong, but it seems like all of those keys can / could be used to set the inversion. does that make sense / answer your question?

valentin-krasontovitsch (approved):
besides [this question in the previous PR](https://github.com/equinor/ert/pull/4103#discussion_r1005538288) I'm happy

DanSava:
> besides [this question in the previous PR](https://github.com/equinor/ert/pull/4103#discussion_r1005538288) I'm happy

So this is the question:

> right, exactly - and it doesn't happen to special keys though, since it's an elif, correct? is that OK? we don't need the checks there? because INVERSION, for instance, which is one of the special keys, has a min and max defined, so we could do the check?

Maybe I was a bit too short in my previous answer let's do the long answer:

Let's take the example for INVERSION which is what I call a "special key". If the user defines it in the config file or the config dictionary like:

```
INVERSION EXACT
or 
INVERSION SUBSPACE_EXACT_R
or 
INVERSION SUBSPACE_EE_R
or
INVERSIONSUBSPACE_RE
```
**Note**: the above options are the only allowed possibilities and the code checks for this
```python 
def handle_special_key_set(self, var_name, value):
 ...
        elif var_name == "INVERSION":
            inversion_str_map = {
                "EXACT": 0,
                "SUBSPACE_EXACT_R": 1,
                "SUBSPACE_EE_R": 2,
                "SUBSPACE_RE": 3,
            }
            if value in inversion_str_map:
                self.set_var("IES_INVERSION", inversion_str_map[value])
            else:
                logger.warning(
                    f"Unknown value {value} used for INVERSION key"
                    f"supported values are {list(inversion_str_map.keys())}"
                )
...
```
So if one of the above options is in the config file or dictionary then we will call 
```
     self.set_var("IES_INVERSION", inversion_str_map[value])
```
which has the min/max check inside, so we still do the min/max check even if we don't really need to because `INVERSION` is a special key that only allows us the set `IES_INVERSION` to be set to 0,1,2 or 3 that are in the min/max bound. 

So in conclusion even if we are dealing with special keys they are just another way of setting values for `IES_INVERSION` or `ENKF_TRUNCATION` and to set the values, in the end, we are still using `self.set_var(key_name, value)` which does the min/max check.

valentin-krasontovitsch:
aaahhhhh, thank you, now i got it : )

src/ert/_c_wrappers/analysis/analysis_module.py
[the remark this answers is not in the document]
DanSava: So a bit more context:

In the C code in analysis_module.cpp there is this snippet of code 
```
        if (strcmp(var_name, "USE_EE") == 0)
            logger->warning("The USE_EE/USE_GE settings have been removed - "
                            "use the INVERSION setting instead");

        if (strcmp(var_name, "USE_GE") == 0)
            logger->warning("The USE_EE/USE_GE settings have been removed - "
                            "use the INVERSION setting instead");
```
And I assumed the keys `USE_EE` and `USE_GE` are deprecated hence the strange message. Also, these two keys are the only ones in the `DEPRECATED_KEYS` list.  I wanted to remove these keys and the `DEPRECATED_KEYS` list but I did not want to break old config files that might still have these keys defined in them. 

> Is that something which is universal across whatever deprecated keys you have used previously?

No this is something special for this set of keys [USE_EE, USE_GE]

src/ert/_c_wrappers/analysis/analysis_module.py
DanSava: I am sorry my I do not understand how the suggestion is more succinct. 

`@overload` has never been used in the ERT codebase before, I don't really want to be the first to start adding it to the codebase.

I never used overloaded functions in python before so I find it a bit hard to understand what you are suggesting.

src/ert/_c_wrappers/analysis/analysis_module.py
eivindjahren: Ok, I read that incorrectly :D

src/ert/_c_wrappers/analysis/analysis_module.py
eivindjahren: We can leave that typing for now, and I can propose a change later

eivindjahren:
I think perhaps you had a rebase that went bad otherwise I have no further comments.

DanSava:
> I think perhaps you had a rebase that went bad (there are unrelated commits in the PR) otherwise I have no further comments.

I am not sure what I did wrong I did cherry-pick before I created this new PR and then rebased and then forced-pushed... anyways the result is that github is back at not updating after a push or force-push. I don't really know what to do except make a new PR

eivindjahren:
> > I think perhaps you had a rebase that went bad (there are unrelated commits in the PR) otherwise I have no further comments.
> 
> I am not sure what I did wrong I did cherry-pick before I created this new PR and then rebased and then forced-pushed... anyways the result is that github is back at not updating after a push or force-push. I don't really know what to do except make a new PR

First do a `git branch my_backup` to create a backup of what you have. Then do `git rebase -i HEAD~10` and pick the commits you want to be in the PR. You can likely resolve any conflict with `git checkout --theirs THE_FILE`.

eivindjahren (approved):
Looks good, just please squash the commits and write a good message and header :)

PR #4158, opened 2022-10-28
Fix config_content.as_dict()

### LINKED ISSUE #4148: Improve config_content.as_dict()
Handle the creation of the config_content dictionary in a better way so that we avoid having to do the following:

```python 
  max_runtime = config_dict.get(ConfigKeys.MAX_RUNTIME, 0)
  if isinstance(max_runtime, list):
      max_runtime = max_runtime[-1]
```
 Currently, any consumer of the config_content dict needs to account for this behaviour if the config KEY is defined multiple times in the file and we are only interested in the final value.

DanSava:
**Issue**
Resolves #4148

**Approach**
Add a list of single-value config keys for each of the keys in the list config_content.as_dict() will return a single value for the resulting dictionary item.

## Pre review checklist

### DIFF
--- a/src/ert/_c_wrappers/config/config_content.py
+++ b/src/ert/_c_wrappers/config/config_content.py
@@ -264,25 +264,3 @@ class ConfigContent(BaseCClass):
 
     def keys(self) -> List[str]:
         return list(self._keys())
-
-    def as_dict(self) -> Dict[str, List[Any]]:
-        # pylint: disable=consider-using-dict-items
-        # (false positive)
-        d: Dict[str, List[Any]] = {}
-        for key in self.keys():
-            item = self[key]
-            if len(item) > 1:
-                d[key] = []
-                for node in item:
-                    values = list(node)
-                    if len(values) > 1:
-                        d[key].append(values)
-                    else:
-                        d[key].append(values[0])
-            else:
-                values = list(item[0])
-                if len(values) > 1:
-                    d[key] = [values]
-                else:
-                    d[key] = values[0]
-        return d
--- a/src/ert/_c_wrappers/config/config_content.py
+++ b/src/ert/_c_wrappers/config/config_content.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Any, Dict, Iterator, List
+from typing import TYPE_CHECKING, Iterator, List
 
 from cwrap import BaseCClass
 
--- a/src/ert/_c_wrappers/enkf/analysis_config.py
+++ b/src/ert/_c_wrappers/enkf/analysis_config.py
@@ -101,9 +101,6 @@ class AnalysisConfig:
         if min_realization == 0:
             min_realization = num_realization
         min_realization = min(min_realization, num_realization)
-        max_runtime = config_dict.get(ConfigKeys.MAX_RUNTIME, 0)
-        if isinstance(max_runtime, list):
-            max_runtime = max_runtime[-1]
 
         config = cls(
             alpha=config_dict.get(ConfigKeys.ALPHA_KEY, 3.0),
@@ -112,7 +109,7 @@ class AnalysisConfig:
             std_cutoff=config_dict.get(ConfigKeys.STD_CUTOFF_KEY, 1e-6),
             stop_long_running=config_dict.get(ConfigKeys.STOP_LONG_RUNNING, False),
             global_std_scaling=config_dict.get(ConfigKeys.GLOBAL_STD_SCALING, 1.0),
-            max_runtime=max_runtime,
+            max_runtime=config_dict.get(ConfigKeys.MAX_RUNTIME, 0),
             min_realization=min_realization,
             update_log_path=config_dict.get(ConfigKeys.UPDATE_LOG_PATH, "update_log"),
             analysis_iter_config=AnalysisIterConfig.from_dict(config_dict),
--- a/tests/unit_tests/c_wrappers/res/config/test_config_content.py
+++ b/tests/unit_tests/c_wrappers/res/config/test_config_content.py
@@ -1,12 +1,8 @@
-from pathlib import Path
-
 from ecl.util.util import StringList
 
 from ert._c_wrappers.config.config_content import ConfigContent
 from ert._c_wrappers.config.config_parser import ConfigParser
 from ert._c_wrappers.config.content_type_enum import ContentTypeEnum
-from ert._c_wrappers.enkf.config_keys import ConfigKeys
-from ert._clib.config_keywords import init_user_config_parser
 
 
 def test_get_executable_list(tmpdir):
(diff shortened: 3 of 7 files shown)


src/ert/_c_wrappers/config/config_content.py
```diff
@@ -7,11 +7,52 @@
...
 
 
+SINGLE_VALUE_KEYS = [
+    ConfigKeys.ANALYSIS_SELECT,
+    ConfigKeys.CONFIG_DIRECTORY,
+    ConfigKeys.DATAROOT,
+    ConfigKeys.DATA_FILE,
+    ConfigKeys.ENSPATH,
+    ConfigKeys.RERUN_KEY,
+    ConfigKeys.ALPHA_KEY,
```
andreas-el, line 23: I have a strong feeling this list was sorted at some point. Suggest to redo that?

DanSava: Good point, it was never sorted but it should be sorted.

tests/unit_tests/c_wrappers/res/config/test_config_content.py
```diff
@@ -2,12 +2,51 @@
...
 from ert._clib.config_keywords import init_user_config_parser
 
+SINGLE_VALUE_KEY_TYPES = {
+    ConfigKeys.ANALYSIS_SELECT: "str",
+    ConfigKeys.CONFIG_DIRECTORY: "path",
+    ConfigKeys.DATAROOT: "file_path",
+    ConfigKeys.DATA_FILE: "file_path",
+    ConfigKeys.ENSPATH: "path",
+    ConfigKeys.RERUN_KEY: "bool",
+    ConfigKeys.ALPHA_KEY: "int",
```
andreas-el, line 18: Sorting again

andreas-el (changes requested):

src/ert/_c_wrappers/config/config_content.py
```diff
@@ -14,42 +14,42 @@
...
-    ConfigKeys.NUM_REALIZATIONS,
     ConfigKeys.NUM_CPU,
+    ConfigKeys.NUM_REALIZATIONS,
     ConfigKeys.OBS_CONFIG,
+    ConfigKeys.QUEUE_SYSTEM,
+    ConfigKeys.RANDOM_SEED,
+    ConfigKeys.RERUN_START_KEY,
+    ConfigKeys.RUNPATH,
+    ConfigKeys.RERUN_KEY,
+    ConfigKeys.REFCASE,
```
andreas-el, line 47:
```
  ConfigKeys.RANDOM_SEED,
    ConfigKeys.REFCASE,
    ConfigKeys.RERUN_KEY,
    ConfigKeys.RERUN_START_KEY,
    ConfigKeys.RUNPATH,
    ConfigKeys.RUNPATH_FILE,
```

src/ert/_c_wrappers/enkf/analysis_config.py
```diff
@@ -112,7 +109,7 @@ def from_dict(cls, config_dict) -> "AnalysisConfig":
             std_cutoff=config_dict.get(ConfigKeys.STD_CUTOFF_KEY, 1e-6),
             stop_long_running=config_dict.get(ConfigKeys.STOP_LONG_RUNNING, False),
             global_std_scaling=config_dict.get(ConfigKeys.GLOBAL_STD_SCALING, 1.0),
-            max_runtime=max_runtime,
+            max_runtime=config_dict.get(ConfigKeys.MAX_RUNTIME, 0),
```
valentin-krasontovitsch, line 112: this change (together with the deleted lines 104-106 above) to how `MAX_RUNTIME` is picked from a config dict seems unrelated to my eyes  - would you mind shortly explaining this change? is it just tidying up, because we do not expect `MAX_RUNTIME` to be a list in the config dict?

src/ert/_c_wrappers/config/config_content.py
```diff
@@ -14,42 +14,42 @@
...
+    ConfigKeys.RUNPATH_FILE,
     ConfigKeys.STOP_LONG_RUNNING,
```
andreas-el, line 49:
```
    ConfigKeys.STD_CUTOFF_KEY,
    ConfigKeys.STOP_LONG_RUNNING,
```

tests/unit_tests/c_wrappers/res/config/test_config_content.py
```diff
@@ -74,9 +112,35 @@ def test_config_content_as_dict(tmpdir):
...
+            "bool": "TRUE",
+            "float": 4.2,
+            "int": 2,
+        }
+
+        with open("config.file", "w") as fileH:
+            for key in SINGLE_VALUE_KEYS:
+                key_type = SINGLE_VALUE_KEY_TYPES[key]
+                fileH.write(f"{key} {type_value_map.get(key_type, key_type)}\n")
+                fileH.write(f"{key} {type_value_map.get(key_type, key_type)}\n")
```
andreas-el, line 141: This is probably obvious, but why do you write twice here? Is it to make sure keys are not duplicated at the assert step?

andreas-el (changes requested):

tests/unit_tests/c_wrappers/res/config/test_config_content.py
```diff
@@ -74,9 +112,35 @@ def test_config_content_as_dict(tmpdir):
...
-                str(Path.cwd() / existing_file_2),
-                str(Path.cwd() / existing_file_1),
-            ],
+            ConfigKeys.DATA_FILE: str(Path.cwd() / existing_file_2),
             ConfigKeys.REFCASE: str(Path.cwd() / existing_file_1),
         }
+
+
+def test_config_content_as_dict_single_value_keys(tmpdir):
+    with tmpdir.as_cwd():
```
valentin-krasontovitsch, line 121: sorry, quick question: i can't seem to figure out where this `tmpdir` is coming from - it seems like a fixture to me, but i cannot find its definition.

a short comment: the pattern of having a tmpdir and chdiring into it, to do stuff, is implemented in the `use_tmpdir` fixture and can be used by just adding `@pytest.mark.usefixtures("use_tmpdir")`
of course then you drop the argument `tmpdir`, I believe, and have to instead access the current directory by using `os.getcwd`... so as i'm not sure what `tmpdir` does exactly here, i'll leave it at your discretion - you probably know about `use_tmpdir` already anyways...

tests/unit_tests/c_wrappers/res/config/test_config_content.py
valentin-krasontovitsch, line 141: the default value in the call to `get` is given in case an unexpected key (or rather a key with unexpected type) appears?

I'm sorry, maybe I'm a bit slow, but I don't quite understand parts of this test - one aspect is rather clear, the double entry that only yields a single value, as conveyed by the assert, but the necessity of the type value map is not immediately clear to me - is this purely technical, or does this also test something?

valentin-krasontovitsch (approved):
besides my little comments and questions, this looks good! it's nice to have a list of the single value keywords ^^

tests/unit_tests/c_wrappers/res/config/test_config_content.py
DanSava: Yes, that is precisely why the key value map is written twice in the file.

src/ert/_c_wrappers/enkf/analysis_config.py
DanSava: > is it just tidying up, because we do not expect MAX_RUNTIME to be a list in the config dict?

Yes. 

Before the change, I was forced to add ln 104-106 and to guard against the possibility of users adding multiple instances of `MAX_RUNTIME` in the same config file. And "there can be only one" (movie quote) `MAX_RUNTIME` and now `config_content.as_dict()` makes sure this is the case.

tests/unit_tests/c_wrappers/res/config/test_config_content.py
DanSava: `tmpdir` is a pytest default fixture. 

And it mainly does exactly what `@pytest.mark.usefixtures("use_tmpdir")` does only that you do not have to call `os.getcwd()` to get the directory you are in. 

Also, I am not a big fun of implementing functionality that I can get for free from a package we already require for tests.

valentin-krasontovitsch: that's why i couldn't find it ^^

ah, that makes a lot of sense! great, carry on : )

src/ert/_c_wrappers/enkf/analysis_config.py
valentin-krasontovitsch: ok cool, now i understand, thanks!

tests/unit_tests/c_wrappers/res/config/test_config_content.py
valentin-krasontovitsch: OK i think i get it now - it's also a short way of supplying (valid according to type) values for all single value config keys. thanks for the explanation!

andreas-el (approved):

eivindjahren (approved):
:rocket:

PR #4165, opened 2022-10-31
Update snake_oil's checked in storage

### LINKED ISSUE #4166: Update checked in storage folder in snake_oil
It's been a while since this was updated.
Some tests will need to be updated.

dafeda:
**Issue**
Resolves #4166

**Approach**

## Pre review checklist


eivindjahren:
This adds data that seems to be about 4 times larger than the current size of the repository. Are we sure we want to do this? What is the runtime cost of generating this on the fly? Also, there are copies of snake-oil other places (semeio for instance). What do we do with those? Would it be better to make snake-oil a standalone integration-test suite (which would also test against other packages)?

eivindjahren (changes requested):
Should probably discuss going directly to not including storage in the repo instead to avoid increasing the download size of `git clone` forever.

dafeda:
> This adds data that seems to be about 4 times larger than the current size of the repository. Are we sure we want to do this? What is the runtime cost of generating this on the fly? Also, there are copies of snake-oil other places (semeio for instance). What do we do with those? Would it be better to make snake-oil a standalone integration-test suite (which would also test against other packages)?

Not a big fan of having storages checked in at all.
I've now run one ensemble experiment and one single realisation run to create two cases; `default` and `default_1`.
Before checking in, I removed all but the folder `realisation-0` from the folder `runpath`.
The currently checked in storage is 8 Mb, while the new one in this PR is 4.2 Mb so that's something.
However, I think we should get rid of it completely.

eivindjahren:
> > This adds data that seems to be about 4 times larger than the current size of the repository. Are we sure we want to do this? What is the runtime cost of generating this on the fly? Also, there are copies of snake-oil other places (semeio for instance). What do we do with those? Would it be better to make snake-oil a standalone integration-test suite (which would also test against other packages)?
> 
> Not a big fan of having storages checked in at all. I've now run one ensemble experiment and one single realisation run to create two cases; `default` and `default_1`. Before checking in, I removed all but the folder `realisation-0` from the folder `runpath`. The currently checked in storage is 8 Mb, while the new one in this PR is 4.2 Mb so that's something. However, I think we should get rid of it completely.

Nice! Just discuss possible actions within sloths and I will not stand in your way :)

dafeda:
This will be solved by making tests not depend on a checked-in storage.

PR #4171, opened 2022-10-31
Let torque driver pick up error codes from jobs

### LINKED ISSUE #3726: Torque driver does not handle non-zero exit code from jobs
**Describe the bug**
The torque driver does not propagate information about failing jobs in the same way as the other drivers. For example, the LSF driver will return `JOB_QUEUE_EXIT` if the LSF job exited with non-zero exit code. The torque driver can produce the following statuses (see also http://docs.adaptivecomputing.com/torque/4-1-3/Content/topics/commands/qstat.htm#standardOutput):

https://github.com/equinor/ert/blob/94a29ddf18ec6f72e94d5cbc6a097c57db67a411/src/libres/lib/job_queue/torque_driver.cpp#L609-L631

This will lead to strange behavior, where ERT will attempt to load data, even though the job failed. The realization will be marked as successful, despite having failed jobs. 

**To reproduce**
Can be reproduced by adding a failing job after the `pole_eval` in the poly example, and using the torque driver. 
Data will be loaded and the realization appear successful, even though the last job failed. 

**Expected behaviour**
Handle failing jobs the same way as e.g. the LSF driver does, where jobs exiting with non-zero error code will propagate `JOB_QUEUE_EXIT` from the driver and not `JOB_QUEUE_DONE`

**Screenshots**
image

**Environment**
 - OS: RHEL7
 - ERT/Komodo release: bleeding (~2.38) on azure
 - Python version: 3.8

**Additional context**
N/A

See also: https://github.com/equinor/ert/issues/3781

berland:
**Issue**
Resolves #3726 

**Approach**
Add `-f` as a hard-coded option supplied to the `qstat` command. This changes the output format from qstat from one-line pr. job to multiple lines pr. job. Only then is the exit code from the cluster made available to the torque driver.

## Pre review checklist

### DIFF
--- a/src/clib/lib/job_queue/torque_driver.cpp
+++ b/src/clib/lib/job_queue/torque_driver.cpp
@@ -1,7 +1,12 @@
 #include <filesystem>
 
+#include <array>
+#include <fstream>
+#include <iostream>
+#include <sstream>
 #include <stdio.h>
 #include <string.h>
+#include <string>
 #include <unistd.h>
 
 #include <ert/res_util/file_utils.hpp>
@@ -530,9 +535,11 @@ torque_driver_get_qstat_status(torque_driver_type *driver,
     job_status_type status = JOB_QUEUE_STATUS_FAILURE;
 
     {
-        const char **argv = (const char **)util_calloc(2, sizeof *argv);
-        argv[0] = driver->qstat_opts;
-        argv[1] = jobnr_char;
+        char const *hard_coded_qstat_option = "-f";
+        /* "qstat -f" means "full"/"long" output
+         * (multiple lines of output pr. job)  */
+        std::array argv{hard_coded_qstat_option,
+                        (const char *)driver->qstat_opts, jobnr_char};
 
         /* The qstat command might fail intermittently for acceptable reasons,
            retry a couple of times with exponential sleep. ERT pings qstat
@@ -541,9 +548,9 @@ torque_driver_get_qstat_status(torque_driver_type *driver,
         int return_value = -1;
         int sleep_time = 2;                 /* seconds */
         int max_sleep_time = 2 * 2 * 2 * 2; /* max 4 attempts */
-        while ((return_value != 0) & (sleep_time < max_sleep_time)) {
+        while ((return_value != 0) && (sleep_time < max_sleep_time)) {
             return_value =
-                util_spawn_blocking(driver->qstat_cmd, 2, (const char **)argv,
+                util_spawn_blocking(driver->qstat_cmd, argv.size(), argv.data(),
                                     tmp_std_file, tmp_err_file);
             if (return_value != 0) {
                 torque_debug(driver,
@@ -553,7 +560,6 @@ torque_driver_get_qstat_status(torque_driver_type *driver,
                 sleep_time *= 2;
             }
         }
-        free(argv);
     }
 
     if (fs::exists(tmp_std_file)) {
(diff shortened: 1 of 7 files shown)


hnformentin:
Nice PR. Could you please take a look on why some tests are not passing? Flakiness?

src/clib/lib/job_queue/torque_driver.cpp
```diff
@@ -530,9 +534,11 @@ torque_driver_get_qstat_status(torque_driver_type *driver,
     job_status_type status = JOB_QUEUE_STATUS_FAILURE;
 
     {
-        const char **argv = (const char **)util_calloc(2, sizeof *argv);
-        argv[0] = driver->qstat_opts;
-        argv[1] = jobnr_char;
+        char const *hard_coded_qstat_option = "-f";
+        const char **argv = (const char **)util_calloc(3, sizeof *argv);
```
pinkwah, line 538: Nitpicks (I realise this isn't new code but still)

Allocate `argv` on the stack.
```suggestion
        const char *argv[3];
```

Alternatively, using fancy C++: (add `#include <vector>` at the top if it doesn't exist)
```c++
    std::vector<const char *> argv{
        hard_coded_qstat_option,
        driver->qstat_opts,
        jobnr_char,
    };
```

Then:
* Remove `free(argv);` on line 562.
* Use `argv.size()` instead of the literal `3`.
* Use `argv.data()` instead of `(const char **)argv` (the cast is unnecessary anyway; the type is already `const char **`)

You can also replace `std::vector` with `std::array` to be invited to the secret cult of C++11. (The difference is marginal and `std::array` is likely slightly harder to use, but would be the most "efficient" thing to do in this case)

src/clib/lib/job_queue/torque_driver.cpp
```diff
@@ -575,74 +581,91 @@ job_status_type torque_driver_parse_status(const char *qstat_file,
...
-                    }
+    /* Parse the qstat output, looking only for requested job_id */
+    if (fs::exists(qstat_file)) {
+
+        std::string job_id_label("Job Id: ");
+        int pos;
+        int job_id = -1;
+        std::string job_state("_void_");
+        int exit_status = 0;
+        std::ifstream qstatoutput(qstat_file);
```
pinkwah, line 604: These days I like to add:
```
qstatoutput.exceptions(qstatoutput.failbit);
```

This will throw a C++ exception if the file doesn't exist (it'll check and potentially fail as soon as you call `.exceptions(`), if you try to read a formatted value that isn't correct (ie, if the buffer doesn't contain a number and you try to read a number, throw exception). Notably, if you get EOF it doesn't throw (until you try to read again I think).

Handle errors with
```c++
try {
    int number;
    qstatoutput >> number;
} catch (std::ios::failure &) {
    // do something else
}
```

*Also*:
Use `qstatoutput.imbue(std::locale::classic());`. This will set C locale to this stream object and will help us avoid the `,` vs `.` problems that we currently have.

See: https://en.cppreference.com/w/cpp/io/basic_ios/fail

src/clib/lib/job_queue/torque_driver.cpp
berland: Thanks, all nitpicks implemented!

berland:
> Nice PR. Could you please take a look on why some tests are not passing? Flakiness?

Tests passing now!

src/clib/lib/job_queue/torque_driver.cpp
```diff
@@ -530,9 +535,9 @@ torque_driver_get_qstat_status(torque_driver_type *driver,
     job_status_type status = JOB_QUEUE_STATUS_FAILURE;
 
     {
-        const char **argv = (const char **)util_calloc(2, sizeof *argv);
-        argv[0] = driver->qstat_opts;
-        argv[1] = jobnr_char;
+        char const *hard_coded_qstat_option = "-f";
```
xjules, line 538: Can this variable name be more explicit? ex. `full_output_option`?

berland: I think it is valuable to tag it in the code that this is an option that the driver hard-codes. But a comment explaining what `-f` means is absolutely warranted!

berland: Comment added.

berland:
Squashed commits.

xjules (approved):
Good job! 🚀

PR #4173, opened 2022-11-01
Remove 'mini_ert' test data fixture

pinkwah:
This is only used in a very small number of tests. In `test_enkf_fs.py` doesn't really use it (it opens the config, asserts that there is no storage with a given name, then creates a storage with that name and now it exists), and in `test_enkf_main.py` it tests `state-map` in a really roundabout way that isn't particularly interesting.


oyvindeide (approved):
LGTM!

PR #4179, opened 2022-11-02
Add default jobname and raise more descriptive config validation exception

### LINKED ISSUE #4174: Add separate exception handling of config checking
Add better handling of config validation

andreas-el:
Resolves #4174 and adds the potentially breaking change of default jobnames.

There will be squashing later on.
- Is the placement of the exception acceptable? I had some issues with circular dependencies on the original placement.
- Verify label

## Pre review checklist

### DIFF
--- a/src/ert/_c_wrappers/enkf/res_config.py
+++ b/src/ert/_c_wrappers/enkf/res_config.py
@@ -795,6 +795,13 @@ class ResConfig:
             return self.num_cpu_from_data_file
         return 1
 
+    def preferred_job_fmt(self) -> str:
+        in_config = self.model_config.getJobnameFormat()
+        if in_config is None:
+            return "JOB%d"
+        else:
+            return in_config
+
     @property
     def errors(self):
         return self._errors
--- a/src/ert/_c_wrappers/enkf/res_config.py
+++ b/src/ert/_c_wrappers/enkf/res_config.py
@@ -10,6 +10,7 @@ from ecl.ecl_util import get_num_cpu as get_num_cpu_from_data_file
 from ecl.util.util import StringList
 
 from ert._c_wrappers.config import ConfigContent, ConfigParser
+from ert._c_wrappers.config.config_parser import ConfigValidationError
 from ert._c_wrappers.enkf.analysis_config import AnalysisConfig
 from ert._c_wrappers.enkf.config_keys import ConfigKeys
 from ert._c_wrappers.enkf.ensemble_config import EnsembleConfig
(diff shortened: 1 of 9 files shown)


src/ert/_c_wrappers/enkf/res_config.py
```diff
@@ -182,17 +182,21 @@ def _alloc_from_content(self, user_config_file=None, config=None):
...
+                        "<CWD>": self.config_path,
+                        "<CONFIG_PATH>": self.config_path,
+                        "<CONFIG_FILE>": os.path.basename(user_config_file),
+                        "<CONFIG_FILE_BASE>": os.path.basename(user_config_file).split(
+                            "."
+                        )[0],
+                    },
+                )
+            except ValueError as err:
+                raise ConfigValidationError(err)
```
andreas-el, line 199: This is not really what i wanted to do, rather i'd like to raise `ConfigValidationError` directly from config_parser.py:
```python
        if validate and not config_content.isValid():
            sys.stderr.write(f"Errors parsing:{config_file} \n")
            for count, error in enumerate(config_content.getErrors()):
                sys.stderr.write(f"  {count:02d}:{error}\n")
            raise ValueError(f"Parsing:{config_file} failed")

        return config_content
```
My problem is that doing so causes a circular dependency somehow. I will look further into this.

andreas-el:
Jenkins retest this please!

src/ert/_c_wrappers/enkf/res_config.py
andreas-el: Aaand it's done! Doomguy says so. :suspect:

eivindjahren: I am guessing you moved `ConfigValidationError` from `res_config.py` to `config_parser.py` which is reasonable :) . I guess alternatively we might have wanted to to have a separate file for `ConfigValidationError`, but I think this is also good.

eivindjahren (approved):
If you don't disagree, I added the breaking change as the first item in the PR title and added the breaking change label. Probably, very few users (if any) will be notice that the default has been added, but I think better safe than sorry in this case. I believe the location of the error is good.

src/ert/_c_wrappers/enkf/res_config.py
andreas-el: Should we separate the exception into a file? That would be super easy.
Also, i'll squash everything to one commit.

valentin-krasontovitsch:
ok great eivind was faster, should have seen it coming XD sorry about my slowness!

PR #4193, opened 2022-11-04
Expose torque_driver to Python using pybind11

berland:
**Issue**
Resolves #my_issue

**Approach**

## Pre review checklist


berland:
Some immediate questions: 

* Should this new python module `torque_driver` be inside contained in say a `job_queue` module instead?
* The pybind module here defines a lambda that gets c++  vectors as input and the `argc` argument is superfluous, kept to make it consistent with the underlying function. There might not be any point in this consistency, further on, the underlying function could be rewritten to c++ too and avoid the argc altogheter.

hnformentin:
Nice PR. Could you please take a look on why some tests are not passing? Flakiness?

hnformentin:
Nice PR. Could you please take a look on why some tests are not passing? Flakiness?

xjules:
> Some immediate questions:
> 
> * Should this new python module `torque_driver` be inside contained in say a `job_queue` module instead?

Not sure we can discuss it, but I would potentially create a stand-alone module?

> * The pybind module here defines a lambda that gets c++  vectors as input and the `argc` argument is superfluous, kept to make it consistent with the underlying function. There might not be any point in this consistency, further on, the underlying function could be rewritten to c++ too and avoid the argc altogheter.

Not sure that I understand what's the worry here.

berland:
> > * The pybind module here defines a lambda that gets c++  vectors as input and the `argc` argument is superfluous, kept to make it consistent with the underlying function. There might not be any point in this consistency, further on, the underlying function could be rewritten to c++ too and avoid the argc altogheter.
> 
> Not sure that I understand what's the worry here.

Should I keep the Python API as similiar as possible to the C-API, or should it be simplified when simplification is more Pythonic?

xjules:
> 

Right. I would prefer Pythonic way to be honest.

berland:
Renamed module from `torque_driver` to `_torque_driver` to give hints of privacy. Considered slightly the possibility of moving this module out of the `ert` hiearchy since it is only there for tests, but this would probably involve noisy boilerplate to expose an additional top level module.

berland:
jenkins please test this

xjules (approved):
LGTM! :rocket:

PR #4202, opened 2022-11-07
Update script/ecl-check

pinkwah:
This script wasn't updated after migration from `libres/` to `src/clib`. I also change the following:

1. Fix off-by-one error by using `.splitlines()` instead of `.split(b"\n")`
2. Use `src/clib/lib`. Ie, don't include tests in the counting
3. Special-case `util_abort` so it gets counted

### DIFF
--- a/script/ecl-check
+++ b/script/ecl-check
@@ -7,18 +7,19 @@ import re
 from typing import Set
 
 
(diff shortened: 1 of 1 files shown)


eivindjahren (approved):
:space_invader:

PR #4205, opened 2022-11-07
Remove unused parameter in model_config_init

eivindjahren:
Cleans up unused parameter in model_config_init

### DIFF
--- a/src/clib/lib/enkf/model_config.cpp
+++ b/src/clib/lib/enkf/model_config.cpp
@@ -236,7 +236,7 @@ model_config_type *model_config_alloc(const config_content_type *config_content,
     model_config_type *model_config = model_config_alloc_empty();
 
     if (config_content)
-        model_config_init(model_config, config_content, data_root, 0, refcase);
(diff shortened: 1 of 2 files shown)


src/clib/lib/enkf/model_config.cpp
```diff
@@ -243,7 +243,7 @@ model_config_type *model_config_alloc(const config_content_type *config_content,
     model_config_type *model_config = model_config_alloc_empty();
 
     if (config_content)
-        model_config_init(model_config, config_content, data_root, 0, joblist,
```
andreas-el, line 246: ./src/clib/lib/include/ert/enkf/model_config.hpp:46:void model_config_init(model_config_type *model_config,
`
void model_config_init(model_config_type *model_config,
                       const config_content_type *, const char *data_root,
                       int ens_size, const ecl_sum_type *refcase);
`

andreas-el (changes requested):

andreas-el (approved):
As long as tests pass, my man approves. :suspect:

PR #4215, opened 2022-11-08
Remove TEST_CONFIG_FILE_KEY testing keyword

### LINKED ISSUE #4229: Remove warnings about test config key
**Describe the bug**
841c599ec548464185b12578bd91d3e1645a3161 replaced an unused config key with a test config key, but the test config key is still (methinks unnecessarily) written to config file in the tests, which leads to warning lines en masse, as the keyword is not recognized by the parser.

**To reproduce**
run tests and see warnings en masse
**Expected behaviour**
see no unnecessary warnings

valentin-krasontovitsch:
Resolves #4229 

~~The config key `TEST_CONFIG_FILE_KEY` is only used in the context of tests, in test code. Writing it to the config file leads to warnings about an unknown config key. This change should suppress those warnings.~~

After a long discussion, with many good ideas, I have arrived at a different solution, please [see below](https://github.com/equinor/ert/pull/4215#issuecomment-1312809102)

## Pre review checklist

### DIFF
--- a/src/ert/_c_wrappers/enkf/res_config.py
+++ b/src/ert/_c_wrappers/enkf/res_config.py
@@ -174,6 +174,22 @@ class ResConfig:
         logger.info("Content of the config_content:")
         logger.info(tmp_dict)
 
+    @staticmethod
+    def _create_pre_defines(
+        config_file_path: str,
+    ) -> dict:
+        date_string = date.today().isoformat()
+        config_file_dir = os.path.abspath(os.path.dirname(config_file_path))
+        config_file_name = os.path.basename(config_file_path)
+        config_file_basename = os.path.splitext(config_file_name)[0]
+        return {
+            "<DATE>": date_string,
+            "<CWD>": config_file_dir,
+            "<CONFIG_PATH>": config_file_dir,
+            "<CONFIG_FILE>": config_file_name,
+            "<CONFIG_FILE_BASE>": config_file_basename,
+        }
+
     # build configs from config file or everest dict
     def _alloc_from_content(self, user_config_file=None, config=None):
         site_config_parser = ConfigParser()
@@ -186,15 +202,7 @@ class ResConfig:
             self.config_path = os.path.abspath(os.path.dirname(user_config_file))
             user_config_content = config_parser.parse(
                 user_config_file,
-                pre_defined_kw_map={
-                    "<CWD>": self.config_path,
-                    "<DATE>": self._current_date_string(),
-                    "<CONFIG_PATH>": self.config_path,
-                    "<CONFIG_FILE>": os.path.basename(user_config_file),
-                    "<CONFIG_FILE_BASE>": os.path.basename(user_config_file).split(".")[
-                        0
-                    ],
-                },
+                pre_defined_kw_map=ResConfig._create_pre_defines(user_config_file),
             )
         else:
             self.config_path = os.getcwd()
@@ -773,10 +781,6 @@ class ResConfig:
         return config_content
 
     @staticmethod
-    def _current_date_string() -> str:
-        return date.today().isoformat()
-
-    @staticmethod
     def _create_substitution_list(
         defines: Dict[str, str],
         data_kw: Dict[str, str],
@@ -786,8 +790,7 @@ class ResConfig:
     ) -> SubstitutionList:
         subst_list = SubstitutionList()
 
-        today_date_string = ResConfig._current_date_string()
-        subst_list.addItem("<DATE>", today_date_string)
+        # only needed by everest's init from `config` path
         subst_list.addItem("<CONFIG_PATH>", config_dir)
 
         for key, value in defines.items():
--- a/tests/test_config_parsing/config_dict_generator.py
+++ b/tests/test_config_parsing/config_dict_generator.py
@@ -1,4 +1,5 @@
 import contextlib
+import datetime
 import glob
 import os
 import os.path
@@ -7,7 +8,6 @@ import stat
 from pathlib import Path
 
 import hypothesis.strategies as st
-import pytest
 from hypothesis import assume
 
 from ert._c_wrappers.enkf import ConfigKeys
@@ -273,6 +273,7 @@ def defines(draw, config_files, cwds):
     pre_defined_kw_map = draw(
         st.fixed_dictionaries(
             {
+                "<DATE>": st.just(datetime.date.today().isoformat()),
                 "<CWD>": cwds,
                 "<CONFIG_PATH>": cwds,
                 "<CONFIG_FILE>": st.just(config_file_name),
@@ -292,7 +293,6 @@ def config_dicts(draw):
     config_dict = draw(
         st.fixed_dictionaries(
             {
-                pytest.TEST_CONFIG_FILE_KEY: config_file_name,
                 ConfigKeys.NUM_REALIZATIONS: positives,
                 ConfigKeys.ECLBASE: st.just(draw(words) + "%d"),
                 ConfigKeys.RUNPATH_FILE: st.just(draw(file_names) + "runpath"),
@@ -305,7 +305,6 @@ def config_dicts(draw):
                 ConfigKeys.UPDATE_LOG_PATH: directory_names(),
                 ConfigKeys.STD_CUTOFF_KEY: small_floats,
                 ConfigKeys.MAX_RUNTIME: positives,
-                ConfigKeys.CONFIG_DIRECTORY: st.just(cwd),
                 ConfigKeys.MIN_REALIZATIONS: positives,
                 ConfigKeys.DEFINE_KEY: defines(config_file_name, st.just(cwd)),
                 ConfigKeys.DATA_KW_KEY: st.dictionaries(words, words),
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,11 +18,6 @@ from ert.shared.feature_toggling import FeatureToggling
 
 from .utils import SOURCE_DIR
 
-
-def pytest_configure():
-    pytest.TEST_CONFIG_FILE_KEY = "__TEST_CONFIG_FILE"
-
-
 # CI runners produce unreliable test timings
 # so too_slow healthcheck and deadline has to
 # be supressed to avoid flaky behavior
@@ -122,8 +117,8 @@ def copy_snake_oil_case(copy_case):
     copy_case("snake_oil")
 
 
-@pytest.fixture()
-def copy_snake_oil_case_storage(_shared_snake_oil_case, tmp_path, monkeypatch):
+@pytest.fixture(name="copy_snake_oil_case_storage")
+def fixture_copy_snake_oil_case_storage(_shared_snake_oil_case, tmp_path, monkeypatch):
     monkeypatch.chdir(tmp_path)
     shutil.copytree(_shared_snake_oil_case, "test_data")
     os.chdir("test_data")
--- a/tests/test_config_parsing/test_analysis_config.py
+++ b/tests/test_config_parsing/test_analysis_config.py
@@ -1,9 +1,7 @@
-import os
-
 import pytest
 from hypothesis import given
 
-from ert._c_wrappers.enkf import AnalysisConfig, ConfigKeys, ResConfig
+from ert._c_wrappers.enkf import AnalysisConfig, ResConfig
 
 from .config_dict_generator import config_dicts, to_config_file
 
@@ -11,10 +9,8 @@ from .config_dict_generator import config_dicts, to_config_file
 @pytest.mark.usefixtures("use_tmpdir")
 @given(config_dicts())
 def test_analysis_config_config_same_as_from_file(config_dict):
-    cwd = os.getcwd()
-    filename = config_dict[pytest.TEST_CONFIG_FILE_KEY]
+    filename = "config.ert"
     to_config_file(filename, config_dict)
-    config_dict[ConfigKeys.CONFIG_DIRECTORY] = cwd
 
     analysis_config_from_file = ResConfig(filename).analysis_config
     analysis_config_from_dict = AnalysisConfig.from_dict(config_dict=config_dict)
--- a/tests/test_config_parsing/test_ensemble_config.py
+++ b/tests/test_config_parsing/test_ensemble_config.py
@@ -1,6 +1,3 @@
-import os
-import os.path
-
 import pytest
 from hypothesis import assume, given
 
@@ -13,10 +10,8 @@ from .config_dict_generator import config_dicts, to_config_file
 @pytest.mark.usefixtures("use_tmpdir")
 @given(config_dicts())
 def test_ensemble_config_from_file_and_dict_coincide(config_dict):
-    cwd = os.getcwd()
-    filename = config_dict[pytest.TEST_CONFIG_FILE_KEY]
+    filename = "config.ert"
     to_config_file(filename, config_dict)
-    config_dict[ConfigKeys.CONFIG_DIRECTORY] = cwd
 
     res_config_from_file = ResConfig(user_config_file=filename)
     res_config_from_dict = ResConfig(config_dict=config_dict)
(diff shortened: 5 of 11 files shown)


valentin-krasontovitsch:
jenkins x-ert-ERT test failed with

```
tests/performance_tests/test_forward_model_script_overhead.py:33: 
/opt/rh/rh-python38/root/usr/lib64/python3.8/subprocess.py:359: in check_call
    retcode = call(*popenargs, **kwargs)
/opt/rh/rh-python38/root/usr/lib64/python3.8/subprocess.py:342: in call
    return p.wait(timeout=timeout)
/opt/rh/rh-python38/root/usr/lib64/python3.8/subprocess.py:1083: in wait
    return self._wait(timeout=timeout)
/opt/rh/rh-python38/root/usr/lib64/python3.8/subprocess.py:1806: in _wait
    (pid, sts) = self._try_wait(0)

self = <subprocess.Popen object at 0x7fb9494f16a0>, wait_flags = 0

    def _try_wait(self, wait_flags):
        """All callers to this function MUST hold self._waitpid_lock."""
        try:
>           (pid, sts) = os.waitpid(self.pid, wait_flags)
E           Failed: Timeout >12.0s

/opt/rh/rh-python38/root/usr/lib64/python3.8/subprocess.py:1764: Failed
```

this seems to be a timeout / flaky thing.

jenkins, retest this please.

oysteoh:
Jenkins: Test this please!

oysteoh:
Sorry, but i'm not following totally what and why here :grimacing:  Do we at all need to add this key? If it is not supported, why do we need it? I guess if not supported there is no functionality around it... Seems weird to introduce code to avoid warning-messages if we can avoid it in the first place. But i probably miss a part of the picture here... :thinking:

oysteoh:
as i said - i might miss some important parts here - but what i'm thinking - why not do i.e this 

```
@given(config_dicts())
def test_from_dict_and_from_file_creates_equal_subst_lists(config_dict):
    filename = "testconfig"
    to_config_file(filename, config_dict)
    res_config_from_file = ResConfig(user_config_file=filename)
    res_config_from_dict = ResConfig(config_dict=config_dict)
    assert (
        res_config_from_file.substitution_list == res_config_from_dict.substitution_list
    )
```

instead of what we have now
```
@given(config_dicts())
def test_from_dict_and_from_file_creates_equal_subst_lists(config_dict):
    filename = config_dict[pytest.TEST_CONFIG_FILE_KEY]    
    to_config_file(filename, config_dict)
    res_config_from_file = ResConfig(user_config_file=filename)
    res_config_from_dict = ResConfig(config_dict=config_dict)
    assert (
        res_config_from_file.substitution_list == res_config_from_dict.substitution_list
    )
```

valentin-krasontovitsch:
during parsing, through the predefined keywords mechanisms, a couple of substitutions are done, one of them for the file name of the config file. these predefined keywords are later passed to the substitution list formerly known as substitution config.
this is in the config content path.

initializing res config from a config dict, we circumvent parsing, and create the dictionary by hand (generally using hypothesis). we still expect to find the substitutions that come from the predefined keywords mechanism of the config file parser. we have to inject them somehow! we hijack the define keyword, and put the substitutions there, during generation of the config dict. then we have to make sure that the config file name we put in the defines is also used for creation of the config file based on the config dict in the test. we have to make the test function aware of the config file name that we generated for the config dict. this is exactly what `pytest.TEST_CONFIG_FILE_KEY = __TEST_CONFIG_FILE` is used for.

sorry about the wall of text.

we will be able to get rid of this once we consolidate / get rid of config content and reimagine parsing.

xjules:
> during parsing, through the predefined keywords mechanisms, a couple of substitutions are done, one of them for the file name of the config file. these predefined keywords are later passed to the substitution list formerly known as substitution config. this is in the config content path.
> 
> initializing res config from a config dict, we circumvent parsing, and create the dictionary by hand (generally using hypothesis). we still expect to find the substitutions that come from the predefined keywords mechanism of the config file parser. we have to inject them somehow! we hijack the define keyword, and put the substitutions there, during generation of the config dict. then we have to make sure that the config file name we put in the defines is also used for creation of the config file based on the config dict in the test. we have to make the test function aware of the config file name that we generated for the config dict. this is exactly what `pytest.TEST_CONFIG_FILE_KEY = __TEST_CONFIG_FILE` is used for.
> 
> sorry about the wall of text.
> 
> we will be able to get rid of this once we consolidate / get rid of config content and reimagine parsing.

Started reading this out of curiosity of why the test fails but must say haven't understand the need for having the `TEST_CONFIG_FILE_KEY` neither unfortunately

oysteoh:
yes, it is funny a small change causing much discussion 😂  This is not your fault, it is ERT which is complicated and i really want to understand why this is i needed so i would be thankful to have an explanation before we merge! 🙏

valentin-krasontovitsch:
of course - let's have a look at it tomorrow then! the more people look at these complicated things, the more confidence we get in the changes we're doing at the moment, too!

oysteoh:
Still after looking into the original commit 841c599ec548464185b12578bd91d3e1645a3161 i cannot understand why we at all need this pytest key. Why not just remove it ? It is not testing any behavior is it? As i wrote further up, why not just do what i suggest in the tests there? Is it possible to explain? :grimacing:  https://github.com/equinor/ert/pull/4215#issuecomment-1308726995

tests/test_config_parsing/config_dict_generator.py
```diff
@@ -389,6 +389,8 @@ def to_config_file(filename, config_dict):  # pylint: disable=too-many-branches
             if keyword == ConfigKeys.DATA_KW_KEY:
                 for define_key, define_value in keyword_value.items():
                     config.write(f"{keyword} {define_key} {define_value}\n")
+            elif keyword == pytest.TEST_CONFIG_FILE_KEY:
```
eivindjahren, line 392: I suggest we inline `pytest.TEST_CONFIG_FILE_KEY`and just write `keyword == "TEST_CONFIG_FILE"` everywhere.

valentin-krasontovitsch: i'm not entirely sure what you mean - instead of holding the value in a variable, you want to repeat the string literal everywhere it's used?

eivindjahren: Yes, that was generally the idea. Since it is just used in tests that seemed unproblematic.

eivindjahren: Actually, I got a better idea, so I am closing this conversation.

eivindjahren:
The way I see it, we have not graduated from just generating a dictionary to write to file as we need to know more things from the generator. I might add that we have for a while had to mock the site config to get it to work (see `tests/test_config_parsing/contest.py`). So let's create a class that holds the relevant context of the config that we generate instead, ie. add the following class to `tests/test_config_parsing/config_dict_generator.py`

```python

@dataclass
class ErtConfig:
  config_dict : Dict[str, Any]
  config_file_name: str

  def write_config(self):
    ....
```

with corresponding generator `ert_configs` made using the composite strategy. Likely we have more context about the config that we want to add such as the location and contents of the site config. This class might also become part of the implementation later.

valentin-krasontovitsch:
OK so I have arrived at the following proposal, which aims to do two things:

- resolve the mess with the `TEST_CONFIG_FILE_KEY` constant, and
- move into a hopefully better direction with respect to the predefined defines

To that end, I have

- removed the highly controversial keyword altogether
  * by replacing it with a string literal `config.ert` in tests that don't care about the predefined defines, and
  * by using a new mechanism to make sure the config file name (and path) end up in the expected defines (in tests for res config's substitution list) [see below]
- removed the `CONFIG_DIRECTORY` key from our config dict generator based tests, as setting that one is not strictly speaking necessary (in all code paths where it's used, we default to cwd, which it was in the tests as well) [SIMPLIFY]
- created a new argument to res config, `config_dict_source_file`, which means to communicate some context about the config dict, and is used to create the predefined defines explicitly, which are again explicitly passed as an argument to the methods creating the substitution list (and it is made clear with a comment that the predefineds:tm:  are part of the defines in config content)

With this I hope to make clearer the distinction between content parsed from the config file (what ends up in config dict) and context content (location and name of the config file). This is moving towards more refined ideas of Morten and Eivind, but the predefineds are still just a free for all dict (sorry Morten), and the context info is not collected in a fancy dataclass (sorry Eivind) - I wanted to leave the changes here small, for now, with room for improvement later.

I hope this solution is more agreeable, but I have also emptied my calendar next week in case there is further need for discussions ; )

tests/test_config_parsing/config_dict_generator.py
valentin-krasontovitsch: this is outdated now, the keyword is gone : )

src/ert/_c_wrappers/enkf/res_config.py
```diff
@@ -104,7 +105,9 @@ def __init__(
                 config=config,
             )
         else:
-            self._alloc_from_dict(config_dict=config_dict)
+            self._alloc_from_dict(
+                config_dict=config_dict, source_file=config_dict_source_file
+            )
```
DanSava, line 110: This seems a bit confusing to me, this seems to suggest that we only have dictionaries that are read from a file. Is this the intended design?

src/ert/_c_wrappers/enkf/res_config.py
```diff
@@ -174,6 +177,22 @@ def _log_config_content(self, config_content: ConfigContent) -> None:
         logger.info("Content of the config_content:")
         logger.info(tmp_dict)
 
+    @staticmethod
+    def _create_predefined_defines(
```
DanSava, line 181: How does `_create_pre_defines` sound? In place of `_create_predefined_defines`.

src/ert/_c_wrappers/enkf/res_config.py
eivindjahren: I am wondering if this isn't less useful. Isn't it better that the list of defines is given explicitly?

valentin-krasontovitsch: about dictionaries only read from file - it was my impression that there's where we moving towards / want to get in the end, and I did not consider other alternatives for getting a dictionary (besides from tests of course). do you have a more accurate naming suggestion, Dan?

Re: Eivind - we can totally change the interface and yank the creation of predefined defines out of resconfig, thanks for the suggestion! would you like me to let it live in...

so i'm gonna 
- change the interface of enkfmain, and let it accept a `predefined_defines: dict` argument, and
- move the convenience method for creating our current default to ...

hmm. i usually don't go higher up in the stack than enkf main, i see ResConfig is initialized

- in `__main__.py`,
- in `cli/main.py`, and
- in `libres_facade.py`

any suggestions as to where the method should live so it is accessible by all three?

src/ert/_c_wrappers/enkf/res_config.py
valentin-krasontovitsch: I'm still rooting for `predefineds`, but i'm not sure if it will stick ; )

will rename once we figure out where to move the method!

src/ert/_c_wrappers/enkf/res_config.py
eivindjahren: What we have thought about is getting a `LibresFacade.from_dict` classmethod analogous to `LibresFacade.from_file` that hides the implementation detail of initializing `ResConfig` and `EnKFMain` mind that this opens up the pandoras box of moving everest away from initializing using `ResConfig(config=...)`.

DanSava: > about dictionaries only read from file - it was my impression that there's where we moving towards / want to get in the end, and I did not consider other alternatives for getting a dictionary (besides from tests of course). do you have a more accurate naming suggestion, Dan?

Even though we are aiming at reading dictionaries from files we are not there yet and currently, there is no instance where a dictionary is read from a file (for ResConfig), so I am not sure it is a good idea to try and implement some distant future functionality now.

But let's say we want to try to do that I am not sure passing a dict and a file_name where that dict is being read from is a winning combination in any kind of future-proofing.  You would be either passing the filepath or the dict, if the implementation requires both then the implementation can be improved :)

My suggestion would be to focus on fixing the tests and not adding/changing functionality so that we can test it.

valentin-krasontovitsch: i guess we can't really future proof at this stage, really I was just trying to move towards the right direction. at the moment, you're of course right, we don't parse a dict from file - but at the same time we generate config dicts for tests, and we want to test those using the code, putting them somewhere. to me, it feels right to think about how things might look like, and nudge the code towards that direction, rather than just nudging the code towards somehow running the tests. but of course the latter strategy might be just right, and we might end up with what we want the code to look like in the end...

> You would be either passing the filepath or the dict

Morten very passionately disagreed, arguing for a dictionary that truly reflects only the contents of the user config file. that necessitates to have something "come along" with the dictionary, to carry meta info, or context, which we need. Of course, one can just as well put everything in the dictionary. I dunno. I think separating it and having the separation clear and out in the open is a good idea, but there's a difference of opinions here...

> focus on fixing the tests and not adding/changing functionality
hmmm. yeah I mean if we can, that would be the best. I'm gonna have to think about that one. of course, I think that's what I originally tried to do - fix the tests, by stuffing the DEFINE keyword with predefineds in the generated config dicts, and using a test internal keyword. and we all know how much good that did me ; )

src/ert/_c_wrappers/enkf/res_config.py
```diff
@@ -91,6 +91,7 @@ def __init__(
         user_config_file: Optional[str] = None,
         config: Optional[Dict[ConfigKeys, Any]] = None,
         config_dict: Optional[Dict[ConfigKeys, Any]] = None,
+        config_dict_source_file: Optional[str] = None,
```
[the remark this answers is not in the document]
valentin-krasontovitsch: We could also follow Eivind's suggestion ... somewhere ... and make the config dict a smarter thing. then it wouldn't be extending the number of arguments of res config, just changing the config dict and putting more information there. [there it is, a comment further up](https://github.com/equinor/ert/pull/4215#issuecomment-1311225883) - what do you think about that approach, Dan?

src/ert/_c_wrappers/enkf/res_config.py
valentin-krasontovitsch: Dan, for an alternative approach to passing the config file, see [my comment below](https://github.com/equinor/ert/pull/4215#discussion_r1021816113), or directly check out [Eivinds idea here](https://github.com/equinor/ert/pull/4215#issuecomment-1311225883)

src/ert/_c_wrappers/enkf/res_config.py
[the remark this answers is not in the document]
DanSava: I think that approach is fine.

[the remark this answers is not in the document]
valentin-krasontovitsch: winner winner chicken dinner! ( blast it, I agree with the approach as well, was just too lazy to actually implement it... :innocent: )

src/ert/_c_wrappers/enkf/res_config.py
eivindjahren: @valentin-krasontovitsch I think going down the route of implementing `LibresFacade.from_dict` and moving everest is too soon right now. Also updating the `cli/gui/everest/semeio` to a new `__init__` is kind of unnecessarily messy as we would rather go straight to `LibresFacade.from_dict/LibresFacade.from_file`.

valentin-krasontovitsch:
okay i have now done the following:

- removed the test keyword `TEST_CONFIG_FILE_KEY`, it's gone
- stuck the predefined defines for test config dicts right in there under the `DEFINE_KEY`
- there is no more extension of res config's init arguments with `source_config_file`
- only make sure that the config file name in the dict and in the config file (in our config dict tests) are equal where it matters, namely whereever the substitution list is consideres
- dropped writing the config dir to the dict under the unused key `CONFIG_DIRECTORY`

waiting eagerly for feedback : )

src/ert/_c_wrappers/enkf/res_config.py
valentin-krasontovitsch: it is done.

eivindjahren (approved):

valentin-krasontovitsch:
many thanks for the approval! :tada:

PR #4223, opened 2022-11-08
Always print blob for failing hypothesis tests

### LINKED ISSUE #4230: print blobs by default for failing hypothesis tests
the current workflow when testing with hypothesis and encountering a failing test is as follows:

- run test
- test fails
- ** go to test code, import `settings` from `hypothesis`
- ** add `@settings(print_blob=True)` above test function
- ** rerun test
- grab blob from test run output, import `reproduce_failure` from `hypothesis`, paste line from output above test
- rerun test
- ???
- SUCCESS
- ** remember to remove the settings import
- remember to remove the reproduce failure import

we should explore whether it's possible to print blobs by default, and whether this has any negative consequences (as it seems like this should be the obvious sane default)

valentin-krasontovitsch:
Resolves #4230

Doesn't change anything when tests pass, ~~makes it (theoretically) possible to reproduce failures in CI,~~ and makes it unnecessary to write `@settings(print_blob=True)` above a failing test, and rerunning, in oder to get the blob during local development.

## Pre review checklist

### DIFF
--- a/tests/test_config_parsing/conftest.py
+++ b/tests/test_config_parsing/conftest.py
@@ -1,4 +1,8 @@
 import pytest
+from hypothesis import settings
+
+settings.register_profile("blobs", print_blob=True)
+settings.load_profile("blobs")
 
 
 @pytest.fixture()


oysteoh:
is this a necessary change? according to this page https://hypothesis.readthedocs.io/en/latest/_modules/hypothesis/_settings.html# it should automatically do this when on i.e GA. And what implications will this have for normal devs ? Seems to me you want to have print_blob on CI but here you are setting it for all.... :thinking:

oysteoh:
And please, create an issue before a PR as agreed on yesterday to actually work on an issue we have agreed to work on :+1:

valentin-krasontovitsch:
no no, we wants it on all! especially for local development.

if tests pass, the consequence for the devs is none.

if a test fails, the consequence for the devs is they get an extra line of output in the tests run, something along the lines of

```
you can reproduce this error by adding @reproduce(somethingsomething)
```

as I sadly failed to convey, the current workflow without this setting when testing with hypothesis and encountering a failing test is as follows:

- run test
- test fails
- ** go to test code, import `settings` from `hypothesis`
- ** add `@settings(print_blob=True)` above test function
- ** rerun test
- grab blob from test run output, import `reproduce_failure` from `hypothesis`, paste line from output above test
- rerun test
- ???
- SUCCESS
- ** remember to remove the settings import
- remember to remove the reproduce failure import

with this change, the steps marked with ** become obsolete.

i hope this makes it clear that this is only an improvement in the developers workflow.

oysteoh:
i suggest you put this :arrow_up: in an issue then and connect your PR to that. 

Another question - do you know of any performance hits or negative effects applying this change? Why is it so "cumbersome" to debug these things in the first place? Seems like a natural thing to have to me - just makes me wonder if there is a reason it is off by default?

berland:
This sounds strange, I thought this is already default behaviour. In https://github.com/equinor/pyscal  these strings "you can reproduce the failure with .." are always emitted when things fail, and I cannot find any mention of any configuration therein that sets this explicitly.

valentin-krasontovitsch:
according to Eivind this used to be the default from many years ago. have you seen the statements lately, Berlend? I can attest with some confidence that I am not seeing the statement unless I add the setting in out current ert tests (also Andreas feels the same way).

I'm gonna investigate in the hypothesis repo / docs and see if i can discover any more information there.

valentin-krasontovitsch:
from the hypothesis changelogs:

> 5.23.0 - 2020-07-26`2
> The default print_blob setting is now smarter. It defaults to True in CI and False for local development.

the bottom of [the reproducing section of hypothesis latest docs](https://hypothesis.readthedocs.io/en/latest/reproducing.html) say:

> By default these messages are not printed. If you want to see these you must set the print_blob setting to True.

looking at the source code of current main, we see as described in the docs that the default for this setting is given by the `is_in_ci` function, which checks for existence of env vars `CI` or `TF_BUILD`. so this change is only an improvement for local development, sorry about being too hasty there!

performance wise, it seems that there is only a hit on performance if there are any `falsifying_examples`, or in other words if a test fails - this change incurs no additional computational burden for runs where all tests pass.

as to why this setting is not switched on always, I cannot say - investigating the changes and commit messages, no reasoning or explanation is given. pending further investigation, it shall remain a mystery.

does this answer all your questions, @oysteoh @berland ? : )

valentin-krasontovitsch:
I have updated the PR's description to reflect the findings about this setting being true in CI by default.

oysteoh (approved):

PR #4197, opened 2022-11-04
Delete site_config.py

mortalisk:
Related to #3986 

## Pre review checklist

- [ ] Added appropriate release note label
- [ ] PR title captures the intent of the changes, and is fitting for release notes.
- [ ] Commit history is consistent and clean, in line with the [contribution guidelines](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md).

Adding labels helps the maintainers when writing release notes. This is the [list of release note labels](https://github.com/equinor/ert/labels?q=release-notes).

PR #4227, opened 2022-11-09
Extend config_content_as_dict

eivindjahren:
Fixes config_content_as_dict so that it correctly handles more types of keywords. This is done so that it correctly handles "INSTALL_JOB_DIRECTORY" which is required for #4197 .

## Pre review checklist

### DIFF
--- a/src/ert/_c_wrappers/enkf/_config_content_as_dict.py
+++ b/src/ert/_c_wrappers/enkf/_config_content_as_dict.py
@@ -1,0 +1,113 @@
+from typing import TYPE_CHECKING, Any, Dict, List
+
+from ert._c_wrappers.enkf.config_keys import ConfigKeys
+
+if TYPE_CHECKING:
+    from ert._c_wrappers.config import ConfigContent
+
+# keywords that have one argument and
+# the last occurrence of the keyword in the file
+# is the value of which we use.
+# ie. if the config contains:
+#
+# MAX_SUBMIT 2
+# MAX_SUBMIT 3
+# ert will use the value 3 for MAX_SUBMIT
+SINGLE_OCCURRENCE_SINGLE_ARG_KEYS = [
+    ConfigKeys.ALPHA_KEY,
+    ConfigKeys.ANALYSIS_SELECT,
+    ConfigKeys.CONFIG_DIRECTORY,
+    ConfigKeys.DATAROOT,
+    ConfigKeys.DATA_FILE,
+    ConfigKeys.ECLBASE,
+    ConfigKeys.ENSPATH,
+    ConfigKeys.GEN_KW_EXPORT_NAME,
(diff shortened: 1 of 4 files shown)


oysteoh:
as i just said to valentin as well and what we discussed yesterday during the retro - we should have issues to connect our PR's to. I guess this PR should be related to issue already exists?

eivindjahren:
@oysteoh I have added link to the PR it relates to.

src/ert/_c_wrappers/enkf/_config_content_as_dict.py
```diff
@@ -0,0 +1,112 @@
...
+
+# keywords that have one argument and
+# the last occurrence of the keyword in the file
+# is the value of which we use.
+# ie. if the config contains:
+#
+# MAX_SUBMIT 2
+# MAX_SUBMIT 3
+# ert will use the value 3 for MAX_SUBMIT
+SINGLE_VALUE_SINGLE_ARG_KEYS = [
```
mortalisk, line 16: Can you call it single occurence?

src/ert/_c_wrappers/enkf/_config_content_as_dict.py
```diff
@@ -0,0 +1,112 @@
...
+
+# Should be interpreted as having three arguments:
+# ["LOCAL", "opt", "the value of the option"]
+JOIN_KEYS = [
+    (ConfigKeys.QUEUE_OPTION, 2),
+    (ConfigKeys.FORWARD_MODEL, 0),
+]
+
+
+def config_content_as_dict(
```
DanSava, line 81: I think this function is trying to do a bit much. Would it be too bad to have code like:
```python
content_dict  = config_content_as_dict(site_config_content)
if user_config_content is not None:
   content_dict.update(config_content_as_dict(user_config_content))
```
but it might be that `   content_dict.update(config_content_as_dict(user_config_content))` does not really work due to having very deep complicated dictionaries.

src/ert/_c_wrappers/enkf/_config_content_as_dict.py
eivindjahren: Done

src/ert/_c_wrappers/enkf/_config_content_as_dict.py
eivindjahren: Yes, it wouldn't really work because we don't do a straight `update`. If there are lists in there we sometimes want to append to those lists.

eivindjahren: I will add some tests that should make it clear.

eivindjahren: Tests added now

DanSava (approved):
LGTM! :+1:

PR #4243, opened 2022-11-10
Implement adaptive localization

dafeda:
Resolves: #4411

Note that the current implementation is naive in that it loops through all parameters.
My understanding is that this is the most accurate but too computationally expensive in practice.
Perhaps we can use this as reference to compare more efficient methods against.

I've added two new keywords; `LOCALIZATION` and `LOCALIZATION_CORRELATION_THRESHOLD` which may be set as for example:

```
ANALYSIS_SET_VAR STD_ENKF LOCALIZATION True
ANALYSIS_SET_VAR STD_ENKF LOCALIZATION_CORRELATION_THRESHOLD 0.2
```

## Pre review checklist


src/ert/analysis/_es_update.py
```diff
@@ -131,6 +132,29 @@ def _create_temporary_parameter_storage(
     return temporary_storage
 
 
+def correlated_parameter_response_pairs(
```
frode-aarstad, line 135: Perhaps prefix with _corr... if its an internal function

frode-aarstad (commented):
Just looking at the code it looks good. Might be worth it to add some comments to the functions if its not obvious whats going on.

src/ert/analysis/_es_update.py
```diff
@@ -131,6 +132,28 @@ def _create_temporary_parameter_storage(
...
+def _correlated_parameter_response_pairs(
+    A: npt.NDArray[np.float_], Y: npt.NDArray[np.float_], correlation_threshold: float
+) -> npt.NDArray[np.int_]:
+    N = A.shape[1]
+    Y_prime = Y - Y.mean(axis=1, keepdims=True)
+    C_YY = Y_prime @ Y_prime.T / (N - 1)
+    Sigma_Y = np.diag(np.sqrt(np.diag(C_YY)))
+
+    A_prime = A - A.mean(axis=1, keepdims=True)
+    C_AA = A_prime @ A_prime.T / (N - 1)
```
geirev, line 144: Are you computing the full error covariance matrix C_AA of A here? This one can become huge, and you only need the diagonal which you can compute with N inner products of each of the N columns of A with themself.

dafeda: I might be misunderstanding something but `A_prime` is here a `1 by N` vector because we are looping over all parameters so `C_AA` is not large.
Were you thinking of something else perhaps?

geirev: Sorry, then I misunderstood (or didn't look carefully enough).

tommyod:
We can speed up the computations by not forming the diagonal matrices. Here's an example showing what I mean, where `Sigma_A` and `Sigma_Y` are never formed as diagonal matrices. Runs around 20x faster than the current code on these sizes.

```python
ensemble_size = 10
S = np.random.randn(1000, ensemble_size)
X_local = np.random.randn(1000, ensemble_size)

# Standard deviations
Sigma_Y = np.std(S, axis=1, ddof=1)
Sigma_A = np.std(X_local, axis=1, ddof=1)

# State-measurement covariance matrix
Y_prime = S - S.mean(axis=1, keepdims=True)
A = X_local - X_local.mean(axis=1, keepdims=True)
C_AY = A @ Y_prime.T / (ensemble_size - 1)

# Absolute values of the correlation matrix
c_AY = np.abs((C_AY / Sigma_Y.reshape(1, -1)) / Sigma_A.reshape(-1, 1))
```

alternatively, if `C_AA` and `C_YY` are needed:

```python
ensemble_size = 10
S = np.random.randn(1000, ensemble_size)
X_local = np.random.randn(1000, ensemble_size)

# Estimate covariance matrix for outputs
Y_prime = S - S.mean(axis=1, keepdims=True)
C_YY = Y_prime @ Y_prime.T / (ensemble_size - 1)
Sigma_Y = np.sqrt(np.diag(C_YY))

A = X_local - X_local.mean(axis=1, keepdims=True)
C_AA = A @ A.T / (ensemble_size - 1)
Sigma_A = np.sqrt(np.diag(C_AA))

# State-measurement covariance matrix
C_AY = A @ Y_prime.T / (ensemble_size - 1)

# State-measurement correlation matrix
c_AY = np.abs((C_AY / Sigma_Y.reshape(1, -1)) / Sigma_A.reshape(-1, 1))
```

Blunde1:
> We can speed up the computations by not forming the diagonal matrices. Here's an example showing what I mean, where `Sigma_A` and `Sigma_Y` are never formed as diagonal matrices. Runs around 20x faster than the current code on these sizes.

I don't see there being anything in the `fit()` or `update()` that is slower than the previous dense matrix inversion of `Sigma_Y` (only needs to be formed once, outside loops) and `Sigma_A`, so I fully support this. It would then even be possible to hope that adaptive localization runs in reasonable time even on Troll. It is at least worth to test it.

dafeda:
Has been merged to main.

PR #4256, opened 2022-11-11
Ensure bin in path for pytest

hnformentin:
Co-authored-by: Zohar Malamant <git@wah.pink>

**Issue**
When running pytest, the path was not properly set and could not find job_dispatch.py even if it was available in the virtual environment.

**Approach**
Pre-append the PATH that has python.

## Pre review checklist

### DIFF
--- a/tests/unit_tests/conftest.py
+++ b/tests/unit_tests/conftest.py
@@ -1,5 +1,7 @@
 # This conftest still exists so that tests files can import ert_utils
 import logging
+import os
+import sys
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


tests/unit_tests/conftest.py
```diff
@@ -32,3 +34,10 @@ def __init__(self, *args, **kwargs):
             super().__init__(*args, **kwargs)
 
     monkeypatch.setattr("ert.cli.main.EvaluatorServerConfig", MockESConfig)
+
+
+@pytest.fixture(scope="session", autouse=True)
+def ensure_bin_in_path():
```
pinkwah, line 40:
```suggestion
def ensure_bin_in_path():
    """
    Running pytest directly without enabling a virtualenv is perfectly valid.
    However, our tests assume that `job_dispatch.py` is in PATH which it may not be.
    This fixture prepends the path to the current Python for tests to pass when not in a virtualenv.
    """
```

pinkwah (approved):

PR #4269, opened 2022-11-14
Resubmit the same event when failed connection

xjules:
**Issue**
Do not pick up new events when Event was not sent due to the connection error

**Solution**
The event from job_dispatch will not be lost , event reporter tries to resubmit until either it
succeed or Finish event was received. This triggers timeout to give it some extra time to submit the event.

Additionally, Client gets ClientConnectionError exception, which bundles / hides all 3rd party
exceptions raised by either while establishing a new connection or sending the event data.

## Pre review checklist

### DIFF
--- a/src/_ert_job_runner/client.py
+++ b/src/_ert_job_runner/client.py
@@ -1,11 +1,27 @@
 import asyncio
+import logging
 import ssl
 from typing import Any, AnyStr, Dict, Optional, Union
 
 import cloudevents
 from websockets.client import WebSocketClientProtocol, connect
 from websockets.datastructures import Headers
-from websockets.exceptions import ConnectionClosedError, ConnectionClosedOK
+from websockets.exceptions import (
+    ConnectionClosedError,
+    ConnectionClosedOK,
+    InvalidHandshake,
+    InvalidURI,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class ClientConnectionError(Exception):
+    pass
+
+
+class ClientConnectionClosedOK(Exception):
+    pass
 
 
 class Client:  # pylint: disable=too-many-instance-attributes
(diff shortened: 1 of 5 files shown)


lars-petter-hauge:
Should we add in the git commit message of `Refactor exceptions in client.py` that `OSerror`and `ClientConnectionRefused` are not exceptions thrown by the methods we are using, to elaborate as to why they are removed?

src/_ert_job_runner/client.py
```diff
@@ -1,11 +1,25 @@
...
+    InvalidURI,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class ClientConnectionError(Exception):
+    def __init__(self, reason: str):
+        super().__init__(reason)
+        self.reason: str = reason
```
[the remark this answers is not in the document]
xjules: Seems valid!

xjules:
> Should we add in the git commit message of `Refactor exceptions in client.py` that `OSerror`and `ClientConnectionRefused` are not exceptions thrown by the methods we are using, to elaborate as to why they are removed?

Can do it. Will create a single commit with a broader description what's going on.

src/_ert_job_runner/client.py
```diff
@@ -74,18 +86,31 @@ async def _send(self, msg: AnyStr) -> None:
                 await self.websocket.send(msg)
```
berland, line 86: It's hard to say which exceptions belong to `get_websocket` and which to `send()`. Is it an idea to split the `try:` block?

xjules: @lars-petter-hauge provided a version, which solves the per-partes try: except: but we both agreed it's not looking as good :) 
```python
async def _handle_exception(self, msg, exc, retry):
    if retry == self._max_retries:
        logger.debug(msg)
        raise ClientConnectionError(msg) from exc
    await asyncio.sleep(0.2 + self._timeout_multiplier * retry)

async def _send(self, msg: AnyStr) -> None:
    for retry in range(self._max_retries + 1):
        if self.websocket is None:
            try:
                self.websocket = await self.get_websocket()
            except (
                InvalidHandshake,
                InvalidURI,
                asyncio.TimeoutError,
                ConnectionRefusedError,
                OSError,
            ) as exception:
                _error_msg = (
                    f"Not able to establish the "
                    f"websocket connection {self.url}! Max retries reached!"
                )
                await self._handle_exception(_error_msg, exception, retry)
                continue
        try:
            await self.websocket.send(msg)
        except ConnectionClosedOK:
            logger.debug("Connection was closed from the server with ok message!")
            raise
        except ConnectionClosedError as exception:
            _error_msg = (
                f"Not been able to send the event"
                f" to {self.url}! Max retries reached!"
            )
            await self._handle_exception(_error_msg, exception, retry)
```

berland: I think this can be split without increasing indendation. But splitting it makes `_send` hard to read. Can we push the exception handling for `get_websocket()` into `get_websocket()` itself? (it could even has its own retry loop)

src/_ert_job_runner/client.py
```diff
@@ -74,18 +86,32 @@ async def _send(self, msg: AnyStr) -> None:
...
-                await asyncio.sleep(0.2 + self._timeout_multiplier * retry)
-                self.websocket = None
+                    _error_msg = (
+                        f"Not been able to send the event"
+                        f" to {self.url}! Max retries reached!"
+                        f" Exception from {type(exception)}: {str(exception)}"
+                    )
+                    raise ClientConnectionError(_error_msg) from exception
+            await asyncio.sleep(0.2 + self._timeout_multiplier * retry)
+            self.websocket = None
```
berland, line 114: Is this indentation change deliberate or by accident? By reading the code it might be that the indent level of this might not matter though due to the `return`.

xjules: This indentation is deliberate as it is valid for both exceptions when we don't raise the exception. We might put a new line in between though.

src/_ert_job_runner/client.py
xjules: We considered it but this would change the behavior of `get_websockets()` since we won't to rise `ClientConnectionError` instead.

berland (approved):
:ship:

PR #4322, opened 2022-11-17
Stop using subst_list C-objects in ext_job

### LINKED ISSUE #3617: Responsibility of substitution is split between subst_list and substituter.py
Currently, substitutions are held by two objects, the cpp `subst_list` in `res_config`  and the python `Substituter` held by `EnKFMain`. Only `Substituter` should be responsible for which substitutions to perform and `subst_list` only the format used to convey those substitutions to the cpp functions. The `subst_list` held by `res_config` should not change after it is done reading from file.

https://github.com/equinor/ert/blob/14f380c2fea77544401b545281b64c6171403087/res/enkf/enkf_main.py#L438-L445

At least the workflows currently get their substitutions from `res_config` and not from `Substituter`.

DanSava:
**Issue**
Partial resolves #3617 

## Pre review checklist


src/ert/_c_wrappers/job_queue/ext_job.py
```diff
@@ -60,24 +57,24 @@ class ExtJob(BaseCClass):
...
     def __init__(
         self,
         config_file: str,
         private: bool,
         name: Optional[str] = None,
         search_PATH: bool = True,
     ):
+        self._name = name
+        self._config_file = config_file
+        self._private = private
```
oysteoh, line 73: i like `self. _is_private` better as i reads better.... But leave it up to you... Just a very very minor comment :sweat_smile:

src/ert/_c_wrappers/job_queue/ext_job.py
```diff
@@ -60,24 +57,24 @@ class ExtJob(BaseCClass):
...
+        self._search_PATH = search_PATH
```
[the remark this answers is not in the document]
DanSava: That is a very good question, the name of the argument in the constructor was `search_PATH` and I just copy pasted it blindly. Will fix

src/ert/_c_wrappers/job_queue/ext_job.py
DanSava: Good suggestion!

src/ert/_c_wrappers/util/substitution_list.py
```diff
@@ -71,6 +71,18 @@ def indexForKey(self, key):
     def substitute(self, to_substitute: str) -> str:
         return self._alloc_filtered_string(to_substitute)
 
+    def substitute_real_iter(
+        self, to_substitute: str, realization: int, iteration: int
+    ) -> str:
+        if f"<GEO_ID_{realization}_{iteration}>" in self:
+            to_substitute = to_substitute.replace(
+                "<GEO_ID>", f"<GEO_ID_{realization}_{iteration}>"
```
andreas-el, line 79: Can you maybe elaborate why this is done differently?
Why is one substitution done before the `_alloc_filtered_string`, even if it preserves the string?

src/ert/_c_wrappers/job_queue/ext_job.py
```diff
@@ -221,7 +230,15 @@ def get_arglist(self) -> List[str]:
...
-        return list(self._get_argvalues())
+        arg_vals = []
+        for arg in self._get_arglist():
+            if self._private_context is not None:
+                arg = self._private_context.substitute(arg)
+            if self._global_context is not None:
+                arg = self._global_context.substitute(arg)
+            arg = self.get_default_mapping().get(arg, arg)
+            arg_vals.append(arg)
+        return arg_vals
```
andreas-el, line 241: I assume the global context should overwrite private context. Is this related to concepts userconfig and siteconfig maybe? What i fail to understand is the the default_mapping on line 239, is there some default substitutions defined that will always be done when working on these configuration files?

src/ert/_c_wrappers/util/substitution_list.py
DanSava: The GEO_ID is a concept coming from "Everest". The "Everest" reservoir optimization application is using the BatchSimulator from ERT to get its results and it is using GEO_ID to identify one run of the whole ensemble.

So from ERT point of view if we run an ensemble of 2 models for 2 iterations we would have the same GEO_ID for the first batch of runs i.e 

|realization| Iteration| geo_id|
| 0              | 0          |0          |    
| 0              | 1          |0          |    
| 1              | 0          |0          |    

--------------------------------------

A new batch of forward models runs
|realization| Iteration| geo_id|
| 2              | 0          |1          |    
| 2              | 1          |1          |    
| 3              | 0          |1          |    

So in order to achieve the correct substitution I am setting `<GEO_ID_{real}_{iter}>` in the list of available substitutions when Everest starts the batch run process and when we encounter a string containing a <GEO_ID> for a given realization and iteration we first substitute `<GEO_ID>` with`<GEO_ID_{real}_{iter}>` so that the final correct substitution is done.

src/ert/_c_wrappers/job_queue/ext_job.py
DanSava: > I assume the global context should overwrite the private context.

Actually is the other way around private context will overwrite global context
example let's say you have in the private context the following substitution defined:
<Location>="Bergen"
And in the global context:
<Location>="Oslo"
default mapping
<Location>="New York"

When we encounter the following arguments: ["We", "are", "going",  "to",  "<Location>"]
for argument "<Location>" the code above will replace "<Location>" with "Bergen"
and not with "New York"

> Is this related to concepts userconfig and siteconfig maybe?
You might consider them a bit similar but the concept is not related to user config and site config. 

> What I fail to understand is  the default_mapping on line 239, is there some default substitutions defined that will always be done when working on these configuration files
So yes there can be a default mapping thing that will be used to substitute arguments from jobs but that is the last thing that happens if it is defined and something else from the private or global context has not replaced it before.

andreas-el: Thanks for the clarification. After the description i see the order of substitution is top-down, and not bottom-up as i first interpreted this to be (looking at re-assignments of arg). :+1:

andreas-el (dismissed):
LGTM

src/ert/_c_wrappers/enkf/runpaths.py
```diff
@@ -32,21 +34,26 @@ def __init__(
         job_name_format: str,
         runpath_format: str,
         filename: PathLike = Path(".ert_runpath_list"),
-        substitute: Callable[[str, int, int], str] = lambda x, *_: x,
+        context: Optional[SubstitutionList] = None,
```
eivindjahren, line 37: I preferred the existing interface of `Runpaths` where it did not directly depend on SubstitutionList. As `Runpaths` was defined, it should suffice to pass it `context.substitute_real_iter`.

tests/unit_tests/c_wrappers/res/enkf/test_enkf_main.py
```diff
@@ -110,26 +109,10 @@ def test_create_run_context(monkeypatch, enkf_main):
...
-
-    enkf_main.set_geo_id("geo_id", realization, iteration)
-
-    assert (
-        enkf_main.substituter.get_substitutions(realization, iteration)["<GEO_ID>"]
-        == geo_id
-    )
+    context = enkf_main.get_context()
+    assert "<RUNPATH>" in context
+    assert context.substitute_real_iter("<ECL_BASE>", 1, iteration) == "name1"
```
eivindjahren, line 114: I think `substitute_real_iter` is a bit of a perplexing name if you are not familiar with what it does. I am not sure how to remedy that though.

eivindjahren (changes requested):
I am not sure that this is the correct direction to take. `Substituter` would be a good starting point for us to use in a python reimplementation of the parser. Furthermore, substitutionlist has several corner-cases and surprising behaviors, not the least of which is that it does substitution when you add new keywords.

I would much rather prefer that we isolate substitutionlist to the old parser and let Substituter take over after the call to `config_content.get_const_define_list()`:

```
substituter = Substituter(dict(config_content.get_const_define_list()))
```
And only use substitution list in the old parser.

tests/unit_tests/c_wrappers/res/enkf/test_enkf_main.py
DanSava: Not really sure how to fix the name such that it captures what is going on which is, it is trying to replace a string that contains <IENS> and <ITER> with a given pair of realisation number and iteration number

src/ert/_c_wrappers/enkf/runpaths.py
DanSava: Passing callable functions defined internally in objects as part of constructor arguments does not really decouple the two objects that much, but that is also a working alternative I will change back.

DanSava:
>I am not sure that this is the correct direction to take. Substituter would be a good starting point for us to use in a python reimplementation of the parser. Furthermore, substitutionlist has several corner-cases and surprising behaviors, not the least of which is that it does substitution when you add new keywords.

As discussed on slack, SubstitutionList might have several corner-cases and surprising behaviors but doing substitutions when a new keyword is added is not really one of them.

src/ert/_c_wrappers/util/substitution_list.py
```diff
@@ -71,6 +71,18 @@ def indexForKey(self, key):
...
+            )
+        to_substitute = self._alloc_filtered_string(to_substitute)
+        to_substitute = to_substitute.replace("<IENS>", str(realization))
```
eivindjahren, line 82: After our talk I had another look at the substlist and I realized that it actually operates differently than I thought. It does substitution until fixed-point when it gets the string. This behavior was in the previous release probably broken by the introduction of substituter, but only for substitutions in `runpaths`. Basically a substitution inside a runpath is not done until fixed point. Luckily we did not introduce substituter more places, but since then it has also been used for printing `jobs.json`.

To remedy this we have to either do the `replace` until fix-point or copy the substlist, insert what we want to substitute, and let the copy do it:

```
substlist = copy.deepcopy(self)
substlist.add_item("<GEO_ID>", f"<GEO_ID_{realization}_{iteration}>")
substlist.add_item("<IENS>", str(realization))
substlist.add_item("<ITER>", str(iteration))
return substlist._alloc_filtered_string(to_substitute)
```

I am exposing the copy action to python in #4386

src/ert/_c_wrappers/enkf/enkf_main.py
```diff
@@ -378,8 +370,7 @@ def addDataKW(self, key: str, value: str) -> None:
         # self.resConfig().substitution_list is still
         # used by workflows to do substitution. For now, we
         # need to update this here.
```
oysteoh, line 372: This comment :arrow_up:  seems a bit outdated as-is now and should maybe be removed?

src/ert/_c_wrappers/util/substitution_list.py
DanSava: I don't understand what you mean by "It does substitution until fixed-point when it gets the string"

Is there any example where your suggested code:
```python 
substlist = copy.deepcopy(self)
substlist.add_item("<GEO_ID>", f"<GEO_ID_{realization}_{iteration}>")
substlist.add_item("<IENS>", str(realization))
substlist.add_item("<ITER>", str(iteration))
return substlist._alloc_filtered_string(to_substitute)
```
does not do the same thing as the code currently in the substitution_list.py ln74-84?

eivindjahren: No, this substitutes `<IENS>` only at the end. I can't really come up with anything that breaks this other than contrived things, but I think that is just a lack of imagination on my part:

```
DEFINE <IT1ER> <ITER>
RUNPATH realization-<REAL>/iter-<IT<ITER>ER>/
```
For the existing code, for realization 0 iteration 1 you get runpath: `"realization-0/iter-<IT1ER>" while if you use the `add_item` code you get `realization-0/iter-1`

eivindjahren: Wait, looking at the subst_list again, I am not sure what it does. It seems to go through the keys only once.

DanSava: If you are a user and you have defines like:
```
DEFINE <IT1ER> <ITER>
RUNPATH realization-<REAL>/iter-<IT<ITER>ER>
```
and expect 

`realization-0/iter-1` as a result you should go to user jail :smile: 

but this will be solved by doing:

``` python 
to_substitute = to_substitute.replace("<IENS>", str(realization))
to_substitute = to_substitute.replace("<IENS>", str(realization))
to_substitute = self._alloc_filtered_string(to_substitute)
```

DanSava: Did some further testing on how the `SubstituteList` works and

```python
    subst_list = SubstitutionList()
    subst_list.addItem("<foo>", "r1")
    subst_list.addItem("<bar_r1>", "r2")
    print(subst_list.substitute("replace this <bar_<foo>>"))
```
this returns 
```python 
replace this r2
```
and this 
```python
    subst_list = SubstitutionList()
    subst_list.addItem("<bar_r1>", "r2")
    subst_list.addItem("<foo>", "r1")
    print(subst_list.substitute("replace this <bar_<foo>>"))
```
returns 
```python 
replace this <bar_r1>
```
so the order in which things are added to the substitution list matters

eivindjahren: Unfortunately we have seen stuff like this in the wild.

eivindjahren (approved):

PR #4351, opened 2022-11-22
Refactor ErtWorkflow to use one codepath

### LINKED ISSUE #4344: Refactor ErtWorkflowList to have single codepath
Make __init__ for ErtWorkflowList use lists instead of dicts. Make the config_dict similar to config_content to get one codepath.

Builds on: https://github.com/equinor/ert/pull/4231

https://github.com/equinor/ert/pull/4231#discussion_r1028023078

> The difference between _handle_config_content_dict and _handle_config_dict is just the lookup of ConfigKeys.Name, ConfigKeys.Path, and ConfigKeys.RUNMODE are indices instead. Because there is no code using _handle_config_dict (except for tests). I suggest we change the expected input for that to be integer indices instead of strings so that we can have just one code path.

https://github.com/equinor/ert/pull/4231#discussion_r1028028142

> Instead of passing dicts to __init__, I would prefer it if we pass workflow_job_info, workflow_job_dir_info, hook_workflow_info and workflow_info to __init__ instead and create a classmethod from_dict which gets the dictionary and creates the class via __init__. See analysis_config.py for an example of what I mean.

andreas-el:
Resolves #4344

**Approach**

## Pre review checklist

### DIFF
--- a/src/ert/_c_wrappers/enkf/ert_workflow_list.py
+++ b/src/ert/_c_wrappers/enkf/ert_workflow_list.py
@@ -13,19 +13,17 @@ logger = logging.getLogger(__name__)
 class ErtWorkflowList:
     def __init__(
         self,
-        content_dict=None,
-        config_dict=None,
+        workflow_job_info: list = None,
+        workflow_job_dir_info: list = None,
+        hook_workflow_info: list = None,
+        workflow_info: list = None,
     ):
-        if content_dict is None and config_dict is None:
-            raise ValueError(
-                "Failed to construct ErtWorkflowList instance with no config object"
-            )
-
-        if content_dict is not None and config_dict is not None:
-            raise ValueError(
-                "Failed to construct ErtWorkflowList "
-                "instance with multiple config object"
-            )
+        workflow_job_info = [] if workflow_job_info is None else workflow_job_info
+        workflow_job_dir_info = (
+            [] if workflow_job_dir_info is None else workflow_job_dir_info
+        )
+        hook_workflow_info = [] if hook_workflow_info is None else hook_workflow_info
+        workflow_info = [] if workflow_info is None else workflow_info
 
         self._workflow_jobs = {}
         self._workflow = {}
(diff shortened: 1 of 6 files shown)


tests/unit_tests/c_wrappers/res/enkf/test_workflow_list.py
```diff
@@ -14,37 +14,22 @@ def test_workflow_list_constructor():
...
         ConfigKeys.LOAD_WORKFLOW_JOB: [
-            {
-                ConfigKeys.NAME: "print_uber",
-                ConfigKeys.PATH: os.getcwd() + "/workflows/UBER_PRINT",
-            },
-            {
-                ConfigKeys.NAME: "HIDDEN_PRINT",
-                ConfigKeys.PATH: os.getcwd() + "/workflows/HIDDEN_PRINT",
-            },
+            [os.getcwd() + "/workflows/UBER_PRINT", "print_uber"],
```
oysteoh, line 17: nitpicking of course - but would it be nicer ( and probably more performant? ) having `cwd = os.getcwd()` and then reuse the variable instead of doing all the gets? :grimacing:

andreas-el: Sure. Already in. :+1:

src/ert/_c_wrappers/enkf/ert_workflow_list.py
```diff
@@ -13,51 +13,59 @@
...
-        self._hook_workflow_info = []
-
-        if content_dict is not None:
-            self._handle_config_content_dict(content_dict)
-
-        if config_dict is not None:
-            self._handle_config_dict(config_dict)
-
-        for workflow_job in self._workflow_job_info:
+        for workflow_job in workflow_job_info:
```
eivindjahren, line 27: This will not work with the default value `None`

andreas-el: Gotcha. :+1:

oysteoh (approved):

PR #4377, opened 2022-11-23
Remove file_sync_timeout improving performance on quick jobs

eivindjahren:
Under certain conditions the job runner would wait 10 seconds before exiting due to a race condition involving the ok file. This ok file is no longer read by ert so there is no longer a race condition.

This PR also removes references to the ok file outside the writing of it which we want to keep for backwards compatability reasons.

Two tests relied on the forward model taking some time so that a stop could be issued before finishing. However, when the sleep was removed, that could fail. The setup for these tests were changed so that the jobs were guaranteed to take some time.

## Pre review checklist

### DIFF
--- a/tests/unit_tests/simulator/test_batch_sim.py
+++ b/tests/unit_tests/simulator/test_batch_sim.py
@@ -4,6 +4,7 @@ import time
 
 import pytest
 
+from ert._c_wrappers.enkf import ResConfig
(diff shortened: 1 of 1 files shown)


mortalisk (approved):
OK file.
Make sure you test with real network disks and real queues and real enkf.

PR #4380, opened 2022-11-23
Don't install `qstat_proxy.sh` to bin

pinkwah:
To avoid installing this file, programmatically add this file's path when loading `clib`. To do this, I had to move `qstat_proxy.sh` to `ert/_c_wrappers/job_queue` so that it would ship with ERT's wheel.

The side effect of this is that `qstat_proxy.sh` is no longer in the user's `PATH`.


src/clib/lib/job_queue/torque_driver.cpp
```diff
@@ -745,3 +750,18 @@ int torque_driver_get_submit_sleep(const torque_driver_type *driver) {
 FILE *torque_driver_get_debug_stream(const torque_driver_type *driver) {
     return driver->debug_stream;
 }
+
+ERT_CLIB_SUBMODULE("torque_driver", m) {
```
berland, line 754: Merge with `_torque_driver` which is already there?

pinkwah: You are right

pinkwah:
Jenkins test this please

berland:
(suggest to squash the two commits)

valentin-krasontovitsch:
might get closed in favor of https://github.com/equinor/ert/pull/4380

berland:
> might get closed in favor of #4380

Circular reference.. :)

berland (approved):
Smart way of deploying this! I think I tried something similar first, but not figure it out how to do it.

PR #4390, opened 2022-11-28
Simplify EnsembleConfig

### LINKED ISSUE #4254: Simplify EnsembleConfig
Blocked by #4253

andreas-el:
Partially Resolves #4254

Will squash prior to merge.
Please take an extra look at the mutex lock, which to me, seems unused.

## Pre review checklist

### DIFF
--- a/src/clib/lib/enkf/ensemble_config.cpp
+++ b/src/clib/lib/enkf/ensemble_config.cpp
@@ -33,14 +33,10 @@ namespace fs = std::filesystem;
 static auto logger = ert::get_logger("ensemble_config");
 
 struct ensemble_config_struct {
-    pthread_mutex_t mutex;
-    char *
-        gen_kw_format_string; /* format string used when creating gen_kw search/replace strings. */
     std::map<std::string, enkf_config_node_type *>
         config_nodes; /* a hash of enkf_config_node instances - which again contain pointers to e.g. field_config objects.  */
     field_trans_table_type *
         field_trans_table; /* a table of the transformations which are available to apply on fields. */
-    bool have_forward_init;
     summary_key_matcher_type *summary_key_matcher;
 };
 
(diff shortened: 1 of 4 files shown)


andreas-el:
Jenkins retest please!

src/ert/_c_wrappers/enkf/ensemble_config.py
```diff
@@ -359,9 +352,6 @@ def from_dict(cls, config_dict):
 
         return ens_config
 
-    def __len__(self):
-        return self._size()
-
```
valentin-krasontovitsch, line 364: i was surprised by this deletion, because i remembered that ensemble size was something used here and there, in particular in libres facade. checking the code, in enkf main, we find

```
self._ensemble_size = self.res_config.model_config.num_realizations
```
it obviously has nothing to do with ensemble config, how silly of me :smiling_face_with_tear:

src/clib/lib/enkf/ensemble_config.cpp
```diff
@@ -151,7 +149,13 @@ enkf_config_node_type *ensemble_config_get_or_create_summary_node(
 
 bool ensemble_config_have_forward_init(
     const ensemble_config_type *ensemble_config) {
-    return ensemble_config->have_forward_init;
+
+    for (auto const &node_it : ensemble_config->config_nodes) {
+        if (enkf_config_node_use_forward_init(node_it.second))
```
valentin-krasontovitsch, line 142: ok just two small questions here -
- in the naming `node_it`, what does the `it` signify wait while writing this i figured out it stands for iterator would you mind spelling it out for the slow ones among us? : D
- the .second throws me off a bit, after checking the type of the `config_nodes` field it becomes clear though. would it be weird to add something like `tuple` to the name of the iterator variable?

valentin-krasontovitsch (approved):
very nice. i like the inlining for simplification. maybe the init_forward is a bit not DRY, but I guess it's OK.
just small rename requests, I'll leave it at your discretion!

src/clib/lib/enkf/ensemble_config.cpp
andreas-el:
```
        for (auto const &[str, enkf_config_node] : self->config_nodes) {
            if (enkf_config_node->forward_init)
```
Will this work? We use C++17, so this should be syntactically ok.

andreas-el: Updated according to slack discussion.

valentin-krasontovitsch: i'm sorry i have no idea - i'm not very familiar with the differences between C++ standards, and don't know which one we're using. try and push, see if it doesn't fail in CI?

andreas-el: This syntax require C++17, it builds locally, so i guess this will be alright. Jenkins has the last word regardless.

andreas-el:
Jenkins retest this please!

PR #4412, opened 2022-12-02
Fix runpath check

frode-aarstad:
**Issue**
Resolves #my_issue

**Approach**

## Pre review checklist

### DIFF
--- a/src/ert/shared/models/base_run_model.py
+++ b/src/ert/shared/models/base_run_model.py
@@ -525,17 +525,26 @@ class BaseRunModel:
         return update_id
 
     def check_if_runpath_exists(self) -> bool:
+        """
+        Determine if the run_path exists by checking if it contains
+        at least one iteration directory for the iterations in the active mask.
+        The run_path can contain one or two %d specifiers ie:
+            "realization-%d/iter-%d/"
+            "realization-%d/"
+        """
         run_path = self.facade.run_path
         start_iteration = self._simulation_arguments.get("start_iteration", 0)
         number_of_iterations = self.facade.number_of_iterations
         active_mask = []
         if "active_realizations" in self._simulation_arguments:
             active_mask = self._simulation_arguments["active_realizations"]
-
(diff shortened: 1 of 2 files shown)


src/ert/shared/models/base_run_model.py
```diff
@@ -535,7 +535,10 @@ def check_if_runpath_exists(self) -> bool:
         idx = [i for i in range(len(active_mask)) if active_mask[i]]
         for i in range(start_iteration, number_of_iterations):
             for j in idx:
-                s = run_path % (j, i)
+                if run_path.count("%d") == 2:
```
eivindjahren, line 538: It is very hard to understand why this is done the way it is. Could you please add some documentation about what is going on.

eivindjahren (changes requested):
Currently, there isn't enough documentation to understand what is going on. Could you please describe what is happening and why we should do it this way? There also needs to be a comment in the code to understand why this is being done.

src/ert/shared/models/base_run_model.py
frode-aarstad: Added a comment to the function to clarify the run_path format

src/ert/shared/models/base_run_model.py
```diff
@@ -525,17 +525,26 @@ def _post_update_data(self, parent_ensemble_id: str, algorithm: str) -> str:
         return update_id
 
     def check_if_runpath_exists(self) -> bool:
+        """
+        Determine if the run_path exists by checking if it contains
+        at least one iteration directory for the iterations in the active mask.
+        The run_path ccan contain one or two %d specifiers ie:
```
eivindjahren, line 531: `can` is misspelled.

frode-aarstad: The we also need to update the docs as they state 

`It should contain at least one %d specifier..`

eivindjahren (approved):

PR #4416, opened 2022-12-02
Do substitution until fixed point

### LINKED ISSUE #4400: Align DEFINE and DATA_KW behavior
- [ ] Communicate to the rest of the team / ert-users on slack/yammer our intentions

Previously, DEFINES only resultet in limited support of mixing "define before use" and "use before define":

The following follows the "define before use" order (`<USER>` is defined before it is used in the definition of `<HOME>`)
```
DEFINE <USER> eja
DEFINE <HOME> /home/<USER>
```
"use before define" is simply the opposite order:
```
DEFINE <HOME> /home/<USER>
DEFINE <USER> eja
```
Mixing these two orders has limited support (you can mix up to three times for a chain of define dependencies).
Also, there is no hard requirement that a key starts `<` and ends with `>` (something we have seen in the wild). This means that substitution can lead to computationally complex behavior.

eivindjahren:
**Issue**
Resolves #4400 

This ensures that order of defines/data_kws do not impact substitution by doing it until fixed-point (with an upper limit on number of iterations to ensure termination).

This means that if keywords are not checked, DEFINES are very complex computationally. In order to restrict how complicated a config can be, we should give warnings if:

1. DEFINE/DATA_KW keywords do not start with `<` and end with `>`.
1. If keywords are nested, e.g. `<ENS-<ITER>>`.

If the above is not followed, there is potential for undefined behavior.

Such a check should probably be done by some mechanism for suggesting migration paths for deprecated configs, and unconventional keywords eventually resulting in hard fails.

## Pre review checklist

### DIFF
--- a/src/clib/lib/res_util/subst_list.cpp
+++ b/src/clib/lib/res_util/subst_list.cpp
@@ -276,7 +276,6 @@ static bool subst_list_replace_strings(const subst_list_type *subst_list,
 */
 bool subst_list_filter_file(const subst_list_type *subst_list,
                             const char *src_file, const char *target_file) {
-    bool match;
     char *backup_file = NULL;
     buffer_type *buffer = buffer_fread_alloc(src_file);
     // Ensure that the buffer is a \0 terminated string:
@@ -298,7 +297,19 @@ bool subst_list_filter_file(const subst_list_type *subst_list,
     }
 
     /* Doing the actual update */
-    match = subst_list_replace_strings(subst_list, buffer);
+    bool matched_at_least_once = false;
+    bool match = true;
+    const int max_iterations = 10000;
+    int iterations = 1;
+    while (match && iterations++ < max_iterations) {
+        match = subst_list_replace_strings(subst_list, buffer);
+        matched_at_least_once |= match;
+    }
+
+    if (iterations >= max_iterations) {
+        throw std::runtime_error(
+            "Reached max iterations while trying to resolve defines");
+    }
 
     /* Writing updated file */
     {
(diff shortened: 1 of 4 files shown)


valentin-krasontovitsch (approved):
looking good!

mortalisk:
Do we simply not allow < or > in define keywords? Or is it complete nested keywords <A> we don't allow?

eivindjahren:
> Do we simply not allow < or > in define keywords? Or is it complete nested keywords we don't allow?

We don't allow nested, unbracketed etc. I will extend the test

src/clib/lib/res_util/subst_list.cpp
```diff
@@ -339,8 +350,19 @@ bool subst_list_update_string(const subst_list_type *subst_list,
 char *subst_list_alloc_filtered_string(const subst_list_type *subst_list,
                                        const char *string) {
     char *filtered_string = util_alloc_string_copy(string);
-    if (subst_list)
-        subst_list_update_string(subst_list, &filtered_string);
+    if (subst_list) {
```
mortalisk, line 353: Do we need to repeat this here?

mortalisk: The following code that does the substitution that is.

tests/test_config_parsing/test_res_config.py
```diff
@@ -114,3 +114,20 @@ def test_res_config_throws_on_missing_forward_model_job(config_dict):
...
+def test_that_non_bracketed_defines_warns(caplog):
+    with open("test.ert", "w", encoding="utf-8") as fh:
+        fh.write(
+            dedent(
+                """
+                NUM_REALIZATIONS  1
+                DEFINE A B
+                """
+            )
+        )
```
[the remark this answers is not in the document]
eivindjahren: Added!

mortalisk (approved):
DEFINE L ME
DEFINE G R
DEFINE T G
DEFINE M E

LGTM

PR #4429, opened 2022-12-05
Move high level tests to bdd specifications

eivindjahren:
Currently, it is difficult to understand what many of the tests are trying to achieve. Generally, the idea is to build in automated quality control. The problem is that code needs to be able to change, while the requirements are kept. 

The purpose of this PR is to make it easier to understand how to change the test when the code changes while still checking for the same defects. In order to do so, tests that check for high level requirements are commented with a [behavioral specifiation](https://en.wikipedia.org/wiki/Behavior-driven_development#Behavioral_specifications).

It then makes it obvious how to change the test when the code changes, and also makes it possible to extend the test towards end-to-end testing by adding steps or extending checks.

## Pre review checklist


tests/specification_tests/end_user_domain/test_history_matching.py
```diff
@@ -0,0 +1,87 @@
...
+            "--target-case",
+            "target",
+            "poly.ert",
+            "--port-range",
+            "1024-65535",
+        ],
+    )
+    run_cli(parsed)
+
+    # THEN I expect the variance to reduce with each step
```
Blunde1, line 45: This is testing "generalized variance", not individual variance elements. Individual variance elements are in fact not guaranteed to decrease after an update (due to existing or induced dependence), but the generalized variance is.

Blunde1: Mentioning it, because intuitively is is easy to think that individual variance elements should decrease, and the comment reinforces this (unfortunately wrong) intuition. The code is correct.

eivindjahren: So the attempt to add documentation revealed this misunderstanding, which I had no idea about reading it the first time.

eivindjahren:
This was discussed in our retrospective, but we have decided to go with another direction.

PR #4440, opened 2022-12-06
Use doctest for _prepare_x_axis test

eivindjahren:
Moves test of _prepare_x_axis to doctest

## Pre review checklist


src/ert/dark_storage/common.py
```diff
@@ -132,6 +132,17 @@ def _get_obs_data(key, obs) -> dict:
 
 
 def _prepare_x_axis(x_axis: List[Union[int, float, str, pd.Timestamp]]) -> List[str]:
+    """Converts the elements of x_axis of an observation to a string suitable
+    for json.
```
pinkwah, line 136:
```suggestion
    """Converts the elements of x_axis of an observation to a string suitable
    for json. If the elements are timestamps, convert to ISO-8601 format.
```

src/ert/dark_storage/common.py
```diff
@@ -132,6 +132,17 @@ def _get_obs_data(key, obs) -> dict:
...
+
+    >>> _prepare_x_axis([1, 2, 3, 4])
+    ['1', '2', '3', '4']
+
+    The x_axis can possibly be a timestamp in which case isoformat is used
```
pinkwah, line 141: Can delete this if you accept previous change.

tests/unit_tests/storage/test_extraction.py
```diff
@@ -15,26 +15,6 @@
 from ert.shared.storage import extraction
 
 
-@pytest.mark.parametrize(
```
pinkwah, line 18: We fundamentally disagree in the purpose of doctests and normal unit tests.

The code example in the docstring is meant for the reader to understand the behaviour of the program. Because it is for the user's eyes, it's nice that we automatically verify that this code is functional.

The unit tests are to test actual behaviour of the program.

The two concepts are disjoint, and as such I don't think it's a good idea to remove this entirely.

src/ert/dark_storage/common.py
eivindjahren: Done

src/ert/dark_storage/common.py
eivindjahren: Added

tests/unit_tests/storage/test_extraction.py
eivindjahren: The reason why I wanted to move this test to the docstring was that it was

1. Comprised of short examples
2. Tested a private member of the file, which should ideally be contained in it.3
3. Is just a unit test of non-user facing features. Ie. it belongs to the programmer domain.

I removed one of the examples in the test as it seemed superfluous.

I don't see a problem with the doctest achieving both quality control of the code and the documentation. On the contrary, that seems efficient. I am not suggesting we move any test to doctest unless it seems to be a nice fit.

Reasons to not convert to doctest would be:

1. The test does not work well as a documenting example
2. It tests user-facing behavior which needs several steps of quality control.

The first case is probably quit obvious why we would want to avoid, the documentation needs to remain good. However, in the second case, using doctest is simply not flexible enough to change with the code while keep testing the same use case. For instance,

```python
def test_that_clean_up_removes_the_dumb_file():
   run()
   assert os.exists("dumb.txt")
   clean_up()
   assert not os.exists("dumb.txt")
```
Is not a good fit for a doctest as it would become more difficult to keep testing the same thing when the code changes to eg.

```python
def test_that_clean_up_removes_the_dumb_file():
   run(clean_up=True)
   assert not os.exists("dumb.txt")
```

pinkwah (approved):

PR #4474, opened 2022-12-12
Provide feedback when using deprecated %d specifiers in RUNPATH

### LINKED ISSUE #4432: Add deprecation warning when parsing the RUNPATH keyword
In particular if the RUNPATH does not contain one or two %d . In accordance with the docs. 

https://github.com/equinor/ert/blob/main/docs/reference/configuration/keywords.rst 

```
 The RUNPATH keyword should give the name of the folders where the ECLIPSE
        simulations are executed. It should contain at least one %d specifier, which
        will be replaced by the realization number when ERT creates the folders.
        Optionally, it can contain one more %d specifier, which will be replaced by
        the iteration number.

        By default, RUNPATH is set to "simulations/realization-%d".

        *Example A:*

        ::

                -- Giving a RUNPATH with just one %d specifer.
                RUNPATH /mnt/my_scratch_disk/realization-%d

        *Example B:*

        ::

                -- Giving a RUNPATH with two %d specifers.
                RUNPATH /mnt/my_scratch_disk/realization-%d/iteration-%d

        The RUNPATH keyword is optional.
```

### LINKED ISSUE #4443: Make runpath file format consistent
There are multiple instances in the code where the runpath file format is replacing 
`realization-%d/iter-%d` with `realization-real_number/iter-iter_number` by using the following pattern:
```
                if run_path.count("%d") == 2:
                    s = run_path % (j, i)
                else:
                    s = run_path % (j)
```
we should try to avoid this by making sure that runpath file format throughout the codebase is of the form `realization-<IENS>/iter-<ITER>`

Related to #4432

andreas-el:
**Issue**
Resolves #4432
Resolves #4443

Updated docs
Updated deprecated syntax for RUNPATH in ert-files, tests and examples

## Pre review checklist


src/ert/_c_wrappers/enkf/model_config.py
```diff
@@ -40,11 +40,29 @@ def __init__(
...
+            logger.warning(
+                "RUNPATH keyword should use syntax "
+                f"`{self.DEFAULT_RUNPATH}` "
+                "instead of deprecated syntax "
+                f"`{runpath_format_string}`"
+            )
+        else:
+            self.runpath_format_string = runpath_format_string
+            logger.error(
+                "RUNPATH keyword should use syntax "
```
eivindjahren, line 60: The error message looks a bit contradicting. Do we want %d or not?.

src/ert/libres_facade.py
```diff
@@ -159,6 +159,9 @@ def get_number_of_iterations(self) -> int:
     def have_observations(self) -> bool:
         return self._enkf_main.have_observations()
 
+    def is_valid_runpath(self) -> bool:
+        return self._enkf_main.getModelConfig().is_valid_runpath()
```
eivindjahren, line 163: I think instead of making yet another method on `libres_facade` based on a stateful value in model_config, I would like `load_results_tool::is_valid_run_path` to be 
```python
def is_valid_run_path(self) -> bool:
  actives = self.facade.get_active_realizations()
  iter_0_paths + set(self.facade.get_run_paths(actives, 0))
  iter_1_paths + set(self.facade.get_run_paths(actives, 1))
  return len(iter_0_paths) == len(actives) and iter_0_paths != iter_1_paths
```

src/ert/_c_wrappers/enkf/model_config.py
andreas-el: Indeed. New suggestion!

src/ert/libres_facade.py
andreas-el: I do agree, but would like to discuss the details regarding this. :+1:

src/ert/gui/tools/load_results/load_results_tool.py
```diff
@@ -14,7 +14,7 @@ def __init__(self, facade):
         )
         self.__import_widget = None
         self.__dialog = None
-        self.setEnabled(self.is_valid_run_path())
+        self.setEnabled(True)
```
andreas-el, line 17: I set this to true, to allow for the test to actually get to evaluate if the runpath is valid

src/ert/_c_wrappers/enkf/model_config.py
```diff
@@ -40,11 +40,21 @@ def __init__(
...
+            logger.warning(
+                "RUNPATH keyword should use syntax "
+                f"`{self.DEFAULT_RUNPATH}` "
+                "instead of deprecated syntax "
+                f"`{runpath_format_string}`"
+            )
+        elif not any(x in runpath_format_string for x in ["<ITER>, <IENS>"]):
+            self.runpath_format_string = runpath_format_string
+            logger.error(
+                "RUNPATH keyword shall use syntax " f"`{self.DEFAULT_RUNPATH}`."
```
eivindjahren, line 56: Maybe `should` instead instead of `shall` ?

eivindjahren (changes requested):
I had one comment regarding the wording of a warning.

Currently, this shows "suggestions" via logger.warn. However, for some of these messages it would be great if we could show them in the suggester gui. Perhaps we should spawn a new issue for that?

src/ert/_c_wrappers/enkf/model_config.py
andreas-el: Sure thing.

And added new issue for suggester ui.
https://github.com/equinor/ert/issues/4543

eivindjahren (approved):

PR #4476, opened 2022-12-12
Fix a bug with fixed point substitution

### LINKED ISSUE #4473: New fixed point substitution breaks  forward model practice possibly copied by users from hm tutorial
In the hm tutorial, the RMS forward model would be called like this:

```
FORWARD_MODEL RMS(<IENS>=<IENS>...)
```

As we do substitutions now, the parameter gets put in a context where IENS is not defined yet, so it is not replaced.

Likely we would want to remedy the situation with two steps:

- [ ] Change substitute_real_iter to make a copy and then appends the per real/iter defines. This reduces the chance that fixed point replace fails. https://github.com/equinor/ert/blob/365be9f9891606af50d85c1aa82a6938e9c18f8b/src/ert/_c_wrappers/util/substitution_list.py#L117
- [ ] Swap the order of private arg and context substitutions in the job printing, and call substitute_real_iter for both https://github.com/equinor/ert/blob/365be9f9891606af50d85c1aa82a6938e9c18f8b/src/ert/_c_wrappers/job_queue/forward_model.py#L59-L61 . This reduces the chance that fixed point substitute fails. One can also create a copy of context, add the private defines to it and then call substitute which further minimizes problems.
- [ ] Provide a deprecation warning whenever a substitution of the form `<ITER>=<ITER>` is added to the substitution list.

eivindjahren:
**Issue**
Resolves #4473 

**Approach**

Adds a warning in case `<ITER>=<ITER>` or `<IENS>=<IENS>` is added to the argument list of a job, which is a common misperception of how the argument list works. In effect, we ignore this assignment. If the user has specified `<ITER>=1` in the argument list that gets priority. If nothing is inserted, then iteration number is inputed.

Also, `substitute_real_iter` is changed to avoid some non-terminating substitutions (the `<ITER>` substitution could be starved before).

## Pre review checklist

### DIFF
--- a/src/ert/_c_wrappers/job_queue/forward_model.py
+++ b/src/ert/_c_wrappers/job_queue/forward_model.py
@@ -1,9 +1,11 @@
+import copy
 import json
 import logging
 import os
 from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, List
+from warnings import warn
 
 from ert._c_wrappers.job_queue import EnvironmentVarlist, ExtJob
 from ert._clib import job_kw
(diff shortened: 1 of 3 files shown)


eivindjahren:
jenkins test ert please

src/ert/_c_wrappers/job_queue/forward_model.py
```diff
@@ -56,7 +58,38 @@ def get_job_data(
...
+                        f"Found <IENS>=<IENS> in {job.name} forward model."
+                        " This argument can safely be removed and setting it in this"
+                        " way is deprecated. Using for instance <IENS>=1 is still "
+                        " possible."
+                    )
+                    copy_private_args.addItem("<IENS>", str(iens))
+                if "<ITER>" not in copy_private_args:
+                    copy_private_args.addItem("<ITER>", str(itr))
+                if "<IENS>" not in copy_private_args:
+                    copy_private_args.addItem("<IENS>", str(iens))
```
DanSava, line 91: This is not really needed

src/ert/_c_wrappers/util/substitution_list.py
```diff
@@ -117,14 +118,13 @@ def substitute(self, to_substitute: str) -> str:
...
-            "<GEO_ID>", self.get(geo_id_key, "<GEO_ID>")
-        )
-        to_substitute = to_substitute.replace("<IENS>", str(realization))
-        to_substitute = to_substitute.replace("<ITER>", str(iteration))
-        return to_substitute
+        if geo_id_key in self:
+            copy_substituter.addItem("<GEO_ID>", self[geo_id_key])
+        copy_substituter.addItem("<IENS>", str(realization))
+        copy_substituter.addItem("<ITER>", str(iteration))
+        return copy_substituter.substitute(to_substitute)
```
DanSava, line 127: Are these changes guarding against recursive defines like 
```
DEFINE <ITER> <ITER>
```
If not then it is not really needed to fix the arglist substitution.  And if they are then what about guarding against 
```
DEFINE <KEY> <KEY>

```
?

DanSava (approved):
It fixes the recursive arg issue for ITER and IENS :+1:

src/ert/_c_wrappers/job_queue/forward_model.py
eivindjahren: Not really, but there are conditions where copy_private_args.substitute may not terminate if these are not added. Those uses are deprecated of course.

src/ert/_c_wrappers/util/substitution_list.py
eivindjahren: I know that sounds like I am paranoid, but we keep discovering real world edge cases, so I thought it would be better to stick with the way of doing it that was more intuitive to the user. I would argue that it is not intuitive for the user that `<IENS>` has a special case were it can get starved from being chosen for substitution. All other keys are chosen fairly (all given one go per round of substitutions).

PR #4496, opened 2022-12-16
Remove keyword RERUN_START

### LINKED ISSUE #4488: remove all traces of config keyword `RERUN_START` except for in suggester

mortalisk:
**Issue**
Resolves #4488

**Approach**

## Pre review checklist


asnyv:
Not sure if it is related or not (and maybe commenting on a PR is not the best way to do it), but there is a keyword `ENKF_RERUN` that I struggle a bit to understand what does. The documentation says:

> This is a boolean switch - TRUE or FALSE. Should the simulation start from time zero after each update.

The default is `False`, but what does that mean? My gut feeling is that it could previously be used to override this `RERUN_START` and that default was `False` to make `RERUN_START` work without having to set `ENKF_RERUN`, but that is just a guess. 
Do you know what `ENKF_RERUN` does (or did)?

mortalisk:
I think @eivindjahren has some insights for you @asnyv 
Also, we should try to update the documentation.

andreas-el:
Did grep for keywords and found no instances. Looks good to me. :+1:

eivindjahren:
> The default is `False`, but what does that mean? My gut feeling is that it could previously be used to override this `RERUN_START` and that default was `False` to make `RERUN_START` work without having to set `ENKF_RERUN`, but that is just a guess. Do you know what `ENKF_RERUN` does (or did)?

I have the following wording from 2010 of the `ENKF_RERUN` keyword: "ENKF_RERUN: Should the simulations be restarted from time zero after each update. " So I am guessing this is whether the simulator (such as OPM flow or eclipse) should use a restart feature or not. It has no effect any more an can be safely removed.

eivindjahren (approved):

PR #4514, opened 2022-12-23
Let delete_file delete broken symlinks

### LINKED ISSUE #4513: DELETE_DIRECTORY forward model job cannot delete directories with internal symlinks
**Describe the bug**
If a directory contains a symlink to a neighbouring file, like in 

```
$ ls -l to_be_deleted/
total 0
lrwxrwxrwx 1 havb havb 15 Dec 23 09:25 link -> link_target.txt
-rw-rw-r-- 1 havb havb  0 Dec 23 09:26 link_target.txt
```
~the DELETE_DIRECTORY job will fail as it will delete `link_target.txt` and then skip `link` and then
the directory is not empty when it finally tries to delete the directory.~
you might be unlucky that `os.walk()` will pick `link_target.txt` before `link`, and then delete `link_target.txt`. When it later in the deletion loop tries to delete `link`, it will do an `os.stat("link")` which will fail because `os.stat()` resolves symlinks.

**To reproduce**
```
def test_that_delete_directory_can_delete_directories_with_internal_symlinks(shell):
    shell.mkdir("to_be_deleted")
    link_target = Path("to_be_deleted/link_target.txt")
    link = Path("to_be_deleted/link")
    Path("to_be_deleted/link_target.txt").write_text("hei", encoding="utf-8")

    os.chdir("to_be_deleted")  # shell.symlink() requires this
    shell.symlink("link_target.txt", "link")
    os.chdir("..")

    assert Path("to_be_deleted/link").exists
    shell.delete_directory("to_be_deleted")
    assert not Path("to_be_deleted").exists
```

**Expected behaviour**
This directory should be deleted successfully.

**Environment**
 - OS: RHEL7
 - ERT/Komodo release: 4.2.0rc2
 - Python version: 3.8

berland:
Since os.stat will resolve symlinks, broken links will not be deleted unless os.lstat is used instead.

This causes issues with deletion of directories with internal symlink, if the link target is deleted prior to the link, the directory deletion will fail.

**Issue**
Resolves #4513 

**Approach**
Change delete_file() to not try to resolve symlinks

## Pre review checklist

### DIFF
--- a/tests/unit_tests/shared/share/test_shell.py
+++ b/tests/unit_tests/shared/share/test_shell.py
@@ -228,35 +228,44 @@ def test_move_pathfile_into_folder_file_exists(shell):
 
 
 @pytest.mark.usefixtures("use_tmpdir")
-def test_delete_file(shell):
+def test_that_delete_file_cannot_delete_directories(shell):
     shell.mkdir("pathx")
     assert b"not a regular file" in shell.delete_file("pathx").stderr
 
-    # deleteFile which does not exist - is silently ignored
-    shell.delete_file("does/not/exist")
 
-    with open("file", "w") as f:
-        f.write("hei")
-    shell.symlink("file", "link")
-    assert os.path.islink("link")
-
-    shell.delete_file("file")
-    assert not os.path.isfile("file")
-    assert os.path.islink("link")
-    shell.delete_file("link")
-    assert not os.path.islink("link")
+@pytest.mark.usefixtures("use_tmpdir")
+def test_that_delete_file_ignores_non_existing_files(shell):
+    assert b"ignored" in shell.delete_file("does/not/exist").stderr
 
 
 @pytest.mark.usefixtures("use_tmpdir")
-def test_delete_directory(shell):
-    # deleteDriecteory which does not exist - is silently ignored
-    shell.delete_directory("does/not/exist")
+def test_that_delete_file_deletes_broken_symlinks(shell):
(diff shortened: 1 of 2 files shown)


tests/unit_tests/shared/share/test_shell.py
```diff
@@ -228,35 +228,44 @@ def test_move_pathfile_into_folder_file_exists(shell):
...
     shell.delete_file("link")
-    assert not os.path.islink("link")
+    assert not Path("link").is_symlink() and not Path("link").exists()
 
 
 @pytest.mark.usefixtures("use_tmpdir")
-def test_delete_directory(shell):
-    # deleteDriecteory which does not exist - is silently ignored
-    shell.delete_directory("does/not/exist")
+def test_that_delete_non_existing_directory_is_silently_ignored(shell):
```
xjules, line 255: I would remove `that` from the test function names, which applies for other tests too.

xjules (changes requested):
Looks good! Just had a small suggestion wrt. naming

tests/unit_tests/shared/share/test_shell.py
berland: I added `_that_` based on the discussion and recommendations in the scout channel on slack around dec 5

xjules: Right, when so it can go as is :)

xjules (approved):
LGTM! :rocket:

PR #4526, opened 2022-12-30
Improve type hints of `_get_A_matrix`

dafeda:
## Pre review checklist


eivindjahren (approved):

eivindjahren (changes requested):
I have more of a question that should be resolved before merging: isn't the analysis module checked by the strict check? If it is, there is no need to check it twice.

dafeda:
> I have more of a question that should be resolved before merging: isn't the analysis module checked by the strict check? If it is, there is no need to check it twice.

Yes you are right. I have removed the extra check but kept the changes to type hints as I think they are better.

eivindjahren (approved):

PR #4529, opened 2023-01-02
Type-check ert.services

pinkwah:
Fixes typing issues with ert.services, allowing it to be automatically type-checked with mypy.

I use the `annotations` from `__future__` (ie, backporting features from future Python versions). This is a 3.11 feature that makes type annotations lazy. In particular, this means that we don't need to wrap types in quotes (so that they're string literals). This should have a marginal performance improvement as generic types are no longer instantiated. (Ie, `List` is a normal Python class, and you pay a start-up penalty for doing `List[str]` as it actually creates this object when the module is initialised, whether you actually use this type at runtime or not)

### DIFF
--- a/.mypy-strict.ini
+++ b/.mypy-strict.ini
@@ -5,7 +5,6 @@ exclude = (?x)(
     src/ert/logging
     | src/ert/shared
     | src/ert/dark_storage
-    | src/ert/services
     | src/ert/_c_wrappers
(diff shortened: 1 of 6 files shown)


oyvindeide (approved):
LGTM!

pinkwah:
Jenkins test this please

PR #4534, opened 2023-01-02
Type-check `ert.data`

pinkwah:
**Achtung!** This deletes some `@deprecated` functions that I cannot find in use anywhere else other than in our test suite. I have checked current semeio, equilibrium, subscript, and everest.


oyvindeide (approved):
LGTM!

PR #4535, opened 2023-01-02
Move ens path from model config to res config

### LINKED ISSUE #4497: Verify new behavior when ens-path is None
Possible introduction of a bug: 

After https://github.com/equinor/ert/pull/4434 -> In the old code it seems like ens-path had a default `DEFAULT_ENSPATH = "storage"`. Currently we are using ` ens_path: str = None` ( in `model_config.py` ). With that i then expect this line to end up in trouble  `(Path(self.getModelConfig().ens_path) / "current_case").write_text(case_name)` ( in `enkf_main.py` -> `switchFileSystem(..)` )

We should look into if we have changed behavior here or if it is actually OK as-is.

valentin-krasontovitsch:
**Issue**
Resolves #4497

**Approach**
See commit message

## Pre review checklist

### DIFF
--- a/src/ert/__main__.py
+++ b/src/ert/__main__.py
@@ -60,7 +60,7 @@ def run_webviz_ert(args):
     if args.config:
         res_config = ResConfig(args.config)
         os.chdir(res_config.config_path)
-        ens_path = res_config.model_config.ens_path
+        ens_path = res_config.ens_path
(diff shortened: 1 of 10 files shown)


andreas-el (approved):
Looks good to me! :suspect:

PR #4567, opened 2023-01-06
Let delete_file() in delete_directory.py delete symlinks

berland:
Since os.stat will resolve symlinks, broken links will not be deleted unless os.lstat is used instead.

Backport of #4514, solves #4513


xjules (approved):
:rocket:

Issue #4566, opened 2023-01-06
Fixed point substitution fails for the forward model keyword


eivindjahren:
Testing of the current release candidate revealed that the following now fails with hitting the substitution limit:
```
DEFINE <RMS_VERSION> 1.2.3
FORWARD_MODEL RMS(<RMS_VERSION>=<RMS_VERSION>)
```

The issue is that the expected behavior of the `FORWARD_MODEL` keyword is like that of an argument list, but in reality the above situation is equivalent to 

```
DEFINE <RMS_VERSION> 1.2.3
FORWARD_MODEL RMS(1.2.3=1.2.3)
```

this has coincidentally worked before the introduction of fixed point substitution as when the forward model was to determine the value of its parameter `<RMS_VERSION>` the following substitutions were available:
```
DEFINE <RMS_VERSION> 1.2.3
DEFINE 1.2.3 1.2.3
```
However, with fixed point substitutions this results in a loop.

One way to solve this, is to turn off substitutions on the line containing the `FORWARD_MODEL` keyword so that the left hand side in the assignments are not substituted, which seems to be what the user expects. This would result in the following substitutions being available for the parameters:
```
DEFINE <RMS_VERSION> 1.2.3
DEFINE <RMS_VERSION> <RMS_VERSION>
```
which does terminate in our fixed point substitution (it would not work in a substitution scheme that does not ensure liveness, but that would create other problems as well).

valentin-krasontovitsch:
question: what's liveness?

eivindjahren:
https://en.wikipedia.org/wiki/Safety_and_liveness_properties

PR #4576, opened 2023-01-06
Provide better error messages for subst failures

eivindjahren:
**Issue**
Because of #4566 , we found out that we need better error messages.

## Pre review checklist

### DIFF
--- a/src/clib/lib/res_util/subst_list.cpp
+++ b/src/clib/lib/res_util/subst_list.cpp
@@ -3,6 +3,8 @@
 #include <ctype.h>
 #include <stdlib.h>
 #include <string.h>
+#include <string>
+#include <vector>
 
 #include <ert/res_util/file_utils.hpp>
 #include <ert/util/buffer.hpp>
@@ -12,6 +14,7 @@
 #include <ert/util/vector.hpp>
 
 #include <ert/res_util/subst_list.hpp>
+#include <fmt/format.h>
 
 namespace fs = std::filesystem;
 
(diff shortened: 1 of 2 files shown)


valentin-krasontovitsch (changes requested):
can we make it better? give more context? like the whole line?

valentin-krasontovitsch:
looking at the code, i see that in `config_parser.cpp`, we try to catch the define error, and append it to the parse errors list - but in testing a run, ert just crashes with the runtime error rather than reporting a parse error. is this because i'm trying to create the infinite loop inside arguments of the forward model keyword?

valentin-krasontovitsch:
python cannot catch a runtime error thrown in C++ code, can it?

oysteoh:
Seems like it works as intended. My only input would be to slightly modify of of the messages to make it more clear, but i would also accept the current message as good enough :)

src/clib/lib/res_util/subst_list.cpp
```diff
@@ -359,8 +361,9 @@ char *subst_list_alloc_filtered_string(const subst_list_type *subst_list,
         }
 
         if (iterations >= max_iterations) {
-            throw std::runtime_error(
-                "Reached max iterations while trying to resolve defines");
+            throw std::runtime_error(fmt::format(
+                "Reached max iterations while trying to resolve defines in {}",
```
oysteoh, line 365: Feel this sentence is not fully explaining what is going on - what about `what(): Reached max iterations while trying to resolve DEFINE for string <RMS_PROJECT>`

oysteoh (approved):

oysteoh:
Very nice! :rocket:

eivindjahren:
> python cannot catch a runtime error thrown in C++ code, can it?

Yes, this is the general problem. Attempting to communicate the error up the chain is going to be painful so I think CLI is as good as it gets.

valentin-krasontovitsch:
turns out if we expose the c/c++ functions through pybind, and throw c++ exceptions, the exceptions get automatically propagated to python and can be caught and handled there. thanks to @pinkwah for the insights! i've created a separate PR where we can try to add some context to the error from further up the calling - #4581

valentin-krasontovitsch (approved):
looks good!

oysteoh:
Jenkins: Test this please!

PR #4582, opened 2023-01-06
Warn when QUEUE_OPTION MAX_RUNNING is invalid / negative

### LINKED ISSUE #4326: You get no warning if you include QUEUE_OPTION LOCAL MAX_RUNNING -1 in your config, it just doesn't start any runs.

andreas-el:
**Issue**
Resolves #4326

This will be a candidate for the upcoming validation improvement.

## Pre review checklist

### DIFF
--- a/src/ert/_c_wrappers/enkf/res_config.py
+++ b/src/ert/_c_wrappers/enkf/res_config.py
@@ -213,6 +213,16 @@ class ResConfig:
         site_config_content = site_config_parser.parse(site_config_location())
         return config_content_as_dict(site_config_content, {})
 
+    def _validate_queue_option_max_running(self, config_path, config_dict):
+        for _, option_name, *values in config_dict.get("QUEUE_OPTION", []):
+            if option_name == "MAX_RUNNING" and int(*values) < 0:
+                raise ConfigValidationError(
+                    config_file=config_path,
+                    errors=[
+                        f"QUEUE_OPTION MAX_RUNNING is negative: {str(*values)}",
+                    ],
+                )
+
     # build configs from config file or everest dict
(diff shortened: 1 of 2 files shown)


src/clib/lib/job_queue/queue_driver.cpp
```diff
@@ -85,6 +85,11 @@ static bool queue_driver_set_generic_option__(queue_driver_type *driver,
         if (strcmp(MAX_RUNNING, option_key) == 0) {
             int max_running_int = 0;
             if (util_sscanf_int(value, &max_running_int)) {
+
+                if (max_running_int < 0) {
```
andreas-el, line 89: This should have no effect when the validation is done in res_config (optionally in queue_config).
Should remove?

eivindjahren: max_running_int = 0 should mean unlimited so something seems off.

eivindjahren: Seems a bit off as max_running = 1 is generally not we want, but rather max_running=0 (unlimited).

eivindjahren (changes requested):

src/clib/lib/job_queue/queue_driver.cpp
andreas-el: I have no objections to this. Question is if this should be removed altogether, since the validation should in theory make sure this will never go into effect.

eivindjahren: Yea, I think remove all together is the best. Adding sanitation to validation just complicates things.

eivindjahren (approved):

PR #4585, opened 2023-01-09
Change infinite substitution loop errors to warn

### LINKED ISSUE #4566: Fixed point substitution fails for the forward model keyword

eivindjahren:
Because of how FORWARD_MODEL works, several users reported issues with uses ala.

   DEFINE <VERSION> 3
   FORWARD_MODEL job(<VERSION>=<VERSION)

which is a widespread use pattern, which coincidentally works. It is actually interpreted as

   DEFINE <VERSION> 3
   FORWARD_MODEL job(3=3)

and it works because there is no scoping of <VERSION> and it is picked up by the job. Until we have a clear course of action we are assuming that infinite loops in defines is due to defines such as 3=3, and is up to the user handle DEFINE resolution.

**Issue**
Resolves #4566

### DIFF
--- a/src/clib/lib/config/config_parser.cpp
+++ b/src/clib/lib/config/config_parser.cpp
@@ -128,18 +128,9 @@ static config_content_node_type *config_content_item_set_arg__(
             int iarg;
             for (iarg = 0; iarg < argc; iarg++) {
 
-                try {
-                    char *filtered_copy = subst_list_alloc_filtered_string(
-                        define_list, stringlist_iget(token_list, iarg + 1));
-                    stringlist_iset_owned_ref(token_list, iarg + 1,
-                                              filtered_copy);
-                } catch (std::runtime_error err) {
-                    std::string error_message = util_alloc_sprintf(
-                        "Could not resolve defines in %s. Defines might have "
-                        "an infinite loop",
-                        stringlist_iget(token_list, iarg + 1));
-                    parse_errors.push_back(error_message);
-                }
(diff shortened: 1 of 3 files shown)


valentin-krasontovitsch (approved):
yasssssss

PR #4609, opened 2023-01-11
Dropdown warn when no observations

### LINKED ISSUE #4606: If no observation, show a warn icon besides the disabled dropdown with info instead of the text in the suggestor

### LINKED ISSUE #4616: Starting a case with empty observation file causing segmentation fault
- ert gui poly.ert

ERT gui is opened

- empty the observations file
- ert gui poly.ert

Segmentation fault

I guess we want at worst an error message which makes sense - at best i want the same behavior as if there is no OBS_CONFIG defined in the config-file

andreas-el:
**Issue**
Resolves #4606
Resolves #4616 

Disabled simulation algorithms that perform updates when no observations are found.
Show warn icon on disabled choices.
Show tooltip on disabled choices.
Show warning icon next to combobox.
Show tooltip on warning icon.
Removed suggestor feedback.
Validate that observations file actually contain something
Added unit-test to verify exception thrown.

Yes, -- i'll squash.

screenshot3

**Approach**

## Pre review checklist

### DIFF
--- a/src/ert/_c_wrappers/enkf/enkf_main.py
+++ b/src/ert/_c_wrappers/enkf/enkf_main.py
@@ -152,6 +152,15 @@ class EnKFMain:
             config.ensemble_config,
         )
         if config.model_config.obs_config_file:
+            if (
+                os.path.isfile(config.model_config.obs_config_file)
+                and os.path.getsize(config.model_config.obs_config_file) == 0
+            ):
+                raise ValueError(
+                    f"Empty observations file: "
+                    f"{config.model_config.obs_config_file}"
+                )
+
             if self._observations.error:
                 raise ValueError(
                     f"Incorrect observations file: "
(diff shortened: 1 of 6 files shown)


andreas-el:
Jenkins test this please!

oysteoh:
Just to check i wanted to use an empy observations file instead of removing OBS_CONFIG - what i expect then is the same behavior as when removing OBS_CONFIG. The file should be found but there are no observations so the model update algorithms should be disabled. What i get is a segmentation fault and crash. Is it possible to look into what causing this? I guess it should at least end up in an informative message and not explosion 😅

eivindjahren (approved):
Genious! I think we should backport this to Version-4.2 as well to show that we take suggestions by testers seriously.

eivindjahren:
style is complaining about `test_es_update` although I am not sure exactly why.

PR #4614, opened 2023-01-11
Add matching list for field transformations

### LINKED ISSUE #4070: Validate unknown function in init transform argument to enkf config nodes early (applies to FIELD config key and perhaps others)
One argument to config keys that define enkf config nodes (which end up in ensemble config, so related to #4254) is an init transform, and there a function can be specified. if that function cannot be found, a util abort is triggered.

we should catch this error in python and return a meaningful error message.

a corresponding test is already defined (and skipped with reference to this issue) in `tests/test_config_parsing/test_ensemble_config.py`.

---

- [ ] Update docs with information on all available init & input and output transformations

andreas-el:
**Issue**
Resolves #4070 

**Approach**
- Synchronized text found in docs describing FIELD and its arguments (INIT, INPUT and OUTPUT)
- Add verification of function names, as a proactive step
- Avoid repetition of code, fetch names from one place
- Removed Rubakumar specials, -- arguably wrong, misleading or both

## Pre review checklist

### DIFF
--- a/src/clib/lib/enkf/field_trans.cpp
+++ b/src/clib/lib/enkf/field_trans.cpp
@@ -153,49 +153,6 @@ void field_trans_table_free(field_trans_table_type *table) {
     free(table);
 }
 
-/* Rubakumar specials: start  */
-#define PERMX_MEAN 100
-#define PERMX_STD 1
-
-#define PERMZ_MEAN 100
-#define PERMZ_STD 1
-
-#define PORO_MEAN 100
-#define PORO_STD 1
-
-static float normalize(float x, float mean, float std) {
-    return (x - mean) / std;
-}
-
-static float denormalize(float x, float mean, float std) {
-    return (x * std) + mean;
-}
-
-static float normalize_permx(float x) {
-    return normalize(x, PERMX_MEAN, PERMX_STD);
-}
-
-static float denormalize_permx(float x) {
-    return denormalize(x, PERMX_MEAN, PERMX_STD);
-}
-
-static float normalize_permz(float x) {
-    return normalize(x, PERMZ_MEAN, PERMZ_STD);
-}
-
-static float denormalize_permz(float x) {
-    return denormalize(x, PERMZ_MEAN, PERMZ_STD);
-}
-
-static float normalize_poro(float x) {
-    return normalize(x, PORO_MEAN, PORO_STD);
-}
-
-static float denormalize_poro(float x) {
-    return denormalize(x, PORO_MEAN, PORO_STD);
-}
-/* Rubakumar specials: end  */
-
 static float field_trans_pow10(float x) { return powf(10.0, x); }
 
 static float trunc_pow10f(float x) {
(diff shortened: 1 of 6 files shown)


src/ert/_c_wrappers/enkf/config/enkf_config_node.py
```diff
@@ -406,6 +410,25 @@ def __repr__(self):
...
+            "EXP",
+            "TRUNC_POW10",
+            "LN0",
+            "DENORMALIZE_PORO",
+            "NORMALIZE_PERMZ",
+            "NORMALIZE_PERMX",
+            "DENORMALIZE_PERMZ",
+            "POW10",
+        ]:
+            raise ValueError(f"FIELD {field_name}:{func_name} is an invalid function")
```
valentin-krasontovitsch, line 430: looks great! only thing i would like to ask about is if it's at all possible to get the list of functions from where they're defined in the C code - or would that be too much trouble to become a bit DRYer?

valentin-krasontovitsch (approved):

src/ert/_c_wrappers/enkf/config/enkf_config_node.py
andreas-el: No this is definitely something i already considered doing.
I guess i also should update the docs while i'm at it. The list of available functions wasn't fully complete either.
I'll give it a shot!

src/clib/lib/enkf/field_trans.cpp
```diff
@@ -239,16 +196,6 @@ field_trans_table_type *field_trans_table_alloc() {
                           "This function will calculate y = exp(x) - 0.000001",
                           field_trans_exp0);
 
-    // Rubakumar specials:
-    field_trans_table_add(table, "NORMALIZE_PERMX", "...", normalize_permx);
-    field_trans_table_add(table, "DENORMALIZE_PERMX", "...", denormalize_permx);
```
eivindjahren, line 244: Do we have any way of knowing whether the "Rubakumar specials" are actually used?

andreas-el: That would have to be looking at the azure logs i guess? Can we search the configurations used/printed there to check that?

andreas-el: I find no hits on azure logs when searching for 'NORMALIZE'. This should catch all Rubakumar specials if they were executed on azure. I'm not sure how far back the logs do go though.
I've verified the search pattern by looking for other valid functions, namely  POW, LOG etc. They are used quite extensively.

As a side note, INPUT_TRANSFORM is present in the code, but not found on azure, nor mentioned in the FMU docs. Not to be mixed with INIT_TRANSFORM which is in use.

andreas-el: Frode wrote on slack: Yes. I did not implement those transformation for the same reasons mentioned above

eivindjahren: I think we can say that this is generally not used then, and in any case we already had a plan to remove it.

src/ert/_c_wrappers/enkf/config/enkf_config_node.py
```diff
@@ -27,6 +27,19 @@
 logger = logging.getLogger(__name__)
 
 
+def fieldFunctionNameList() -> List[str]:
```
eivindjahren, line 30: Maybe use a constant list for this?

```python
FIELD_FUNCTION_NAMES = [
        "LN",
        "LOG10",
        "EXP0",
        "LOG",
        "EXP",
        "TRUNC_POW10",
        "LN0",
        "POW10",
    ]
```

eivindjahren (changes requested):

src/ert/_c_wrappers/enkf/config/enkf_config_node.py
andreas-el: Definitely. :+1:

eivindjahren (approved):

PR #4663, opened 2023-01-17
Avoid util_abort when INCLUDEing non existing file in config

### LINKED ISSUE #4640: Using a include statement with a missing file does not produce an error message in the suggester
When the included file does not exist, the INCLUDE statement will add an error to the list but then attempt to open and parse it anyways leading to an abort.

https://github.com/equinor/ert/blob/3eeb38fd43ffa694191abc6c85769ca8dfc9e818/src/clib/lib/config/config_parser.cpp#L546-L555
5

DanSava:
**Issue**
Resolves #4640

**Approach**
Do not parse non-existing include file from the config file and throw `ConfigValidationError` if non existing file is included. 

## Pre review checklist

### DIFF
--- a/src/clib/lib/config/config_parser.cpp
+++ b/src/clib/lib/config/config_parser.cpp
@@ -547,11 +547,11 @@ config_parse__(config_parser_type *config, config_content_type *content,
                     std::string error_message = util_alloc_sprintf(
                         "%s file:%s not found", include_kw, include_file);
                     content->parse_errors.push_back(error_message);
+                } else {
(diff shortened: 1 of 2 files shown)


eivindjahren (approved):
We should probably backport this, although it is not a blocker for the release.

DanSava:
> We should probably backport this, although it is not a blocker for the release.

I don't think we introduced this issue with out config changes so I am not really sure why would we need to backport this. It can wait an be added to the next release

PR #4664, opened 2023-01-17
Stop substituting in FORWARD_MODEL argument names

### LINKED ISSUE #4634: Make FORWARD_MODEL actually do what the documentation says
@eivindjahren  please add to this issue sub-issues if possible and also the background and intended way of solving it when you are done discussing possible solutions.

This epic should have issues both related to solving the argument-list, resolving of keys and also validation

mortalisk:
**Issue**
Resolves #4634

**Approach**
Disable substitutions when parsing FORWARD_MODEL. Instead do substitutions only on the private arg values, first with global defines and then with private defines. This should cause the argument list to behave a little more like one would expect according to the documentation and general common sense.

## Pre review checklist


eivindjahren (approved):
:brain:

PR #4683, opened 2023-01-19
Remove nonsensical error when executable does not exist

### LINKED ISSUE #4673: Missing workflow job executable may result in error message "Path to executable: {exec.sh} is actually a directory"

DanSava:
**Issue**
Resolves #4673

**Approach**
Remove the code that generates the confusing error message now the error will report that the executable is not found if the executable value is not properly formatted. 

## Pre review checklist

### DIFF
--- a/src/clib/lib/config/config_schema_item.cpp
+++ b/src/clib/lib/config/config_schema_item.cpp
@@ -409,19 +409,6 @@ bool config_schema_item_validate_set(const config_schema_item_type *item,
 
                         free(relocated);
 
-                        if (strstr(value, "/")) {
-                            /*
-                             * value not naked name, which means it was relative path
-                             * that wasn't found
-                             */
-
-                            error_list.push_back(std::string(
-                                util_alloc_sprintf("Path to executable:%s is "
-                                                   "actually a directory",
-                                                   value)));
-                            break;
-                        }
-
                         /*
                          * res_env_alloc_PATH_executable aborts if some parts of the path is
                          * not an existing dir, so call it only when its an absolute path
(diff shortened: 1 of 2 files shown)


oysteoh:
Is it possible to reproduce this running the GUI today? I'm not able to manipulate the config file to trigger this behavior :thinking:

andreas-el (approved):
Sounds good to me! :guitar: :sound:

src/clib/lib/config/config_schema_item.cpp
```diff
@@ -409,19 +409,6 @@ bool config_schema_item_validate_set(const config_schema_item_type *item,
 
                         free(relocated);
 
-                        if (strstr(value, "/")) {
-                            /*
-                             * value not naked name, which means it was relative path
-                             * that wasn't found
-                             */
-
-                            error_list.push_back(std::string(
```
eivindjahren, line 418: Looking at the history, I may have messed this message up. I think perhaps that the code works in such a way that if value contains "/" then the executable does not exist. If I follow the logic then it is supposedly a relative path that wasn't found. Is that still covered?

DanSava: If there is a relative path that is not found or not fully resolved yet when the check is running then I guess an abort will happen. 
```
                        /*
                         * res_env_alloc_PATH_executable aborts if some parts of the path is
                         * not an existing dir, so call it only when its an absolute path
                         */
                        char *path_exe = res_env_alloc_PATH_executable(value);
```

DanSava:
> Is it possible to reproduce this running the GUI today? I'm not able to manipulate the config file to trigger this behavior thinking

I could not reproduce it using the GUI either but I managed to do it with the test that I changed in this PR.

src/clib/lib/config/config_schema_item.cpp
eivindjahren: Can we somehow provoke such an abort in a test?

DanSava: I am struggling a bit to do it but the execution logic for this piece of code is a bit complex and it is taking some time

PR #4688, opened 2023-01-19
Add startup suggestion for removing deprecated keyword PLOT_SETTINGS

### LINKED ISSUE #4658: Add 'PLOT_SETTINGS' to suggester
As per https://wiki.equinor.com/wiki/index.php/Debugging , the old deprecation message for PLOT_SETTINGS is confusing. It has been removed so now the error message is "Unrecognized keyword". Add a better message to suggester.

mortalisk:
**Issue**
Resolves #4658

This should give a user confidence to take action towards removing the keyword.

## Pre review checklist

### DIFF
--- a/src/ert/_c_wrappers/enkf/_deprecation_migration_suggester.py
+++ b/src/ert/_c_wrappers/enkf/_deprecation_migration_suggester.py
@@ -22,7 +22,12 @@ class DeprecationMigrationSuggester:
         "SCHEDULE_PREDICTION_FILE",
         "MULTFLT",
     ]
-    JUST_REMOVE_KEYWORDS = ["UMASK", "LOG_FILE", "LOG_LEVEL", "ENKF_RERUN"]
+    JUST_REMOVE_KEYWORDS = [
+        "UMASK",
+        "LOG_FILE",
+        "LOG_LEVEL",
+        "ENKF_RERUN",
+    ]
     RSH_KEYWORDS = ["RSH_HOST", "RHS_COMMAND"]
 
     def _add_deprecated_keywords_to_parser(self):
@@ -32,6 +37,7 @@ class DeprecationMigrationSuggester:
             self._parser.add(kw)
         for kw in self.RSH_KEYWORDS:
             self._parser.add(kw)
+        self._parser.add("PLOT_SETTINGS")
         self._parser.add("HAVANA_FAULT")
         self._parser.add("REFCASE_LIST")
         self._parser.add("RFTPATH")
@@ -122,6 +128,13 @@ class DeprecationMigrationSuggester:
             "DELETE_RUNPATH",
             "The DELETE_RUNPATH keyword would clear the runpath directories "
             "between runs. It was removed in 2017 and no longer has any effect.",
+        )
+        add_suggestion(
+            "PLOT_SETTINGS",
+            f"The keyword {kw} no longer has any effect, and can be safely removed."
+            " The PLOT_SETTINGS keyword was once used to set up parameters for"
+            " visualising results after running ERT. This is now done in the graphical"
+            " user interface instead.",
         )
 
         return suggestions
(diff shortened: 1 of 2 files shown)


src/ert/_c_wrappers/enkf/_deprecation_migration_suggester.py
```diff
@@ -22,7 +22,12 @@ def __init__(self, parser: "ConfigParser"):
         "SCHEDULE_PREDICTION_FILE",
         "MULTFLT",
     ]
-    JUST_REMOVE_KEYWORDS = ["UMASK", "LOG_FILE", "LOG_LEVEL", "ENKF_RERUN"]
+    JUST_REMOVE_KEYWORDS = [
+        "UMASK",
+        "LOG_FILE",
+        "LOG_LEVEL",
+        "ENKF_RERUN",
```
oysteoh, line 29: shouldn't `PLOT_SETTINGS` have an entry in this list?

eivindjahren: The keywords in that list is for those keywords we have nothing further t say about it than it can be removed.

eivindjahren:
The `PLOT_SETTINGS` keyword was made to be note recognized in sep 2020 (ert version 2.15 2bc412637a03e27c9d390a150e911670bbff6929)

eivindjahren:
It was made deprecated in 2019-10-09,  b394480403 .

eivindjahren:
Found this documentation about how it was used back in 2017:
```
.. _plot_settings:
.. topic:: PLOT_SETTINGS

        The :code:`PLOT_SETTINGS` keyword is a "master keyword" which
        can be used to configure some aspects of the plotting. These
        settings will affect the default behaviour when you create a
        new plot, you can still changes these settings interactively.

        When using the :code:`PLOT_SETTINGS` keyword you supply a
        secondary keyword and a values as the tow arguments:

        ::

           PLOT_SETTINGS SHOW_REFCASE False

        Will make sure that your plots are created without the refcase
        plotted as default. The available secondary keys are:

        SHOW_REFCASE : Default True
        SHOW_HISTORY : Default True     
```

eivindjahren:
Use of `PLOT_SETTINGS SHOW_HISTORY` had an effect april 2019: 9c2c70e5ae42ecd087e65f46afee1443d6154b6a

src/ert/_c_wrappers/enkf/_deprecation_migration_suggester.py
```diff
@@ -123,5 +129,12 @@ def add_suggestion(kw, suggestion):
             "The DELETE_RUNPATH keyword would clear the runpath directories "
             "between runs. It was removed in 2017 and no longer has any effect.",
         )
+        add_suggestion(
+            "PLOT_SETTINGS",
+            f"The keyword {kw} no longer has any effect, and can be safely removed."
```
eivindjahren, line 134: Maybe we should say "The keyword PLOT_SETTINGS was removed in 2019 and has no effect. It can safely be removed. It was used for controlling settings for outputting qc plots to disk and what plots were shown in the GUI. All plots can now be selected on request from the GUI, and there is also an alternative view accessible by using `ert viz` in the commandline.

andreas-el: I like how the feedback actually gives back some information regarding when it was removed, and also what the original purpose was. If we're going nitpick on this, the feedback should be reversed, with explanation first, and wrapping it up with "It was removed in year-no, no longer has any effect and can be safely removed."

And maybe go for either f-string kw substitution or hard-coded string in the next?

eivindjahren: Maybe, but I also think stating the most relevant information for the user first may help them action on it quicker.

eivindjahren (approved):
:safety_vest:

PR #4699, opened 2023-01-23
Remove special casing from AnalysisConfig

mortalisk:
While working on a parsing prototype, found that there is some code paths that in only used in testing. Removing those and updating tests.

## Pre review checklist

### DIFF
--- a/src/ert/_c_wrappers/enkf/analysis_config.py
+++ b/src/ert/_c_wrappers/enkf/analysis_config.py
@@ -49,11 +49,7 @@ class AnalysisConfig:
 
     def _copy_modules(self):
         for element in self._analysis_copy:
-            if isinstance(element, list):
-                src_name, dst_name = element
-            else:
-                src_name = element[ConfigKeys.SRC_NAME]
-                dst_name = element[ConfigKeys.DST_NAME]
+            src_name, dst_name = element
 
             module = self._modules.get(src_name)
             if module is not None:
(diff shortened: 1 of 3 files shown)


src/ert/_c_wrappers/enkf/analysis_config.py
```diff
@@ -49,11 +49,7 @@ def __init__(
 
     def _copy_modules(self):
         for element in self._analysis_copy:
-            if isinstance(element, list):
-                src_name, dst_name = element
-            else:
-                src_name = element[ConfigKeys.SRC_NAME]
-                dst_name = element[ConfigKeys.DST_NAME]
+            src_name, dst_name = element
```
andreas-el, line 52: I take it that we should always get lists of tuples from now on.
But should we then express this in the __init__ using optionals ?

``` python
def __init__(
        self,
        ...
        analysis_copy=None,
        analysis_set_var=None,
        analysis_select=None,
    ):
```

andreas-el (changes requested):

src/ert/_c_wrappers/enkf/analysis_config.py
mortalisk: what() ?

andreas-el: Discussed and was decided not to follow up on.

andreas-el (approved):
Shoot for the stars. :rocket:

PR #4698, opened 2023-01-23
Catch only ConfigValidationError in ERT GUI

DanSava:
**Issue**
Resolves #4691 

## Pre review checklist

- [x] Added appropriate release note label
- [x] PR title captures the intent of the changes, and is fitting for release notes.
- [x] Commit history is consistent and clean, in line with the [contribution guidelines](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md).

Adding labels helps the maintainers when writing release notes. This is the [list of release note labels](https://github.com/equinor/ert/labels?q=release-notes).

PR #4598, opened 2023-01-10
Fix runpath format displayed in QT GUI

DanSava:
**Issue**
Resolves #4577 

**Approach**
Format the runpath for all the simulation modes

## Pre review checklist

- [x] Added appropriate release note label
- [x] PR title captures the intent of the changes, and is fitting for release notes.
- [x] Commit history is consistent and clean, in line with the [contribution guidelines](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md).

Adding labels helps the maintainers when writing release notes. This is the [list of release note labels](https://github.com/equinor/ert/labels?q=release-notes).

PR #4709, opened 2023-01-24
Backporting missing bug fixes 

DanSava:
**Issue**
Backporting:
#4598 
#4698


andreas-el (approved):
Changes look good! :+1:

andreas-el:
The everest failure is likely due to newer changes applied to everest which is not present in this release.

PR #4716, opened 2023-01-24
Catch mismatch between expected and given arguments for forward model jobs

### LINKED ISSUE #4694: Config Validation - Catch mismatch between expected and given arguments for forward model jobs
# background

tracked by #4599 

Forward model jobs are given in the command line according to the following format:

```
FORWARD_MODEL JOB_NAME(<ARG_A>=VALUE_A, <ARG_B>=VALUE_B)
```

where `JOB_NAME` refers to a job installed by a certain mechanism, and the key value pairs inside parantheses get parsed to a so-called "private args" substitution list.

the job referred to by name is defined through a job file that for instance looks like this:

```
$ cat src/ert/shared/share/ert/forward-models/old_style/COPY_FILE
STDOUT      copy_file.stdout
STDERR      copy_file.stderr
EXECUTABLE  ../../shell_scripts/copy_file.py
ARGLIST     <FROM> <TO>

MIN_ARG    2
MAX_ARG    2
ARG_TYPE 0 STRING
ARG_TYPE 1 STRING
```

observe the `ARGLIST` line. the arg list supplied on said line gets passed as arguments to the executable referenced by `EXECUTABLE`, after applying substitutions. We apply both substitutions from the so-called global scope, i.e. substitutions made up of predefined defines and defines parsed from the config file.

## goals

validate that the correct number of arguments is given in the forward model definition, compared to the job definition file.

this means we have to make sure that after applying all substitutions, there are no more elements of the form `<SOMETHING>` in the arg list.

also, it might be sensible to alert the user when a key value pair supplied in the forward model has no effect on the arg list, as that is probably not intended.

to summarize:

- [x] validate that enough substitutions are available to resolve all expected substitutions in the arg list of the job definition file
- [x] check that all supplied private args are used / necessary / have an effect

when we discover an unresolved argument, or a private arg without that does not resolve, we issue a warning.

## caveats / questions

- arg which uses pointy brackets and is not intended for substitution

are these even permitted? -> No! We catch these by interpreting them as unresolved intended substitutions, and report them as warnings to the user

- arg resolved both by global and private substitution

presumably, as long as something resolves the to be substituted strings, we're happy? -> yes, we are. we don't mind who or what does the job, as long as it gets done.

- some jobs have defaults, e.g.

```
$ cat src/ert/shared/share/ert/forward-models/old_style/ECLIPSE100
EXECUTABLE ../res/script/ecl100.py
DEFAULT <VERSION> version
DEFAULT <NUM_CPU> 1
DEFAULT <OPTS> ""
ARGLIST <ECLBASE> -v <VERSION> -n <NUM_CPU> <OPTS>
```
should make sure to consider that

valentin-krasontovitsch:
**Issue**
Resolves #4694 

**Approach**
We warn if

- any magic string in the arg list of a job referenced by a forward model could not be resolved with substitutions or defauklts, and
- a private arg key=value substitution is given in the forward model definition that has no effect 

## Pre review checklist

### DIFF
--- a/src/ert/_c_wrappers/job_queue/ext_job.py
+++ b/src/ert/_c_wrappers/job_queue/ext_job.py
@@ -10,6 +10,8 @@ from ert._c_wrappers.config import ConfigParser, ConfigValidationError, ContentT
 from ert._c_wrappers.util import SubstitutionList
 from ert._clib.job_kw import type_from_kw
 
+_SUBSTITUTED_AT_EXECUTION_TIME: List[str] = ["<ITER>", "<IENS>"]
+
 
 class ExtJobInvalidArgsException(BaseException):
     pass
@@ -233,6 +235,39 @@ class ExtJob:
         )
 
     def validate_args(self, context: SubstitutionList) -> None:
+        self._validate_all_passed_private_args_have_an_effect()
+        self._validate_all_magic_string_in_arglist_get_resolved(context)
+
+    def _validate_all_passed_private_args_have_an_effect(self) -> None:
+        """raises InvalidArgsException if validation fails"""
+        # private args are always applied first, so we can grab the keys of the private
+        # args list, and for every key, check if it matches on a substring of any
+        # argument from the argument list
+        relevant_private_args_keys = [
+            key
+            for key in self.private_args.keys()
+            if key not in _SUBSTITUTED_AT_EXECUTION_TIME
+        ]
+        unused_private_args_keys = list(
+            filter(
+                lambda private_arg_key: all(
+                    private_arg_key not in arg for arg in self.arglist
+                ),
+                relevant_private_args_keys,
+            )
+        )
+        if unused_private_args_keys:
+            unused_private_args_representation = [
+                f"{key}={self.private_args[key]}" for key in unused_private_args_keys
+            ]
+            raise ExtJobInvalidArgsException(
+                "following arguments to job have no effect (values after "
+                f"applying defines): {','.join(unused_private_args_representation)}"
+            )
+
+    def _validate_all_magic_string_in_arglist_get_resolved(
+        self, context: SubstitutionList
+    ) -> None:
         """raises InvalidArgsException if validation fails"""
         args_substituted_with_private_list = [
             (arg, self.private_args.substitute(arg)) for arg in self.arglist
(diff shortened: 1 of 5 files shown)


valentin-krasontovitsch:
everest fail is occuring elsewhere, too - Peter is informed and preparing a fix (location of a file accessed in a test has changed)

valentin-krasontovitsch:
methinks tests are mostly failing due to azure infrastructure problems. marking as ready for review.

oysteoh:
Jenkins: Test this please!

src/ert/_c_wrappers/job_queue/ext_job.py
```diff
@@ -33,6 +38,8 @@ class ExtJob:
     define_args: SubstitutionList = field(default_factory=SubstitutionList)
     help_text: str = ""
 
+    _SUBTITUTED_AT_EXECUTION_TIME: List[str] = field(default=("<ITER>", "<IENS>"))
```
eivindjahren, line 41: Since this is a dataclass `_SUBSTITUTED_AT_EXECUTION_TIME` becomes an input parameter to `__init__`. I suggest we make this a constant outside the class.

eivindjahren (changes requested):
I had a small comment, but other than that I think this is fine because we are just logging exceptions for now. We really need to see what effect this will have before we potentially do any more action.

src/ert/_c_wrappers/job_queue/ext_job.py
valentin-krasontovitsch: ah right of course outside the class, in the file... thank you : )

valentin-krasontovitsch: fixed.

src/ert/_c_wrappers/job_queue/ext_job.py
```diff
@@ -1,14 +1,21 @@
...
 from collections import defaultdict
 from dataclasses import dataclass, field
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Tuple
 
 from ert._c_wrappers.config import ConfigParser, ConfigValidationError, ContentTypeEnum
 from ert._c_wrappers.util import SubstitutionList
 from ert._clib.job_kw import type_from_kw
 
+_SUBTITUTED_AT_EXECUTION_TIME: List[str] = field(default=("<ITER>", "<IENS>"))
```
DanSava, line 13: Small fixup: SUBTITUTED -> SUBSTITUTED

valentin-krasontovitsch: thank you! fixed.

eivindjahren (approved):

PR #4717, opened 2023-01-24
Fix parsing of forward model with quotes

eivindjahren:
When quoting a proper substring of the right side of an argument assignment in a forward model, as in

```
FORWARD_MODEL COPY_FILE(<FROM>=foo,<TO>=storage/"bar--bar">)
```

the as dict method would introduce a space on the first quotation mark, which is undesired behaviour. This commit fixes that and introduces a regression test.

## Pre review checklist


valentin-krasontovitsch (approved):
thanks again for making the backport!

PR #4730, opened 2023-01-26
Move pylint command to script/pylint

pinkwah:
This makes it easier for developers to run pylint


pinkwah:
Yes, but I noticed that eg. `script/ecl-check` fails on `isort`. I reckon we do this in a separate PR where we enable `pylint`, `isort`, and `black` all at once for `script` directory.

ManInFez (approved):
:clap:

PR #4744, opened 2023-01-30
Refactor workflowjob status

### LINKED ISSUE #4701: Separate WorkflowJob data from the status of the running job
After converting `WorkflowJob` to python, it is now clear that `WorkflowJob` does many different things:

1. It is used to load `ErtPlugins`.
2. It is used to load `ErtScript`.
3. It is used to denote an executable that can be called in a workflow.
4. It holds the information of a running exectuable/script/plugin.

Of those, the 4th point seems to have a big impact and be reasonably easy to to fix: push out the properties `isCancelled`, `hasFailed`, `isRunning` `stdoutdata` and `stderrdata` to its own class, say `JobRun` with methods `cancel(self)` and a function `run_job(self, job: WorkflowJob, ert: EnkfMain, arguments: List[Any], verbose: bool=False) -> JobRun`.

DanSava:
**Issue**

## Pre review checklist

- [ ] Updated documentation

### DIFF
--- a/src/ert/_c_wrappers/job_queue/ert_script.py
+++ b/src/ert/_c_wrappers/job_queue/ert_script.py
@@ -5,7 +5,7 @@ import sys
 import traceback
 from typing import TYPE_CHECKING, Any, Callable, List, Type
 
-from ert._c_wrappers.job_queue.run_status import RunStatus
+from ert._c_wrappers.job_queue.run_status import RunTracker
 
 if TYPE_CHECKING:
     from ert._c_wrappers.enkf import EnKFMain
@@ -20,7 +20,7 @@ class ErtScript:
                 "ErtScript implementations must provide a method run(self, ert, ...)"
             )
         self.__ert = ert
-        self.__run_status = RunStatus()
+        self.__run_status = RunTracker()
 
     def ert(self) -> "EnKFMain":
         logger.info(f"Accessing EnKFMain from workflow: {self.__class__.__name__}")
--- a/src/ert/_c_wrappers/job_queue/ert_script.py
+++ b/src/ert/_c_wrappers/job_queue/ert_script.py
@@ -5,6 +5,8 @@ import sys
 import traceback
 from typing import TYPE_CHECKING, Any, Callable, List, Type
 
+from ert._c_wrappers.job_queue.run_status import RunStatus
+
 if TYPE_CHECKING:
     from ert._c_wrappers.enkf import EnKFMain
 
@@ -18,36 +20,18 @@ class ErtScript:
                 "ErtScript implementations must provide a method run(self, ert, ...)"
             )
         self.__ert = ert
-
-        self.__is_cancelled = False
-        self.__failed = False
-        self._stdoutdata = ""
-        self._stderrdata = ""
-
-    @property
-    def stdoutdata(self) -> str:
-        if isinstance(self._stdoutdata, bytes):
-            self._stdoutdata = self._stdoutdata.decode()
-        return self._stdoutdata
-
-    @property
-    def stderrdata(self) -> str:
-        if isinstance(self._stderrdata, bytes):
-            self._stderrdata = self._stderrdata.decode()
-        return self._stderrdata
+        self.__run_status = RunStatus()
 
     def ert(self) -> "EnKFMain":
         logger.info(f"Accessing EnKFMain from workflow: {self.__class__.__name__}")
         return self.__ert
 
-    def isCancelled(self) -> bool:
-        return self.__is_cancelled
-
-    def hasFailed(self) -> bool:
-        return self.__failed
+    @property
+    def run_status(self):
+        return self.__run_status
 
     def cancel(self):
-        self.__is_cancelled = True
+        self.run_status.cancel()
 
     def cleanup(self):
         """
@@ -73,20 +57,19 @@ class ErtScript:
                 arguments.append(None)
 
         try:
-            return self.run(*arguments)
+            self.run_status.start()
+            msg = self.run(*arguments)
+            self.run_status.finish(msg)
         except AttributeError as e:
             error_msg = str(e)
             if not hasattr(self, "run"):
                 error_msg = "No 'run' function implemented"
             self.output_stack_trace(error=error_msg)
-            return None
         except KeyboardInterrupt:
             error_msg = "Script cancelled (CTRL+C)"
             self.output_stack_trace(error=error_msg)
-            return None
         except Exception as e:
             self.output_stack_trace(str(e))
-            return None
         finally:
             self.cleanup()
 
@@ -95,11 +78,12 @@ class ErtScript:
 
     def output_stack_trace(self, error: str = ""):
         stack_trace = error or "".join(traceback.format_exception(*sys.exc_info()))
-        sys.stderr.write(
+        msg = (
             f"The script '{self.__class__.__name__}' caused an "
             f"error while running:\n{str(stack_trace).strip()}\n"
         )
-        self.__failed = True
+        sys.stderr.write(msg)
+        self.run_status.error(msg)
 
     @staticmethod
     def loadScriptFromFile(path) -> Callable[["EnKFMain"], "ErtScript"]:
--- a/src/ert/_c_wrappers/job_queue/external_ert_script.py
+++ b/src/ert/_c_wrappers/job_queue/external_ert_script.py
@@ -12,7 +12,7 @@ class ExternalErtScript(ErtScript):
         self.__executable = executable
         self.__job = None
 
-    def run(self, *args) -> None:
+    def run(self, *args) -> str:
         command = [self.__executable]
         command.extend([str(arg) for arg in args])
 
@@ -23,13 +23,15 @@ class ExternalErtScript(ErtScript):
         # The job will complete before stdout and stderr is returned
         stdoutdata, stderrdata = self.__job.communicate()
 
-        self._stdoutdata = codecs.decode(stdoutdata, "utf8", "replace")
-        self._stderrdata = codecs.decode(stderrdata, "utf8", "replace")
+        stdoutdata = codecs.decode(stdoutdata, "utf8", "replace")
+        stderrdata = codecs.decode(stderrdata, "utf8", "replace")
 
-        sys.stdout.write(self._stdoutdata)
+        sys.stdout.write(stderrdata)
 
         if self.__job.returncode != 0:
-            raise Exception(self._stderrdata)
+            raise RuntimeError(stderrdata)
+
+        return stdoutdata
 
     def cancel(self):
         super().cancel()
--- a/src/ert/_c_wrappers/job_queue/run_status.py
+++ b/src/ert/_c_wrappers/job_queue/run_status.py
@@ -1,51 +1,51 @@
 from dataclasses import dataclass
-from enum import Enum
+from enum import Enum, auto
 
 
-class StatusEnum(Enum):
-    RUNNING = "RUNNING"
-    FINISHED = "FINISHED"
-    CANCELED = "CANCELED"
-    FAILED = "FAILED"
-    NOT_STARTED = "NOT_STARTED"
+class RunStatus(Enum):
+    RUNNING = auto()
+    FINISHED = auto()
+    CANCELED = auto()
+    FAILED = auto()
+    NOT_STARTED = auto()
 
 
 @dataclass
-class RunStatus:
-    status: StatusEnum = StatusEnum.NOT_STARTED
+class RunTracker:
+    status: RunStatus = RunStatus.NOT_STARTED
     stdoutdata: str = ""
     stderrdata: str = ""
 
     def start(self):
-        self.status = StatusEnum.RUNNING
+        self.status = RunStatus.RUNNING
 
     def cancel(self):
-        self.status = StatusEnum.CANCELED
+        self.status = RunStatus.CANCELED
 
     def finish(self, msg=""):
         if isinstance(msg, bytes):
             msg = msg.decode()
         self.stdoutdata = msg
-        if self.status == StatusEnum.RUNNING:
-            self.status = StatusEnum.FINISHED
+        if self.status == RunStatus.RUNNING:
+            self.status = RunStatus.FINISHED
 
     def error(self, msg=""):
         if isinstance(msg, bytes):
             msg = msg.decode()
-        self.status = StatusEnum.FAILED
+        self.status = RunStatus.FAILED
         self.stderrdata = msg
 
     def has_finished(self):
-        return self.status == StatusEnum.FINISHED
+        return self.status == RunStatus.FINISHED
 
     def was_canceled(self):
-        return self.status == StatusEnum.CANCELED
+        return self.status == RunStatus.CANCELED
 
     def has_failed(self):
-        return self.status == StatusEnum.FAILED
+        return self.status == RunStatus.FAILED
 
     def is_running(self):
-        return self.status == StatusEnum.RUNNING
+        return self.status == RunStatus.RUNNING
 
     def not_started(self):
-        return self.status == StatusEnum.NOT_STARTED
+        return self.status == RunStatus.NOT_STARTED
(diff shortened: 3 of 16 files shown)


src/ert/_c_wrappers/job_queue/ert_script.py
```diff
@@ -81,34 +57,33 @@ def initializeAndRun(
...
-                    f"Script '{self.__class__.__name__}' "
-                    "has not implemented a 'run' function"
-                )
-            self.outputStackTrace(e)
-            return None
+                error_msg = "No 'run' function implemented"
+            self.output_stack_trace(error=error_msg)
         except KeyboardInterrupt:
-            return f"Script '{self.__class__.__name__}' cancelled (CTRL+C)"
+            error_msg = "Script cancelled (CTRL+C)"
```
eivindjahren, line 69: Seems like mypy is not agreeing with what you give to the `error` argument here as you seem to give it str in one setting and ArgumentError in the other. Should probably give it the union type in the `outputStackTrace` function or explicitly convert the error to a string.

src/ert/_c_wrappers/job_queue/external_ert_script.py
```diff
@@ -23,13 +23,15 @@ def run(self, *args) -> None:
...
-        self._stderrdata = codecs.decode(stderrdata, "utf8", "replace")
+        stdoutdata = codecs.decode(stdoutdata, "utf8", "replace")
+        stderrdata = codecs.decode(stderrdata, "utf8", "replace")
 
-        sys.stdout.write(self._stdoutdata)
+        sys.stdout.write(stderrdata)
 
         if self.__job.returncode != 0:
-            raise Exception(self._stderrdata)
+            raise Exception(stderrdata)
```
eivindjahren, line 32: Raising a general exception here seems not a good idea, probably want a custom error contected to a handler somewhere or a `RuntimeError`.

tests/unit_tests/c_wrappers/res/job_queue/test_workflow_runner.py
```diff
@@ -95,7 +97,7 @@ def test_workflow_failed_job():
 
     workflow_runner = WorkflowRunner(workflow, ert=None)
 
-    assert not workflow_runner.isRunning()
+    assert not workflow_runner.workflow_status == RunStatus.RUNNING
```
eivindjahren, line 100: Linting is complaining about using `not a == b` instead of `a != b`.

src/ert/_c_wrappers/job_queue/ert_script.py
```diff
@@ -37,9 +36,6 @@ def stderrdata(self) -> str:
             self._stderrdata = self._stderrdata.decode()
         return self._stderrdata
 
-    def isVerbose(self):
-        return self.__verbose
```
eivindjahren, line 41: Could you split the first commit into "Clean up workflow verbose flag" and "refactor function_ert_script.py". I am having some problem seeing what goes where.

src/ert/_c_wrappers/job_queue/ert_script.py
```diff
@@ -60,8 +60,6 @@ def initializeAndRun(
         argument_types: List[Type[Any]],
         argument_values: List[str],
     ):
-        self.__failed = False
```
eivindjahren, line 69: The message "Clenup error messages in ert script" has a typo and could be more descriptive of what is happening.

src/ert/_c_wrappers/job_queue/ert_script.py
```diff
@@ -18,36 +20,18 @@ def __init__(self, ert: "EnKFMain"):
                 "ErtScript implementations must provide a method run(self, ert, ...)"
             )
         self.__ert = ert
-
-        self.__is_cancelled = False
-        self.__failed = False
```
eivindjahren, line 24: The commit message "Workflow state/status refactor" should be in imperative form "Refactor workflow state and status". It is also a bit big so if there is any way of splitting it that makes sense I would greatly appreciate it.

src/ert/_c_wrappers/job_queue/run_status.py
```diff
@@ -0,0 +1,36 @@
+from dataclasses import dataclass
+from typing import Optional
+
+
+@dataclass
+class RunStatus:
+    RUNNING = "RUNNING"
+    FINISHED = "FINISHED"
```
eivindjahren, line 8: This class is a bit difficult to understand as it combines the concept of a Enum (RUNNING, FINISHED, CANCELED, FAILED) but is actually a dataclass. I think it would make sense for to actually hold the script that is cancelled as the "cancel" method doesn't actually cancel the job it just sets the status.

eivindjahren (changes requested):

src/ert/_c_wrappers/job_queue/run_status.py
eivindjahren: Maybe have a 
```
class RunStatus(Enum):
   RUNNING = auto()
   FINISHED = auto()
   CANCELED = auto()
   FAILED = auto()
```

and a separate:
```
@dataclass
class RunningJob:
    status: RunStatus = RunStatus.RUNNING
    stdoutdata: str = ""
    stderrdata: str = ""
    script: ErtScript

    def start(self, args):
        self.status = RunStatus.RUNNING
        self.script.run(args)

    def cancel(self):
        self.status = RunStatus.CANCELED
        self.script.cancel()
    ...
```

src/ert/_c_wrappers/job_queue/run_status.py
```diff
@@ -0,0 +1,36 @@
...
+        if self.status == RunStatus.RUNNING:
+            self.status = RunStatus.FINISHED
+
+    def error(self, msg=""):
+        if isinstance(msg, bytes):
+            msg = msg.decode()
+        self.status = RunStatus.FAILED
+        self.stderrdata = msg
+
+    def __eq__(self, other):
```
eivindjahren, line 35: This __eq__ has a bug, and also, dataclasses have a separate mechanism for tweaking __eq__ through the `field` function.

src/ert/_c_wrappers/job_queue/run_status.py
DanSava: I understand that RunStatus might be a bit hard to understand initially, maybe having it as a @dataclass is also adding to the confusion, it is designed to be a minimal internal status tracker for other objects, like ErtScript and Workflow. The aim is not to add another wrapper,  I think there are enough wrapper classes already with WorflowJob and Plugin and workflowRunner and maybe some other class that I don't remember now.

src/ert/_c_wrappers/job_queue/run_status.py
DanSava: It's not a bug it is a workaround to allow for code like `if export_job.run_status == RunStatus.FAILED:` in place of `if export_job.run_status.status == RunStatus.FAILED:`

eivindjahren: That is very confusing though :D

eivindjahren: Generally, a reader will assume, that `a == b` will have the same value as `b==a` and this doesn't do that.

DanSava: These are some excellent points, and I don't like overwriting internal class functionality, but sometimes I forget the mantra that if you can do something it does not mean you should do it :)).  I will add the extra lines to do the proper status checks.

src/ert/_c_wrappers/job_queue/ert_script.py
DanSava: I can rename the commit splitting it up into multiple logical parts is a bit difficult and it will not make it less big, but the refactor is not critical so I can just drop the commit.

src/ert/_c_wrappers/job_queue/run_status.py
DanSava: It might be true that Workflow is doing too many things it could just be a list of jobs, but that is a bigger redesign than I planned on tackling with these changes so we can just do this in the future when it becomes necessary.

eivindjahren: Ok :)

src/ert/_c_wrappers/job_queue/workflow_job.py
```diff
@@ -137,30 +136,20 @@ def run(self, ert: "EnKFMain", arguments: List[Any], verbose: bool = False) -> A
         if self.internal and self.script is not None:
             script_obj = ErtScript.loadScriptFromFile(self.script)
             self.__script = script_obj(ert)
-            result = self.__script.initializeAndRun(
-                self.argumentTypes(), arguments, verbose=verbose
-            )
-
-        elif self.internal and self.script is None:
```
eivindjahren, line 144: The if dropped `self.script is None` here. What is the implication of that.

src/ert/_c_wrappers/job_queue/function_ert_script.py
```diff
@@ -7,46 +7,33 @@
...
-        parsed_argument_types = []
-
+    def run(self, *args):
+        ert = self.ert()
         if ert is not None:
-            self.__function = ResPrototype(f"void* {function_name}(void*, stringlist)")
+            str_args = StringList()
+            for arg in args:
+                str_args.append(arg)
+            pointer = ert.from_param(ert) if hasattr(ert, "from_param") else ert
```
eivindjahren, line 20: I hope this functionality is not in use, so could we add a log-statement to see if that is the case?

src/ert/_c_wrappers/job_queue/ert_script.py
```diff
@@ -77,26 +75,25 @@ def initializeAndRun(
         try:
             return self.run(*arguments)
         except AttributeError as e:
+            error_msg = e
             if not hasattr(self, "run"):
-                self.__failed = True
-                return (
-                    f"Script '{self.__class__.__name__}' "
```
eivindjahren, line 89: Cleanup is a bit unspecsific in the commit title. Could you add a message to the body describing a bit more what happens?

src/ert/_c_wrappers/job_queue/workflow_job.py
```diff
@@ -147,9 +147,7 @@ def run(self, ert: "EnKFMain", arguments: List[Any]) -> Any:
             self.__script = ExternalErtScript(ert, self.executable)
         else:
             raise UserWarning("Unknown script type!")
-        result = self.__script.initializeAndRun(
-            self.argumentTypes(), arguments
-        )
+        result = self.__script.initializeAndRun(self.argumentTypes(), arguments)
```
eivindjahren: Seems like some difference in black makes this line change from one commit to the next. Could you run black on both commits so that the diff is smaller?

src/ert/_c_wrappers/job_queue/run_status.py
```diff
@@ -1,36 +1,48 @@
 from dataclasses import dataclass
-from typing import Optional
+from enum import Enum
 
 
-@dataclass
-class RunStatus:
+class StatusEnum(Enum):
```
eivindjahren, line 5: I think maybe name the Enum `RunStatus` and the dataclass `RunResult`.

src/ert/_c_wrappers/job_queue/workflow_job.py
DanSava: Very good question, I think the implication of dropping the `self.script is None` is that nothing changes.

Let's try to reason through this beautifully designed if, elfi chain of logic:

For the sake of being less verbose, I will assume `self.internal` and `self.script` are replaced by internal and script:

We have 
```
If internal and script is not None:
   case1()
elif internal and script is None:
   case2()
elif not internal:
   case3()
else:
   case4()
```


|   internal |   script	|   case	|  
|---	|---	|---	|
|   True	|   True	|   1	|
|   True	|   False	|   2	|
|   False	|   True	|   3	|
|   False	|   False	|   3	||


So what do we learn from the table above? 

1. That if `internal` is `False` we are always running `case3`  and it does not really matter if `script` is `True` or `False`.
2. We never reach `case4` hence we never raise that very useful `UserWarning("Unknown script type")` [Something I just realized by making the table above and maybe I should remove that final `else` if it is unreachable code]
3. It is redundant to check if the `script` is `False` to verify if we need to run `case2` because  if `internal` is `True` and we got to check if we need to run `case2` that means `script` definitely is `False` so we don't need to check it

src/ert/_c_wrappers/job_queue/function_ert_script.py
DanSava: I also hope this functionality is not in use, can you point me to an example of logging similar to what you would like to have logged here?

src/ert/_c_wrappers/job_queue/run_status.py
DanSava: I am not sure about the suggested names, can you give me a few reasons why renaming these would make things better?

src/ert/_c_wrappers/job_queue/function_ert_script.py
DanSava: Dropped changes to function_ert_script they were not needed

src/ert/_c_wrappers/job_queue/ert_script.py
DanSava: Moved changes to a separate PR https://github.com/equinor/ert/pull/4831 and added more info to the commit message

src/ert/_c_wrappers/job_queue/run_status.py
eivindjahren: The name currently has two non-descriptive parts: `Status` which could be anything (what is it the status of?) and `Enum` which just reiterates the type.

src/ert/_c_wrappers/job_queue/run_status.py
```diff
@@ -0,0 +1,51 @@
+from dataclasses import dataclass
+from enum import Enum
+
+
+class StatusEnum(Enum):
+    RUNNING = "RUNNING"
+    FINISHED = "FINISHED"
+    CANCELED = "CANCELED"
+    FAILED = "FAILED"
+    NOT_STARTED = "NOT_STARTED"
```
eivindjahren, line 10: Its more common to use 

```python
from enum import Enum, auto

class StatusEnum(Enum):
  RUNNING = auto()
  FINISHED = auto()
  CANCELED = auto()
  FAILED = auto()
  NOT_STARTED = auto()
```
https://docs.python.org/3/library/enum.html#enum.auto

src/ert/_c_wrappers/job_queue/workflow_job.py
```diff
@@ -137,56 +136,25 @@ def run(self, ert: "EnKFMain", arguments: List[Any], verbose: bool = False) -> A
...
-
-        elif self.internal and self.script is None:
+        elif self.internal:
             self.__script = FunctionErtScript(
                 ert,
                 self.function,
                 self.argumentTypes(),
                 argument_count=len(arguments),
             )
-            result = self.__script.initializeAndRun(
```
eivindjahren, line 151: I believe that that there is no valid value for `self.function` any more, so either we can go and see in the `src/ert/clib.so` for any function matching the signature `void * f(void*, stringlist_type*)` and if there is no such function we simply remove this code and make it into an error message. If we do keep this code around we should log usage to try to understand whether it is used. Also in order to call `FunctionErtScript`, `self.function` has to be not `None` so that should be in the `if` statement.

src/ert/_c_wrappers/job_queue/run_status.py
DanSava: Hmm how about renaming `RunStatus` to `RunMonitor` or `RunTracker`, because that is the main purpose of the class keeps track or monitors the state of a run also renaming `StatusEnum` to `RunStatus`.

eivindjahren: Perfect!

DanSava:
These changes simplify the status tracking for workflowrunner, workflows, workflowjobs and script, but ultimately this refactor is not needed and does not solve any current issue.  Also, I think that a more thorough redesign of the whole workflowrunner, workflows, workflowjobs and script is needed before we can attempt any further significant refactor. 

Closing the PR without merging.

PR #4802, opened 2023-02-03
Fs merge

ManInFez:
**Issue**
Resolves #my_issue

**Approach**

## Pre review checklist

- [ ] Updated documentation


src/ert/_c_wrappers/enkf/enkf_main.py
```diff
@@ -362,47 +473,74 @@ def sample_prior(
...
-                gen_kw_node = enkf_node.asGenKw()
-                keys = [val[0] for val in gen_kw_node.items()]
+            impl_type = config_node.getImplementationType()
+            if impl_type == ErtImplType.FIELD:
+                for _, realization_nr in enumerate(active_realizations):
+                    init_file = config_node.get_init_file_fmt()
+                    if "%d" in init_file:
+                        init_file = init_file % realization_nr
+                    grid = xtgeo.grid_from_file(self.ensembleConfig().grid_file)
+                    props = xtgeo.gridproperty_from_file(
```
dafeda, line 488: This assumes that the files containing grid-properties have some extension like `GRDECL` and will fail if not.
`main` works even if the files have no such extension, for example, `PORO_0`, `PORO_1` etc. which makes this regression of sorts.
Thoughts @frode-aarstad  and @oyvindeide ?

src/ert/_c_wrappers/enkf/config/gen_kw_config.py
```diff
@@ -146,3 +154,51 @@ def get_priors(self) -> List["PriorDict"]:
...
+        active_realizations: List[int],
+        nr_samples: int,
+    ) -> npt.NDArray[np.double]:
+        parameter_values = []
+        for key in keys:
+            key_hash = sha256(
+                global_seed.encode("utf-8")
+                + f"{parameter_group_name}:{key}".encode("utf-8")
+            )
+            seed = np.frombuffer(key_hash.digest(), dtype="uint32")
```
dafeda, line 198: Is it correct that the idea of `key_hash` here is to get the same seed for a specific `global_seed` and a specific parameyer/key pair?
What is the idea behind using `np.frombuffer`? AFAIK it returns an array and I'm not sure why an array is needed as a seed here.

src/ert/_c_wrappers/enkf/enkf_main.py
```diff
@@ -362,47 +473,74 @@ def sample_prior(
...
+                        init_file, name=parameter, grid=grid
+                    )
+
+                    data = props.values1d.data
```
dafeda, line 492: Are we sure we want both masked and unmasked entries from `values1d`?

src/ert/_c_wrappers/enkf/enkf_main.py
frode-aarstad: This is correct. We are adressing this in a new issue now that we store transformation, etc long with the field data

src/ert/_c_wrappers/enkf/enkf_main.py
```diff
@@ -362,47 +473,74 @@ def sample_prior(
...
+                        init_file = init_file % realization_nr
+                    storage.save_surface_file(
+                        config_node.getKey(), realization_nr, init_file
                     )
+            else:
+                raise NotImplementedError(f"{impl_type} is not supported")
         for realization_nr in active_realizations:
-            if state_map[realization_nr] in [STATE_UNDEFINED, STATE_LOAD_FAILURE]:
-                state_map[realization_nr] = RealizationStateEnum.STATE_INITIALIZED
+            storage.update_realization_state(
```
dafeda, line 535: `save_field_data`, `save_gen_kw` and `save_surface_file` all change realization state from `STATE_UNDEFINED` to `STATE_INITIALIZED`. Why do we here want to change state from `STATE_UNDEFINED` and `STATE_LOAD_FAILURE` to `STATE_INITIALIZED`?

src/ert/_c_wrappers/enkf/enkf_main.py
```diff
@@ -107,44 +201,61 @@ def _generate_parameter_files(
...
 
-        node_eclfile = node.get_enkf_outfile()
+        if type_ == ErtImplType.GEN_KW:
+            _generate_gen_kw_parameter_file(
+                fs,
+                iens,
+                node.getKeywordModelConfig(),
+                node.get_enkf_outfile(),
+                Path(run_path),
+                exports,
```
dafeda, line 230: Should we consider not passing the `exports` dict and instead return from `_generate_gen_kw_parameter_file` and the other `_generate...` functions?

src/ert/_c_wrappers/enkf/enkf_main.py
ManInFez: it is old behaviour. I guess it is to make it possible for re-running failed realizations

dafeda: How do we know that realization failed when its state gets set to `STATE_INITIALIZED` if it has state `STATE_LOAD_FAILURE`?

ManInFez: Overwriting the state STATE_LOAD_FAILURE only happens if you are writing parameters to an existing case, essentially saying it is now good to go for a new evaluation. The user should now whether the specific realization failed prior to rerunning a case, but once the rerun is initialized the information is lost if that specific realization failed in a previous evaluation.

src/ert/_c_wrappers/enkf/enkf_main.py
ManInFez: Not sure, we should make an issue for this. Could you write one up?

dafeda: Have you perhaps looked into this @frode-aarstad ?

src/ert/_c_wrappers/enkf/enkf_main.py
ManInFez: good point! made an issue https://github.com/equinor/ert/issues/4813

src/ert/_c_wrappers/enkf/config/gen_kw_config.py
ManInFez: This is the behavior currently on main: https://github.com/equinor/ert/blob/b828844409c74bcd22bd57f7196c9861a4bb8694/src/ert/_c_wrappers/enkf/data/gen_kw.py#L144 

Probably worth a discussion :shrug:

src/ert/_c_wrappers/enkf/enkf_main.py
frode-aarstad: I am not sure either. As it stands `data` is currently unmasked (an ndarray).

PR #4818, opened 2023-02-07
Fix a bug where job_name did not get substituted with DEFINEs

eivindjahren:
Resolves an issue where user reported that substitutions did not happen in job names:
user-image

## Pre review checklist

- [x] Updated documentation

### DIFF
--- a/src/ert/_c_wrappers/enkf/res_config.py
+++ b/src/ert/_c_wrappers/enkf/res_config.py
@@ -264,7 +264,10 @@ class ResConfig:
             + self.substitution_list.get("<CONFIG_FILE>", "")
         )
         jobs = []
-        for job_name, args in config_dict.get(ConfigKeys.FORWARD_MODEL, []):
+        for unsubstituted_job_name, args in config_dict.get(
+            ConfigKeys.FORWARD_MODEL, []
+        ):
+            job_name = self.substitution_list.substitute(unsubstituted_job_name)
             try:
                 job = copy.deepcopy(self.installed_jobs[job_name])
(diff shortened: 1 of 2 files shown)


src/ert/_c_wrappers/enkf/res_config.py
```diff
@@ -265,6 +265,7 @@ def _alloc_from_dict(self, config_dict):
         )
         jobs = []
         for job_name, args in config_dict.get(ConfigKeys.FORWARD_MODEL, []):
+            job_name = self.substitution_list.substitute(job_name)
```
valentin-krasontovitsch, line 268: would it be reasonable to use a different name for the variable holding the substituted job name?

src/ert/_c_wrappers/enkf/res_config.py
```diff
@@ -265,6 +265,7 @@ def _alloc_from_dict(self, config_dict):
...
             try:
                 job = copy.deepcopy(self.installed_jobs[job_name])
             except KeyError as err:
```
valentin-krasontovitsch, line 271: in line 455, github says:

> `Use 'self != other' instead of 'not self == other'`

valentin-krasontovitsch (approved):
great, besides the two little points - i'll leave adressing them at your discretion

src/ert/_c_wrappers/enkf/res_config.py
eivindjahren: That is going to do an infinite recursion and so its just wrong. This is being fixed by #4743

PR #4825, opened 2023-02-07
Fix issue with env variables not being inserted in forward model jobs

eivindjahren:
Fix an issue where environment variables did not enter the forward model.

## Pre review checklist

- [x] Updated documentation

### DIFF
--- a/src/clib/lib/config/config_parser.cpp
+++ b/src/clib/lib/config/config_parser.cpp
@@ -10,6 +10,7 @@
 #include <ert/util/hash.hpp>
 #include <ert/util/parser.hpp>
 #include <ert/util/path_stack.hpp>
+#include <optional>
 
 #include <ert/config/config_parser.hpp>
 
@@ -435,6 +436,10 @@ bool config_parser_add_key_values(
     return new_node != NULL;
 }
 
+static std::optional<std::string> char_to_optstr(std::unique_ptr<char> p) {
+    return p ? std::optional<std::string>(p.get()) : std::nullopt;
+}
+
 /**
    This function parses the config file 'filename', and updated the
    internal state of the config object as parsing proceeds. If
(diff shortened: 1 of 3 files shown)


src/clib/lib/config/config_parser.cpp
```diff
@@ -565,31 +574,54 @@ config_parse__(config_parser_type *config, config_content_type *content,
...
-                    token_list, 2, active_tokens, " ");
+                if (active_tokens < 3) {
+                    content->parse_errors.push_back(fmt::format(
+                        "Keyword:{} must have two or more arguments.",
+                        define_kw));
+                    return;
+                }
 
-                config_content_add_define(content, key, value);
+                const std::string key = util_alloc_string_copy(
```
andreas-el, line 584: Is it necessary to util_allow_string_copy this? Also, should you free the char* string if so?

src/clib/lib/config/config_parser.cpp
```diff
@@ -565,31 +574,54 @@ config_parse__(config_parser_type *config, config_content_type *content,
...
-                               "arguments. \n",
-                               __func__, define_kw);
-
-                char *key = (char *)util_alloc_string_copy(
-                    stringlist_iget(token_list, 1));
-                char *value = stringlist_alloc_joined_substring(
-                    token_list, 2, active_tokens, " ");
+                if (active_tokens < 3) {
+                    content->parse_errors.push_back(fmt::format(
+                        "Keyword:{} must have two or more arguments.",
```
andreas-el, line 579: Feedback is now correct? I take it that active_tokens 3 implies two arguments?

eivindjahren: Yes :)

src/clib/lib/config/config_parser.cpp
```diff
@@ -565,31 +576,53 @@ config_parse__(config_parser_type *config, config_content_type *content,
...
+                            char_to_optstr(res_env_isscanf_alloc_envvar(
+                                stringlist_iget(token_list.get(), iarg),
+                                env_offset));
+                        if (!env_var.has_value())
+                            break;
+
+                        const char *env_value =
+                            getenv(env_var.value().substr(1).c_str());
+                        if (env_value != nullptr) {
+                            char *new_value = util_string_replace_alloc(
```
andreas-el, line 601: This one will need to be freed also? Do you even need to allocate for this?

src/clib/lib/config/config_parser.cpp
```diff
@@ -565,31 +576,53 @@ config_parse__(config_parser_type *config, config_content_type *content,
...
+                                                      new_value);
+                        } else {
+                            env_offset += 1;
+                            logger->warning(
+                                "Environment variable: {} is not defined",
+                                env_var.value());
+                        }
+                    }
+                }
+                const std::string value = stringlist_alloc_joined_substring(
```
andreas-el, line 614: Do we need allocation for this?

src/clib/lib/config/config_parser.cpp
```diff
@@ -435,6 +436,12 @@ bool config_parser_add_key_values(
     return new_node != NULL;
 }
 
+static std::optional<std::string> char_to_optstr(char *p) {
```
andreas-el, line 439: As a user of this function, i'd expect it not to free my string when converting?

src/clib/lib/config/config_parser.cpp
eivindjahren: Ah, no, not any more as assigning `char*` to `std::string` is a copy action. Fixed it :)

src/clib/lib/config/config_parser.cpp
eivindjahren: Fixed this with use of unique_ptr

src/clib/lib/config/config_parser.cpp
eivindjahren: Fixed it with use of unique_ptr

src/clib/lib/config/config_parser.cpp
eivindjahren: Yea, I changed it to take unique_ptr as pass by value so that the expectation is that it takes ownership.

andreas-el (approved):
As far as i can see, this seems okay. :+1:

PR #4830, opened 2023-02-08
Reduce memory usage in parameter update step

### LINKED ISSUE #4805: Potentially excessive memory usage in `_es_update.py`
The function that calls the ensemble smoother first creates an in-memory storage called `temp_storage`:

https://github.com/equinor/ert/blob/b828844409c74bcd22bd57f7196c9861a4bb8694/src/ert/analysis/_es_update.py#L150-L151

Next, we loop over all update-steps and create the parameter matrix `A`:

https://github.com/equinor/ert/blob/b828844409c74bcd22bd57f7196c9861a4bb8694/src/ert/analysis/_es_update.py#L178

In the most common use case with one update step and when all parameters are active, i.e., when doing a global update, we end up with two copies of all parameters in memory. 

I think we can re-factor to avoid creating `A` and `A_with_rowscaling` and instead update the parameters in `temp_storage` but  let's discuss.

dafeda:
**Issue**
Resolves #4805 

**Approach**

## Pre review checklist

- [ ] Updated documentation


src/ert/analysis/_es_update.py
```diff
@@ -175,23 +159,43 @@ def analysis_ES(
...
+        for param_name, param_matrix in temp_storage.items():
+            current_param = [p for p in update_step.parameters if p.name == param_name][
+                0
+            ]
+
+            if active_indices := current_param.active_list.get_active_index_list():
+                temp_storage[param_name][
+                    active_indices, :
+                ] = ies.ensemble_smoother_update_step(
+                    S,
```
ManInFez, line 175:
```suggestion
        for parameter in update_step.parameters:
            if active_indices := current_param.active_list.get_active_index_list():
                temp_storage[parameter.name][
                    active_indices, :
                ] = ies.ensemble_smoother_update_step(
                    S,
```

ManInFez: same for IES

tests/unit_tests/analysis/test_es_update.py
```diff
@@ -160,8 +160,8 @@ def test_that_posterior_has_lower_variance_than_prior(copy_case):
         (
             [
                 0.5895781800838542,
-                -0.732401196722254,
-                -1.0913320833716744,
+                -1.6225405348397028,
```
ManInFez, line 163: why do we get different results?

dafeda: This is copied from the commit message:

Note that `projection` is now being passed to the smoothers.
This is because we only want to do projection when
the total number of parameters is less than ensemble_size - 1.
Because of this, the snapshot test_localization needs updating
as it ran using projection when it should not have.

So, pretty sure that the previous numbers were created with projection turned on when it should have been turned off.

src/ert/analysis/_es_update.py
dafeda: Much better, thanks 👍

ManInFez:
could you remove the "temporary print for debugging" commit?

dafeda:
> could you remove the "temporary print for debugging" commit?

Fixed, thanks 👍

PR #4874, opened 2023-02-14
Update case info tab after case is initialized

### LINKED ISSUE #4864: When using "initialize from scratch" in the "Manage cases" tool, case info panel list not updated
**Describe the bug**
When initialize from scratch is used in the manage cases tool, the case info list is not updated

**To reproduce**
1. `pip install ert`
2. `ert gui poly.ert`
3.  click on Manage cases
4. create a new case named 'new_case'
5. click on 'initialize from scratch'
6. select case "new_case"
7. click on initialize
8. Go to the "Case info" panel
9. The list says STATE_UNDEFINED when it should say STATE_INITIALIZED.

The list will update if you close the Manage cases dialog and reopen it.

DanSava:
**Issue**
Resolves #4864 

**Approach**
Update the case info list after initializing a case from scratch or from another existing case.

## Pre review checklist

- [x] Updated documentation

### DIFF
--- a/src/ert/gui/tools/manage_cases/case_init_configuration.py
+++ b/src/ert/gui/tools/manage_cases/case_init_configuration.py
@@ -70,6 +70,7 @@ class CaseInitializationConfigurationPanel(QTabWidget):
         self.addInitializeFromScratchTab()
         self.addInitializeFromExistingTab()
         self.addShowCaseInfo()
+        self.currentChanged.connect(self.on_tab_changed)
 
     def addCreateNewCaseTab(self):
         panel = QWidget()
@@ -235,7 +236,7 @@ class CaseInitializationConfigurationPanel(QTabWidget):
 
         layout.addLayout(row1)
 
-        self._case_info_area = QTextEdit()
+        self._case_info_area = QTextEdit(objectName="html_text")
         self._case_info_area.setReadOnly(True)
         self._case_info_area.setMinimumHeight(300)
 
@@ -250,8 +251,6 @@ class CaseInitializationConfigurationPanel(QTabWidget):
 
         self.addTab(case_widget, "Case info")
 
-        self._showInfoForCase()
-
     def _showInfoForCase(self, case_name=None):
         if case_name is None:
             case_name = self.ert.storage_manager.current_case.case_name
@@ -265,3 +264,8 @@ class CaseInitializationConfigurationPanel(QTabWidget):
         html += "</table>"
 
         self._case_info_area.setHtml(html)
+
+    @showWaitCursorWhileWaiting
+    def on_tab_changed(self, p_int):
+        if self.tabText(p_int) == "Case info":
+            self._showInfoForCase()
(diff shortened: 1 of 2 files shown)


src/ert/gui/tools/manage_cases/case_init_configuration.py
```diff
@@ -262,3 +261,8 @@ def _showInfoForCase(self, case_name=None):
         html += "</table>"
 
         self._case_info_area.setHtml(html)
+
+    @showWaitCursorWhileWaiting
+    def on_tab_changed(self, p_int):
+        if p_int == 3:
```
andreas-el, line 267: This is probably fine, but can we instead do something like:
``` python
if case_widget.currentWidget().objectName() is "html_text":
    self._showInfoForCase()
```
I dislike the fixed indices, because re-ordering will break stuff.
But that being said, if there is no super-easy alternative to this, we should let this pass.

src/ert/gui/tools/manage_cases/case_init_configuration.py
```diff
@@ -232,7 +233,7 @@ def addShowCaseInfo(self):
 
         layout.addLayout(row1)
 
-        self._case_info_area = QTextEdit()
+        self._case_info_area = QTextEdit(objectName="html_text")
```
andreas-el, line 236: Why did you set this name specifically?
Is this being used somewhere i can't see, or was intended to be used?

andreas-el (changes requested):
Please look at the questions, -- i'll approve if none apply.

src/ert/gui/tools/manage_cases/case_init_configuration.py
DanSava: We can do something like:

```python 
if self.tabText(p_int) == "Case info":
    self._showInfoForCase()
```

In this case, the renaming will break stuff, but maybe magic strings are better than magic numbers.

src/ert/gui/tools/manage_cases/case_init_configuration.py
DanSava: It is used in the new test to select this specific `QTextEdit` widget. 

```python
html_edit = tool.findChild(QTextEdit, name="html_text")
```

The name can be removed but if in the future somebody will add another anther QTextEdit the test will fail.

andreas-el: No, keep that as is. I must have overlooked the test when searching. :+1:

src/ert/gui/tools/manage_cases/case_init_configuration.py
andreas-el: Sure. This would be equivalent to my proposal.
I see i had 'case_widget' where i should have had 'self'
Both rely on the object-text or object-name.

andreas-el (approved):
Looks good to me!
Just need to fix the failing test.

eivindjahren (approved):
:flags:

PR #4876, opened 2023-02-14
Rename ResConfig to ErtConfig in integrationtest

lars-petter-hauge:
Between the PR with test ran and it was merged, the Config was renamed to ErtConfig.


oyvindeide (approved):

lars-petter-hauge:
fixed by https://github.com/equinor/ert/pull/4877

PR #4882, opened 2023-02-15
Fix logging messages that failed with telemetry

eivindjahren:
Fixes log messages that caused telemetry to crash.

## Pre review checklist

- [x] Updated documentation

### DIFF
--- a/src/ert/gui/main.py
+++ b/src/ert/gui/main.py
@@ -88,11 +88,11 @@ def _start_initial_gui_window(args, log_handler):
         ]
     except ConfigValidationError as error:
         messages.append(str(error))
-        logger.info("Error in config file shown in gui: '%s'", error)
+        logger.info("Error in config file shown in gui: '%s'", str(error))
         return _setup_suggester(messages), None
 
     for job in res_config.forward_model_list:
-        logger.info("Config contains forward model job %s", job)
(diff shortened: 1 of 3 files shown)


src/ert/gui/main.py
```diff
@@ -88,11 +88,11 @@ def _start_initial_gui_window(args, log_handler):
         ]
     except ConfigValidationError as error:
         messages.append(str(error))
-        logger.info("Error in config file shown in gui: '%s'", error)
+        logger.info("Error in config file shown in gui: '%s'", str(error))
```
DanSava, line 91: Is there a special reason why you did not change:
`logger.info("Error in config file shown in gui: '%s'", str(error))`
to 
`logger.info(f"Error in config file shown in gui: '{error}'")`

eivindjahren: It does mean slightly different things to the azure logger. In the second case you just get a message, in the other case the data is send separately.

DanSava (approved):
Cool fix!

src/ert/gui/main.py
DanSava: Ok, and why is it preferable to send data separately?

eivindjahren: There is a limit on the message length, and the data is treated as json.

PR #4908, opened 2023-02-17
Improve error message in ExtJobInvalidArgsException

eivindjahren:
Improve the error message for ExtJob validation

## Pre review checklist

- [x] Updated documentation

### DIFF
--- a/src/ert/_c_wrappers/job_queue/ext_job.py
+++ b/src/ert/_c_wrappers/job_queue/ext_job.py
@@ -287,8 +287,8 @@ class ExtJob:
                 f"{key}={self.private_args[key]}" for key in unused_private_args_keys
             ]
             raise ExtJobInvalidArgsException(
-                "following arguments to job have no effect (values after "
-                f"applying defines): {','.join(unused_private_args_representation)}"
+                f"following arguments to job {self.name!r} have no effect (values after"
+                f" applying defines): {','.join(unused_private_args_representation)}"
             )
 
     def _validate_all_magic_string_in_arglist_get_resolved(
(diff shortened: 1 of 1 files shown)


src/ert/_c_wrappers/job_queue/ext_job.py
```diff
@@ -287,8 +287,8 @@ def _validate_all_passed_private_args_have_an_effect(self) -> None:
                 f"{key}={self.private_args[key]}" for key in unused_private_args_keys
             ]
             raise ExtJobInvalidArgsException(
-                "following arguments to job have no effect (values after "
-                f"applying defines): {','.join(unused_private_args_representation)}"
+                f"following arguments to job {self.name!r} have no effect (values after"
```
DanSava, line 290: What does the `!r` do? A quick google search yielded: 

`!r - convert the value to a string using repr().`  I am not sure this applies to how it is used in this instance. 

Also if there is a good reason to have the `!r` should we consider to add it also to:

```python
raise ExtJobInvalidArgsException(
    f"Job {self.name} has unresolved arguments after "
    "applying argument substitutions, defines and default values: "
    f"{', '.join(unresolved_args_representation)}"
)
```

valentin-krasontovitsch (approved):
genious, thank you! which moron missed adding the job name in the original creation of this error message (me. it was I what did it)

src/ert/_c_wrappers/job_queue/ext_job.py
valentin-krasontovitsch: good point! what it does here is we see in the logs `... to job "COPY_FILE" have ...` instead of `... to job COPY_FILE have ...`

this is generally good practice because it can help make obvious what is a variable value and what is "just text" in logs - i think there was a discussion about it on some PR, let me see if i can find it

valentin-krasontovitsch: here was some discussion at least: https://github.com/equinor/ert/pull/4855

DanSava: I actually tested an example and realized the brilliance of `!r`, now I know for sure we need to use 

```python
raise ExtJobInvalidArgsException(
    f"Job {self.name!r} has unresolved arguments after "
    "applying argument substitutions, defines and default values: "
    f"{', '.join(unresolved_args_representation)}"
)
```

eivindjahren: So the idea was to use `!r` to unify how we give feedback with values that are from the config. Before we sometimes used backticks, or single quotes. This way it should be consistent.

DanSava (approved):
I also approve :smile_cat:

src/ert/_c_wrappers/job_queue/ext_job.py
eivindjahren: Added the additional `!r`

PR #4932, opened 2023-02-22
Workflows with errors are no longer loaded, before they would give error when run

### LINKED ISSUE #4931: Ert will attempt to run a partially created workflow, when before it would just fail
In 065bada071e66ea6d57dcd069566dfb0403b1b3b the workflows was changed from being parsed when its run to be parsed when the config file is read. After 682fcdbd66bf772ddcd5c573107b74e33ccf670c workflows will silently fail on missing jobnames.

We should change this so that appropriate warnings are shown for bad workflows, and that they are not installed if there are errors.

eivindjahren:
**Issue**
Resolves #4931 

**Approach**
Stop loading workflows with errors and show an appropriate ConfigWarning which shows in the GUI.

## Pre review checklist

- [x] Updated documentation

### DIFF
--- a/src/ert/_c_wrappers/enkf/ert_config.py
+++ b/src/ert/_c_wrappers/enkf/ert_config.py
@@ -488,9 +488,22 @@ class ErtConfig:
 
         for work in workflow_info:
             filename = os.path.basename(work[0]) if len(work) == 1 else work[1]
-            workflows[filename] = Workflow.from_file(
-                work[0], substitution_list, workflow_jobs
-            )
+            try:
+                existed = filename in workflows
+                workflows[filename] = Workflow.from_file(
+                    work[0], substitution_list, workflow_jobs
+                )
+                if existed:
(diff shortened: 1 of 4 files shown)


src/ert/_c_wrappers/enkf/ert_config.py
```diff
@@ -488,9 +488,22 @@ def _workflows_from_dict(cls, content_dict, substitution_list):
...
-                work[0], substitution_list, workflow_jobs
-            )
+            try:
+                existed = filename in workflows
+                workflows[filename] = Workflow.from_file(
+                    work[0], substitution_list, workflow_jobs
+                )
+                if existed:
+                    warnings.warn(
+                        f"Workflow {filename!r} was added twice",
```
mortalisk, line 498: Perhaps "Two workflows with name x configured, only the last one will be used"

mortalisk (approved):

PR #4938, opened 2023-02-22
Improve exception handling around init/loading of enkf nodes

### LINKED ISSUE #4900: unreadable field init file leads to segfault
**To reproduce**
1. `pip install ert`
2. get the ert repo and move to `test-data/snake_oil_field`
3. modify the field init files in the `fields` directory to be unreadable by your user
4. run `ert test_run snake_oil_field.ert`, or `ert gui snake_oil_field.ert` and start test run
5. segfault

**Expected behaviour**
an error message telling you that the init file(s) are unreadable

valentin-krasontovitsch:
**Issue**
Resolves #4900

**OBS**
we're using a kind of integration test approach. it takes like 2 minutes to run one of the tests, which is quite long. i've tried looking at this before, but it seems like we might not be able to do much about it because of how we set up different components of ert in threads, and their communication with each other.

**Approach**
See commit messages.

## Pre review checklist

- [X] SKIPPED! PR title captures the intent of the changes, and is fitting for release notes.
- [X] IRRELEVANT Updated documentation
- [X] Ensured new behaviour is tested

### DIFF
--- a/src/clib/lib/config/config_parser.cpp
+++ b/src/clib/lib/config/config_parser.cpp
@@ -1,3 +1,4 @@
+#include <fcntl.h>
 #include <filesystem>
 
 #include <stdio.h>
@@ -657,15 +658,26 @@ config_parse(config_parser_type *config, const char *filename,
         hash_iter_free(keys);
     }
 
-    if (util_file_readable(filename)) {
+    int test_file_readability = open(filename, O_RDONLY);
+    if (test_file_readability == -1) {
+        auto error_message =
+            fmt::format("Could not open file:{} for parsing", filename);
+        content->parse_errors.push_back(error_message);
+    }
+    int test_file_readability_close = close(test_file_readability);
+    if (test_file_readability_close == -1) {
+        auto error_message =
+            fmt::format("unexpected failure to close `{}` while "
+                        "checking if readable for parsing",
+                        filename, strerror(errno));
+        content->parse_errors.push_back(error_message);
+    }
+
+    if (test_file_readability != -1 && test_file_readability_close != -1) {
         path_stack_type *path_stack = path_stack_alloc();
         config_parse__(config, content, path_stack, filename, comment_string,
                        include_kw, define_kw, unrecognized_behaviour, validate);
         path_stack_free(path_stack);
-    } else {
-        std::string error_message =
-            util_alloc_sprintf("Could not open file:%s for parsing", filename);
-        content->parse_errors.push_back(error_message);
     }
 
     if (content->parse_errors.size() == 0)
--- a/src/clib/lib/config/config_parser.cpp
+++ b/src/clib/lib/config/config_parser.cpp
@@ -1,5 +1,5 @@
-#include <fcntl.h>
 #include <filesystem>
+#include <fstream>
 
 #include <stdio.h>
 #include <stdlib.h>
@@ -658,22 +658,29 @@ config_parse(config_parser_type *config, const char *filename,
         hash_iter_free(keys);
     }
 
-    int test_file_readability = open(filename, O_RDONLY);
-    if (test_file_readability == -1) {
-        auto error_message =
-            fmt::format("Could not open file:{} for parsing", filename);
+    std::ifstream file_handler;
+    file_handler.exceptions(std::ifstream::failbit | std::ifstream::badbit);
+    bool file_readable_check_succeeded = true;
+    try {
+        file_handler.open(filename);
+    } catch (std::ios_base::failure &err) {
+        file_readable_check_succeeded = false;
+        auto error_message = fmt::format(
+            "could not open file `{}` for parsing - {}", filename, err.what());
         content->parse_errors.push_back(error_message);
     }
-    int test_file_readability_close = close(test_file_readability);
-    if (test_file_readability_close == -1) {
+    try {
+        file_handler.close();
+    } catch (std::ios_base::failure &err) {
+        file_readable_check_succeeded = false;
         auto error_message =
             fmt::format("unexpected failure to close `{}` while "
                         "checking if readable for parsing",
-                        filename, strerror(errno));
+                        filename, err.what());
         content->parse_errors.push_back(error_message);
     }
 
-    if (test_file_readability != -1 && test_file_readability_close != -1) {
+    if (file_readable_check_succeeded) {
         path_stack_type *path_stack = path_stack_alloc();
         config_parse__(config, content, path_stack, filename, comment_string,
                        include_kw, define_kw, unrecognized_behaviour, validate);
--- a/src/clib/lib/enkf/enkf_node.cpp
+++ b/src/clib/lib/enkf/enkf_node.cpp
@@ -257,8 +257,22 @@ bool enkf_node_store(enkf_node_type *enkf_node, enkf_fs_type *fs,
 bool enkf_node_try_load(enkf_node_type *enkf_node, enkf_fs_type *fs,
                         node_id_type node_id) {
     if (enkf_node_has_data(enkf_node, fs, node_id)) {
-        enkf_node_load(enkf_node, fs, node_id);
-        return true;
+        try {
+            enkf_node_load(enkf_node, fs, node_id);
+            return true;
+        } catch (std::range_error &err) {
+            std::string error_message = fmt::format(
+                "failed to load node `{}` of type `{}` - {}",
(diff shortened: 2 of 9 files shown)


tests/unit_tests/c_wrappers/res/enkf/test_field_export.py
```diff
@@ -60,3 +64,57 @@ def test_field_export_many(snake_oil_field_example):
     assert os.path.isfile("export/with/path/PERMX_0.grdecl")
     assert os.path.isfile("export/with/path/PERMX_2.grdecl")
     assert os.path.isfile("export/with/path/PERMX_4.grdecl")
+
+
+def test_field_init_file_not_readable(copy_case):
+    copy_case("snake_oil_field")
```
eivindjahren, line 70: Likely the test could run faster if you used a smaller example. snake_oil_field is quite extensive.

valentin-krasontovitsch: indeed! i'm gonna see if i can fix : )

i also realize now that you're right, both examples don't fail at startup, but after running, so what I wrote in the OBS is not actually relevant here

valentin-krasontovitsch: OK looking at it more carefully, the first test, the one without `FORWARD_INIT:True`, fails while trying to sample priors, I believe at the start of running simulations, and that one could fail faster (experience when running in command line: the exception gets logged after like 3 seconds, but the main process dies only after one minute)

valentin-krasontovitsch: made it faster!

src/clib/lib/config/config_parser.cpp
```diff
@@ -657,15 +658,26 @@ config_parse(config_parser_type *config, const char *filename,
         hash_iter_free(keys);
     }
 
-    if (util_file_readable(filename)) {
+    int test_file_readability = open(filename, O_RDONLY);
```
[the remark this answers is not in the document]
valentin-krasontovitsch: do you think based on the fact that this code is gonna vanish in lastest a few months (both the config parser and the enkf nodes are being rewritten in python currently) we can stick to this "old-school" solution for the time being? :innocent:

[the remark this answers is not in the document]
valentin-krasontovitsch: fixed it, it's now all new c++ and shiny and pretty. please take another look @eivindjahren

src/clib/lib/enkf/enkf_node.cpp
```diff
@@ -419,11 +433,21 @@ void enkf_node_deserialize(enkf_node_type *enkf_node, enkf_fs_type *fs,
...
-        char *init_file =
-            enkf_config_node_alloc_initfile(enkf_node->config, NULL, iens);
-        bool init = enkf_node->initialize(enkf_node->data, iens, init_file);
-        free(init_file);
-        return init;
+        std::unique_ptr<char> init_file(
+            enkf_config_node_alloc_initfile(enkf_node->config, NULL, iens));
+        try {
+            bool init =
+                enkf_node->initialize(enkf_node->data, iens, init_file.get());
```
eivindjahren, line 440: This could be just `return enkf_node->initialize(enkf_node->data, iens, init_file.get());` now.

tests/unit_tests/cli/test_integration_cli.py
```diff
@@ -436,3 +436,60 @@ def test_that_prior_is_not_overwritten_in_ensemble_experiment(
...
+        [
+            TEST_RUN_MODE,
+            config_file_name,
+        ],
+    )
+
+    try:
+        run_cli(parsed)
+    except ErtCliError as err:
+        assert "failed to open" in str(err)
```
eivindjahren, line 461: Should probably add to the test that it did not fail to close the file.

src/clib/lib/config/config_parser.cpp
```diff
@@ -657,15 +658,33 @@ config_parse(config_parser_type *config, const char *filename,
...
-    if (util_file_readable(filename)) {
+    std::ifstream file_handler;
+    file_handler.exceptions(std::ifstream::failbit | std::ifstream::badbit);
+    bool file_readable_check_succeeded = true;
+    try {
+        file_handler.open(filename);
+    } catch (std::ios_base::failure &err) {
+        file_readable_check_succeeded = false;
+        auto error_message = fmt::format(
+            "could not open file `{}` for parsing - {}", filename, err.what());
```
eivindjahren, line 669: This failure condition is untested now right?

src/clib/lib/enkf/enkf_node.cpp
valentin-krasontovitsch: nooooooo we add like tons of extra useful information in the catch block - it's fine to see what file caused the error, but the node type tells you what keyword in your config file you need to find, and the node name gives you the second token in the config file line! superbly useful for finding where to fix it in the config file. i feel. would you agree?

src/clib/lib/enkf/field.cpp
```diff
@@ -825,16 +826,35 @@ bool field_fload_typed(field_type *field, const char *filename,
...
+        throw std::runtime_error(error_meessage);
+    }
 
-        return field_fload_typed(field, filename, file_type, keep_inactive);
-    } else
-        return false;
+    field_file_format_type file_type = field_config_guess_file_type(filename);
+    if (file_type == UNDEFINED_FORMAT) {
+        std::string error_meessage =
+            fmt::format("could not automagically infer type "
```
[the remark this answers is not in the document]
valentin-krasontovitsch: but but but OK : )

tests/unit_tests/cli/test_integration_cli.py
valentin-krasontovitsch: hm do you mean that we should that it did not throw an exception because of troubles with closing the file? isn't that implicitly covered by checking that "open" is in the error message?

src/clib/lib/config/config_parser.cpp
valentin-krasontovitsch: AHHHH thank you eivind. guess what. trying to add a test i've discovered, why, yes, you guessed it, yet another bug :exploding_head:

while working on these changes, i did the dumbs and got a bunch of tests failing, one of them something with observations. that's how i figured that ah, we use the parser for parsing the observation config (i had done the dumbs in the confuig parser, you see)

so i figured i make a test where the obs config is not readable. UTIL ABORT! :tada:

valentin-krasontovitsch: hm the OBS_CONFIG file seems to (at least at first, not sure about what happens after the util abort) parsed by `ecl`'s `lib/util/parser.cpp` - so maybe i just need to make an issue about this, and try to trigger / cover the config parsing error another way

src/clib/lib/enkf/field.cpp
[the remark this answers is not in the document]
valentin-krasontovitsch: fixed!

src/clib/lib/enkf/gen_kw_config.cpp
```diff
@@ -102,6 +104,14 @@ void gen_kw_config_set_parameter_file(gen_kw_config_type *config,
         config_content_type *content =
             config_parse(parser, parameter_file, "--", NULL, NULL, NULL,
                          CONFIG_UNRECOGNIZED_ADD, false);
+        if (!content->valid) {
+            logger->warning(
```
eivindjahren, line 108: This shows up as multiple logging statements in azure, which is difficult to deal with. Could we make it into one statement?

valentin-krasontovitsch:
@eivindjahren the three new tests run in 18 seconds (with `-n 3` on my machine)

oyvindeide (changes requested):
Please refrain from merging this in its current state, this causes all kinds of conflicts with our feature branch which we are aiming to merge next week. Making tests is fine, but `enkf_node.cpp` and `field.cpp` in particular will just create a lot of extra work.

eivindjahren:
> Please refrain from merging this in its current state, this causes all kinds of conflicts with our feature branch which we are aiming to merge next week. Making tests is fine, but `enkf_node.cpp` and `field.cpp` in particular will just create a lot of extra work.

@oyvindeide This fixes a bug in the existing code and adds a regression test. This therefore creates more safety for you to merge.

PR #4995, opened 2023-03-02
Fix failed workflow jobs failing with unexpected exception, now warns and skips loading

### LINKED ISSUE #4992: Ert eagerly loads wf jobs which is an unintended breaking change
A user may install workflow jobs that will not work with the current version of ert (but perhaps an older version). Currently this hard fails where before this would fail if the user tries to run the workflow job. This should be changed to not install the job and give a warning.

eivindjahren:
**Issue**
Resolves #4992
image
## Pre review checklist

- [x] Updated documentation

### DIFF
--- a/src/ert/_c_wrappers/job_queue/workflow_job.py
+++ b/src/ert/_c_wrappers/job_queue/workflow_job.py
@@ -35,6 +35,10 @@ def _workflow_job_config_parser() -> ConfigParser:
 _config_parser = _workflow_job_config_parser()
 
 
+class ErtScriptLoadFailure(ValueError):
+    pass
+
+
 @dataclass
 class WorkflowJob:
     name: str
@@ -48,6 +52,18 @@ class WorkflowJob:
 
     def __post_init__(self):
         self.__running = False
+        self._ert_script: Optional[type] = None
+        if self.script is not None and self.internal:
+            try:
+                self._ert_script = ErtScript.loadScriptFromFile(
+                    self.script,
+                )  # type: ignore
+            # Bare Exception here as we have no control
+            # of exceptions in the loaded ErtScript
+            except Exception as err:  # noqa
+                raise ErtScriptLoadFailure(
+                    f"Failed to load {self.name}: {err}"
+                ) from err
 
     @staticmethod
     def _make_arg_types_list(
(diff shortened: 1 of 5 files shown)


src/ert/_c_wrappers/job_queue/workflow_job.py
```diff
@@ -48,6 +52,14 @@ class WorkflowJob:
 
     def __post_init__(self):
         self.__running = False
+        self._ert_script: Optional[type] = None
+        if self.script is not None and self.internal:
+            try:
+                self._ert_script = ErtScript.loadScriptFromFile(self.script)  # type: ignore
+            except Exception as err:
```
berland, line 59: Can you add the pylint exception here right away (and explain why), then whenever that pylint issue on bare exceptions, it will not be touched :)

valentin-krasontovitsch, line 59: this is a very wide exception type to except - can we narrow it down?

tests/test_config_parsing/test_res_config.py
```diff
@@ -708,3 +708,38 @@ def test_that_workflows_with_errors_are_not_loaded():
...
+    does not work with the current versions of plugins installed in the system,
+    but could have worked with an older or newer version of the packages installed.
+
+    Therefore the user should be warned about workflow jobs that have issues, and not be
+    able to run those later.
+    """
+    test_config_file_name = "test.ert"
+    Path("wfs").mkdir()
+    Path("wfs/WFJOB").write_text("SCRIPT wf_script.py\nINTERNAL True", encoding="utf-8")
+    # intentionally misspelled WFJOB as WFJAB
```
valentin-krasontovitsch, line 729: i guess this comment is no longer relevant?

src/ert/_c_wrappers/job_queue/workflow_job.py
valentin-krasontovitsch: `loadScriptFromFile` it self only raises `ValueError`, maybe we can catch that one and have an "unexpected error ... " for the catch-all class afterwards?

valentin-krasontovitsch (changes requested):
just a few comments, otherwise looking good!

src/ert/_c_wrappers/job_queue/workflow_job.py
eivindjahren: Well the script can throw any exception it wants and we will have a hard time controlling it. We want any failure to load to result in this message.

tests/test_config_parsing/test_res_config.py
eivindjahren: Fixed

src/ert/_c_wrappers/job_queue/workflow_job.py
eivindjahren: Done

valentin-krasontovitsch: fair enough.

valentin-krasontovitsch (approved):
eggcelent! :ship: it

Issue #4945, opened 2023-02-23
test_config_content segfaults on Azure


berland:
**Describe the bug**
```
[havb@s034-lcatop01:~/projects/ert/tests] main$ pytest unit_tests/c_wrappers/res/config/test_config_content.py 
[havb@s034-lcatop01:~/projects/ert/tests] segfault-troll-get-the-juices$ pytest -s unit_tests/c_wrappers/res/config/test_config_content.py 
============================================================================ test session starts =============================================================================
platform linux -- Python 3.8.14, pytest-7.2.1, pluggy-1.0.0
PyQt5 5.15.7 -- Qt runtime 5.15.2 -- Qt compiled 5.15.2
benchmark: 4.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /private/havb/projects/ert, configfile: pyproject.toml
plugins: hypothesis-6.68.2, xdist-3.2.0, timeout-2.1.0, snapshot-0.9.0, raises-0.11, qt-4.2.0, mock-3.10.0, cov-4.0.0, benchmark-4.0.0, asyncio-0.20.3, anyio-3.6.1, dash-2.7.0, ert-storage-0.3.19.dev2+ge34666d, webviz-config-0.5.2a0
asyncio: mode=auto
collected 1 item                                                                                                                                                             

unit_tests/c_wrappers/res/config/test_config_content.py 
Error message: util_is_file: stat(/opt/cycle/jetpack/bin/MY_EXECUTABLE) failed: Permission denied 

See file: /tmp/ert_abort_dump.havb.20230224-125557.log for more details of the crash.
Setting the environment variable "ERT_SHOW_BACKTRACE" will show the backtrace on stderr.
Fatal Python error: Aborted

Current thread 0x00002b3c0bd12a00 (most recent call first):
  File "/prog/res/komodo/bleeding-py38-rhel7/root/lib/python3.8/site-packages/cwrap/prototype.py", line 221 in __call__
  File "/prog/res/komodo/bleeding-py38-rhel7/root/lib64/python3.8/site-packages/ert/_c_wrappers/config/config_parser.py", line 144 in add_key_value
  File "/private/havb/projects/ert/tests/unit_tests/c_wrappers/res/config/test_config_content.py", line 19 in test_get_executable_list
[...]

PR #5027, opened 2023-03-06
Improve error handling around executables in configuration

### LINKED ISSUE #4960: util abort raised if CONFIG_EXECUTABLE is missing permissions
The call to a `res_env_alloc_PATH_executable` in `config_schema_item.cpp` will call `util_abort` if it doesn't have permissions to access the executable:

https://github.com/equinor/ert/blob/c29496db6d4a79f35c32268f8ee23286be924d7f/src/clib/lib/config/config_schema_item.cpp#L417

So a external job with the config:

```
EXECUTABLE /private/exec
```

where ert does not have access to `/private/` will result in an abort signal instead of an error message.

valentin-krasontovitsch:
**Issue**
Resolves #4960
Related to #4945

**Approach**
See commits

NB: we log a warning when there are directories in the PATH that are not accessible / do not exist. this might create noise for users. please take this under consideration during review. testing locally, for me (ouch my setup), running `ert gui poly.ert` with added line `JOBSCRIPT less` produces

```
Could not search PATH dir /snap/dotnet-sdk/current for executable less - filesystem error: directory iterator cannot open directory: No such file or directory [/snap/dotnet-sdk/current]
2023-03-06 15:51:51,305 [WARNING] res.job_queue: Could not search PATH dir /home/val/adb-fastboot/platform-tools for executable less - filesystem error: directory iterator cannot open directory: No such file or directory [/home/val/adb-fastboot/platform-tools]
2023-03-06 15:51:51,305 [WARNING] res.job_queue: Could not search PATH dir /home/val/code/go/bin for executable less - filesystem error: directory iterator cannot open directory: No such file or directory [/home/val/code/go/bin]
2023-03-06 15:51:51,305 [WARNING] res.job_queue: Could not search PATH dir /snap/dotnet-sdk/current for executable less - filesystem error: directory iterator cannot open directory: No such file or directory [/snap/dotnet-sdk/current]
2023-03-06 15:51:51,305 [WARNING] res.job_queue: Could not search PATH dir /home/val/adb-fastboot/platform-tools for executable less - filesystem error: directory iterator cannot open directory: No such file or directory [/home/val/adb-fastboot/platform-tools]
2023-03-06 15:51:51,305 [WARNING] res.job_queue: Could not search PATH dir /home/val/code/go/bin for executable less - filesystem error: directory iterator cannot open directory: No such file or directory [/home/val/code/go/bin]
```

which is quite a bit of noise...

## Pre review checklist

- [x] Updated documentation

### DIFF
--- a/src/clib/lib/res_util/res_env.cpp
+++ b/src/clib/lib/res_util/res_env.cpp
@@ -1,3 +1,4 @@
+#include <filesystem>
 #include <string>
 #include <vector>
 
@@ -9,6 +10,8 @@
 #include <ert/res_util/string.hpp>
 #include <ert/util/buffer.hpp>
 #include <ert/util/util.hpp>
+
+namespace fs = std::filesystem;
 
 void res_env_unsetenv(const char *variable) { unsetenv(variable); }
 
@@ -66,6 +69,11 @@ char *res_env_alloc_PATH_executable(const char *executable) {
         int ipath = 0;
 
         for (auto path : path_list) {
+            try {
+                auto dir_iter = fs::directory_iterator(fs::path(path));
+            } catch (fs::filesystem_error &err) {
+                continue;
+            }
             char *current_attempt =
                 util_alloc_filename(path.c_str(), executable, NULL);
 
--- a/src/ert/_c_wrappers/enkf/lark_parser.py
+++ b/src/ert/_c_wrappers/enkf/lark_parser.py
@@ -279,6 +279,16 @@ class _TreeToDictTransformer:
                             f"Could not find executable {val.value!r}",
                             config_file=val.filename,
                         )
+                if not os.access(path, os.X_OK):
+                    context = (
+                        f"{val.value!r} which was resolved to {path!r}"
+                        if val.value != path
+                        else f"{val.value!r}"
+                    )
+                    raise ConfigValidationError(
+                        f"File not executable: {context}",
+                        config_file=val.filename,
+                    )
                 return path
             return str(val)
 
--- a/src/clib/lib/config/config_schema_item.cpp
+++ b/src/clib/lib/config/config_schema_item.cpp
@@ -6,6 +6,8 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include <fmt/format.h>
+
 #include <ert/res_util/res_env.hpp>
 #include <ert/util/hash.hpp>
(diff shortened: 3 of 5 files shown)


eivindjahren (approved):
Genious

src/clib/lib/res_util/res_env.cpp
```diff
@@ -66,6 +71,14 @@ char *res_env_alloc_PATH_executable(const char *executable) {
         int ipath = 0;
 
         for (auto path : path_list) {
+            try {
+                auto dir_iter = fs::directory_iterator(fs::path(path));
+            } catch (fs::filesystem_error &err) {
+                logger->warning(
+                    "Could not search PATH dir {} for executable {} - {}", path,
```
eivindjahren, line 78: Changed my mind here. We should probably silently ignore a directory in PATH we cannot open. That is what other tools do.

valentin-krasontovitsch: fixed!

src/ert/_c_wrappers/enkf/lark_parser.py
```diff
@@ -279,6 +279,16 @@ def convert(
                             f"Could not find executable {val.value!r}",
                             config_file=val.filename,
                         )
+                if not os.access(path, os.X_OK):
```
xjules, line 282: Just wondering if this belongs to the issue? Also this `_convert` only returns fill path, so just not sure it this check was necessary :)

eivindjahren: Since the new and old parser are run against the same tests, unless we specifically say it is only for the old/new parser, it's best to update both at the same time.

valentin-krasontovitsch: @xjules `_convert`?

the test i added to test the old parser basically made the new parser fail, and discovered something that should have been covered in the new parser. that's why it's included here -without this change, the test doesn't pass.

xjules: Sure, makes sense :+1:

valentin-krasontovitsch: ah now I see - you expect `convert` to do a very simple operation, while i added validation. @mortalisk could you suggest a better place for validation? or is this sensible, and the method should rather be renamed?

PR #5028, opened 2023-03-06
CaseSelector keeps track of global current_case

pinkwah:
Previously, each CaseSelector would only set its currently selected case
when `ertChanged` happens, usually only after simulations have
run (although this didn't always happen either.) Because ErtNotifier has
slightly more fine-grained signals, we can listed for
`current_case_changed` specifically and update the ComboBox' current
index.

Note how each CaseSelector listens to the global current_case, as well
as emits it, there is a potential for an infinite loop where each
CaseSelector responds to the change in the global variable by setting
it, which causes it to respond to the change etcetera. This, however,
does not happen because `setCurrentIndex` does not emit a signal if the
new index is already the current index. In theory we make a few too many
calls to this signal but in practice it is fine and is probably worth
the code cleanliness.

This commit also adds a `ertChanged` signal emission when the simulation
runner dialog is closed. Finer-grained controls are still required by
this makes it so that the CaseSelector boxes are updated more correctly.

### DIFF
--- a/src/ert/gui/ertwidgets/caselist.py
+++ b/src/ert/gui/ertwidgets/caselist.py
@@ -101,11 +101,11 @@ class CaseList(QWidget):
         )
         new_case_name = dialog.showAndTell()
         if not new_case_name == "":
-            self.storage.create_ensemble(
-                self.storage.create_experiment(),
+            ensemble = self.storage.create_experiment().create_ensemble(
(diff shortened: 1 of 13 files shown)


dafeda (approved):
Testet by click about and I think this looks good 👍

pinkwah:
Jenkins test this please

PR #5059, opened 2023-03-09
Refactor code in lark_parser, also fix define/include interactions

### LINKED ISSUE #5015: Create test/refactor handling of exceeding argc_max, possibly touch other uncovered code
There is no test for the following the argc_max error handling in the lark parser but it might not be needed due to similar handling on line 429. Consider removing.

https://github.com/equinor/ert/blob/ef9ff34565eb76610ffe9346dd3c5f0c9c5a2011/src/ert/_c_wrappers/enkf/lark_parser.py#L289

mortalisk:
**Issue**
Resolves #5015

**Approach**
Big refactor of lark_parser. Uses larks transformers more, to massage the tree into something easier to deal with.

## Pre review checklist

- [x] Updated documentation

### DIFF
--- a/src/ert/_c_wrappers/enkf/config_keywords.py
+++ b/src/ert/_c_wrappers/enkf/config_keywords.py
@@ -1,7 +1,12 @@
+import os
+import shutil
 from enum import Enum
-from typing import List, Mapping, Optional
+from typing import Any, List, Mapping, Optional, Union
 
 from pydantic import BaseModel
+
+from ert._c_wrappers.config import ConfigValidationError
+from ert._c_wrappers.enkf.lark_parser_common import FileContextToken, Instruction
 
 # These keys are used as options in KEY:VALUE statements
 BASE_SURFACE_KEY = "BASE_SURFACE"
@@ -152,12 +157,136 @@ class SchemaItem(BaseModel):
     # overwrite any previous value set
     multi_occurrence: bool = False
     expand_envvar: bool = True
+    # Index of tokens to do substitution from until end,
+    # 0 means no substitution, as keyword is never substituted
+    substitute_from: int = 1
     required_set: bool = False
     required_children_value: Mapping[str, List[str]] = {}
     # Allowed values for arguments, if empty, all values allowed
     common_selection_set: List[str] = []
     # Allowed values for specific arguments, if no entry, all values allowed
     indexed_selection_set: Mapping[int, List[str]] = {}
+
+    def check_valid(self, val: "FileContextToken", index: int):
+        if (
+            index in self.indexed_selection_set
+            and val not in self.indexed_selection_set[index]
+        ):
+            raise ConfigValidationError(
+                f"{self.kw!r} argument {index!r} must be one of"
+                f" {self.indexed_selection_set[index]!r} was {val.value!r}",
+                config_file=val.filename,
+            )
+
+    def convert(
+        self, token: FileContextToken, index: int
+    ) -> Optional[Union[str, int, float]]:
+        self.check_valid(token, index)
+        if not len(self.type_map) > index:
+            return token
+        val_type = self.type_map[index]
+        if val_type is None:
+            return token
+        if val_type == SchemaType.CONFIG_BOOL:
+            if token.lower() == "true":
+                return True
+            elif token.lower() == "false":
+                return False
+            else:
+                raise ConfigValidationError(
+                    f"{self.kw!r} must have a boolean value"
+                    f" as argument {index + 1!r}",
+                    config_file=token.filename,
+                ) from None
+        if val_type == SchemaType.CONFIG_INT:
+            try:
+                return int(token)
+            except ValueError:
+                raise ConfigValidationError(
+                    f"{self.kw!r} must have an integer value"
+                    f" as argument {index + 1!r}",
+                    config_file=token.filename,
+                ) from None
+        if val_type == SchemaType.CONFIG_FLOAT:
+            try:
+                return float(token)
+            except ValueError:
+                raise ConfigValidationError(
+                    f"{self.kw!r} must have a number as argument {index + 1!r}",
+                    config_file=token.filename,
+                ) from None
+        if val_type in [SchemaType.CONFIG_PATH, SchemaType.CONFIG_EXISTING_PATH]:
+            path = str(token)
+            if not os.path.isabs(token):
+                path = os.path.normpath(
+                    os.path.join(os.path.dirname(token.filename), token)
+                )
+            if val_type == SchemaType.CONFIG_EXISTING_PATH and not os.path.exists(path):
+                err = f'Cannot find file or directory "{token.value}" \n'
+                if path != token:
+                    err += f"The configured value was {path!r} "
+                raise ConfigValidationError(err, config_file=token.filename)
+            return path
+        if val_type == SchemaType.CONFIG_EXECUTABLE:
+            path = str(token)
+            if not os.path.isabs(token) and not os.path.exists(token):
+                path = shutil.which(token)
+                if path is None:
+                    raise ConfigValidationError(
+                        f"Could not find executable {token.value!r}",
+                        config_file=token.filename,
+                    )
+            if not os.access(path, os.X_OK):
+                context = (
+                    f"{token.value!r} which was resolved to {path!r}"
+                    if token.value != path
+                    else f"{token.value!r}"
+                )
+                raise ConfigValidationError(
+                    f"File not executable: {context}",
+                    config_file=token.filename,
+                )
+            return path
+        return str(token)
+
+    def apply_constraints(self, args: List[Any]) -> Union[List[Any], Any]:
+        args = [
+            self.convert(x, i) if isinstance(x, FileContextToken) else x
+            for i, x in enumerate(args)
+        ]
+        if self.argc_min != -1 and len(args) < self.argc_min:
+            raise ConfigValidationError(
+                f"{self.kw} must have at least {self.argc_min} arguments"
+            )
+        if self.argc_max != -1 and len(args) > self.argc_max:
+            raise ConfigValidationError(
+                f"{self.kw} must have maximum {self.argc_max} arguments"
+            )
+        if self.argc_max == 1 and self.argc_min == 1:
+            return args[0]
+        return args
+
+    def join_args(self, line: List[Any]) -> List[Any]:
+        n = self.join_after
+        if 0 < n < len(line):
+            joined = FileContextToken.join_tokens(line[n:], " ")
+            new_line = line[0:n]
+            if len(joined) > 0:
+                new_line.append(joined)
+            return new_line
+        return line
+
+
+def check_required(
+    schema: Mapping[str, SchemaItem],
+    config_dict: Mapping[str, Instruction],
+    filename: str,
+):
+    for constraints in schema.values():
+        if constraints.required_set and constraints.kw not in config_dict:
+            raise ConfigValidationError(
+                f"{constraints.kw} must be set.", config_file=filename
+            )
 
 
 def float_keyword(keyword):
@@ -170,10 +299,6 @@ def int_keyword(keyword):
 
 def string_keyword(keyword):
     return SchemaItem(kw=keyword, type_map=[SchemaType.CONFIG_STRING])
-
-
-def bool_keyword(keyword):
-    return SchemaItem(kw=keyword, type_map=[SchemaType.CONFIG_BOOL])
 
 
 def path_keyword(keyword):
@@ -214,6 +339,7 @@ def forward_model_keyword():
         argc_min=0,
         argc_max=CONFIG_DEFAULT_ARG_MAX,
         multi_occurrence=True,
+        substitute_from=0,
     )
 
 
@@ -227,7 +353,25 @@ def simulation_job_keyword():
 
 
 def data_kw_keyword():
-    return SchemaItem(kw=DATA_KW_KEY, required_set=False, argc_min=2, argc_max=2)
+    return SchemaItem(
+        kw=DATA_KW_KEY,
+        required_set=False,
+        argc_min=2,
+        argc_max=2,
+        multi_occurrence=True,
+        substitute_from=2,
+    )
+
+
+def define_keyword():
+    return SchemaItem(
+        kw=DEFINE_KEY,
+        required_set=False,
+        argc_min=2,
+        argc_max=2,
+        multi_occurrence=True,
+        substitute_from=2,
+    )
 
 
 def history_source_keyword():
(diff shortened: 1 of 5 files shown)


src/ert/_c_wrappers/enkf/lark_parser.py
```diff
@@ -45,6 +46,25 @@ def __str__(self):
         return self.value
 
 
+def assert_is_token(
```
eivindjahren, line 49: I think this should be renamed to `_assert_is_token` to make it clear that this is private to the file.

src/ert/_c_wrappers/enkf/lark_parser.py
```diff
@@ -45,6 +46,25 @@ def __str__(self):
...
+    if not isinstance(val, FileContextToken):
+        # Should be impossible to reach this through the parser
+        raise ConfigValidationError(
+            f"Unexpected value in keyword: {kw!r}, value: {val!r}",
+            config_file=kw.filename if kw else None,
+        )
+    return True
+
+
+def assert_is_tree(val: Any, filename) -> TypeGuard[FileContextToken]:
```
eivindjahren, line 61: Same rename: `_assert_is_tree`.

src/ert/_c_wrappers/enkf/lark_parser.py
```diff
@@ -164,6 +193,13 @@ class Instruction:
     args: List[FileContextToken]
 
 
+def assert_is_instruction(val: Any) -> TypeGuard[Instruction]:
```
eivindjahren, line 196: Please rename to `_assert_is_instruction`

src/ert/_c_wrappers/enkf/lark_parser.py
```diff
@@ -45,6 +46,25 @@ def __str__(self):
...
+        raise ConfigValidationError(
+            f"Unexpected value in keyword: {kw!r}, value: {val!r}",
+            config_file=kw.filename if kw else None,
+        )
+    return True
+
+
+def assert_is_tree(val: Any, filename) -> TypeGuard[FileContextToken]:
+    if not isinstance(val, Tree):
+        # Should be impossible to reach this through the parser
```
eivindjahren, line 63: Maybe make this a docstring to the assert instead:
```python
def assert_is_token(val, filename):
    """Raises an error if val is not of type FileContextToken.
    used to indicate ensure that a value fetched from the ParseTree is actually a token,
   and not Tree. This should be guaranteed by the grammar, but such a case is indicated by 
   this function for two reasons:
   1. as a type guard
   2. Just in case the grammar is more permissive than expected. Then the user still gets some reasonable
       error message.
    """

File: src/ert/_c_wrappers/enkf/lark_parser.py [hunk 5]
```diff
@@ -45,6 +46,25 @@ def __str__(self):
         return self.value
 
 
+def assert_is_token(
+    val: Any, kw: Optional[FileContextToken] = None
+) -> TypeGuard[FileContextToken]:
```
eivindjahren, line 51: Please create a test that checks that the typeguards actually create reasonable error messages.

src/ert/_c_wrappers/enkf/lark_parser.py
```diff
@@ -366,28 +396,19 @@ def get_values(item: SchemaItem, line: List[FileContextToken]):
...
-        if not isinstance(value, FileContextToken):
-            raise ConfigValidationError(
-                f"Cannot define {kw!r} to {value!r}", config_file=self.config_file
-            )
+        assert_is_token(value, kw)
         self.data_kws.append([kw, substitute(self.defines, value)])
 
     def keyword(self, tree: Tree):
         arguments = []
         kw = tree.children[0]
```
eivindjahren, line 410: After talking in person we came up with this idea: 
```python
kw = fetch_token(tree.children[0])

File: src/ert/_c_wrappers/enkf/lark_parser.py [hunk 7]
```diff
@@ -366,28 +396,19 @@ def get_values(item: SchemaItem, line: List[FileContextToken]):
...
-        if not isinstance(kw, FileContextToken):
-            raise ConfigValidationError(
-                f"Unrecognized keyword {kw!r}", config_file=self.config_file
-            )
+        assert_is_token(kw, kw)
```
eivindjahren, line 411: This is going raise and exception while raising an exception

mortalisk: I heard you like exceptions...

src/ert/_c_wrappers/enkf/lark_parser.py
```diff
@@ -45,6 +102,73 @@ def __str__(self):
...
+
+@dataclass
+class JobInstruction:
+    keyword: FileContextToken
+    job_name: FileContextToken
+    args: List[FileContextToken]
+
+    @staticmethod
+    def _get_arglist(
+        arglist_node: Tree,
```
yngve-sk, line 144: More specific type annotation of tree? Not sure what its nodes contain

src/ert/_c_wrappers/enkf/lark_parser.py
```diff
@@ -68,59 +200,60 @@ def __default_token__(self, token) -> FileContextToken:
...
 
-arg: NUMBER | STRING | UNQUOTED
-arglist: UNQUOTED kw_list
 
-kw_list: "(" [ kw_pair ("," kw_pair)*] ")"
-kw_val: NUMBER | UNQUOTED_NO_EQ | STRING | ENV_STRING
-kw_pair: KW_NAME "=" kw_val
-KW_NAME_STRICT: "<" UNQUOTED_KW ">"
-KW_NAME: UNQUOTED_KW | "<" KEYWORD_NAME ">"
+def get_filename_from_include(node: Tree, config_file: str) -> FileContextToken:
```
yngve-sk, line 241: More specific type annotation?

src/ert/_c_wrappers/enkf/lark_parser.py
```diff
@@ -68,59 +200,60 @@ def __default_token__(self, token) -> FileContextToken:
         return FileContextToken(token, self.filename)
 
 
-grammar = r"""
-WHITESPACE: (" "|"\t")+
-%ignore WHITESPACE
+class InstructionTransformer(Transformer):
+    def job_instruction(self, node):
```
yngve-sk, line 204: (also applies for `regular_instruction`) Not seeing this used, assume it is used by some lark magic in the background, maybe just a short comment on how/why this is needed, as it is not used "directly" in code but only appears in the grammar itself, but maybe this is also just me not understanding exactly how lark works.

src/ert/_c_wrappers/enkf/lark_parser.py
```diff
@@ -68,59 +200,60 @@ def __default_token__(self, token) -> FileContextToken:
...
+        and all(isinstance(n, FileContextToken) for n in node.children)
+        and node.data == data_type
+    )
 
-LETTER: UCASE_LETTER | LCASE_LETTER
-WORD: LETTER+
 
-CHAR: /[&$\[\]=,.\*a-zæøåA-ZÆØÅ10-9_%:\<\>\/-]/
-UNQUOTED: CHAR+
+def get_define_key_value(node: Tree) -> FileContextToken:
```
yngve-sk, line 226: Tree[FileContextToken]?

src/ert/_c_wrappers/enkf/lark_parser.py
```diff
@@ -153,333 +185,89 @@ def substitute(
...
-                item.argc_max != CONFIG_DEFAULT_ARG_MAX
-                and len(instruction.args) > item.argc_max
-            ):
-                raise ConfigValidationError(
-                    f"{kw!r} takes maximum {item.argc_max} arguments",
-                    config_file=kw.filename,
-                )
 
-        self.keywords.append(instruction)
+def _substitute_args(args, constraints, defines):
```
yngve-sk, line 235: I think this could be type annotated, `_substitute` always returns a `str`, so this should return the type of args (which is ..?) OR at least a list of `str`s

src/ert/_c_wrappers/enkf/lark_parser.py
```diff
@@ -153,333 +185,89 @@ def substitute(
...
-        if isinstance(node.children[0], Token):
-            continue  # This is either a newline or a comment
-        if node.children[0].data == "include":
-            inc_node = node.children[0]
-
-            if len(inc_node.children) > 1:
+        kw, *args = node
+        if kw == "DEFINE":
+            constraints = define_keyword()
+            args = constraints.join_args(args)
```
yngve-sk, line 261: Similar comment to that on _tree_to_dict and type annotations. Also, the args are transformed by first join_args, then _substitute_args (but not apply_constraints? I guess you know they are already applied when we are at this point maybe?, but just mentioning it).

src/ert/_c_wrappers/enkf/lark_parser_common.py
```diff
@@ -0,0 +1,58 @@
...
+        return inst
+
+    def __repr__(self):
+        return f"{self.value!r}"
+
+    def __str__(self):
+        return self.value
+
+    @classmethod
+    def join_tokens(cls, tokens, separator=" "):
```
yngve-sk, line 34: Seems like someone forgot to add type annotations

```python
def join_tokens(
        cls, tokens: List["FileContextToken"], separator=" "
    ) -> "FileContextToken":
```

yngve-sk (commented):
Looks a lot better than before, a few unclear things / proposed improvements are in detail comments

src/ert/_c_wrappers/enkf/config_keywords.py
```diff
@@ -152,13 +157,119 @@ class SchemaItem(BaseModel):
...
+        if index in self.indexed_selection_set:
+            if val not in self.indexed_selection_set[index]:
+                raise ConfigValidationError(
+                    f"{self.kw!r} argument {index!r} must be one of"
+                    f" {self.indexed_selection_set[index]!r} was {val.value!r}",
+                    config_file=val.filename,
+                )
+
+    def convert(
+        self, val: FileContextToken, index: int
```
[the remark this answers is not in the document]
mortalisk: I think it's correct. A FileContextToken is a string, which has a few additional fields. It is a bit weird. The .value is a string, which is not a FileContextToken.

src/ert/_c_wrappers/enkf/config_keywords.py
```diff
@@ -152,13 +157,119 @@ class SchemaItem(BaseModel):
...
+    def check_valid(self, val: "FileContextToken", index: int):
+        if index in self.indexed_selection_set:
+            if val not in self.indexed_selection_set[index]:
+                raise ConfigValidationError(
+                    f"{self.kw!r} argument {index!r} must be one of"
+                    f" {self.indexed_selection_set[index]!r} was {val.value!r}",
+                    config_file=val.filename,
+                )
+
+    def convert(
```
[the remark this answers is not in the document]
mortalisk: What is a value? It is still a string, potentially. Should it be named token_to_int_or_float_or_boolean_or_string ?

src/ert/_c_wrappers/enkf/config_keywords.py
```diff
@@ -152,13 +157,129 @@ class SchemaItem(BaseModel):
...
+        if self.argc_min != -1 and len(args) < self.argc_min:
+            raise ConfigValidationError(
+                f"{self.kw} must have at least {self.argc_min} arguments"
+            )
+        if self.argc_max != -1 and len(args) > self.argc_max:
+            raise ConfigValidationError(
+                f"{self.kw} must have maximum {self.argc_max} arguments"
+            )
+        if self.argc_max == 1 and self.argc_min == 1:
+            return args[0]
```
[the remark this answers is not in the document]
mortalisk: Well, some part of the code has to do this weird thing. And the weird thing is what the schema dictates it should be, so I think the weird thing belongs in the schema functionality. It also simplifies a lot the parsing code, which now can ignore this particular weirdness.

src/ert/_c_wrappers/enkf/lark_parser.py
```diff
@@ -153,333 +185,89 @@ def substitute(
...
+
+    for node in tree.children:
+        try:
+            kw, *args = node
+            if kw not in schema:
+                warnings.warn(f"unknown key {kw!r}", category=ConfigWarning)
+                continue
+            constraints = schema[kw]
+
+            args = constraints.join_args(args)
```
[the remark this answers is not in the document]
mortalisk: They all return List[FileContextToken], except apply_constraints, which returns what ever you are supposed to put into the dictionary.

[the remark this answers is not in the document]
mortalisk: I take it back, they are either List[FileContextToken], or the schema will tell you to not do joining or substitution, and return what ever was given.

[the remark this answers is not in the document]
mortalisk: In other words, the actual type is dependent on the schema item in question.

src/ert/_c_wrappers/enkf/lark_parser.py
mortalisk: It's a bit complicated again, the actual type is dependent on the logic of what we're doing. If you are supposed to substitute, you can expect the particular item to be a FileContextToken, otherwise it can be anything. And it will return the same as given in the case of substitution, but not the other functions.

src/ert/_c_wrappers/enkf/lark_parser.py
mortalisk: There are no actual constraints on the defines, except the joining

src/ert/_c_wrappers/enkf/lark_parser.py
mortalisk: I have added a few isinstance(x, FileContextToken) here and there, to make it clear that the function will only do stuff to FileContextToken, and otherwise return what ever was there. This is strictly not needed, if the code is correct otherwise, but does make it clearer.

Did similar another place.

yngve-sk (approved):
Looks good! Most important comments are addressed 👍

tests/test_config_parsing/test_ert_config_parsing.py
```diff
@@ -829,3 +830,70 @@ def test_that_include_take_into_account_path():
         "job1",
         "job2",
     ]
+
+
+@pytest.mark.usefixtures("use_tmpdir", "set_site_config")
+def test_that_substitution_happens_for_include():
+    """
+    Tests that use_new_parser resolves an issue
+    with the old parser where the first relative path
```
eivindjahren, line 839: This docstring doesn't describe what is in test

src/ert/_c_wrappers/enkf/lark_parser_common.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def __str__(self):
+        return self.value
+
+    @classmethod
+    def join_tokens(
+        cls, tokens: List["FileContextToken"], separator=" "
+    ) -> "FileContextToken":
+        first = tokens[0]
+        min_start_pos = min(x.start_pos for x in tokens)
```
eivindjahren, line 40: linting doesn't like that x.start_pos is optional

PR #5060, opened 2023-03-09
Describe squashing more precisely

valentin-krasontovitsch:
We update the contributing guidelines to specify details around squashing and commit history.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,19 +27,29 @@ We expect commit messages to follow this style:
 1. Wrap the body at 72 characters
 1. Use the body to explain what and why vs. how
 
-This list is taken from [here](https://chris.beams.io/posts/git-commit/). 
+This list is taken from [here](https://chris.beams.io/posts/git-commit/).
 
-Also, focus on making clear the reasons why you made the change in the first place—the way things worked before the change (and what was wrong with that), the way they work now, and why you decided to solve it the way you did. A commit body is required for anything except very small changes.
+Also, focus on making clear the reasons why you made the change in the first
+place—the way things worked before the change (and what was wrong with that),
+the way they work now, and why you decided to solve it the way you did. A
+commit body is required for anything except very small changes.
 
-(*) Tip for making sure all tests passes, try out --exec while rebasing. You can then have all tests run per commit in a single command.
+(*) Tip for making sure all tests passes, try out --exec while rebasing. You
+can then have all tests run per commit in a single command.
+
 ## Pull Request Process
 
 1. Work on your own fork of the main repo
+1. Squash/organize your work into meaningful atomic commits, if possible.
 1. Push your commits and make a draft pull request using the pull request template.
 1. Check that your pull request passes all tests.
-1. When all tests have passed and your are happy with your changes, change your pull request to "ready for review"
-   and ask for a code review.
-1. When your code has been approved—rebase, squash and merge your changes.
+1. While you wait, carefully review the diff yourself.
+1. When all tests have passed and your are happy with your changes, change your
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -27,19 +27,29 @@ We expect commit messages to follow this style:
...
-1. When all tests have passed and your are happy with your changes, change your pull request to "ready for review"
-   and ask for a code review.
-1. When your code has been approved—rebase, squash and merge your changes.
+1. While you wait, carefully review the diff yourself.
+1. When all tests have passed and your are happy with your changes, change your
+   pull request to "ready for review" and ask for a code review.
+1. When your code has been approved
+  * rebase onto the base branch
+  * squash whatever still needs squashing, and
+  * merge.
```
eivindjahren, line 52: This three point list seems a bit weird, I believe we don't generally allow "merge" at least and require linear history. "rebase onto the base branch" is a bit vague about what you want the reader to do.

valentin-krasontovitsch: yeah i'm not super happy with it either, thanks for picking up on it.
so semantics about `merge` - i think a fast-forward merge is still a merge.
`rebase onto the base branch` - what i want the user to do, is if their branch is based on an old version of say `main`, then they should update it first by rebasing onto `upstream/main`. one could reformulate as
> update your branch by rebasing onto the base branch if necessary
but one could also say that the github gui already offers "rebase and merge" which does exactly that, so maybe the point is moot

eivindjahren: I think we can link to this: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#require-linear-history as the minimum requirement we have. It is enforced by our github settings so it shouldn't be a problem. Beyond that, I, as a reviewer, usually assume that the author either uses `--fixup` to show his intent of how the commits are going to look like after merging, or that everything is squashed. If I am unsure, I will ask and indicate which action I think is appropriate.

eivindjahren: One thing that is important, is that the pr author has created a PR which is contained enough in scope so that we only get 1 main commit, with perhaps 1-2 smaller commits that fix minor related issues.

eivindjahren: If there is a completely unrelated commit, then I would prefer it had a separate PR, no matter how small.

valentin-krasontovitsch: i have

- linked to the resource
- included what you describe about marking with fixup for commits that happen during review
- added a paragraph on scoping of PRs
- fixed up the three weird points

CONTRIBUTING.md
```diff
@@ -27,19 +27,41 @@ We expect commit messages to follow this style:
...
+1. Squash/organize your work into meaningful atomic commits, if possible.
 1. Push your commits and make a draft pull request using the pull request template.
 1. Check that your pull request passes all tests.
-1. When all tests have passed and your are happy with your changes, change your pull request to "ready for review"
-   and ask for a code review.
-1. When your code has been approved—rebase, squash and merge your changes.
+1. While you wait, carefully review the diff yourself.
+1. When all tests have passed and your are happy with your changes, change your
+   pull request to "ready for review" and ask for a code review.
+1. As a courtesy to the reviewer(s), mark commits that react to review comments
```
eivindjahren, line 60: Maybe make it more clear that this point is optional: "As a courtesy to the reviewer(s), you may mark commits that react to review comments with `fixup`".

valentin-krasontovitsch: it is done!

eivindjahren (approved):

PR #5066, opened 2023-03-10
Verify if active realizations is less than minimum set

### LINKED ISSUE #3727: Warn user instead of error after run when active realisations < MIN_REALIZATIONS
**Describe the bug**

If the user specifies a set of realizations that is smaller than `MIN_REALIZATIONS` in the config file, all forward models can succeed and still ert will produce a failure. Ert has enough information before running that this will fail.

**To reproduce**
1. `pip install ert`
2. Make sure `MIN_REALIZATIONS` > 1
3. `ert gui my_config.ert`
4. Run experiment and specify single realization

**Expected behaviour**
A warning that user has specified less than `MIN_REALIZATIONS` from config and have them either specify more realizations or change `MIN_REALIZATIONS`.

**Environment**
 - OS: [e.g. RHEL7]: all
 - ERT/Komodo release: main
 - Python version: all
 - Remote/HPC execution involved: [N/A] (valid for both)

**Other**
https://github.com/equinor/ert/pull/3426
This PR changed the behaviour, as ert used to take the minimum between chosen realizations and `MIN_REALIZATIONS`. This issue proposes that the change should stay, but we can stop users from running an experiment that ultimately will fail.

andreas-el:
**Issue**
Resolves #3727 

**Approach**

## Pre review checklist

- [x] Updated documentation

### DIFF
--- a/src/ert/shared/models/base_run_model.py
+++ b/src/ert/shared/models/base_run_model.py
@@ -488,6 +488,28 @@ class BaseRunModel:
             return
         errors = []
 
+        active_mask = self._simulation_arguments.get("active_realizations", [])
+        active_realizations_count = len(
+            [i for i in range(len(active_mask)) if active_mask[i]]
+        )
+
+        min_realization_count: int = 0
+
+        if self._ert:
+            min_realization_count = (
+                self.ert().analysisConfig().minimum_required_realizations
+            )
+
+        if (
+            "SingleTestRun" not in str(type(self))
+            and active_realizations_count < min_realization_count
+        ):
+            raise ValueError(
+                f"Number of active realizations ({active_realizations_count}) is less "
+                f"than the specified MIN_REALIZATIONS in the config file "
+                f"({min_realization_count})"
+            )
+
         current_case = self._simulation_arguments.get("current_case", None)
         target_case = self._simulation_arguments.get("target_case", None)
 
--- a/tests/unit_tests/cli/test_integration_cli.py
+++ b/tests/unit_tests/cli/test_integration_cli.py
@@ -574,3 +574,57 @@ def run_ert_test_run(config_file: str) -> None:
         ],
     )
     run_cli(parsed)
+
+
+@pytest.mark.parametrize(
+    "mode",
+    [
+        pytest.param(ENSEMBLE_SMOOTHER_MODE),
+        pytest.param(ITERATIVE_ENSEMBLE_SMOOTHER_MODE),
+        pytest.param(ES_MDA_MODE),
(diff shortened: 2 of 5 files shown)


oysteoh:
Seems to work as expected, but I wonder if the message could be slightly different... :thinking: At least I suggest an alternative and then we can discuss... 
Instead of 
`Number of active realizations is less than the specified MIN_REALIZATIONS in the config file: 6 < 10`
would this look nicer? 
`Number of active realizations ( 6 ) is less than the specified MIN_REALIZATIONS in the config file ( 10 )`

andreas-el:
> Seems to work as expected, but I wonder if the message could be slightly different... 🤔 At least I suggest an alternative and then we can discuss... Instead of `Number of active realizations is less than the specified MIN_REALIZATIONS in the config file: 6 < 10` would this look nicer? `Number of active realizations ( 6 ) is less than the specified MIN_REALIZATIONS in the config file ( 10 )`

Sure thing. I picked that message since it was already present in the repo for this specific problem. The problem was that it was raised *after* execution.

oysteoh:
> > Seems to work as expected, but I wonder if the message could be slightly different... thinking At least I suggest an alternative and then we can discuss... Instead of `Number of active realizations is less than the specified MIN_REALIZATIONS in the config file: 6 < 10` would this look nicer? `Number of active realizations ( 6 ) is less than the specified MIN_REALIZATIONS in the config file ( 10 )`
> 
> Sure thing. I picked that message since it was already present in the repo for this specific problem. The problem was that it was raised _after_ execution.

But please just suggest other ways of writing it as well - or just disregard my comment if not agree. I just thought it looked a bit silly with the `: 6 < 10` statement  .. :smile:

andreas-el:
> > > Seems to work as expected, but I wonder if the message could be slightly different... thinking At least I suggest an alternative and then we can discuss... Instead of `Number of active realizations is less than the specified MIN_REALIZATIONS in the config file: 6 < 10` would this look nicer? `Number of active realizations ( 6 ) is less than the specified MIN_REALIZATIONS in the config file ( 10 )`
> > 
> > 
> > Sure thing. I picked that message since it was already present in the repo for this specific problem. The problem was that it was raised _after_ execution.
> 
> But please just suggest other ways of writing it as well - or just disregard my comment if not agree. I just thought it looked a bit silly with the `: 6 < 10` statement .. smile

I have absolutely no objections to this. Already changed the PR to this proposal. :+1:

oysteoh (approved):

andreas-el:
Jenkins test this please!

PR #5100, opened 2023-03-15
For several config errors, show more simultaneously in the "Some problems detected" window 

### LINKED ISSUE #5099: Collect multiple errors with file context when parsing config files

yngve-sk:
**Issue**
Previously targeted towards 5099

image

**Approach**
Base idea: Each ConfigValidationError can store information about several error, and most errors should be pinpointed to a text segment in the config file.

Motivation & planned future usage:
Validation errors are caught inside a try-except, and combined so that there is a "top-level" error with all the future error information. This still lets us have local control over error handling and smoothly transition into a consolidated error handling for the ert config parsing.

First step: Introduce `errors_list` field to ConfigValidationError, which can be made use of gradually.

Example usage (with errors)
```python
errors: [ConfigValidationError] = []
for items in some_list:
   if item.fail(): 
      errors.append(
         ConfigValidationError(info=ErrorInfo(line, end_line, column, end_column, message="",data={})
      )
...
# ...gathering up more errors

raise ConfigValidationError.combined(errors)
```

## Pre review checklist

- [ ] Updated documentation

### DIFF
--- a/src/ert/_c_wrappers/config/config_parser.py
+++ b/src/ert/_c_wrappers/config/config_parser.py
@@ -1,5 +1,6 @@
 import os.path
-from typing import Optional
+from dataclasses import dataclass
+from typing import List, Optional, Union
 
 from cwrap import BaseCClass
 from ecl.util.util import StringHash
@@ -9,14 +10,40 @@ from ert._c_wrappers.config.config_content import ConfigContent
 from ert._c_wrappers.config.unrecognized_enum import UnrecognizedEnum
 
 
+@dataclass()
+class Location:
+    filename: str
+    start_pos: Optional[int] = None
+    line: Optional[int] = None
+    column: Optional[int] = None
+    end_line: Optional[int] = None
+    end_column: Optional[int] = None
+    end_pos: Optional[int] = None
+
+
 class ConfigWarning(UserWarning):
     pass
 
 
 class ConfigValidationError(ValueError):
-    def __init__(self, errors: str, config_file: Optional[str] = None) -> None:
-        self.config_file = config_file
+    def __init__(
+        self,
+        errors: str,
+        config_file: Optional[str] = None,
+        location: Optional[Union[str, Location]] = None,
+    ) -> None:
+        if config_file:
+            self.location = Location(config_file)
+        elif location is not None:
+            if isinstance(location, Location):
+                self.location = location
+            else:
+                self.location = Location(location)
+        else:
+            self.location = Location(filename="")
+
         self.errors = errors
+
         super().__init__(
             (
                 f"Parsing config file `{self.config_file}` "
@@ -25,6 +52,68 @@ class ConfigValidationError(ValueError):
             if self.config_file
             else f"{self.errors}"
         )
+
+    def replace(self, old_text: str, new_text: str):
+        return ConfigValidationError(
+            errors=self.errors.replace(old_text, new_text),
+            config_file=self.config_file,
+            location=self.location,
+        )
+
+    @property
+    def config_file(self):
+        return self.location.filename
+
+    @config_file.setter
+    def config_file(self, config_file):
+        self.location.filename = config_file
+
+    def get_error_messages(self):
+        return [self.errors]
+
+
+class CombinedConfigError(ConfigValidationError):
+    def __init__(
+        self,
+        errors: Optional[
+            List[Union[ConfigValidationError, "CombinedConfigError"]]
+        ] = None,
+    ):
+        self.errors = []
+
+        for err in errors or []:
+            self.add_error(err)
+
+    def __str__(self):
+        return ", ".join(str(x) for x in self.errors)
+
+    def is_empty(self):
+        return len(self.errors) == 0
+
+    def add_error(self, error: Union[ConfigValidationError, "CombinedConfigError"]):
+        if isinstance(error, CombinedConfigError):
+            self.errors.append(*error.errors)
+        else:
+            self.errors.append(error)
+
+    def get_error_messages(self):
+        all_messages = []
+        for e in self.errors:
+            all_messages.append(*e.get_error_messages())
+
+        return all_messages
+
+    def find_matching_error(self, match: str) -> Optional[ConfigValidationError]:
+        return next(x for x in self.errors if match in str(x))
+
+    @property
+    def config_file(self):
+        return self.errors[0].location.filename
+
+    @config_file.setter
+    def config_file(self, config_file: str):
+        for err in self.errors:
+            err.config_file = config_file
 
 
 class ConfigParser(BaseCClass):
@@ -140,8 +229,11 @@ class ConfigParser(BaseCClass):
         config_content.setParser(self)
 
         if validate and not config_content.isValid():
-            raise ConfigValidationError(
-                config_file=config_file, errors=config_content.getErrors()
+            raise CombinedConfigError(
+                errors=[
+                    ConfigValidationError(errors=x, config_file=config_file)
+                    for x in config_content.getErrors()
+                ],
             )
 
         return config_content
--- a/src/ert/_c_wrappers/config/config_parser.py
+++ b/src/ert/_c_wrappers/config/config_parser.py
@@ -75,9 +75,7 @@ class ConfigValidationError(ValueError):
 class CombinedConfigError(ConfigValidationError):
     def __init__(
         self,
-        errors: Optional[
-            List[Union[ConfigValidationError, "CombinedConfigError"]]
-        ] = None,
+        errors: Optional[List[ConfigValidationError]] = None,
     ):
         self.errors = []
 
@@ -90,7 +88,7 @@ class CombinedConfigError(ConfigValidationError):
     def is_empty(self):
         return len(self.errors) == 0
 
-    def add_error(self, error: Union[ConfigValidationError, "CombinedConfigError"]):
+    def add_error(self, error: ConfigValidationError):
         if isinstance(error, CombinedConfigError):
             self.errors.append(*error.errors)
         else:
--- a/src/ert/_c_wrappers/config/config_parser.py
+++ b/src/ert/_c_wrappers/config/config_parser.py
@@ -90,14 +90,14 @@ class CombinedConfigError(ConfigValidationError):
 
     def add_error(self, error: ConfigValidationError):
         if isinstance(error, CombinedConfigError):
-            self.errors.append(*error.errors)
+            self.errors += error.errors
         else:
             self.errors.append(error)
 
     def get_error_messages(self):
         all_messages = []
         for e in self.errors:
-            all_messages.append(*e.get_error_messages())
+            all_messages += e.get_error_messages()
 
         return all_messages
 
--- a/src/ert/_c_wrappers/config/config_parser.py
+++ b/src/ert/_c_wrappers/config/config_parser.py
@@ -90,6 +90,9 @@ class CombinedConfigError(ConfigValidationError):
 
     def add_error(self, error: ConfigValidationError):
         if isinstance(error, CombinedConfigError):
+            if error == self:
+                raise ValueError("Cannot add a combined configuration error to itself")
+
             self.errors += error.errors
         else:
             self.errors.append(error)
--- a/src/ert/_c_wrappers/enkf/ert_config.py
+++ b/src/ert/_c_wrappers/enkf/ert_config.py
@@ -595,7 +595,9 @@ class ErtConfig:
                 )
 
         for hook_name, mode_name in hook_workflow_info:
-            if mode_name not in [runtime.name for runtime in HookRuntime.enums()]:
+            try:
+                run_mode = HookRuntime.from_string(mode_name)
+            except ValueError:
                 combined_error.add_error(
                     ConfigValidationError(
                         errors=f"Run mode {mode_name!r} not supported for Hook "
@@ -604,7 +606,9 @@ class ErtConfig:
                 )
                 continue
 
-            if hook_name not in workflows:
+            try:
+                workflow = workflows[hook_name]
+            except KeyError:
                 combined_error.add_error(
                     ConfigValidationError(
                         errors=f"Cannot setup hook for non-existing jo"
@@ -613,9 +617,7 @@ class ErtConfig:
                 )
                 continue
 
-            hooked_workflows[HookRuntime.from_string(mode_name)].append(
-                workflows[hook_name]
-            )
+            hooked_workflows[run_mode].append(workflow)
 
         if not combined_error.is_empty():
             raise combined_error
--- a/src/ert/_c_wrappers/enkf/ert_config.py
+++ b/src/ert/_c_wrappers/enkf/ert_config.py
@@ -394,6 +394,7 @@ class ErtConfig:
                             config_file=config_file,
                         )
                     )
+                    continue
 
             try:
                 job.validate_args(substitution_list)
--- a/src/ert/_c_wrappers/enkf/ert_config.py
+++ b/src/ert/_c_wrappers/enkf/ert_config.py
@@ -612,8 +612,8 @@ class ErtConfig:
             except KeyError:
                 combined_error.add_error(
                     ConfigValidationError(
-                        errors=f"Cannot setup hook for non-existing jo"
-                        f"b name {hook_name!r}"
+                        errors=f"Cannot setup hook for non-existing workflow named "
+                        f"{hook_name!r}"
                     )
                 )
                 continue
--- a/src/ert/_c_wrappers/enkf/ert_config.py
+++ b/src/ert/_c_wrappers/enkf/ert_config.py
@@ -11,7 +11,11 @@ from typing import Any, ClassVar, Dict, List, Mapping
 import pkg_resources
 
 from ert._c_wrappers.config import ConfigParser
-from ert._c_wrappers.config.config_parser import ConfigValidationError, ConfigWarning
+from ert._c_wrappers.config.config_parser import (
+    CombinedConfigError,
+    ConfigValidationError,
+    ConfigWarning,
+)
 from ert._c_wrappers.enkf.analysis_config import AnalysisConfig
 from ert._c_wrappers.enkf.config_keys import ConfigKeys
 from ert._c_wrappers.enkf.ensemble_config import EnsembleConfig
@@ -84,10 +88,12 @@ class ErtConfig:
         user_config_dict = ErtConfig.read_user_config(
             user_config_file, use_new_parser=use_new_parser
         )
+
         config_dir = os.path.abspath(os.path.dirname(user_config_file))
         ErtConfig._log_config_file(user_config_file)
         ErtConfig._log_config_dict(user_config_dict)
         ErtConfig.apply_config_content_defaults(user_config_dict, config_dir)
+
         return ErtConfig.from_dict(user_config_dict)
 
     @classmethod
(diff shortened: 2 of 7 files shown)


eivindjahren:
Looks good! Needs to have it pass the style check first. I also think it would be a good idea to just use it in one instance so that the code is actually in use from the get go. If not, we might quickly figure out that we need to redesign it. I think https://github.com/equinor/ert/blob/3bd1bb11cdd712ddf071929d4657b453ac33e1af/src/ert/_c_wrappers/enkf/ert_config.py#L334 would be a good candidate.

src/ert/_c_wrappers/enkf/ert_config.py
```diff
@@ -330,13 +344,25 @@ def read_forward_model(
...
-                        f"Could not find job {job_name!r} in list of installed jobs: "
-                        f"{list(installed_jobs.keys())!r}"
-                    ),
-                    config_file=config_file,
-                ) from err
+                errors.append(
+                    ConfigValidationError(
+                        info=ErrorInfo(
+                            message="",
+                            data={},
```
eivindjahren, line 351: Rather than having a dummy ErrorInfo,  I think we should leave it as None for now.

src/ert/_c_wrappers/config/config_parser.py
```diff
@@ -13,10 +14,35 @@ class ConfigWarning(UserWarning):
     pass
 
 
+@dataclass
+class ErrorInfo:
+    line: int
+    end_line: int
+    column: int
```
eivindjahren, line 21: Maybe we should delay implementing this until we need it.

src/ert/_c_wrappers/config/config_parser.py
```diff
@@ -13,10 +14,35 @@ class ConfigWarning(UserWarning):
...
+    end_column: int
+    message: Optional[str]
+    data: Optional[dict]
+    original_error: Optional[Exception]
```
eivindjahren, line 25: You could make a classmethod `from_token` that returns None if not a token and The location info from the token otherwise.

eivindjahren: I think you might want this instead:
```python
@dataclass
class LocationInfo:
   line: int
   column: int
   file: str
```

src/ert/_c_wrappers/config/config_parser.py
```diff
@@ -13,10 +14,35 @@ class ConfigWarning(UserWarning):
...
+
+
 class ConfigValidationError(ValueError):
-    def __init__(self, errors: str, config_file: Optional[str] = None) -> None:
+    def __init__(
+        self,
+        errors: Optional[str],  # Old setup
+        config_file: Optional[str] = None,  # Old setup
```
eivindjahren, line 32: Having two different concepts, old and new, in the same class is going to become too complicate.

It seems to me that the fields you need are
```
self.errors = List[Tuple[str, LocationInfo]]
```

src/ert/_c_wrappers/config/config_parser.py
```diff
@@ -26,6 +52,43 @@ def __init__(self, errors: str, config_file: Optional[str] = None) -> None:
...
 
+    @classmethod
+    def attempt(cls, *lambdas: [Callable[[], Any]]):
+        """
+        Helper function for running a set of functions and gathering up the errors.
+        To be used as an alternative to using a try-catch for every single
+        error-raising routine. It is assumed that every lambda function provided to
+        this is responsible for raising a ConfigValidationError (possibly combined)
+        when something goes wrong.
+        :param lambdas: the functions to be run
```
eivindjahren, line 63: I think we can add this method when we start using it in a realistic scenario.

src/ert/_c_wrappers/config/config_parser.py
```diff
@@ -26,6 +52,43 @@ def __init__(self, errors: str, config_file: Optional[str] = None) -> None:
...
+
+        return errors, results
+
+    @classmethod
+    def combined(cls, errors: ["ConfigValidationError"]):
+        """
+        Creates a combined ConfigValidationError with the same information as a list of
+        other ConfigValidationError instances.
+        :param errors a list of ConfigValidationErrors
+        """
```
eivindjahren, line 83: Instead of the `combined` method, I think we can use the separate `CombinedConfigError` I suggested above.

src/ert/_c_wrappers/config/config_parser.py
yngve-sk: On second thought, I think we can just not introduce a class for this at all, and use FileContextToken directly instead? 
FCToken does exactly the same as it does, so it would be introducing a differently named class containing exactly the same info

src/ert/_c_wrappers/config/config_parser.py
```diff
@@ -1,18 +1,51 @@
...
 
 from ert._c_wrappers import ResPrototype
 from ert._c_wrappers.config.config_content import ConfigContent
 from ert._c_wrappers.config.unrecognized_enum import UnrecognizedEnum
 
 
+class FileContextToken(Token):
+    filename: str
+
+    def __new__(cls, token, filename):
```
eivindjahren, line 17: `FileContextToken` should probably just be imported from `.lark_parser`.

src/ert/_c_wrappers/enkf/ert_config.py
```diff
@@ -81,14 +81,26 @@ def __post_init__(self):
...
         cls, user_config_file, use_new_parser: bool = USE_NEW_PARSER_BY_DEFAULT
     ) -> "ErtConfig":
-        user_config_dict = ErtConfig.read_user_config(
-            user_config_file, use_new_parser=use_new_parser
-        )
+        try:
+            user_config_dict = ErtConfig.read_user_config(
+                user_config_file, use_new_parser=use_new_parser
+            )
+        except ConfigValidationError as err:
```
eivindjahren, line 88: This is left over test code

src/ert/_c_wrappers/enkf/ert_config.py
```diff
@@ -81,14 +81,26 @@ def __post_init__(self):
...
         config_dir = os.path.abspath(os.path.dirname(user_config_file))
         ErtConfig._log_config_file(user_config_file)
         ErtConfig._log_config_dict(user_config_dict)
         ErtConfig.apply_config_content_defaults(user_config_dict, config_dir)
-        return ErtConfig.from_dict(user_config_dict)
+
+        try:
+            return ErtConfig.from_dict(user_config_dict)
+        except ConfigValidationError as err:
+            raise err
```
eivindjahren, line 101: This is left-over test code

src/ert/_c_wrappers/config/config_parser.py
```diff
@@ -27,6 +40,77 @@ def __init__(self, errors: str, config_file: Optional[str] = None) -> None:
         )
 
 
+class SingleConfigError(ConfigValidationError):
+    def __init__(self, error: str, filename: str, location: Optional[Location] = None):
+        self.error = error
```
eivindjahren, line 45: This starting to look good! I just realized over the weekend that we probably could make `ConfigValidationError` have the following `__init__`:

```python
class ConfigValidationError(ValueError):
    def __init__(self, errors: str, location: Optional[Union[str,Location]] = None) -> None:
```
then we don't need to introduce `SingleConfigError`.

yngve-sk: I added this in now, removing SingleConfigError. I added (temporarily) some stuff to ConfigValidationError to support its future + current use, planning to remove the redundant stuff when the "old way" of calling/instantiating it is deprecated:

in ConfigValidationError: remove config_file getter/setter
in ConfigValidationError.init: remove config_file argument
in ConfigValidationError: take in only one error at a time

src/ert/_c_wrappers/config/config_parser.py
```diff
@@ -9,13 +10,38 @@
...
 
 
 class ConfigValidationError(ValueError):
-    def __init__(self, errors: str, config_file: Optional[str] = None) -> None:
-        self.config_file = config_file
+    def __init__(
+        self,
+        errors: str,
+        config_file: Optional[str] = None,
+        location: Optional[Union[str, Location]] = None,
```
eivindjahren, line 33: I think with this design we can just use the type `Optional[Location]` instead of `Optional[Union[str, Location]]`

src/ert/_c_wrappers/enkf/enkf_main.py
```diff
@@ -6,10 +6,11 @@
 import os
 from datetime import datetime
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Dict, List, Mapping, Optional, Sequence, Union
+from typing import Any, Dict, List, Mapping, Optional, Sequence, TYPE_CHECKING, Union
 
 import numpy as np
 import xtgeo
+from ert._clib import trans_func  # noqa: no_type_check
```
eivindjahren, line 13: I think perhaps you have `isort` incorrectly set up if it is moving this import for you.

src/ert/_c_wrappers/enkf/ert_config.py
```diff
@@ -319,25 +331,31 @@ def _validate_ensemble_config(cls, ensemble_config, config_path):
...
+                all_errors.append(
+                    ConfigValidationError(
+                        errors=(
+                            f"Could not find job {job_name!r} in list of installed "
+                            f"jobs: "
+                            f"{list(installed_jobs.keys())!r}"
+                        ),
+                        config_file=config_file,
+                    )
+                )
```
eivindjahren, line 357: Perhaps it would simplify the logic a bit to use a continue here instead of relying on the condition `job is not None` ?

src/ert/_c_wrappers/enkf/ert_config.py
```diff
@@ -350,27 +368,36 @@ def read_forward_model(
...
-                    raise ConfigValidationError(
-                        errors=f"{err}: 'FORWARD_MODEL {job_name}({args})'\n",
-                        config_file=config_file,
-                    ) from err
+                    all_errors.append(
+                        ConfigValidationError(
+                            errors=f"{err}: 'FORWARD_MODEL {job_name}({args})'\n",
+                            config_file=config_file,
+                        )
+                    )
```
eivindjahren, line 376: Need to use `continue` after the except.

src/ert/_c_wrappers/config/config_parser.py
```diff
@@ -9,13 +10,38 @@
...
+    ) -> None:
+        if config_file:
+            self.location = Location(config_file)
+        elif location is not None:
+            if isinstance(location, Location):
+                self.location = location
+            else:
+                self.location = Location(location)
+        else:
+            self.location = Location(filename="")
```
eivindjahren, line 43: def get_error_messages(self):
   return [self.message]

src/ert/_c_wrappers/enkf/ert_config.py
yngve-sk: Done

src/ert/_c_wrappers/enkf/ert_config.py
yngve-sk: Done, added continue after missing fwd models / sim jobs

src/ert/_c_wrappers/config/config_parser.py
```diff
@@ -26,6 +52,22 @@ def __init__(self, errors: str, config_file: Optional[str] = None) -> None:
...
+        self.location.filename = config_file
+
+
+class CombinedConfigError(ConfigValidationError):
+    def __init__(self, errors: Optional[List[ConfigValidationError]] = None):
+        self.errors = errors if errors is not None else []
+
+    def __str__(self):
+        return ", ".join(str(x) for x in self.errors)
+
```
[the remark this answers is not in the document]
yngve-sk: Done

src/ert/_c_wrappers/enkf/ert_config.py
```diff
@@ -548,25 +596,35 @@ def _workflows_from_dict(cls, content_dict, substitution_list):
 
         for hook_name, mode_name in hook_workflow_info:
             if mode_name not in [runtime.name for runtime in HookRuntime.enums()]:
-                raise ConfigValidationError(
-                    errors=[f"Run mode {mode_name!r} not supported for Hook Workflow"]
+                combined_error.add_error(
+                    ConfigValidationError(
+                        errors=f"Run mode {mode_name!r} not supported for Hook "
+                        f"Workflow"
+                    )
```
eivindjahren, line 603: Need to `continue` if this error is hit, otherwise `HookRuntime.from_string(mode_name)` will raise. Seeing as this did not fail the tests, we should probably add a test that would have caught this error. It would have to be a unit test because parsing also guards against this error. Still, it is not correct behavior for the `_workflows_from_dict` function.

src/ert/_c_wrappers/job_queue/workflow.py
```diff
@@ -68,6 +69,8 @@ def from_file(
             content = parser.parse(
                 to_compile, unrecognized=UnrecognizedEnum.CONFIG_UNRECOGNIZED_ERROR
             )
+        except CombinedConfigError as err:
```
eivindjahren, line 72: Here we will have to replace the config_file location which has been set to a temporary directory, with `src_file`.

src/ert/gui/main.py
```diff
@@ -276,6 +276,7 @@ def _setup_suggester(
     suggest_layout = QVBoxLayout()
     buttons_layout = QHBoxLayout()
 
+    print(errors)
```
eivindjahren, line 279: This looks like it is just a debugging print.

src/ert/_c_wrappers/enkf/ert_config.py
yngve-sk: Nice catch. I did add the continue and will add the test that fails without the continue

src/ert/_c_wrappers/job_queue/workflow.py
yngve-sk: Done, with this logic it is now also assumed that the config_file will be set for all the "child" errors of the `CombinedConfigError` raised from parsing, since the operation is setting the `config_file` for an already combined error (with potentially not the same file origin? I'm guessing not but don't know it for sure)

src/ert/gui/main.py
yngve-sk: Oops, fixed

src/ert/gui/main.py
```diff
@@ -129,7 +129,7 @@ def _start_initial_gui_window(
         ]
 
     except ConfigValidationError as error:
-        error_messages.append(str(error))
+        error_messages.append(*error.get_error_messages())
```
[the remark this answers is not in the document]
yngve-sk: Using += syntax now

yngve-sk:
Jenkins test this please

src/ert/_c_wrappers/enkf/ert_config.py
```diff
@@ -351,28 +388,37 @@ def read_forward_model(
                         for key, val in args:
                             job.private_args.addItem(key, val)
                 except ValueError as err:
-                    raise ConfigValidationError(
-                        errors=f"{err}: 'FORWARD_MODEL {job_name}({args})'\n",
-                        config_file=config_file,
-                    ) from err
+                    combined_error.add_error(
```
eivindjahren, line 391: We will also need to `continue` from this error case, otherwise `job.validate_args` will later create nonsensical error messages.

yngve-sk: Fixed

src/ert/_c_wrappers/job_queue/workflow.py
yngve-sk: A bit too magical indeed, this will be resolved by another future pass where the `config_file` parameter is deprecated and the logic is somehow unified for the two in a (ideally) consistent manner

src/ert/_c_wrappers/config/config_parser.py
```diff
@@ -26,6 +53,68 @@ def __init__(self, errors: str, config_file: Optional[str] = None) -> None:
...
+    def get_error_messages(self):
+        return [self.errors]
+
+
+class CombinedConfigError(ConfigValidationError):
+    def __init__(
+        self,
+        errors: Optional[
+            List[Union[ConfigValidationError, "CombinedConfigError"]]
```
eivindjahren, line 79: There is no need for `Union` here as `CombinedConfigError` is a subtype of `ConfigValidationError`.

src/ert/_c_wrappers/config/config_parser.py
```diff
@@ -26,6 +53,68 @@ def __init__(self, errors: str, config_file: Optional[str] = None) -> None:
...
+
+    def __str__(self):
+        return ", ".join(str(x) for x in self.errors)
+
+    def is_empty(self):
+        return len(self.errors) == 0
+
+    def add_error(self, error: Union[ConfigValidationError, "CombinedConfigError"]):
+        if isinstance(error, CombinedConfigError):
+            self.errors.append(*error.errors)
```
eivindjahren, line 95: `self.errors.append(*error.errors)` crashes with `list.append() takes exactly one argument`. You need to do `self.errors += error.errors`

eivindjahren: Seems like we may have to create a test that hits this condition.

eivindjahren:
Could maybe be idea to specify which queue `MAX_RUNNING` is negative for. Adding the following to the config file:
```
QUEUE_OPTION LSF MAX_RUNNING -1
QUEUE_OPTION LOCAL MAX_RUNNING -1
QUEUE_OPTION SLURM MAX_RUNNING -1
```

results in
image

yngve-sk:
> Could maybe be idea to specify which queue `MAX_RUNNING` is negative for. Adding the following to the config file:
> 
> ```
> QUEUE_OPTION LSF MAX_RUNNING -1
> QUEUE_OPTION LOCAL MAX_RUNNING -1
> QUEUE_OPTION SLURM MAX_RUNNING -1
> ```
> 
> results in image

Done, though improving error messages and their overall level of detail will maybe be done in a separate pass/PR?

src/ert/_c_wrappers/enkf/ert_config.py
```diff
@@ -547,33 +596,52 @@ def _workflows_from_dict(cls, content_dict, substitution_list):
...
-                raise ConfigValidationError(
-                    errors=[
-                        f"Cannot setup hook for non-existing job name {hook_name!r}"
-                    ]
+            try:
+                workflow = workflows[hook_name]
+            except KeyError:
+                combined_error.add_error(
+                    ConfigValidationError(
+                        errors=f"Cannot setup hook for non-existing jo"
```
eivindjahren, line 615: This error message was actually always wrong, could we change it now and just have a separate commit for it? IT should say `non-existing workflow` instead of `non-existing job`.

yngve-sk: Done

yngve-sk:
Update: We had a discussion on how to do this and think there may be other ways. I think some more discussion on this would be good before finalizing as all options seem to have some pros and cons, and one doesn't stand out as "obviously" the best. 

The main goal of this is to just collect all errors without halting execution, then acting on them somehow at the end of the call in `from_dict` or `from_file`.

Current approach is maybe a bit noisy with all the raising/catching. Alternatives are 
(1) just passing a list or object/container as argument and adding to it. Benefits: Easy to understand, less added lines per function, and it matches more closely up with what we are trying to do (collecting messages instead of raising)
(2) using a fn and passing in errors there (seems inferior)
(3) maybe too magical and diff to understand without being the person implementing it, but ... using decorators + yielding of error info, which would enable us to turn on/off the "instant-stopping" effect of errors as they are raised

yngve-sk:
From what I can currently tell I'd argue that passing an object/list as a parameter to collect "errors" is more clear and introduces less bloat to the code than using try-catch, in spite of it introducing another argument to some functions. Some ref:

| Feature \| Approach ->                                           | Combined "upwards" error propagation | Passing object | Passing array | Decorator                                         | Python warnings |
|-----------------------------------------------------|------------------------------------|----------------|---------------|--------------------------------------------------|-----------------|
| Each function responsible for submitting own errors | Yes                                | Yes            | Yes           | Yes, would also have to use yield in error-submitting functions | Yes             |
| Possible to immediately raise error by toggling bool flag | Yes, flag would have to live within individual error | Yes            | No            | Yes                                              | No(?)           |
| Noisy?                                              | Quite some try/catch/raise, using errors might be deprecated because we are really collecting WARNINGS | No             | No            | No                                               | No              |
| Easy to read/understand                             | Yes but a bit noisy                | Yes            | Yes           | Not immediately                                  | Yes             |
| Easy to avoid overlapping scope/sessions when running multiple parser passes | Yes                                | Yes            | Yes           | Maybe                                            | No              |


Transitioning from what we have now to passing list/object should be very fast, as its impact on the control flow is exactly the same, only the syntax is a bit different. Doing a quick draft PR with a list as a minimal example of what this would look like, then we could just create a container class overriding `__setitem__` IF we need to intercept something.

yngve-sk:
new PR with using list as parameter for collecting errors: https://github.com/equinor/ert/pull/5155

PR #5108, opened 2023-03-16
Push ConfigValidationError for empty single value keywords to suggestor

### LINKED ISSUE #5106: Empty single value keywords will result in index error
For the keywords `ANALYSIS_SELECT`, `STD_CUTOFF`, `ENKF_ALPHA`, `MAX_SUBMIT`, `NUM_CPU`, `MAX_RUNTIME`, `ITER_COUNT`, `UPDATE_LOG_PATH`, `MIN_REALIZATIONS`, `ITER_CASE`, `GEN_KW_TAG_FORMAT`, `RANDOM_SEED`, `WORKFLOW_JOB_DIRECTORY`,
`INSTALL_JOB_DIRECTORY`, and `LICENSE_PATH`, giving no arguments will result in the following index error:

```
ERT crashed unexpectedly with "Expected 0 <= index < 0, was 0 <= -1 < 0"
  File "/prog/res/komodo/2023.03.01-py38-rhel7/root/lib64/python3.8/site-packages/ert/__main__.py", line 564, in main
    args.func(args)
  File "/prog/res/komodo/2023.03.01-py38-rhel7/root/lib64/python3.8/site-packages/ert/__main__.py", line 202, in run_gui_wrapper
    run_gui(args)
  File "/prog/res/komodo/2023.03.01-py38-rhel7/root/lib64/python3.8/site-packages/ert/gui/main.py", line 40, in run_gui
    window, ens_path = _start_initial_gui_window(args, log_handler)
  File "/prog/res/komodo/2023.03.01-py38-rhel7/root/lib64/python3.8/site-packages/ert/gui/main.py", line 86, in _start_initial_gui_window
    suggestions += ErtConfig.make_suggestion_list(args.config)
  File "/prog/res/komodo/2023.03.01-py38-rhel7/root/lib64/python3.8/site-packages/ert/_c_wrappers/enkf/ert_config.py", line 214, in make_suggestion_list
    return DeprecationMigrationSuggester(
  File "/prog/res/komodo/2023.03.01-py38-rhel7/root/lib64/python3.8/site-packages/ert/_c_wrappers/enkf/_deprecation_migration_suggester.py", line 66, in suggest_migrations
    defines = config_content_as_dict(content, {}).get("DEFINE", [])
  File "/prog/res/komodo/2023.03.01-py38-rhel7/root/lib64/python3.8/site-packages/ert/_c_wrappers/enkf/_config_content_as_dict.py", line 89, in config_content_as_dict
    content_dict[key] = item.getValue()
  File "/prog/res/komodo/2023.03.01-py38-rhel7/root/lib64/python3.8/site-packages/ert/_c_wrappers/config/config_content.py", line 159, in getValue
    node = self[item_index]
  File "/prog/res/komodo/2023.03.01-py38-rhel7/root/lib64/python3.8/site-packages/ert/_c_wrappers/config/config_content.py", line 149, in __getitem__
    raise IndexError(
IndexError: Expected 0 <= index < 0, was 0 <= -1 < 0
```

DanSava:
**Issue**
Resolves #5106 

**Approach**
Validate config file when generation suggestions

## Pre review checklist

- [ ] Updated documentation


src/ert/_c_wrappers/enkf/_deprecation_migration_suggester.py
```diff
@@ -61,8 +62,12 @@ def _add_deprecated_keywords_to_parser(self):
     def suggest_migrations(self, filename: str):
         suggestions = []
         content = self._parser.parse(
-            filename, pre_defined_kw_map=self._pre_defines, validate=False
+            filename, pre_defined_kw_map=self._pre_defines, validate=True
```
eivindjahren, line 65: `validate=True` means that the below raise is done by `self._parser.parse` so it is redundant.

Also, this is too soon to do the validation because we do not yet have the site config parsed. Therefore this will give false positives. Instead we need to not validate at this point. Check `content.isValid()` and then if that is false, return empty list.

eivindjahren (changes requested):
I had a minor comment about the raising now happening twice, but I am also wondering if this may not be the right approach any more. Now that we have a lot more control over the parsing, would it make sense to move the adding of deprecated keyword to the parser to `ert_config` and rather use `warnings.warn(category=ErtConfigDeprecationWatning)` to send deprecations to the GUI? 

We have had at least two problems related to the parsing happening twice now (which was the reason why we put `validate=False` there in the first place) so perhaps this separate step is causing more trouble than it is worth.

DanSava:
> I had a minor comment about the raising now happening twice, but I am also wondering if this may not be the right approach any more. Now that we have a lot more control over the parsing, would it make sense to move the adding of deprecated keyword to the parser to `ert_config` and rather use `warnings.warn(category=ErtConfigDeprecationWatning)` to send deprecations to the GUI?

I guess it is possible to implement what you suggested but that is a significant redesign. How about we have a fix for the issue first and then make a new issue for the redesign?

PR #5110, opened 2023-03-20
Avoid Pandas deprecation warning on frame.append

berland:
This warning is now removed:

loader.py:125: FutureWarning: The frame.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead.
    data = data.append(gen_data)

The merging of dataframes has also been postponed until all data has been collected, which might give a small speed increase.

**Issue**
Resolves noise in test logs

**Approach**
Refactor slightly and use pd.concat instead.

## Pre review checklist

- [ ] Updated documentation

### DIFF
--- a/src/ert/data/loader.py
+++ b/src/ert/data/loader.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from functools import partial
-from typing import TYPE_CHECKING, Callable, Optional, Protocol, Sequence
+from typing import TYPE_CHECKING, Callable, List, Optional, Protocol, Sequence
 
 import pandas as pd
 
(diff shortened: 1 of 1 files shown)


src/ert/data/loader.py
```diff
@@ -118,11 +118,10 @@ def _load_general_response(
             for key in facade.all_data_type_keys()
             if facade.is_gen_data_key(key) and data_key in key
         ]
-        data = pd.DataFrame()
-
+        gen_datas: List[pd.DataFrame] = []
         for time_step in time_steps:
-            gen_data = facade.load_gen_data(ensemble, data_key, time_step).T
-            data = data.append(gen_data)
+            gen_datas.append(facade.load_gen_data(ensemble, data_key, time_step).T)
```
hnformentin, line 123: Why dataS in plural?

berland: Because it is multiple "gen_data" objects, as given by the typing specification.

hnformentin: sure, but the ordinary definition for data referring to `The word data is a plural noun so write "data are`, I would think that we keep `gen_data` unless data refers to something different than the usual.
I don't think this is the case here, but maybe you know a specific reason?

berland: I was more thinking of "gen_data" as a specific thing, for which these rules don't apply. But I can rename it to `gen_data_list` instead.

hnformentin: Nice!

hnformentin (approved):
Thanks for the patience with review :-)

PR #5111, opened 2023-03-20
Solve four pylint issues

berland:
**Issue**
Adds to #3309

**Approach**
Do as suggested, add one exception for FileContextToken in lark parser. Looks legit.

## Pre review checklist

- [ ] Updated documentation


src/ert/_c_wrappers/enkf/lark_parser.py
```diff
@@ -23,6 +23,7 @@
 class FileContextToken(Token):
     filename: str
 
+    # pylint: disable=signature-differs
```
berland, line 26: @mortalisk , can you verify this?

hnformentin: What are you suggesting @mortalisk to verify? I could not understand. 
Would you like that this issue is reviewed by him for some reason or was only this comment?

mortalisk: So, it's complaining that we changed the __Init__ signature as compared with the super-class? I think that is fine, don't get why it is complaining about that in the first place.

berland: Fine with me, @hnformentin  you may review the rest :)

src/ert/shared/storage/extraction.py
```diff
@@ -16,13 +16,13 @@ def _prepare_x_axis(x_axis: List[Union[int, float, str, pd.Timestamp]]) -> List[
...
 
 
-def _get_obs_data(key, obs) -> dict:
-    return dict(
-        name=key,
-        x_axis=obs.columns.get_level_values(0).to_list(),
-        values=obs.loc["OBS"].to_list(),
-        errors=obs.loc["STD"].to_list(),
-    )
+def _get_obs_data(key, obs) -> Mapping[str, Any]:
```
hnformentin, line 19: Do you know if it would be possible and correct ` Mapping[Optional[str]]`

.pylintrc
```diff
@@ -531,5 +531,5 @@ min-public-methods=2
 
 # Exceptions that will emit a warning when being caught. Defaults to
 # "BaseException, Exception".
-overgeneral-exceptions=BaseException,
-                       Exception
+overgeneral-exceptions=builtins.BaseException,
+                       builtins.Exception
```
hnformentin, line 535: Could you please point where I can learn the difference between builtins and "non-builtins" exceptions?
I tried to find it but was not trivial to get a useful link...of course it does not seem really relevant if you don't have anything in particular to add on that.

hnformentin (approved):
A comment for learning purpose and another for preference, if  Mapping[Optional[str]] is compliant I would prefer that, but not strong preference.
Please reply before merging :-) 

Otherwise, LGTM 🎉

src/ert/shared/storage/extraction.py
berland: `Mapping` should have two arguments, https://docs.python.org/3/library/typing.html#typing.Mapping , so `Optional[str]` would probably give an error. `str, Any` is not the same as `Optional[str]`, `[str, Any] `just means that the `Mapping` (the dict) is indexed by strings, but we don't specify the types these string keys link to.

.pylintrc
berland: Ah, this I forgot to mention in the log. But given a new enough pylint and python, it will tell you that not specifying "builtins." as here is deprecated.

src/ert/shared/storage/extraction.py
hnformentin: Interesting. Thanks for explaining!

.pylintrc
hnformentin: Basically they will stop supporting if `builtins` is not used. Interesting.

PR #5115, opened 2023-03-20
Fix bug where inactive mask was not used for fields

### LINKED ISSUE #5086: When updating FIELD parameters, both active and inactive grid-cells are used
**Describe the bug**
When updating FIELD parameters, both active and inactive grid-cells are used.
Behaviour in stable and on main prior to merging of `enkf_fs_refactor` branch is that only inactive cells are used.

**To reproduce**
1. `pip install ert`
2. Run an update of a config with FIELD parameters, for example REEK on stable.
3. Run the same update on main.
4. Results will be different. (There are probably other reason for the difference as well, but this is one)

**Expected behaviour**
Behaviour should be the same before and after merge of `enkf_fs_refactor` branch.

frode-aarstad:
**Issue**
Resolves #5086 

**Approach**

## Pre review checklist

- [ ] Updated documentation


src/ert/_c_wrappers/enkf/enkf_main.py
```diff
@@ -487,18 +491,31 @@ def sample_prior(
...
+                            name=parameter,
+                            grid=grid,
                         )
+                        data = props.get_npvalues1d(order="C", fill_value=-np.nan)
                     except PermissionError as err:
                         context_message = (
                             f"Failed to open init file for parameter {parameter!r}"
                         )
                         raise RuntimeError(context_message) from err
+                    except ValueError:
```
dafeda, line 508: In which cases do we expect to get a `ValueError`?
Is it when `grid_file` is for example made using `libecl` or something like that?

src/ert/_c_wrappers/enkf/enkf_main.py
```diff
@@ -487,18 +491,31 @@ def sample_prior(
...
+                        ecl_grid = EclGrid(grid_file)
+                        with cwrap.open(init_file, "rb") as f:
+                            param = EclKW.read_grdecl(f, parameter)
+
+                        mask = [False if e else True for e in ecl_grid.export_actnum()]
```
dafeda, line 513: Perhaps this is a bit easier to read:

`[not e for e in ecl_grid.export_actnum()]`

src/ert/storage/local_ensemble.py
```diff
@@ -670,10 +727,30 @@ def save_field(
...
         output_path = self.mount_point / f"realization-{realization}"
         Path.mkdir(output_path, exist_ok=True)
-        np.save(f"{output_path}/{parameter_name}", data)
+
+        if unmasked:
+            grid = self.experiment.grid
+            assert grid
+            if isinstance(grid, xtgeo.Grid):
+                masked_data = [np.nan] * grid.ntotal
+                for i, e in enumerate(grid.actnum_indices):
```
dafeda, line 740: Have you perhaps tried writing this without the for-loop and instead relying on numpy-indexing?
Something like:

```python
masked_data[grid.actnum_indices] = data[grid.actnum_indices]
```

Fields can have tens of millions of elements so it might be worth a try.

frode-aarstad: Agree

src/ert/_c_wrappers/enkf/enkf_main.py
```diff
@@ -436,20 +440,32 @@ def sample_prior(
...
-                    try:
-                        props = xtgeo.gridproperty_from_file(
-                            init_file, name=parameter, grid=grid
-                        )
-                    except PermissionError as err:
-                        context_message = (
-                            f"Failed to open init file for parameter {parameter!r}"
+
+                    grid = ensemble.experiment.grid
+                    assert grid
```
dafeda, line 445: Do we want to assert with an error message or do you think this is fine as is?

dafeda:
Commits can perhaps be squashed.

src/ert/_c_wrappers/enkf/enkf_main.py
frode-aarstad: Ill remove the asserts before merging

PR #5134, opened 2023-03-23
Interpret zero length output from qstat as failure

berland:
The return code from qstat is believed to be flaky. Rather base
the qstat success on whether the command returns non-empty output.

This relies on:
 * ERT never asks the queue system for status unless it
   has actually submitted something
 * ERT only asks for status for jobs it has submitted, and
   when a job is finished, it is still reported by the queue
   system (accomplished by the -x default option).

This PR also adds a test verifying that the qstat_proxy
would forward a nonzero exit code.

## Pre review checklist

- [ ] Updated documentation


berland:
Closing this, it is more likely that #5169 is the solution.

tests/unit_tests/c_wrappers/res/job_queue/test_job_queue_manager_torque.py
```diff
@@ -135,6 +135,9 @@ def create_qstat_f_output(
 """
 )
 
+# A qstat command that lies about its return code on first invocation, then succeeds:
```
valentin-krasontovitsch, line 138: perhaps make it clear that this variant of qstat not only lies about its return code upon first invocation, but also has no output / produces no file (at least if i understood this correctly) - 

```suggestion
# A qstat command that has no output but rc 0 on first invocation, and then succeeds:
```

valentin-krasontovitsch (approved):
only one comment about a comment, i'll leave it up to you : )

PR #5155, opened 2023-03-28
Error collection & refactoring in config parser & lark parser

yngve-sk:
**Issue**
Resolves https://github.com/equinor/ert/issues/5099

**Approach**
Uses list passed as parameters to catch errors from parsing, while still raising it at the end as a ConfigValidationError

## Pre review checklist

- [ ] Updated documentation


src/ert/_c_wrappers/enkf/ert_config.py
```diff
@@ -252,13 +298,17 @@ def read_user_config(
...
+        cls, config_path, config_dict, collected_errors: List[ErrorInfo]
+    ):
+        for queue_name, option_name, *values in config_dict.get("QUEUE_OPTION", []):
             if option_name == "MAX_RUNNING" and int(*values) < 0:
-                raise ConfigValidationError(
-                    errors=[
-                        f"QUEUE_OPTION MAX_RUNNING is negative: {str(*values)!r}",
-                    ],
+                collected_errors.append(
+                    ErrorInfo(
```
mortalisk, line 307: perhaps we can have an extra parameter to ErrorInfo, to which we feed potential FileContextTokens, and if they are tokens, ErrorInfo can copy all the relevant info from those tokens. If they are not, which will be the case in tests and in everests case, which do not use the parser, it can simply give general errors without filename, line, etc.

src/ert/_c_wrappers/enkf/ert_config.py
```diff
@@ -70,50 +74,94 @@ class ErtConfig:
...
+        if collected_errors is None:
+            collected_errors = []
+            do_raise_errors = True
+
         substitution_list = SubstitutionList.from_dict(config_dict=config_dict)
         config_dir = substitution_list.get("<CONFIG_PATH>", "")
         config_file = substitution_list.get("<CONFIG_FILE>", "no_config")
         config_file_path = os.path.join(config_dir, config_file)
 
-        ErtConfig._validate_dict(config_dict, config_file)
```
mortalisk, line 100: We probably now desire to run as muc of the code as possible. Previously the code has been structured to stop early. I therefore propose that we ditch this _validate_dict, and let each part of the code that needs to detect error conditions which implies it cannot run, to detect and give an error in the list about that, and then simply skip calling that piece of the code, instead of skipping everything like before.

src/ert/_c_wrappers/config/config_parser.py
```diff
@@ -9,6 +10,60 @@
...
+        ):
+            token = originates_from_keyword.keyword_token
+        elif originates_from is not None and hasattr(originates_from, "token"):
+            token = originates_from.token
+        elif originates_from_these is not None:
+            tokens = [x.token for x in originates_from_these if hasattr(x, "token")]
+            # Merge the token positions etc
+
+            if len(tokens) > 0:
+                from ert._c_wrappers.enkf.lark_parser_common import FileContextToken
```
DanSava, line 53: This should be a top level import

src/ert/_c_wrappers/enkf/ert_config.py
```diff
@@ -69,58 +75,113 @@ class ErtConfig:
...
-            if self.user_config_file
-            else os.getcwd()
-        )
-
     @classmethod
     def from_file(
-        cls, user_config_file, use_new_parser: bool = USE_NEW_PARSER_BY_DEFAULT
+        cls,
+        user_config_file,
+        collected_errors: List[ErrorInfo] = None,
```
eivindjahren, line 82: I don't think it is a good idea to add this additional parameter. It seems to only be used in a test.

src/ert/_c_wrappers/enkf/ert_config.py
```diff
@@ -69,58 +75,113 @@ class ErtConfig:
...
+        self.config_path = (
+            os.path.dirname(os.path.abspath(self.user_config_file))
+            if self.user_config_file
+            else os.getcwd()
+        )
 
     @classmethod
-    def from_dict(cls, config_dict) -> "ErtConfig":
+    def from_dict(
+        cls, config_dict, collected_errors: Optional[List[ErrorInfo]] = None
```
eivindjahren, line 123: Same with the addtional collected_errors parameter here. It doesn't seem to be used for anything.

src/ert/_c_wrappers/enkf/ert_config.py
```diff
@@ -69,58 +75,113 @@ class ErtConfig:
...
+    ) -> "ErtConfig":
+        do_raise_errors = False
+        if collected_errors is None:
+            collected_errors = []
+            do_raise_errors = True
```
eivindjahren, line 128: This logic is fairly hard to follow. It would seem to me that we should always raise errors when there are some. in this function. Also, relying on the function not raising under certain conditions is going to be ahrd to manage. It would make more sense to always raise at the end, with the option of other parts of the code raising early if there are problems from which we cannot  continue.

src/ert/_c_wrappers/enkf/ert_config.py
mortalisk: When implementing the linter, this will be used I guess?

src/ert/_c_wrappers/enkf/queue_config.py
```diff
@@ -19,7 +19,7 @@ class QueueConfig:
     )
 
     @classmethod
-    def from_dict(cls, config_dict) -> QueueConfig:
+    def from_dict(cls, config_dict, collected_errors: List[ErrorInfo]) -> QueueConfig:
```
eivindjahren, line 22: The interface for this function becomes very difficult to follow, except in the context of `ErtConfig.from_file`. I suggest that it should raise ConfigValidationError that `ErtConfig.from_file` catches and then appends to its list of collected errors.

src/ert/parsing/config_errors.py
```diff
@@ -17,3 +19,14 @@ def __init__(self, errors: str, config_file: Optional[str] = None) -> None:
             if self.config_file
             else f"{self.errors}"
         )
+
+    @classmethod
+    def raise_from_collected(cls, collected_errors: List[ErrorInfo]):
+        if len(collected_errors) > 0:
```
eivindjahren, line 25: For the GUI we need the list of errors as separate boxes, so I suggest going back to the approach of keeping the list of errors as a field.

src/ert/_c_wrappers/enkf/queue_config.py
```diff
@@ -40,7 +40,19 @@ def from_dict(cls, config_dict) -> QueueConfig:
         max_submit = config_dict.get("MAX_SUBMIT", 2)
         queue_options = defaultdict(list)
         for driver, option_name, *values in config_dict.get("QUEUE_OPTION", []):
-            queue_driver_type = QueueDriverEnum.from_string(driver + "_DRIVER")
+            try:
+                queue_driver_type = QueueDriverEnum.from_string(driver + "_DRIVER")
+            except ValueError as err:
+                collected_errors.append(
+                    ErrorInfo(
```
eivindjahren, line 47: This is checked in the validation step I believe.

PR #5159, opened 2023-03-28
Add MAX_RUNTIME option to Torque driver

### LINKED ISSUE #5121: Support MAX_RUNTIME in Torque driver
The Torque driver does not support the option MAX_RUNTIME that LSF and Slurm supports

The same feature as LSF

None

kwinkunks:
**Issue**
Resolves #5121 

**Approach**
There is currently a MAX_RUNTIME option for LSF and Slurm, but not for the Torque driver we're using on Azure. This commit adds the option, following the pattern of the recently added TIMEOUT option.

It also adds to test_job_torque.cpp, which tests that the setting and getting machinery works. There is no test for the setting actually being used in the qsub command, or that it actually sets an effective time limit, however.

Also adds documentation of the option.

## Pre review checklist

- [x] Updated documentation

### DIFF
--- a/docs/reference/configuration/keywords.rst
+++ b/docs/reference/configuration/keywords.rst
@@ -655,10 +655,13 @@ Keywords controlling the simulations
 
         ::
 
-                -- Let each realizations run for 50 seconds
+                -- Let each realization run for 50 seconds
                 MAX_RUNTIME 50
 
         The MAX_RUNTIME key is optional.
+
+        Note that if you are using a Slurm queue, then you need to pass the option
+        explcitly as a queue option. See :ref:`Slurm's options <slurm_smax_runtime>`.
 
 
 Parameterization keywords
@@ -1857,24 +1860,6 @@ bjobs.
         ::
 
                 QUEUE_OPTION TORQUE TIMEOUT 126
-
-.. _torque_max_runtime:
-.. topic:: MAX_RUNTIME
-
-        By default, there is no limit on how long jobs are allowed to run for.
-        In other words, the queue will wait for ever for a job to finish. The
-        ``MAX_RUNTIME`` option sets a limit in seconds. If the job has not
-        finished within the limit, it will be killed.
-
-        To set a time limit of 90 seconds, use this option:
-
-        ::
-
-                QUEUE_OPTION TORQUE MAX_RUNTIME 90
-
-        Specifying a time of 0 s will be interpreted as unlimited runtime.
-        Internally, the option is passed to the TORQUE resource list as
-        ``walltime``. 
 
 
 .. _configuring_the_slurm_queue:
--- a/docs/reference/configuration/keywords.rst
+++ b/docs/reference/configuration/keywords.rst
@@ -1855,7 +1855,27 @@ bjobs.
         flakyness.
 
         ::
+
                 QUEUE_OPTION TORQUE TIMEOUT 126
+
+.. _torque_max_runtime:
+.. topic:: MAX_RUNTIME
+
+        By default, there is no limit on how long jobs are allowed to run for.
+        In other words, the queue will wait for ever for a job to finish. The
+        ``MAX_RUNTIME`` option sets a limit in seconds. If the job has not
+        finished within the limit, it will be killed.
+
+        To set a time limit of 90 seconds, use this option:
+
+        ::
+
+                QUEUE_OPTION TORQUE MAX_RUNTIME 90
+
+        Specifying a time of 0 s will be interpreted as unlimited runtime.
+        Internally, the option is passed to the TORQUE resource list as
+        ``walltime``. 
+
 
 .. _configuring_the_slurm_queue:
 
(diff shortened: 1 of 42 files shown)


berland:
I speculate the LSF driver itself will kill long-running jobs, which makes it clear to ERT why this job failed. Are we able to give a correct error message to the user if we let the queue system implement MAX_RUNTIME on behalf of ERT?

kwinkunks:
@berland Good question.

The function `job_status_type torque_driver_parse_status()` in `torque_driver.cpp` is responsible for parsing the output of `qstat -f`. However, **past** (i.e. completed) jobs do not show up in that output for very long. So our killed jobs may or may not be there.

Even if they are there, jobs that were killed are reported by `qstat -fx` with `job_status` code `F` (finished), so that's no good. They look exactly like properly finished jobs.

However, also in the `qstat -fx` output, killed jobs seem to have `Exit_status` of `-29`, as far as I can tell. [This resource](https://dl.technion.ac.il/docs/cis/public/usg/tamnun/PBS%20exit%20codes.html) suggests it should be `137` so :shrug: 

The most unequivocal evidence that a job was killed is captured in the error file, which the function `torque_driver_get_qstat_status()` seems to call `ert-qstat-err`. If `ert-qstat-std` does not exist, then the error file is reported with `fprintf` with "reading qstat status failed"" --- but really this means that the **job** potentially failed I think, not that qstat failed.

If a job is killed by insufficient `walltime` then its stderr file looks like:

    =>> PBS: job killed: walltime 10 exceeded limit 5

If we wanted to get more information out of that file, I think we'd need to change this code:

```
    if (fs::exists(tmp_std_file)) {
        status = torque_driver_parse_status(tmp_std_file, jobnr_char);
        if (status != JOB_QUEUE_STATUS_FAILURE) {
            unlink(tmp_std_file);
            unlink(tmp_err_file);
        }
    } else
        fprintf(stderr,
                "No such file: %s - reading qstat status failed\n"
                "stderr: %s\n",
                tmp_std_file, tmp_err_file);
    free(tmp_std_file);
    free(tmp_err_file);
```

AFAICT the Slurm driver checks for CANCELLED status and sets `JOB_QUEUE_IS_KILLED`. I guess we're looking for something similar.

kwinkunks:
@kvashchuka I don't quite know what to make of some of these cancelled and failed checks. These are the **Required** ones...

- [ ] Python package / test-linux (3.8, ubuntu-latest, gui-test)
- [ ] Python package / CMake (ubuntu-latest) 
- [ ] Style / check-style (3.10)

`clang-format` says this in CI:

    would reformat src/clib/lib/job_queue/torque_driver.cpp
    would reformat src/clib/old_tests/job_queue/test_job_torque.cpp
    
But I can't get `ert/script/clang-format` to run properly on my machine.

xjules:
The following failed - not sure if that is related though:
```FAILED tests/unit_tests/gui/test_main_window.py::test_that_the_plot_window_contains_the_expected_elements - AssertionError: assert ['default', 'default_0'] == ['default', '..., 'default_3']
  Right contains 3 more items, first extra item: 'default_1'
  Full diff:
  - ['default', 'default_0', 'default_1', 'default_2', 'default_3']
  + ['default', 'default_0']
===== 1 failed, 61 passed, 1340 deselected, 1 warning in 126.35s (0:02:06) =====
```

src/clib/lib/job_queue/torque_driver.cpp
```diff
@@ -317,6 +340,15 @@ stringlist_type *torque_driver_alloc_cmd(torque_driver_type *driver,
         free(resource_string);
     }
 
+    if (driver->max_runtime != 0) {
+        stringlist_append_copy(argv, "-l");
+        char *walltime_string;
+        walltime_string =
+            util_alloc_sprintf("walltime=%d", driver->max_runtime);
```
xjules, line 347: I guess that `walltime` was tested - is it in seconds?

valentin-krasontovitsch: little handy habit i learned from prometheus was to add units to your measurement variable names, but i'm not sure if it's applicable / practical here

kvashchuka: yes, it is in seconds, goos idea to add units @valentin-krasontovitsch, thank you! I think I will refactor

hnformentin:
There is some reviewers already, so moved the issue to in review. Hope this is correct :-)

docs/reference/configuration/keywords.rst
```diff
@@ -655,11 +655,14 @@ Keywords controlling the simulations
...
         ::
 
-                -- Let each realizations run for 50 seconds
+                -- Let each realization run for 50 seconds
                 MAX_RUNTIME 50
 
         The MAX_RUNTIME key is optional.
 
+        Note that if you are using a Slurm queue, then you need to pass the option
+        explcitly as a queue option. See :ref:`Slurm's options <slurm_smax_runtime>`.
```
berland, line 664: typo explcitly

kvashchuka: fixed:)

sondreso:
Is this still being worked on? 🙂

kwinkunks:
@sondreso The conclusion was that the feature is already implemented as a 'global' configuration, and the queue option implementation for SLURM should be removed (it can use the global setting). 

It will presumably be implemented in the refactor.

Closing this as redundant.

PR #5182, opened 2023-03-30
Solve race condition in qstat_proxy.sh

### LINKED ISSUE #5168: Solve possible race condition in qstat_proxy.sh
There is potentially a race condition in the between the two if-statements in:

https://github.com/equinor/ert/blob/ce2f84f4c4d7263a7b8aadaa3942db9df49ac12c/src/ert/_c_wrappers/job_queue/qstat_proxy.sh#L92-L104

where the `$proxyfile` has zero content in the second `if`-sentence while it had non-zero content in the first. This might explain errors seen in https://github.com/equinor/scout/issues/791

berland:
The code read the proxyfile from disk multiple times. Other processes can start writing to the file during the duration of this script, which can explain observations where this script gives zero exit code and zero content.

The merge of error detection and filtering into one step avoids the possible race condition.

The filtering step is simplified from awk into grep, this reduces cpu cost and code complexity, at the expense of including garbage lines in the output that the torque driver must (continue to) ignore.

**Issue**
Resolves #5168 

**Approach**
Merge two steps into one, and simplify.

## Pre review checklist

- [ ] Updated documentation

### DIFF
--- a/src/ert/_c_wrappers/job_queue/qstat_proxy.sh
+++ b/src/ert/_c_wrappers/job_queue/qstat_proxy.sh
@@ -29,6 +29,10 @@ CACHE_INVALID=10
 
 QSTAT=`which qstat 2>/dev/null`
 
+# This is a number that must be larger then largest amount of lines
+# the qstat backend can display pr. job
+MAX_JOB_LINES=200
+
 QSTAT_OPTIONS=""
 # while the first arg starts with a - character:
 while [ "$1" != "${1#-}" ]; do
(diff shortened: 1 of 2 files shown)


src/ert/_c_wrappers/job_queue/qstat_proxy.sh
```diff
@@ -84,30 +88,22 @@ fi
...
+if [ $proxyage_seconds -gt $CACHE_INVALID ]; then
+    fail "qstat_proxy: proxyfile $proxyfile was too old ($proxyage_seconds seconds). Giving up!"
 fi
 
-# Extract the job id from the proxyfile:
 if [ -n "$1" ]; then
-    awk "BEGIN { RS=\"Job Id: \"} /^$1/,/Job/ {printf \"Job Id: \"; print}" $proxyfile \
-        | grep -v ^$
-    exit 0
+    grep -A $MAX_JOB_LINES "Job Id: $1" $proxyfile || fail "qstat: Unknown Job Id $1"
```
berland, line 101: Noe that this grep filter is quite simple, if you as for job id say `100` it would also give you job id `1000`. But that should be acceptable, ERT makes sure it fetches the correct job id.

xjules: ` "Job Id: $1$" ` where the last `$1` marks end of line - would it work?

berland: That means we require any namespace to always be included, `15449.s034-lcam` is a typical job id with namespace, but `qstat 15449` also works for the backend qstat. That means filtering on end-of-line is risky and I prefer not to do it. ERTs torque driver handles the namespace issue.

src/ert/_c_wrappers/job_queue/qstat_proxy.sh
```diff
@@ -84,30 +88,22 @@ fi
 # The file is potentially updated:
 proxyage_seconds=`file_age_seconds $proxyfile`
 
-if [ $proxyage_seconds -gt $CACHE_INVALID ]; then
-    echo "qstat_proxy: proxyfile $proxyfile was too old ($proxyage_seconds seconds). Giving up!"
+fail() {
+    echo "$1"
```
xjules, line 92: echo "Job id: $1"?

tests/unit_tests/c_wrappers/res/job_queue/test_ert_qstat_proxy.py
```diff
@@ -91,7 +93,9 @@ def test_proxyfile_not_exists(tmpdir, monkeypatch):
             EXAMPLE_QSTAT_CONTENT.splitlines()
         )
     else:
-        assert len(result.stdout.splitlines()) == 5
+        lines = result.stdout.splitlines()
+        assert lines[0].decode("utf-8") == "Job Id: 15399.s034-lcam"
+        assert len(lines) >= 5
```
xjules, line 98: Why >= 5?

berland: Since qstat_proxy.sh now allows itself to give garbage lines (that means irrelevant lines for other jobs) it makes less sense to be precise on the number of lines.

xjules: Right, so that's the `grep` statement - now I got it I think :+1:

xjules (approved):
Very good! Hope it helps. :rocket:

Issue #5216, opened 2023-04-11
Unify queue option keyword interface


kwinkunks:
Assuming we agree that we should keep `MAX_RUNTIME` and `MAX_SUBMIT` as 'top-level' options, I recommend the following to unify the queue APIs a bit:

- [ ] ~Change the current [`MAX_RUNTIME` PR](https://github.com/equinor/ert/pull/5159) to use the top-level option instead.~ (Allegedly already done.)
- [ ] ~Create an issue to change the Slurm option `MAX_RUNTIME` to use the top-level option instead and deprecate the queue-specific option.~ Will be implemented this way.
- [x] Verify that `MAX_SUBMIT` is used by all queues. If not, make issue for this.
- [ ] ~Create an issue to implement `TIMEOUT` on LSF and Slurm queues, replacing `BJOBS_TIMEOUT` and `SQUEUE_TIMEOUT`.~ Will decide on this later.
- [ ] ~Create an issue to implement `INCLUDE_HOST` and `EXCLUDE_HOST` for Torque, and `INCLUDE_HOST` on LSF.~ Not decided on even using this param.
- [ ] ~Create an issue to implement `SUBMIT_SLEEP` and `DEBUG_OUTPUT` for Slurm.~ Will implement `SUBMIT_SLEEP` globally, and `DEBUG_OUTPUT` is derprecated.

## Current LSF driver options

- `QUEUE_OPTION LSF LSF_SERVER be-grid01` &mdash; unique to LSF
- `QUEUE_OPTION LSF LSF_QUEUE name_of_queue` &mdash; unique to LSF
- `QUEUE_OPTION LSF LSF_RESOURCE resource_string` &mdash; unique to LSF
- `QUEUE_OPTION LSF LSF_RSH_CMD name_of_queue` &mdash; unique to LSF
- `QUEUE_OPTION LSF LSF_LOGIN_SHELL name_of_queue` &mdash; unique to LSF
- `QUEUE_OPTION LSF BSUB_CMD command` &mdash; unique to LSF
- `QUEUE_OPTION LSF BJOBS_CMD command` &mdash; unique to LSF
- `QUEUE_OPTION LSF BKILL_CMD command` &mdash; unique to LSF
- `QUEUE_OPTION LSF BHIST_CMD command` &mdash; unique to LSF
- `QUEUE_OPTION LSF BJOBS_TIMEOUT 0` &mdash; (equivalent to `TIMEOUT` in the other queues? If so, implemented everywhere)
[...]

PR #5218, opened 2023-04-11
Change Torque option TIMEOUT to QUEUE_QUERY_TIMEOUT

berland:
The term TIMEOUT is overloaded, and something more specific is needed to distuingish this particular timeout from other timeouts. This particular timeout is specific for Torque only, and should not be confused with similar timeouts in other drivers.

This change is breaking, as ert configs using the previous option name will not have the intended effect. But usage of this option outside of developer testing should be marginal enough to not categorize this as a breaking change causing version bump.

**Issue**
Problem exposed in #5216 

**Approach**
Rename

## Pre review checklist

- [x] Updated documentation

### DIFF
--- a/docs/reference/configuration/keywords.rst
+++ b/docs/reference/configuration/keywords.rst
@@ -1855,7 +1855,7 @@ bjobs.
         flakyness.
 
         ::
-                QUEUE_OPTION TORQUE TIMEOUT 126
+                QUEUE_OPTION TORQUE QUEUE_QUERY_TIMEOUT 126
 
 .. _configuring_the_slurm_queue:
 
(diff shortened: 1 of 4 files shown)


hnformentin:
I see you did not check the testing box. Is it possible to test somehow?
Do you think it would be worth to include in some documentation (not sure where), even if not a breaking change?

berland:
I have not done a manual test on Azure yet no (it is time-consuming). Not certain how critical it is.

It should be and will be mentioned in the release notes.

hnformentin (approved):
Seems it is covered then.

I was wondering actually how to test this option, even if manually. Would be a way to make Torque to "sleep" on purpose?

If you have an idea, could you please express it before merging? I will catch up afterwards. Thank you.

berland:
I have done some testing interactively by now (until the cluster was shut down yesterday around 15:00).

By setting the QUEUE_QUERY_TIMEOUT to 1, you will get immediate failure for all but one realization (poly example is enough) due to the way the qstat_proxy works. Setting it to larger than 2 allows the qstat_proxy to work as intended, and setting it to >60 seconds makes ERT a little resilient to qstat failing.

PR #5223, opened 2023-04-12
Fix strange `gui` case name appearing for each run of EnsembleSmoother

### LINKED ISSUE #5141: Unexpected case "gui"
**Describe the bug**
After running a config, a new case called "gui" is created

**To reproduce**
1. `pip install ert`
2. `ert gui poly.ert`
3. Run Smoother, select yes to question of whether to use default case
4. Create plot
5. Observe the case list, there is a case gui with the second iteration results

**Expected behaviour**
Two cases, default with the first iteration result, default_smoother_update with the updated results.

**Screenshots**

**Environment**
 - OS: MacOS
 - ERT main, as of 24 March
 - Python version, 3.9

**Additional context**

DanSava:
**Issue**
Resolves #5141

**Approach**
Replace the `gui` name with current selected case name with a `_prior` suffix. So if a user runs EnsembleSmoother with the current case set for the `default` case. after the smoother runs two new cases are created: 

1. `default_prior`
2. `default_smoother_update`

image

## Pre review checklist

- [ ] Updated documentation

### DIFF
--- a/src/ert/gui/simulation/ensemble_smoother_panel.py
+++ b/src/ert/gui/simulation/ensemble_smoother_panel.py
@@ -91,7 +91,7 @@ class EnsembleSmootherPanel(SimulationConfigPanel):
     def getSimulationArguments(self):
         arguments = Arguments(
             mode="ensemble_smoother",
-            current_case="gui",
+            current_case=f"{self.notifier.current_case_name}_prior",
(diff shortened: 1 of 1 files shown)


andreas-el (approved):
This seems like a fair solution. Undoubtedly better than what was already there. 👍

PR #5263, opened 2023-04-21
Ignore B028 in annotations

eivindjahren:
These errors are:

> B028: No explicit stacklevel keyword argument found. The warn method from the warnings module uses a stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called. It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user. 

These are unhelpful as we have legitemate reasons to not add stacklevel to warnings

## Pre review checklist

- [ ] Updated documentation


berland:
Should we not have B011 enabled on the production code (and possibly disabled on test code)?

eivindjahren:
@berland Maybe, but personally I would rather have the solution be raising some Exception in `src/` rather than using `assert`.

berland:
https://github.com/equinor/ert/pull/5269 makes skipping B011 unneeded.

eivindjahren:
@berland I updated by no longer skipping B011 :)

berland (approved):
:+1:

PR #5286, opened 2023-04-25
Fix flake8-bugbear's B007 on unused loop control variables

berland:
**Issue**
Resolves flake8-bugbear B007 which says:
> B007: Loop control variable not used within the loop body. If this is intended, start the name with an underscore.

**Approach**
Rewrite.

## Pre review checklist

- [ ] Updated documentation

### DIFF
--- a/src/ert/gui/plottery/plots/ccsp.py
+++ b/src/ert/gui/plottery/plots/ccsp.py
@@ -41,7 +41,7 @@ def plotCrossCaseStatistics(figure, plot_context, case_to_data_map, _observation
         "p67": {},
         "p90": {},
     }
-    for case_index, (case, data) in enumerate(case_to_data_map.items()):
+    for case_index, data in enumerate(case_to_data_map.values()):
         case_indexes.append(case_index)
         std_dev_factor = config.getStandardDeviationFactor()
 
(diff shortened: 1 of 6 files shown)


src/ert/gui/plottery/plots/ccsp.py
```diff
@@ -41,7 +41,7 @@ def plotCrossCaseStatistics(figure, plot_context, case_to_data_map, _observation
         "p67": {},
         "p90": {},
     }
-    for case_index, (case, data) in enumerate(case_to_data_map.items()):
+    for case_index, data in enumerate(case_to_data_map.values()):
```
hnformentin, line 44: why do we need `data` here? Is the usage of `_` correct in this case?

```suggestion
    for case_index, _ in enumerate(case_to_data_map.values()):
```

src/ert/shared/models/iterated_ensemble_smoother.py
```diff
@@ -155,7 +155,7 @@ def runSimulations(
                 iteration=current_iter,
             )
             update_success = False
-            for iteration in range(analysis_config.num_retries_per_iter):
+            for _iteration in range(analysis_config.num_retries_per_iter):
```
hnformentin, line 158: The same here, do we need to name `_iteration` or is it only for readability?

src/ert/gui/plottery/plots/ccsp.py
berland: `_ ` is typical to use for variables that you don't really need. Since `enumerate` yields tuples, we have this dummy variable name to get the tuple unpacked.

hnformentin (approved):
I understand you prefer to keep a name. It makes sense for readability indeed.
LGTM!

PR #5297, opened 2023-04-26
Bump the queue_query timeout in Torque driver

berland:
Adding 64 seconds .

62 seconds have been proven occasionally to be insufficient.

**Issue**
Possibly resolves flakyness whenever the PBS queue system in Azure has a bad moment.

**Approach**
Increase timeout

## Pre review checklist

- [x] Updated documentation

### DIFF
--- a/docs/reference/configuration/keywords.rst
+++ b/docs/reference/configuration/keywords.rst
@@ -1847,7 +1847,7 @@ bjobs.
         intermittently not respond and give error messages when submitting jobs
         or asking for job statuses. The timeout (in seconds) determines how long
         ERT will wait before it will give up. Applies to job submission (qsub)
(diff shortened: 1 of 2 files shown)


hnformentin (commented):
If I understood correctly you updated the documentation, so the box could be checked. Is it correct?

Do you know if we have control of all timeouts we increased in order to reduce flakyness in Aure/github actions?

hnformentin (approved):
No code need to be changed I think, but please reply the comments :-)
Thanks!

berland:
> If I understood correctly you updated the documentation, so the box could be checked. Is it correct?

Correct! Fixed.

> Do you know if we have control of all timeouts we increased in order to reduce flakyness in Aure/github actions?

I don't think we have a list.

Issue #5262, opened 2023-04-20
Possible race condition in qstat_proxy.sh


berland:
The `grep` command will not care about any file locks from `flock`:

https://github.com/equinor/ert/blob/0849e45f77f1263337fc0ea715261122e40a07c8/src/ert/_c_wrappers/job_queue/qstat_proxy.sh#L101

berland:
Reproduction of race condition has been attempted unsuccessfully:

Preparation:
```
$ mkdir somedir; cd somedir
$ echo "foo" > containsfoo
``` 

One shell is running the command `set -e; while true; do grep foo containsfoo; done`. Start this. When/if the terminal closes, you have reproduced a race condition.

In another terminal do `$ while true; do echo "foo" > containsfoo.tmp && mv containsfoo.tmp containsfoo; done`

This is performing the action by the backend qstat, and we are testing really the atomicity of `mv` versus the `grep` command, if there are occasional times when `grep` sees the wrong file content due to the file shuffling. 

Results:

Over some minutes, no reproduction of a race conditions has been observed.

berland:
`man 2 rename` essentially claims that the `mv` command in this situation is atomic.

berland:
Some explanation is certainly due :)

I think the statement in the first message here is correct. From reading up on  flock, it is true that grep will not care about any locks made by flock.

The symptom we observe in some of the Azure runs, is that the queue driver fails hard for some realizations with JOB_QUEUE_STATUS_FAILURE. This happens at least if the driver gets an empty file to process, and a zero exit status. Then I have spent time thinking whether it is possible that the qstat_proxy could cause this scenario, and that lead me to the statement in this issue title. 

The proxy shell script tries to mitigate this problem by a mv command, which is probably atomic on /tmp. What is up for review, is that the mv command is proven (by the tests described above) to be atomic in this context, and thus the grep line cannot be responsible for the errors we have seen.

valentin-krasontovitsch:
thanks for the detailed explanation : )

I'm gonna try to understand a bit better by reading the docs and playing with the things, and perhaps approach you with further questions ^^

valentin-krasontovitsch:
ok, little control question, i'm probably missing something: if we suspect the problem to be an invocation of grep which potentially ignores an flock, and if we have control of the flock, why not invoke grep by wrapping it with flock?

berland:
[...]

PR #5324, opened 2023-05-02
Fix possible error source in qstat-proxy script

valentin-krasontovitsch:
The qstat proxy script calls the qstat backend, and records the result in a file used as a caching proxy for the backend.

When it updates that cache file by calling the backend, it stores the result in a temp file first, and then moves that temp file to the actual cache file location (for race conditions reasons).

This moving of the temp file happens regardless of the return code of the qstat backend. We mitigate here by only moving if the call to the backend was successful.

Possibly related to https://github.com/equinor/scout/issues/803 and https://github.com/equinor/ert/issues/5262


berland (approved):
👍🏻

PR #5325, opened 2023-05-02
Reformat lint cli message

### LINKED ISSUE #5300: Create problem-matcher friendly CLI output of errors
The problem matcher needs to match an error to its filename, line, column, etc on a per-error basis, so the output grouped by filename does not work. It should be formatted like this (inspired by flake8 and other linters etc): 
```filename:start_line:start_column:end_column:message```. Also, all messages output by linter should not contain newlines

### LINKED ISSUE #5319: lark parse will raise an UnexpectedInputError which we should wrap in ErrorInfo
This error messagewill format with new-lines which may not show well in the gui tab:

https://github.com/equinor/ert/blob/daea5dfe98e874b157491162eccf93115d1189a2/src/ert/parsing/lark_parser.py#L324-L325

The error contains location info:

https://github.com/lark-parser/lark/blob/2de51ceaa542d4be322cba034ee29a2d887bfd19/lark/exceptions.py#L37

We should extract that info and put it into the ErrorInfo object.

### PR DESCRIPTION
**Pull request by yngve-sk**
**Issue**
Resolves #5300

Partially fixes #5319 (adding error info & localization to the errors in that method should be a separate PR)

**Approach**
* Format CLI message in standard linting format
* Remove newlines from errors
* Unexpected characters message without newlines

## Pre review checklist

- [ ] Updated documentation

### DIFF
--- a/src/ert/_c_wrappers/enkf/ensemble_config.py
+++ b/src/ert/_c_wrappers/enkf/ensemble_config.py
@@ -389,7 +389,7 @@ class EnsembleConfig(BaseCClass):
         elif not Path(base_surface).exists():
             errors.append(f"BASE_SURFACE:{base_surface} not found")
         if errors:
-            errors = "\n".join(errors)
+            errors = ". ".join(errors)
             raise ConfigValidationError(
                 f"SURFACE {name} incorrectly configured: {errors}"
             )
--- a/src/ert/_c_wrappers/enkf/ensemble_config.py
+++ b/src/ert/_c_wrappers/enkf/ensemble_config.py
@@ -389,7 +389,7 @@ class EnsembleConfig(BaseCClass):
         elif not Path(base_surface).exists():
             errors.append(f"BASE_SURFACE:{base_surface} not found")
         if errors:
-            errors = ". ".join(errors)
+            errors = ";".join(errors)
             raise ConfigValidationError(
                 f"SURFACE {name} incorrectly configured: {errors}"
             )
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: src/ert/_c_wrappers/enkf/ensemble_config.py [hunk 1]
```diff
@@ -389,7 +389,7 @@ def get_surface_node(surface: List[str]) -> SurfaceConfig:
         elif not Path(base_surface).exists():
             errors.append(f"BASE_SURFACE:{base_surface} not found")
         if errors:
-            errors = "\n".join(errors)
+            errors = ". ".join(errors)
```
DanSava, line 392: Not sure how this will show in the suggestor dialog but would it make sense to separate the errors with `;`? It might be a bit easier to separate them when reading multiple errors.

src/ert/parsing/config_errors.py
```diff
@@ -40,36 +42,25 @@ def _get_error_message(cls, info: ErrorInfo) -> str:
...
-
-        result = ";".join(
-            [
-                f"Parsing config file `{filename}` resulted in the errors: \n"
-                + "\n".join(
-                    [
-                        indent_only_first_line(err.message_with_location, "  * ")
-                        for err in info_list
-                    ]
+        nice_messages = []
```
DanSava, line 53: Nice code cleanup :+1: would you consider renaming this variable to just `messages`? I know it is not as nice as `nice_messages` (pun intended) but a bit shorter  :smile:.

DanSava (approved):
Looks good :+1:  just some very minor comments.

src/ert/parsing/config_errors.py
yngve-sk: Done!

src/ert/_c_wrappers/enkf/ensemble_config.py
yngve-sk: Makes sense, and done :)

PR #5346, opened 2023-05-08
Mitigate partial coverage uploads from ci.

eivindjahren:
To stop partial uploads to codecov which makes the coverage report weird, we stop canceling on the main branch which codecov uses to compair against.

Also, we fail the coverage when the upload fails. This happens very rarely, but frequent enough (roughly once every other week). For PR's, most checks are optional except the cmake one, which is fairly fast to rerun.

## Pre review checklist

- [ ] Updated documentation


andreas-el (approved):
LGTM! :bacon:

PR #5357, opened 2023-05-09
Change some BaseCEnum classes to python Enums

eivindjahren:
Move some enums into python as they are no longer used. Other enums are just removed as they are used nowhere.

## Pre review checklist

- [ ] Updated documentation


src/ert/load_status.py
```diff
@@ -0,0 +1,13 @@
+from enum import Enum
+from typing import NamedTuple, Optional
+
+
+class LoadStatus(Enum):
+    LOAD_SUCCESSFUL = 0
+    LOAD_FAILURE = 2
+    TIME_MAP_FAILURE = 3
```
valentin-krasontovitsch, line 8: i see that you kept the values of `0,2,3` from before - any reason to keep skipping `1`?

valentin-krasontovitsch (approved):
beautiful

src/ert/load_status.py
eivindjahren: Probably existed a value for one which was removed.

PR #5391, opened 2023-05-11
Use 32 bit datatypes in tests

berland:
**Issue**
Resolves hundreds of warnings from the test suite

**Approach**
Use special function from hypothesis to ensure correct dtypes.

## Pre review checklist

- [ ] Updated documentation

### DIFF
--- a/tests/performance_tests/performance_utils.py
+++ b/tests/performance_tests/performance_utils.py
@@ -17,14 +17,14 @@ from ert.dark_storage import enkf
 
 def write_summary_spec(file, keywords):
     content = [
-        ("INTEHEAD", [1, 100]),
+        ("INTEHEAD", array([1, 100], dtype=numpy.int32)),
         ("RESTART ", [b"        "] * 8),
-        ("DIMENS  ", [1 + len(keywords), 10, 10, 10, 0, -1]),
+        ("DIMENS  ", array([1 + len(keywords), 10, 10, 10, 0, -1], dtype=numpy.int32)),
         ("KEYWORDS", [f"{x: <8}" for x in ["TIME"] + keywords]),
         ("WGNAMES ", [b":+:+:+:+"] * (len(keywords) + 1)),
-        ("NUMS    ", [-32676] + ([0] * len(keywords))),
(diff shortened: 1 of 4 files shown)


tests/test_config_parsing/summary_generator.py
```diff
@@ -231,15 +232,21 @@ class Unsmry:
...
     def to_file(
         self, filelike, file_format: ecl_data_io.Format = ecl_data_io.Format.UNFORMATTED
     ):
         ecl_data_io.write(filelike, self.to_ecl(), file_format)
 
 
-positive_floats = st.floats(min_value=0.1, allow_nan=False, allow_infinity=False)
+positive_floats = from_dtype(
+    np.dtype(np.float32),
+    min_value=0.10000000149011612,  # 0.1 rounded to float32
```
xjules, line 246: Hm, why not `min_value=0.1`?

tests/performance_tests/performance_utils.py
```diff
@@ -17,14 +17,14 @@
 
 def write_summary_spec(file, keywords):
     content = [
-        ("INTEHEAD", [1, 100]),
+        ("INTEHEAD", array([1, 100], dtype=numpy.int32)),
```
xjules, line 20: btw, you can also do `array(...).astype(np.int32)`, which applies for all the others.

berland: are you sure that will not create the array with double values and then do the astype-conversion?

xjules: Good point!

tests/test_config_parsing/summary_generator.py
xjules: Can you put this line as a comment there ? Just that people / us would understand when we look at it in a year or so :smile:

berland: Comment updated! :)

berland: No, changed to this instead:
`
+    min_value=np.float32(0.1),
`

xjules (approved):
Very good! :rocket:

PR #5397, opened 2023-05-11
Improve parsing of options

dafeda:
## Pre review checklist

- [ ] Updated documentation

### DIFF
--- a/src/ert/_c_wrappers/enkf/ensemble_config.py
+++ b/src/ert/_c_wrappers/enkf/ensemble_config.py
@@ -38,56 +38,48 @@ def _get_abs_path(file):
     return file
 
 
-def _get_filename(file):
-    if file is not None:
-        file = os.path.basename(file)
-    return file
+def _parse_options(option_list: List[str], offset: int) -> Dict[str, str]:
+    """Converts a list of options given to keywords such as GEN_DATA into a dictionary.
 
+    Args:
+        option_list: The list of options. Each option should be a string
+            formatted as 'KEY:VALUE'.
+        offset: The index in the list from where to start parsing the options.
 
-def _option_dict(option_list: List[str], offset: int) -> Dict[str, str]:
-    """Gets the list of options given to a keywords such as GEN_DATA.
+    Returns:
+        A dictionary where the keys are the options and the values
+        are the arguments for those options.
 
-    The first step of parsing will separate a line such as
-
-      GEN_DATA NAME INPUT_FORMAT:ASCII RESULT_FILE:file.txt REPORT_STEPS:3
-
-    into
-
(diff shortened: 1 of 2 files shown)


src/ert/_c_wrappers/enkf/ensemble_config.py
```diff
@@ -115,6 +104,8 @@ def _str_to_bool(txt: str) -> bool:
     Any text which is not correctly identified as true or false returns False, but
     with a failure message written to the log:
 
+    >>> import sys
```
eivindjahren, line 107: Probably a good idea to change this to a regular test, instead of a doctest.

src/ert/_c_wrappers/enkf/ensemble_config.py
```diff
@@ -38,56 +38,45 @@ def _get_abs_path(file):
...
+    for option in option_list[offset:]:
+        try:
+            key, value = option.split(":")
 
-    """
-    option_dict = {}
-    for option_pair in option_list[offset:]:
-        if not isinstance(option_pair, str):
+        except ValueError:
             logger.warning(
```
eivindjahren, line 65: This was added a while ago, before we started using ConfigWarning to signal warnings that are supposed to be shown to the user. Could we use `warnings.warn(msg, category=ConfigWarning)` instead of `logger.warning` while we are at it.

src/ert/_c_wrappers/enkf/ensemble_config.py
dafeda: What if we instead simplify the `doctest` to just test that `_str_to_bool("fail")` returns `False`?
I don't think there's much benefit to test that the logger logs.

eivindjahren: Sure

src/ert/_c_wrappers/enkf/ensemble_config.py
dafeda: Just so that we are on the same page.
The code as it is in `main` today would not raise any warnings if this test is run.
That is, it ignores both "FORWARD_INIT:Frue" and "not_an_opt".
My suggested change would still ignore "FORWARD_INIT:Frue" but would raise on "not_an_opt" because it can't split it.

eivindjahren: It is still not raise though, it just makes the warning visible to the user.

dafeda: I'm just saying that the way the logic works on `main` ignores things that the current code will not ignore.

eivindjahren: Would be nice to have a test for the new behavior then. Also, if there is a new error that is expected in case of user error, then we should probably reraise it as `ConfigValidationError` to show it correctly to the user.

dafeda:
Closing because stale.

PR #5425, opened 2023-05-16
Revert "Explicitly use mixed format when converting to dates, avoiding warning"

berland:
Reverts equinor/ert#5417

The warning comes with Pandas 2.0.1, not with Pandas 1.5.3. 

#5417 fixed the warning in Pandas 2.0.1, but errors hard in Pandas 1.5.3.


xjules (approved):
:+1:

Issue #5355, opened 2023-05-09
Remove old parser


eivindjahren:
On hold until we have had the feature-flag flipped defaulting to new parser in stable for at least one month ( merged in #5361 )

- [x] Remove old suggester `./src/ert/_c_wrappers/enkf/_deprecation_migration_suggester.py`

PR #5450, opened 2023-05-22
Decouple parsing from content_type_enum

### LINKED ISSUE #5449: Uncouple from ContentTypeEnum
Much of the parsing code makes use of ContentTypeEnum, which ties into the old parser and its C++ code.

yngve-sk:
**Issue**
Resolves #5449 
Contributes towards #5355

### DIFF
--- a/src/ert/parsing/config_schema_item.py
+++ b/src/ert/parsing/config_schema_item.py
@@ -1,6 +1,6 @@
 import os
 import shutil
-from typing import Any, List, Mapping, Optional, TypeVar, Union
+from typing import List, Mapping, Optional, TypeVar, Union
 
 from pydantic import BaseModel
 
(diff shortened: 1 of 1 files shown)


src/ert/parsing/config_schema_item.py
```diff
@@ -28,7 +28,7 @@ class SchemaItem(BaseModel):
     # The maximum number of arguments: -1 means no upper limit
     argc_max: int = 1
     # A list of types for the items. Set along with argc_minmax()
-    type_map: List[Optional[SchemaItemType]] = []
+    type_map: List[Optional[Any]] = []
```
eivindjahren, line 31: Think this should be the way the was.

eivindjahren (approved):
Need to remove the `Any` change, otherwise it is good.

PR #5476, opened 2023-05-25
Use new parser for workflows

### LINKED ISSUE #5475: New parser for workflows
The new parser is not yet added for workflows

yngve-sk:
**Issue**
Resolves #5475 

**Approach**
Use new parser for workflows
* Lark parser has to create arbitrary schema items as keywords can be anything but `DEFINE`, thus it is necessary to override `__getitem__` in `WorkflowSchemaDict`.
* The order of jobs in a workflow matters, so we use the line number of the keyword to order them after being parsed by the lark parser. Lark parser parses them into a dict keyed by the job name, with one arglist (possibly empty) per occurrence. It is therefore necessary to attach the context to the arglist parsed in by lark, and this is done the config schema item.

## Pre review checklist

- [ ] Updated documentation

### DIFF
--- a/src/ert/_c_wrappers/enkf/ert_config.py
+++ b/src/ert/_c_wrappers/enkf/ert_config.py
@@ -644,7 +644,7 @@ class ErtConfig:
                     work[0],
                     substitution_list,
                     workflow_jobs,
-                    use_new_parser=USE_NEW_PARSER_BY_DEFAULT,
+                    use_new_parser=use_new_parser,
                 )
                 if existed:
                     warnings.warn(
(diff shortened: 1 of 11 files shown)


src/ert/_c_wrappers/enkf/ert_config.py
```diff
@@ -644,6 +644,7 @@ def _workflows_from_dict(
                     work[0],
                     substitution_list,
                     workflow_jobs,
+                    use_new_parser=USE_NEW_PARSER_BY_DEFAULT,
```
eivindjahren, line 647: This should be set to the same as the `use_new_parser` parameter and not the constant.

andreas-el (changes requested):

andreas-el (approved):
LGTM! :rocket:

PR #5484, opened 2023-05-25
Substitute existing defines in lark parser

### LINKED ISSUE #5485: DEFINE not overwriting old defines in new parser
**Describe the bug**
When parsing defines with the lark parser, re-defines are appended to the list of defines, rather than substituting the define.

yngve-sk:
**Issue**
Resolves  #5485 

**Approach**

## Pre review checklist

- [ ] Updated documentation

### DIFF
--- a/src/ert/parsing/lark_parser.py
+++ b/src/ert/parsing/lark_parser.py
@@ -4,7 +4,7 @@ import logging
 import os
 import os.path
 import warnings
-from typing import Any, List, Optional, Tuple, Union
+from typing import List, Optional, Tuple, Union
 
 from lark import Discard, Lark, Token, Transformer, Tree, UnexpectedCharacters
 from typing_extensions import Self
@@ -143,11 +143,9 @@ _parser = Lark(grammar, propagate_positions=True)
 logger = logging.getLogger(__name__)
 
 
-def _substitute(
+def _substitute_token(
     defines: Defines,
-    token: Union[
-        FileContextToken, List[Tuple[FileContextToken]], Tuple[FileContextToken]
-    ],
+    token: FileContextToken,
     expand_env: bool = True,
 ) -> str:
     #    if isinstance(token, (list, tuple)):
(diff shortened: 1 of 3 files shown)


tests/test_config_parsing/test_ert_config_parsing.py
```diff
@@ -1021,3 +1026,35 @@ def test_that_multiple_errors_are_shown_for_forward_model():
...
+
+
+@pytest.mark.usefixtures("use_tmpdir")
+def test_that_redefines_are_applied_correctly():
+    test_config_file_name = "test.ert"
+    test_config_contents = dedent(
+        """
+        NUM_REALIZATIONS  1
+        DEFINE <A> 2
+        DEFINE B <A>
```
eivindjahren, line 1038: Would be good to test that intermediate usages are substituted. So try a multioccurrance `KW <A>` and check that it first gets argument `2` and then later `3`.

src/ert/parsing/lark_parser.py
```diff
@@ -236,9 +250,11 @@ def _substitute_args(
 ) -> List[Any]:
     if constraints.substitute_from < 1:
         return args
+
     return [
         _substitute(defines, x, constraints.expand_envvar)
```
eivindjahren, line 255:
```python
def _substitute_all(i, constraints, defines, x, expand_envvar):
   if i+1 >= constraints.substitute_from and isinstance(x, FileContextToken):
      return _substitute(defines, x, constraints.expand_envvar)
   if isinstance(x, (list, tuple)):
      return [_substitute_all(i, defines, t, expand_env) for t in token]
   return x
   
 return [
        _substitute_all(defines, x, constraints.expand_envvar)
        if i + 1 >= constraints.substitute_from
        and isinstance(x, (FileContextToken, list))
        else x
        for i, x in enumerate(args)
    ]
````

File: src/ert/parsing/lark_parser.py [hunk 3]
```diff
@@ -210,9 +214,19 @@ def _tree_to_dict(
             args = _substitute_args(args, constraints, defines)
             value_list = constraints.apply_constraints(args, kw, cwd)
 
-            if constraints.multi_occurrence:
```
eivindjahren, line 213: Does this work for SET_ENV also?
```
DEFINE <X> 3
SETENV VAR <X>
DEFINE <X> 4
SETENV VAR2 <X>
```

src/ert/parsing/lark_parser.py
```diff
@@ -231,17 +240,52 @@ def _tree_to_dict(
...
+            # It is a list of keyword tuples
+            return [substitute_arglist_tuple(x) for x in arg]
+
+        raise ValueError(
+            f"Expected "
+            f"Union[FileContextToken, List[Tuple[FileContextToken]]], "
+            f"got {arg}"
+        )
+
+    return [substitute_arg(i, arg) for i, arg in enumerate(args)]
```
eivindjahren, line 288: I was thinking that maybe `return [substitute_arg(arg) if (i + 1) >= constraints.substitute_from else arg for i, arg in enumerate(args)]` might be a little easier to read.

src/ert/parsing/lark_parser.py
```diff
@@ -231,17 +240,52 @@ def _tree_to_dict(
...
-        if i + 1 >= constraints.substitute_from and isinstance(x, FileContextToken)
-        else x
-        for i, x in enumerate(args)
-    ]
+
+    def substitute_arglist_tuple(
+        tup: Tuple[FileContextToken],
+    ) -> Tuple[FileContextToken]:
+        tuple_kw = tup[0]
+        tuple_args = tup[1:]
```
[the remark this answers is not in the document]
yngve-sk: Fixed. 

Was not really sure if it can somehow in some cases receive more, though it seems it is not supported in currently:

```
FORWARD_MODEL MAKE_SYMLINK(<U>=<A> <B> <C>)
FORWARD_MODEL MAKE_SYMLINK(<U>=<A>,<B>,<C>)
```

fails in the new parsre, so assuming it is only two

src/ert/parsing/lark_parser.py
```diff
@@ -210,9 +209,19 @@ def _tree_to_dict(
             args = _substitute_args(args, constraints, defines)
             value_list = constraints.apply_constraints(args, kw, cwd)
 
-            if constraints.multi_occurrence:
-                arglist = config_dict.get(kw, [])
-                arglist.append(value_list)  # type: ignore
+            arglist = config_dict.get(kw, [])
+            if kw == "DEFINE":
+                define_key, *define_args = value_list
+                existing_define = next(
```
eivindjahren, line 215: So there were some questions regarding whether this is expected behavior regarding situations where order is significant. Order should never be significant given that the user writes `define_key` as "<ARG>" which I believe we require now. However, if the `define_key` for instance was `<A<B><C>>` and you had other dines such as `<AB<C>>` and `<A<B>C>` then order would be significant. I am wondering whether we should write this as

`define = [kw, arg for kw, arg in arglist if kw != define_key] + [(define_key, define_args)]` to have a consistent order of defines.

PR #5502, opened 2023-05-30
Make experiment server websockets>=11.0 compatible

### LINKED ISSUE #5199: Unpin websockets<11.0
According to the changelog, websockets 11 comes with some breaking changes resulting unit tests to fail. The main reason is the experiment server which is touched by the following two changes mainly:

- Closing a connection without an empty close frame is OK.

Receiving an empty close frame now results in [ConnectionClosedOK](https://websockets.readthedocs.io/en/stable/reference/exceptions.html#websockets.exceptions.ConnectionClosedOK) instead of [ConnectionClosedError](https://websockets.readthedocs.io/en/stable/reference/exceptions.html#websockets.exceptions.ConnectionClosedError).

As a consequence, calling WebSocket.close() without arguments in a browser isn’t reported as an error anymore.

- [serve()](https://websockets.readthedocs.io/en/stable/reference/asyncio/server.html#websockets.server.serve) times out on the opening handshake after 10 seconds by default.

You can adjust the timeout with the open_timeout parameter. Set it to [None](https://docs.python.org/3/library/constants.html#None) to disable the timeout entirely.

See here for the complete list of changes:

https://websockets.readthedocs.io/en/stable/project/changelog.html

xjules:
- We need to call `dispatcher.send` within the same connection establish context since connection is now being closed automatically when existing the contextmanager.
- ~~We need to provide at least one awaitable after creating the server task, since connect itself has await-able when specifying `open_timeout` parameter in client.connect function but currently it looks to be ignored (?) in Python 3.8 https://github.com/python-websockets/websockets/blob/1bf73423044dedc325435d261a39473337f5ddcf/src/websockets/legacy/client.py#L646~~
Seems like using `async for x in connect(x...): ` solves the problem.

**Issue**
Resolves #5199 

**Approach**
Remove dispatcher factory and provide `await asyncio.sleep()`

## Pre review checklist

- [ ] Updated documentation

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -147,7 +147,7 @@ args = dict(
         "scipy<1.9.2",
         "sqlalchemy",
         "uvicorn >= 0.17.0",
-        "websockets >= 9.0.1, <11.0",
+        "websockets",
         "httpx",
         "tables",
         "xarray",
--- a/tests/unit_tests/experiment_server/test_server.py
+++ b/tests/unit_tests/experiment_server/test_server.py
@@ -1,10 +1,11 @@
-from typing import AsyncContextManager, Awaitable, Callable
+from typing import AsyncContextManager
 from unittest.mock import AsyncMock
 
 import pytest
 from cloudevents.conversion import to_json
 from cloudevents.http import CloudEvent
-from websockets.client import WebSocketClientProtocol
+from websockets.client import connect
+from websockets.exceptions import ConnectionClosed
 
 import ert.experiment_server
 from ert.experiment_server._experiment_protocol import Experiment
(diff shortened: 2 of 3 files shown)


tests/unit_tests/experiment_server/test_server.py
```diff
@@ -15,19 +16,16 @@
...
         experiment = AsyncMock(Experiment)
         experiment_server.add_experiment(experiment)
 
-        async with dispatcher_factory as make_dispatcher:
-            dispatcher = await make_dispatcher(experiment_server)
+        # we need to use awaitable here due to open_timeout is "problematic"
+        # in websockets for python3.8
+        await asyncio.sleep(0)
 
+        async with connect(experiment_server._config.dispatch_uri) as dispatcher:
```
berland, line 28: Using this:
```
       async for dispatcher in connect(experiment_server._config.dispatch_uri, open_timeout=None):
            try:
                event = CloudEvent(
                {
                    "type": "test.event",
                    "source": "test_receiving_event_from_cluster",
                }
            )
                await dispatcher.send(to_json(event).decode())
                break
            except websockets.ConnectionClosed:
                continue
```
will not require the use of asyncio.sleep(). I am not getting any use from open_timeout yet, but my hunch is that `async for` is the way to go for this kind of robust reconnect.

setup.py
```diff
@@ -147,7 +147,7 @@ def package_files(directory):
         "scipy<1.9.2",
         "sqlalchemy",
         "uvicorn >= 0.17.0",
-        "websockets >= 9.0.1, <11.0",
+        "websockets >= 9.0.1",
```
berland, line 150: 9.0.1 is from may 2021, maybe old enough that we can drop it as a lower bound (we probably don't even know if that lower bound is sufficient)

berland:
Since we are only fixing the test code in this PR, do we know if the production code is not hit when we bump websockets - in case the production code is using the same pattern as the test code prior to this PR?

xjules:
> Since we are only fixing the test code in this PR, do we know if the production code is not hit when we bump websockets - in case the production code is using the same pattern as the test code prior to this PR?

As mentioned, I run the poly example and worked as it should. No LSF test yet though.

berland (approved):
LGTM;

PR #5528, opened 2023-06-02
Update eclrun tests

berland:
**Issue**
Maintenance of test_eclrun. It should not change any behaviour.

Note: These tests for eclipse are skipped in CI, one needs to supply `--eclipse-simulator` to  `pytest` to activate them. Works on my machine.

Reviewer may decide whether to squash commits.

## Pre review checklist

- [ ] Updated documentation

### DIFF
--- a/tests/unit_tests/shared/share/test_ecl_run.py
+++ b/tests/unit_tests/shared/share/test_ecl_run.py
@@ -3,6 +3,7 @@ import re
 import shutil
 import stat
 from distutils.spawn import find_executable
+from pathlib import Path
 from subprocess import PIPE, Popen
 
 import pytest
@@ -21,16 +22,12 @@ from ._import_from_location import import_from_location
 
 ecl_config = import_from_location(
     "ecl_config",
-    os.path.join(
-        SOURCE_DIR, "src/ert/shared/share/ert/forward-models/res/script/ecl_config.py"
-    ),
+    SOURCE_DIR / "src/ert/shared/share/ert/forward-models/res/script/ecl_config.py",
 )
 
 ecl_run = import_from_location(
     "ecl_run",
-    os.path.join(
-        SOURCE_DIR, "src/ert/shared/share/ert/forward-models/res/script/ecl_run.py"
-    ),
+    SOURCE_DIR / "src/ert/shared/share/ert/forward-models/res/script/ecl_run.py",
 )
 
 
@@ -51,8 +48,8 @@ def find_version(output):
     return re.search(r"flow\s*([\d.]+)", output).group(1)
 
 
-@pytest.fixture()
-def init_ecl100_config(monkeypatch, tmpdir):
+@pytest.fixture(name="init_ecl100_config")
+def fixture_init_ecl100_config(monkeypatch, tmpdir):
     ecl14_prefix = "/prog/ecl/grid/2014.2/bin/linux_x86_64/"
     ecl19_prefix = "/prog/res/ecl/grid/2019.3/bin/linux_x86_64/"
     mpi_prefix = "/prog/ecl/grid/tools/linux_x86_64/intel/mpi/5.0.2.044/"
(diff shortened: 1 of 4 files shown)


tests/unit_tests/shared/share/test_ecl_run.py
```diff
@@ -92,14 +89,14 @@ def init_ecl100_config(monkeypatch, tmpdir):
...
+        with open("ecl100_config.yml", "w", encoding="utf-8") as filehandle:
+            filehandle.write(yaml.dump(conf))
         monkeypatch.setenv("ECL100_SITE_CONFIG", "ecl100_config.yml")
         yield
 
 
-@pytest.fixture
-def init_flow_config(monkeypatch, tmpdir):
+@pytest.fixture(name="init_flow_config")
+def fixture_init_flow_config(monkeypatch, tmpdir):
```
hnformentin, line 99: Why `name=init_flow_config` needs to be there? I could not find it quickly on a search...

tests/unit_tests/shared/share/test_ecl_run.py
```diff
@@ -375,54 +370,31 @@ def test_run(init_ecl100_config, source_root):
...
     )
     econfig = ecl_config.Ecl100Config()
-    sim = econfig.sim("2014.2")
+    sim = econfig.sim("2019.3")
     erun = ecl_run.EclRun("SPE1.DATA", sim)
     erun.runEclipse()
 
     ok_path = os.path.join(erun.runPath(), f"{erun.baseName()}.OK")
-    log_path = os.path.join(erun.runPath(), f"{erun.baseName()}.LOG")
+    prt_path = os.path.join(erun.runPath(), f"{erun.baseName()}.PRT")
```
hnformentin, line 378: What kind of extension is `prt`? I result of a search and chatGPT summarised in a way that does not make sense for me here 🤔 : 

> ".prt" (Part) file extension:
> CAD Software: In the field of computer-aided design (CAD), ".prt" files are often associated with 3D models created using software like PTC Creo, Siemens NX, or SolidWorks. These files contain the geometric and structural information of a part or component.
> Printer-related: ".prt" can also be used as an abbreviation for "print" in certain applications, where the file may contain instructions or settings specific to printing a document or image.

tests/unit_tests/shared/share/test_ecl_run.py
berland: It is `pylint` that gives a warning on `redefined-outer-name` prior to this PR. The solution to that pylint issue is to do like this. 
https://docs.pytest.org/en/stable/reference/reference.html#pytest-fixture
https://itecnote.com/tecnote/python-pytest-fixtures-redefining-name-from-outer-scope-pylint/

tests/unit_tests/shared/share/test_ecl_run.py
berland: ChatGPT is utterly wrong here.. PRT is a extension name chosen by the Eclipse simulator and is shorthand for "PRINT". We call the PRT-file for the PRINT-file from Eclipse, representing the same thing as a log file (why there is sometimes a LOG file as well I do not know the details of).

tests/unit_tests/shared/share/test_ecl_run.py
hnformentin: perfect!

tests/unit_tests/shared/share/test_ecl_run.py
hnformentin: Given the confusion I made, what do you think about renaming to `print_file_path` instead of `prt_file`?
But maybe that is obvious for people in general and then the comment can be considered irrelevant :-)

berland: I can remove, or at least comment. This is only obvious to Eclipse people, not to ERT-programmers.

hnformentin (approved):
I think squashing would be better because tests were not run in each individual commit.
But if possible it would be nice to have a main commit messages and the other commits squashed inside.
I left a comment regarding `prt`, feel free to rename it or not.

PR #5532, opened 2023-06-05
Migration fails gracefully

pinkwah:
Add a try-catch wrapper around migration that logs the error and continues.

Resolve: #5530


pinkwah:
NOT tested on Drogon.

oyvindeide (approved):

PR #5534, opened 2023-06-06
Trust a nonzero exit value from qstat

berland:
This fixes a bug in 7f6cfb2, where the exit code was disregarded fully, which lead to qstat outputs like 'Unknown Job Id' being interpreted as successful qstat runs.

**Issue**
Resolves non-working hm-tutorial nightly runs.

**Approach**
Given by the PR title.

## Pre review checklist

- [ ] Updated documentation

### DIFF
--- a/src/clib/lib/job_queue/torque_driver.cpp
+++ b/src/clib/lib/job_queue/torque_driver.cpp
@@ -615,16 +615,22 @@ torque_driver_get_qstat_status(torque_driver_type *driver,
            retry a couple of times with exponential sleep. ERT pings qstat
            every second for every realization, thus the initial sleep time
            is 2 seconds. */
+        int return_value = -1;
         bool qstat_succeeded = false;
         int retry_interval = 2; /* seconds */
         int slept_time = 0;
         while ((!qstat_succeeded) && (slept_time <= driver->timeout)) {
-            util_spawn_blocking(driver->qstat_cmd, argv.size(), argv.data(),
-                                tmp_std_file, tmp_err_file);
-            if (fs::exists(tmp_std_file) && (fs::file_size(tmp_std_file) > 0)) {
-                // A non-existing or zero length output from qstat is interpreted
-                // as a failure. ERT never calls qstat unless it has already submitted
-                // something.
+            return_value =
+                util_spawn_blocking(driver->qstat_cmd, argv.size(), argv.data(),
+                                    tmp_std_file, tmp_err_file);
+            // A non-zero return value is trusted, but a zero return-value
+            // is not trusted unless the output has nonzero length.
+            // ERT never calls qstat unless it has already submitted something, and
+            // can therefore assume that qstat results about Unknown Job Id are
+            // failures (these have nonzero output length, but return value != 0)
+            // that should trigger retries.
+            if (std::error_code ec; fs::file_size(tmp_std_file, ec) > 0 && ec &&
+                                    return_value == 0) {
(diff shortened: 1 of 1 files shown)


src/clib/lib/job_queue/torque_driver.cpp
```diff
@@ -615,13 +615,17 @@ torque_driver_get_qstat_status(torque_driver_type *driver,
...
-            util_spawn_blocking(driver->qstat_cmd, argv.size(), argv.data(),
-                                tmp_std_file, tmp_err_file);
-            if (fs::exists(tmp_std_file) && (fs::file_size(tmp_std_file) > 0)) {
+            return_value =
+                util_spawn_blocking(driver->qstat_cmd, argv.size(), argv.data(),
+                                    tmp_std_file, tmp_err_file);
+            // A zero return-value is not trusted, but a nonzero is.
+            if (fs::exists(tmp_std_file) && (fs::file_size(tmp_std_file) > 0) &&
+                return_value == 0) {
                 // A non-existing or zero length output from qstat is interpreted
```
xjules, line 629: I guess this comment does not make sense then or?

berland: Yes, it does make sense. But I can see it can be hard to make sense of it.

berland: Comments rewritten.

xjules (approved):
LGTM! :rocket:

src/clib/lib/job_queue/torque_driver.cpp
```diff
@@ -615,16 +615,22 @@ torque_driver_get_qstat_status(torque_driver_type *driver,
...
+            return_value =
+                util_spawn_blocking(driver->qstat_cmd, argv.size(), argv.data(),
+                                    tmp_std_file, tmp_err_file);
+            // A non-zero return value is trusted, but a zero return-value
+            // is not trusted unless the output has nonzero length.
+            // ERT never calls qstat unless it has already submitted something, and
+            // can therefore assume that qstat results about Unknown Job Id are
+            // failures (these have nonzero output length, but return value != 0)
+            // that should trigger retries.
+            if (fs::exists(tmp_std_file) && (fs::file_size(tmp_std_file) > 0) &&
```
pinkwah, line 632: `fs::exists` returns true if the "file" exists. But file here means a UNIX file, and directories are files too. This also doesn't cover potential permission issues. These errors would through an exception, potentially causing ERT to crash for a unreasonably.

Consider instead to use `fs::file_size` directly and treating IO errors as failure.

```suggestion
            std::error_code ec;
            if (fs::file_size(tmp_std_file, ec) > 0 && ec &&
```

`ec` is convertible to `bool`, and if `ec` is truthy, then there was no error. `fs::file_size` returns `std::uintmax_t -1`, but because it's an unsigned type, this becomes the largest possible value, so we have to check that the file-size is non-zero _and_ that there was no error obtaining the file-size.

https://en.cppreference.com/w/cpp/error/error_code
https://en.cppreference.com/w/cpp/filesystem/file_size

pinkwah (dismissed):
Please address the above comment. But feel free to ignore it if you wish I am not your mother. ;)

src/clib/lib/job_queue/torque_driver.cpp
berland: thanks! Will fix

pinkwah (approved):

PR #5535, opened 2023-06-06
Fix intercomponent communication instability when running many realizations (>=200)

valentin-krasontovitsch:
**Issue**
Related to #5273, might solve it

NB: one left issue is that the proposed pattern in queue.py will continue to try to reconnect basically forever (OBS: maybe not if the socket gets closed, and there is nobody listening on the port anymore!) - this should perhaps be adjusted, which is sadly non-trivial based on how the loop is set up (where to catch the exception to count reconnects, how to recover from possible partial data submission, ...)

**Approach**
We traced the symptoms back to the websockets server and client.
We increase timeouts, and change the websockets connection usage in the job queue.
We also dramatically increase loggin - *NB* that particular last commit can be dropped or modified at reviewer's discretion.

## Pre review checklist

- [X] NOT RELEVANT Updated documentation
- [ ] STILL WORKING ON IT - Ensured new behaviour is tested

### DIFF
--- a/src/_ert_job_runner/client.py
+++ b/src/_ert_job_runner/client.py
@@ -14,7 +14,10 @@ from websockets.exceptions import (
     InvalidURI,
 )
 
+logging.basicConfig(level=logging.DEBUG)
+
 logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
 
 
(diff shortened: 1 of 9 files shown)


valentin-krasontovitsch:
closing in favor of merged #5597

PR #5568, opened 2023-06-13
Avoid trusting zero exit code from qstat (backport to version-5.0)

berland:
Experience shows that a zero exit code from qstat (or qstat_proxy) is sometimes zero also in case of failures. So to trust a zero exit code, we also require that the output file has length more than zero.

Failure codes from qstat is trusted as before.

This relies on:
 * ERT never asks the queue system for status unless it has actually submitted something
 * ERT only asks for status for jobs it has submitted, and when a job is finished, it is still reported by the queue system (accomplished by the -x default option).

This PR also adds a test verifying that the qstat_proxy would forward a nonzero exit code.


hnformentin (approved):
LGTM 🚀

Issue #5096, opened 2023-03-15
ERT crashed unexpectedly with "'Namespace' object has no attribute 'current_case'"


berland:
**Describe the bug**
Running es_mda on jenkins-hm-tutorial ends with the error message:
```
ERT crashed unexpectedly with "'Namespace' object has no attribute 'current_case'".
See logfile(s) for details:
   /private/havb/jenkins-hm-tutorial/reek_hm/r006/ert/model/logs/ert-log-2023-03-15T1316.txt

```
**To reproduce**
Steps to reproduce the behaviour:
1. ssh to a stavanger rgs-node
2. cd somewhere; tar xzvf /project/oompf/jenkins-ert-test-data/reek_hm_20221202_py3.tar.gz 
3. cd reek_hm/r006/ert/model
4. `source /prog/res/komodo/bleeding/enable`
5. `ert es_mda hm_analysis.ert`

Sourcing anything but bleeding works fine.

**Expected behaviour**
No crash

**Environment**
 - OS:RHEL7
 - ERT/Komodo release: bleeding (ert 5.x)
 - Python version: 3.8
 - Remote/HPC execution involved: no

**Additional context**
Extract from ert log:
```
2023-03-15 13:17:04,084 - root - MainThread - DEBUG - Resetting environment variable ERT_SITE_CONFIG
2023-03-15 13:17:04,084 - root - MainThread - DEBUG - Deleting temporary directory for site-config
2023-03-15 13:17:04,084 - ert.__main__ - MainThread - ERROR - ERT crashed unexpectedly with "'Namespace' object has no attribute 'current_case'"
Traceback (most recent call last):
  File "/prog/res/komodo/bleeding-py38-rhel7/root/lib64/python3.8/site-packages/ert/__main__.py", line 580, in main
    args.func(args)
  File "/prog/res/komodo/bleeding-py38-rhel7/root/lib64/python3.8/site-packages/ert/cli/main.py", line 124, in run_cli
    model = create_model(
  File "/prog/res/komodo/bleeding-py38-rhel7/root/lib64/python3.8/site-packages/ert/cli/model_factory.py", line 41, in create_model
    return _setup_multiple_data_assimilation(ert, storage, args, experiment_id)
  File "/prog/res/komodo/bleeding-py38-rhel7/root/lib64/python3.8/site-packages/ert/cli/model_factory.py", line 98, in _setup_multiple_data_assimilation
    "target_case": _target_case_name(ert, args, format_mode=True),
[...]

PR #5593, opened 2023-06-19
Fix ES-MDA CLI parser missing current_case

pinkwah:
Resolves: #5096 

The CLI is a mess and I am not sure whether this should even be used. This functionality means that not adding any options would default target case to `default_%d`.

See: https://github.com/equinor/ert/blob/d66e7d2f8a87b2a7bc77d8af9fe888a22a880b5f/src/ert/cli/model_factory.py#L173

Tested by going into `poly_example` and doing `ert es_mda poly.ert`, noting that no options were required and everything is happy.

### DIFF
--- a/src/ert/__main__.py
+++ b/src/ert/__main__.py
@@ -419,6 +419,13 @@ def get_ert_parser(parser: Optional[ArgumentParser] = None) -> ArgumentParser:
         ES_MDA_MODE, description=es_mda_description, help=es_mda_description
     )
     es_mda_parser.add_argument(
+        "--current-case",
+        type=valid_name,
+        default="default",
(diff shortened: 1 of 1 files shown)


pinkwah:
What do you mean? In GUI it defaults to `default_%d` also. The GUI user isn't required to set target case explicitly, and neither will the CLI user.

pinkwah:
In fact, I'd argue this change reverts it to how ERT 4 was. Consider that `current_case` has been changed, and prior to enkf_fs_refactor, it's a variable that always exists and therefore this original issue would not occur (it'd default to `default_%d`)

oyvindeide (approved):

PR #5597, opened 2023-06-19
Resolve timeout problems around running >= 200 realizations

### LINKED ISSUE #5273: Running ert with >= 200 realizations fails

valentin-krasontovitsch:
**Issue**
Resolves #5273

**Approach**
The failing symptoms occur in components communicating over websockets, in particular the ensemble evaluator as a server and the job queue, gui/cli, and job dispatch scripts as clients.

There are several improvements around the communication between those components.
The main (suspected) fix is backing of the so-called `JobQueueNode` busy business, by letting it sleep for a longer amount of time instead of one second between each querying of the queue system.
Rationale behind this is that the `JobQueueNode` keeps track of the state of a whole realization, with coarse resolution (not considering forward models). For long running realizations (>= 1h) it is quite unnecessary to query the state every second, as it does not change as long as the realization is running.
Given that each `JobQueueNode` runs in its own thread, @berland had the idea that this might starve components running in other threads for resources, which was somewhat confirmed by looking at timing logs (how long did some processing take) and their respective timestamps.

## Pre review checklist

- [X] NOT RELEVANT - Updated documentation

### DIFF
--- a/src/ert/_c_wrappers/job_queue/job_queue_node.py
+++ b/src/ert/_c_wrappers/job_queue/job_queue_node.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import random
 import time
 from threading import Lock, Semaphore, Thread
 from typing import TYPE_CHECKING, Optional
@@ -189,13 +190,30 @@ class JobQueueNode(BaseCClass):
 
         current_status = self.refresh_status(driver)
 
+        # in the following loop, we increase the sleep time between loop iterations as
+        # long running realizations do not change state often, and too frequent querying
+        # with many realizations starves other threads for resources.
+        initial_sleep_seconds = 1
+        time_sleep_seconds = initial_sleep_seconds
+        max_sleep_seconds = 30
+        time_until_longer_sleep_seconds = 30
+        use_random_sleep_offset = False
         while self.is_running(current_status):
             if (
                 self._start_time is None
                 and current_status == JobStatusType.JOB_QUEUE_RUNNING
             ):
                 self._start_time = time.time()
-            time.sleep(1)
+            if (
+                self._start_time is not None
+                and time_sleep_seconds != max_sleep_seconds
+                and time.time() - self._start_time > time_until_longer_sleep_seconds
+            ):
+                time_sleep_seconds = max_sleep_seconds
+                use_random_sleep_offset = True
+            time.sleep(
+                time_sleep_seconds + use_random_sleep_offset * random.randint(-5, 5)
+            )
             if self._max_runtime is not None and self.runtime >= self._max_runtime:
                 self._kill(driver)
                 # We sometimes end up in a state where we are not able to kill it,
--- a/src/ert/_c_wrappers/job_queue/queue.py
+++ b/src/ert/_c_wrappers/job_queue/queue.py
@@ -12,24 +12,14 @@ import threading
 import time
 from collections import deque
 from threading import Semaphore
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Callable,
-    Dict,
-    Iterable,
-    List,
-    Mapping,
-    Optional,
-    Union,
-)
+from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Union
 
 from cloudevents.conversion import to_json
 from cloudevents.http import CloudEvent
 from cwrap import BaseCClass
-from websockets.client import connect
+from websockets.client import WebSocketClientProtocol, connect
 from websockets.datastructures import Headers
-from websockets.exceptions import ConnectionClosedError
+from websockets.exceptions import ConnectionClosed
 
 import _ert_com_protocol
 from ert._c_wrappers import ResPrototype
@@ -231,14 +221,11 @@ class JobQueue(BaseCClass):
         self._free()
 
     def is_active(self) -> bool:
-        for job in self.job_list:
-            if job.thread_status in (
-                ThreadStatus.READY,
-                ThreadStatus.RUNNING,
-                ThreadStatus.STOPPING,
-            ):
-                return True
-        return False
+        return any(
+            job.thread_status
+            in (ThreadStatus.READY, ThreadStatus.RUNNING, ThreadStatus.STOPPING)
+            for job in self.job_list
+        )
 
     def fetch_next_waiting(self) -> Optional[JobQueueNode]:
         for job in self.job_list:
@@ -385,9 +372,7 @@ class JobQueue(BaseCClass):
     async def _publish_changes(
         ens_id: str,
         changes: Dict[int, str],
-        ws_uri: str,
-        ssl_context: Optional[Union[bool, ssl.SSLContext]],
-        headers: Mapping[str, str],
+        ee_connection: WebSocketClientProtocol,
     ) -> None:
         events = deque(
             [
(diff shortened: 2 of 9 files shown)


src/ert/_c_wrappers/job_queue/job_queue_node.py
```diff
@@ -170,13 +171,26 @@ def _job_monitor(self, driver, pool_sema, max_submit):
 
         current_status = self.refresh_status(driver)
 
+        initial_sleep_seconds = 1
+        time_sleep_seconds = initial_sleep_seconds
+        max_sleep_seconds = 30
```
xjules, line 176: Just wondering if we should decrease it to 15 secs.

src/ert/_c_wrappers/job_queue/job_queue_node.py
```diff
@@ -170,13 +171,26 @@ def _job_monitor(self, driver, pool_sema, max_submit):
...
-            time.sleep(1)
+            if (
+                self._start_time is not None
+                and time_sleep_seconds != max_sleep_seconds
+                and time.time() - self._start_time > 30
+            ):
+                time_sleep_seconds = max_sleep_seconds
+                use_random_sleep_offset = True
+            time.sleep(
+                time_sleep_seconds + use_random_sleep_offset * random.randint(-5, 5)
```
xjules, line 192: I like this step, which desyncronize threads!

src/ert/_c_wrappers/job_queue/queue.py
```diff
@@ -372,47 +374,67 @@ async def _queue_changes(
...
+        ws_uri: str,
+        ens_id: str,
+        pool_sema: threading.BoundedSemaphore,
+        evaluators: List[Callable[..., Any]],
+        ssl_context,
+        headers,
+    ) -> None:
+        reconnects = 0
+        max_reconnects = 10
+        async for websocket in connect(
```
xjules, line 400: The async itself contains reconnecting pattern, where to be honest to sure how to combine with the ert's one, ie. `reconnects` and `max_reconnects`.

src/ert/ensemble_evaluator/_builder/_legacy.py
```diff
@@ -161,6 +163,7 @@ def _evaluate(self) -> None:
         asyncio.set_event_loop(asyncio.new_event_loop())
 
         if self._config is None:
+            # TODO should event loop be closed?
```
xjules, line 166: I think it will close by itself.

src/ert/_c_wrappers/job_queue/job_queue_node.py
valentin-krasontovitsch: we can, no problem

src/ert/_c_wrappers/job_queue/job_queue_node.py
valentin-krasontovitsch: thank you : )

src/ert/_c_wrappers/job_queue/queue.py
valentin-krasontovitsch: could you link to the documentation for the contained reconnect pattern? i could not find it : (

src/ert/_c_wrappers/job_queue/job_queue_node.py
valentin-krasontovitsch: done.

src/ert/ensemble_evaluator/_builder/_legacy.py
```diff
@@ -141,13 +141,15 @@ def evaluate(self, config: "EvaluatorServerConfig") -> None:
         if not config:
             raise ValueError("no config for evaluator")
         self._config = config
+        logger.info("$$$ we are here. in evaluate.")
```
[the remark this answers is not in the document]
valentin-krasontovitsch: i would like to review the logging commit together, to see if we should drop the whole thing, or keep some of the logs, if that's OK

src/ert/_c_wrappers/job_queue/queue.py
xjules: It's in the code only https://github.com/python-websockets/websockets/blob/1bf9d1d766c80da4887240737266926f173fbcef/src/websockets/legacy/client.py#L594-L595

valentin-krasontovitsch: sadly there seems to be no way offered by the API to hook into that code, nor any other way apparent to me.

src/ert/ensemble_evaluator/_builder/_legacy.py
```diff
@@ -141,13 +141,15 @@ def evaluate(self, config: "EvaluatorServerConfig") -> None:
...
         get_event_loop().run_until_complete(
             wait_for_evaluator(
                 base_url=self._config.url,
                 token=self._config.token,
                 cert=self._config.cert,
             )
         )
+        logger.info("$$$ evaluate: done waiting for evaluator")
```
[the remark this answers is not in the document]
valentin-krasontovitsch: it's gone.

src/ert/ensemble_evaluator/_builder/_legacy.py
valentin-krasontovitsch: it's gone.

src/ert/ensemble_evaluator/dispatch.py
```diff
@@ -39,17 +40,21 @@ def _done_callback(self, *args):
 
     async def _work(self):
         t0 = time.time()
-        event_buffer, self._buffer = (
-            self._buffer[: self._max_batch],
-            self._buffer[self._max_batch :],
-        )
+        event_buffer = []
+        for _ in range(self._max_batch):
```
xjules, line 44: There is a slicing analogy / substitution for deque: https://stackoverflow.com/questions/7064289/use-slice-notation-with-collections-deque

valentin-krasontovitsch: tried to do a bit of benchmarking, not sure how reliable the data i produced was, but i tested the same thing with lists and dequeu, which is supposed to resemble what happens in ert, and the list approach was 2 times faster. so i'm dropping this commit altogether.

valentin-krasontovitsch: just for reference, i was a bit discouraged of trying to slice the deque using itertools by this quote that i found in [a forum](https://discuss.python.org/t/slicing-notation-with-deque-and-multi-dimensional-slicing-on-data-structures/2791/3):

> The deque type is a linked list and indexing past 0 and -1 is inefficient (read, takes O(N) time), so slicing those doesn’t make much sense.

it also raises the good (or at least practically, it should be solved) question of what's the return type of the call to islice (it's not a deque a priori), and the islice call doesn't pop elements, so we need to calls, and ... bla bla bla this was just to share some more thoughts around this idea

src/ert/_c_wrappers/job_queue/queue.py
```diff
@@ -449,30 +455,23 @@ async def execute_queue_via_websockets(  # pylint: disable=too-many-arguments
...
+            # initial publish
+            async with connect(
+                ws_uri,
+                ssl=ssl_context,
+                extra_headers=headers,
+                open_timeout=60,
+                ping_timeout=60,
+                ping_interval=60,
+                close_timeout=60,
+            ) as websocket:
```
xjules, line 467: wondering whether `websocket` is the best name here. What about being explicit `queue_connection` or something along those lines?

valentin-krasontovitsch: indeed - websocket is a bit generic, I guess - or could you make more specific in which sense `websocket` is not the best name?

I'm having a hard time coming up with something better though :sweat_smile: 

I feel in the suggested `queue_connection`, the `queue` part is a bit redundant, as it's used within the `JobQueue` class - leaving just `connection, which is an alternative to `websocket`, but perhaps just as generic or bad.

Maybe it could be named according to what it connects to, like `evaluator_connection` - does that make more sense? it gives a coupling to the name of the ensemble evaluator though. 

mind also that the connection information is passed in as a parameter called `ws_uri`, which in itself is also a very generic name. perhaps it would be more consistent to either leave the contested name to be `websocket`, or also change e.g. `ws_uri` to be more descriptive?

naming, eh? one of the hardest problems...

xjules: 1. What about `ee_connection` then? Since ee is used in our code base to refer to ensemble_evaluator.
2. ws_uri -> `ee_uri` ?

valentin-krasontovitsch: sure! let me fix it now

valentin-krasontovitsch: done!

kvashchuka:
I will dare to ask - should we add a test for this fix?

valentin-krasontovitsch:
we've discussed the testing. at the moment, we are in the unfortunate situation of being pressed for time.

furthermore, the initial problems we were encountering and trying to fix were difficult to encapsulate in a unit test.

now we have an easy way of reproducing, using an integration test (run ert with this config) based on the poly example.

from yesterday's standup / discussions i understood that running that breaking example might bring down the computer it's running on, but i'm willing to take that risk (on a github action runner) ; )

i suggest to look into it after this is merged and the komodo build is under way.

xjules (approved):
Wonderful job @valentin-krasontovitsch and co.! Let's get it in! :rocket:

PR #5626, opened 2023-06-23
Run integration tests with Eclipse in komodo

berland:
Ported from the Jenkins setup. The contents of res-testdata/Equinor is no longer in use and not included in the port.

**Issue**
Resolves missing running of integration tests with Eclipse in Komodo nightly testing.

**Approach**
Port Jenkins seteup.

## Pre review checklist

- [ ] Updated documentation


ci/testkomodo.sh
```diff
@@ -25,6 +25,7 @@ start_tests () {
     # Allow xvfb to find a new server
     pushd ${CI_TEST_ROOT}/tests
     xvfb-run -s "-screen 0 640x480x24" --auto-servernum python -m \
-    pytest -m "not requires_window_manager"
+        pytest -m "not requires_window_manager"
```
xjules, line 28: There is an extra indentation or ?

ci/testkomodo.sh
```diff
@@ -25,6 +25,7 @@ start_tests () {
...
+    pytest -m "requires_eclipse" --eclipse-simulator
```
berland, line 29: The  part `-m "requires_eclipse"` does the subset selection of 17 eclipse related tests.  These 17 are skipped by default, but the option `--eclipse-simulator` turns them on.

xjules: Right, now it makes sense :+1:

berland:
Before merge: Check if this requires /global/bin needs to be in PATH (which it is not by default for action runners)

berland:
> Before merge: Check if this requires /global/bin needs to be in PATH (which it is not by default for action runners)

Check done. Removing `/global../ from `$PATH` and then running pytest manually goes fine. ✔️

oysteoh (approved):

PR #5675, opened 2023-07-04
Change BaseRunModel to throw an exception with stacktrace

pinkwah:
Previously, if an exception occurred during ensemble evaluation, ERT would throw an exception but with no stack trace. Instead, we save the exception instead of just the message.

### DIFF
--- a/src/ert/run_models/base_run_model.py
+++ b/src/ert/run_models/base_run_model.py
@@ -7,7 +7,17 @@ import time
 import uuid
 from contextlib import contextmanager
 from pathlib import Path
-from typing import TYPE_CHECKING, Callable, Dict, Iterator, List, Optional, Union
+from typing import (
+    TYPE_CHECKING,
+    Callable,
+    Dict,
+    Generator,
+    List,
+    MutableSequence,
+    Optional,
+    Type,
+    Union,
+)
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


berland:
* Can you rewrite the git commit message to imperative form?
* Missing reelase note label

Will this look different for the user in the GUI? (have not tried to reproduce it)

src/ert/run_models/base_run_model.py
```diff
@@ -90,8 +101,9 @@ def __init__(
         self._job_start_time: int = 0
         self._job_stop_time: int = 0
         self._indeterminate: bool = False
-        self._fail_message: str = ""
-        self._failed: bool = False
+        self._failed = False
```
berland, line 104: why was typing removed?

src/ert/run_models/base_run_model.py
```diff
@@ -90,8 +101,9 @@ def __init__(
...
+        self._exception: Optional[Exception] = None
+        self._messages: MutableSequence[str] = []
```
berland, line 106: It looks like these `_messages` are always "error messages". If so, should we rename the variable accordingly?

src/ert/run_models/base_run_model.py
pinkwah: Not necessary. `False` has type `bool`.

xjules: `self._failed: bool = False` ?

pinkwah:
This does not change the behaviour of the GUI. Only the CLI checks whether there was an exception.

src/ert/run_models/base_run_model.py
berland: We would like `mypy` to err if you later do `self._failed = 0`. Will mypy err like this if you remove `: bool`?

berland (approved):
Happy reviewer. 
Only missing release not label.

PR #5679, opened 2023-07-05
Remove `subst_list.cpp`

pinkwah:


eivindjahren (approved):
:+1:

src/ert/_c_wrappers/util/substitution_list.py
```diff
@@ -1,141 +1,160 @@
...
-        c_ptr = self._alloc(None)
 
-        if c_ptr:
-            super().__init__(c_ptr)
-        else:
-            raise ValueError("Failed to construct subst_list instance.")
-
-    @classmethod
-    def from_dict(cls, config_dict) -> SubstitutionList:
+class SubstitutionList(_UserDict):
```
eivindjahren, line 26: We should eventually just have `SubstitutionList` be a `Dict[str, str]` and have functions that work on that class, but we don't need to do that now.

src/ert/_c_wrappers/util/substitution_list.py
```diff
@@ -1,141 +1,160 @@
...
-            return self._get_value(key)
-        else:
-            raise KeyError(f"No such key:{key}")
-
     def add_from_string(self, string):
-        _clib.subst_list.subst_list_add_from_string(self, string)
-
-    def get(self, key, default=None):
-        return self[key] if key in self else default
+        _add_from_string(self, string)
```
eivindjahren, line 52: You could just inline this function.

PR #5722, opened 2023-07-14
GUI: Handle SIGINT and exceptions in tests

pinkwah:


valentin-krasontovitsch (approved):
Love it! If this will affect users (i'm not sure how much they will notice during running) - will you add a release label and formulate the PR title into a for users understandable / meaningful message please?

pinkwah:
I don't think this will affect users in any way they care.

PR #5724, opened 2023-07-17
Raise error on faulty result file and missing report steps

### LINKED ISSUE #5712: assert config_node is not None

yngve-sk:
**Issue**
Resolves #5712 

**Approach**
Raise localized error upon faulty `RESULT_FILE` without `%d`, or missing `REPORT_STEPS` of `GEN_DATA` entry.

## Pre review checklist

- [x] Updated documentation

### DIFF
--- a/src/ert/_c_wrappers/enkf/ensemble_config.py
+++ b/src/ert/_c_wrappers/enkf/ensemble_config.py
@@ -247,7 +247,7 @@ class EnsembleConfig:
         report_steps = rangestring_to_list(options.get(ConfigKeys.REPORT_STEPS, ""))
 
         if os.path.isabs(res_file) or "%d" not in res_file:
-            result_file_context = next(
+            result_file_context: ContextValue = next(
                 x for x in gen_data if x.startswith("RESULT_FILE:")
             )
             raise ConfigValidationError.from_info(
--- a/src/ert/_c_wrappers/enkf/ensemble_config.py
+++ b/src/ert/_c_wrappers/enkf/ensemble_config.py
@@ -19,6 +19,7 @@ from ert.config.response_config import ResponseConfig
 from ert.config.summary_config import SummaryConfig
 from ert.config.surface_config import SurfaceConfig
 from ert.parsing import ConfigValidationError, ConfigWarning, ErrorInfo
+from ert.parsing.context_values import ContextList, ContextValue
 from ert.storage.field_utils.field_utils import Shape, get_shape
 from ert.validation import rangestring_to_list
 
@@ -233,7 +234,7 @@ class EnsembleConfig:
             self.addNode(self.get_field_node(field, grid_file, dims))
 
     @staticmethod
-    def gen_data_node(gen_data: List[str]) -> Optional[GenDataConfig]:
+    def gen_data_node(gen_data: ContextList[ContextValue]) -> Optional[GenDataConfig]:
         options = _option_dict(gen_data, 1)
         name = gen_data[0]
         res_file = options.get(ConfigKeys.RESULT_FILE)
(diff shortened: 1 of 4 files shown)


src/ert/parsing/lark_parser.py
```diff
@@ -362,7 +362,9 @@ def _handle_includes(
                         filename=config_file,
                     ).set_context(error_context)
                 )
-                continue
+
+                args = args[0:1]
+                # continue
```
hnformentin, line 367: why continue commented?

src/ert/_c_wrappers/enkf/ensemble_config.py
```diff
@@ -246,21 +247,29 @@ def gen_data_node(gen_data: List[str]) -> Optional[GenDataConfig]:
         report_steps = rangestring_to_list(options.get(ConfigKeys.REPORT_STEPS, ""))
 
         if os.path.isabs(res_file) or "%d" not in res_file:
-            logger.error(
-                f"The RESULT_FILE:{res_file} setting for {name} is invalid - "
-                "must have an embedded %d - and be a relative path"
+            result_file_context = next(
+                x for x in gen_data if x.startswith("RESULT_FILE:")
```
hnformentin, line 251: Would `error_message` be a more meaningful name?

src/ert/parsing/lark_parser.py
yngve-sk: Oops, should be removed instead. Omiting the `continue` will make it just ignore superfluous tokens, but still give an error about them.

src/ert/_c_wrappers/enkf/ensemble_config.py
yngve-sk: Hmm I'm not fully sure about this. This variable is a `ContextValue`, which has context information about the location of the word in the config file. I did type annotate the variable to make it a bit more clear

src/ert/_c_wrappers/enkf/ensemble_config.py
```diff
@@ -246,21 +247,29 @@ def gen_data_node(gen_data: List[str]) -> Optional[GenDataConfig]:
...
-        elif not report_steps:
-            logger.error(
-                "The GEN_DATA keywords must have a REPORT_STEPS:xxxx defined"
-                "Several report steps separated with ',' and ranges with '-'"
-                "can be listed"
+            raise ConfigValidationError.from_info(
+                ErrorInfo(
+                    filename=result_file_context.token.filename,
+                    message=f"The RESULT_FILE:{res_file} setting for {name} is "
+                    f"invalid - must have an embedded %d - and be a relative path",
```
valentin-krasontovitsch, line 257: just a litte suggestion regarding punctuation, you can skip this change if you want:
```suggestion
                    f"invalid - must have an embedded %d and be a relative path",
```

src/ert/_c_wrappers/enkf/ensemble_config.py
```diff
@@ -246,21 +247,29 @@ def gen_data_node(gen_data: List[str]) -> Optional[GenDataConfig]:
...
             )
-        else:
-            gdc = GenDataConfig(
-                name=name, input_file=res_file, report_steps=report_steps
+
+        if not report_steps:
+            raise ConfigValidationError.from_info(
+                ErrorInfo(
+                    filename=gen_data.keyword_token.filename,
+                    message="The GEN_DATA keywords must have a REPORT_STEPS:xxxx"
```
valentin-krasontovitsch, line 265: again just language and perhaps make the argument we expect more explicit using quotes:
```suggestion
                    message="The GEN_DATA keywords must have 'REPORT_STEPS:xxxx'"
```

src/ert/parsing/lark_parser.py
```diff
@@ -362,7 +362,8 @@ def _handle_includes(
                         filename=config_file,
                     ).set_context(error_context)
                 )
-                continue
+
+                args = args[0:1]
```
valentin-krasontovitsch, line 366: this seems a bit unexpected / out of place. what is happening here, and how is it related to the issue this PR fixes? or does this fall under camp-site rule?

tests/test_config_parsing/test_parser_error_collection.py
```diff
@@ -121,9 +121,17 @@ def assert_that_config_leads_to_error(
 
     with pytest.raises(ConfigValidationError) as caught_error:
         ErtConfig.from_file(config_filename)
+        # If the ert config did not raise any errors
+        # we manually raise an "empty" error to make
+        # this raise an assertion error that can be
+        # acted upon from assert_that_config_does_not_lead_to_error
+        raise ConfigValidationError(errors=[])
```
valentin-krasontovitsch, line 128: this seems hacky / funky upon first read :grimacing: in a function called `asser_that_config_leads_to_error`, you would expect that the config file contents you pass in always lead to an error, otherwise you would use a different function. so why the need to artificially raise an error?

seems like it might cause false positives when using that function with a config file that one thinks should lead to an error, but it don't :thinking:

EDIT: i now see the catch in line 132. i guess it protects exactly against what i'm describing here :sweat_smile:

valentin-krasontovitsch (approved):
looking good! just a few minor comments, and one question

src/ert/parsing/lark_parser.py
yngve-sk: On a second look, this change makes sense but is not directly related to this PR, just a very tiny thing I noticed doing this, so I guess it does fall under the camp-site rule.

If an include has multiple arguments, it will now keep only the first, and (1) lint that the rest are superfluous, and (2) potentially also lint errors pertaining to the actually included file. With the continue in place, it will only do (1) because the continue stops the execution.

valentin-krasontovitsch: put it into its own commit, write a test if you can, and then you will have earned a(nother) gold star from me : ) (and can merge it in this PR, i think)

PR #5733, opened 2023-07-20
Solve flake8-bugbears B017

### LINKED ISSUE #5284: Solve flake8-bugbears B017
```
$ pip install flake8-bugbear
$ flake8 | grep B017 | wc -l 
8
```

`flake8-bugbear` says:

> B017 `assertRaises(Exception)` and `pytest.raises(Exception)` should be considered evil. They can lead to your test passing even if the code being tested is never executed due to a typo. Assert for a more specific exception (builtin or custom), or use `assertRaisesRegex` (if using `assertRaises`), or add the `match` keyword argument (if using `pytest.raises`), or use the context manager form with a target.

which makes sense. We should fix these to avoid being exposed to the test itself failing.

jonathan-eq:
**Issue**
Resolves #5284

Flake8-bugbear error B017 states that assertRaises(Exception) and pytest.raises(Exception) should be considered evil. They can lead to your test passing even if the code being tested is never executed due to a typo. Assert for a more specific exception (builtin or custom), or use assertRaisesRegex (if using assertRaises), or add the match keyword argument (if using pytest.raises), or use the context manager form with a target.

**Approach**
Updated affected tests to specify which exception type should be raised, and the message (if there is one.)

## Pre review checklist

- [X] Updated documentation
- [X] Ensured new behaviour is tested

### DIFF
--- a/tests/unit_tests/services/test_base_service.py
+++ b/tests/unit_tests/services/test_base_service.py
@@ -143,7 +143,7 @@ sys.exit(1)
 """
 )
 def test_authtoken_fail(server):
-    with pytest.raises(Exception):
+    with pytest.raises(ServerBootFail):
         server.fetch_conn_info()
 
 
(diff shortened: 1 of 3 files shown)


pinkwah (approved):
:tada:

PR #5799, opened 2023-08-03
Do no create experiment and case on start-up

dafeda:
This fixes the bug where we end up with two default-cases when running a single experiment.

I propose we initially set the drop-down list of cases as inactive:

Screenshot 2023-08-07 at 09 19 37

The drop-down list will be set to active when gui updates. The drop-down list stays inactive while the simulation window is open though. This could perhaps be fixed.

Screenshot 2023-08-07 at 09 22 09

Once closed though, the list is updated:

Screenshot 2023-08-07 at 09 24 28

The `Initialize` button in "Initialize from scratch" is now deactivated if there are no cases:

Screenshot 2023-08-07 at 12 24 37

## Pre review checklist

- [ ] Updated documentation

### DIFF
--- a/src/ert/gui/ertwidgets/caseselector.py
+++ b/src/ert/gui/ertwidgets/caseselector.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Iterable, Optional
 
-from qtpy.QtCore import Qt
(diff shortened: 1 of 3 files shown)


oyvindeide (approved):
Behaves as I would expect, LGTM!

PR #5800, opened 2023-08-03
Make ert.storage use typing.Protocol

pinkwah:
This is a change in the direction of RemoteStorage.


pinkwah:
I had a think and I don't think using `typing.Protocol` is the way to go at this time. I would probably rather use implementation classes that are wrapped in abstracted classes.

PR #5828, opened 2023-08-07
Several mitigations against crashes when running with many realizations

### LINKED ISSUE #5273: Running ert with >= 200 realizations fails

valentin-krasontovitsch:
**Issue**
Resolves #5273 

**Approach**
- rewrite snapshot class in ensemble evaluator to be faster
- remove code that was tracking ensemble evaluator and crashing ert upon connection errors
- move batching dispatcher into own thread
- move forward model done callback (IO FS storage heavy) into own process

## Pre review checklist

- [X] IRRELEVANT - Updated documentation
- [X] Ensured new behaviour is tested: Johan Sverdrup case ran through with this once, should test with other users

### DIFF
--- a/src/ert/ensemble_evaluator/dispatch.py
+++ b/src/ert/ensemble_evaluator/dispatch.py
@@ -1,6 +1,6 @@
 import asyncio
-import contextlib
 import logging
+import threading
 import time
 import traceback
 from collections import OrderedDict, defaultdict
@@ -10,98 +10,101 @@ from ert.ensemble_evaluator import identifiers
 logger = logging.getLogger(__name__)
 
 
-class BatchingDispatcher:
-    def __init__(self, loop, timeout, max_batch=1000):
-        self._timeout = timeout
+class BatchingDispatcher:  # pylint: disable=too-many-instance-attributes
+    def __init__(self, sleep_between_batches_seconds, max_batch=1000):
+        self._sleep_between_batches_seconds = sleep_between_batches_seconds
         self._max_batch = max_batch
 
         self._LOOKUP_MAP = defaultdict(list)
         self._running = True
+        self._finished = False
         self._buffer = []
+        self._buffer_lock = threading.Lock()
 
-        # Schedule task
-        self._task = asyncio.ensure_future(self._job(), loop=loop)
-        self._task.add_done_callback(self._done_callback)
+        self._dispatcher_thread = threading.Thread(
+            name="ert_ee_batch_dispatcher",
+            target=self.run_dispatcher,
+        )
+        self._dispatcher_thread.start()
 
-    def _done_callback(self, *args):
-        try:
-            failure = self._task.exception()
-            if failure is not None:
-                logger.error(f"exception in batcher: {failure}")
-                trace_info = traceback.format_exception(
-                    type(failure), failure, failure.__traceback__
-                )
-                logger.error("".join(trace_info))
-            else:
-                logger.debug("batcher finished normally")
-                return
-        except asyncio.CancelledError as ex:
-            logger.warning(f"batcher was cancelled: {ex}")
-
-        # call any registered handlers for FAILED. since we don't have
-        # an event, pass empty list and let handler decide how to proceed
-        funcs = self._LOOKUP_MAP[identifiers.EVTYPE_ENSEMBLE_FAILED]
-        asyncio.gather(*[f([]) for f, _ in funcs])
-
-    async def _work(self):
+    def _work(self):
         if len(self._buffer) == 0:
             logger.debug("no events to be processed in queue")
             return
 
-        t0 = time.time()
-        batch_of_events_for_processing, self._buffer = (
-            self._buffer[: self._max_batch],
-            self._buffer[self._max_batch :],
-        )
-        left_in_queue = len(self._buffer)
-
-        function_to_events_map = OrderedDict()
-        for f, event in batch_of_events_for_processing:
-            if f not in function_to_events_map:
-                function_to_events_map[f] = []
-            function_to_events_map[f].append(event)
-
-        def done_logger(_):
-            logger.debug(
-                f"processed {len(batch_of_events_for_processing)} events in "
-                f"{(time.time()-t0):.6f}s. "
-                f"{left_in_queue} left in queue"
+        with self._buffer_lock:
+            t0 = time.time()
+            batch_of_events_for_processing, self._buffer = (
+                self._buffer[: self._max_batch],
+                self._buffer[self._max_batch :],
             )
+            left_in_queue = len(self._buffer)
+        function_to_events_map = OrderedDict()
+        for func, event in batch_of_events_for_processing:
+            if func not in function_to_events_map:
+                function_to_events_map[func] = []
+            function_to_events_map[func].append(event)
 
-        events_handling = asyncio.gather(
-            *[f(events) for f, events in function_to_events_map.items()]
+        for func, events in function_to_events_map.items():
+            func(events)
+
+        logger.debug(
+            f"processed {len(batch_of_events_for_processing)} events in "
+            f"{(time.time()-t0):.6f}s. "
+            f"{left_in_queue} left in queue"
         )
-        events_handling.add_done_callback(done_logger)
-        await events_handling
 
-    async def _job(self):
+    def run_dispatcher(self):
+        try:
+            self._job()
+        except Exception as failure:  # pylint: disable=broad-exception-caught
+            logger.exception(f"exception in batching dispatcher: {failure}")
+            trace_info = traceback.format_exception(
+                type(failure), failure, failure.__traceback__
+            )
+            logger.exception(f"{trace_info}")
+        else:
+            logger.debug("batcher finished normally")
+            return
+        finally:
+            self._finished = True
+
+        # call any registered handlers for FAILED. since we don't have
+        # an event, pass empty list and let handler decide how to proceed
+        funcs = self._LOOKUP_MAP[identifiers.EVTYPE_ENSEMBLE_FAILED]
+        for f in funcs:
+            f([])
+
+        logger.debug("Dispatcher thread exiting.")
+
+    def _job(self):
         while self._running:
-            await asyncio.sleep(self._timeout)
-            await self._work()
-
+            if len(self._buffer) < self._max_batch:
+                time.sleep(self._sleep_between_batches_seconds)
+            else:
+                # sleep(0) is used to play nice with other threads.
+                # It will release the GIL and let other threads run.
+                time.sleep(0)
+            self._work()
         # Make sure no events are lingering
-        await self._work()
+        self._work()
 
-    async def join(self):
+    async def wait_until_finished(self):
         self._running = False
-        # if result is exception it should have been handled by
-        # done-handler, but also avoid killing the caller here
-        with contextlib.suppress(BaseException):
-            await self._task
+        while not self._finished:
+            await asyncio.sleep(0.01)
 
-    def register_event_handler(self, event_types, function, batching=True):
+    def register_event_handler(self, event_types, function):
         if not isinstance(event_types, set):
             event_types = {event_types}
         for event_type in event_types:
-            self._LOOKUP_MAP[event_type].append((function, batching))
+            self._LOOKUP_MAP[event_type].append(function)
 
     async def handle_event(self, event):
-        for function, batching in self._LOOKUP_MAP[event["type"]]:
-            if batching:
-                if self._task.done():
-                    raise asyncio.InvalidStateError(
-                        "trying to handle event after batcher is done"
-                    )
+        if not self._running:
+            raise asyncio.InvalidStateError(
+                "trying to handle event after batcher is done"
+            )
+        for function in self._LOOKUP_MAP[event["type"]]:
+            with self._buffer_lock:
                 self._buffer.append((function, event))
-            else:
-                await function(event)
--- a/src/ert/ensemble_evaluator/dispatch.py
+++ b/src/ert/ensemble_evaluator/dispatch.py
@@ -16,7 +16,17 @@ class BatchingDispatcher:  # pylint: disable=too-many-instance-attributes
         self._sleep_between_batches_seconds = sleep_between_batches_seconds
         self._max_batch = max_batch
 
-        self._LOOKUP_MAP: Mapping[set, Callable] = defaultdict(lambda: (lambda _: None))
+        def log_unknown_event_type(events: List[CloudEvent]) -> None:
+            if len(events) > 0:
+                event_type = events[0]["type"]
+                logger.warning(
+                    "tried to lookup handle function for unknown "
+                    f"event type: {event_type}"
+                )
+
+        self._LOOKUP_MAP: Dict[str, Callable[[List[CloudEvent]], None]] = defaultdict(
+            lambda: log_unknown_event_type
+        )
         self._running = True
         self._finished = False
         self._buffer = []
--- a/src/ert/ensemble_evaluator/dispatch.py
+++ b/src/ert/ensemble_evaluator/dispatch.py
@@ -4,7 +4,9 @@ import threading
 import time
 import traceback
 from collections import OrderedDict, defaultdict
-from typing import Callable, Mapping
+from typing import Callable, Collection, Dict, List, Tuple
+
+from cloudevents.http import CloudEvent
 
 from ert.ensemble_evaluator import identifiers
 
@@ -12,7 +14,9 @@ logger = logging.getLogger(__name__)
 
 
 class BatchingDispatcher:  # pylint: disable=too-many-instance-attributes
-    def __init__(self, sleep_between_batches_seconds, max_batch=1000):
+    def __init__(
+        self, sleep_between_batches_seconds: int, max_batch: int = 1000
+    ) -> None:
         self._sleep_between_batches_seconds = sleep_between_batches_seconds
         self._max_batch = max_batch
 
@@ -29,7 +33,7 @@ class BatchingDispatcher:  # pylint: disable=too-many-instance-attributes
         )
         self._running = True
         self._finished = False
-        self._buffer = []
+        self._buffer: List[Tuple[Callable[[List[CloudEvent]], None], CloudEvent]] = []
         self._buffer_lock = threading.Lock()
 
         self._dispatcher_thread = threading.Thread(
@@ -38,7 +42,7 @@ class BatchingDispatcher:  # pylint: disable=too-many-instance-attributes
         )
         self._dispatcher_thread.start()
 
-    def _work(self):
+    def _work(self) -> None:
         if len(self._buffer) == 0:
             logger.debug("no events to be processed in queue")
             return
@@ -50,7 +54,9 @@ class BatchingDispatcher:  # pylint: disable=too-many-instance-attributes
                 self._buffer[self._max_batch :],
             )
             left_in_queue = len(self._buffer)
-        function_to_events_map = OrderedDict()
+        function_to_events_map: Dict[
+            Callable[[List[CloudEvent]], None], List[CloudEvent]
+        ] = OrderedDict()
         for func, event in batch_of_events_for_processing:
             if func not in function_to_events_map:
                 function_to_events_map[func] = []
@@ -65,7 +71,7 @@ class BatchingDispatcher:  # pylint: disable=too-many-instance-attributes
             f"{left_in_queue} left in queue"
         )
 
-    def run_dispatcher(self):
+    def run_dispatcher(self) -> None:
         try:
             self._job()
         except Exception as failure:  # pylint: disable=broad-exception-caught
@@ -87,7 +93,7 @@ class BatchingDispatcher:  # pylint: disable=too-many-instance-attributes
 
         logger.debug("Dispatcher thread exiting.")
 
-    def _job(self):
+    def _job(self) -> None:
         while self._running:
             if len(self._buffer) < self._max_batch:
                 time.sleep(self._sleep_between_batches_seconds)
@@ -99,16 +105,18 @@ class BatchingDispatcher:  # pylint: disable=too-many-instance-attributes
         # Make sure no events are lingering
         self._work()
 
-    async def wait_until_finished(self):
+    async def wait_until_finished(self) -> None:
         self._running = False
         while not self._finished:
             await asyncio.sleep(0.01)
 
-    def set_event_handler(self, event_types: set, function: Callable):
+    def set_event_handler(
+        self, event_types: Collection[str], function: Callable[[List[CloudEvent]], None]
+    ) -> None:
         for event_type in event_types:
             self._LOOKUP_MAP[event_type] = function
 
-    async def handle_event(self, event):
+    async def handle_event(self, event: CloudEvent) -> None:
         if not self._running:
             raise asyncio.InvalidStateError(
                 "trying to handle event after batcher is done"
--- a/src/ert/ensemble_evaluator/dispatch.py
+++ b/src/ert/ensemble_evaluator/dispatch.py
@@ -4,6 +4,7 @@ import threading
 import time
 import traceback
 from collections import OrderedDict, defaultdict
+from typing import Callable, Mapping
 
 from ert.ensemble_evaluator import identifiers
 
@@ -15,7 +16,7 @@ class BatchingDispatcher:  # pylint: disable=too-many-instance-attributes
         self._sleep_between_batches_seconds = sleep_between_batches_seconds
         self._max_batch = max_batch
 
-        self._LOOKUP_MAP = defaultdict(list)
+        self._LOOKUP_MAP: Mapping[set, Callable] = defaultdict(lambda: (lambda _: None))
         self._running = True
         self._finished = False
         self._buffer = []
@@ -71,9 +72,8 @@ class BatchingDispatcher:  # pylint: disable=too-many-instance-attributes
 
         # call any registered handlers for FAILED. since we don't have
         # an event, pass empty list and let handler decide how to proceed
-        funcs = self._LOOKUP_MAP[identifiers.EVTYPE_ENSEMBLE_FAILED]
-        for f in funcs:
-            f([])
+        failure_func = self._LOOKUP_MAP[identifiers.EVTYPE_ENSEMBLE_FAILED]
+        failure_func([])
 
         logger.debug("Dispatcher thread exiting.")
 
@@ -94,17 +94,15 @@ class BatchingDispatcher:  # pylint: disable=too-many-instance-attributes
         while not self._finished:
             await asyncio.sleep(0.01)
 
-    def register_event_handler(self, event_types, function):
-        if not isinstance(event_types, set):
-            event_types = {event_types}
+    def set_event_handler(self, event_types: set, function: Callable):
         for event_type in event_types:
-            self._LOOKUP_MAP[event_type].append(function)
+            self._LOOKUP_MAP[event_type] = function
 
     async def handle_event(self, event):
         if not self._running:
             raise asyncio.InvalidStateError(
                 "trying to handle event after batcher is done"
             )
-        for function in self._LOOKUP_MAP[event["type"]]:
-            with self._buffer_lock:
-                self._buffer.append((function, event))
+        function = self._LOOKUP_MAP[event["type"]]
+        with self._buffer_lock:
+            self._buffer.append((function, event))
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -2,7 +2,6 @@ import asyncio
 import logging
 import pickle
 import threading
-import time
 from contextlib import asynccontextmanager, contextmanager
 from http import HTTPStatus
 from typing import Optional, Set
@@ -10,7 +9,6 @@ from typing import Optional, Set
 import cloudevents.exceptions
 import cloudpickle
 import websockets
-from aiohttp import ClientError
 from cloudevents.conversion import to_json
 from cloudevents.http import CloudEvent, from_json
 from websockets.exceptions import ConnectionClosedError
@@ -33,7 +31,6 @@ from .identifiers import (
     EVTYPE_ENSEMBLE_STARTED,
     EVTYPE_ENSEMBLE_STOPPED,
 )
-from .monitor import Monitor
 from .snapshot import PartialSnapshot
 from .state import (
     ENSEMBLE_STATE_CANCELLED,
@@ -365,10 +362,9 @@ class EnsembleEvaluator:
         loop.run_until_complete(self.evaluator_server())
         logger.debug("Server thread exiting.")
 
-    def run(self) -> Monitor:
+    def _start_running(self) -> None:
         self._ws_thread.start()
         self._ensemble.evaluate(self._config)
-        return Monitor(self._config.get_connection_info())
 
     def _stop(self):
         if not self._done.done():
@@ -395,45 +391,8 @@ class EnsembleEvaluator:
             self._loop.call_soon_threadsafe(self._stop)
 
     def run_and_get_successful_realizations(self) -> int:
-        monitor = self.run()
-        unsuccessful_connection_attempts = 0
-        while True:
-            try:
-                for _ in monitor.track():
-                    unsuccessful_connection_attempts = 0
-                break
-            except ConnectionClosedError as e:
-                logger.debug(
-                    "Connection closed unexpectedly in "
-                    f"run_and_get_successful_realizations: {e}"
-                )
-            except (ConnectionRefusedError, ClientError) as e:
-                unsuccessful_connection_attempts += 1
-                logger.debug(
-                    f"run_and_get_successful_realizations caught {e}."
-                    f"{unsuccessful_connection_attempts} unsuccessful attempts"
-                )
-                if (
-                    unsuccessful_connection_attempts
-                    == _MAX_UNSUCCESSFUL_CONNECTION_ATTEMPTS
-                ):
-                    logger.debug("Max connection attempts reached")
-                    self._signal_cancel()
-                    break
-
-                sleep_time = 0.25 * 2**unsuccessful_connection_attempts
-                logger.debug(
-                    f"Sleeping for {sleep_time} seconds before attempting to reconnect"
-                )
-                time.sleep(sleep_time)
-            except BaseException:  # pylint: disable=broad-except
-                logger.exception("unexpected error: ")
-                # We really don't know what happened...  shut down and
-                # get out of here. Monitor is stopped by context-mgr
-                self._signal_cancel()
-                break
-
-        logger.debug("Waiting for evaluator shutdown")
+        self._start_running()
+        logger.debug("Started evaluator, joining until shutdown")
         self._ws_thread.join()
         logger.debug("Evaluator is done")
         return self._ensemble.get_successful_realizations()
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -71,12 +71,12 @@ class EnsembleEvaluator:
 
         for e_type, f in (
             (EVGROUP_FM_ALL, self._fm_handler),
-            (EVTYPE_ENSEMBLE_STARTED, self._started_handler),
-            (EVTYPE_ENSEMBLE_STOPPED, self._stopped_handler),
-            (EVTYPE_ENSEMBLE_CANCELLED, self._cancelled_handler),
-            (EVTYPE_ENSEMBLE_FAILED, self._failed_handler),
+            ({EVTYPE_ENSEMBLE_STARTED}, self._started_handler),
+            ({EVTYPE_ENSEMBLE_STOPPED}, self._stopped_handler),
+            ({EVTYPE_ENSEMBLE_CANCELLED}, self._cancelled_handler),
+            ({EVTYPE_ENSEMBLE_FAILED}, self._failed_handler),
         ):
-            self._dispatcher.register_event_handler(e_type, f)
+            self._dispatcher.set_event_handler(e_type, f)
 
         self._result = None
         self._ws_thread = threading.Thread(
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -4,13 +4,24 @@ import pickle
 import threading
 from contextlib import asynccontextmanager, contextmanager
 from http import HTTPStatus
-from typing import Optional, Set
+from typing import (
+    Any,
+    AsyncIterator,
+    Callable,
+    Dict,
+    Generator,
+    List,
+    Optional,
+    Set,
+    Tuple,
+)
 
 import cloudevents.exceptions
 import cloudpickle
 import websockets
 from cloudevents.conversion import to_json
 from cloudevents.http import CloudEvent, from_json
+from websockets.datastructures import Headers, HeadersLike
 from websockets.exceptions import ConnectionClosedError
 from websockets.legacy.server import WebSocketServerProtocol
 
@@ -59,7 +70,7 @@ class EnsembleEvaluator:
         self._done = self._loop.create_future()
 
         self._clients: Set[WebSocketServerProtocol] = set()
-        self._dispatchers_connected: Optional[asyncio.Queue] = None
+        self._dispatchers_connected: Optional[asyncio.Queue[None]] = None
         self._snapshot_mutex = threading.Lock()
         self._dispatcher = BatchingDispatcher(
             sleep_between_batches_seconds=2,
@@ -81,14 +92,14 @@ class EnsembleEvaluator:
         )
 
     @property
-    def config(self):
+    def config(self) -> EvaluatorServerConfig:
         return self._config
 
     @property
-    def ensemble(self):
+    def ensemble(self) -> Ensemble:
         return self._ensemble
 
-    def _fm_handler(self, events):
+    def _fm_handler(self, events: List[CloudEvent]) -> None:
         with self._snapshot_mutex:
             snapshot_update_event = self.ensemble.update_snapshot(events)
         send_future = asyncio.run_coroutine_threadsafe(
@@ -96,7 +107,7 @@ class EnsembleEvaluator:
         )
         send_future.result()
 
-    def _started_handler(self, events):
+    def _started_handler(self, events: List[CloudEvent]) -> None:
         if self.ensemble.status != ENSEMBLE_STATE_FAILED:
             with self._snapshot_mutex:
                 snapshot_update_event = self.ensemble.update_snapshot(events)
@@ -105,7 +116,7 @@ class EnsembleEvaluator:
             )
             send_future.result()
 
-    def _stopped_handler(self, events):
+    def _stopped_handler(self, events: List[CloudEvent]) -> None:
         if self.ensemble.status != ENSEMBLE_STATE_FAILED:
             self._result = events[0].data  # normal termination
             with self._snapshot_mutex:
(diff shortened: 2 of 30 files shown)


src/ert/ensemble_evaluator/dispatch.py
```diff
@@ -10,51 +10,37 @@
 logger = logging.getLogger(__name__)
 
 
-class BatchingDispatcher:
-    def __init__(self, loop, timeout, max_batch=1000):
+class BatchingDispatcher:  # pylint: disable=too-many-instance-attributes
+    def __init__(self, timeout, max_batch=1000):
```
xjules, line 14: I guess that the batch size remains on 1000 then?

valentin-krasontovitsch: yepp that's right, there doesn't seem to be a need to modify that parameter at this time

src/ert/ensemble_evaluator/dispatch.py
```diff
@@ -74,20 +60,47 @@ def done_logger(_):
         events_handling.add_done_callback(done_logger)
         await events_handling
 
+    def run_dispatcher(self, loop):
+        try:
+            loop.run_until_complete(self._job())
+        except asyncio.CancelledError as ex:
+            logger.warning(f"batcher was cancelled: {ex}")
+        except Exception as failure:  # pylint: disable=broad-exception-caught
+            logger.error(f"exception in batcher: {failure}")
```
xjules, line 69: I would perhaps rephrase it `Exception in run disbatcher event loop` to indicate where this was cauth?

valentin-krasontovitsch: personally, i think `event loop` is a technical detail that does not add much valuable information (we also get a stack trace, after all), so i would skip that part. we can change `batcher` to `batching dispatcher` though. and `logger.error` should be `logger.exception` I think :thinking:

valentin-krasontovitsch, line 69:
```suggestion
            logger.exception(f"exception in batching dispatcher: {failure}")
```

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -91,25 +92,41 @@ def ensemble(self):
...
+        send_future.result()
 
     async def _started_handler(self, events):
         if self.ensemble.status != ENSEMBLE_STATE_FAILED:
-            snapshot_update_event = self.ensemble.update_snapshot(events)
-            await self._send_snapshot_update(snapshot_update_event)
+            with self._snapshot_mutex:
+                snapshot_update_event = self.ensemble.update_snapshot(events)
+            send_future = asyncio.run_coroutine_threadsafe(
+                self._send_snapshot_update(snapshot_update_event), self._loop
```
valentin-krasontovitsch, line 104: now that we run send snapshot in the ensemble evaluator's asyncio loop, can we remove all asyncio stuff from batchy dispatchy @JHolba ?

src/ert/ensemble_evaluator/dispatch.py
```diff
@@ -74,20 +60,47 @@ def done_logger(_):
...
     async def join(self):
         self._running = False
         # if result is exception it should have been handled by
         # done-handler, but also avoid killing the caller here
-        with contextlib.suppress(BaseException):
-            await self._task
+        while not self._finished:
+            await asyncio.sleep(0.01)
 
     def register_event_handler(self, event_types, function, batching=True):
```
xjules, line 105: This seems to be always true (by default we always use batching). Maybe removing the parameter would simplify the code (removing the `if batching`) as I don't see that we can get rid of it any time soon.

src/ert/ensemble_evaluator/evaluator_tracker.py
```diff
@@ -144,7 +145,7 @@ def track(
                     indeterminate=self._model.isIndeterminate(),
                     progress=self._progress(),
                     iteration=iter_,
-                    snapshot=snapshot,
+                    snapshot=copy.deepcopy(snapshot),
```
xjules, line 148: Why are we copying the snapshot here?

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -104,110 +102,181 @@ def convert_iso8601_to_datetime(
...
-        self._snapshot = copy.copy(snapshot) if snapshot else None
+    def __init__(self, snapshot: Optional["Snapshot"] = None) -> None:
+        self._realization_states = defaultdict(dict)
+        """A shallow dictionary of realization states. The key is a string with
+        realization number, pointing to a dict with members active (bool),
+        start_time (datetime), end_time (datetime) and status (str)."""
 
-    def update_status(self, status: str) -> None:
-        self._apply_update(SnapshotDict(status=status))
+        self._step_states = defaultdict(dict)
```
xjules, line 119: Out of context a bit comment, but do we need steps? As far as I understand we only have a single step.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -104,110 +102,181 @@ def convert_iso8601_to_datetime(
...
     def update_step(
         self, real_id: str, step_id: str, step: "Step"
     ) -> "PartialSnapshot":
-        if self._snapshot is None:
-            raise UnsupportedOperationException(
-                f"cannot update step for {self.__class__} without providing a snapshot"
-            )
-        self._apply_update(
-            SnapshotDict(reals={real_id: RealizationSnapshot(steps={step_id: step})})
+        step_idx = (real_id, step_id)
```
xjules, line 163: Referring to my previews comment on necessity of steps :)

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -104,110 +102,181 @@ def convert_iso8601_to_datetime(
...
+            {
+                "status": real.status,
+                "active": real.active,
+                "start_time": real.start_time,
+                "end_time": real.end_time,
+            }
         )
-        self._data = _recursive_update(self._data, dictionary, check_key=False)
-        self._snapshot.merge(dictionary)
+        if real_id not in self._realization_states:
```
xjules, line 155: Since you use `defaultdict(dict)` you can remove this if statement.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -104,110 +102,181 @@ def convert_iso8601_to_datetime(
...
     ) -> "PartialSnapshot":
-        self._apply_update(
-            SnapshotDict(
-                reals={
-                    real_id: RealizationSnapshot(
-                        steps={step_id: Step(jobs={job_id: job})}
-                    )
-                }
+        job_idx = (real_id, step_id, job_id)
+        if job_idx not in self._job_states:
```
xjules, line 182: Since you use `defaultdict(dict)` you can remove this if statement. :top:

src/ert/ensemble_evaluator/dispatch.py
valentin-krasontovitsch: removed the logic and renamed `timeout` - see if you like it:
https://github.com/equinor/ert/pull/5828/commits/31ba8baff50d190b1037540a658184b018ac8836

src/ert/ensemble_evaluator/evaluator_tracker.py
valentin-krasontovitsch: that is a good question that we discussed the other day again as well. we try to explain a bit in the commit message:

    Make deep copy of full snapshot event
    
    This must be done as partial snapshots may modify the full snapshot held
    by the evaluator tracker.

basically a few lines above we have

```
    self._iter_snapshot[iter_] = snapshot
```

where we store the received snapshot as internal state in the evaluator tracker (a perhaps reasonable question at this point is why does the evaluator tracker need to store that state, too?) - when we yield it for further consumption, whatever grabs it can (and does) manipulate the snapshot (at least the metadata is manipulated by the gui).
that last point is also something that could / should be addressed in a refactor of the whole snapshot handling.

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: plan of refactor of this whole thing, so far planned for after fixes are merged - remove steps : )

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: yeah would be awesome to remove steps. i totally wanna do it, too.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -104,110 +102,181 @@ def convert_iso8601_to_datetime(
...
-                step.status == state.STEP_STATE_SUCCESS
-                and not self._snapshot.all_steps_finished(real_id)
-            ):
-                pass
-            else:
-                raise ValueError(
-                    f"unknown step status {step.status} for real: {real_id} step: "
-                    + f"{step_id}"
-                )
+        if step_idx not in self._step_states:
```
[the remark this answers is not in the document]
valentin-krasontovitsch: indeed! good catch, thank you!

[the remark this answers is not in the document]
valentin-krasontovitsch: fix is here: https://github.com/equinor/ert/pull/5828/commits/0c798435df846f964e3a30b5ae549c62e4218dc7

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: thank you, fixed:
https://github.com/equinor/ert/pull/5828/commits/f2d3bdbe134157b2135914591c926e5224027666

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: thank you, fixed:
https://github.com/equinor/ert/pull/5828/commits/f2d3bdbe134157b2135914591c926e5224027666

src/ert/ensemble_evaluator/snapshot.py
xjules: Very good! Then could you perhaps make an issue for step removal?

src/ert/ensemble_evaluator/evaluator_tracker.py
xjules: Thanks for the nice explanation! When you are writting the issue for refactoring you could include "to consider removal of snapshot from evaluator tracker".

src/ert/ensemble_evaluator/dispatch.py
xjules: Very good! :smile:

src/ert/ensemble_evaluator/dispatch.py
```diff
@@ -74,34 +60,59 @@ def done_logger(_):
...
 
     async def handle_event(self, event):
-        for function, batching in self._LOOKUP_MAP[event["type"]]:
-            if batching:
-                if self._task.done():
-                    raise asyncio.InvalidStateError(
-                        "trying to handle event after batcher is done"
-                    )
+        for function in self._LOOKUP_MAP[event["type"]]:
+            if not self._running:
```
xjules, line 113: I guess this can go above the `for loop`

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: here it is:

https://github.com/equinor/ert/issues/5866

src/ert/gui/model/snapshot.py
```diff
@@ -115,11 +114,12 @@ def prerender(
                 snapshot.data()[ids.REALS].keys(), key=int
             )
             metadata[SORTED_JOB_IDS] = {}
-            for real_id, real in snapshot.data()[ids.REALS].items():
+            for real_id in snapshot.reals.keys():
                 metadata[SORTED_JOB_IDS][real_id] = {}
-                for step_id, step in real[ids.STEPS].items():
+                for step_id in snapshot.steps(real_id).keys():
```
xjules, line 119: I know refactoring will tackle this, just it hurts looking at redundant for loops in python :laughing:

src/ert/gui/model/snapshot.py
```diff
@@ -295,13 +293,12 @@ def _add_snapshot(self, snapshot: Snapshot, iter_: int):
                 NodeType.REAL,
             )
             snapshot_tree.add_child(real_node)
-            for step_id, step in real[ids.STEPS].items():
-                step_node = Node(step_id, {ids.STATUS: step[ids.STATUS]}, NodeType.STEP)
+            for step_id, step in snapshot.steps(real_id).items():
```
xjules, line 296: When changing the way we access the metadata, why not to employ it further up in 
```python
real_id,
                {
                    ids.STATUS: real[ids.STATUS],
                    ids.ACTIVE: real[ids.ACTIVE],
                    REAL_JOB_STATUS_AGGREGATED: metadata[REAL_JOB_STATUS_AGGREGATED][
                        real_id
                    ],
                    REAL_STATUS_COLOR: metadata[REAL_STATUS_COLOR][real_id],
                },
```
and thus getting rid of `real = snapshot.data()[ids.REALS][real_id]` ?

src/ert/ensemble_evaluator/evaluator_tracker.py
valentin-krasontovitsch: included in [the issue](https://github.com/equinor/ert/issues/5866)

src/ert/gui/model/snapshot.py
valentin-krasontovitsch: yes, it is : )

src/ert/gui/model/snapshot.py
valentin-krasontovitsch: we did a whole refactor of the gui snapshot handling - we wanted to keep the changes minimal at first, to get a fix merged that can be used by users.

cf. #5850

src/ert/job_queue/job_queue_node.py
```diff
@@ -394,3 +473,22 @@ def wait_for(self) -> None:
 
     def _set_thread_status(self, new_status: ThreadStatus) -> None:
         self._thread_status = new_status
+
+
+class ProcessWithException(mp.Process):
```
xjules, line 478: It would be handy to have a small doc / comment for why we have this class in the first place.

src/ert/job_queue/job_queue_node.py
```diff
@@ -196,6 +219,62 @@ def run_timeout_callback(self) -> None:
         if self.callback_timeout:
             self.callback_timeout(*self.callback_arguments)
 
+    # this function only works on systems where multiprocessing.Process uses forking
+    def run_done_callback_forking(self) -> Tuple[LoadStatus, str]:
```
xjules, line 223: I like this part a lot! :tada: very nice!

src/ert/ensemble_evaluator/dispatch.py
valentin-krasontovitsch: fixed in 811e1af82fdc61eb75f7772aca114793dc997448

src/ert/job_queue/job_queue_node.py
valentin-krasontovitsch: all honor and glory goes to @JHolba, and his house

src/ert/job_queue/job_queue_node.py
valentin-krasontovitsch: I could not add more information then

- it's for doing stuff in a subprocess, and
- we add exception handling

so from that point of view i think the name does the trick

but maybe you would like to see some other information, that i'm missing / not thinking of? what would you like to include / see in such a doc or comment?

src/ert/ensemble_evaluator/dispatch.py
valentin-krasontovitsch: i also took the liberty, after looking at that code, of [simplifying further](https://github.com/equinor/ert/pull/5828/commits/81b0aa0ecfca33959d49b5257fb026945cd99405) - check it out and let me know what you think

src/ert/ensemble_evaluator/evaluator.py
valentin-krasontovitsch: Jon is working on it, resolving this comment

src/ert/job_queue/job_queue_node.py
xjules: What I meant is the by default we won't get any stacktrace from a Process, therefore `wait_and_throw_if_exception` needs to be called explicitly or something along those lines :)

tests/unit_tests/cli/test_integration_cli.py
```diff
@@ -798,3 +798,43 @@ def test_failing_job_cli_error_message():
             assert substring in f"{error}"
     else:
         pytest.fail(msg="Expected run cli to raise ErtCliError!")
+
+
+@pytest.mark.integration_test
+def test_cli_test_run_catches_forward_model_ok_callback_exception(
```
xjules, line 804: Nice!

valentin-krasontovitsch: thank you : )

src/ert/job_queue/job_queue_node.py
valentin-krasontovitsch: ah, indeed - the usage. good point, thank you!

added in 6b543a112aa94b79b275312fa0770ab58660651c

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -18,11 +12,114 @@
...
+
+            # second dispatcher informs that job 0 is running
+            send_dispatch_event(
+                dispatch2,
+                identifiers.EVTYPE_FM_JOB_RUNNING,
+                f"/ert/ensemble/{evaluator.ensemble.id_}/real/1/step/0/job/0",
+                "event1",
+                {"current_memory_usage": 1000},
+            )
+            send_dispatch_event(
```
xjules, line 62: I guess we miss here: # second dispatcher informs that job 1 is running

valentin-krasontovitsch, line 62:
```suggestion
            # second dispatcher informs that job 1 is running 
            send_dispatch_event(
```

valentin-krasontovitsch: thank you, fixed in 345d9a44e209df99ca069c722964144a747caf7c

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -18,11 +12,114 @@
...
+        # take down first monitor by leaving context
+
+    with Client(
+        url + "/dispatch",
+        cert=cert,
+        token=token,
+        max_retries=1,
+        timeout_multiplier=1,
+    ) as dispatch2:
+        # second dispatcher informs that job 0 is done
```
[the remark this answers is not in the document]
valentin-krasontovitsch: honestly i didn't care much about / read the comments, they're copypasta :grimacing: 
these are actually job runners, yes.
i suppose they were called dispatchers because they connect to the `/dispatch` endpoint of the server.

do you want me to rename them?

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -18,11 +12,114 @@
...
+            identifiers.EVTYPE_FM_JOB_FAILURE,
+            f"/ert/ensemble/{evaluator.ensemble.id_}/real/1/step/0/job/1",
+            "event_job_1_fail",
+            {identifiers.ERROR_MSG: "error"},
+        )
+
+    # we have to wait for the dispatcher to process the events, and for the internal
+    # ensemble state to get updated before connecting and getting a full ensemble
+    # snapshot
+    time.sleep(2)
```
xjules, line 106: Now I'm confused a bit. From the context above I understood that dispatcher is essentially job_runner.Client, which doesn't process anything. But from this one it seems that you refer to BatchingDispatcher. Maybe I don't see inside so well :smile:

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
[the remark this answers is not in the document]
xjules: More the comment bellow :arrow_down:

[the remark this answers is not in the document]
xjules: I guess it will be helpful to make the comments right for the next iteration of developers to find their bearings :)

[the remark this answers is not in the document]
valentin-krasontovitsch: comments should be fixed by eb1b445df37b665e44acdc16af14d010950a7cfc

renaming task added to refactor issue

src/ert/ensemble_evaluator/dispatch.py
```diff
@@ -1,107 +1,106 @@
...
-class BatchingDispatcher:
-    def __init__(self, loop, timeout, max_batch=1000):
-        self._timeout = timeout
+class BatchingDispatcher:  # pylint: disable=too-many-instance-attributes
+    def __init__(self, sleep_between_batches_seconds, max_batch=1000):
+        self._sleep_between_batches_seconds = sleep_between_batches_seconds
         self._max_batch = max_batch
 
-        self._LOOKUP_MAP = defaultdict(list)
+        self._LOOKUP_MAP: Mapping[set, Callable] = defaultdict(lambda: (lambda _: None))
```
xjules, line 19: As discussed, it would be nice to have a default function, which says "event not handled" instead of just an empty function. But I like the simplification :)

src/ert/gui/model/snapshot.py
```diff
@@ -295,11 +295,12 @@ def _add_snapshot(self, snapshot: Snapshot, iter_: int):
...
+                step_node = Node(step_id, {ids.STATUS: step.status}, NodeType.STEP)
                 real_node.add_child(step_node)
                 for job_id in metadata[SORTED_JOB_IDS][real_id][step_id]:
-                    job = step[ids.JOBS][job_id]
+                    job = snapshot.get_job(real_id, step_id, job_id)
+                    # job = step[ids.JOBS][job_id]
```
berland: Supposing this line should be taken out.

src/ert/ensemble_evaluator/dispatch.py
eivindjahren: Should probably be `Mapping[str, Callable]` also ert.ensemble_evaluator.identifiers should not be a package but a [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum) (see for instance `ConfigKeys` for how to set that up with python 3.8).

src/ert/ensemble_evaluator/dispatch.py
```diff
@@ -1,107 +1,106 @@
...
+        failure_func([])
 
+        logger.debug("Dispatcher thread exiting.")
+
+    def _job(self):
+        while self._running:
+            if len(self._buffer) < self._max_batch:
+                time.sleep(self._sleep_between_batches_seconds)
+            else:
+                time.sleep(0)
```
eivindjahren, line 85: `sleep(0)` needs some explanation.

src/ert/ensemble_evaluator/dispatch.py
xjules: > Also, this used to default to a list so that you could have multiple even handlers for each event which is a common assumption. So probably change this to Mapping[str, List[Callable[[], Any]]

This is sort-of the point of simplification that currently we do have only a single event handler function. Basically if we need to have multiple handler functions than we can adjust I guess.

eivindjahren: > > Also, this used to default to a list so that you could have multiple even handlers for each event which is a common assumption. So probably change this to Mapping[str, List[Callable[[], Any]]
> 
> This is sort-of the point of simplification that currently we do have only a single event handler function. Basically if we need to have multiple handler functions then we can adjust I guess.

Then you probably want to rename `add` to `set`

src/ert/job_queue/job_queue_node.py
```diff
@@ -196,6 +219,62 @@ def run_timeout_callback(self) -> None:
...
+        callback_arguments: CallbackArgs,
+        callback_status_shared: "Synchronized[c_int]",
+        status_msg_shared: "SynchronizedString",
+    ) -> None:
+        callback_status: Optional[LoadStatus]
+        status_msg: str
+        callback_status, status_msg = self.done_callback_function(*callback_arguments)
+
+        if callback_status is not None:
+            callback_status_shared.value = callback_status.value  # type: ignore
```
eivindjahren, line 275: This function is never called in tests. Is it still used?

src/ert/job_queue/job_queue_node.py
```diff
@@ -394,3 +473,26 @@ def wait_for(self) -> None:
...
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        mp.Process.__init__(self, *args, **kwargs)
+        self._parent_connection, self._child_connection = mp.Pipe(False)
+        self._exception = None
+
+    def run(self) -> None:
+        try:
+            mp.Process.run(self)
+            self._child_connection.send(None)
```
eivindjahren, line 491: `run` of `ProcessWithException` is not called in tests.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -104,110 +102,175 @@ def convert_iso8601_to_datetime(
...
-            raise UnsupportedOperationException(
-                f"trying to mutate {self.__class__} without providing a snapshot is "
-                + "not supported"
-            )
-        dictionary = update.dict(
-            exclude_unset=True, exclude_none=True, exclude_defaults=True
+        self._snapshot = snapshot
+
+    @property
+    def status(self) -> str:
```
eivindjahren, line 135: This property seems to not be used.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -104,110 +102,175 @@ def convert_iso8601_to_datetime(
...
+            self._realization_states[real_id].update(
+                {"status": _STEP_STATE_TO_REALIZATION_STATE[step_status]}
+            )
+        elif (
+            step_status == state.REALIZATION_STATE_FINISHED
+            and self._snapshot.all_steps_finished(real_id)
+        ):
+            real_state["status"] = state.REALIZATION_STATE_FINISHED
+        elif (
+            step_status == state.STEP_STATE_SUCCESS
```
eivindjahren, line 206: The fallthrough case of STEP_STATE_SUCESS is never hit in any test.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -104,110 +102,175 @@ def convert_iso8601_to_datetime(
...
+            _dict["metadata"] = self._metadata
+        if self._ensemble_state:
+            _dict["status"] = self._ensemble_state
+        if self._realization_states:
+            _dict["reals"] = self._realization_states
+
+        for step_index_tuple, step_state in self._step_states.items():
+            real_id = step_index_tuple[0]
+            step_id = step_index_tuple[1]
+            if "reals" not in _dict:
```
eivindjahren, line 231: In the tests, "reals" is always in _dict. Is this block redundant?

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -266,113 +333,126 @@ def from_cloudevent(self, event: CloudEvent) -> "PartialSnapshot":
...
+            job_idx[2]: Job(**job_data)
+            for job_idx, job_data in self._my_partial._job_states.items()
+            if job_idx[0] == real_id and job_idx[1] == step_id
+        }
 
     def get_real(self, real_id: str) -> "RealizationSnapshot":
-        if real_id not in self._data[ids.REALS]:
-            raise ValueError(f"No realization with id {real_id}")
-        return RealizationSnapshot(**self._data[ids.REALS][real_id])
+        return RealizationSnapshot(**self._my_partial._realization_states[real_id])
```
eivindjahren, line 414: This function is never called in the tests. Is it no longer used?

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -451,15 +533,46 @@ def add_job(  # pylint: disable=too-many-arguments
...
             name=name,
             stdout=stdout,
             stderr=stderr,
+            current_memory_usage=current_memory_usage,
+            max_memory_usage=max_memory_usage,
         )
         return self
 
     def add_metadata(self, key: str, value: Any) -> "SnapshotBuilder":
         self.metadata[key] = value
```
eivindjahren, line 547: add_metadata is not called in the tests. Is it no longer used?

src/ert/gui/model/snapshot.py
```diff
@@ -103,7 +103,6 @@ def prerender(
         # If there are no realizations, there's nothing to prerender.
         if not snapshot.data().get(ids.REALS):
             return None
```
eivindjahren, line 105: `snapshot.data().get(ids.REALS)` is always truthy in the tests. Is this case not needed?

src/ert/gui/model/snapshot.py
```diff
@@ -245,9 +245,7 @@ def _add_partial_snapshot(self, partial: PartialSnapshot, iter_: int):
 
                         for attr in (ids.CURRENT_MEMORY_USAGE, ids.MAX_MEMORY_USAGE):
                             if job.get(ids.DATA) and attr in job.get(ids.DATA):
-                                job_node.data[ids.DATA] = job_node.data[ids.DATA].set(
-                                    attr, job.get(ids.DATA).get(attr)
-                                )
+                                job_node.data[ids.DATA][attr] = job.get(attr)
```
eivindjahren, line 248: This line is never hit in the tests. Maybe add a test? Otherwise just remove the loop.

src/ert/ensemble_evaluator/dispatch.py
```diff
@@ -1,107 +1,106 @@
...
                 function_to_events_map[f] = []
             function_to_events_map[f].append(event)
 
-        def done_logger(_):
-            logger.debug(
-                f"processed {len(batch_of_events_for_processing)} events in "
-                f"{(time.time()-t0):.6f}s. "
-                f"{left_in_queue} left in queue"
-            )
+        for f, events in function_to_events_map.items():
```
berland, line 49: Can we spell out this `f` to something more descriptive, like `events_processor` ?  (pylint would complain about such short variable names)

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -104,110 +102,175 @@ def convert_iso8601_to_datetime(
     return parse(timestamp)
 
 
+DICT_SEP = "/"
```
berland, line 105: I think this was only used for the dict flattening code, which is now gone, thus this line can go away.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -104,110 +102,175 @@ def convert_iso8601_to_datetime(
...
-    def __init__(self, snapshot: "Snapshot") -> None:
-        """Create a PartialSnapshot. If no snapshot is provided, the object is
-        a immutable POD, and any attempt at mutating it will raise an
-        UnsupportedOperationException."""
-        self._data: TPMap[str, Any] = pyrsistent.m()
-        self._snapshot = copy.copy(snapshot) if snapshot else None
+    def __init__(self, snapshot: Optional["Snapshot"] = None) -> None:
+        self._realization_states = defaultdict(dict)
+        """A shallow dictionary of realization states. The key is a string with
+        realization number, pointing to a dict with members active (bool),
```
berland, line 116: "members" -> "keys"

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -104,110 +102,175 @@ def convert_iso8601_to_datetime(
...
+
+    def _check_state_after_step_update(
+        self, real_id: str, step_id: str
+    ) -> "PartialSnapshot":
+        step = self._step_states[(real_id, step_id)]
+        step_status = step["status"]
+
+        real_state = self._realization_states[real_id]
+        if (
+            "status" in real_state
```
berland, line 192: Can we use `real_state.get("status")` to replace line 192?

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -104,110 +102,175 @@ def convert_iso8601_to_datetime(
...
+
+        self._job_states[job_idx].update(job_update)
+        self._snapshot._my_partial._job_states[job_idx].update(job_update)
+        return self
+
+    def _check_state_after_step_update(
+        self, real_id: str, step_id: str
+    ) -> "PartialSnapshot":
+        step = self._step_states[(real_id, step_id)]
+        step_status = step["status"]
```
berland, line 188: Possibly safer to use `step.get("status")`

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -104,110 +102,175 @@ def convert_iso8601_to_datetime(
...
+            _dict["reals"][real_id]["steps"][step_id]["jobs"][job_id] = job_values_dict
+
+        return _dict
 
-    def data(self) -> TPMap[str, Any]:
-        return self._data
+    def data(self) -> Mapping[str, Any]:
+        return self.to_dict()
+
+    def _recursive_merge(self, other: "PartialSnapshot") -> "PartialSnapshot":
```
berland, line 259: Rename to `_merge`

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -104,110 +102,175 @@ def convert_iso8601_to_datetime(
...
+            self._realization_states[real_id].update(other_real_data)
+        for step_id, other_step_data in other._step_states.items():
+            self._step_states[step_id].update(other_step_data)
+        for job_id, other_job_data in other._job_states.items():
+            self._job_states[job_id].update(other_job_data)
+        return self
 
     # pylint: disable=too-many-branches
     def from_cloudevent(self, event: CloudEvent) -> "PartialSnapshot":
+        # pylint: disable=too-many-statements   # TODO: fix this before merge
```
berland, line 273: I am okay with not fixing this before merge.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -451,15 +533,46 @@ def add_job(  # pylint: disable=too-many-arguments
...
+            max_memory_usage=max_memory_usage,
         )
         return self
 
     def add_metadata(self, key: str, value: Any) -> "SnapshotBuilder":
         self.metadata[key] = value
         return self
+
+
+def _from_old_super_nested_dict(data: Mapping[str, Any]) -> PartialSnapshot:
```
berland, line 551: rename to `_from_nested_dict`

src/ert/ensemble_evaluator/dispatch.py
JHolba: sleep(0) is a way to play nice with other threads. you give control back to whatever scheduler is running and let it choose which thread to run next.

src/ert/ensemble_evaluator/dispatch.py
xjules: That sounds like a good suggestion @valentin-krasontovitsch

src/ert/ensemble_evaluator/dispatch.py
xjules: I guess this comes from the empirical experience to prevent the constant processing execution. I guess you can shed more light into it @valentin-krasontovitsch

src/ert/job_queue/job_queue_node.py
JHolba: It should be called in tests, but it may not be picked up by coverage due to being run in a different process?

src/ert/job_queue/job_queue_node.py
JHolba: It should be called in tests, but it may not be picked up by coverage due to being run in a different process?

src/ert/ensemble_evaluator/dispatch.py
JHolba: I added the if else so that we do not spend 2 seconds doing nothing if our queue is already full. This change could possibly have negative consequences as it will also cause more frequent updates in gui, but @valentin-krasontovitsch has another PR that makes the gui updates several orders of magnitude quicker.

The sleep(0) I added because AFAIK it is considered good practice to give the scheduler a possibility to switch to other running threads in between work.

src/ert/job_queue/job_queue_node.py
JHolba: done_callback_wrapper is the target of the ProcessWithException created in run_done_callback_forking

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: fixed

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: removed comment

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: fixed.

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: fixed

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: it's gone.

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: fixed

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: it's used in `_builder/_ensemble.py`

valentin-krasontovitsch: something is funky with the coverage calc - [here](https://app.codecov.io/gh/equinor/ert/pull/5828/blob/src/ert/ensemble_evaluator/_builder/_ensemble.py#L148), one usage of the method seems covered:

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: done.

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: step is scheduled to go out the door, so i'm gonna say this is a problem for future me

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: sometimes we just like to code defensively. especially when we don't really have a good understanding of the whole codebase's usage of the structures involved.

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: it is used - in the FUTURE :tm:

src/ert/ensemble_evaluator/dispatch.py
JHolba: Added a comment in source code about the sleep

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: it's gone.

src/ert/gui/model/snapshot.py
valentin-krasontovitsch: it's impossible to know, in the real world (without reading and understanding like LOTS of code)

src/ert/ensemble_evaluator/dispatch.py
JHolba: renamed f to func

src/ert/gui/model/snapshot.py
valentin-krasontovitsch: fixed in gui rewrite cf #5850

src/ert/ensemble_evaluator/dispatch.py
```diff
@@ -1,107 +1,106 @@
...
-        with contextlib.suppress(BaseException):
-            await self._task
-
-    def register_event_handler(self, event_types, function, batching=True):
-        if not isinstance(event_types, set):
-            event_types = {event_types}
+        while not self._finished:
+            await asyncio.sleep(0.01)
+
+    def register_event_handler(self, event_types: set, function: Callable):
```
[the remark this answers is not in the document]
valentin-krasontovitsch: thank you!

src/ert/ensemble_evaluator/dispatch.py
valentin-krasontovitsch: renamed method, default handler changed to logger

src/ert/gui/model/snapshot.py
valentin-krasontovitsch: yassss : D thanks for catching this!

it's already gone though.

src/ert/ensemble_evaluator/snapshot.py
valentin-krasontovitsch: something funky, maybe mocking only. we should stop mocking.

valentin-krasontovitsch:
We have run a git rebase with exec of a hand-picked selection of tests that have been found to conveniently run in parallel and should cover all the important stuff - invocation used is

```
env ERT_SHOW_BACKTRACE=yasss python3 -m pytest -n16 \
  --ignore tests/performance_tests \
  --ignore tests/unit_tests/data/test_integration_data.py \
  --ignore tests/unit_tests/gui/simulation/view/test_realization.py \
  --ignore tests/unit_tests/analysis/test_es_update.py \
  --ignore tests/unit_tests/test_libres_facade.py \
  --ignore tests/unit_tests/all/plugins/test_export_misfit.py \
  --ignore tests/unit_tests/c_wrappers/res/enkf/test_load_forward_model.py \
  --ignore tests/unit_tests/c_wrappers/res/job_queue/test_ert_qstat_proxy.py \
  -k "not test_many_concurrent_qstat_invocations" \
  --durations=10 --durations-min=60 \
  tests
```

this chooses ~ 1430 tests. these all passed for most commits, a few commits had one failing test which i attribute to flakyness - it was one of
```
- tests/unit_tests/gui/simulation/test_run_dialog.py::test_large_snapshot - \
  pytestqt.exceptions.TimeoutError: waitUntil timed out in 5000 milliseconds
- tests/unit_tests/services/test_storage_service.py::test_init_service_server_start_conf_info_is_stale - \
  AssertionError: Expected 'mock' to be called once. Called 0 times.
```

well to be honest the first test failing is because our changes in the first commit make the gui handling of snapshots slower, but the test doesn't fail for all commits (without us ever doing anything for the gui after the first commit), that's why i call it flaky. not sure about the second one, have just seen it before, and couldn't figure out what's going on.

i at least feel quite happy with the resulting commits (we're still inspecting the code by hand after the rebase).

xjules (approved):
Amazing job @valentin-krasontovitsch and @JHolba (and to all participants)! I think once the tests pass we get push the merge button :rocket:

PR #5839, opened 2023-08-08
Update performance benchmarks comparison baseline

jonathan-eq:
Updated baseline for comparing performance benchmarks with output result in artifact from github actions


jonathan-eq:
Verified locally

xjules (approved):
Looks fine to me, but I guess we need to confirm this new baseline with @EJahren if this agress our expectantions.

jonathan-eq:
Eivind confirmed 👍

PR #5842, opened 2023-08-08
Fix a few gui test waits / timeouts

valentin-krasontovitsch:
[Small fix in gui test](https://github.com/equinor/ert/pull/5842/commits/def80e5356a00f87e766a1e30a2f549441fd19fe) 

We have to wait a wee bit for the done button to become visible, it
seems.

[Fix too short wait in gui test](https://github.com/equinor/ert/pull/5842/commits/63ad38f60f516f522257da3f7682a099cb2a0f3a) 

Sometimes the wait for the single shot fire to handle a pop up dialog
seems to short, and we get a hanging test in the gui that eventually
gets cancelled by timeout in CI. We conservativally increase the wait
for the single shot to not introduce flakiness.

### DIFF
--- a/tests/unit_tests/gui/simulation/test_run_dialog.py
+++ b/tests/unit_tests/gui/simulation/test_run_dialog.py
@@ -3,6 +3,7 @@ from pathlib import Path
 from unittest.mock import Mock, patch
 
 import pytest
+from pytestqt.qtbot import QtBot
 from qtpy import QtWidgets
 from qtpy.QtCore import Qt, QTimer
 from qtpy.QtWidgets import QMessageBox, QToolButton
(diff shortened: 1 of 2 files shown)


tests/unit_tests/gui/simulation/test_run_dialog.py
```diff
@@ -35,12 +36,12 @@ def test_success(runmodel, qtbot, mock_tracker):
 
     with qtbot.waitExposed(widget, timeout=30000):
         qtbot.waitUntil(lambda: widget._total_progress_bar.value() == 100)
-        assert widget.done_button.isVisible()
+        qtbot.waitUntil(widget.done_button.isVisible, timeout=100)
```
eivindjahren, line 39: Are you sure this isn' flaky? Couldn't there be some time between the total_progress_bar goes to 100 and the done button is visible?

valentin-krasontovitsch: i think it's supposed to be instant - that's why there was no timeout.
but on inspecting the code a bit more carefully that is responsible for the done button, i don't really see in the immediate vicinity anything about progress being 100 - so it's at least not immediately clear.

do you suggest increasing the timeout i added?

PR #5861, opened 2023-08-10
Don't allow creating new case with existing name.

pinkwah:


dafeda (approved):
Tried it and it looks good.

I suggest the following as PR-title as I think it's a bit clearer:

"Don't allow creating new case with existing name."

PR #5865, opened 2023-08-11
Exit un-runnable job and fail job

### LINKED ISSUE #5684: job_dispatch should fail if not possible to execute job
How to reproduce: 
- Remove `#!/usr/bin/env python` from `poly_eval.py` in the poly-example
- Open ERT gui and launch `poly.ert` 

You will now see ERT "hanging" waiting for something to happen. In the terminal you will see an exception from the attempt of running `poly_eval.py`

```
(2023.05.02-py38) /datadisk/repos/ert/test-data/poly_example (main) $ ert gui poly.ert
  File "/prog/res/komodo/2023.05.02-py38-rhel7/root/bin/job_dispatch.py", line 8, in <module>
    sys.exit(main())
  File "/prog/res/komodo/2023.05.02-py38-rhel7/root/lib64/python3.8/site-packages/_ert_job_runner/job_dispatch.py", line 16, in main
    job_runner_main(sys.argv)
  File "/prog/res/komodo/2023.05.02-py38-rhel7/root/lib64/python3.8/site-packages/_ert_job_runner/cli.py", line 119, in main
    for job_status in job_runner.run(parsed_args.job):
  File "/prog/res/komodo/2023.05.02-py38-rhel7/root/lib64/python3.8/site-packages/_ert_job_runner/runner.py", line 57, in run
    for status_update in job.run():
  File "/prog/res/komodo/2023.05.02-py38-rhel7/root/lib64/python3.8/site-packages/_ert_job_runner/job.py", line 90, in run
    proc = Popen(
  File "/opt/rh/rh-python38/root/usr/lib64/python3.8/subprocess.py", line 858, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/opt/rh/rh-python38/root/usr/lib64/python3.8/subprocess.py", line 1704, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
OSError: [Errno 8] Exec format error: '/datadisk/repos/ert/test-data/poly_example/poly_eval.py'

```

Outcome:
- Job should be flagged as failed as it would if any other exception inside the job and ERT should complete. Ideally the message to the user should be somewhat discriptive if possible as the exception itself in the terminal is not giving much help to the user if they are not familiar with this.

DanSava:
**Issue**
Resolves #5684

## Pre review checklist

- [ ] Updated documentation

### DIFF
--- a/src/_ert_job_runner/job.py
+++ b/src/_ert_job_runner/job.py
@@ -81,17 +81,39 @@ class Job:
         environment = self.job_data.get("environment")
         if environment is not None:
             environment = {**os.environ, **environment}
-        proc = Popen(
-            arg_list,
-            stdin=stdin,
-            stdout=stdout,
-            stderr=stderr,
-            env=environment,
-        )
+
+        def ensure_file_handles_closed():
+            if stdin is not None:
+                stdin.close()
+            if stdout is not None:
+                stdout.close()
+            if stderr is not None:
+                stderr.close()
+
+        try:
+            proc = Popen(
+                arg_list,
+                stdin=stdin,
+                stdout=stdout,
+                stderr=stderr,
+                env=environment,
+            )
+            process = Process(proc.pid)
+        except OSError as e:
+            msg = f"{e.strerror} {e.filename}"
+            if e.strerror == "Exec format error" and e.errno == 8:
+                msg = (
+                    f"Missing execution format information in file: {e.filename!r}."
+                    f"Most likely you are missing and should add "
+                    f"'#!/usr/bin/env python' to the top of the file: "
+                )
+            stderr.write(msg)
+            ensure_file_handles_closed()
+            yield Exited(self, e.errno).with_error(msg)
+            return
 
         exit_code = None
 
-        process = Process(proc.pid)
         max_memory_usage = 0
         while exit_code is None:
             try:
@@ -158,14 +180,7 @@ class Job:
             if target_file_error:
                 yield exited_message.with_error(target_file_error)
                 return
-
-        if stdin is not None:
-            stdin.close()
-        if stdout is not None:
-            stdout.close()
-        if stderr is not None:
-            stderr.close()
-
+        ensure_file_handles_closed()
         yield exited_message
 
     def _assert_arg_list(self):
--- a/tests/unit_tests/c_wrappers/job_runner/test_job.py
+++ b/tests/unit_tests/c_wrappers/job_runner/test_job.py
@@ -1,4 +1,5 @@
 import os
+import stat
 from unittest.mock import PropertyMock, patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


eivindjahren:
Not really familiar with this, perhaps @xjules could have a look as well?

oysteoh:
Doing what i described in the issue now fail and that is good. I'm not sure though that the error message still makes sense at that the users are able to figure out what is wrong. In the case as described the shebang is removed. The error message which i'm getting is:

```
Experiment failed!
Realization: 0 failed after reaching max submit (2):
	job poly_eval failed with: 'Exec format error in /datadisk/repos/ert/test-data/poly_example/poly_eval.py'
	stderr file: '',
	its contents:	
	<Not written by:poly_eval>
``` 

Is there anything we can do to make it easier to understand what is wrong in this specific case?

DanSava:
> Is there anything we can do to make it easier to understand what is wrong in this specific case?

We can change the error message, but I was guessing that missing `#!/usr/bin/env python` is not the only reason that a job will not be able to start. So I thought that preserving the error message should be ok.

oysteoh:
> > Is there anything we can do to make it easier to understand what is wrong in this specific case?
> 
> We can change the error message, but I was guessing that missing `#!/usr/bin/env python` is not the only reason that a job will not be able to start. So I thought that preserving the error message should be ok.

Yup, i guess it is not - and the error message usually then makes sense ( just tried a bit with various failures ). But for this specific case it still is not easy to understand what is wrong. I also have a question - does it preserve the old behavior when there is an error in one job among many others which succeed? Now i got this popup at the end showing a error message, but how will this be if there is just one job failing and potentially among many others in many differen realisations?

DanSava:
> > > Is there anything we can do to make it easier to understand what is wrong in this specific case?
> > 
> > 
> > We can change the error message, but I was guessing that missing `#!/usr/bin/env python` is not the only reason that a job will not be able to start. So I thought that preserving the error message should be ok.
> 
> Yup, i guess it is not - and the error message usually then makes sense ( just tried a bit with various failures ). But for this specific case it still is not easy to understand what is wrong. I also have a question - does it preserve the old behavior when there is an error in one job among many others which succeed? Now i got this popup at the end showing a error message, but how will this be if there is just one job failing and potentially among many others in many differen realisations?

That is a good point will test for that.

DanSava:
> I also have a question - does it preserve the old behavior when there is an error in one job among many others which succeed? Now i got this popup at the end showing a error message, but how will this be if there is just one job failing and potentially among many others in many differen realisations?

Tested and it behaves like any other job that fails among other jobs that succeed.

src/_ert_job_runner/job.py
```diff
@@ -81,17 +81,38 @@ def run(self):
...
+                stdout=stdout,
+                stderr=stderr,
+                env=environment,
+            )
+            process = Process(proc.pid)
+        except OSError as e:
+            msg = f"{e.strerror} {e.filename}"
+            if e.strerror == "Exec format error" and e.errno == 8:
+                msg = (
+                    f"Missing execution format information "
```
oysteoh, line 106: Very good detecting this very issue. A bit hard to read the message - maybe "clean it up"  into something like i.e 

```
f"Missing execution format information in file: '{e.filename}'."
f"Most likely you are missing and should add following to the top of the file: '#!/usr/bin/env python'"
```

oysteoh (approved):

PR #5871, opened 2023-08-11
Dont run test summary if no input file is missing

valentin-krasontovitsch:
## Pre review checklist

- [X] IRRELEVANT Updated documentation
  * gui test times out, is cancelled (gui tests "hang"), no junit file created AND we dont attempt to use the file: https://github.com/equinor/ert/actions/runs/5854251234/job/15869878650?pr=5871

### DIFF
--- a/.github/workflows/test_ert.yml
+++ b/.github/workflows/test_ert.yml
@@ -48,6 +48,7 @@ jobs:
       run: |
         pytest tests --junit-xml=junit.xml -sv --mpl \
           -m "requires_window_manager" --benchmark-disable
+          sleep 900000
 
     - name: Unit Test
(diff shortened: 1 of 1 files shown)


valentin-krasontovitsch:
so i thought just putting as a condition for running the summary thing "not cancelled" would work, because in case of a timeout the github action logs say "operation was cancelled", but that does not make the `cancelled()` [github workflow thingy](https://docs.github.com/en/actions/learn-github-actions/expressions#cancelled) true : /
so probably one needs to somehow register if there is a file, and then use that in the condition for the summary job, but it will get too convoluted and i'm running out of time, so i'm closing this.

PR #5920, opened 2023-08-21
Remove unused lsf submit internal

eivindjahren:
## Pre review checklist

- [ ] Updated documentation


src/clib/lib/job_queue/lsf_driver.cpp
```diff
@@ -195,78 +184,22 @@ int lsf_job_parse_bsub_stdout(const char *bsub_cmd, const char *stdout_file) {
 }
 
 static void lsf_driver_internal_error(const lsf_driver_type *driver) {
```
berland, line 186: Can't this function also be purged?

andreas-el: Is it not used in function `lsf_driver_assert_submit_method`?

eivindjahren: Technically still used for lsf_driver_assert_submit_method, and I thought it would be a good idea to have a look at the use of that function separately.

berland (approved):
✂️

PR #5921, opened 2023-08-21
Mark flaky test with xfail

valentin-krasontovitsch:
We have to investigate and figure out why this test is flaky.

In the meantime, to not disturb development, we mark it with xfail.


eivindjahren (approved):
5/7 almost perfect

PR #5960, opened 2023-08-25
Let ERT handle a flaky qdel as for qsub and qstat

### LINKED ISSUE #5121: Support MAX_RUNTIME in Torque driver

berland:
Use the same exponential retry schema as for qsub
and qstat, and conserve stdout/stderr for future debugging

**Issue**
Resolves #5121 

## Pre review checklist

- [x] Updated documentation

### DIFF
--- a/docs/reference/configuration/keywords.rst
+++ b/docs/reference/configuration/keywords.rst
@@ -639,7 +639,7 @@ Keywords controlling the simulations
 
         ::
 
-                -- Let each realizations run for 50 seconds
+                -- Let each realization run for a maximum of 50 seconds
                 MAX_RUNTIME 50
 
         The MAX_RUNTIME key is optional.
@@ -1869,7 +1869,8 @@ bjobs.
         ERT will do exponential sleeps, starting at 2 seconds, and the provided
         timeout is a maximum. Let the timeout be sums of series like 2+4+8+16+32+64
         in order to be explicit about the number of retries. Set to zero to disallow
-        flakyness.
+        flakyness, setting it to 2 will allow for one re-attempt, and 6 will give two
+        re-attempts.
 
         ::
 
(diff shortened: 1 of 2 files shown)


docs/reference/configuration/keywords.rst
```diff
@@ -1869,7 +1869,8 @@ bjobs.
         ERT will do exponential sleeps, starting at 2 seconds, and the provided
         timeout is a maximum. Let the timeout be sums of series like 2+4+8+16+32+64
         in order to be explicit about the number of retries. Set to zero to disallow
-        flakyness.
+        flakyness, setting it to 2 will allow for one re-attempt, and 6 will give two
```
xjules, line 1872: This is quite cumbersome to understand to be honest. If instead they could specify `QUEUE_QUERY_MAX_ATTEMPTS` which internally would transfer to TIMEOUT; that would be much nice I think :)

src/clib/lib/job_queue/torque_driver.cpp
```diff
@@ -845,10 +845,57 @@ job_status_type torque_driver_get_job_status(void *__driver, void *__job) {
...
+                                      (const char **)&job->torque_jobnr_char,
+                                      tmp_std_file, tmp_err_file);
+        if (return_value != 0) {
+            if (slept_time + retry_interval <= driver->timeout) {
+                torque_debug(driver,
+                             "qdel failed for job %s with exit code "
+                             "%d, retrying in %d seconds",
+                             job->torque_jobnr_char, return_value,
+                             retry_interval);
+                sleep(retry_interval);
```
xjules, line 875: I see that on Torque we have some `usleep(rand() % 2000000); // max 2 seconds` in addition - why is that?; ie. why not here?

berland: Comment in the code: "Sleep some more at random, to avoid synchronized retries from all threads". This is relevant for qstat which is ddos'ing the queue server. qdel should not need this drastic measure.

xjules (approved):
Well done! :rocket:

PR #5982, opened 2023-08-28
Add type hinting to `ert.shared.plugins`

pinkwah:
This module used the `decorator` package for a single function. We can re-write the decorator to be a standard Python decorator with type-hinting, and remove a dependency on this `decorator` package.


eivindjahren (approved):
:+1:

PR #5993, opened 2023-08-29
Add warning when overwriting QUEUE_OPTION

### LINKED ISSUE #5715: LSF_SERVER config linting
Have had a user-support case where the user got an exception like this and was not able to continue

```
ssh: Could not resolve hostname $ert_lsf_server: Name or service not known
Submitting: /usr/bin/ssh $ERT_LSF_SERVER bsub -o /private/ooh/test/poly_example/poly_out/realization-1/iter-0/poly_1.LSF-stdout -q mr -J poly_1 -n 1 -R "select[cs && x86_64Linux]" -P user:ooh /prog/res/komodo/2023.06.01-py38-rhel7/root/lib/python3.8/site-packages/equilibrium/bin/komodo_job_dispatch /private/ooh/test/poly_example/poly_out/realization-1/iter-0 
ssh: Could not resolve hostname $ert_lsf_server: Name or service not known
```

Reason for this is that she had 
```
QUEUE_OPTION LSF LSF_SERVER $ERT_LSF_SERVER
```
in her config-file. This is not the way we do it anymore and this is configured through ert-configurations. What happens here is that LSF_SERVER is set to $ERT_LSF_SERVER ( not resolved since the variable does not exist ) and overwrites what we have in ert-configurations. 

Not sure if this is a big problem - but in general - should we warn users about overwrites or in some way mitigate at least things like this to happen as it is not very user-friendly!

Update 31-08-23:
This would be useful for inspecting all QUEUE_OPTION provided. If any (for a given queue system) have duplicates, we know that they will be overwritten.

larsevj:
**Issue**
Resolves #5715 

## Pre review checklist

- [x] Updated documentation

### DIFF
--- a/src/ert/config/queue_config.py
+++ b/src/ert/config/queue_config.py
@@ -9,8 +9,7 @@ from typing import Any, Dict, List, Tuple, Union, no_type_check
 
 from ert import _clib
 
-from .parsing import (ConfigDict, ConfigValidationError, ConfigWarning,
-                      ErrorInfo)
+from .parsing import ConfigDict, ConfigValidationError, ConfigWarning, ErrorInfo
 from .queue_system import QueueSystem
 
 GENERIC_QUEUE_OPTIONS: List[str] = ["MAX_RUNNING"]
@@ -96,7 +95,7 @@ class QueueConfig:
             else:
                 queue_options[queue_system].append(option_name)
 
-        if queue_system == QueueSystem.LSF and queue_options[QueueSystem.LSF]:
+        if selected_queue_system == QueueSystem.LSF and queue_options[QueueSystem.LSF]:
             _validate_lsf_options(queue_options[QueueSystem.LSF])
 
         if (
--- a/src/ert/config/queue_config.py
+++ b/src/ert/config/queue_config.py
@@ -3,13 +3,14 @@ from __future__ import annotations
 import re
 import shutil
 import warnings
-from collections import defaultdict
+from collections import Counter, defaultdict
 from dataclasses import dataclass, field
 from typing import Any, Dict, List, Tuple, Union, no_type_check
 
 from ert import _clib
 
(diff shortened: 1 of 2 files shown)


src/ert/config/queue_config.py
```diff
@@ -111,6 +115,28 @@ def create_local_copy(self) -> QueueConfig:
         )
 
 
+def _validate_lsf_options(lsf_options: List[Tuple[str, str]]) -> None:
+    # Checks currently only for overwriting of 'QUEUE_OPTIONS LSF LSF_SERVER'
+    keywords_to_check = ["LSF_SERVER"]
+    for keyword in keywords_to_check:
```
andreas-el, line 121: There are very many different `QUEUE_OPTION` for the different `QUEUE_SYSTEM` available.
Would it be an idea to test if there are _ANY_ duplicates here? The end result would be the same regardless.
The function could also cover other systems, i.e. TORQUE, LOCAL, SLURM etc.

I guess we can omit `MAX_RUNNING` as we already know this might show up multiple times.

tests/unit_tests/config/test_queue_config.py
```diff
@@ -127,3 +133,18 @@ def test_torque_queue_config_invalid_memory_pr_job(memory_with_unit_str):
         f.write(f"QUEUE_OPTION TORQUE MEMORY_PER_JOB {memory_with_unit_str}")
     with pytest.raises(ConfigValidationError):
         ErtConfig.from_file(filename)
+
+
+@pytest.mark.usefixtures("use_tmpdir")
+def test_queue_option_LSF_SERVER_set_by_user_warning():
```
andreas-el, line 139: The site wide fixture does not contain `QUEUE_OPTION`, but you could add a fixture instead that has some in `conftest.py`? The test is a lot better if we actually use site-config & user-config to test this.

PR #6037, opened 2023-09-04
Print overwritten Queue Options values to log files

### LINKED ISSUE #5715: LSF_SERVER config linting

larsevj:
**Issue**
Resolves #5715 

**Approach**

## Pre review checklist

- [ ] Wait for tests in CI to pass (see "checks" below)
- [ ] Updated documentation
- [ ] Ensured new behaviour is tested (opened GUI? screenshots? tried workflows?)

## Pre merge checklist
- [x] Commit history is consistent and clean, in line with the [contribution
  guidelines](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md).

### DIFF
--- a/src/ert/config/queue_config.py
+++ b/src/ert/config/queue_config.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 import re
 import shutil
 from collections import defaultdict
@@ -100,6 +101,15 @@ class QueueConfig:
         ):
             _validate_torque_options(queue_options[QueueSystem.TORQUE])
 
+        if (
+            selected_queue_system != QueueSystem.LOCAL
+            and queue_options[selected_queue_system]
+        ):
+            _check_for_overwritten_queue_system_options(
+                selected_queue_system,
+                queue_options[selected_queue_system],
+            )
+
         return QueueConfig(job_script, max_submit, selected_queue_system, queue_options)
 
     def create_local_copy(self) -> QueueConfig:
@@ -109,6 +119,27 @@ class QueueConfig:
             QueueSystem.LOCAL,  # type: ignore
             self.queue_options,
         )
+
+
+def _check_for_overwritten_queue_system_options(
+    selected_queue_system: QueueSystem,
+    queue_system_options: List[Union[Tuple[str, str], str]],
+) -> None:
+    def generate_dict(
+        option_list: List[Union[Tuple[str, str], str]]
+    ) -> Dict[str, List[str]]:
+        temp_dict: Dict[str, List[str]] = defaultdict(list)
+        for option_string in option_list:
+            if isinstance(option_string, tuple) and (option_string[0] != "MAX_RUNNING"):
+                temp_dict.setdefault(option_string[0], []).append(option_string[1])
+        return temp_dict
+
+    for option_name, option_values in generate_dict(queue_system_options).items():
+        if len(option_values) > 1:
+            logging.info(
+                f"Overwriting QUEUE_OPTION {selected_queue_system} {option_name}:"
+                f" \n Old value: {option_values[0]} \n New value: {option_values[-1]}"
+            )
 
 
 def _validate_torque_options(torque_options: List[Tuple[str, str]]) -> None:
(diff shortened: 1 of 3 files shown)


src/ert/config/queue_config.py
```diff
@@ -111,6 +120,23 @@ def create_local_copy(self) -> QueueConfig:
         )
 
 
+def _check_for_overwritten_queue_system_options(
+    selected_queue_system: QueueSystem, queue_system_options: List[Tuple[str, str]]
+) -> None:
+    temp_dict: Dict[str, List[str]] = defaultdict(list)
+    for option_string in queue_system_options:
```
sregales-TNO, line 127: why are we checking if instance is tuple if we are expecting `List[Tuple[str, str]]`
If the type hint is not what we expect, then we either explicitly hint unstructured data type or data needs to sturctured before referencing this function.
then the function can be simple
```python
for option_name, option_value in queue_system_options:
    temp_dict.setdefault(option_name, []).append(option_value)
```

src/ert/config/queue_config.py
```diff
@@ -111,6 +120,23 @@ def create_local_copy(self) -> QueueConfig:
...
+        if isinstance(option_string, tuple):
+            option_name = option_string[0]
+            option_value = option_string[1]
+            temp_dict.setdefault(option_name, []).append(option_value)
+    for option_name, option_values in temp_dict.items():
```
sregales-TNO, line 132: if we split the function in two, the first and second for loop. then we can make the first into a generator. because the actual behaviour of this function is done in the second part. the first seems to just "structurize" the data

src/ert/config/queue_config.py
larsevj: I had to modify and complicate this part because I discovered that the option_value might be left unspecified in the config file in which case there is only one value to unpack.

src/ert/config/queue_config.py
```diff
@@ -100,6 +101,14 @@ def from_dict(cls, config_dict: ConfigDict) -> QueueConfig:
         ):
             _validate_torque_options(queue_options[QueueSystem.TORQUE])
 
+        if (
+            selected_queue_system != QueueSystem.LOCAL
+            and queue_options[selected_queue_system]
+        ):
+            _check_for_overwritten_queue_system_options(
+                selected_queue_system, queue_options[selected_queue_system]
```
sregales-TNO, line 109: > I had to modify and complicate this part because I discovered that the option_value might be left unspecified in the config file in which case there is only one value to unpack.
If that is the case, try passing a filter or filter generator to the function.
example
```python
(options_string for opitons_string in queue_options[selected_queue_system] if isinstance(option_string, tuple))
```
This allows you to loop once through the data set instead of twice.
This is a matter of opinion but it seems expensive doing two consecutive loops on the same data set for a log.

sregales-TNO (approved):
LGTM

Issue #6045, opened 2023-09-05
flaky test: test_ert_qstat_proxy


eivindjahren:
```
FAILED tests/unit_tests/job_queue/test_ert_qstat_proxy.py::test_many_concurrent_qstat_invocations
                    assert state == CacheState.CACHE_EXISTS
>                   assert (
                        process.returncode == 0
                    ), "Check for race condition if AssertionError"
E                   AssertionError: Check for race condition if AssertionError
E                   assert 1 == 0
E                    +  where 1 = <Popen: returncode: 1 args: [PosixPath('/opt/hostedtoolcache/Python/3.10.12/...>.returncode
```

berland:
The error has been reproduced in https://github.com/equinor/ert/actions/runs/6085593995/job/16510174916?pr=6046

```
tests/unit_tests/job_queue/test_ert_qstat_proxy.py::test_many_concurrent_qstat_invocations[59-1000] 011111111111111111111111111111111111111111111111111111111111111111111111111111111110FAILED
...
FAILED tests/unit_tests/job_queue/test_ert_qstat_proxy.py::test_many_concurrent_qstat_invocations[59-1000] - AssertionError: Check for race condition if AssertionError, we failed on invocation 84 out of 800
assert 1 == 0
```

tldr: There is an out-of-order problem in the loop over subprocesses, the 84th process has probably been lucky to fail before process 83 (or process 83 has been unlucky to have been scheduled late) returned with zero exit code (it found the cache file).

This can be mitigated by either increasing the sleeptime (makes the whole test slower), or perhaps accepting a few out-of-order subprocesses. These two measures will only make the test less flaky, not non-flaky.

PR #6046, opened 2023-09-05
pytest-repeat in order to trigger qstat_concurrency flakyness

berland:
This is a PR trying to trigger flakyness for qstat flakyness.

There is probably a combination of machine load and machine hardware that can trigger this.

Aids debugging #6045


eivindjahren:
This should be closed now right?

berland:
Solved in #6065

PR #6057, opened 2023-09-06
Update docs of theory behind ensemble methods

dafeda:
**Issue**
Resolves #my_issue

**Approach**

## Pre review checklist

- [ ] Updated documentation
- [ ] Ensured new behaviour is tested (opened GUI? screenshots? tried workflows?)

## Pre merge checklist

### DIFF
--- a/docs/theory/ensemble_based_methods.rst
+++ b/docs/theory/ensemble_based_methods.rst
@@ -159,8 +159,10 @@ equation for how the boat moves in time, and at selected points in time :math:`t
 :math:`\sigma_k`: The uncertainty in the position at time :math:`t_k`.
 
 :math:`x_k^{\ast}`: The *estimated/forecasted* position at time :math:`t_k`.
-   This is the position estimated from :math:`x_{k-1}` and :math:`g(x,t)`, but
+   This is the position estimated from :math:`x_{k-1}` using :math:`g(x,t)`, but
    before the observed data :math:`d_k` are taken into account.
+
+:math:`\sigma_k^{\ast}`: The uncertainty of estimate / forcast.
 
 :math:`d_k`: The observed values that are used in the updating process.
    The :math:`d_k` values are measured with a process external to the model updating.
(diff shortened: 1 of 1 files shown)


docs/theory/ensemble_based_methods.rst
```diff
@@ -159,9 +159,11 @@ equation for how the boat moves in time, and at selected points in time :math:`t
 :math:`\sigma_k`: The uncertainty in the position at time :math:`t_k`.
 
 :math:`x_k^{\ast}`: The *estimated/forecasted* position at time :math:`t_k`.
-   This is the position estimated from :math:`x_{k-1}` and :math:`g(x,t)`, but
+   This is the position estimated from :math:`x_{k-1}` using :math:`g(x,t)`, but
    before the observed data :math:`d_k` are taken into account.
 
+:math:`\sigma_k^{\ast}`: The uncertainty of estimate / forcast.
+
```
Blunde1, line 166: Could we be more clear that it belongs together with `x_k^*` by writing out the math-quantity, e.g
"The uncertainty associate estimate / forecast, :math:`x_k^{\ast}`, at time k."

docs/theory/ensemble_based_methods.rst
```diff
@@ -175,25 +177,40 @@ equation for how the boat moves in time, and at selected points in time :math:`t
...
+   x_k = Kd_k + (1 - K)x_k^{\ast} = x_k^{\ast} + K(x_k^{\ast} - d_k)
+
+:math:`K` is the Kalman gain which is a scaling term that chooses a value between the prediction and measurement.
+
+The udpated uncertainty is given by
 
 .. math::
 
-   \sigma_k^2 = \sigma_k^i{2\ast}\left(1 - \frac{\sigma_k^{2\ast}}{\sigma_d^2 + \sigma_k^{2\ast}}\right)
+   \sigma_k^2 = \frac{\sigma_k^{2*} \sigma_d^2}{\sigma_k^{2*} + \sigma_d^2} = K\sigma_d^2 = (1 - K)\sigma_k^{2*}
```
Blunde1, line 209: This is, generally, much better than before! Very good

docs/theory/ensemble_based_methods.rst
```diff
@@ -175,25 +177,40 @@ equation for how the boat moves in time, and at selected points in time :math:`t
...
+.. math:: 
+   
+   K=\frac{\sigma_k^{2*}}{\sigma_k^{2*} + \sigma_d^2}
+
+which allows us to write the udpated position as
+
+.. math:: 
+   x_k = Kd_k + (1 - K)x_k^{\ast} = x_k^{\ast} + K(x_k^{\ast} - d_k)
+
+:math:`K` is the Kalman gain which is a scaling term that chooses a value between the prediction and measurement.
```
Blunde1, line 203: It took some time for me to understand what "chooses a value between the prediction and measurement" means. But yes, this is good and meaningful. 

Would it be easier to understand if we are more explicit, i.e.
":math:`K` is the Kalman gain which is a weight term, :math: `0<K<1`, that chooses an update value :math: `x_k` that lies between the prediction and measurement."
I prefer "weight" over "scaling", because a weight is typically between 0 and 1, while scaling is to me a (slightly more) general map than a weighting.

If we want to be rigorous, we could add "The K as a weight term holds for the one-dimensional case, but has similar properties in the general sense."

docs/theory/ensemble_based_methods.rst
```diff
@@ -318,15 +335,15 @@ where :math:`G_j = \nabla g(x_j)`.  The clever trick in ensemble methods is to r
...
-Covariances :math:`\bar{C}_{xy}`, :math:`\bar{C}_{xx}`, and :math:`\bar{C}_{dd}` are
+Covariances :math:`\bar{C}_{xy}`, :math:`\bar{C}_{yy}`, and :math:`\bar{C}_{dd}` are
 estimated from the ensemble and the state vector is updated according to:
 
 .. math::
    \begin{align}
-   x_j^a &= x_j^f + \bar{C}_{xy}(\bar{C}_{xy}^{f}\bar{C}_{xx}^{-1}\bar{C}_{xy}+\bar{C}_{dd})^{-1}(d_j-y_j^f)\\
-   X^a &= X^f + \bar{C}_{xy}(\bar{C}_{xy}^{f}\bar{C}_{xx}^{-1}\bar{C}_{xy}+\bar{C}_{dd})^{-1}(D-Y_f).
+   x_j^a &= x_j^f + \bar{C}_{xy}(\bar{C}_{yy} + \bar{C}_{dd})^{-1}(d_j - y_j^f)\\
+   X^a &= X^f + \bar{C}_{xy}(\bar{C}_{yy} + \bar{C}_{dd})^{-1}(D - Y_f).
```
Blunde1, line 346: This is
1. more aligned with what we do
2. easier to read and understand

I think the previous equations perhaps is a remnant form the equations when the linear map, `H`, from forecast to response is known. Then it is natural to write it out in a similar fashion. This is however not the case for us, generally.

Blunde1 (approved):
In general, the documentation is improved. I built the new docs and everything also runs.

I have a few comments for praise, and one or two with a small change request. They are however optional and can be discussed. This is because everything is correct theoretically, and it is more about how to present the message / clarity.

I therefore hit "approve", and you can choose if you want to put them in, in which case I will hit "approve" again.

docs/theory/ensemble_based_methods.rst
dafeda: I've replaced the current description of `K` with the one you propose.

PR #6075, opened 2023-09-11
Update docs of distributions

dafeda:
**Issue**
Resolves #my_issue

**Approach**

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist


docs/reference/configuration/data_types.rst
```diff
@@ -72,167 +61,351 @@ This section only applies if a fixed seed is used:
...
 
-A stochastic variable is uniformly distributed if has a constant
-probability density on a closed interval. Thus, the uniform
-distribution is completely characterized by it's minimum and maximum
-value. To assign a uniform distribution to a variable, use the keyword
-UNIFORM, which takes a minimum and a maximum value for a the
-variable. Here is an example, which assigns a uniform distribution
-between 0 and 1 to a variable ``VAR4``:
+The ``TRUNCATED_NORMAL`` keyword is utilized to assign a truncated normal distribution to a variable.
+This disribution works as follows:
```
berland, line 126: disribution -> distribution

docs/reference/configuration/data_types.rst
```diff
@@ -72,167 +61,351 @@ This section only applies if a fixed seed is used:
...
+  VAR TRUNCATED_NORMAL <mean> <standard_deviation> <min> <max>
+
+Parameters
+~~~~~~~~~~
+- **<mean>**: Specifies the mean of the normal distribution prior to truncation.
+- **<standard_deviation>**: Indicates the standard deviation of the distribution before truncation.
+- **<min>**: Determines the lower truncation limit.
+- **<max>**: Establishes the upper truncation limit.
+
+Example
```
berland, line 147: Emphasize that this is an example showing how this should not be used? "Bad example", followed by a "Good example"?

dafeda: What's a good example?

docs/reference/configuration/data_types.rst
dafeda: Fixed 👍

docs/reference/configuration/data_types.rst
berland: Where the clamping is an "exceptional event", so that there is no visible spikes at the truncation limit.

dafeda: I have added an example of what I think is reasonable usage of this distribution.

docs/reference/configuration/data_types.rst
```diff
@@ -72,167 +61,352 @@ This section only applies if a fixed seed is used:
...
+~~~~~~
 ::
 
-   VAR1   NORMAL    0   1
+  VAR NORMAL <mean_value> <standard_deviation>
+
+Parameters
+~~~~~~~~~~
+- **<mean_value>**: Specifies the mean of the normal distribution.
+- **<standard_deviation>**: Indicates the standard deviation of the normal distribution.
```
Blunde1, line 79: Probably re-use "Specifies" from above instead of "Indicates".
Or just start with "The" skipping entirely.

docs/reference/configuration/data_types.rst
```diff
@@ -72,167 +61,352 @@ This section only applies if a fixed seed is used:
...
-LOGNORMAL
-,,,,,,,,,
+LOGNORMAL: Log Normal Distribution
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-A stochastic variable is log normally distributed if the logarithm of the
-variable is normally distributed. In other words, if X is normally distributed,
-then Y = exp(X) is log normally distributed.
+The ``LOGNORMAL`` keyword is used to assign a log normal prior to a variable. A variable is considered log normally distributed if the logarithm of that variable follows a normal distribution. 
+In mathematical terms, if :math:`X` is normally distributed, then :math:`Y = e^X` is log normally distributed.
```
Blunde1, line 98: Would prefer skipping "In mathematical terms, " and start at "If". We are already talking mathematically.

docs/reference/configuration/data_types.rst
```diff
@@ -72,167 +61,352 @@ This section only applies if a fixed seed is used:
...
 
-A log normal prior is suited to model positive quantities with a heavy tail
-(tendency to take large values). To set a log normal prior, use the keyword
-LOGNORMAL. It takes two arguments, the mean and standard deviation of the
-*logarithm* of the variable:
+Log normal priors are especially suitable for modeling positive quantities that exhibit a heavy tail, indicating a tendency for the quantity to occasionally take large values.
```
Blunde1, line 100: I had to think about what "positive quantities" was. I think the "quantities" word throws me off. Can we change to "values", or do a bigger change to something like "modeling values that are exclusively greater than zero..." or "that are known to always be greater than zero".

docs/reference/configuration/data_types.rst
```diff
@@ -72,167 +61,352 @@ This section only applies if a fixed seed is used:
...
 
-   VAR2   LOGNORMAL  0  1
+  VAR LOGNORMAL <log_mean> <log_standard_deviation>
 
-TRUNCATED_NORMAL
-,,,,,,,,,,,,,,,,,
+Parameters
+~~~~~~~~~~
+- **<log_mean>**: Represents the mean of the logarithm of the variable.
+- **<log_standard_deviation>**: Denotes the standard deviation of the logarithm of the variable.
```
Blunde1, line 111: Would prefer skipping "Represents" and "Denots" to start directly at "The". This is exactly what it is and it is okay to be clear about it.

docs/reference/configuration/data_types.rst
```diff
@@ -72,167 +61,352 @@ This section only applies if a fixed seed is used:
...
+
+Example
+~~~~~~~
+For a variable with its logarithm having a mean of 0 and a standard deviation of 1:
+::
+
+   VAR LOGNORMAL 0 1
 
-This *TRUNCATED_NORMAL* distribution works as follows:
+.. image:: fig/lognormal.png
```
Blunde1, line 120: How about "Histogram from values sampled from a lognormal variable specified with log-mean of 0 and log standard deviation 1."

docs/reference/configuration/data_types.rst
```diff
@@ -72,167 +61,352 @@ This section only applies if a fixed seed is used:
...
+~~~~~~
 ::
 
-   VAR3  TRUNCATED_NORMAL  1  0.25   0  10
+  VAR TRUNCATED_NORMAL <mean> <standard_deviation> <min> <max>
 
+Parameters
+~~~~~~~~~~
+- **<mean>**: Specifies the mean of the normal distribution prior to truncation.
+- **<standard_deviation>**: Indicates the standard deviation of the distribution before truncation.
```
Blunde1, line 143: Again, I would prefer to start directly at "The".

docs/reference/configuration/data_types.rst
```diff
@@ -72,167 +61,352 @@ This section only applies if a fixed seed is used:
...
-this distribution for a situation with mean 1, standard deviation 0.25
-and hard limits 0 and 10:
+The ``TRUNCATED_NORMAL`` keyword is utilized to assign a truncated normal distribution to a variable.
+This distribution works as follows:
 
+   1. Draw a random variable :math:`X \sim N(\mu,\sigma)`.
+   2. Clamp :math:`X` to the interval [min, max].
+
+It's important to note that this is **not** a proper normal distribution due to the constraints applied.
+Thus, the act of clamping values to the `[min,max]` interval should be an exceptional event.
```
Blunde1, line 132: Most of the distributions we expose are not normal distributions. Should it be an exceptional event to use non-normal distributions, or why/how is this different?

docs/reference/configuration/data_types.rst
```diff
@@ -72,167 +61,352 @@ This section only applies if a fixed seed is used:
...
+~~~~~~~~~~
+- **<mean>**: Specifies the mean of the normal distribution prior to truncation.
+- **<standard_deviation>**: Indicates the standard deviation of the distribution before truncation.
+- **<min>**: Determines the lower truncation limit.
+- **<max>**: Establishes the upper truncation limit.
+
+Example
+~~~~~~~
+Reasonable usage of this distribution where the number of values being clamped is relatively low.
+::
```
Blunde1, line 150: Why is this reasonable compared to the example below? Or, why is excessive clamping problematic?

docs/reference/configuration/data_types.rst
```diff
@@ -72,167 +61,352 @@ This section only applies if a fixed seed is used:
...
+Syntax
+~~~~~~
+::
+
+  VAR UNIFORM <min_value> <max_value>
+
+Parameters
+~~~~~~~~~~
+- **<min_value>**: Specifies the lower bound of the uniform distribution.
+- **<max_value>**: Sets the upper bound of the uniform distribution.
```
Blunde1, line 179: Start with "The". I would prefer the same with parameters for the distributions below as well.

docs/reference/configuration/data_types.rst
```diff
@@ -72,167 +61,352 @@ This section only applies if a fixed seed is used:
...
 
-ERRF
-,,,,,
+ERRF: Error Function-Based Prior
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The ERRF keyword is used to define a prior resulting from applying the error
-function to a normally distributed variable with mean 0 and variance 1. The
-keyword takes four arguments:
+The ``ERRF`` keyword allows creating prior distributions derived from applying the error function to a standard normal variable (with mean 0 and variance 1).
```
Blunde1, line 284: We do not only apply the error function. I think this is more precise: "...derived from applying the normal CDF (involving the error function) to a standard normal variable. Note that the CDF is not necessarily the standard normal, as `SKEWNESS` and `WIDTH` corresponds to its negative mean and standard deviation respectively. This allows flexibility in creating distributions of diverse shapes and symmetries."

docs/reference/configuration/data_types.rst
```diff
@@ -72,167 +61,352 @@ This section only applies if a fixed seed is used:
...
+~~~~~
+Keep in mind the interactions between the parameters, especially when both ``SKEWNESS`` and ``WIDTH`` are adjusted.
+Their combination can result in a wide range of distribution shapes.
+
+DERRF: Discrete Error Function-Based Distribution
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The ``DERRF`` keyword is a discrete version of the ``ERRF`` keyword.
+It is designed for creating distributions based on the error function but with discrete output values.
+This keyword facilitates the creation of distributions with various shapes and asymmetries.
```
Blunde1, line 334: How about "... facilitates sampling from discrete distributions with various shapes and asymmetries"

Blunde1 (commented):
Most comments are preferences I have to be more explicit. A few are corrections. The docs looks much better now, in particular the figures are guaranteed to be very illuminating for users. Later iterations could perhaps include the exact pdf's we are smapling from. I was unsure about the size of the figures, they are quite large, but I do not have strong preferences and thus not requesting anything here.

docs/reference/configuration/data_types.rst
```diff
@@ -23,24 +23,13 @@ The data types in ERT can be classified based on two main criteria:
 Note: All data types share a common namespace, meaning that each keyword must
 be globally unique.
 
-Parameters
-----------
-
-The *parameter* datatypes will serve as input to the simulator. For a pure
-uncertainty study they will pass right through, model updating update
-parameters. Sample internally or load externally.
```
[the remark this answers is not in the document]
dafeda: I just don't think the description does not provide any value.

docs/reference/configuration/data_types.rst
dafeda: I think this is talking about the `TRUNCATED_NORMAL` (or clamped normal) specifically but not sure.
Perhaps it is because you can end up with too high probabilities of end-points because of the clamping?

docs/reference/configuration/data_types.rst
dafeda: I have updated all (I hope) definitions to just start with "The" as you suggest.

docs/reference/configuration/data_types.rst
dafeda: Fixed

docs/reference/configuration/data_types.rst
dafeda: Updated to: "...modeling positive values..."

docs/reference/configuration/data_types.rst
dafeda: Fixed

docs/reference/configuration/data_types.rst
dafeda: Fixed

docs/reference/configuration/data_types.rst
dafeda: Done

docs/reference/configuration/data_types.rst
dafeda: Fixed

docs/reference/configuration/data_types.rst
dafeda: Fixed

dafeda:
> Most comments are preferences I have to be more explicit. A few are corrections. The docs looks much better now, in particular the figures are guaranteed to be very illuminating for users. Later iterations could perhaps include the exact pdf's we are smapling from. I was unsure about the size of the figures, they are quite large, but I do not have strong preferences and thus not requesting anything here.

FYI, the sizes of the figures are what we get by default when saving from ERT.

docs/reference/configuration/data_types.rst
dafeda: Not sure about this.
I've heard people talk about this and that it can lead to too high probabilities of end-points.

Blunde1: I think the answer is in "more" clamping results in implicit increased non-linearity of the forward model (from gaussian sample --> desired distribution --> forward model result). So indeed the validity of data assimilation linearization can be put more into question with increased clamping. I do not think it will necessarily be worse than other distributions though, e.g. using "errf" and "derrf" we can induce a great deal of non-linearity. Since we are not doing any warnings for non-linear distributions, and we will have non-linearity due to other parts of the forward model anyway, should we explicitly warn against it here?

dafeda: We should perhaps create a document describing the effect of these transformations, but I agree with you that we should not warn about it here.
I've removed it and the figure showing "bad" clamping.

Blunde1 (approved):
LGTM, great work.

PR #6122, opened 2023-09-20
Count memory usage for grandchildren in job_dispatch

### LINKED ISSUE #6108: ERT reports wrong max memory usage for forward models
**Describe the bug**
When running RMS, which is typically very memory hungry, the memory usage reported by ERT is typically only around 12 megabytes. This is clearly wrong.

**To reproduce**
Run an ERT model that includes RMS.

Watch the memory usage in the GUI and compare alongside with e.g. `htop` (logged into the relevant compute node)

**Expected behaviour**
Comparable memory figures reported by both `htop` and ERT.

**Environment**
 - OS: RHEL7
 - ERT/Komodo release: 2023.06
 - Python version: 3.8

Reported in: https://app.slack.com/client/T02JL00JU/CDPJULWR4/thread/CDPJULWR4-1695038357.128309

berland:
job_dispatch/job_runner only picked up the memory usage for its direct child. When the direct child is only a wrapper for some potentially memory hungry processes, the memory usage reported from job_dispatch is essentially garbage.

This PR lets the job runner scan through all its grandchildren when estimating memory usage for a started forward model.

NB: This counts RSS (resident set size). When summing over multiple processes, the RSS sum will count shared memory pages multiple times, thus the sum is likely to be an overestimate for the actual memory load imposed. This is probably less of a problem, as the target memory consumption numbers that are of interest, stem from big datasets which are not shared among processes.

**Issue**
Resolves #6108 

**Approach**
Use the `children(recursive=True)` function in `psutil` to find all descendants.

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/tests/unit_tests/job_runner/test_job.py
+++ b/tests/unit_tests/job_runner/test_job.py
@@ -34,6 +34,47 @@ def test_run_with_process_failing(
 
 
 @pytest.mark.usefixtures("use_tmpdir")
+def test_memory_usage_counts_grandchildren():
+    scriptname = "recursive_memory_hog.py"
+    with open(scriptname, "w", encoding="utf-8") as script:
+        script.write(
+            """#!/usr/bin/env python
+import os
+import sys
+import time
+
+counter = int(sys.argv[1])
+numbers = list(range(int(1e6)))
+if counter > 0:
+    parent = os.fork()
+    if not parent:
+        os.execv(sys.argv[0], [sys.argv[0], str(counter - 1)])
+time.sleep(0.3)"""  # Too low sleep will make the test faster but flaky
+        )
+    executable = os.path.realpath(scriptname)
+    os.chmod(scriptname, stat.S_IRWXU | stat.S_IRWXO | stat.S_IRWXG)
+
+    def max_memory_per_subprocess_layer(layers: int) -> int:
+        job = Job(
+            {
+                "executable": executable,
+                "argList": [str(layers)],
+            },
+            0,
+        )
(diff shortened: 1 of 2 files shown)


tests/unit_tests/job_runner/test_job.py
```diff
@@ -33,6 +33,47 @@ def test_run_with_process_failing(
...
+        script.write(
+            """#!/usr/bin/env python
+import os
+import sys
+import time
+
+counter = int(sys.argv[1])
+numbers = list(range(int(1e6)))
+if counter > 0:
+    parent = os.fork()
```
xjules, line 49: How this will work on Mac?

tests/unit_tests/job_runner/test_job.py
```diff
@@ -33,6 +33,47 @@ def test_run_with_process_failing(
...
+    if not parent:
+        os.execv(sys.argv[0], [sys.argv[0], str(counter - 1)])
+time.sleep(0.3)"""  # Too low sleep will make the test faster but flaky
```
xjules, line 52: Also wondering if `sleep(0.3)` is enough?

berland: It seems enough. Making it too high will make the test slower so that we don't want either.

tests/unit_tests/job_runner/test_job.py
berland: I have no idea if this will work differently on Mac... Passing tests is noted :) (docs do not hint towards there being a mac differnce here..?)

PR #6130, opened 2023-09-21
Divide outside of floor

### LINKED ISSUE #6129: `trans_derrf` only yields maximum two outcomes no matter `_steps` corresponding to `NBINS`
**Describe the bug**
Implementation is wrong for 
```
    @staticmethod
    def trans_derrf(x: float, arg: List[float]) -> float:
        '''Observe that the argument of the shift should be \"+\"'''
        _steps, _min, _max, _skew, _width = int(arg[0]), arg[1], arg[2], arg[3], arg[4]
        y = math.floor(
            _steps
            * 0.5
            * (1 + math.erf((x + _skew) / (_width * math.sqrt(2.0))))
            / (_steps - 1)
        )
        return _min + y * (_max - _min)
```
when when `_steps` is > 2 it yields maximum two outcomes.
This does not match intended behaviour https://fmu-docs.equinor.com/docs/ert/reference/configuration/data_types.html#derrf

Blunde1:
**Issue**
Resolves #6129 

**Approach**
The part
```
0.5
            * (1 + math.erf((x + _skew) / (_width * math.sqrt(2.0))))
        ) 
```
is the cdf of a gaussian, so it yields something on the interval 0,1.
multiplying by steps and taking floor then yields steps different possibilities (intended)
the division should occur outside of floor to standardize to the 0, 1 interval again.

### DIFF
--- a/tests/unit_tests/config/test_transfer_functions.py
+++ b/tests/unit_tests/config/test_transfer_functions.py
@@ -1,6 +1,7 @@
 import numpy as np
 from hypothesis import given
 from hypothesis import strategies as st
+from scipy.stats import norm
 
 from ert.config import TransferFunction
 
@@ -79,3 +80,80 @@ def test_that_truncated_normal_stretches(x, arg):
         return
     result = TransferFunction.trans_truncated_normal(x, arg)
     assert np.isclose(result, expected)
+
+
+def valid_derrf_parameters():
+    """All elements in R, min<max, and width>0"""
+    steps = st.integers(min_value=2, max_value=1000)
+    min_max = (
+        st.tuples(
+            st.floats(
+                min_value=-1e6, max_value=1e6, allow_nan=False, allow_infinity=False
+            ),
+            st.floats(
+                min_value=-1e6, max_value=1e6, allow_nan=False, allow_infinity=False
+            ),
+        )
+        .map(sorted)
+        .filter(lambda x: x[0] < x[1])  # filter out edge case of equality
+    )
+    skew = st.floats(allow_nan=False, allow_infinity=False)
+    width = st.floats(
+        min_value=0.01, max_value=1e6, allow_nan=False, allow_infinity=False
+    )
+    return min_max.flatmap(
+        lambda min_max: st.tuples(
+            steps, st.just(min_max[0]), st.just(min_max[1]), skew, width
+        )
+    )
+
+
+@given(st.floats(allow_nan=False, allow_infinity=False), valid_derrf_parameters())
+def test_that_derrf_is_within_bounds(x, arg):
+    """The result shold always be between (or equal) min and max"""
+    result = TransferFunction.trans_derrf(x, arg)
+    assert arg[1] <= result <= arg[2]
+
+
+@given(
+    st.lists(st.floats(allow_nan=False, allow_infinity=False), min_size=2),
+    valid_derrf_parameters(),
+)
+def test_that_derrf_creates_at_least_steps_or_less_distinct_values(xlist, arg):
+    """derrf cannot create more than steps distinct values"""
+    res = [TransferFunction.trans_derrf(x, arg) for x in xlist]
+    assert len(set(res)) <= arg[0]
+
+
+@given(st.floats(allow_nan=False, allow_infinity=False), valid_derrf_parameters())
+def test_that_derrf_corresponds_scaled_binned_normal_cdf(x, arg):
+    """Check correspondance to normal cdf with -mu=_skew and sd=_width"""
+    _steps, _min, _max, _skew, _width = arg
+    q_values = np.linspace(start=0, stop=1, num=_steps)
+    q_checks = np.linspace(start=0, stop=1, num=_steps + 1)[1:]
+    p = norm.cdf(x, loc=-_skew, scale=_width)
+    bin_index = np.digitize(p, q_checks, right=True)
+    expected = q_values[bin_index]
+    # scale and ensure ok numerics
+    expected = _min + expected * (_max - _min)
+    if expected > _max or expected < _min:
+        np.clip(expected, _min, _max)
+    result = TransferFunction.trans_derrf(x, arg)
+    assert np.isclose(result, expected)
+
+
+@given(
+    st.tuples(
+        st.floats(allow_nan=False, allow_infinity=False),
+        st.floats(allow_nan=False, allow_infinity=False),
+    )
+    .map(sorted)
+    .filter(lambda x: x[0] < x[1]),
+    valid_derrf_parameters(),
+)
+def test_that_derrf_is_non_strictly_monotone(x_tuple, arg):
+    """`derrf` is a non-strict monotone function"""
+    x1, x2 = x_tuple
+    assert TransferFunction.trans_derrf(x1, arg) <= TransferFunction.trans_derrf(
+        x2, arg
+    )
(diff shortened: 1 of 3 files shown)


dafeda (approved):
Beautiful!
Remember to capitalize your commit message.

andreas-el:
You should know that the failing tests were setup as part of the refactoring process (moving from C++ to python), and to me it looks like the original had this bug to begin with.
```diff
C++:
-static double trans_derrf(double x, const std::vector<double> arg) {
-    int steps = arg[0];
-    double min = arg[1];
-    double max = arg[2];
-    double skewness = arg[3];
-    double width = arg[4];
-    double y;
-
-    y = floor(steps * 0.5 * (1 + erf((x + skewness) / (width * sqrt(2.0)))) /
-              (steps - 1));
-    return min + y * (max - min);
-}

Python:
+    def trans_derrf(x: float, arg: List[float]) -> float:
+        '''Observe that the argument of the shift should be \"+\"'''
+        _steps, _min, _max, _skew, _width = int(arg[0]), arg[1], arg[2], arg[3], arg[4]
+        y = math.floor(
+            _steps
+            * 0.5
+            * (1 + math.erf((x + _skew) / (_width * math.sqrt(2.0))))
+            / (_steps - 1)
+        )
+        return _min + y * (_max - _min)
```

Blunde1:
If `_skew` is too large relative to variance, we get a probability rounded of to exactly `1.0`
The resulted binning is then between `min` and `max+1`.
This is a problem in the original `TransferFunction.trans_errf` as well
```
  def test_that_derrf_is_within_bounds(x, arg):
        result = TransferFunction.trans_derrf(x, arg)
>       assert arg[1] <= result <= arg[2]
E       assert 2.0 <= 1.0
E       Falsifying example: test_that_derrf_is_within_bounds(
E           x=0.0,
E           arg=(2, 0.0, 1.0, 9.0, 1.0),
E       )
```
All hail hypothesis for finding this!

Blunde1:
Let's resolve #6166 as well

Blunde1:
This is ready for review

Blunde1:
Only coverage tests fail.
Rebase + one more test that makes sure the initial problem #6129 using normal cdf

dafeda:
I may have found an example that fails:

`@reproduce_failure('6.84.3', b'AXicY2BgZIAAEL1nAQOMt/8/BDDgAIwAHtgJGQ==')`

Could you try running `test_that_derrf_is_within_bounds` with that decorator and see if you can reproduce the error. 
I may have done something wrong...

I also think that we should be able to get rid of the for-loop using for example `np.digitize`.
Here's perhaps a way:

```python
    def trans_derrf(x: float, arg: List[float]) -> float:
        """
        Bin the result of `trans_errf` with `min=0` and `max=1` to closest of `nbins`
        linearly spaced values on [0,1]. Finally map [0,1] to [min, max].
        """
        _steps, _min, _max, _skew, _width = (
            int(arg[0]),
            arg[1],
            arg[2],
            arg[3],
            arg[4],
        )
        q_values = np.linspace(start=0, stop=1, num=_steps)
        q_checks = np.linspace(start=0, stop=1, num=_steps + 1)[1:]
        y = TransferFunction.trans_errf(x, [0, 1, _skew, _width])
        bin_index = np.digitize(y, q_checks, right=True)
        y_binned = q_values[bin_index]
        result = _min + y_binned * (_max - _min)
        if result > _max or result < _min:
            warnings.warn(
                "trans_derff suffered from catastrophic loss of precision, clamping to min,max",
                stacklevel=1,
            )
            return np.clip(result, _min, _max)
        return result
```

Blunde1:
Reproducing the test I see
```
x = -1.0, arg = (2, -1.7976931348623157e+308, 9.9792015476736e+291, 0.0, 1.0)

    @given(st.floats(allow_nan=False, allow_infinity=False), valid_derrf_parameters())
    @reproduce_failure("6.84.3", b"AXicY2BgZIAAEL1nAQOMt/8/BDDgAIwAHtgJGQ==")
    def test_that_derrf_is_within_bounds(x, arg):
        """The result shold always be between (or equal) min and max"""
        result = TransferFunction.trans_derrf(x, arg)
>       assert arg[1] <= result <= arg[2]
E       assert -1.7976931348623157e+308 <= nan
```
It seems that `min` and `max` are so far apart that simple arithmetic fails, which will likely be a problem for all code unless special care is taken.
I think we do not take special care of this, so unless I am wrong (which I might of course be), it makes sense to me tobound the values of `valid_derrf_parameters()`

Blunde1:
> I also think that we should be able to get rid of the for-loop using for example `np.digitize`. Here's perhaps a way:
> 
> ```python
>         bin_index = np.digitize(y, q_checks, right=True)
>         y_binned = q_values[bin_index]
> ```

This is better, thanks.

Blunde1:
Hypothesis found through testing `test_that_derrf_corresponds_scaled_binned_normal_cdf` (@reproduce_failure("6.84.3", b"AXicY2DADxhxSgAAAFUAAw==")) that to mimic the previously loop implementation we should have apply `np.digitize` with `right=False`.
What is actually correct? 
The documentation says
image
reproducing the `floor()` functionality in the original implementation, I think `right=False` is correct.

tests/unit_tests/config/test_transfer_functions.py
```diff
@@ -79,3 +80,82 @@ def test_that_truncated_normal_stretches(x, arg):
...
+
+
+@given(
+    st.lists(st.floats(allow_nan=False, allow_infinity=False), min_size=2),
+    valid_derrf_parameters(),
+)
+def test_that_derrf_creates_at_least_steps_or_less_distinct_values(xlist, arg):
+    """derrf cannot create more than steps distinct values"""
+    res = []
+    for x in xlist:
```
dafeda, line 125: A bit more Pythonic perhaps:

`res = [TransferFunction.trans_derrf(x, arg) for x in xlist]`

dafeda:
Tried running it a few times and `test_that_derrf_corresponds_scaled_binned_normal_cdf` fails with:

```
E       assert False
E        +  where False = <function isclose at 0x10fb3b870>(0.2500038146972659, 0.25000762939453125)
E        +    where <function isclose at 0x10fb3b870> = np.isclose
E       Falsifying example: test_that_derrf_corresponds_scaled_binned_normal_cdf(
E           x=1.0,
E           arg=(2, -34359738368.0, 0.2500038146972659, 0.0, 1.0),
E       )
E       
E       You can reproduce this example by temporarily adding @reproduce_failure('6.84.3', b'AXicY2CAAUYgPiAFYjVAeRwMeAAjADnbAWY=') as a decorator on your test case
```

tests/unit_tests/config/test_transfer_functions.py
Blunde1: Thanks, updated with [84d5b33](https://github.com/equinor/ert/pull/6130/commits/84d5b33a158e25a446f0f26bacbaa922460763ab)

Blunde1:
> Tried running it a few times and `test_that_derrf_corresponds_scaled_binned_normal_cdf` fails with:
> 
> ```
> E       assert False
> E        +  where False = <function isclose at 0x10fb3b870>(0.2500038146972659, 0.25000762939453125)
> E        +    where <function isclose at 0x10fb3b870> = np.isclose
> E       Falsifying example: test_that_derrf_corresponds_scaled_binned_normal_cdf(
> E           x=1.0,
> E           arg=(2, -34359738368.0, 0.2500038146972659, 0.0, 1.0),
> E       )
> E       
> E       You can reproduce this example by temporarily adding @reproduce_failure('6.84.3', b'AXicY2CAAUYgPiAFYjVAeRwMeAAjADnbAWY=') as a decorator on your test case
> ```

This is again due to precision. Resolving by restricting valid_derrf_parameters() even further

PR #6150, opened 2023-09-22
Mark qstat_proxy concurrency as flaky

### LINKED ISSUE #6138: qstat test is still flaky
```
    @pytest.mark.skipif(sys.platform.startswith("darwin"), reason="No flock on MacOS")
    def test_many_concurrent_qstat_invocations(tmpdir, monkeypatch):
        """Run many qstat invocations simultaneously, with a mocked backend qstat
        script that logs how many times it is invoked, then we assert that it has
        not been called often.
    
        If this test has enough invocations (or the hardware is slow enough), this test
        will also pass the timeout in the script yielding a rerun of the backend. In that
        scenario there is a risk for a race condition where the cache file is blank when
        another process is reading from it. This test also assert that failures are only
        allowed to happen in a sequence from the second invocation (the first process
        succeeds because it calls the backend, the second one fails because there is no
        cache file and it is locked.)
    
        This test will dump something like
        01111111111000000000000000000000000000000000000 to stdout
        where each digit is the return code from the proxy script. Only one sequence of 1's
        is allowed after the start, except for some noise at the end of the sequence of 1's
        due to out-of-order finalization of the subprocesses started in this test.
        """
        starttime = time.time()
        sleeptime = 0.02  # seconds between each qstat process started in this test
        # (the mocked qstat backend sleeps for 0.5 seconds to facilitate races)
        cache_timeout = 2  # This is CACHE_TIMEOUT in the shell script
        invocations = int(cache_timeout / sleeptime * 2)
    
        monkeypatch.chdir(tmpdir)
        Path("log").mkdir()  # The mocked backend writes to this directory
        subprocesses = []
        with testpath.MockCommand("qstat", python=MOCKED_QSTAT_BACKEND_LOGGING):
            for _ in range(invocations):
                # pylint: disable=consider-using-with
                # process.wait() is called below
                subprocesses.append(
                    subprocess.Popen(
                        [PROXYSCRIPT, "15399", PROXYFILE_FOR_TESTS],
                        stdout=subprocess.DEVNULL,
                    )
                )
                # The asserts below in this test function depend on each subprocess
                # to finish in order of invocation:
                time.sleep(sleeptime)
    
            class CacheState(enum.Enum):
                # Only consecutive transitions are allowed.
                FIRST_INVOCATION = 0
                FIRST_HOLDS_FLOCK = 1
                CACHE_EXISTS = 2
    
            state = None
            cache_hits = 0
            for _, process in enumerate(subprocesses):
                process.wait()
                if state is None:
                    if process.returncode == 0:
                        state = CacheState.FIRST_INVOCATION
>                   assert state is not None, "First invocation should not fail"
E                   AssertionError: First invocation should not fail
E                   assert None is not None
```
https://github.com/equinor/ert/actions/runs/6271871608/job/17032295425?pr=6137

berland:
This fails seldom enough for simple rerunning to be sufficient.

**Issue**
Resolves #6138 

**Approach**
♻️ 

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/tests/unit_tests/job_queue/test_ert_qstat_proxy.py
+++ b/tests/unit_tests/job_queue/test_ert_qstat_proxy.py
@@ -241,6 +241,7 @@ def test_options_passed_through_proxy(tmpdir, options, expected, monkeypatch):
     # (the output from the mocked qstat happens to adhere to json syntax)
 
 
+@flaky(max_runs=3, min_passes=1)
 @pytest.mark.skipif(sys.platform.startswith("darwin"), reason="No flock on MacOS")
 def test_many_concurrent_qstat_invocations(tmpdir, monkeypatch):
     """Run many qstat invocations simultaneously, with a mocked backend qstat


berland:
Superseded by another PR

PR #6211, opened 2023-09-30
Create param data in same way as response data

dafeda:
## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/storage/local_experiment.py
+++ b/src/ert/storage/local_experiment.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import json
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Dict, Generator, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, Generator, List, Optional
 from uuid import UUID
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


src/ert/storage/local_experiment.py
```diff
@@ -147,10 +148,6 @@ def __init__(  # pylint: disable=too-many-arguments
         with open(response_file, "w", encoding="utf-8") as f:
             json.dump(response_data, f, default=str)
 
-    @property
-    def ensembles(self) -> Generator[LocalEnsembleAccessor, None, None]:
-        yield from super().ensembles  # type: ignore
-
```
kvashchuka, line 153: why is this function not needed?

dafeda: My thinking is that it's not needed since it just calls its parents `ensembles`.

This also makes the typing misleading as the parent is a `LocalExperiment` reader and its `ensemble` property has the signature: `def ensembles(self) -> Generator[LocalEnsembleReader, None, None]:`.

kvashchuka: Okay, I checked where `ensembles()` function is called, and to me it looks like indeed only parent function is used 👍

PR #6226, opened 2023-10-02
Update status border color based on job status

### LINKED ISSUE #6191: ERT gui realization status border lagging
**Describe the bug**
When running ert gui, it seems like the jobs status and realization status (box inside-border) get updates from completely different sources. Either that, or the realization status is lagging behind.
We have observed yellow subpixel/cell indicating a running job, while the realization status still shows blue (waiting).
See screenshots attached (from Azure)

The easiest way to reproduce this is to run on Azure, since that is the slowest queue we have.
It should be possible to reproduce this other places also.

**To reproduce**
1. run ert gui
2. see if there is a slight lag between job status and realization status

**Expected behaviour**
The realization status should follow the job status accurately. If a job is running, then the realization should for sure indicate running too.
Screenshot 2023-09-27 at 12 41 17
Screenshot 2023-09-27 at 12 46 46

andreas-el:
Use inner status colours to determine the overall status drawn outside the realization boxes

Resolves #6191 

Tested on Azure.

**Approach**

## Pre review checklist

- [x] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/gui/simulation/view/realization.py
+++ b/src/ert/gui/simulation/view/realization.py
@@ -1,7 +1,8 @@
 import math
+from typing import Final
 
 from qtpy.QtCore import QModelIndex, QRect, QSize, Qt, Signal
-from qtpy.QtGui import QColorConstants, QImage, QPainter, QPen
+from qtpy.QtGui import QColor, QColorConstants, QImage, QPainter, QPen
 from qtpy.QtWidgets import (
     QAbstractItemView,
     QListView,
@@ -12,8 +13,16 @@ from qtpy.QtWidgets import (
     QWidget,
 )
 
+from ert.ensemble_evaluator import state
 from ert.gui.model.real_list import RealListModel
-from ert.gui.model.snapshot import RealJobColorHint, RealLabelHint, RealStatusColorHint
+from ert.gui.model.snapshot import RealJobColorHint, RealLabelHint
+
+COLOR_UNKNOWN: Final[QColor] = QColor(*state.COLOR_UNKNOWN)
+COLOR_WAITING: Final[QColor] = QColor(*state.COLOR_WAITING)
+COLOR_PENDING: Final[QColor] = QColor(*state.COLOR_PENDING)
+COLOR_RUNNING: Final[QColor] = QColor(*state.COLOR_RUNNING)
+COLOR_FINISHED: Final[QColor] = QColor(*state.COLOR_FINISHED)
+COLOR_FAILED: Final[QColor] = QColor(*state.COLOR_FAILED)
 
 
 class RealizationWidget(QWidget):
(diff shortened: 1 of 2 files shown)


src/ert/gui/simulation/view/realization.py
```diff
@@ -140,3 +149,27 @@ def _paint_inner_grid(self, painter: QPainter, rect: QRect, colors) -> None:
...
+
+
+def determine_realization_status(job_colors) -> QColor:
+    states = [
+        COLOR_UNKNOWN,
+        COLOR_FINISHED,
+        COLOR_WAITING,
+        COLOR_PENDING,
+    ]
+    real_status_color = COLOR_UNKNOWN
```
jonathan-eq, line 161: It took a while before I realized "real" here is short for realization. I think updating this variable name to "realization_status_color" would make sense

jonathan-eq (changes requested):

larsevj (changes requested):
One question regarding failed state.

src/ert/gui/simulation/view/realization.py
andreas-el: Sure thing, the naming was originally due to it holding the queue status, while now it will hold the most prominent job status.

src/ert/gui/simulation/view/realization.py
```diff
@@ -140,3 +141,27 @@ def _paint_inner_grid(self, painter: QPainter, rect: QRect, colors) -> None:
...
+        if color in [QColor(*state.COLOR_FAILED), QColor(*state.COLOR_RUNNING)]:
+            real_status_color = color
+            break
+
+        color_index = states.index(color)
+
+        if color_index > current_color_index:
+            real_status_color = states[color_index]
+            current_color_index = color_index
+
```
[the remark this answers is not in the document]
andreas-el: Can it occur? Both yes and no. The jobs are executed sequentially, so in that respect no. But when I do this on the snake oil test-data (raise exception on 2 out of 3 jobs), the first job very shortly revisits running state for some reason, so yes. I would still argue we should keep the failed state, as this is the overall state for that realization.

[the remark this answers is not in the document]
andreas-el: I ended selecting this solution even though mine only looped n/2 during execution (but full n when box is green).
I also tried caching this but realized that just the hash function would probably be slower than iteration.

src/ert/gui/simulation/view/realization.py
```diff
@@ -140,3 +148,16 @@ def _paint_inner_grid(self, painter: QPainter, rect: QRect, colors) -> None:
...
+    states = [
+        COLOR_UNKNOWN,
+        COLOR_FINISHED,
+        COLOR_WAITING,
+        COLOR_PENDING,
+        COLOR_RUNNING,
+        COLOR_FAILED,
+    ]
+
+    return max(job_colors, key=lambda x: states.index(x), default=COLOR_UNKNOWN)
```
larsevj, line 163: Think pylint complains because one may replace lambda by `key = states.index`.

larsevj (approved):
LGTM

src/ert/gui/simulation/view/realization.py
andreas-el: Then Ill fix that too. 👍

jonathan-eq (approved):
Great job Andreas!

PR #6247, opened 2023-10-05
Convert simulation arguments to dataclass

dafeda:
## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/cli/model_factory.py
+++ b/src/ert/cli/model_factory.py
@@ -22,6 +22,13 @@ from ert.run_models import (
     MultipleDataAssimilation,
     SingleTestRun,
 )
+from ert.run_models.run_arguments import (
+    EnsembleExperimentRunArguments,
+    ESMDARunArguments,
+    ESRunArguments,
+    SIESRunArguments,
+    SingleTestRunArguments,
+)
 from ert.validation import ActiveRange
 
 if TYPE_CHECKING:
@@ -62,13 +69,14 @@ def create_model(
 def _setup_single_test_run(
     ert: EnKFMain, storage: StorageAccessor, args: Namespace, experiment_id: UUID
 ) -> SingleTestRun:
-    simulations_argument = {
-        "active_realizations": [True],
-        "current_case": args.current_case,
-        "simulation_mode": "Single test run",
-    }
-    model = SingleTestRun(simulations_argument, ert, storage, experiment_id)
-    return model
+    return SingleTestRun(
+        SingleTestRunArguments(
+            current_case=args.current_case,
+        ),
+        ert,
+        storage,
+        experiment_id,
+    )
 
 
 def _setup_ensemble_experiment(
@@ -90,37 +98,33 @@ def _setup_ensemble_experiment(
             stacklevel=1,
         )
 
-    simulations_argument = {
-        "active_realizations": active_realizations,
-        "iter_num": int(args.iter_num),
-        "current_case": args.current_case,
-        "simulation_mode": "Ensemble experiment",
-    }
-    model = EnsembleExperiment(
-        simulations_argument, ert, storage, ert.get_queue_config(), experiment_id
-    )
-
-    return model
-
-
-def _setup_ensemble_smoother(
-    ert: EnKFMain, storage: StorageAccessor, args: Namespace, experiment_id: UUID
-) -> EnsembleSmoother:
-    simulations_argument = {
-        "active_realizations": _realizations(args, ert.getEnsembleSize()),
-        "current_case": args.current_case,
-        "target_case": _target_case_name(ert, args, format_mode=False),
-        "analysis_module": "STD_ENKF",
-        "simulation_mode": "Ensemble smoother",
-    }
-    model = EnsembleSmoother(
-        simulations_argument,
+    return EnsembleExperiment(
+        EnsembleExperimentRunArguments(
+            active_realizations=active_realizations,
+            current_case=args.current_case,
+            iter_num=int(args.iter_num),
+        ),
         ert,
         storage,
         ert.get_queue_config(),
         experiment_id,
     )
-    return model
+
+
+def _setup_ensemble_smoother(
+    ert: EnKFMain, storage: StorageAccessor, args: Namespace, experiment_id: UUID
+) -> EnsembleSmoother:
+    return EnsembleSmoother(
+        ESRunArguments(
+            active_realizations=_realizations(args, ert.getEnsembleSize()),
+            current_case=args.current_case,
+            target_case=_target_case_name(ert, args, format_mode=False),
+        ),
+        ert,
+        storage,
+        ert.get_queue_config(),
+        experiment_id,
+    )
 
 
 def _setup_multiple_data_assimilation(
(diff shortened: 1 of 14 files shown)


src/ert/run_models/run_arguments.py
```diff
@@ -0,0 +1,69 @@
+from dataclasses import dataclass, field
+from typing import List, Optional
+
+
+@dataclass
+class SingleTestRunArguments:
+    simulation_mode: str = field(init=False, default="Single test run")
```
pinkwah, line 7: Unused afaik, but:

```suggestion
    simulation_mode: Literal["Single test run"] = "Single test run"
```

This means that the value can only ever be the literal value. It's also useful in Pydantic (for future reference):

```py
class Foo(BaseModel):
  kind: Literal["Foo"] = "Foo"
  intvalue: int

class Bar(BaseModel):
  kind: Literal["Bar"] = "Bar"
  strvalue: str

class MyData(BaseModel):
  foo_or_bar: Union[Foo, Bar]

MyData.parse_raw('{"foo_or_bar": {"kind": "Foo", "intvalue": 42}}')  #=> OK
MyData.parse_raw('{"foo_or_bar": {"kind": "Bar", "strvalue": "some_str"}}')  #=> OK
MyData.parse_raw('{"foo_or_bar": {"kind": "Foo", "strvalue": "some_str"}}')  #=> Error! `Foo` does not have `strvalue`
```

Eivind says that we can use `from pydantic.dataclasses import dataclass` instead of `from pydantic import BaseModel` to have the same API as the built-in `dataclass`, but using Pydantic features like validators. I haven't used this so I leave it as an excercise to the reader.

src/ert/run_models/run_arguments.py
```diff
@@ -0,0 +1,69 @@
...
+
+    active_realizations: List[bool]
+    current_case: str
+    target_case: str
+    iter_num: Optional[int] = 0
+    item_number: Optional[int] = 0
```
pinkwah, line 17: I think this was a typo on my part and was supposed to be `iter_num`. (Which is why it's nice to have mypy to avoid dumb typos).

src/ert/run_models/run_arguments.py
```diff
@@ -0,0 +1,69 @@
+from dataclasses import dataclass, field
+from typing import List, Optional
+
+
+@dataclass
+class SingleTestRunArguments:
+    simulation_mode: str = field(init=False, default="Single test run")
+
+    active_realizations: List[bool]
```
pinkwah, line 9: I hate the prefix "active". Everything is "active" is not informative. Maybe just `realizations` or `realizations_mask`?

src/ert/run_models/run_arguments.py
```diff
@@ -0,0 +1,69 @@
...
+    current_case: str
```
pinkwah, line 10:
```suggestion
    experiment_name: str
```

src/ert/run_models/run_arguments.py
```diff
@@ -0,0 +1,69 @@
...
+    target_case: str
```
pinkwah, line 11:
```suggestion
    ensemble_name_format: str
```
(Unsure about this)

src/ert/run_models/run_arguments.py
```diff
@@ -0,0 +1,69 @@
...
+    iter_num: Optional[int] = 0
```
pinkwah, line 16:
```suggestion
    override_iteration_number: Optional[int] = None
```

This is used for explicitly setting the iteration number, regardless of what it should actually be. Really bad behaviour imo but we can be more explicit on what it does.

src/ert/run_models/run_arguments.py
```diff
@@ -0,0 +1,69 @@
...
+    item_number: Optional[int] = 0
+    prev_successful_realizations: Optional[int] = 0
```
pinkwah: I am anal about mixing short and long forms of words. Also, the "previous" makes no sense to me? What if this were `successful_realizations`?

src/ert/run_models/run_arguments.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@dataclass
+class ESMDARunArguments:
+    simulation_mode: str = field(init=False, default="Multiple data assimilation")
+    analysis_module: str = field(init=False, default="STD_ENKF")
+
+    active_realizations: List[bool]
+    target_case: str
+    weights: str
```
pinkwah, line 54: Instead of using `str` here, you can have the type be `Sequence[float]` and have a Pydantic validator convert from `str`. Move the code from `MultipleDataAssimilation.parseWeights` here as a validator.

This makes it so that the correctness of `weights` is an invariant, rather something that the user code has to remember to call.

src/ert/run_models/run_arguments.py
dafeda: Did not know this was unused, and thanks for the trick 👍

src/ert/run_models/run_arguments.py
dafeda: I agree that `active` is not great, but I don't think we should make this change in this PR.

src/ert/run_models/run_arguments.py
dafeda: Agree that it's not the best name, but this discussion is best left for another PR I think.

src/ert/run_models/run_arguments.py
dafeda: In this PR, I want to make the smallest change possible that replaces `Dict[str, Any]` with dataclasses for simulation arguments.

src/ert/run_models/ensemble_smoother.py
```diff
@@ -40,7 +43,7 @@ def __init__(
             experiment_id,
             phase_count=2,
         )
-        self._current_case_name: str = simulation_arguments["current_case"]
+        self._current_case_name: str = simulation_arguments.current_case
```
oyvindeide, line 46: Not used as far as I can tell

oyvindeide (commented):
Looks like a good improvement 👍

src/ert/cli/model_factory.py
```diff
@@ -62,13 +69,15 @@ def create_model(
...
-    simulations_argument = {
-        "active_realizations": [True],
-        "current_case": args.current_case,
-        "simulation_mode": "Single test run",
-    }
-    model = SingleTestRun(simulations_argument, ert, storage, experiment_id)
-    return model
+    return SingleTestRun(
+        SingleTestRunArguments(
+            active_realizations=[True],
```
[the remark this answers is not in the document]
dafeda: Not sure I understand what you mean.
`SingleTestRunArguments` does not have an `analysis_module`.
Did you perhaps mean something else?

src/ert/cli/model_factory.py
```diff
@@ -134,42 +139,38 @@ def _setup_multiple_data_assimilation(
...
-        "simulation_mode": "Multiple data assimilation",
-    }
-    model = MultipleDataAssimilation(
-        simulations_argument,
+    return MultipleDataAssimilation(
+        ESMDARunArguments(
+            active_realizations=_realizations(args, ert.getEnsembleSize()),
+            target_case=_target_case_name(ert, args, format_mode=True),
+            weights=args.weights,
+            num_iterations=len(args.weights),
```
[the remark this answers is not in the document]
dafeda: Not sure I understand how these could be part in post_init without passing `args` and `ert` to the dataclass.

src/ert/run_models/ensemble_smoother.py
dafeda: Thanks, fixed 👍

src/ert/cli/model_factory.py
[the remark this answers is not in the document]
oyvindeide: Yes, meant that you could put: `active_realizations` in `__post_init__` of `SingleTestRunArguments` as it is constant.

src/ert/cli/model_factory.py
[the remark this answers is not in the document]
oyvindeide: `ESMDARunArguments` has `self.weights`, so could calculate `self.num_iterations`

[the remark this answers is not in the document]
dafeda: That makes sense, thanks 👍

src/ert/run_models/run_arguments.py
```diff
@@ -0,0 +1,81 @@
+from dataclasses import dataclass
+from typing import List, Optional, Union
+
+
+@dataclass
+class SingleTestRunArguments:
+    active_realizations: List[bool]
```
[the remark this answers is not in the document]
dafeda: Agree 👍

dafeda:
I have included your suggestions in the lates push @oyvindeide . Sorry for force pushing 😬

oyvindeide (approved):
LGTM! Nice improvement!

PR #6339, opened 2023-10-16
Compute cross-correlation matrices without matrix inversion

### LINKED ISSUE #6330: Compute cross-correlation matrices without matrix inversion
It is not necessary to form `C_AA` or `Sigma_A` and call` np.linalg.inv(Sigma_A)`  and `np.linalg.inv(Sigma_Y)` instead we could form the matrix of inverse SD's `1/(sd_x*sd_y)` to array multiply with `C_AY` . We then avoid some very expensive matrix operations, in particular the inversions that are `O(n^3)`, and we would have a great very many of. 

https://github.com/equinor/ert/blob/20cfcdefd1b19124a89f347ae6c8afae2d71896e/src/ert/analysis/_es_update.py#L477-L490

Use @tommyod comment https://github.com/equinor/ert/pull/4243#issuecomment-1738643115 to do this in a nice numpy-way

Blunde1:
**Issue**
Resolves #6330 

**Approach**
Use @tommyod approach https://github.com/equinor/ert/pull/4243#issuecomment-1738643115

### DIFF
--- a/src/ert/analysis/_es_update.py
+++ b/src/ert/analysis/_es_update.py
@@ -444,11 +444,10 @@ def analysis_ES(
             temp_storage = _create_temporary_parameter_storage(
                 source, iens_active_index, param_group.name
             )
-
             if module.localization():
                 Y_prime = S - S.mean(axis=1, keepdims=True)
                 C_YY = Y_prime @ Y_prime.T / (ensemble_size - 1)
-                Sigma_Y = np.diag(np.sqrt(np.diag(C_YY)))
+                Sigma_Y = np.std(S, axis=1, ddof=1)
                 batch_size: int = 1000
                 correlation_threshold = module.localization_correlation_threshold(
                     ensemble_size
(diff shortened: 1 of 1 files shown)


src/ert/analysis/_es_update.py
```diff
@@ -476,17 +473,18 @@ def analysis_ES(
                 batches = _split_by_batchsize(np.arange(0, num_params), batch_size)
                 for param_batch_idx in tqdm(batches):
                     X_local = temp_storage[parameter.name][param_batch_idx, :]
-                    A = X_local - X_local.mean(axis=1, keepdims=True)
-                    C_AA = A @ A.T / (ensemble_size - 1)
+
+                    # Standard deviations
+                    Sigma_Y = np.std(S, axis=1, ddof=1)
```
dafeda, line 478: I think this can be moved outside of the `for param_batch_idx` loop.

src/ert/analysis/_es_update.py
```diff
@@ -476,17 +473,18 @@ def analysis_ES(
...
+                    Sigma_A = np.std(X_local, axis=1, ddof=1)
 
                     # State-measurement covariance matrix
+                    Y_prime = S - S.mean(axis=1, keepdims=True)
```
dafeda, line 482: I think this can also be moved outside of the loop.

src/ert/config/analysis_module.py
```diff
@@ -33,6 +34,23 @@ class VariableInfo(TypedDict):
...
+
+def correlation_threshold(ensemble_size: int, user_defined_threshold: float) -> float:
+    """Decides whether or not to use user-defined or default threshold.
+
+    Default threshold taken from luo2022,
+    Continuous Hyper-parameter OPtimization (CHOP) in an ensemble Kalman filter
+    Section 2.3 - Localization in the CHOP problem
+    """
+    default_threshold = 3 / math.sqrt(ensemble_size)
+    if user_defined_threshold == -1:
```
tommyod, line 50: Please don't compare floats using `==`.

------------

In fact, I think using `None` is better here. More typical Python, more readable:

```python
if user_defined_threshold is None:
    return default_threshold
```

------------

Come to think of it, I believe the (1) decision to use the default or user defined threshold and (2) retrieving the default threshold are two separate concerns. I think this is better:

```python
threshold = # either a float in [0, 1], or None
self.threshold = correlation_threshold(ensemble_size) if threshold is None else threshold
```
In other words, remove `user_defined_threshold` entirely. The decision is already taken before the function is called (by setting it to `-1` / `None`), so no need to propagate the data down into the function.

tommyod (commented):
Small comment on a small function. Take it or leave it :)

src/ert/config/analysis_module.py
dafeda: I agree with this.
Could you make an issue with this?

dafeda (approved):
Nice 👍

PR #6372, opened 2023-10-18
Remove util string copy

eivindjahren:
## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/clib/lib/job_queue/lsf_driver.cpp
+++ b/src/clib/lib/job_queue/lsf_driver.cpp
@@ -224,7 +224,7 @@ static void lsf_driver_internal_error() {
 char *
 alloc_composed_resource_request(const lsf_driver_type *driver,
                                 const std::vector<std::string> &select_list) {
-    char *resreq = util_alloc_string_copy(driver->resource_request);
+    char *resreq = driver->resource_request;
     std::string excludes_string = ert::join(select_list, " && ");
 
     char *req = NULL;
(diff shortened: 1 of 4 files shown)


src/clib/lib/job_queue/lsf_driver.cpp
```diff
@@ -291,11 +303,11 @@ static char *alloc_quoted_resource_string(const lsf_driver_type *driver) {
 
     char *quoted_resource_request = NULL;
     if (req) {
-        if (driver->submit_method == LSF_SUBMIT_REMOTE_SHELL)
+        if (driver->submit_method == LSF_SUBMIT_REMOTE_SHELL) {
             quoted_resource_request = util_alloc_sprintf("\"%s\"", req);
-        else
-            quoted_resource_request = util_alloc_string_copy(req);
-        free(req);
+            free(req);
```
andreas-el, line 308: You free the char* only if `LSF_SUBMIT_REMOTE_SHELL`, but on line 269 you free the char* regardless of use.
Is this intentional?

andreas-el (changes requested):
Had a question

src/clib/lib/job_queue/lsf_driver.cpp
eivindjahren: Yes, this means that req is stolen to quoted_resource_request instead of having a copy and then free the old version which was the old behavior.

PR #6395, opened 2023-10-23
Test property generalized variance increases in cutoff

### LINKED ISSUE #6224: Test that the generalized variance of the adaptive localization resulting ensemble falls between ES and prior

Blunde1:
**Issue**
Resolves #6224 #6225

### DIFF
--- a/tests/unit_tests/analysis/test_adaptive_localization.py
+++ b/tests/unit_tests/analysis/test_adaptive_localization.py
@@ -10,6 +10,8 @@ from ert.cli.main import run_cli
 from ert.config import ErtConfig
 from ert.storage import open_storage
 
+random_seed_line = "RANDOM_SEED 1234\n\n"
+
 
 def run_cli_ES_with_case(poly_config):
     config_name = poly_config.split(".")[0]
@@ -42,9 +44,7 @@ def run_cli_ES_with_case(poly_config):
 
 
 @pytest.mark.integration_test
-def test_that_adaptive_localization_with_cutoff_1_equals_ensemble_prior(copy_case):
-    copy_case("poly_example")
-    random_seed_line = "RANDOM_SEED 1234\n\n"
+def test_that_adaptive_localization_with_cutoff_1_equals_ensemble_prior(copy_poly_case):
     set_adaptive_localization_1 = dedent(
         """
         ANALYSIS_SET_VAR STD_ENKF LOCALIZATION True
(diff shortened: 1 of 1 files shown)


Blunde1:
@eivindjahren the tests seem to work, but hypothesis runs into issues, likely due to running ert many times? Is there any way around this, (except refactor the code, which we will indeed do), or should I avoid hypothesis for this right now? You can see my setup in commented out lines of code.

@pinkwah is there any way around the 
`prior_cov = pd.DataFrame(prior_sample_c1.values).cov().to_numpy()`
where we have used storage
```
 storage_path = ErtConfig.from_file(poly_config).ens_path
    with open_storage(storage_path) as storage:
        prior_ensemble = storage.get_ensemble_by_name(prior_sample_name)
        prior_sample = prior_ensemble.load_parameters("COEFFS")
```
for some other code I see a different (old?) usage of storage that directly provides a dataframe. A dataframe is what I want, so I may run `.cov()`.
https://github.com/equinor/ert/blob/3f657a88839c5702d4675ba176f68de8b5600e54/tests/unit_tests/analysis/test_es_update.py#L193-L206

eivindjahren:
Ah, I think I know what the problem is. You are copying in the polycase into the test, but when you run this with hypothesis, there is no automatic cleanup. In order to fix this you can probably use [tmp_path_factory](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmp-path-factory-fixture):

```python
def test_that_posterior_generalized_variance_increases_in_cutoff(tmp_path_factory, source_root, c1_and_c2):
        c1, c2 = c1_and_c2
        tmp_path = tmp_path_factory.mktemp("poly_case")
        shutil.copytree(os.path.join(source_root, "test-data", "poly_case"), tmp_path)
        monkeypatch.chdir(tmp_path)
```

We should probably change the copy_case fixture anyways so that it guarantees a new directory every time we call it: 
https://github.com/equinor/ert/pull/6396

kvashchuka:
Even with https://github.com/equinor/ert/pull/6396, we run into problem
```
TimeoutError: Not able to acquire lock for: /private/var/folders/nj/5n4_0y3d68bdhxrjtl5p_qd00000gp/T/pytest-of-AKVAS/pytest-4/poly_example0/test_data/storage. 
You may already be running ERT, or another user is using the same ENSPATH.
```
Therefore, I think we can use a pytest parametrization for this test, and hypothesis when and if we refactor the code in a way described in https://github.com/equinor/ert/issues/6334

tests/unit_tests/analysis/test_adaptive_localization.py
```diff
@@ -102,3 +101,63 @@ def test_that_adaptive_localization_with_cutoff_0_equals_ESupdate(copy_case):
...
+        f.writelines(lines)
+
+    prior_sample_cutoff1, posterior_sample_cutoff1 = run_cli_ES_with_case(
+        "poly_localization_cutoff1.ert"
+    )
+    prior_sample_cutoff2, posterior_sample_cutoff2 = run_cli_ES_with_case(
+        "poly_localization_cutoff2.ert"
+    )
+
+    prior_cov = pd.DataFrame(prior_sample_cutoff1.values).cov().to_numpy()
```
tommyod, line 153: I would personally just use [numpy.cov](https://numpy.org/doc/stable/reference/generated/numpy.cov.html). No reason to introduce a DataFrame and leave numpy-land if all you need is the covariance.

tommyod (commented):
Small comment for consideration :)

kvashchuka (approved):

PR #6392, opened 2023-10-23
Replace custom debug options with logging

eivindjahren:
Now puts output into jobqueue-log.txt:
```
...
2023-10-30 07:42:59,032 - ert.job_queue.torque_driver - Thread-25 (_job_monitor) - DEBUG - Submit arguments: qsub -k oe -l nodes=1:ppn=1 -N SNAKE_OIL_FIELD -r n .../snake_oil/storage/snake_oil/runpath/realization-21/iter-0/qsub_script.sh 
2023-10-30 07:42:59,033 - ert.job_queue.torque_driver - Thread-27 (_job_monitor) - DEBUG - Submit arguments: qsub -k oe -l nodes=1:ppn=1 -N SNAKE_OIL_FIELD -r n .../snake_oil/storage/snake_oil/runpath/realization-23/iter-0/qsub_script.sh 
...
```

## Pre review checklist

- [x] Read through the code changes carefully after finishing work
- [x] Make sure tests pass locally (after every commit!)
- [x] Prepare changes in small commits for more convenient review (optional)
- [x] PR title captures the intent of the changes, and is fitting for release notes.
- [x] Updated documentation
- [x] Ensured that unit tests are added for all new behavior (See 
    [Ground Rules](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md#ground-rules)),
    and changes to existing code have good test coverage.

## Pre merge checklist
- [x] Added appropriate release note label
- [x] Commit history is consistent and clean, in line with the [contribution guidelines](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md).

PR #6413, opened 2023-10-24
Add queue option validation to queue config

### LINKED ISSUE #6391: Failed conversion of queue driver option values are silently ignored
In the following `self.set_option` will return False if the option string failed to convert to int/bool/float etc. However, this is never checked and the option is silently not set.

https://github.com/equinor/ert/blob/8c54c23329e169bd4cda584bf61a5fa85cde83a4/src/ert/job_queue/driver.py#L55

Also, this conversion happens when the `start simulations` button is pressed. We would like to signal such a failure as soon as the application starts so that it can be shown in the `problems detected` window.

xjules:
**Issue**
Resolves #6391 

**Approach**
Add validation of option types to each driver.

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist


src/ert/config/queue_config.py
```diff
@@ -139,17 +113,130 @@ def generate_dict(option_list: List[Tuple[str, str]]) -> Dict[str, List[str]]:
             )
 
 
-def _validate_torque_options(torque_options: List[Tuple[str, str]]) -> None:
-    for option_strings in torque_options:
+memory_options: Mapping[str, List[str]] = {
```
berland, line 116: Hmm, should we not categorize either by data format (string/float/int) or by topic, not by both?

berland: I see, memory has its own syntax. Should we not call this a string option, and then apply an extra layer of validation after the format of the value has been accepted?

berland: I suggest adding a map from options to regular expressions that are also verified in a loop, and include only memory_per_job as of now.

src/ert/config/queue_config.py
```diff
@@ -139,17 +113,130 @@ def generate_dict(option_list: List[Tuple[str, str]]) -> Dict[str, List[str]]:
...
+        if option_name in string_options[queue_type] and not isinstance(
+            option_value, str
+        ):
+            raise ConfigValidationError(
+                f"The value '{option_value}' is not valid string "
+                f"for the {queue_type} option "
+                f"{option_name}, it must be of string type."
+            )
+        if option_name in float_options[queue_type] and (
+            re.match(r"^\d+(\.\d+)?$", option_value) is None and option_value != ""
```
berland, line 221: That means we disallow leading zeros. Maybe ok.

berland: What about negative numbers?

src/ert/config/queue_config.py
```diff
@@ -139,17 +113,130 @@ def generate_dict(option_list: List[Tuple[str, str]]) -> Dict[str, List[str]]:
...
+            re.match(r"^\d+$", option_value) is None and option_value != ""
+        ):
+            raise ConfigValidationError(
+                f"The value '{option_value}' is not valid positive integer "
+                f" for the {queue_type} option {option_name}."
+            )
+        if option_name in bool_options[queue_type] and not option_value in [
+            "TRUE",
+            "FALSE",
+            "",
```
berland, line 237: My gut tells me that we should perhaps allow lower-case values of this. Maybe also `0` and `1`. Yes/no?

src/ert/config/queue_config.py
```diff
@@ -139,17 +113,130 @@ def generate_dict(option_list: List[Tuple[str, str]]) -> Dict[str, List[str]]:
...
+}
+
+
+def _validate_queue_options(
+    queue_options: List[Tuple[str, str]], queue_type: str
+) -> None:
+    for option_strings in queue_options:
         option_name = option_strings[0]
         option_value = option_strings[1]
         if (
```
berland, line 201: will it make sense to do `if option_value == "": continue` here to make the conditionals below shorter?

tests/unit_tests/config/test_ert_config.py
```diff
@@ -744,11 +744,11 @@ def test_that_queue_config_dict_negative_value_invalid(
     with config_generator(tmp_path_factory) as config_values:
         config_dict = config_values.to_config_dict("test.ert", os.getcwd())
         config_dict[ConfigKeys.QUEUE_OPTION].append(
-            ["LSF", "MAX_RUNNING", "-6"],
+            [config_dict[ConfigKeys.QUEUE_SYSTEM], "MAX_RUNNING", "-6"],
         )
         with pytest.raises(
             expected_exception=ConfigValidationError,
-            match="QUEUE_OPTION MAX_RUNNING is negative",
+            match="not valid positive integer",
```
berland, line 751: will the string `MAX_RUNNING` be presented to the user still?

berland:
If you expose the `string_options` in `queue_config.py` to `tests/unit_tests/config/config_dict_generator.py` I think you can make the tests quite smart.

src/ert/config/queue_config.py
xjules: yes, makes sense!

src/ert/config/queue_config.py
```diff
@@ -139,17 +113,130 @@ def generate_dict(option_list: List[Tuple[str, str]]) -> Dict[str, List[str]]:
...
+    ],
+    "SLURM": [
+        "SQUEUE_TIMEOUT",
+    ],
+    "TORQUE": ["SUBMIT_SLEEP", "QUEUE_QUERY_TIMEOUT"],
+    "LOCAL": [],
+}
+
+bool_options: Mapping[str, List[str]] = {
+    "LSF": ["DEBUG_OUTPUT"],
```
larsevj, line 188: What is the reason for `LSF DEBUG_OUTPUT` being validated both as bool and as string, whereas `TORQUE DEBUG_OUTPUT` is only validated as string?

src/ert/config/queue_config.py
```diff
@@ -139,17 +113,130 @@ def generate_dict(option_list: List[Tuple[str, str]]) -> Dict[str, List[str]]:
...
 
-def _validate_torque_options(torque_options: List[Tuple[str, str]]) -> None:
-    for option_strings in torque_options:
+memory_options: Mapping[str, List[str]] = {
+    "LSF": [],
+    "SLURM": [],
+    "TORQUE": ["MEMORY_PER_JOB"],
+    "LOCAL": [],
+}
+
```
larsevj, line 122: I can not seem to get this list of queue options that is validated to match with the list of `VALID_QUEUE_OPTIONS` found at the start of this file. Why is this?

src/ert/config/queue_config.py
```diff
@@ -139,17 +113,130 @@ def generate_dict(option_list: List[Tuple[str, str]]) -> Dict[str, List[str]]:
...
+bool_options: Mapping[str, List[str]] = {
+    "LSF": ["DEBUG_OUTPUT"],
+    "SLURM": [],
+    "TORQUE": [],
+    "LOCAL": [],
+}
+
+
+def _validate_queue_options(
+    queue_options: List[Tuple[str, str]], queue_type: str
```
larsevj, line 196: Maybe consider another name for this, since this is not the same as the class object (dictionary) `queue_options`?

larsevj (changes requested):

src/ert/config/queue_config.py
berland: NB: This will change in https://github.com/equinor/ert/pull/6392, the options will just go away and we make a debug file in any case.

src/ert/config/queue_config.py
xjules: Do we need negative numbers?

src/ert/config/queue_config.py
xjules: Union of all  these lists should `VALID_QUEUE_OPTIONS`. So the regex checks is essentially an extra verification layer on top of `VALID_QUEUE_OPTIONS`

src/ert/config/queue_config.py
xjules: I've gone for `_validate_queue_driver_settings`?

larsevj: I should probably have formulated it better, apologies for that. My concern was about the variable name `queue_options` not necessarily the function name.

src/ert/config/queue_config.py
xjules: @larsevj you can here for more details: https://fmu-docs.equinor.com/docs/ert/reference/configuration/keywords.html#torque-debug-output
It is different for torque and lsf at the moment.

src/ert/config/queue_config.py
xjules: Renamed to `positive_number_options`.

src/ert/config/queue_config.py
xjules: For better code understanding I've decided to keep it explicit.

src/ert/config/queue_config.py
xjules: Hm, it's part of the docs and can be found in the code: https://github.com/equinor/ert/blob/39b7d047e5de932092251fd44a9fc212ed1460b7/src/clib/lib/job_queue/torque_driver.cpp#L210

src/ert/config/queue_config.py
```diff
@@ -139,17 +113,139 @@ def generate_dict(option_list: List[Tuple[str, str]]) -> Dict[str, List[str]]:
...
+        "INCLUDE_HOST",
+        "EXCLUDE_HOST",
+    ],
+    "TORQUE": [
+        "QSUB_CMD",
+        "QSTAT_CMD",
+        "QDEL_CMD",
+        "QSTAT_OPTIONS",
+        "QUEUE",
+        "DEBUG_OUTPUT",
```
berland, line 152: DEBUG_OUTPUT is now deprecated as #6392 is merged.

src/ert/config/queue_config.py
```diff
@@ -139,17 +113,139 @@ def generate_dict(option_list: List[Tuple[str, str]]) -> Dict[str, List[str]]:
...
+        if option_name in queue_bool_options[queue_type] and not option_value in [
+            "TRUE",
+            "FALSE",
+            "0",
+            "1",
+            "T",
+            "F",
+        ]:
+            raise ConfigValidationError.with_context(
+                f"The '{option_value}' for {option_name} must be one of TRUE, FALSE, T, F, 1, 0 "
```
berland, line 248: Hmm, do we want to not hint the users about every option here? Maybe only say TRUE or FALSE, in case we want to cleanup later

berland: and if so, relax 'must' to 'should'.

src/ert/config/queue_config.py
```diff
@@ -139,17 +113,139 @@ def generate_dict(option_list: List[Tuple[str, str]]) -> Dict[str, List[str]]:
...
+                f" for the {queue_type}.",
```
berland, line 249: This is probably independent of queue type?

src/ert/job_queue/driver.py
```diff
@@ -52,7 +52,8 @@ def create_driver(cls, queue_config: QueueConfig) -> "Driver":
         driver = Driver(queue_config.queue_system)
         if queue_config.queue_system in queue_config.queue_options:
             for setting in queue_config.queue_options[queue_config.queue_system]:
-                driver.set_option(*setting)
+                if not driver.set_option(*setting):
+                    raise ValueError(f"Queue option not set {setting}")
```
berland, line 56: Can we in some way blow up something with this? 🤔

tests/unit_tests/config/test_ert_config.py
```diff
@@ -707,8 +707,8 @@ def test_default_ens_path(tmpdir):
     "max_running_value, expected_error",
     [
         (100, None),  # positive integer
-        (-1, "QUEUE_OPTION MAX_RUNNING is negative"),  # negative integer
-        ("not_an_integer", "QUEUE_OPTION MAX_RUNNING is not an integer"),  # non-integer
+        (-1, "is not a valid positive integer"),  # negative integer
```
berland, line 710: Should we not assert that the string MAX_RUNNING is part of the error? (via regexp if needed)

berland: ah, tested further below. Do we need this test here then?

src/ert/job_queue/driver.py
xjules: That's the whole point I think at least, that those errors should be caught before we hit start simulation.

src/ert/config/queue_config.py
xjules: Not really. There are some, which might be different for each type; eg. DEBUG_OUTPUT

src/ert/config/queue_config.py
xjules: Right, that I guess I can revert the comment bellow :)

src/ert/job_queue/driver.py
berland: Me thinking there are rogue ERT configs out in the wild. Hopefully the user will get warnings from the new validation part before we hit this. Maybe we are quite sure on that part. I assume it is impossible to reach this ValueError if we go through the validated config, and only if we have rogue self-written application utilizing the c-driver?

berland:
Noting that this PR will allow f.ex. `QUEUE_OPTION LOCAL MAX_RUNNING fooo` if the queue system is set to Torque. Ert-main will not allow fooo here.

src/ert/config/queue_config.py
xjules: I needed to put DEBUG_OUTPUT back until it is completely removed.

src/ert/config/queue_config.py
xjules: done

tests/unit_tests/config/test_ert_config.py
xjules: Not sure, maybe we can skip it?

berland: The code below lacks testing of "-1". The happy path ("100") is probably tested within the hypothesis setup I suppose.

xjules:
> Noting that this PR will allow f.ex. `QUEUE_OPTION LOCAL MAX_RUNNING fooo` if the queue system is set to Torque. Ert-main will not allow fooo here.

I see that as up discussion whether we allow some general configs for instance related to other queues (LOCAL) to be set when using TORQUE for instance.

src/ert/job_queue/driver.py
xjules: I see your point point. If i get your point right, instead of raising an error we should provide a a warning? The problem with warning that this might get ignored :)

berland:
> I see that as up discussion whether we allow some general configs for instance related to other queues (LOCAL) to be set when using TORQUE for instance.

Should be possible to issue `ConfigWarning` for errors in not-selected queues.

src/ert/config/queue_config.py
```diff
@@ -62,8 +62,8 @@ def from_dict(cls, config_dict: ConfigDict) -> QueueConfig:
             if queue_system != selected_queue_system:
                 warnings.warn(
                     ConfigWarning.with_context(
-                        f"Setting '{option_name}'for {queue_system.name} but "
-                        f"{selected_queue_system.name} is selected. ",
+                        f"Setting '{option_name}' for {queue_system.name} but "
+                        f"{selected_queue_system.name} is selected.",
```
berland, line 66: Will this give warnings if I have valid Torque queue options but Local queue is selected? I don't think I want that, I only want warnings if I have invalid Torque queue options and local is selected.

xjules: I think we should be validated only when the queue is chosen otherwise just give a warning that you are setting something that is not in used anyway.

berland: Nah, I think it must to be possible to have ert config files that can be used both on LSF and Torque without warnings.

xjules: Now we raise warnings if different queue is set wrongly.

src/ert/config/queue_config.py
```diff
@@ -95,11 +73,14 @@ def from_dict(cls, config_dict: ConfigDict) -> QueueConfig:
                     " usually provided by the site-configuration file, beware that"
                     " you are effectively replacing the default value provided."
                 )
-        if (
-            selected_queue_system == QueueSystem.TORQUE
-            and queue_options[QueueSystem.TORQUE]
-        ):
-            _validate_torque_options(queue_options[QueueSystem.TORQUE])
+
+        for queue_system_val in used_queue_systems:
```
berland, line 77: Can we instead of `used_queue_systems` do  something like `queue_options.keys()`?

src/ert/config/queue_config.py
```diff
@@ -139,17 +120,166 @@ def generate_dict(option_list: List[Tuple[str, str]]) -> Dict[str, List[str]]:
             )
 
 
-def _validate_torque_options(torque_options: List[Tuple[str, str]]) -> None:
-    for option_strings in torque_options:
+queue_memory_options: Mapping[str, List[str]] = {
+    "LSF": [],
```
berland, line 124: Is it more consistent to use `QueueSystem.LSF` instead of the string `LSF`?

berland (changes requested):

src/ert/config/queue_config.py
xjules: I had that before, but I changed it due to know these `*_options` are exposed in tests and didn't want to import all the Cwrap classes there.

berland: ok!

berland (approved):
Lets go!

larsevj (approved):

PR #6414, opened 2023-10-24
Test torque driver through slurm

berland:
The slurm-wlm-torque adds a compatibility layer so that there is a qsub/qstat/qdel shell API for ERTs torque driver to use.

This commit will set it up and run the poly-case through it.

**Issue**
Resolves #my_issue

**Approach**

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist


.github/workflows/test_ert_with_slurm.yml
```diff
@@ -71,6 +71,36 @@ jobs:
...
+        EOF
+
+        time ert ensemble_experiment poly.ert
+
+
+    - name: Test Torque driver through slurm
+      run: |
+        set -e
+
+        sudo apt install slurm-wlm-torque
```
xjules, line 89: what's `slurm-wlm-torque`?

berland: That is the compatibility layer so that a Slurm cluster can expose a Torque shell api

.github/workflows/test_ert_with_slurm.yml
```diff
@@ -71,6 +71,36 @@ jobs:
...
+
+        # Patch qstat from slurm-wlm-torque (patch is submitted upstream):
+        cp /usr/bin/qstat qstat
+        # shellcheck disable=SC2016
```
xjules, line 93: very good! didn't know about this :)

xjules:
Looks good, but something is off since tests are failing.

.github/workflows/test_ert_with_slurm.yml
```diff
@@ -71,6 +71,36 @@ jobs:
...
+        sed -i '516 i printf("    Exit_status = %s\\n", $job->{"exit_code"});' qstat
+        cp -a test-data/poly_example poly_torque
+
+        cat << EOF >> poly_torque/poly.ert
+        NUM_REALIZATIONS 5
+        QUEUE_SYSTEM TORQUE
+        QUEUE_OPTION TORQUE QSTAT_CMD $(pwd)/qstat
+        QUEUE_OPTION TORQUE QSTAT_OPTIONS -f
```
xjules, line 101: I get the `QSTAT_CMD`, but is `-f ` required by SLURM?

berland: Yes, it means "full output" which is the only thing our parser can parse. We need to set it, because the default is "-f -x", while the "-x" option is unsupported in slurm-wlm-torque (thus we should not have it as default in the driver, but in ert-configurations)

berland: added comment on qstat_options and keep_qsub_output

xjules: nice!

xjules (approved):
Nice job! Let's slurm :)

PR #6428, opened 2023-10-25
Fix Ert GUI crashes on esc press

### LINKED ISSUE #6336: Pressing ESC in when running experiment will crash Ert
Pressing ESC in rundialog will exit the window, but Ert will crash with;

```
QThread: Destroyed while thread is still running
[1]    40034 abort      ert gui test-data/poly_example/poly.ert
```

upon completing the ensemble.

larsevj:
**Issue**
Resolves #6336 

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist


src/ert/gui/simulation/run_dialog.py
```diff
@@ -401,3 +401,10 @@ def toggle_detailed_progress(self):
     def _on_finished(self):
         for file_dialog in self.findChildren(FileDialog):
             file_dialog.close()
+
+    def keyPressEvent(self, q_key_event):
+        if q_key_event.key() == Qt.Key_Escape:
+            # self.kill_button.click()
+            self.close()
+        else:
```
larsevj, line 408: Unsure what to do with the other key presses? Is there any other uses for them? Pressing enter will select the "blue" button at least.

andreas-el: Usually there is a selected button/element that would be triggered if ENTER-key was pressed. I guess that is reasonable as-is today. What seems to be missing is the possibility to tab between the elements. There is usually a predetermined sequence that allows users to tab between a selected set of elements.
Regarding the enter key though, id argue that passing on the signal (like you do) makes most sense.

andreas-el (approved):
Squash those commits, and then we're good to go!
Good work! 🚀

Issue #6218, opened 2023-10-02
Remove Stage, Step and Job from ensemble_evaluator._builder


pinkwah:
`Job` is just a container for `ert.config.ext_job.ExtJob`, so let `Realization` contain a list of `ExtJob`s maybe.

The other parts have no value.

berland:
Rougly 600 lines less code from the PRs solving this issue.

PR #6430, opened 2023-10-26
Remove step states from snapshots

berland:
The step state is moved upwards to the realization.

**Issue**
Partially solves #6218 

**Approach**
Cut code,  and merge datastructure.

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/ensemble_evaluator/identifiers.py
+++ b/src/ert/ensemble_evaluator/identifiers.py
@@ -17,13 +17,13 @@ STDERR = "stderr"
 STDOUT = "stdout"
 STEPS = "steps"
 
-EVTYPE_FM_STEP_FAILURE = "com.equinor.ert.forward_model_step.failure"
-EVTYPE_FM_STEP_PENDING = "com.equinor.ert.forward_model_step.pending"
-EVTYPE_FM_STEP_RUNNING = "com.equinor.ert.forward_model_step.running"
-EVTYPE_FM_STEP_SUCCESS = "com.equinor.ert.forward_model_step.success"
-EVTYPE_FM_STEP_UNKNOWN = "com.equinor.ert.forward_model_step.unknown"
-EVTYPE_FM_STEP_WAITING = "com.equinor.ert.forward_model_step.waiting"
-EVTYPE_FM_STEP_TIMEOUT = "com.equinor.ert.forward_model_step.timeout"
+EVTYPE_REALIZATION_FAILURE = "com.equinor.ert.realization.failure"
+EVTYPE_REALIZATION_PENDING = "com.equinor.ert.realization.pending"
+EVTYPE_REALIZATION_RUNNING = "com.equinor.ert.realization.running"
+EVTYPE_REALIZATION_SUCCESS = "com.equinor.ert.realization.success"
+EVTYPE_REALIZATION_UNKNOWN = "com.equinor.ert.realization.unknown"
+EVTYPE_REALIZATION_WAITING = "com.equinor.ert.realization.waiting"
+EVTYPE_REALIZATION_TIMEOUT = "com.equinor.ert.realization.timeout"
 
 EVTYPE_FM_JOB_START = "com.equinor.ert.forward_model_job.start"
 EVTYPE_FM_JOB_RUNNING = "com.equinor.ert.forward_model_job.running"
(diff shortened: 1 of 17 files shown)


src/ert/ensemble_evaluator/identifiers.py
```diff
@@ -17,28 +17,28 @@
...
 STEPS = "steps"
 
-EVTYPE_FM_STEP_FAILURE = "com.equinor.ert.forward_model_step.failure"
-EVTYPE_FM_STEP_PENDING = "com.equinor.ert.forward_model_step.pending"
-EVTYPE_FM_STEP_RUNNING = "com.equinor.ert.forward_model_step.running"
-EVTYPE_FM_STEP_SUCCESS = "com.equinor.ert.forward_model_step.success"
-EVTYPE_FM_STEP_UNKNOWN = "com.equinor.ert.forward_model_step.unknown"
-EVTYPE_FM_STEP_WAITING = "com.equinor.ert.forward_model_step.waiting"
-EVTYPE_FM_STEP_TIMEOUT = "com.equinor.ert.forward_model_step.timeout"
+EVTYPE_REALIZATION_FAILURE = "com.equinor.ert.realization.failure"
```
xjules, line 20: These are the new realization statuses.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -118,18 +103,18 @@ def update_metadata(self, metadata: Dict[str, Any]) -> None:
...
-        step_update = _filter_nones(
-            {
-                "status": step.status,
-                "start_time": step.start_time,
-                "end_time": step.end_time,
-            }
+    def update_realization(
+        self,
+        real_id: str,
+        status: Optional[str] = None,
```
xjules, line 109: Is this fine that it can be `None`?

src/ert/job_queue/queue.py
```diff
@@ -50,30 +50,30 @@
...
 
-
-_FM_STEP_FAILURE = "com.equinor.ert.forward_model_step.failure"
-_FM_STEP_PENDING = "com.equinor.ert.forward_model_step.pending"
-_FM_STEP_RUNNING = "com.equinor.ert.forward_model_step.running"
-_FM_STEP_SUCCESS = "com.equinor.ert.forward_model_step.success"
-_FM_STEP_UNKNOWN = "com.equinor.ert.forward_model_step.unknown"
-_FM_STEP_WAITING = "com.equinor.ert.forward_model_step.waiting"
-
+# Duplicated from ert.ensemble_evaluator.identifiers to avoid circular imports
```
xjules, line 53: What about putting those `identifiers.py` into some standalone module?

tests/unit_tests/ensemble_evaluator/ensemble_evaluator_utils.py
```diff
@@ -103,13 +103,6 @@ def _evaluate(self, url):
             event_id = event_id + 1
             for real in range(0, self.test_reals):
                 job_failed = False
-                send_dispatch_event(
```
xjules, line 106: Should we send a corresponding `REALIZATION_UNKNOW` status event?

xjules (changes requested):
Looks very good! Just had a small comment here and there. Mainly though, if we can move the `identifiers.py`?

src/ert/ensemble_evaluator/identifiers.py
berland: I think I will call these *renamed* events.  There were realization *states* before, but they were implicitly updated based on the step states by the function `_propagate_step_update_to_realization`. Now we update them directly using their own events.

src/ert/job_queue/queue.py
berland: That could probably solve the circular uglyness yes, but I think that should be its own PR :)

berland: https://github.com/equinor/ert/issues/6457

src/ert/ensemble_evaluator/snapshot.py
berland: Hmm, that would be if you have an update for start- or end-time without having an update for the state. Not entirely sure if this is a scenario that actually occurs.

berland: You are right, it cannot happen that `status is None`. Will fix.

tests/unit_tests/ensemble_evaluator/ensemble_evaluator_utils.py
berland: Added that. But don't know what it is good for.

src/ert/ensemble_evaluator/snapshot.py
berland: Now changed. This implies that a `dict.get()` in `from_cloudevent` had to be changed to a direct accessor which can yield a `KeyError`. A buggy job_runner can then crash the evaluator by sending unknown events.

src/ert/job_queue/queue.py
berland: (the duplication was there prior to this PR)

src/ert/ensemble_evaluator/snapshot.py
berland: test_dark_storage triggered this KeyError. Fixed by not mapping event type to state before we know what kind of event we are talking about.

xjules (approved):
Looks good now! Could you squash / split commits?

PR #6446, opened 2023-10-29
Force fortran order for row scaling

dafeda:
Row scaling calls C++ code via pybind11 that expects Fortran order.

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/analysis/_es_update.py
+++ b/src/ert/analysis/_es_update.py
@@ -186,15 +186,33 @@ def _get_param_with_row_scaling(
     temp_storage: TempStorage,
     parameter: RowScalingParameter,
 ) -> List[Tuple[npt.NDArray[np.double], RowScaling]]:
+    """The row-scaling functionality is implemented in C++ and is made
+    accessible through the pybind11 library.
+    pybind11 requires that numpy arrays passed to it are in
+    Fortran-contiguous order (column-major), which is different from
+    numpy's default row-major (C-contiguous) order.
+    To ensure compatibility, numpy arrays are explicitly converted to Fortran order.
+    It's important to note that if an array originally in C-contiguous
+    order is passed to a function expecting Fortran order,
+    pybind11 will automatically create a Fortran-ordered copy of the array.
+    """
     matrices = []
+
     if parameter.index_list is None:
         matrices.append(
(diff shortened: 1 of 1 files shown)


oddvarlia (commented):
When I tested this with your change and without your change I got the same result (The order is not changed), but it does not mean the the memory layout is not changed.  But anyway it looks like it works using the version without this.  So when the scaling parameter is delivered from the non-adaptive localisation as a c-order flatten masked array, the result seems to be correct. I assume parameter.row_scaling contains the scaling factors calculated by the non-adaptive localisation code where the scaling factors are delivered to rowscaling object. Maybe I misunderstand numpy, but numpy.asfortranarray(vector) rearrange in memory to be fortran contiguous, but it does not change the index order like what numpy.ndarray.flatten(order=something) may do. So for me it looks like the index order is not change and that works anyway, but probably using fortrancontiguous input to the C++ library is more efficient?

dafeda:
> When I tested this with your change and without your change I got the same result (The order is not changed), but it does not mean the the memory layout is not changed. But anyway it looks like it works using the version without this. So when the scaling parameter is delivered from the non-adaptive localisation as a c-order flatten masked array, the result seems to be correct. I assume parameter.row_scaling contains the scaling factors calculated by the non-adaptive localisation code where the scaling factors are delivered to rowscaling object. Maybe I misunderstand numpy, but numpy.asfortranarray(vector) rearrange in memory to be fortran contiguous, but it does not change the index order like what numpy.ndarray.flatten(order=something) may do. So for me it looks like the index order is not change and that works anyway, but probably using fortrancontiguous input to the C++ library is more efficient?

The main reason to convert to Fortran order here is because `pybind11` will make a copy of the array if it's C-order.
The results should however stay the same as you observe.

oyvindeide (approved):
> The main reason to convert to Fortran order here is because pybind11 will make a copy of the array if it's C-order.
The results should however stay the same as you observe.

That is not very obvious, so perhaps add a comment about that?

dafeda:
> > The main reason to convert to Fortran order here is because pybind11 will make a copy of the array if it's C-order.
> > The results should however stay the same as you observe.
> 
> That is not very obvious, so perhaps add a comment about that?

Good point. I've added a comment.

PR #6456, opened 2023-10-31
Use QDir search paths instead of resourceIcon

pinkwah:
Make use of the "correct" way of adding search paths, rather than using the resourceIcon and co.

The most Qt way would be using `QResource` and the `qrc` compiler that ships with Qt. It takes an XML as input and produces a Qt resource file, which is a compressed archive of our contents. This, however, would mean having to pip-reinstall ERT every time we change the resources, which is a pain. Using this method allows us to keep distributing files as part of the Python wheel while being able to have Qt look up files directly.


aronhoyer (approved):
looks good! 🚀

PR #6491, opened 2023-11-03
Use plain qstat by default

berland:
The proxy script is deploy-specific and should not be the default in the driver.

**Issue**
Resolves future bug

**Approach**
Move the qstat_proxy configuration to ert-configurations. 

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist


berland:
Relates to https://github.com/equinor/ert-configurations/pull/79

jonathan-eq (approved):
LGTM, but I dont understand why the proxy script would be deploy specific? Wouldnt we use the proxy on every instance of TORQUE to avoid dos-ing the server?

berland:
> LGTM, but I dont understand why the proxy script would be deploy specific? Wouldnt we use the proxy on every instance of TORQUE to avoid dos-ing the server? EDIT: Is this only Equinor Azure specific?

I think the reason the Equinor Torque cluster is ddos-able is due to some DNS dependency (which is also flaky) when it is processing the qstat request. 

As an example, a Torque interface to a slurm cluster set up ad-hoc on a single multi-core computer should have no problem processing the qstat requests.

berland:
That said, we will however build something similar to the qstat proxy into a new queue code, maybe that is a reason to keep it.

berland:
Lets keep it in for now.

PR #6501, opened 2023-11-06
Fix the realization number for the stdout and stderr files

### LINKED ISSUE #6486: Stdout window header truncates realization numbers
User reported bug from ert-dev:

FYI: Seems to be a bug in the "realization" header / stdout window when you are not running in a "standard range": Here realization 693 get a header as realization 2. (Ran real 691, 692, 693 ...)

image

larsevj:
**Issue**
Resolves #6486 

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/gui/model/snapshot.py
+++ b/src/ert/gui/model/snapshot.py
@@ -352,7 +352,6 @@ class SnapshotModel(QAbstractItemModel):
     def _real_data(self, _index: QModelIndex, node: Node, role: int):
         if role == RealJobColorHint:
             colors: List[QColor] = []
-            assert node.parent  # mypy
 
             is_running = False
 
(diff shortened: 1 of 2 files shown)


larsevj:
Did a small refactoring in `_job_data` to avoid setting `real = node.parent` multiple times. Not strictly necessary in order to resolve the issue.

andreas-el:
> Did a small refactoring in `_job_data` to avoid setting `real = node.parent` multiple times. Not strictly necessary in order to resolve the issue.

But nice, since it was probably wrong to begin with to set that variable twice.

What did concern me a bit though, was if all nodes actually had a parent. If that isn't the case, we now enforce the system to stop. I've tried to see if this is the case, but could not find any situation where that happened.

Ref.
```
def _job_data(self, index: QModelIndex, node: Node, role: int):
         assert node.parent
         real = node.parent
 ```

andreas-el (approved):
Nice! Seems to do the trick. Good work! 🚀

larsevj:
> > Did a small refactoring in `_job_data` to avoid setting `real = node.parent` multiple times. Not strictly necessary in order to resolve the issue.
> 
> But nice, since it was probably wrong to begin with to set that variable twice.
> 
> What did concern me a bit though, was if all nodes actually had a parent. If that isn't the case, we now enforce the system to stop. I've tried to see if this is the case, but could not find any situation where that happened.
> 
> Ref.
> 
> ```
> def _job_data(self, index: QModelIndex, node: Node, role: int):
>          assert node.parent
>          real = node.parent
> ```

Yes that assert part is probably a leftover from before, and it might be best to just remove it? I believe that `_job_data` will only be called on job - nodes, which should all have a realization node as parent?

PR #6515, opened 2023-11-08
Cast queue option to string when validating against regex

yngve-sk:


xjules (changes requested):
let's do:
`option_value = str(option_value)`
on line 225

xjules (approved):
Nice job @yngve-sk !

Issue #6517, opened 2023-11-08
Different memory notations on different queues


yngve-sk:
Ert users should be able to specify memory options for queues, even though they internally have different memory usage notations. 

| Queue | Accepts only ints | Suffix | Link | Ert docs |
|-----------|----------|----------|---------|--------|
| SLURM | Yes (default to M) | K, M, G, T | [Link](https://slurm.schedmd.com/sbatch.html) | int format only | 
| TORQUE | No | kb, mb, gb | [Link](https://docs.adaptivecomputing.com/torque/4-0-2/Content/topics/commands/qsub.htm) | `"mb"` or `"gb"` |
| LSF |  Yes (in MB) | KB,K,MB,M,GB,G,TB,T,PB,P,EB,E,ZB,Z  | [Link](https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=requirements-resource-requirement-strings)| `LSF_RESOURCE` documented (a bit unclear how to specify mem usage) |

Related command line options per queue:
SLURM (int defaults to M) (ert docs describe INT only)
	`-- mem` [KMGT]
	`-- mem-per-cpu` [KMGT]
	`-- mem-per-gpu` [KMGT]
	`-- tmp` [KMGT]

TORQUE (Requires notation b, kb, mb, gb, tb, w, kw, mw, gw, tw) (ert docs say "mb, gb")
	`-- mem` (amount of physical memory used by job)
	`-- pvmem` (virtual memory used by any single process in job)
	`-- vmem` (virtual memory used by all concurrent processes in job)

	
LSF (No notation, number in MB) (not described in docs, describes a LSF_RESOURCE, outdated?)
	`-- M` (memory per core process memory limit)
	`-- R` (memory per core minimum available memory per core)

Other inconsistencies:
* `src/clib/tests/job_queue/test_job_slurm_driver.cpp` has a test that asserts `SLURM` has a "mb" suffix

berland:
Could it be an idea to standardize on one notation, and choose the same as Dask? I can't find it specified, but I see examples like "5 GiB" and "4 MiB"

PR #6521, opened 2023-11-08
Add corrections for memory usage specifications per driver

yngve-sk:
**Issue**
Resolves https://github.com/equinor/ert/issues/6517

### DIFF
--- a/src/ert/config/queue_config.py
+++ b/src/ert/config/queue_config.py
@@ -64,7 +64,7 @@ class QueueConfig:
             if queue_options[queue_system_val]:
                 _validate_queue_driver_settings(
                     queue_options[queue_system_val],
-                    queue_system_val.name,
+                    QueueSystem(queue_system_val).name,
                     throw_error=(queue_system_val == selected_queue_system),
                 )
 
--- a/src/ert/config/queue_config.py
+++ b/src/ert/config/queue_config.py
@@ -10,7 +10,13 @@ from typing import Any, Dict, List, Mapping, Tuple, no_type_check
 
 from ert import _clib
 
-from .parsing import ConfigDict, ConfigValidationError, ConfigWarning, QueueSystem
+from .parsing import (
+    ConfigDict,
+    ConfigValidationError,
+    ConfigWarning,
+    MaybeWithContext,
+    QueueSystem,
+)
 
 GENERIC_QUEUE_OPTIONS: List[str] = ["MAX_RUNNING"]
 VALID_QUEUE_OPTIONS: Dict[Any, List[str]] = {
@@ -108,9 +114,42 @@ def _check_for_overwritten_queue_system_options(
 
 queue_memory_options: Mapping[str, List[str]] = {
     "LSF": [],
-    "SLURM": [],
+    "SLURM": ["MEMORY_PER_CPU", "MEMORY"],
     "TORQUE": ["MEMORY_PER_JOB"],
     "LOCAL": [],
+}
+
+
+@dataclass
+class QueueMemoryStringFormat:
+    suffixes: List[str]
+
+    def validate(self, mem_str_format: str) -> bool:
+        return (
+            re.match(
+                r"\d+(" + "|".join(self.suffixes) + ")$",
+                mem_str_format,
+            )
+            is not None
+        )
+
+
+def parse_slurm_memopt(s: str) -> str:
+    return s.lower().replace("b", "").upper()
+
+
+def parse_torque_memopt(s: str) -> str:
+    if re.match(r"\d+[kgmt](?!\w)", s, re.IGNORECASE):
+        return s.lower() + "b"
+    if re.match(r"^\d+$", s):
+        return s + "kb"
+
+    return s
+
+
+queue_memory_usage_formats: Mapping[str, QueueMemoryStringFormat] = {
+    "SLURM": QueueMemoryStringFormat(suffixes=["", "K", "M", "G", "T"]),
+    "TORQUE": QueueMemoryStringFormat(suffixes=["kb", "mb", "gb", "KB", "MB", "GB"]),
 }
 
 queue_string_options: Mapping[str, List[str]] = {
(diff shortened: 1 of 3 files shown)


src/ert/config/queue_config.py
```diff
@@ -108,11 +115,68 @@ def generate_dict(option_list: List[Tuple[str, str]]) -> Dict[str, List[str]]:
...
+            "P",
+            "EB",
+            "E",
+            "ZB",
+            "Z",
+        ],
+        parse=parse_lsf_memopt,
+    ),
+    "SLURM": QueueMemoryStringFormat(
+        suffixes=["", "K", "M", "G", "T"], parse=parse_slurm_memopt
```
kwinkunks, line 172: I did not check these.

src/ert/config/queue_config.py
```diff
@@ -108,11 +115,68 @@ def generate_dict(option_list: List[Tuple[str, str]]) -> Dict[str, List[str]]:
...
+    if re.match(r"^\d+$", s):
+        return s + "mb"
+
+    return s
+
+
+queue_memory_usage_formats: Mapping[str, QueueMemoryStringFormat] = {
+    "LSF": QueueMemoryStringFormat(
+        suffixes=[
+            "",
```
kwinkunks, line 153: Using `bsub` directly, I can only get `rusage` to work without any units, and AFAICT we pass the `LSF_RESOURCE` text verbatim to `bsub -R`... so I'm not sure we can pass any units at all.

kwinkunks:
As you get to the bottom of this, I think it would be good to add something to the docs explaining these options.

I am 90% sure, BTW, that all of the queue systems are actually using power-of-2 units. So `1GB` is actually 2<sup>30</sup> bytes, so 1 073 741 824 bytes, which is 1.07 'real' GB.

src/ert/config/queue_config.py
```diff
@@ -108,11 +115,68 @@ def generate_dict(option_list: List[Tuple[str, str]]) -> Dict[str, List[str]]:
...
+            "ZB",
+            "Z",
+        ],
+        parse=parse_lsf_memopt,
+    ),
+    "SLURM": QueueMemoryStringFormat(
+        suffixes=["", "K", "M", "G", "T"], parse=parse_slurm_memopt
+    ),
+    "TORQUE": QueueMemoryStringFormat(
+        suffixes=["kb", "mb", "gb"],
```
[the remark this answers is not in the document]
yngve-sk: Fixed

src/ert/config/queue_config.py
yngve-sk: Re-checked, pasted from https://slurm.schedmd.com/sbatch.html
```
--mem=<size>[units]
Specify the real memory required per node. Default units are megabytes. Different units can be specified using the suffix [K|M|G|T]. Default value is DefMemPerNode and the maximum value is MaxMemPerNode. If configured, both parameters can be seen using the scontrol show config command. This parameter would generally be used if whole nodes are allocated to jobs (SelectType=select/linear). Also see --mem-per-cpu and --mem-per-gpu. The --mem, --mem-per-cpu and --mem-per-gpu options are mutually exclusive. If --mem, --mem-per-cpu or --mem-per-gpu are specified as command line arguments, then they will take precedence over the environment.
NOTE: A memory size specification of zero is treated as a special case and grants the job access to all of the memory on each node.
```

What do we feel about this part? 
```NOTE: A memory size specification of zero is treated as a special case and grants the job access to all of the memory on each node.``` disable/disallow/warn?

File: src/ert/config/queue_config.py (hunk 2 above)
Comment (yngve-sk) [reply src/ert/config/queue_config.py:153]: That is a bit odd, are we using an older version maybe? Official LSF docs on resource req strings (https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=requirements-resource-requirement-strings) give these examples: `bsub -C 500MB -M 1G -S 1TB -F 1GB -R "rusage[mem=512MB:swp=1GB:tmp=1TB]" sleep 100`. From the snake oil test config we have `"select[x86_64Linux] same[type:model]"` as value for the `LSF_RESOURCE`, so I'm not sure LSF_RESOURCE is even a pure memory specification keyword...

Comment (yngve-sk) [reply src/ert/config/queue_config.py:153]: TLDR; Omitting "parsing" of the LSF stuff as part of this, not finding any direct ways to specify mem usage on LSF other than as part of the resource string

File: src/ert/config/queue_config.py (hunk 3 above)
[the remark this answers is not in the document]
Comment (yngve-sk) [reply src/ert/config/queue_config.py]: Ref, this test in `test_queue_config.py`
```@pytest.mark.usefixtures("use_tmpdir", "set_site_config")
@pytest.mark.parametrize("memory_with_unit_str", ["1", "gb", "mb", "1 gb", "1kb"])
def test_that_invalid_memory_pr_job_raises_validation_error(memory_with_unit_str):
    filename = "config.ert"
    with open(filename, "w", encoding="utf-8") as f:
        f.write("NUM_REALIZATIONS 1\n")
        f.write("QUEUE_SYSTEM TORQUE\n")
        f.write(f"QUEUE_OPTION TORQUE MEMORY_PER_JOB {memory_with_unit_str}")
    with pytest.raises(ConfigValidationError):
        ErtConfig.from_file(filename)
```

src/ert/config/queue_config.py
```diff
@@ -206,23 +250,55 @@ def throw_error_or_warning(
...
-                ),
-                option_value,
-                throw_error,
-            )
+        elif option_name in queue_memory_options[queue_type]:
+            option_format = queue_memory_usage_formats[queue_type]
+            original_value = option_value
+            parsed_value = option_format.parse(str(original_value))
+
+            needs_correction = parsed_value != original_value
```
xjules, line 265: I would put this logic into a standalone pre-step to have a "nicer" separation.

src/ert/config/queue_config.py
```diff
@@ -205,24 +254,61 @@ def throw_error_or_warning(
         )
 
 
+def _try_correct_queue_driver_memory_notations(
```
xjules, line 257: `_correct_and_validate_queue_driver_memory_options`?

xjules (approved):
Very god job! :rocket:

src/ert/config/queue_config.py
kwinkunks: I think you have to allow it to avoid future confusion, and we have other params that work in a similar way (timouts, MAX_RUNTIME)

PR #6522, opened 2023-11-08
Remove usage of ert/util in job_queue_node

eivindjahren:
## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist


src/clib/lib/job_queue/job_node.cpp
```diff
@@ -177,7 +135,7 @@ void job_queue_node_set_status(job_queue_node_type *node,
         return;
 
     logger->debug("Set {}({}) to {}", node->job_name, node->queue_index,
-                  job_status_names.at(new_status).c_str());
+                  job_status_names.at(new_status));
```
andreas-el, line 138: debug takes std::string and not char * ?

src/clib/lib/job_queue/job_node.cpp
```diff
@@ -209,8 +167,7 @@ ERT_CLIB_SUBMODULE("queue", m) {
         std::optional<std::string> error_msg = std::nullopt;
 
         if (current_status & JOB_QUEUE_RUNNING && !node->confirmed_running) {
-            node->confirmed_running =
-                !node->status_file || fs::exists(node->status_file);
+            node->confirmed_running = !fs::exists(node->status_file);
```
andreas-el, line 170: Is this equivalent with the replaced code?
It used to be true if string was empty OR the file actually existed?

andreas-el (changes requested):
Had some comments we should look into

src/clib/lib/job_queue/job_node.cpp
eivindjahren: I think it works with either through pybind.

src/clib/lib/job_queue/job_node.cpp
eivindjahren: It was for the possibility of setting status_file to nullptr, which we don't do.

src/clib/lib/job_queue/job_node.cpp
andreas-el: Probably.

src/clib/lib/job_queue/job_node.cpp
eivindjahren: Oh shit, I guess we need to fix that. We don't really use that method any more, but we might want to investigate what is going on.

andreas-el (approved):
Great work! 🚀

PR #6524, opened 2023-11-08
Fix iteration number in GUI when using ensemble experiment

### LINKED ISSUE #6516: Iteration number displayed in UI labels is not consistent with iteration specified
**Describe the bug**
When running the poly-example and specifying iteration-2, you would expect that Iteration 2 is presented throughout the UI.
Rundialog has conflicting iteration numbers.
When opening STDOUT or STDERR files in rundialog, the Iteration is also wrong.

**Expected behaviour**
Iteration 2 being displayed throughout the Ert UI.

**Environment**

OS: macOS
ERT/Komodo release: main
Python version: 3.11
Remote/HPC execution involved: no

Screenshot 2023-11-08 at 09 47 50
Screenshot 2023-11-08 at 09 50 25
Screenshot 2023-11-08 at 09 50 37

larsevj:
**Issue**
Resolves #6516 

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/gui/simulation/run_dialog.py
+++ b/src/ert/gui/simulation/run_dialog.py
@@ -30,7 +30,7 @@ from ert.gui.ertnotifier import ErtNotifier
 from ert.gui.ertwidgets.message_box import ErtMessageBox
 from ert.gui.model.job_list import JobListProxyModel
 from ert.gui.model.progress_proxy import ProgressProxyModel
-from ert.gui.model.snapshot import FileRole, RealIens, IterNum, SnapshotModel
+from ert.gui.model.snapshot import FileRole, IterNum, RealIens, SnapshotModel
 from ert.gui.tools.file import FileDialog
 from ert.gui.tools.plot.plot_tool import PlotTool
 from ert.run_models import BaseRunModel
--- a/src/ert/gui/simulation/run_dialog.py
+++ b/src/ert/gui/simulation/run_dialog.py
@@ -30,7 +30,7 @@ from ert.gui.ertnotifier import ErtNotifier
 from ert.gui.ertwidgets.message_box import ErtMessageBox
 from ert.gui.model.job_list import JobListProxyModel
 from ert.gui.model.progress_proxy import ProgressProxyModel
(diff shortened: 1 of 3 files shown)


src/ert/gui/simulation/run_dialog.py
```diff
@@ -230,7 +230,7 @@ def _select_real(self, index):
         iter_ = index.model().get_iter()
         self._job_model.set_real(iter_, real)
         self._job_label.setText(
-            f"Realization id {index.data(RealIens)} in iteration {iter_}"
+            f"Realization id {index.data(RealIens)} in iteration {index.data(IterNum)}"
```
andreas-el, line 233: This might be correct, but the `iter_` that was earlier used in the text, is also used/passed to the set_real-function. Please verify that this is the intended behaviour.

larsevj: I think the iteration will still be mapped to `0` in the underlying data. For instance the `.get_iter()` function seems to return the index of the rowsInserted into the snapshot model, and I guess this will always start at `0`. I guess then that the only thing to do is to display the correct iteration id?

larsevj: I think `iter` is somewhat "misused" in the GUI components, as it is there used to keep track of the indicies of for instance the RealizationWidget. Since we only run one iteration then this index must be `0` and not the real iteration number.

andreas-el: I see this function leads to a signal, `self.real_changed.emit(iter_, real)`.
I can't immediately see that this signal is caught anywhere, can you look into that?

larsevj: I cant find any receivers either. Tried to check by running `self.receivers(self.real_changed)` which returns `0`. Not entirely sure if this is reliable though.

larsevj: And cant see any obvious difference when uncommenting the emit line.

andreas-el (approved):
Nice work! 🚀

PR #6540, opened 2023-11-10
Ensure queue_system in QueueConfig is correct type

yngve-sk:
**Issue**
Resolves queue driver initialization failing due to queue_type being a str, not `QueueSystem`, specifically happens in Everest in everest2res. Ref: https://github.com/equinor/everest/actions/runs/6786195687/job/18553630322?pr=2153


DanSava (approved):
Nice one :rocket:

PR #6548, opened 2023-11-10
Have restart_from combobox in es_mda be default disabled

### LINKED ISSUE #6544: ES_MDA restart_from combobox should have been disabled
The restart_from combobox should have been disabled but it not.
Toggling the checkbox will enable / disable correctly, but the initial state is wrong.
Screenshot 2023-11-10 at 10 39 17

andreas-el:
**Issue**
Resolves #6544 

The combobox will be enabled when being populated, so we need to listen for the content in the box to change to make a decision on if the box should be enabled or not.
Slight refactor to toggle-function.

**Approach**

## Pre review checklist

- [x] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/gui/simulation/multiple_data_assimilation_panel.py
+++ b/src/ert/gui/simulation/multiple_data_assimilation_panel.py
@@ -71,12 +71,12 @@ class MultipleDataAssimilationPanel(SimulationConfigPanel):
         layout.addRow("Active realizations:", self._active_realizations_field)
 
         self._restart_box = QCheckBox("")
-        self._restart_box.toggled.connect(self.restart_run)
+        self._restart_box.toggled.connect(self.restart_run_toggled)
         layout.addRow("Restart run:", self._restart_box)
 
         self._case_selector = CaseSelector(notifier)
+        self._case_selector.case_populated.connect(self.restart_run_toggled)
         layout.addRow("Restart from:", self._case_selector)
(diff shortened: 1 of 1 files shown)


larsevj (commented):
I see that there is little validation for if the cases you provide are "valid". For instance if you restart from an empty case or a fully finished experiment then ert seems to be stuck. But this is maybe hard to avoid, and not really related to this issue.

DanSava (approved):
LGTM :guitar:

andreas-el:
> I see that there is little validation for if the cases you provide are "valid". For instance if you restart from an empty case or a fully finished experiment then ert seems to be stuck. But this is maybe hard to avoid, and not really related to this issue.

I'd be happy to change the PR if the issue you describe is related or a cause of the change in the PR. But to me this sounds like a different issue? Would it make sense to create a separate issue for the aforementioned situation that can occur?

larsevj:
> > I see that there is little validation for if the cases you provide are "valid". For instance if you restart from an empty case or a fully finished experiment then ert seems to be stuck. But this is maybe hard to avoid, and not really related to this issue.
> 
> I'd be happy to change the PR if the issue you describe is related or a cause of the change in the PR. But to me this sounds like a different issue? Would it make sense to create a separate issue for the aforementioned situation that can occur?

Yes it is a different issue, just something I noticed. I can look into creating a separate issue.

PR #6558, opened 2023-11-13
Unify execute API of JobQueue, refactor websocket connection

berland:
* Put websocket connection handling into only one (async) function, that will run until it is finished.
* Reduce the execute functions to only one, letting both legacy ensemble and simulation context use it. 

The simulation_context now has to call it in an async loop, and it is also doing slightly more work (logging changes) than before.

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/ensemble_evaluator/_builder/_legacy.py
+++ b/src/ert/ensemble_evaluator/_builder/_legacy.py
@@ -158,7 +158,7 @@ class LegacyEnsemble(Ensemble):
         # Set up the timeout-mechanism
         timeout_queue = asyncio.Queue()  # type: ignore
         # Based on the experiment id the generator will
-        # give a function returning cloud event or protobuf
+        # give a function returning cloud event
         event_creator = self.generate_event_creator(experiment_id=experiment_id)
         on_timeout, send_timeout_future = self.setup_timeout_callback(
             timeout_queue, cloudevent_unary_send, event_creator
@@ -169,16 +169,11 @@ class LegacyEnsemble(Ensemble):
         if not self._config:
             raise ValueError("no config")  # mypy
 
-        # event for normal evaluation, will be overwritten later in case of failure
-        # or cancellation
-        result = event_creator(identifiers.EVTYPE_ENSEMBLE_STOPPED, None)
-
         try:
-            # Dispatch STARTED-event
-            out_cloudevent = event_creator(identifiers.EVTYPE_ENSEMBLE_STARTED, None)
-            await cloudevent_unary_send(out_cloudevent)
+            await cloudevent_unary_send(
+                event_creator(identifiers.EVTYPE_ENSEMBLE_STARTED, None)
+            )
 
-            # Submit all jobs to queue and inform queue when done
             for real in self.active_reals:
                 self._job_queue.add_realization(real, callback_timeout=on_timeout)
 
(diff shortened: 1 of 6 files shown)


berland:
Tested manually with GUI for poly_example, bigpoly and also with everest's egg model.

berland:
Up for discussion: Should this be squashed?

src/ert/job_queue/queue.py
```diff
@@ -230,143 +251,111 @@ def _translate_change_to_cloudevent(
...
+    async def _jobqueue_publisher(self) -> None:
+        assert self._changes_to_publish is not None
+        ee_headers = Headers()
+        if self._ee_token is not None:
+            ee_headers["token"] = self._ee_token
+
+        if self._ee_uri is None:
+            # If no ensemble evaluator present, we will publish to the log
+            while (
+                change := await self._changes_to_publish.get()
```
jonathan-eq, line 277: I did not know about this (walrus operator :=). Thanks🎉

jonathan-eq (approved):
Great Job!

src/ert/ensemble_evaluator/_builder/_legacy.py
```diff
@@ -199,46 +194,35 @@ async def _evaluate_inner(  # pylint: disable=too-many-branches
...
-            result = event_creator(identifiers.EVTYPE_ENSEMBLE_FAILED, None)
+            result = identifiers.EVTYPE_ENSEMBLE_FAILED
 
-        else:
-            logger.debug("ensemble finished normally")
-
-        finally:
-            await timeout_queue.put(None)  # signal to exit timer
-            await send_timeout_future
+        await timeout_queue.put(None)  # signal to exit timer
```
xjules, line 221: I see that you've removed the finally section. Isn't point of finally that it gets always executed even when there was an exception?

src/ert/job_queue/queue.py
```diff
@@ -230,54 +240,48 @@ def _translate_change_to_cloudevent(
...
 
-    @staticmethod
     async def _publish_changes(
-        ens_id: str,
-        changes: Dict[int, str],
-        ee_connection: WebSocketClientProtocol,
+        self, changes: Dict[int, str], ee_connection: WebSocketClientProtocol
     ) -> None:
+        assert self.ens_id is not None  # mypy
         events = deque(
```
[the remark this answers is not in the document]
berland: probably a local and minor optimization thing to have a double-ended queue.

src/ert/job_queue/queue.py
```diff
@@ -441,27 +430,23 @@ def changes_without_transition(self) -> Tuple[Dict[int, str], List[JobStatus]]:
...
-        ens_id: str,
-        dispatch_url: str,
-        cert: Optional[str],
-        token: Optional[str],
         experiment_id: Optional[str] = None,
     ) -> None:
         for q_index, q_node in enumerate(self.job_list):
             cert_path = f"{q_node.run_path}/{CERT_FILE}"
-            if cert is not None:
+            if self._ee_cert is not None:
```
xjules, line 437: Was it not clear when conninfo was passed directly into the function as a parameter?

src/ert/ensemble_evaluator/_builder/_legacy.py
berland: Yes, but since we capture absolutely every exception, what comes after the `try:` block will get executed nonetheless.

src/ert/simulator/simulation_context.py
```diff
@@ -72,7 +73,7 @@ def _run_forward_model(
             )
         ]
 
-    job_queue.execute_queue(queue_evaluators)
+    asyncio.run(job_queue.execute(evaluators=queue_evaluators))
```
xjules, line 76: `run_until_complete`? calling `asyncio.run` always creates a new event loop actually, but maybe that's what we want here?

src/ert/job_queue/queue.py
xjules: The only thing is that Queue does not need to know about connection setup between ee and job_runner; hence my comment on passing it rather as a function params :shrug:

src/ert/simulator/simulation_context.py
berland: asyncio.run() is the Python 3.7 way of doing asyncio.get_event_loop().run_until_complete() (which is from python 3.4)

berland: yes, we do need a new event loop here, because we are coming from a totally sync context (everest)

src/ert/ensemble_evaluator/_builder/_legacy.py
xjules: Right, I've missed the `except Exception:`. Then it's fine I guess.

xjules:
> Up for discussion: Should this be squashed? (there are some minor details to be fixed if not squashed, some print-statements and maybe one mypy issue)

I would the fixup commits at least :)

src/ert/job_queue/queue.py
berland: That is correct, I am removing that "feature" to have different connection info.

PR #6587, opened 2023-11-16
Fix deprecation info for Analysis Module

andreas-el:
The deprecation suggestions where initially testing for `ENKF_NCOMP 1`
instead of the proper syntax
`ANALYSIS_SET_VAR <MODULE> ENKF_NCOMP 1`
Also fixed for ENKF_SUBSPACE_DIMENSION and ENKF_FORCE_NCOMP

**Issue**
Resolves #my_issue

**Approach**

## Pre review checklist

- [x] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/config/analysis_module.py
+++ b/src/ert/config/analysis_module.py
@@ -110,8 +110,8 @@ def get_mode_variables(mode: AnalysisMode) -> Dict[str, "VariableInfo"]:
 
 
 class AnalysisModule:
-    TRUNC_ALTERNATE_KEYS = ["ENKF_NCOMP", "ENKF_SUBSPACE_DIMENSION"]
-    SPECIAL_KEYS = ["INVERSION", *TRUNC_ALTERNATE_KEYS]
+    DEPRECATED_KEYWORDS = ["ENKF_NCOMP", "ENKF_SUBSPACE_DIMENSION"]
+    JUST_REMOVE_KEYWORDS = ["ENKF_FORCE_NCOMP"]
 
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


src/ert/config/analysis_module.py
```diff
@@ -223,7 +217,7 @@ def set_var(self, var_name: str, value: Union[float, int, bool, str]) -> None:
                     value = str(value).lower() != "false"
 
                 var["value"] = var["type"](value)
-        else:
+        elif var_name not in self.JUST_REMOVE_KEYWORDS:
```
DanSava, line 220: Is there a conditional branch where `var_name` is in `self.JUST_REMOVE_KEYWORDS` is handled?  Also as far as I can tell now unknown `var_names` will no longer throw a config validation error. Is this intended behaviour?

DanSava (commented):
Looks good I just have some small questions

src/ert/config/analysis_module.py
andreas-el: It's actually reverse. We don't raise if the keyword is in that list, otherwise (for all unknown `var_names`) we raise the error.
The keywords found in `self.JUST_REMOVE_KEYWORDS` have are already been suggested in the UI at this point, and should be disregarded since they have no effect.

I can take a look and see if I can do this in a more readable way?

andreas-el:
```
        elif var_name in self.JUST_REMOVE_KEYWORDS:
            pass
        else:
            raise ConfigValidationError(
                f"Variable {var_name!r} not found in {self.name!r} analysis module"
            )
```
I guess this is might be more readable, avoiding the `not` in there?

DanSava (approved):
Good one sorry for the confusing questions :rocket:

PR #6597, opened 2023-11-17
Fix simulation_context

xjules:
Fix simulation context to match current queue API.

### DIFF
--- a/src/ert/job_queue/queue.py
+++ b/src/ert/job_queue/queue.py
@@ -18,9 +18,8 @@ from websockets.client import WebSocketClientProtocol, connect
 from websockets.datastructures import Headers
 from websockets.exceptions import ConnectionClosed
 
-from ert.constant_filenames import CERT_FILE, JOBS_FILE
-
 from ert.callbacks import forward_model_ok
+from ert.constant_filenames import CERT_FILE, JOBS_FILE
 from ert.load_status import LoadStatus
 
 from .driver import Driver
@@ -113,7 +112,10 @@ class JobQueue:
             for real in self._realizations
         )
 
-    def count_status(self, state: RealizationState) -> int:
+    def real_state(self, iens: int) -> RealizationState:
+        return self._realizations[iens].current_state
+
+    def count_real_state(self, state: RealizationState) -> int:
         return len([real for real in self._realizations if real.current_state == state])
 
     async def run_done_callback(self, state: RealizationState):
@@ -153,11 +155,8 @@ class JobQueue:
     def queue_size(self) -> int:
         return len(self._realizations)
 
-    def count_running(self) -> int:
-        return sum(
-            real.current_state == RealizationState.RUNNING
-            for real in self._realizations
-        )
+    def _add_realization(self, realization: QueueableRealization) -> None:
+        self._realizations.append(RealizationState(self, realization, retries=1))
 
     def max_running(self) -> int:
         max_running = 0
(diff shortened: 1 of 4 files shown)


src/ert/simulator/simulation_context.py
```diff
@@ -153,25 +153,25 @@ def isRunning(self) -> bool:
         return self._sim_thread.is_alive() or self._job_queue.is_active()
 
     def getNumPending(self) -> int:
-        return self._job_queue.count_status(JobStatus.PENDING)  # type: ignore
+        return self._job_queue.count_status(RealizationState.PENDING)  # type: ignore
```
berland, line 156: I think we can remove the type ignores now that cwrap is out

src/ert/simulator/simulation_context.py
```diff
@@ -153,25 +153,25 @@ def isRunning(self) -> bool:
...
 
     def getNumRunning(self) -> int:
-        return self._job_queue.count_status(JobStatus.RUNNING)  # type: ignore
```
berland, line 159: I think these five functions are only u sed in `batch_simulator_context()` line 55-60. We can get rid of these functions and rather provide these statistics directly from job_queue.

src/ert/job_queue/queue.py
```diff
@@ -110,8 +110,11 @@ def is_active(self) -> bool:
             for real in self._realizations
         )
 
-    def count_status(self, state: RealizationState) -> int:
-        return len([real for real in self._realizations if real.current_state == state])
+    def real_state(self, iens: int) -> RealizationState:
```
kwinkunks, line 113: Suggest using the full name, `realization_state`, if that's what it is. 'Real state' could be misinterpreted.

kwinkunks: Or potentially find another abbreviation, like `rlz`?

xjules: `realization_state` sounds good enough!

src/ert/simulator/simulation_context.py
```diff
@@ -152,26 +152,11 @@ def isRunning(self) -> bool:
...
-    def getNumSuccess(self) -> int:
-        return self._job_queue.count_status(JobStatus.SUCCESS)  # type: ignore
-
-    def getNumFailed(self) -> int:
-        return self._job_queue.count_status(JobStatus.FAILED)  # type: ignore
-
-    def getNumWaiting(self) -> int:
-        return self._job_queue.count_status(JobStatus.WAITING)  # type: ignore
-
     def didRealizationSucceed(self, iens: int) -> bool:
```
kwinkunks, line 155: Do we really want Camelcase function names?

xjules: Yes, there are some other functions in `simulation_context` using Camelcase, but I would keep the current API names unchanged.

src/ert/job_queue/queue.py
```diff
@@ -151,12 +154,6 @@ def queue_size(self) -> int:
...
 
-    def count_running(self) -> int:
-        return sum(
-            real.current_state == RealizationState.RUNNING
-            for real in self._realizations
-        )
-
     def max_running(self) -> int:
         max_running = 0
         if self.driver.has_option("MAX_RUNNING"):
```
kwinkunks, line 159: Since all the drivers have this option, and we know this is a positive integer option (I tihnk?), is it not possible to store the correct type already (avoid cast to `int`) and just go straight to `get`, and then do the thing `if not None`?

src/ert/simulator/simulation_context.py
xjules: done

src/ert/job_queue/queue.py
xjules: I'll create an issue to exploit padentic for queueconfig.

kwinkunks (approved):
This looks good to me... but that may not be sufficient criterion :D

PR #6611, opened 2023-11-20
Add support for SUBMIT_SLEEP

berland:
Written assuming there will be a global option to set SUBMIT_SLEEP, but will currently pick from the selected queue system.


src/ert/job_queue/queue.py
```diff
@@ -159,6 +159,12 @@ def count_running(self) -> int:
             for real in self._realizations
         )
 
+    def count_submitted(self) -> int:
```
xjules, line 162: Now you can do `self.count_realization_state(RealizationState.SUBMITTED)`

xjules:
I think you need to rebase

berland:
Should be ready now, with tests

src/ert/job_queue/queue.py
```diff
@@ -166,27 +169,37 @@ def max_running(self) -> int:
...
+        time_of_last_submission: datetime.datetime = (
+            datetime.datetime.now() - submit_delta
+        )
+        while self.is_active():
+            while (
+                self.count_realization_state(RealizationState.RUNNING)
+                + self.count_realization_state(RealizationState.SUBMITTED)
+                < self.max_running()
+            ):
+                if datetime.datetime.now() - time_of_last_submission >= submit_delta:
```
xjules, line 186: I thought that this check is not required if we do `await asyncio.sleep(self._queue_config.submit_sleep)` or?

berland: removed now

JHolba (approved):

PR #6614, opened 2023-11-21
Remove JobStatus

xjules:
:scissors: JobStatus and inject states directly into RealizationState


kwinkunks (approved):
This looks much nicer :+1:

PR #6624, opened 2023-11-21
Group ee variables in dataclass

JHolba:


berland:
This needs the tests in tests/unit_tests/simulator/test_simulation_context.py (at least) to pass, which is currently blocked by another bug.

PR #6627, opened 2023-11-21
Rename internal functions and attributes on state change

berland:
Maybe better names..


xjules (approved):
I think it makes it more explicit! Very good!

PR #6631, opened 2023-11-22
Rename RealizationState to RealizationStorageState and move it to storage module

xjules:
**Issue**
In the upcoming Queue we would like reserve RealizationState to describe the execution state when running a single realization and therefore suggest renaming the current `RealizationState` to `RealizationStorageState`

### DIFF
--- a/src/ert/analysis/_es_update.py
+++ b/src/ert/analysis/_es_update.py
@@ -27,7 +27,7 @@ from iterative_ensemble_smoother.experimental import (
 )
 
 from ert.config import Field, GenKwConfig, SurfaceConfig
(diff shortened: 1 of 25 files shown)


src/ert/analysis/_es_update.py
```diff
@@ -492,10 +492,13 @@ def analysis_ES(
...
-                observation_values,
-                observation_errors,
-                update_snapshot,
+            (
+                S,
+                (
+                    observation_values,
+                    observation_errors,
+                    update_snapshot,
+                ),
```
berland, line 501: Is this some peculiar black version that has done this?

berland (approved):

PR #6644, opened 2023-11-23
Make run_done_callback async

berland:
Make run_done_callback async
    
    This mitigates slightly blocking the async event loop
    by at least yielding wherever possible, at least in between
    every read/write to storage.

### DIFF
--- a/src/ert/callbacks.py
+++ b/src/ert/callbacks.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import logging
 import time
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


src/ert/callbacks.py
```diff
@@ -59,6 +61,7 @@ def _write_responses_to_storage(
                 f"Loaded {config.name}",
                 extra={"Time": f"{(time.perf_counter() - start_time):.4f}s"},
             )
+            await asyncio.sleep(0)
```
xjules, line 64: I guess (all these) await asyncio.sleep(0) is for the sake of yielding right?

berland: yep!

xjules (approved):
To be honest, it looks good to me :) :rocket:

PR #6649, opened 2023-11-23
Distinguish between RealizationState and RealizationStorageState in SimulationContext

xjules:
Distinguish between RealizationState and RealizationStorageState in SimulationContext

### DIFF
--- a/src/ert/simulator/simulation_context.py
+++ b/src/ert/simulator/simulation_context.py
@@ -10,8 +10,8 @@ import numpy as np
 
 from ert.config import HookRuntime
 from ert.enkf_main import create_run_path
-from ert.job_queue import JobQueue
-from ert.realization_state import RealizationState
+from ert.job_queue import JobQueue, RealizationState
+from ert.realization_state import RealizationState as RealizationStorageState
 from ert.run_context import RunContext
 from ert.runpaths import Runpaths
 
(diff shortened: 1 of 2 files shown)


berland (approved):
👍🏻

src/ert/simulator/simulation_context.py
```diff
@@ -68,12 +68,15 @@ def _run_forward_model(
...
-                job_queue.stop_long_running_jobs,
+                job_queue.stop_long_running_realizations,
                 ert.ert_config.analysis_config.minimum_required_realizations,
             )
         ]
 
-    asyncio.run(job_queue.execute(evaluators=queue_evaluators))
+    try:
+        asyncio.run(job_queue.execute(evaluators=queue_evaluators))  # type: ignore
+    except Exception:
```
berland, line 78: bare exceptions are not a good idea.  Being bare, it should at least be re-raised?

src/ert/simulator/simulation_context.py
```diff
@@ -68,12 +68,15 @@ def _run_forward_model(
...
+        job_queue.kill_all_jobs()
```
berland, line 79: This trickles down to an  asyncio.create_task() in realization_state.py. Will that work outside the async context?

src/ert/simulator/simulation_context.py
berland: There is a bare exception catch in queue.py already. Should not be needed.

xjules: Yes, but the point is that this gets raised in here only.

src/ert/simulator/simulation_context.py
xjules: That's what I don't know :man_shrugging:

berland: It won't work. You'll get "RunTimeError: no running event loop"

src/ert/simulator/simulation_context.py
berland: I think the queue should be responsible for catching the exception(s) here yes.

src/ert/simulator/simulation_context.py
xjules: removed!

PR #6657, opened 2023-11-23
Fix error reporting, unit_tests/cli/test_integration_cli.py passes.

berland:
Report stderr from realization the same way as c-code did
    
    This reimplements job_queue_node_fscanf_EXIT in job_node.cpp
    
    It has a small breakage in the contents of the ERROR file
    left in runpath in that it is now valid XML as opposed to
    invalid XML produced by the file.py reporter before.


pyproject.toml
```diff
@@ -54,6 +54,7 @@ dependencies=[
     "typing_extensions",
     "jinja2",
     "lark",
+    "lxml",
```
berland, line 57: Ouch, not in komodo..

kwinkunks (approved):
This looks great.

PR #6658, opened 2023-11-23
Fix regression from async forward_model

berland:


tests/unit_tests/ensemble_evaluator/conftest.py
```diff
@@ -64,9 +64,10 @@ def queue_config_fixture():
 @pytest.fixture
 def make_ensemble_builder(queue_config):
     def _make_ensemble_builder(monkeypatch, tmpdir, num_reals, num_jobs, job_sleep=0):
+        forward_mock = AsyncMock(return_value=(LoadStatus.LOAD_SUCCESSFUL, ""))
         monkeypatch.setattr(
             "ert.job_queue.queue.forward_model_ok",
-            lambda _: (LoadStatus.LOAD_SUCCESSFUL, ""),
+            AsyncMock(return_value=(LoadStatus.LOAD_SUCCESSFUL, "")),
```
kwinkunks, line 70: Everything async is new to me, but I wondered if we want to assert that this was awaited?

berland: It will error if not awaited.

kwinkunks (approved):
Looks good!

PR #6668, opened 2023-11-24
Fix bugs uncovered in simulation context

berland:
Poking an internal variable in queue, _queue_stopped. 

There should perhaps be an official way in order to stop the queue from a sync context


kwinkunks (approved):
Looks good :+1:

PR #6670, opened 2023-11-24
Mypys

berland:


src/ert/job_queue/realization_state.py
```diff
@@ -135,24 +136,27 @@ def on_enter_EXIT(self) -> None:
...
 
     def on_enter_DO_KILL(self) -> None:
         asyncio.create_task(self.jobqueue.driver.kill(self))
+
+
+def log_info_from_exit_file(exit_file_path: Path) -> None:
+    if not exit_file_path.exists():
+        return
+    exit_file = etree.parse(exit_file_path)
+    filecontents: Dict[str, str] = defaultdict(str)
```
kwinkunks, line 155: It's nice to move this out to a function, but unpacking from the etree to a dict, then from the dict to the f-strings feels a little roundabout. It was maybe a bit repetitive before, but I think it was clearer (though I tripped a little on the order).

Maybe you could use something like `f"...".format(**filecontents)` to save the unpacking? And even just use a `list` instead of a `dict` and unpack like `*filecontents` into unlabeled `{}` pairs?

IDK, not a bit deal either way, just a thought.

berland: Added an alternative, check if it looks better (I find it hard to decide..)

kwinkunks (approved):
Nice. I think it's an improvement... I guess type hinting forces you to cast to `str` even though you don't really need to.

PR #6685, opened 2023-11-28
Fix mypy erros in realization_state.py

xjules:
Fix mypy erros in realization_state.py


kwinkunks:
I think @berland already changed this  https://github.com/equinor/ert/pull/6670

Issue #6702, opened 2023-11-29
test_simulation_context - fix skipped tests


xjules:
Make  `tests/unit_tests/simulator/test_simulation_context.py` pass with no `pytest.mark.skip`

PR #6707, opened 2023-11-29
Fix sim ctx

berland:
Fix test_simulation_context by fixing a bug in isRealizationFinished()

Solves #6702


xjules (approved):
Nice! Now it makes sense! :rocket:

Issue #6715, opened 2023-11-29
test_driver either refactor or remove the tests.


xjules:
Just remove the xfail / skip decorator and make the tests run.
Validate through config_dict_generate if it is tested already. 
Also the tests might have been previously needed due to C-wrap so might not be necessary now.

PR #6719, opened 2023-11-29
rm test_driver.py

berland:
Remove test_driver, superfluous
 
We don't need testing of get_options, and that
we get an LSF driver when we ask for it

Solves #6715


xjules:
I think it contains some weird commits.

JHolba (commented):
looks good now

xjules (approved):
:scissors:

Issue #6705, opened 2023-11-29
Map out RealizationState state diagram


pinkwah:
Definition of done: There is a document describing the state machine, and how changes in the state of the job on the cluster affects the state of the realisation in ert.

It should be documented in the ert docs.

pinkwah:
Moving this to backlog as I'm uncertain about the scope. Will discuss in 2024.

pinkwah:
We should document the Scheduler as things complete. The RealizationState does not exist anymore per-se.

PR #6723, opened 2023-11-29
Add developer documentation on Scheduler

pinkwah:
Resolves: #6705 

Todo:
- [ ] Add motivation (why are we implementing this ourselves and not using dask-jobqueue?)


pinkwah:
Closing as I doubt whether this is a good way to go about documenting an internal part of Ert.

PR #6769, opened 2023-12-07
Add LocalDriver for Scheduler

### LINKED ISSUE #6759: Move LOCAL driver from the feature branch under the Scheduler module.
It is up for discussion, but since Scheduler requires new python driver it would be reasonable to move it under the scheduler module first.

pinkwah:
Resolves #6759

### DIFF
--- a/src/ert/scheduler/driver.py
+++ b/src/ert/scheduler/driver.py
@@ -1,0 +1,91 @@
+from __future__ import annotations
+
+import asyncio
+import os
+from abc import ABC, abstractmethod
+from enum import Enum
+from typing import (
+    MutableMapping,
+    Optional,
+    Tuple,
+)
+
+
+class JobEvent(Enum):
+    STARTED = 0
+    COMPLETED = 1
+    FAILED = 2
+    ABORTED = 3
+
+
+class Driver(ABC):
+    """Adapter for the HPC cluster."""
+
+    event_queue: asyncio.Queue[Tuple[int, JobEvent]]
+
+    def __init__(self) -> None:
+        self.event_queue = asyncio.Queue()
+
+    @abstractmethod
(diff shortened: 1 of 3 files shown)


src/ert/scheduler/driver.py
```diff
@@ -0,0 +1,72 @@
...
+            preexec_fn=os.setpgrp,
+        )
+        await self.event_queue.put((iens, JobEvent.STARTED))
+        try:
+            if await proc.wait() == 0:
+                await self.event_queue.put((iens, JobEvent.COMPLETED))
+            else:
+                await self.event_queue.put((iens, JobEvent.FAILED))
+        except asyncio.CancelledError:
+            proc.kill()
```
xjules, line 71: `proc.terminate()` ?

xjules (approved):
Nice! Let's bring it in :rocket:

src/ert/scheduler/driver.py
```diff
@@ -0,0 +1,89 @@
...
+    STARTED = 0
+    COMPLETED = 1
+    FAILED = 2
+    ABORTED = 3
+
+
+class Driver(ABC):
+    """Adapter for the HPC cluster."""
+
+    event_queue: asyncio.Queue[Tuple[int, JobEvent]]
```
xjules, line 22: We need to do typing.Queue here:
```from typing import Queue
event_queue: Queue[Tuple[int, JobEvent]] ....
```

xjules (changes requested):
Small update required with typing

src/ert/scheduler/driver.py
pinkwah: I think the solution is `from __future__ import annotations`

xjules: Or that :)

xjules (approved):
Re-approved :+1:

PR #6772, opened 2023-12-08
Move LocalDriver into separate file from Driver

pinkwah:


src/ert/scheduler/local_driver.py
```diff
@@ -0,0 +1,43 @@
...
+
+from ert.scheduler.driver import Driver, JobEvent
+
+
+class LocalDriver(Driver):
+    def __init__(self) -> None:
+        super().__init__()
+        self._tasks: MutableMapping[int, asyncio.Task[None]] = {}
+
+    async def submit(self, iens: int, executable: str, /, *args: str, cwd: str) -> None:
```
xjules, line 15: Shall we maybe add here:
```
if self.event_queue:
        self.event_queue = asyncio.Queue()
```

xjules (approved):
Let's move it in! Py38 scheduler fix will come on top.

Issue #6760, opened 2023-12-05
FeatureToggling: Toggle Scheduler


pinkwah:
We have four drivers: `LOCAL`, `LSF`, `TORQUE` and `SLURM`. We want to be able to make the `LOCAL` driver use the new Scheduler by default in production, while the other drivers still use the legacy `JobQueue`.

We should modify the `FeatureToggling` so that:
- There is a default, force-enable and force-disable.
- `--enable-FEATURE` will force-enable, `--disable-FEATURE` will force-disable.
- Consumers of `FeatureToggling` can determine which of the three states it is.

Bonus points for environment variables, so as to make it easier to run CI. For each feature, look for `ERT_{FEATURE}` environment variable:
- If empty string or unset, use default.
- If `0`, force-disable.
- All other non-zero, non-empty values means force-enable.

The order for something to be enabled or disabled is:
1. Has the user provided explicit command line arguments? (`--enable-scheduler` or `--disable-scheduler`)
2. Does the user have explicit environment variables set? (`ERT_SCHEDULER=1` or `ERT_SCHEDULER=0`)
3. Use the default value for the given toggle.

xjules:
This has been done already.

PR #6778, opened 2023-12-11
Feature toggling can now be done from environment variables

JHolba:
**Issue**
#6760 

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/shared/feature_toggling.py
+++ b/src/ert/shared/feature_toggling.py
@@ -1,29 +1,50 @@
 import logging
+import os
 from argparse import ArgumentParser
 from copy import deepcopy
-from typing import Optional
+from typing import Dict, Optional, Union
 
 from ert.namespace import Namespace
 
 
 class _Feature:
-    def __init__(self, default_enabled: bool, msg: Optional[str] = None) -> None:
-        self.is_enabled = default_enabled
+    def __init__(
+        self, default: Optional[bool], msg: Optional[str] = None, optional: bool = False
+    ) -> None:
+        self._value = default
         self.msg = msg
+        self.optional = optional
+
+    def validate_value(self, value: Union[bool, str, None]) -> Optional[bool]:
+        if type(value) is bool or value is None:
+            return value
+        elif value.lower() in ["true", "1"]:
+            return True
+        elif value.lower() in ["false", "0"]:
+            return False
+        elif self.optional and value.lower() in ["default", ""]:
+            return None
+        else:
+            raise ValueError(
+                f"This option can only be set to {'True/1, False/0 or Default/<empty>' if self.optional else 'True/1 or False/0'}"
+            )
+
+    @property
+    def value(self) -> Optional[bool]:
+        return self._value
+
+    @value.setter
+    def value(self, value: Optional[bool]) -> None:
+        self._value = self.validate_value(value)
 
 
 class FeatureToggling:
-    _conf_original = {
-        "new-storage": _Feature(
-            default_enabled=False,
-            msg=(
-                "The new storage solution is experimental! "
-                "Thank you for testing our new features."
-            ),
-        ),
+    _conf_original: Dict[str, _Feature] = {
         "scheduler": _Feature(
-            default_enabled=False,
-            msg="Use Scheduler instead of JobQueue",
+            default=None,
+            msg="Default value for use of Scheduler has been overridden\n"
+            "This is experimental and may cause problems",
+            optional=True,
         ),
     }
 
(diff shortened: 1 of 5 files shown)


src/ert/shared/feature_toggling.py
```diff
@@ -61,6 +63,18 @@ def update_from_args(args: Namespace) -> None:
...
 
+    @staticmethod
+    def update_from_environment() -> None:
+        pattern = re.compile("(?:ERT_)(ENABLE_|DISABLE_)(.*)")
+        args = dict()
+        for key, value in os.environ.items():
+            if pattern.match(key):
+                feature_name = key[4:]
+                feature_name_arg = feature_name.replace("-", "_").lower()
+                args[feature_name_arg] = value
```
pinkwah, line 74: Your order of operations in the `str.replace` function is wrong (it needs to replace `_` with `-`). Please add a test for this.

Also, I reckon explicit disable doesn't make sense in environment variables, where an empty string and `0` already mean disable? Like, what is the correct behaviour of `ERT_ENABLE_SCHEDULER=1 ERT_DISABLE_SCHEDULER=1`, or `ERT_DISABLE_SCHEDULER= ert test_run`?

```suggestion
        pattern = re.compile("^ERT_(ENABLE_|DISABLE_)(.+)$")
        args: MutableMapping[str, str] = {}
        for key, value in os.environ.items():
            if (match := pattern.match(key)):
                feature = match[2].replace("_", "-").lower()
                args[feature] = match[1] == "ENABLE_"
```

tests/unit_tests/shared/test_feature_toggling.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+@pytest.fixture(autouse=True)
+def mocked_valid_file(monkeypatch):
+    monkeypatch.setattr(ert.__main__, "valid_file", Mock(return_value=""))
+
+
+def test_feature_toggling_from_env(monkeypatch):
+    envs = {"ERT_ENABLE_SCHEDULER": True}
+    monkeypatch.setattr(os, "environ", envs)
```
pinkwah, line 19:
```suggestion
    monkeypatch.setenv("ERT_ENABLE_SCHEDULER", "1")
```

Note that `True` is not a valid value in `os.environ`.

tests/unit_tests/shared/test_feature_toggling.py
```diff
@@ -0,0 +1,69 @@
...
+
+    assert not FeatureToggling.is_enabled("scheduler")
+
+    FeatureToggling.update_from_environment()
```
pinkwah, line 23: `FeatureToggling` is a singleton. How are you ensuring that the tests don't leak?

Hint: Use a `pytest.fixture` with `autouse=True` in this module that sets the values to some default.

tests/unit_tests/shared/test_feature_toggling.py
```diff
@@ -0,0 +1,69 @@
...
+    assert not FeatureToggling.is_enabled("new-storage")
+
+    FeatureToggling.update_from_args(parsed)
+
+    assert FeatureToggling.is_enabled("scheduler")
+    assert FeatureToggling.is_enabled("new-storage")
+
+
+def test_feature_toggling_both(monkeypatch):
+    envs = {"ERT_ENABLE_SCHEDULER": False}
```
pinkwah, line 49: Same comment about `monkeypatch.setenv`. Also, `False` is not a valid value (`os.environ` is `str -> str` mapping)

tests/unit_tests/shared/test_feature_toggling.py
```diff
@@ -0,0 +1,69 @@
...
+
+import ert.__main__
+from ert.__main__ import ert_parser
+from ert.cli import TEST_RUN_MODE
+from ert.shared.feature_toggling import FeatureToggling
+
+
+@pytest.fixture(autouse=True)
+def mocked_valid_file(monkeypatch):
+    monkeypatch.setattr(ert.__main__, "valid_file", Mock(return_value=""))
```
pinkwah, line 14: What's the purpose of this?

pinkwah (commented):
Can we also have a test where there are opposing enable and disable envvar and args?

Ie, if the user starts with `ERT_ENABLE_SCHEDULER=0` and `--enable-scheduler`. Here, the argument should take precedence to the environment variable.

pinkwah:
Maybe we can drop `ENABLE`/`DISABLE` from envvars. So, `ERT_SCHEDULER=1` (or any non-empty non-zero value) is explicit enable, `ERT_SCHEDULER=0` is explicit disable and `ERT_SCHEDULER=` (empty string) is the same as not setting the variable, ie. use default value.

tests/unit_tests/shared/test_feature_toggling.py
JHolba: ert_parser tries to validate that the ert config file exists. there is little point in making a temp file for this, so I override the test instead.

src/ert/shared/feature_toggling.py
JHolba: it is done that way on purpose to align with how update_from_args works. this way I could reuse all of update_from_args without making changes to it.

tests/unit_tests/shared/test_feature_toggling.py
JHolba: good catch

JHolba: I am setting all of environ to make sure that no other env vars are set. In case a developer may have set some ENABLE_/DISABLE_ in their profile or something.

src/ert/ensemble_evaluator/_builder/_legacy.py
```diff
@@ -45,7 +45,7 @@ def __init__(
         if not queue_config:
             raise ValueError(f"{self} needs queue_config")
 
-        if FeatureToggling.is_enabled("scheduler"):
+        if FeatureToggling.value("scheduler") == "new":
```
xjules, line 48: `"scheduler" == "new"` what's the idea behind these? I thought the `scheduler` is already `new` queue.

pinkwah:
I think this implements something we don't need at this time (and possibly ever). The solution where it was either `force on`, `force off` or `default` with the arguments `--enable-scheduler` and `--disable-scheduler` was the cleanest and is exactly to the point of what we wish to do.

I don't see the short-term usefulness of having an enum-based system. So unless there is something specific, this is a lot of code to do something that looks like a hack.

We all discussed and agreed that `--enable-scheduler` and `--disable-scheduler` is the way to go. Obviously things can change, but they need to be discussed.

xjules:
I think the PR can perhaps contain a description (in nutshell) what does it actually want to solve since I also was (in my mind at least) convinced that we had decided to go for the `--enable-scheduler` and `--disable-scheduler` flags only. 
Nevertheless, when looking at the original issue (https://github.com/equinor/ert/issues/6760) it is not really clear from there. @pinkwah could you maybe update the origin issue on:

> We all discussed and agreed that --enable-scheduler and --disable-scheduler is the way to go

if this represents essentially **two flags** that should be added - also in the form of env variable?

pinkwah:
I've amended the original issue with what we discussed a week ago.

I don't believe we talked about environment variables, though, so it's more my take on it: I don't think we need to have separate environment variables as they can fit all 3 states without much difficulty.
- Unset or empty: use default (eg. `ERT_SCHEDULER=`)
- Zero: force-disable (`ERT_SCHEDULER=0`)
- Non-zero, non-empty: force-enable (`ERT_SCHEDULER=1`)

tests/unit_tests/shared/test_feature_toggling.py
JHolba: this was fixed

src/ert/ensemble_evaluator/_builder/_legacy.py
JHolba: changed to True/False/None

tests/unit_tests/shared/test_feature_toggling.py
JHolba: fixed

tests/unit_tests/shared/test_feature_toggling.py
JHolba: fixed

JHolba:
The PR now uses --enable-scheduler --disable-scheduler.

I rewrote the env var handling before the comment from @pinkwah . Can change it again if someone feels strongly about this.
It uses ERT_FEATURE_SCHEDULER as env flag.
The env value is case insensitive.
true/1 enables scheduler.
false/0 disabled it.
empty string and "default" will set it to default.
all other values will be rejected.

src/ert/shared/feature_toggling.py
```diff
@@ -1,71 +1,132 @@
...
             ),
         ),
         "scheduler": _Feature(
-            default_enabled=False,
-            msg="Use Scheduler instead of JobQueue",
+            default=None,
+            msg="Force enable/disable the usage of the new scheduler.",
+            optional=True,
+        ),
+        "old-storage": _Feature(
```
pinkwah, line 57: There is no such thing as `old-storage` and `new-storage` has not been doing anything since b0ef3f5f6d8d92af79b3cc5e3769a93ab68c52a3 . You can remove this.

We can maybe make `FeatureToggling` not be a singleton for testing purposes. Eg:
```py
class FeatureToggling:
    def __init__(self, config: Mapping[str, _Feature]) -> None:
        self._config

    # [..]

FEATURE_TOGGLING = FeatureToggling({
    "scheduler": _Feature(...)
})

def is_enabled(feature: str) -> bool:
    return FEATURE_TOGGLING.is_enabled(feature)
```

Everywhere uses `is_enabled` function directly while feature toggling tests can test that eg. hyphenated names work correctly by creating a new instance of `FeatureToggling`.

pinkwah (approved):
Looks good! I have some comments that may be addressed, but that can also be done in a later PR.

eivindjahren:
I suggest simplifying and just using env vars. We don't need big machinery for this: https://github.com/equinor/ert/pull/6886

pinkwah:
> I suggest simplifying and just using env vars. We don't need big machinery for this: #6886

I agree about simplifying, I do not agree about env vars or the approach that your PR has taken. Jon's solution is more robust and user-friendly and we can simplify later. (In particular, we might not need `FeatureToggling` as a class, but it can be a function `def use_scheduler() -> bool`. But it cannot be a global variable as it would make it too difficult to systems-test.

eivindjahren:
> > I suggest simplifying and just using env vars. We don't need big machinery for this: #6886
> 
> I agree about simplifying, I do not agree about env vars or the approach that your PR has taken. Jon's solution is more robust and user-friendly and we can simplify later. (In particular, we might not need `FeatureToggling` as a class, but it can be a function `def use_scheduler() -> bool`. But it cannot be a global variable as it would make it too difficult to systems-test.

The issue of user-friendiness is moot as ideally no user will ever want or need to see this. Later if a user needs to opt out, they can get the exact syntax to do so from us.

I don't see how a global variable makes this difficult to system test or why this is not robust. If you want we can make a function out of it, but then you have to make sure to show validation only once.

src/ert/shared/feature_toggling.py
JHolba: code was rewritten. this is not relevant anymore

src/ert/shared/feature_toggling.py
JHolba: old-storage was something I was testing with. came back when I messed up a rebase/squash. removed new-storage and all code referencing it now.
Changing FeatureToggling to not be a singleton can be a different PR/issue

PR #6782, opened 2023-12-11
Add retry loop into the scheduler handling resubmitting when running state failed 

### LINKED ISSUE #6771: Implement retry functionality in the Scheduler
image

In the new Scheduler / Job class this should handle retry functionality when forward model fails (`RUNNING`) and we should go back to `WAITING` state.

xjules:
**Issue**
Resolves #6771 

**Approach**
Use while retry to iterate from running to waiting states. It includes a
simple test to check if job has started several times. Max_submit is a function
parameter of job.__call__ that is passed on from scheduler.

Additionally, function driver.finish will implement the basic clean up
functionally. For the local driver it makes sure that all tasks have
been awaited correctly.

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/scheduler/driver.py
+++ b/src/ert/scheduler/driver.py
@@ -3,10 +3,7 @@ from __future__ import annotations
 import asyncio
 from abc import ABC, abstractmethod
 from enum import Enum
-from typing import (
-    Optional,
-    Tuple,
-)
+from typing import Optional, Tuple
 
 
 class JobEvent(Enum):
@@ -48,3 +45,7 @@ class Driver(ABC):
     @abstractmethod
     async def poll(self) -> None:
         """Poll for new job events"""
+
+    @abstractmethod
+    async def finish(self) -> None:
+        """make sure that all the jobs / realizations are complete."""
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import logging
 from enum import Enum
 from typing import TYPE_CHECKING
 
@@ -15,6 +16,8 @@ from ert.scheduler.driver import Driver
 if TYPE_CHECKING:
     from ert.ensemble_evaluator._builder._realization import Realization
     from ert.scheduler.scheduler import Scheduler
+
+logger = logging.getLogger(__name__)
 
 
 class State(str, Enum):
@@ -62,12 +65,8 @@ class Job:
     def driver(self) -> Driver:
         return self._scheduler.driver
 
-    async def __call__(
-        self, start: asyncio.Event, sem: asyncio.BoundedSemaphore
-    ) -> None:
-        await start.wait()
+    async def _submit_and_run_once(self, sem: asyncio.BoundedSemaphore) -> None:
         await sem.acquire()
-
         try:
             await self._send(State.SUBMITTING)
             await self.driver.submit(
@@ -81,6 +80,7 @@ class Job:
             while not self.returncode.done():
                 await asyncio.sleep(0.01)
             returncode = await self.returncode
+
             if (
                 returncode == 0
                 and forward_model_ok(self.real.run_arg).status
@@ -89,6 +89,8 @@ class Job:
                 await self._send(State.COMPLETED)
             else:
                 await self._send(State.FAILED)
+                self.returncode = asyncio.Future()
+                self.started = asyncio.Event()
 
         except asyncio.CancelledError:
             await self._send(State.ABORTING)
@@ -98,6 +100,26 @@ class Job:
             await self._send(State.ABORTED)
         finally:
             sem.release()
+
+    async def __call__(
+        self, start: asyncio.Event, sem: asyncio.BoundedSemaphore, max_submit: int = 2
+    ) -> None:
+        await start.wait()
+
+        for _ in range(max_submit):
+            await self._submit_and_run_once(sem)
+
+            if self.returncode.done() or self.aborted.is_set():
+                break
+            else:
+                message = f"Realization: {self.iens} failed, resubmitting"
+                logger.warning(message)
+        else:
+            message = (
+                f"Realization: {self.iens} "
+                f"failed after reaching max submit {max_submit}"
+            )
+            logger.error(message)
 
     async def _send(self, state: State) -> None:
         status = STATE_TO_LEGACY[state]
--- a/src/ert/scheduler/local_driver.py
+++ b/src/ert/scheduler/local_driver.py
@@ -13,6 +13,7 @@ class LocalDriver(Driver):
         self._tasks: MutableMapping[int, asyncio.Task[None]] = {}
 
     async def submit(self, iens: int, executable: str, /, *args: str, cwd: str) -> None:
+        await self.kill(iens)
         self._tasks[iens] = asyncio.create_task(
             self._wait_until_finish(iens, executable, *args, cwd=cwd)
         )
@@ -20,8 +21,13 @@ class LocalDriver(Driver):
     async def kill(self, iens: int) -> None:
         try:
             self._tasks[iens].cancel()
+            await self._tasks[iens]
+            del self._tasks[iens]
         except KeyError:
             return
+
+    async def finish(self) -> None:
+        await asyncio.gather(*self._tasks.values())
 
     async def _wait_until_finish(
         self, iens: int, executable: str, /, *args: str, cwd: str
--- a/tests/unit_tests/scheduler/test_scheduler.py
+++ b/tests/unit_tests/scheduler/test_scheduler.py
@@ -1,7 +1,6 @@
 import asyncio
 import json
 import shutil
-from dataclasses import asdict
 from pathlib import Path
 from typing import Sequence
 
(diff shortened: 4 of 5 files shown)


tests/unit_tests/scheduler/test_scheduler.py
```diff
@@ -108,3 +107,14 @@ async def test_cancel(tmp_path: Path, realization):
 
     assert (tmp_path / "a").exists()
     assert not (tmp_path / "b").exists()
+
+
+async def test_that_max_submit_was_reached(tmp_path: Path, realization):
```
xjules, line 112: note to self: patch the logger and get the error message

src/ert/scheduler/job.py
```diff
@@ -66,36 +69,56 @@ async def __call__(
...
+                ):
+                    await self._send(State.COMPLETED)
+                else:
+                    await self._send(State.FAILED)
+                    retries += 1
+                    retry = retries < self._scheduler._max_submit
+                    if retry:
+                        message = f"Realization: {self.iens} failed, resubmitting"
+                        logger.warning(message)
+                        print(message)
```
berland, line 104: leftover

src/ert/scheduler/job.py
```diff
@@ -66,36 +69,56 @@ async def __call__(
...
+                if (
+                    returncode == 0
+                    and forward_model_ok(self.real.run_arg).status
+                    == LoadStatus.LOAD_SUCCESSFUL
+                ):
+                    await self._send(State.COMPLETED)
+                else:
+                    await self._send(State.FAILED)
+                    retries += 1
+                    retry = retries < self._scheduler._max_submit
```
berland, line 100: The scheduler should probably propagate this max_submit down to the Job instance upon creation so we don't need to jump up to fetch it.

src/ert/scheduler/job.py
```diff
@@ -63,41 +66,58 @@ def driver(self) -> Driver:
...
+
+                await self._send(State.STARTING)
+                await self.started.wait()
+
+                await self._send(State.RUNNING)
+                while not self.returncode.done():
+                    await asyncio.sleep(0.01)
+                returncode = await self.returncode
+                # we need to make sure that the task has finished too
+                await self.driver.wait(self.real.iens)
```
berland, line 91: Is it optimal to use the name `wait()` for this, as it looks like the same as the lines with `self.started.wait()`, while the latter is waiting for an asyncio.Event.

berland: Maybe a name sufficiently clear to avoid the comment.

src/ert/scheduler/job.py
```diff
@@ -63,41 +66,58 @@ def driver(self) -> Driver:
...
-            await self._send(State.ABORTING)
-            await self.driver.kill(self.iens)
-
-            await self.aborted.wait()
-            await self._send(State.ABORTED)
-        finally:
-            sem.release()
+        retries = 0
+        retry: bool = True
+        while retry:
```
pinkwah, line 74: Maybe separate the retry functionality into its own function? This is beginning to be a mess.

Eg:
```py
async def __call__(...):
    for _ in range(max_submit):
        if await self.actually_do_the_thing(...):
            # SUCCESS!
    else:
        # FAILURE (No more retries)
```

Also, using a normal loop will probably avoid having to do these `retries`, `retry` variables and manually checking whether you should escape.

src/ert/scheduler/job.py
```diff
@@ -63,41 +66,58 @@ def driver(self) -> Driver:
...
-        except asyncio.CancelledError:
-            await self._send(State.ABORTING)
-            await self.driver.kill(self.iens)
-
-            await self.aborted.wait()
-            await self._send(State.ABORTED)
-        finally:
-            sem.release()
+        retries = 0
+        retry: bool = True
```
pinkwah, line 73:
```suggestion
        retry = True
```

`True` has type `bool`, just like `0` has type `int`. No need to specify it explicitly.

pinkwah (changes requested):

src/ert/scheduler/local_driver.py
```diff
@@ -23,6 +23,9 @@ async def kill(self, iens: int) -> None:
         except KeyError:
             return
 
+    async def wait(self, iens: int) -> None:
```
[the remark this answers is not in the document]
xjules: It could be an Event that is triggered?

src/ert/scheduler/job.py
xjules: > Maybe separate the retry functionality into its own function? This is beginning to be a mess.

Agree, I saw it coming, will give it some tries to split it in 2.

src/ert/scheduler/job.py
xjules: > Why can't the LocalDriver itself do the right thing when the job is ended / restarted?

Since we want to have it non-blocking and thus we need to await out of the scope. The thing that Subprocess has ended does not guarantee that the task has ended too, ie. got failures on tests.

src/ert/scheduler/local_driver.py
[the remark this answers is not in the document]
pinkwah: I mean, what are we `wait` for in LSF? The job being completed/failed is already covered by `Job`'s `returncode`.

src/ert/scheduler/job.py
pinkwah: The fact that it's a subprocess is implementation detail. `await self.returncode` already means "wait for the job to end". Do we need another `wait` function to do this?

src/ert/scheduler/local_driver.py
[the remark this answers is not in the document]
xjules: For localdriver, this task was never awaited: 
```python
self._tasks[iens] = asyncio.create_task(
            self._wait_until_finish(iens, executable, *args, cwd=cwd)
        )
```

We need to take the task awaiting out from the driver to the Job level (ie. `await driver.wait()`) to have the driver atomic functions (init, kill, poll) non-blocking. For LSF :shrug: maybe empty then?

src/ert/scheduler/job.py
xjules: Correct, but as you said it's the job that has finished not the task. You can try to remove it and the test with retries will fail. I guess this `await task` does more to that async context.

src/ert/scheduler/job.py
xjules: Now it's split.

src/ert/scheduler/job.py
```diff
@@ -81,6 +80,9 @@ async def __call__(
             while not self.returncode.done():
                 await asyncio.sleep(0.01)
             returncode = await self.returncode
+            # we need to make sure that the task has finished too
+            await self.driver.wait(self.real.iens)
```
berland, line 84: I think the presence of a comment here means that the function in the driver should have a clearer name.

src/ert/scheduler/driver.py
```diff
@@ -48,3 +45,11 @@ async def kill(self, iens: int) -> None:
     @abstractmethod
     async def poll(self) -> None:
         """Poll for new job events"""
+
+    @abstractmethod
+    async def wait(self, iens: int) -> None:
+        """Blocks the execution of a job associated with a realization.
```
berland, line 51: I think it is too vague what is to be waited for by in implementations of this function. And should we use "block", as we never block anything, do we?

src/ert/scheduler/job.py
```diff
@@ -99,6 +103,27 @@ async def __call__(
...
             sem.release()
 
+    async def __call__(
+        self, start: asyncio.Event, sem: asyncio.BoundedSemaphore, max_submit: int = 2
+    ) -> None:
+        await start.wait()
+
+        for _ in range(max_submit):
+            await self._execute(sem)
+            # job has succesfully finished or was aborted
```
berland, line 113: don't need the comment, readable from code

src/ert/scheduler/job.py
```diff
@@ -62,12 +65,8 @@ def iens(self) -> int:
     def driver(self) -> Driver:
         return self._scheduler.driver
 
-    async def __call__(
-        self, start: asyncio.Event, sem: asyncio.BoundedSemaphore
-    ) -> None:
-        await start.wait()
+    async def _execute(self, sem: asyncio.BoundedSemaphore) -> None:
```
berland, line 68: hmm, maybe be more precise in the function name,  "_submit_and_run_once"?

tests/unit_tests/scheduler/test_scheduler.py
```diff
@@ -108,3 +107,16 @@ async def test_cancel(tmp_path: Path, realization):
...
```
berland, line 112: should we ensure that the edge_case max_submit=1 is covered?

(what about max_submit=0, is that covered by the config parser?)

src/ert/scheduler/job.py
xjules: Right, that was not clear that the name was an issue :)

src/ert/scheduler/driver.py
xjules: We do when we call block, but you are correct it's not the job execution but the event loop only  I guess

tests/unit_tests/scheduler/test_scheduler.py
xjules: Can do a parametriization ranging from 1 to 3? Regarding 0 that's a matter of config and not this test.

pinkwah (approved):
LGTM!

PR #6785, opened 2023-12-12
Support max_runtime in scheduler

### LINKED ISSUE #6784: Add support for MAX_RUNTIME in scheduler

berland:
**Issue**
Resolves #6784

**Approach**

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/ensemble_evaluator/_builder/_legacy.py
+++ b/src/ert/ensemble_evaluator/_builder/_legacy.py
@@ -87,6 +87,9 @@ class LegacyEnsemble(Ensemble):
         cloudevent_unary_send: Callable[[CloudEvent], Awaitable[None]],
         event_generator: Callable[[str, Optional[int]], CloudEvent],
     ) -> Tuple[Callable[[int], None], asyncio.Task[None]]:
+        """This function is reimplemented inside the Scheduler and should
+        be removed when Scheduler is the only queue code."""
+
         def on_timeout(iens: int) -> None:
             timeout_queue.put_nowait(
                 event_generator(identifiers.EVTYPE_REALIZATION_TIMEOUT, iens)
@@ -173,14 +176,16 @@ class LegacyEnsemble(Ensemble):
         is a function (or bound method) that only takes a CloudEvent as a positional
         argument.
         """
-        # Set up the timeout-mechanism
-        timeout_queue = asyncio.Queue()  # type: ignore
-        # Based on the experiment id the generator will
-        # give a function returning cloud event
         event_creator = self.generate_event_creator(experiment_id=experiment_id)
-        on_timeout, send_timeout_future = self.setup_timeout_callback(
-            timeout_queue, cloudevent_unary_send, event_creator
-        )
+        timeout_queue: Optional[asyncio.Queue[Any]] = None
+        if not FeatureToggling.is_enabled("scheduler"):
+            # Set up the timeout-mechanism
+            timeout_queue = asyncio.Queue()
+            # Based on the experiment id the generator will
+            # give a function returning cloud event
+            on_timeout, send_timeout_future = self.setup_timeout_callback(
+                timeout_queue, cloudevent_unary_send, event_creator
+            )
 
         if not self.id_:
             raise ValueError("Ensemble id not set")
@@ -235,8 +240,10 @@ class LegacyEnsemble(Ensemble):
             )
             result = identifiers.EVTYPE_ENSEMBLE_FAILED
 
-        await timeout_queue.put(None)  # signal to exit timer
-        await send_timeout_future
+        if not isinstance(self._job_queue, Scheduler):
+            assert timeout_queue is not None
+            await timeout_queue.put(None)  # signal to exit timer
+            await send_timeout_future
 
         # Dispatch final result from evaluator - FAILED, CANCEL or STOPPED
         await cloudevent_unary_send(event_creator(result, None))
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -2,13 +2,15 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import uuid
 from enum import Enum
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
 
 from cloudevents.conversion import to_json
 from cloudevents.http import CloudEvent
 
 from ert.callbacks import forward_model_ok
+from ert.ensemble_evaluator.identifiers import EVTYPE_REALIZATION_TIMEOUT
 from ert.job_queue.queue import _queue_state_event_type
 from ert.load_status import LoadStatus
 from ert.scheduler.driver import Driver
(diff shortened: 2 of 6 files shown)


src/ert/scheduler/job.py
```diff
@@ -47,12 +47,18 @@ class Job:
...
+        scheduler: Scheduler,
+        real: Realization,
+        callback_timeout: Optional[Callable[[int], None]] = None,
+    ) -> None:
         self.real = real
         self.started = asyncio.Event()
         self.returncode: asyncio.Future[int] = asyncio.Future()
         self.aborted = asyncio.Event()
         self._scheduler = scheduler
+        self._callback_timeout: Optional[Callable[[int], None]] = callback_timeout
```
xjules, line 61: Do we need this function as an attribute?

src/ert/scheduler/job.py
```diff
@@ -78,6 +85,8 @@ async def __call__(
             await self.started.wait()
 
             await self._send(State.RUNNING)
+            if self.real.max_runtime is not None and self.real.max_runtime > 0:
+                cancel_task = asyncio.create_task(self._cancel_from_timeout())
```
xjules, line 89: instead of `cancel_task` maybe `max_runtime_tasl`?

src/ert/scheduler/job.py
berland: The timeout is set up by `_legacy.py` and we need to keep in sync with how the existing job_queue operates.

pinkwah: Maybe we can have a
```py
if isinstance(self._job_queue, Scheduler):
```
in `_legacy.py`? I reckon that is less dirty than having this `callback_timeout`.

src/ert/scheduler/job.py
```diff
@@ -67,6 +73,7 @@ async def __call__(
     ) -> None:
         await start.wait()
         await sem.acquire()
+        cancel_task: Optional[asyncio.Task[None]] = None
```
pinkwah, line 76: Could we use `asyncio.wait_for` instead of creating our own cancel task?

src/ert/scheduler/job.py
```diff
@@ -78,6 +85,8 @@ async def __call__(
             await self.started.wait()
 
             await self._send(State.RUNNING)
+            if self.real.max_runtime is not None and self.real.max_runtime > 0:
+                cancel_task = asyncio.create_task(self._max_runtime_task())
```
pinkwah, line 89: This is a sidenote and I think `asyncio.wait_for` is better solution overall, but I discovered that you can do `asyncio.current_task()` to get the current `asyncio.Task`, which you can pass to a cancel task.

src/ert/scheduler/job.py
berland: Fixed!

src/ert/scheduler/job.py
berland: It looks like `wait_for` is non-trivial because of the ugly:
```
            while not self.returncode.done():
                await asyncio.sleep(0.01)
            returncode = await self.returncode
```
which we then would have to wrap in yet another function (?)

src/ert/ensemble_evaluator/_builder/_legacy.py
```diff
@@ -165,14 +168,15 @@ async def _evaluate_inner(  # pylint: disable=too-many-branches
...
-        timeout_queue = asyncio.Queue()  # type: ignore
-        # Based on the experiment id the generator will
-        # give a function returning cloud event
-        event_creator = self.generate_event_creator(experiment_id=experiment_id)
-        on_timeout, send_timeout_future = self.setup_timeout_callback(
-            timeout_queue, cloudevent_unary_send, event_creator
-        )
+        if not isinstance(self._job_queue, Scheduler):
+            # Set up the timeout-mechanism
+            timeout_queue = asyncio.Queue()  # type: ignore
```
xjules, line 173: This is a bit puzzling. Does not `on_timeout` require `timeout_queue`?
I am asking because we don't create it for scheduler:
```python
if not isinstance(self._job_queue, Scheduler):
            # Set up the timeout-mechanism
            timeout_queue = asyncio.Queue()  # type: ignore
```

src/ert/scheduler/job.py
```diff
@@ -78,9 +87,32 @@ async def __call__(
...
+                    await asyncio.sleep(0.01)
+                returncode = await future
+                return returncode
+
+            try:
+                returncode = await asyncio.wait_for(
+                    threadsafer_future_wait(self.returncode),
+                    timeout=self.real.max_runtime,
+                )
+            except asyncio.exceptions.TimeoutError:
```
xjules, line 102: `asyncio.TimeoutError`

src/ert/scheduler/job.py
```diff
@@ -78,9 +87,32 @@ async def __call__(
             await self.started.wait()
 
             await self._send(State.RUNNING)
-            while not self.returncode.done():
-                await asyncio.sleep(0.01)
-            returncode = await self.returncode
+
+            async def threadsafer_future_wait(future: asyncio.Future[_T]) -> _T:
+                while not future.done():
```
xjules, line 92: I think this should be:
```python
try:
    while not future.done():
            await asyncio.sleep(0.01)
        returncode = await future
        return returncode
    timeout_event = CloudEvent(
                    {
                        "type": EVTYPE_REALIZATION_TIMEOUT,
                        "source": f"/ert/ensemble/{self._scheduler._ens_id}/real/{self.iens}",
                        "id": str(uuid.uuid1()),
                    }
                )
    assert self._scheduler._events is not None
    await self._scheduler._events.put(to_json(timeout_event))

```
`wait_for` automatically cancells the task that was waited_for.

src/ert/ensemble_evaluator/_builder/_legacy.py
berland: No, it is not needed in `_legacy.py` when we are doing the scheduler thing, since we move the entire thing down to `job.py` (and thereby always try to send the event even if `_legacy.py` has not asked for it)

src/ert/scheduler/job.py
berland: you want the except clause to happen inside the `threadsafer_future_wait` or do you want to avoid `threadsafer_future_wait`?

xjules: It depends what we want to achieve on the cancellation due to the `max_runtime`. Based on the docs on `wait_for`, firstly the task gets cancelled, then the cancellation errors gets propagated and turns into the `TimeoutError`.

xjules:
> I added a commit that refactored the PR to use asyncio.wait_for, but I am not sure I liked that alternative better than the _max_runtime_task.

When studying both approaches, I'm bit more inclined (since it cleaner tbh) towards having a separate task that just asyncio.sleeps and cancel jobs in the end. Also wait_for might be just another version of the same approach anyway.

pinkwah (approved):
:tada:

PR #6787, opened 2023-12-12
Have integration tests run with both scheduler and job queue

### LINKED ISSUE #6756: Identify tests that can be run for both queues, eg. using feature toggling
We should prioritize those tests that use LOCAL driver.

Goal: the tests that can be run using both queues (ie. feature toggled) shall be moved to `tests/test_integration` module. The name `test_integration` can be replaced by something more descriptive.

jonathan-eq:
**Issue**
Resolves #6756

Add pytest scheduler custom mark and fixture, indicating that an integration test will be ran with both old job queue and new scheduler. 
The commit in this PR also moves all integration tests (atleast those marked with integration_test) to a new integration test directory.


tests/conftest.py
```diff
@@ -234,13 +234,38 @@ def excepthook(cls, exc, tb):
...
 
 
+def pytest_generate_tests(metafunc):
+    print(f"{metafunc=}")
+    if "scheduler" in metafunc.definition.own_markers:
+        print("FOUND SET_ENABLE_SCHEDULER")
+        metafunc.fixturenames.insert(0, "set_enable_scheduler")
+        metafunc.parametrize("set_enable_scheduler", [True, False], indirect=True)
+
+@pytest.fixture(params=[True, False])
```
jonathan-eq, line 244: This works until you add the parms for the fixture

berland:
Recommend to do `pre-commit install` in your local project folder, and  it will be harder to miss on style issues :)

berland:
> The commit in this PR also moves all integration tests (atleast those marked with integration_test) to a new integration test directory.

Should the moving around of tests be in a separate commit?

jonathan-eq:
I only moved parts of the files, but that is by far the main part of the commit. I think I would reword the commit message instead

berland:
For the tests that use the pytest snapshot feature, you need to `git mv` the snapshot files as well.

pyproject.toml
```diff
@@ -150,6 +150,7 @@ markers = [
     "script",
     "slow",
     "unstable",
+    "scheduler"
```
berland, line 153: Use the opportunity to sort this list :)

tests/conftest.py
```diff
@@ -245,6 +245,20 @@ def excepthook(cls, exc, tb):
     monkeypatch.setattr(sys, "excepthook", excepthook)
 
 
+@pytest.fixture(params=[False, True])
+def try_queue_and_scheduler(request, monkeypatch):
```
berland, line 249: For some reason, the test that is being decorated with this fixture also needs to specify the `monkeypatch` fixture. Don't know why (otherwise it will silently ignore the request to enable scheduler).

tests/integration_tests/analysis/test_es_update.py
```diff
@@ -102,9 +103,10 @@ def test_that_posterior_has_lower_variance_than_prior(
     )
 
 
+@pytest.mark.scheduler
 @pytest.mark.integration_test
 def test_that_surfaces_retain_their_order_when_loaded_and_saved_by_ert(
-    copy_case
+    copy_case, try_queue_and_scheduler
```
berland, line 109: monkeypatch needed here?

jonathan-eq: I forgot to commit the changes...

berland:
Still two snapshots missing?

xjules:
There are snapshots to be updated @jonathan-eq

jonathan-eq:
Yes, it seems they updated ES-MDA so I will wait until they update the snapshots

PR #6788, opened 2023-12-12
Implement MockDriver

pinkwah:
LocalDriver differs from the HPC drivers in that it is made of three
parts that are run in sequence. `init` the subprocess, `wait` for the
process to complete and `kill` when the user wants to cancel. Between
the three parts the driver needs to send `JobEvent`s to the `Scheduler`.

This commit implements a `MockDriver`, where the user can optionally
specify a simplified version of each of `init`, `wait` or `kill`,
depending on what they wish to do.

### DIFF
--- a/src/ert/scheduler/local_driver.py
+++ b/src/ert/scheduler/local_driver.py
@@ -2,9 +2,14 @@ from __future__ import annotations
 
 import asyncio
 import os
-from typing import MutableMapping
+from asyncio.subprocess import Process
+from typing import (
+    MutableMapping,
+)
 
 from ert.scheduler.driver import Driver, JobEvent
+
+_TERMINATE_TIMEOUT = 10.0
 
 
 class LocalDriver(Driver):
@@ -15,7 +20,7 @@ class LocalDriver(Driver):
     async def submit(self, iens: int, executable: str, /, *args: str, cwd: str) -> None:
         await self.kill(iens)
         self._tasks[iens] = asyncio.create_task(
-            self._wait_until_finish(iens, executable, *args, cwd=cwd)
+            self._run(iens, executable, *args, cwd=cwd)
         )
 
     async def kill(self, iens: int) -> None:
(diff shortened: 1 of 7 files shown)


src/ert/scheduler/local_driver.py
```diff
@@ -23,26 +66,81 @@ async def kill(self, iens: int) -> None:
...
+    async def _wait(proc: Process) -> bool:
+        return await proc.wait() == 0
+
+    @staticmethod
+    async def _kill(proc: Process) -> None:
+        proc.terminate()
+        await proc.wait()
+
+
+class MockDriver(TaskDriver[None]):
```
berland, line 115: Could/should we move this class to the tests? As in `conftest.py` or just at the top of the driver test files?

tests/unit_tests/scheduler/test_scheduler.py
```diff
@@ -73,38 +43,48 @@ async def test_empty():
...
-    sch.add_realization(realization, callback_timeout=lambda _: None)
+    async def init(iens, exec, *args, cwd):
+        future.set_result(iens)
 
-    create_jobs_json(tmp_path, [step])
-    sch.add_dispatch_information_to_jobs_file()
+    driver = MockDriver(init=init)
+
+    sch = scheduler.Scheduler(driver)
+    sch.add_realization(realization, callback_timeout=lambda _: None)
```
berland, line 55: can remove `callback_timeout` from this function call.

tests/unit_tests/scheduler/test_local_driver.py
```diff
@@ -0,0 +1,72 @@
...
+    trap "" 1 2 3 4 5 6 7 8 10 11 12 13 14 15
+    sleep 10
+    """
+    )
+
+    driver = LocalDriver()
+
+    await driver.submit(42, "/bin/sh", tmp_path / "script", cwd=tmp_path)
+    assert await driver.event_queue.get() == (42, JobEvent.STARTED)
+    await driver.kill(42)
```
berland, line 58: Should we have some timout code here to assert that we are not at anytime ending waiting for 10 seconds for this test to pass?

src/ert/scheduler/local_driver.py
```diff
@@ -2,19 +2,62 @@
...
+        self, iens: int, executable: str, /, *args: str, cwd: str
+    ) -> _T_co:
+        ...
+
+
+_WaitFn = Callable[[_T_contra], Awaitable[bool]]
+_KillFn = Callable[[_T_contra], Awaitable[None]]
+
+
+class TaskDriver(Driver, Generic[_T]):
```
xjules, line 35: I'm bit lost what's this `_T` template. Can it be anything else than `asyncio.subprocess.Process`? Regardless the driver we use we need to execute bsub via the `Process`? Or is it about hiding the polling mechanism in a Process wrapper?

pinkwah: It can in the case of `MockDriver`, where it is `None`. The `TaskDriver` has no polling. LFS and co will likely inherit directly from `Driver`.

src/ert/scheduler/local_driver.py
pinkwah: My concern is that this will make us lose type information that may be useful for people writing tests. But I don't really know.

tests/unit_tests/scheduler/test_local_driver.py
berland: pytest.timeout for 9 seconds might suffice?

src/ert/scheduler/local_driver.py
berland: We can change name from "Mock" to something else, as it is a "functional" driver, but parametrized. "ParametrizedDriver", "ZeroDriver". Avoiding mock makes it more natural for having it in the source.

tests/unit_tests/scheduler/test_local_driver.py
```diff
@@ -0,0 +1,72 @@
...
+async def test_success(tmp_path):
+    driver = LocalDriver()
+
+    # LocalDriver doesn't have a polling task
+    assert driver.create_poll_task() is None
+
+    await driver.submit(42, "/usr/bin/env", "touch", "testfile", cwd=tmp_path)
+    assert await driver.event_queue.get() == (42, JobEvent.STARTED)
+    assert await driver.event_queue.get() == (42, JobEvent.COMPLETED)
+
```
berland, line 18: I like these test functions! 💯

tests/unit_tests/scheduler/test_local_driver.py
pinkwah: Oh, I didn't notice you were referring to this specific test. Yes, I added `pytest.mark.timeout` and this test fails now, so it was indeed waiting the entire duration. :rofl:

berland: (that is strange, as I tested it and then it worked fine!)

src/ert/scheduler/local_driver.py
berland: "EmptyDriver" also popped out from my head.

berland (approved):
👌🏻

PR #6790, opened 2023-12-12
Add "done_callback" to every task for debugging

pinkwah:
Currently, tasks that throw exceptions silently stop. This commit makes it so that all tasks are created with a done_callback that checks whether they stopped due to an exception.


berland (approved):
Looks good!

PR #6798, opened 2023-12-13
Add `background_tasks` for canceling tasks

pinkwah:
This contextmanager makes it simpler to ensure that infinitely-running tasks get canceled when appropriate, even when an exception occurs.


src/ert/scheduler/scheduler.py
```diff
@@ -122,23 +123,19 @@ async def execute(
         if queue_evaluators is not None:
             logger.warning(f"Ignoring queue_evaluators: {queue_evaluators}")
 
-        publisher_task = asyncio.create_task(self._publisher())
-        poller_task = self.driver.create_poll_task()
-        event_queue_task = asyncio.create_task(self._process_event_queue())
+        async with background_tasks() as bt:
```
berland, line 126: I am not a fan of two-letter shortenings of objects, either avoid or use the opportunity to describe the code even more precisely, perhaps `as cancel_when_execute_is_done:`

berland (approved):
👌🏻

PR #6803, opened 2023-12-14
Test ensemble_evaluator with new scheduler

berland:
Adds testing and fixes some problems. 

Currently it allows for a change in events emitted by the local driver from legacy to new. This is up for discussion.

Should be merged after #6787 as code is stolen therefrom

**Approach**
Test unit_tests/ensemble_evaluator with scheduler using new fixture.

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -23,7 +23,7 @@ logger = logging.getLogger(__name__)
 class State(str, Enum):
     WAITING = "WAITING"
     SUBMITTING = "SUBMITTING"
-    STARTING = "STARTING"
+    PENDING = "PENDING"
     RUNNING = "RUNNING"
     ABORTING = "ABORTING"
     COMPLETED = "COMPLETED"
@@ -34,7 +34,7 @@ class State(str, Enum):
 STATE_TO_LEGACY = {
     State.WAITING: "WAITING",
     State.SUBMITTING: "SUBMITTED",
-    State.STARTING: "PENDING",
+    State.PENDING: "PENDING",
     State.RUNNING: "RUNNING",
     State.ABORTING: "DO_KILL",
     State.COMPLETED: "SUCCESS",
@@ -73,7 +73,7 @@ class Job:
                 self.real.iens, self.real.job_script, cwd=self.real.run_arg.runpath
             )
 
-            await self._send(State.STARTING)
+            await self._send(State.PENDING)
             await self.started.wait()
 
             await self._send(State.RUNNING)
@@ -126,7 +126,7 @@ class Job:
         event = CloudEvent(
             {
                 "type": _queue_state_event_type(status),
-                "source": f"/etc/ensemble/{self._scheduler._ens_id}/real/{self.iens}",
+                "source": f"/ert/ensemble/{self._scheduler._ens_id}/real/{self.iens}",
                 "datacontenttype": "application/json",
             },
             {
--- a/tests/unit_tests/ensemble_evaluator/test_async_queue_execution.py
+++ b/tests/unit_tests/ensemble_evaluator/test_async_queue_execution.py
@@ -9,6 +9,8 @@ from websockets.server import serve
 from ert.async_utils import get_event_loop
 from ert.ensemble_evaluator._wait_for_evaluator import wait_for_evaluator
 from ert.job_queue import Driver, JobQueue
+from ert.scheduler import Scheduler, create_driver
+from ert.shared.feature_toggling import FeatureToggling
 
 
 async def mock_ws(host, port, done):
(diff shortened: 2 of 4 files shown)


tests/unit_tests/ensemble_evaluator/test_async_queue_execution.py
```diff
@@ -65,12 +74,21 @@ async def test_happy_path(
...
-    event_0 = from_json(mock_ws_task.result()[0])
-    assert event_0["source"] == "/ert/ensemble/ee_0/real/0"
-    assert event_0["type"] == "com.equinor.ert.realization.waiting"
-    assert event_0.data == {"queue_event_type": "WAITING"}
-
-    end_event_index = len(mock_ws_task.result()) - 1
-    end_event = from_json(mock_ws_task.result()[end_event_index])
-    assert end_event["type"] == "com.equinor.ert.realization.success"
-    assert end_event.data == {"queue_event_type": "SUCCESS"}
+    if FeatureToggling.is_enabled("scheduler"):
```
berland, line 77: Look here for the change of events emitted by the new local driver. Also note that "waiting" is translated to "SUBMITTED", but not for the legacy driver/job_queue_node(?).

src/ert/scheduler/job.py
```diff
@@ -104,7 +104,7 @@ async def _send(self, state: State) -> None:
         event = CloudEvent(
             {
                 "type": _queue_state_event_type(status),
-                "source": f"/etc/ensemble/{self._scheduler._ens_id}/real/{self.iens}",
+                "source": f"/ert/ensemble/{self._scheduler._ens_id}/real/{self.iens}",
```
pinkwah, line 107: No way did I really write `etc` and not notice??? :sweat_smile:

berland: Your finger muscle memory is forgiven

src/ert/scheduler/job.py
```diff
@@ -73,7 +73,7 @@ async def _submit_and_run_once(self, sem: asyncio.BoundedSemaphore) -> None:
                 self.real.iens, self.real.job_script, cwd=self.real.run_arg.runpath
             )
 
-            await self._send(State.STARTING)
+            await self._send(State.STARTING)  # aka PENDING
```
pinkwah, line 76: Is `PENDING` a better name? Should we change it back so that we don't need this comment?

tests/unit_tests/ensemble_evaluator/test_async_queue_execution.py
pinkwah: I think this is because the initial state is assumed to be `WAITING` in Scheduler, so the first new event is `SUBMITTED`, while JobQueue starts with `NOT_ACTIVE`, so the first event is `WAITING`.

I don't see the need for two different states that both say "we're not doing anything yet". Maybe we should consider removing `NOT_ACTIVE` from JobQueue to avoid this `if`? Alternatively, Scheduler might send `WAITING` explicitly.

src/ert/scheduler/job.py
berland: Perhaps not better by itself, but more familiar for legacy ert devs..

berland: Will change back to PENDING. I will claim it means exactly the same, and also makes as much sense as STARTING for the LocalDriver.

tests/unit_tests/ensemble_evaluator/test_async_queue_execution.py
berland: I am fine with differing behaviour for now, so we can reduce the number of states while transitioning to Scheduler. Lets reconsider if a case pops up where we need to distuingish between NOT_ACTIVE and WAITING in Scheduler.

tests/unit_tests/ensemble_evaluator/test_ensemble_legacy.py
```diff
@@ -99,7 +89,11 @@ def _test_run_and_cancel_legacy_ensemble(tmpdir, make_ensemble_builder, monkeypa
 
 
 @pytest.mark.timeout(10)
-def test_run_legacy_ensemble_exception(tmpdir, make_ensemble_builder, monkeypatch):
+def test_run_legacy_ensemble_with_bare_exception(
+    tmpdir, make_ensemble_builder, monkeypatch
+):
+    """This test function is not ported to Scheduler, as it will not
+    catch general exceptions."""
```
berland, line 96: The logs have been searched through, and the general exception that the legacy queue can catch has not happened the last 3 months.

pinkwah (approved):
:100:

PR #6805, opened 2023-12-14
Add fixture for running tests with scheduler and job queue

jonathan-eq:
Adds fixture for running integration tests with both scheduler and job queue. The fixture might be dependent on the test having the monkeypatch fixture explicitly used.
The commit also adds a marker for scheduler tests. The marker can take skip=True as a parameter, to only run the test with job queue.


berland (approved):
perfect!

PR #6816, opened 2023-12-15
Implement adaptive batch size

dafeda:
The `batch_size` has so far been hard-coded to `1000`.
I chose this value based on the memory available while working on Troll.
This was before we implemented the streaming algorithm and there was not much available memory.
A larger batch size means fewer loops and better performance so we want it to be as large as we can fit in memory.
This PR proposes a batch size that is a function of available memory that should increase performance quite a bit.

## Benchmarks

Drogon run (with seismic):
Ran twice, once manually and once as part of an `Ensemble Smoother` run.
In both cases updates took approximately 1h 30min, which is down from approx. 10h with a fixed batch size of 1000.

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -135,6 +135,7 @@ types = [
     "types-decorator",
     "types-docutils",
     "types-tqdm",
+    "types-psutil"
 ]
 
 [tool.setuptools]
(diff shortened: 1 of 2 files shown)


oyvindeide (approved):
Good job!

Issue #6779, opened 2023-12-11
Resolve overloaded usage of the term Job in new Scheduler vs Snapshot


berland:
https://github.com/equinor/ert/blob/650c5cae21001dc7121528a7826a2c51f5dd807c/src/ert/scheduler/job.py#L43

while we still have 

https://github.com/equinor/ert/blob/35d1984b1fd055ded396a24c0f77bfe9bd585676/src/ert/ensemble_evaluator/snapshot.py#L377

which is a different kind of "Job".

berland:
Suggestion: "ExecutingRealization"

pinkwah:
I don't understand the dislike of the term "job". The `Job` class is a representation of the thing one "submits" and "kills" on the cluster. Per the documentation:
- LSF `bsub` ["Submits a **job** to LSF [..]"](https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=SSWRJV_10.1.0/lsf_command_ref/bsub.man_top.1.htm)
- LSF `bkill` ["Sends signals to kill, suspend, or resume unfinished **jobs**"](https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=reference-bkill)
- PBS `qsub` ["To submit a script is to create a batch **job** that executes the script."](https://linuxcommandlibrary.com/man/qsub#description)
- PBS `qdel` ["The qdel command deletes **jobs** [..]"](https://manpages.org/qdel)
- SLURM `sbatch` ["Option(s) define multiple **jobs** in a co-scheduled heterogeneous **job**. For more details about heterogeneous **jobs** see the document"](https://slurm.schedmd.com/sbatch.html#SECTION_SYNOPSIS)
- SLURM `scancel` ["Used to signal **jobs** or **job** steps that are under the control of Slurm."](https://slurm.schedmd.com/scancel.html#SECTION_NAME)

LSF is a "workload management platform and **job scheduler** for distributed high performance computing (HPC)." The Wikipedia article is called [Job scheduler](https://en.wikipedia.org/wiki/Job_scheduler). The other common term is "batch processing" I guess. A forward model would be a "batch script"?

I wouldn't be this anal about it if "job" wasn't such a widespread and understood concept within the realm of HPC. We ought really not reinvent terms if there are fitting existing terms for it already, and `Job` is the perfect term for this concept.

berland:
[...]

PR #6829, opened 2023-12-19
Refactor JobQueue and Scheduler

pinkwah:
Initialisation of `JobQueue` and `Scheduler` is moved to where they are used.
This means that the current event loop is the same during initialisation
and use.

Changes:
- `add_realization`: We can simply pass realisations as a list to the
  given executor.
- `set_ee_info`: We can simply pass this information to the queues.
- `CONCURRENT_INITIALIZATION`: No longer passed to the queue as it's a
  constant.
- `timeout_callback`: Now passed to the queue, rather than with each realisation.
- `queue_evaluators`: Concept removed. Only used for
  "min_required_realizations", which we can call directly.

### DIFF
--- a/src/ert/config/queue_config.py
+++ b/src/ert/config/queue_config.py
@@ -95,6 +95,13 @@ class QueueConfig:
             self.queue_options,
         )
 
+    @property
+    def max_running(self) -> int:
+        for key, val in self.queue_options.get(self.queue_system, []):
+            if key == "MAX_RUNNING":
+                return int(val)
+        return 0
+
 
 def _check_for_overwritten_queue_system_options(
     selected_queue_system: QueueSystem,
--- a/src/ert/ensemble_evaluator/_builder/_legacy.py
+++ b/src/ert/ensemble_evaluator/_builder/_legacy.py
@@ -4,8 +4,18 @@ import asyncio
 import logging
 import threading
 import uuid
-from functools import partial, partialmethod
-from typing import TYPE_CHECKING, Any, Awaitable, Callable, Dict, List, Optional, Tuple
+from functools import partialmethod
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Awaitable,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Protocol,
+    Tuple,
+)
 
 from cloudevents.http.event import CloudEvent
 
@@ -31,6 +41,11 @@ logger = logging.getLogger(__name__)
 event_logger = logging.getLogger("ert.event_log")
 
 
+class _KillAllJobs(Protocol):
+    def kill_all_jobs(self) -> None:
+        ...
+
+
 class LegacyEnsemble(Ensemble):
     def __init__(
         self,
@@ -42,16 +57,9 @@ class LegacyEnsemble(Ensemble):
         id_: str,
     ) -> None:
         super().__init__(reals, metadata, id_)
-        if not queue_config:
-            raise ValueError(f"{self} needs queue_config")
 
-        if FeatureToggling.is_enabled("scheduler"):
-            if queue_config.queue_system != QueueSystem.LOCAL:
-                raise NotImplementedError()
-            driver = create_driver(queue_config)
-            self._job_queue = Scheduler(driver)
-        else:
-            self._job_queue = JobQueue(queue_config)
+        self._queue_config = queue_config
+        self._job_queue: Optional[_KillAllJobs] = None
         self.stop_long_running = stop_long_running
         self.min_required_realizations = min_required_realizations
         self._config: Optional[EvaluatorServerConfig] = None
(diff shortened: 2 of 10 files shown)


src/ert/config/queue_config.py
```diff
@@ -95,6 +95,13 @@ def create_local_copy(self) -> QueueConfig:
             self.queue_options,
         )
 
+    @property
+    def max_running(self) -> int:
+        for key, val in self.queue_options.get(self.queue_system, []):
+            if key == "MAX_RUNNING":
+                return int(val)
+        return 1
```
berland, line 103: Why 1? Docs says that zero is default and it will imply no limit (doc states that max_running is equal to number of realizations)

berland: https://github.com/equinor/ert/pull/6834

src/ert/ensemble_evaluator/_builder/_legacy.py
```diff
@@ -31,6 +41,11 @@
 event_logger = logging.getLogger("ert.event_log")
 
 
+class _JobQueue(Protocol):
```
berland, line 44: Oh no, yet another usage of "Job"... 😢 https://github.com/equinor/ert/issues/6779

src/ert/ensemble_evaluator/_builder/_legacy.py
```diff
@@ -180,43 +188,48 @@ async def _evaluate_inner(  # pylint: disable=too-many-branches
...
+                # Tell queue to pass info to the jobs-file
+                # NOTE: This touches files on disk...
+                queue.add_dispatch_information_to_jobs_file()
 
-            sema = threading.BoundedSemaphore(value=CONCURRENT_INTERNALIZATION)
-            result: str = await self._job_queue.execute(
-                sema,
-                queue_evaluators,
-            )
+                print(f"{min_required_realizations=}")
```
berland, line 231: leftover

src/ert/config/queue_config.py
pinkwah: I wasn't sure so I put a different number temporarily. 0 makes sense (maybe num CPUs for LOCAL driver?)

src/ert/simulator/simulation_context.py
```diff
@@ -61,19 +60,9 @@ def _run_forward_model(
...
-        queue_evaluators = [
-            partial(
-                job_queue.stop_long_running_jobs,
-                ert.ert_config.analysis_config.minimum_required_realizations,
-            )
-        ]
-
-    asyncio.run(job_queue.execute(evaluators=queue_evaluators))
+    asyncio.run(
+        job_queue.execute(ert.ert_config.analysis_config.minimum_required_realizations)
```
berland, line 64: Do we want to remove the stop_long_running feature from everest?

tests/unit_tests/status/test_tracking_integration.py
```diff
@@ -40,6 +40,12 @@ def check_expression(original, path_expression, expected, msg_start):
     match_found = False
     for match in jsonpath_expr.find(original):
         match_found = True
+
+        if match.value != expected:
+            import pdb
+
+            pdb.set_trace()
```
berland, line 47: ah, this is not in ready for review yet :)

src/ert/ensemble_evaluator/_builder/_legacy.py
pinkwah: It's a `Protocol` class for `JobQueue`. I don't believe I need it though.

src/ert/simulator/simulation_context.py
pinkwah: Nope. Mistake!

tests/unit_tests/status/test_tracking_integration.py
pinkwah: ;)

src/ert/ensemble_evaluator/_builder/_legacy.py
pinkwah: Renamed to `_KillAllJobs` since that is the only thing it does.

src/ert/job_queue/queue.py
```diff
@@ -102,7 +100,18 @@ def __repr__(self) -> str:
     def __str__(self) -> str:
         return self.__repr__()
 
-    def __init__(self, queue_config: QueueConfig):
+    def __init__(
+        self,
+        queue_config: QueueConfig,
+        realizations: Optional[Sequence[Realization]] = None,
+        *,
```
berland, line 107: Should not this star be one line up, as `realizations` looks like a named argument to me?

berland: If it forces realizations to be specified before ens_id, it is perfectly fine with me.

berland (approved):
It does look less stupid now!

PR #6837, opened 2023-12-21
Rename Job to ForwardModel in Snapshot

### LINKED ISSUE #6779: Resolve overloaded usage of the term Job in new Scheduler vs Snapshot

berland:
**Issue**
Resolves #6779 

**Approach**

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist


berland:
I think this one is ready. There are certainly more "job" stuff lying around, but not feasible put everything in one PR.

I suggest merging, not squashing. In case something goes wrong, bisecting the commits in this PR might be beneficial.

pinkwah (approved):
It might be that this should rather be called "Steps", but this is definitely an improvement over what was. LGTM

berland:
> It might be that this should rather be called "Steps", but this is definitely an improvement over what was. LGTM

What is called "ForwardModel" in this PR is in line with what the user writes in their config files, one line with a `FORWARD_MODEL` for every thing they need to run. This is where ERT has its own definition of a forward model opposed to the literature on Ensemble methods. In the latter there is only one "forward model" for a realization (the simulation that brings your data forward in time), while in ERT each realization consists of multiple forward models (you could say that the term "step" is implicit).

Issue #6773, opened 2023-12-08
Run all integration tests with both queues for the LocalDriver


xjules:
The goal is to run all ert tests with both implementations of the LocalDriver and Queues.

Trick to expose these tests is to do `raise ValueError` or something in `queue.py::__init__()`. All tests that then fail should get the fixture `try_queue_and_scheduler` applied.
---

List of issues related to running those tests is coming here:

PR #6852, opened 2023-12-28
Fix scheduler tests with tracking and failure messages

### LINKED ISSUE #6789: Add run_done_callback to scheduler
The "run_done_callback" is in `job.py` but only as an MVP. The output from `forward_model_ok` is not handled correctly yet.

berland:
**Issue**
Resolves #6789 and #6773 

**Approach**
See commit messages.

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -63,6 +63,7 @@ class Job:
         self.aborted = asyncio.Event()
         self._scheduler: Scheduler = scheduler
         self._callback_status_msg: str = ""
+        self._requested_max_submit: Optional[int] = None
 
     @property
     def iens(self) -> int:
@@ -103,9 +104,11 @@ class Job:
 
                 if callback_status == LoadStatus.LOAD_SUCCESSFUL:
                     await self._send(State.COMPLETED)
-                elif callback_status == LoadStatus.TIME_MAP_FAILURE:
-                    await self._send(State.FAILED)
-                else:  # LoadStatus.LOAD_FAILURE
+                else:
+                    assert callback_status in (
+                        LoadStatus.LOAD_FAILURE,
+                        LoadStatus.TIME_MAP_FAILURE,
+                    )
                     await self._send(State.FAILED)
 
             else:
@@ -126,6 +129,7 @@ class Job:
     async def __call__(
         self, start: asyncio.Event, sem: asyncio.BoundedSemaphore, max_submit: int = 2
     ) -> None:
+        self._requested_max_submit = max_submit
         await start.wait()
 
         for attempt in range(max_submit):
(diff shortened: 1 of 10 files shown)


src/ert/scheduler/job.py
```diff
@@ -57,7 +61,9 @@ def __init__(self, scheduler: Scheduler, real: Realization) -> None:
         self.started = asyncio.Event()
         self.returncode: asyncio.Future[int] = asyncio.Future()
         self.aborted = asyncio.Event()
-        self._scheduler = scheduler
+        self._scheduler: Scheduler = scheduler
+        self._callback_status_msg: str = ""
+        self._max_submit: int = 2
```
xjules, line 66: Why is `_max_submit` set as an attribute here? I think this one was suppose to just come as function parameter for `__call__`?

src/ert/scheduler/job.py
```diff
@@ -87,12 +93,22 @@ async def _submit_and_run_once(self, sem: asyncio.BoundedSemaphore) -> None:
...
+                if self._callback_status_msg != "":
+                    self._callback_status_msg = status_msg
+                else:
+                    self._callback_status_msg += (
+                        f"\nstatus from done callback: {status_msg}"
+                    )
+
+                if callback_status == LoadStatus.LOAD_SUCCESSFUL:
+                    await self._send(State.COMPLETED)
+                elif callback_status == LoadStatus.TIME_MAP_FAILURE:
```
xjules, line 107:
```python
else:
   await self._send(State.FAILED)
```

src/ert/scheduler/job.py
```diff
@@ -111,6 +127,7 @@ async def _submit_and_run_once(self, sem: asyncio.BoundedSemaphore) -> None:
     async def __call__(
         self, start: asyncio.Event, sem: asyncio.BoundedSemaphore, max_submit: int = 2
     ) -> None:
+        self._max_submit = max_submit
```
xjules, line 130: Why an attribute?

berland: It is only for the sake of the error message. The class `Job` does not hold the config, so it needs to keep track of `max_submit` from `__call__` in order to know how to make the error message. The attribute name can be anonymized further at least.

src/ert/scheduler/job.py
berland: will fix!

Looking forward to 3.10:
```
                match callback_status:
                    case LoadStatus.LOAD_SUCCESSFUL:
                        await self._send(State.COMPLETED)
                    case LoadStatus.LOAD_FAILURE | LoadStatus.TIME_MAP_FAILURE | _:
                        await self._send(State.FAILED)
```

xjules:
A question on a general note. Since we are using async context we do not need to worry that forward_model_ok is being executed in job.py. Since the main idea (at least what I [thought is](https://github.com/equinor/ert/issues/6789)) is to execute forward_model_ok for each realization in a sequence from scheduler. But I said async context will take care of it?

berland:
I think we still need to worry, we are executing the done callback in the async context and it will be blocking while it is happening. There are some tricks from the branch `python_jobqueue` to apply at least until we solve it using multiprocessing or threads.

xjules:
You need to rebase but otherwise it is good to go :+1:

xjules (approved):
Nice job @berland ! :rocket:

PR #6857, opened 2023-12-28
Support max_running in Scheduler

### LINKED ISSUE #6850: Scheduler does not handle max_running=0 properly
If max_running is set to zero in the ert configuration, this should mean the same as no limit. The current code seemingly does not support max_running=0.

https://github.com/equinor/ert/blob/508fa719c9340705633b92a1abc0b93c4026ccee/src/ert/scheduler/scheduler.py#L122

berland:
**Issue**
Resolves #6850

**Approach**
just_do_it

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/tests/unit_tests/scheduler/test_scheduler.py
+++ b/tests/unit_tests/scheduler/test_scheduler.py
@@ -215,6 +215,8 @@ async def test_max_running(max_running, mock_driver, storage, tmp_path):
         await asyncio.sleep(0.01)
         runs.append(False)
 
+    # Ensemble size must be larger than max_running to be able
+    # to expose issues related to max_running
     ensemble_size = max_running * 3 if max_running > 0 else 10
 
     ensemble = storage.create_experiment().create_ensemble(
@@ -229,7 +231,7 @@ async def test_max_running(max_running, mock_driver, storage, tmp_path):
         mock_driver(wait=wait), realizations, max_running=max_running
     )
 
-    await sch.execute()
+    assert await sch.execute() == EVTYPE_ENSEMBLE_STOPPED
 
     currently_running = 0
     max_running_observed = 0
--- a/tests/unit_tests/scheduler/test_scheduler.py
+++ b/tests/unit_tests/scheduler/test_scheduler.py
@@ -204,6 +204,39 @@ async def test_max_runtime(realization, mock_driver):
     assert timeouteventfound
 
 
+@pytest.mark.parametrize("max_running", [0, 1, 2, 10])
+async def test_max_running(max_running, mock_driver, storage, tmp_path):
+    currently_running = 0
+    max_running_observed = 0
+
+    async def wait():
+        nonlocal currently_running
+        currently_running += 1
+        await asyncio.sleep(0.01)
+        currently_running -= 1
+
+    ensemble_size = max_running * 3 if max_running > 0 else 10
+
+    ensemble = storage.create_experiment().create_ensemble(
+        name="foo", ensemble_size=ensemble_size
+    )
+    realizations = [
+        create_stub_realization(ensemble, tmp_path, iens)
+        for iens in range(ensemble_size)
+    ]
+
+    sch = scheduler.Scheduler(
+        mock_driver(wait=wait), realizations, max_running=max_running
+    )
+    task = asyncio.create_task(sch.execute())
+    effective_max_running = max_running if max_running else ensemble_size
+    while not task.done():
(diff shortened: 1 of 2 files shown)


tests/unit_tests/scheduler/test_scheduler.py
```diff
@@ -204,6 +204,39 @@ async def wait():
...
+@pytest.mark.parametrize("max_running", [0, 1, 2, 10])
+async def test_max_running(max_running, mock_driver, storage, tmp_path):
+    currently_running = 0
+    max_running_observed = 0
+
+    async def wait():
+        nonlocal currently_running
+        currently_running += 1
+        await asyncio.sleep(0.01)
+        currently_running -= 1
```
pinkwah, line 216: We could change this so that instead of incrementing and decrementing `currently_running`, we can use a boolean list where `True` means started and `False` means stopped. We can then find `max_running_observed` by:
```py
async def wait():
    nonlocal runs
    runs.append(True)
    await asyncio.sleep(0.1)
    runs.append(False)

# ...

max_running_observed = 0
currently_running = 0
for run in runs:
    currently_running += 1 if run else -1
    max_running_observed = max(max_running_observed, currently_running)
```

Very advent of code. Also this way we don't rely on timings (other than `wait`'s `asyncio.sleep(0.1)`)

berland: Done!

tests/unit_tests/scheduler/test_scheduler.py
```diff
@@ -204,6 +205,44 @@ async def wait():
...
+    realizations = [
+        create_stub_realization(ensemble, tmp_path, iens)
+        for iens in range(ensemble_size)
+    ]
+
+    sch = scheduler.Scheduler(
+        mock_driver(wait=wait), realizations, max_running=max_running
+    )
+
+    await sch.execute()
```
xjules, line 232: Just to be sure: ` assert await sch.execute() == EVTYPE_ENSEMBLE_STOPPED` ?

tests/unit_tests/scheduler/test_scheduler.py
```diff
@@ -204,6 +205,44 @@ async def wait():
...
 
 
+@pytest.mark.parametrize("max_running", [0, 1, 2, 10])
+async def test_max_running(max_running, mock_driver, storage, tmp_path):
+    runs: List[bool] = []
+
+    async def wait():
+        nonlocal runs
+        runs.append(True)
+        await asyncio.sleep(0.01)
```
jonathan-eq, line 215: Can this be replaces with `await asyncio.sleep(0)`?

tests/unit_tests/scheduler/test_scheduler.py
```diff
@@ -204,6 +205,44 @@ async def wait():
...
+        runs.append(False)
+
+    ensemble_size = max_running * 3 if max_running > 0 else 10
```
jonathan-eq, line 218: Does this mean we should have 3x more ensembles than max_running? Maybe have it as a variable instead of magic number

jonathan-eq (changes requested):
Very clean code. Nice!

tests/unit_tests/scheduler/test_scheduler.py
xjules: I'd say await `asyncio.sleep(0.1)` just to be sure

tests/unit_tests/scheduler/test_scheduler.py
berland: Can do, but it adds assertions that are not related to the purpose of the test. Can it cause unneccesary maintenance later on to have bonus asserts?

tests/unit_tests/scheduler/test_scheduler.py
berland: I put a small number there to let the async event loop have some time to let all realizations being started. But I tried with 0 and the test passed for at least 500 realizations on local iron. I think I will keep it at 0.01 as an insurance against flakiness. Bumping to 0.1 I will do if we see flakiness on this, but I think 0.01 is sufficient when even zero works fine.

tests/unit_tests/scheduler/test_scheduler.py
berland: We want more realizations than max_running, if not we are exposing the scheduler for any problem related to max_running. Can add a comment, but we basically just need `ensemble_size > max_running`

jonathan-eq (approved):
Very good!

PR #6858, opened 2023-12-28
Support submit sleep in scheduler

### LINKED ISSUE #6856: Implement SUBMIT_SLEEP for Scheduler
The legacy Torque and LSF drivers support a driver option called SUBMIT_SLEEP.

This should be reimplemented in scheduler.py (it does not have to be a driver thing).

For the implementation in feature branch, see:
https://github.com/equinor/ert/blob/f49e93f9677c9c6cfde0d1e3fc51c2f01c36c5bc/src/ert/scheduler/scheduler.py#L224
https://github.com/equinor/ert/blob/python_jobqueue/tests/unit_tests/scheduler/test_scheduler.py#L148-L160

In LSF driver:
https://github.com/equinor/ert/blob/211888d7bdfc1e600db1126d9fa3ad186330a269/src/clib/lib/job_queue/lsf_driver.cpp#L696

berland:
**Issue**
Resolves #6856 

**Approach**
Use a singleton object to be responsible for determining whether  `Job` needs to wait more before it proceeds when doing its `__call__`.

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/config/queue_config.py
+++ b/src/ert/config/queue_config.py
@@ -32,6 +32,7 @@ VALID_QUEUE_OPTIONS: Dict[Any, List[str]] = {
 class QueueConfig:
     job_script: str = shutil.which("job_dispatch.py") or "job_dispatch.py"
     max_submit: int = 2
+    submit_sleep: float = 0.0
     queue_system: QueueSystem = QueueSystem.LOCAL
     queue_options: Dict[QueueSystem, List[Tuple[str, str]]] = field(
         default_factory=dict
@@ -48,6 +49,7 @@ class QueueConfig:
         )
         job_script = job_script or "job_dispatch.py"
         max_submit: int = config_dict.get("MAX_SUBMIT", 2)
+        submit_sleep: float = config_dict.get("SUBMIT_SLEEP", 0.0)
         queue_options: Dict[QueueSystem, List[Tuple[str, str]]] = defaultdict(list)
         for queue_system, option_name, *values in config_dict.get("QUEUE_OPTION", []):
             if option_name not in VALID_QUEUE_OPTIONS[queue_system]:
@@ -67,6 +69,12 @@ class QueueConfig:
                     " usually provided by the site-configuration file, beware that"
                     " you are effectively replacing the default value provided."
                 )
+            if (
+                values
+                and option_name == "SUBMIT_SLEEP"
+                and selected_queue_system == queue_system
+            ):
+                submit_sleep = float(values[0])
 
         for queue_system_val in queue_options:
             if queue_options[queue_system_val]:
@@ -85,12 +93,15 @@ class QueueConfig:
                 queue_options[selected_queue_system],
             )
 
-        return QueueConfig(job_script, max_submit, selected_queue_system, queue_options)
+        return QueueConfig(
+            job_script, max_submit, submit_sleep, selected_queue_system, queue_options
+        )
 
     def create_local_copy(self) -> QueueConfig:
         return QueueConfig(
             self.job_script,
             self.max_submit,
+            self.submit_sleep,
             QueueSystem.LOCAL,
             self.queue_options,
         )
--- a/src/ert/scheduler/scheduler.py
+++ b/src/ert/scheduler/scheduler.py
@@ -5,6 +5,7 @@ import json
 import logging
 import os
 import ssl
+import time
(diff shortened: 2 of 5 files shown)


berland:
By adding `"SUBMIT_SLEEP"` to `GENERIC_QUEUE_OPTIONS` in `queue_config.py` this can be tested to work with the local driver on the poly case. Will not add that to the PR as it breaks the legacy JobQueue.

jonathan-eq (changes requested):
Looks good, very readable code!

berland:
There should be some tests for this togheter with MAX_RUNNING. I don't think combining these feature will currently give expected results.

src/ert/config/queue_config.py
```diff
@@ -48,6 +49,7 @@ def from_dict(cls, config_dict: ConfigDict) -> QueueConfig:
         )
         job_script = job_script or "job_dispatch.py"
         max_submit: int = config_dict.get("MAX_SUBMIT", 2)
+        submit_sleep: float = config_dict.get("SUBMIT_SLEEP", 0.0)
```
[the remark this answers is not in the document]
berland: Agree. Will change to `0.0` everywhere.

tests/unit_tests/scheduler/test_scheduler.py
```diff
@@ -342,3 +344,94 @@ async def wait(iens):
 
     assert await sch.execute(min_required_realizations=5) == EVTYPE_ENSEMBLE_STOPPED
     assert killed_iens == [6, 7, 8, 9]
+
+
+@pytest.mark.parametrize(
+    "submit_sleep, iens_stride, realization_runtime, expected_max_running",
+    [(0, 1, 0.1, 10), (0.1, 1, 0.1, 2), (0.1, 2, 0.1, 2)],
+)
+async def test_submit_sleep(
```
berland, line 353: This funcion maybe represents a strange way to test submit_sleep, by asserting on a correct max_running throughout the simulations which is a consequence of different submit_sleeps found by trial and error more or less.  Perhaps drop this and only test via the function below.

src/ert/scheduler/scheduler.py
```diff
@@ -57,6 +76,8 @@ def __init__(
         self.driver = driver
         self._tasks: MutableMapping[int, asyncio.Task[None]] = {}
 
+        self.submit_sleep_state = SubmitSleeper(submit_sleep)
```
xjules, line 79: Shall we just do?:
```python
self.submit_sleep_state = None
if submit_sleep > 0:
    self.submit_sleep_state = SubmitSleeper(submit_sleep)
```
And then in job:
```python
if self._scheduler.submit_sleep_state:
   await self._scheduler.submit_sleep_state.sleep_until_we_can_submit()
```

xjules:
It needs rebasing.

tests/unit_tests/scheduler/test_scheduler.py
berland: resolved.

src/ert/scheduler/scheduler.py
berland: Fixed

src/ert/config/queue_config.py
```diff
@@ -67,6 +69,12 @@ def from_dict(cls, config_dict: ConfigDict) -> QueueConfig:
                     " usually provided by the site-configuration file, beware that"
                     " you are effectively replacing the default value provided."
                 )
+            if (
+                values
```
xjules, line 73: Shall we produce a warning that `setting SUBMIT_SLEEP for a different queue`?

berland: In this PR, you are only allowed to set it for the LSF queue, for which there is no driver for Scheduler. So that should not be necessary. 

Through this PR it submit_sleep is not really usable from an ert config. I think we can accept that until the LSF driver is merged, because then it should work immediately.

When job_queue is purged, we can add SUBMIT_SLEEP to GENERIC_QUEUE_OPTIONS  and it will work for all queues.

berland: (you will get a ConfigValidationError if you currently use SUBMIT_SLEEP with the local queue)

xjules (approved):
Nice job @berland ! :rocket:

jonathan-eq (approved):
LGTM!

PR #6864, opened 2023-12-29
Add `EventSender` to resolve dependency cycle

pinkwah:
`Scheduler` has a reference to `Job` and `Job` has a reference to `Scheduler`. Adding `EventSender` lets us resolve this cycle as now `Scheduler` has a reference to `Job`s and `EventSender`, but each `Job` only refers to `EventSender`.

### DIFF
--- a/src/ert/scheduler/event_sender.py
+++ b/src/ert/scheduler/event_sender.py
@@ -1,0 +1,68 @@
+from __future__ import annotations
+
+import asyncio
+import ssl
+from typing import TYPE_CHECKING, Any, Mapping, Optional
+
+from cloudevents.conversion import to_json
+from cloudevents.http import CloudEvent
+from websockets import Headers, connect
+
+if TYPE_CHECKING:
+    from ert.ensemble_evaluator.identifiers import EvGroupRealizationType
+
+
+class EventSender:
+    def __init__(
+        self,
+        ens_id: Optional[str],
+        ee_uri: Optional[str],
+        ee_cert: Optional[str],
+        ee_token: Optional[str],
+    ) -> None:
+        self.ens_id = ens_id
+        self.ee_uri = ee_uri
+        self.ee_cert = ee_cert
+        self.ee_token = ee_token
+        self.events: asyncio.Queue[CloudEvent] = asyncio.Queue()
+
+    async def send(
+        self,
+        type: EvGroupRealizationType,
+        source: str,
+        attributes: Optional[Mapping[str, Any]] = None,
+        data: Optional[Mapping[str, Any]] = None,
+    ) -> None:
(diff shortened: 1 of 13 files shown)


xjules:
This needs rebasing first

berland:
Can this solve https://github.com/equinor/ert/blob/881a350c2fef60983a1042309d82ad39aeac8178/src/ert/scheduler/job.py#L28-L29 ?

src/ert/scheduler/scheduler.py
```diff
@@ -57,38 +63,32 @@ def __init__(
         self.driver = driver
         self._tasks: MutableMapping[int, asyncio.Task[None]] = {}
 
-        self._jobs: MutableMapping[int, Job] = {
-            real.iens: Job(self, real) for real in (realizations or [])
+        self._jobs: Mapping[int, Job] = {
+            real.iens: Job(real, on_complete=self._on_job_complete)
```
xjules, line 67: Looks good! But (just our of curiosity mostly), why not to to use:
`self._tasks[iens].add_done_callback(self._on_job_complete)`?

pinkwah: Because it's not when the task is done running, it's when the job is `COMPLETED` successfully.

xjules: Yes, but we can check in `_on_job_complete` that `self.state == COMPLETED`?

src/ert/scheduler/event_sender.py
```diff
@@ -0,0 +1,69 @@
...
+            ssl=tls,
+            extra_headers=headers,
+            open_timeout=60,
+            ping_timeout=60,
+            ping_interval=60,
+            close_timeout=60,
+        ):
+            while True:
+                event = await self.events.get()
+                print(f"==SENDING {event=}")
```
berland, line 68: leftover?

src/ert/scheduler/job.py
```diff
@@ -87,13 +95,15 @@ def running_duration(self) -> float:
             return time.time() - self._start_time
         return 0
 
-    async def _submit_and_run_once(self, sem: asyncio.BoundedSemaphore) -> None:
+    async def _submit_and_run_once(
+        self, sem: asyncio.BoundedSemaphore, driver: Driver
+    ) -> None:
```
berland, line 100: is this driver arg related to the commit message, or is it some other refactoring?

pinkwah: It's part of the commit. The `Job` object obtains a `Driver` only when it is supposed to use it, not before.

pinkwah:
Closing as this doesn't really solve anything. There will be a cyclic dependency regardless.

PR #6865, opened 2023-12-29
Add stop_long_running_jobs functionality to Scheduler

### LINKED ISSUE #6710: Implement  stop_long_running_jobs in the Scheduler
Reimplement functionality for killing jobs that take more than 125% of the average runtime.
This relates to the newly introduced `Scheduler` and `LocalDriver` classes.
This should include the test as well.

xjules:
**Issue**
Resolves #6710 

**Approach**
This adds a task which handles long running jobs

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist

### DIFF
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import time
 import uuid
 from enum import Enum
 from pathlib import Path
@@ -67,6 +68,8 @@ class Job:
         self._scheduler: Scheduler = scheduler
         self._callback_status_msg: str = ""
         self._requested_max_submit: Optional[int] = None
+        self._start_time: Optional[float] = None
+        self._end_time: Optional[float] = None
 
     @property
     def iens(self) -> int:
@@ -75,6 +78,14 @@ class Job:
     @property
     def driver(self) -> Driver:
         return self._scheduler.driver
+
+    @property
+    def running_duration(self) -> float:
+        if self._start_time:
+            if self._end_time:
+                return self._end_time - self._start_time
+            return time.time() - self._start_time
+        return 0
 
     async def _submit_and_run_once(self, sem: asyncio.BoundedSemaphore) -> None:
         await sem.acquire()
@@ -88,6 +99,7 @@ class Job:
 
             await self._send(State.PENDING)
             await self.started.wait()
+            self._start_time = time.time()
 
             await self._send(State.RUNNING)
             if self.real.max_runtime is not None and self.real.max_runtime > 0:
@@ -178,6 +190,10 @@ class Job:
         self.state = state
         if state in (State.FAILED, State.ABORTED):
             await self._handle_failure()
+
+        if state == State.COMPLETED:
+            self._end_time = time.time()
+            await self._scheduler.completed_jobs.put(self.iens)
 
         status = STATE_TO_LEGACY[state]
         event = CloudEvent(
--- a/src/ert/scheduler/scheduler.py
+++ b/src/ert/scheduler/scheduler.py
@@ -8,14 +8,7 @@ import ssl
 from collections import defaultdict
 from dataclasses import asdict
 from pathlib import Path
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Dict,
-    MutableMapping,
-    Optional,
-    Sequence,
-)
+from typing import TYPE_CHECKING, Any, Dict, MutableMapping, Optional, Sequence
 
 from pydantic.dataclasses import dataclass
 from websockets import Headers
@@ -69,6 +62,10 @@ class Scheduler:
         }
 
         self._events: asyncio.Queue[Any] = asyncio.Queue()
+        self._average_job_runtime: float = 0
+        self._completed_jobs_num: int = 0
+        self.completed_jobs: asyncio.Queue[int] = asyncio.Queue()
+
         self._cancelled = False
         self._max_submit = max_submit
         self._max_running = max_running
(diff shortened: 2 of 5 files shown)


src/ert/scheduler/scheduler.py
```diff
@@ -83,8 +76,31 @@ def kill_all_jobs(self) -> None:
         for task in self._tasks.values():
             task.cancel()
 
-    def stop_long_running_jobs(self, minimum_required_realizations: int) -> None:
-        pass
+    async def _stop_long_running_jobs(self, minimum_required_realizations: int) -> None:
+        LONG_RUNNING_FACTOR = 1.25
```
pinkwah, line 80: Style: `UPPER_SNAKE_CASE` variable defined inside a function. Should be moved to class or module namespace.

Also, what does this control and why 1.25? Add comment?

src/ert/scheduler/scheduler.py
```diff
@@ -83,8 +76,31 @@ def kill_all_jobs(self) -> None:
...
+                ) / len(completed_jobs)
+                for job_id, task in self._tasks.items():
+                    if (
+                        self._jobs[job_id].running_duration
+                        > LONG_RUNNING_FACTOR * average_runtime
+                        and not task.done()
+                    ):
+                        task.cancel()
+                        await task
+            await asyncio.sleep(0.1)
```
pinkwah, line 103: Perhaps this whole mechanism should be handled by `_process_event_queue()`? This task loops through all realisations multiple times, every 0.1 seconds.

The only time something can happen is when a job starts or finishes (either fail or success). Otherwise, if no job has finished since last loop, all jobs' `running_duration` have increased by `0.1`, but `average_runtime` stays the same. So once `len(completed_jobs) >= minimum_required_realizations` is `True`, ERT will keep on running until the jobs get 90% done and _then_ it kills them, and keep checking every 0.1s even though the time when we should start killing jobs is entirely predictable.

Instead, `_process_event_queue()` could have this logic that is only executed when a job starts or stops. This would mean no more checking every 100ms, which is nice imo because performance.

tests/unit_tests/scheduler/conftest.py
```diff
@@ -10,19 +10,26 @@ def __init__(self, init=None, wait=None, kill=None):
...
-            result = await self._mock_wait()
+            if self._mock_wait.__code__.co_argcount > 0:
+                result = await self._mock_wait(iens)
+            else:
+                result = await self._mock_wait()
             return True if result is None else bool(result)
         return True
 
-    async def _kill(self, *args):
+    async def _kill(self, iens, *args):
```
pinkwah, line 27: `_kill` takes no more arguments

```suggestion
    async def _kill(self, iens):
```

pinkwah (commented):
Looks pretty good, but functionality is a bit weird for me still. Also, wrt. performance, we should try to avoid sleep loops if we can help it.

src/ert/scheduler/scheduler.py
xjules: Right, I gave it up for discussion since in "jobqueue" it is this weird constant (1.25), which is essentially never changed. Neverthelles I would either have as 1) just a number or 2) if this is to change at some point it should be a function parameter (defaulted to 1.25).

src/ert/scheduler/job.py
```diff
@@ -76,6 +79,14 @@ def iens(self) -> int:
     def driver(self) -> Driver:
         return self._scheduler.driver
 
+    @property
+    def running_duration(self) -> float:
+        if self._start_time:
+            if self._end_time:
+                return self._end_time - self._start_time
+            return time.time() - self._start_time
+        return -1
```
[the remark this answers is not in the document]
xjules: I can do `0` yes. Initially I thought it might be beneficial to have `-1` as saying I have not started yet - but `0` would server the same purpose :)

src/ert/scheduler/scheduler.py
xjules: I agree and makes sense, which brings to question why it is not so in old jobqueue - from where I got inspired? :)

src/ert/scheduler/scheduler.py
xjules: FYI, I opted for nr 2.

src/ert/scheduler/job.py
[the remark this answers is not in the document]
xjules: Weirdly enough this does not work: `return (self._end_time or time.time()) - (self._start_time or time.time())`

pinkwah (approved):
I still don't like how we are checking every 100ms when I think we can do better, but I think having basic functionality trumps this sort of performance consideration at this time.

src/ert/scheduler/job.py
```diff
@@ -67,6 +68,8 @@ def __init__(self, scheduler: Scheduler, real: Realization) -> None:
         self._scheduler: Scheduler = scheduler
         self._callback_status_msg: str = ""
         self._requested_max_submit: Optional[int] = None
+        self._start_time: Optional[float] = None
```
berland, line 71: why not datetime.datetime?

berland: maybe faster

xjules: It all boils down to floats, but tbh tried datetime first but got a mypy error at some point.

src/ert/scheduler/scheduler.py
```diff
@@ -83,8 +80,29 @@ def kill_all_jobs(self) -> None:
         for task in self._tasks.values():
             task.cancel()
 
-    def stop_long_running_jobs(self, minimum_required_realizations: int) -> None:
-        pass
+    async def _update_avg_job_runtime(self) -> None:
+        while True:
+            job_id = await self.completed_jobs.get()
```
berland, line 85: use `iens` instead of `job_id`

tests/unit_tests/scheduler/test_scheduler.py
```diff
@@ -270,3 +270,34 @@ async def init(iens, *args, **kwargs):
...
+    killed_iens = []
+
+    async def kill(iens):
+        nonlocal killed_iens
+        killed_iens.append(iens)
+
+    async def wait(iens):
+        # all jobs with iens > 5 will sleep for 10 seconds and should be killed
+        if iens < 6:
+            await asyncio.sleep(0.5)
```
berland, line 286: will 0.1 also work and give faster test execution?

xjules: Will try.

PR #6884, opened 2024-01-03
Clarify fixture param for try_queue_and_scheduler

jonathan-eq:
Now, when a test using the try_queue_and_scheduler-fixture will have the ids "using_job_queue" or "using_scheduler" instead of True or False. This will make it easier to see whether the test is failing with scheduler or job queue.


jonathan-eq:
Now, failing tests will look like this.
image
Way more descriptive than just True/False.

berland (approved):
nice!

PR #6886, opened 2024-01-03
Remove the feature toggle in favor of env vars

eivindjahren:
**Issue**
Resolves #my_issue

**Approach**

## Pre review checklist

- [ ] Updated documentation

## Pre merge checklist


pinkwah:
I'm gonna give some pushback:
1. I would rather have it be an option to use command-line arguments that you have to type every time instead of an environment variable that the user may put in their `.cshrc` and forget. We can use envvars ourselves but tell users to use args.
2. We need to ensure that this functionality is easily changable from tests, as we currently have a fixture that parametrises over which scheduler should be used. If it's a global variable imported in every module that uses it, we would keep having to monkeypatch every module. The `FeatureToggling` singleton allows for easy patching during tests.

src/ert/feature_toggling.py
```diff
@@ -0,0 +1,3 @@
+import os
+
+NEW_SCHEDULER_ENABLED = "ERT_FEATURE_SCHEDULER" in os.environ
```
pinkwah, line 3: This doesn't allow for explicit disabling via `ERT_FEATURE_SCHEDULER=0 ert ..`

Besides, it's not the "new" scheduler, it's the scheduler. The alternative is the job queue.

eivindjahren: Changed to ERT_FEATURE_SCHEDULER. You can now turn it off by doing `ERT_FEATURE_SCHEDULER=OFF ert`.

eivindjahren: Added some validation and a warning if the feature is turned on.

eivindjahren:
> I'm gonna give some pushback:
> 
> 1. I would rather have it be an option to use command-line arguments that you have to type every time instead of an environment variable that the user may put in their `.cshrc` and forget. We can use envvars ourselves but tell users to use args.
> 2. We need to ensure that this functionality is easily changable from tests, as we currently have a fixture that parametrises over which scheduler should be used. If it's a global variable imported in every module that uses it, we would keep having to monkeypatch every module. The `FeatureToggling` singleton allows for easy patching during tests.

It is actually really easy to set this in tests with the `env` keyword in github actions. The current FeatureToggle class still sets a global variable in the end and you have to monkeypatch to change this dynamically.

I have never seen an actual user use a feature toggle flag without being explicitly told to do so.

src/ert/feature_toggling.py
```diff
@@ -0,0 +1,13 @@
+import os
+import warnings
+
+if (_sched_str := os.environ.get("ERT_FEATURE_SCHEDULER")) not in ["ON", "OFF", None]:
```
pinkwah, line 4: `1` and `0` are what is commonly used. A reasonable solution is to have `""` mean unset, `"0"` mean false and any other non-zero non-empty value be true. However, it would be reasonable to add other false strings like `"false"` or `"off"`.

Let's not create code that deviates from how the rest of the universe does it. `1` is on.

pinkwah:
When you import a symbol, you are creating a new reference to it in your new module. Monkeypatching replaces the reference in that one location.

```py
# a.py 
FOO = 0

# b.py
from a import FOO

# c.py
import a, b

a.FOO = 1
print(f"{a.FOO=}")  #=> 1
print(f"{b.FOO=}")  #=> 0

# d.py
import pytest
import a, b

monkeypatch = pytest.MonkeyPatch()
monkeypatch.setattr(a, "FOO", 1)
print(f"{a.FOO=}")  #=> 1
print(f"{b.FOO=}")  #=> 0
```

Hence, if you want to be able to monkeypatch things easily, you need to proxy it via a function.

pinkwah (commented):
But again, I am reiterating that I don't like this solution. I'd like to be able to use arguments as well as environment variables.

We've spent enough time on feature toggling and I much prefer Jon's PR to this one so I would like to get his work in. The next time we can look at feature toggling is when we delete it completely after scheduler runs in production successfully in a few months' time.

PR #6887, opened 2024-01-03
Add TorqueDriver

### LINKED ISSUE #6780: Create PBS (TORQUE) driver for new Scheduler
MVP for something that uses `qsub`, `qdel`, `qstat` called `PBSDriver`/`TorqueDriver` that is capable of running poly_example on the cluster.

pinkwah:
Resolves #6780

### DIFF
--- a/src/ert/scheduler/event.py
+++ b/src/ert/scheduler/event.py
@@ -1,0 +1,19 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Union
+
+
+@dataclass
+class StartedEvent:
+    iens: int
+
+
+@dataclass
+class FinishedEvent:
+    iens: int
+    returncode: int
+    aborted: bool = False
+
+
+Event = Union[StartedEvent, FinishedEvent]
--- a/src/ert/scheduler/local_driver.py
+++ b/src/ert/scheduler/local_driver.py
@@ -5,7 +5,8 @@ import os
 from asyncio.subprocess import Process
 from typing import MutableMapping
 
-from ert.scheduler.driver import Driver, JobEvent
+from ert.scheduler.driver import Driver
+from ert.scheduler.event import FinishedEvent, StartedEvent
 
 _TERMINATE_TIMEOUT = 10.0
 
@@ -15,7 +16,9 @@ class LocalDriver(Driver):
         super().__init__()
         self._tasks: MutableMapping[int, asyncio.Task[None]] = {}
 
-    async def submit(self, iens: int, executable: str, /, *args: str, cwd: str) -> None:
+    async def submit(
+        self, iens: int, executable: str, /, *args: str, cwd: str, name: str = "dummy"
+    ) -> None:
         await self.kill(iens)
         self._tasks[iens] = asyncio.create_task(
             self._run(iens, executable, *args, cwd=cwd)
@@ -40,20 +43,21 @@ class LocalDriver(Driver):
                 *args,
                 cwd=cwd,
             )
-        except Exception as exc:
-            print(f"{exc=}")
-            await self.event_queue.put((iens, JobEvent.FAILED))
+        except FileNotFoundError:
+            # /bin/sh uses returncode 127 for FileNotFound, so copy that
+            # behaviour.
+            await self.event_queue.put(FinishedEvent(iens=iens, returncode=127))
             return
 
-        await self.event_queue.put((iens, JobEvent.STARTED))
+        await self.event_queue.put(StartedEvent(iens=iens))
         try:
-            if await self._wait(proc):
-                await self.event_queue.put((iens, JobEvent.COMPLETED))
-            else:
-                await self.event_queue.put((iens, JobEvent.FAILED))
+            returncode = await self._wait(proc)
+            await self.event_queue.put(FinishedEvent(iens=iens, returncode=returncode))
         except asyncio.CancelledError:
-            await self._kill(proc)
-            await self.event_queue.put((iens, JobEvent.ABORTED))
+            returncode = await self._kill(proc)
+            await self.event_queue.put(
+                FinishedEvent(iens=iens, returncode=returncode, aborted=True)
+            )
 
     async def _init(
         self, iens: int, executable: str, /, *args: str, cwd: str
@@ -66,18 +70,18 @@ class LocalDriver(Driver):
             preexec_fn=os.setpgrp,
         )
 
-    async def _wait(self, proc: Process) -> bool:
+    async def _wait(self, proc: Process) -> int:
         """This method exists to allow for mocking it in tests"""
-        return await proc.wait() == 0
+        return await proc.wait()
 
-    async def _kill(self, proc: Process) -> None:
+    async def _kill(self, proc: Process) -> int:
         """This method exists to allow for mocking it in tests"""
         try:
             proc.terminate()
             await asyncio.wait_for(proc.wait(), _TERMINATE_TIMEOUT)
         except asyncio.TimeoutError:
             proc.kill()
-            await asyncio.wait_for(proc.wait(), _TERMINATE_TIMEOUT)
+        return await proc.wait()
 
     async def poll(self) -> None:
         """LocalDriver does not poll"""
--- a/src/ert/scheduler/torque_driver.py
+++ b/src/ert/scheduler/torque_driver.py
@@ -1,0 +1,127 @@
+from __future__ import annotations
+
+import asyncio
+import shlex
+from asyncio.subprocess import PIPE
+from typing import Literal, Mapping, MutableMapping, Optional, Tuple, Union
+
+from pydantic import BaseModel, Field
+from typing_extensions import Annotated
+
+from ert.scheduler.driver import Driver
+from ert.scheduler.event import Event, FinishedEvent, StartedEvent
+
+_POLL_PERIOD = 2.0  # seconds
+JobState = Literal["C", "E", "H", "Q", "R", "T", "W", "F"]
+
+
+class FinishedJob(BaseModel):
+    job_state: Literal["F"]
+    returncode: Annotated[int, Field(alias="Exit_status")]
+
+
+class QueuedJob(BaseModel):
+    job_state: Literal["H", "Q"]
+
+
+class RunningJob(BaseModel):
+    job_state: Literal["R"]
+
+
+AnyJob = Annotated[
+    Union[FinishedJob, QueuedJob, RunningJob], Field(discriminator="job_state")
+]
+
+
+class _Stat(BaseModel):
+    jobs: Annotated[Mapping[str, AnyJob], Field(alias="Jobs")]
+
+
+class TorqueDriver(Driver):
+    def __init__(self, *, queue_name: Optional[str] = None) -> None:
+        super().__init__()
+
+        self._queue_name = queue_name
+        self._jobs: MutableMapping[str, Tuple[int, JobState]] = {}
+        self._iens2jobid: MutableMapping[int, str] = {}
+
+    async def submit(
+        self, iens: int, executable: str, /, *args: str, cwd: str, name: str = "dummy"
+    ) -> None:
+        script = (
+            "#!/usr/bin/env bash\n"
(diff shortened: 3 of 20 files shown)


tests/integration_tests/scheduler/mock_torque_bin/runner
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+job=$1
+
+function handle_sigterm {
+    # Torque uses (256 + SIGNAL) as the returncode
+    echo "271" > "${job}.returncode"
```
berland, line 6: write to a torque-specific filename, so that the lsf and slurm PR's can utilize this same runner and just let it write to more returncode files as needed.

tests/integration_tests/scheduler/mock_torque_bin/qsub
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
berland, line 1: Suggest to have these mocked queue binaries in a common directory,  `mock_queue_bin` or something

tests/integration_tests/scheduler/mock_torque_bin/runner
```diff
@@ -0,0 +1,14 @@
...
+    exit 0
+}
+
+trap handle_sigterm SIGTERM
+
+echo "$$" > "${job}.pid"
+sh "${job}.script" > "${job}.stdout" 2> "${job}.stderr"
```
berland, line 13: The process started here is not killed when this script is killed. I have an extra `kill $child_pid` in `handle_sigterm` and 
```
sh "${job}.script" > "${job}.stdout" 2> "${job}.stderr" &
child_pid=$!

wait
```

src/ert/scheduler/local_driver.py
```diff
@@ -42,18 +45,17 @@ async def _run(self, iens: int, executable: str, /, *args: str, cwd: str) -> Non
...
-            if await self._wait(proc):
-                await self.event_queue.put((iens, JobEvent.COMPLETED))
-            else:
-                await self.event_queue.put((iens, JobEvent.FAILED))
+            returncode = await self._wait(proc)
+            await self.event_queue.put(FinishedEvent(iens=iens, returncode=returncode))
         except asyncio.CancelledError:
             await self._kill(proc)
-            await self.event_queue.put((iens, JobEvent.ABORTED))
+            returncode = (proc and proc.returncode) or 0
```
xjules, line 57: If something went wrong and we cancel why to set returncode/signal to `0`? Not sure If I understand the semantics in the signal value.

src/ert/scheduler/event.py
```diff
@@ -0,0 +1,24 @@
...
+
+@dataclass
+class StartedEvent:
+    iens: int
+
+
+@dataclass
+class FinishedEvent:
+    iens: int
+    returncode: int
```
xjules, line 15: Could you maybe add a comment, which will describe some fixed `returncode` and `signal` values and what do they mean?

tests/integration_tests/scheduler/mock_torque_bin/runner
pinkwah: I didn't read this and ended up doing:
```sh
sh "${job}.script" ... &
wait $!
echo $? > "${job}.returncode"
```

src/ert/scheduler/torque_driver.py
```diff
@@ -0,0 +1,128 @@
...
+            f"exec -a {shlex.quote(executable)} {shlex.quote(executable)} {shlex.join(args)}\n"
+        )
+
+        arg_queue_name = ["-q", self._queue_name] if self._queue_name else []
+
+        process = await asyncio.create_subprocess_exec(
+            "qsub",
+            "-koe",  # Discard stdout/stderr of job
+            "-rn",  # Don't restart on failure
+            f"-N{name}",  # Set name of job
```
xjules, line 64: Do we need a name, can the name just be "job_iens" or something like that?

src/ert/scheduler/torque_driver.py
```diff
@@ -0,0 +1,128 @@
...
+            job_id = self._iens2jobid[iens]
+
+            proc = await asyncio.create_subprocess_exec("qdel", job_id)
+            await proc.wait()
+        except KeyError:
+            return
+
+    async def poll(self) -> None:
+        while True:
+            if not self._jobs.keys():
```
xjules, line 87: maybe sleep before if and you don't have to have it 2x

src/ert/scheduler/torque_driver.py
```diff
@@ -0,0 +1,128 @@
...
+            *arg_queue_name,
+            "-",
+            stdin=PIPE,
+            stdout=PIPE,
+            stderr=DEVNULL,
+        )
+        job_id, _ = await process.communicate(script.encode())
+        job_id_ = job_id.decode("utf-8").strip()
+        self._jobs[job_id_] = (iens, "Q")
+        self._iens2jobid[iens] = job_id_
```
xjules, line 74: Correct me if I am wrong (which probably I am), but don't we want to create polling task at the end of submit?

src/ert/scheduler/torque_driver.py
```diff
@@ -0,0 +1,128 @@
...
+
+    async def kill(self, iens: int) -> None:
+        try:
+            job_id = self._iens2jobid[iens]
+
+            proc = await asyncio.create_subprocess_exec("qdel", job_id)
+            await proc.wait()
```
xjules, line 81: `del self._iens2jobi[iens]`?

src/ert/scheduler/torque_driver.py
```diff
@@ -0,0 +1,128 @@
...
+                continue
+
+            proc = await asyncio.create_subprocess_exec(
+                "qstat",
+                "-fxFjson",
+                *self._jobs.keys(),
+                stdout=asyncio.subprocess.PIPE,
+            )
+            stdout, _ = await proc.communicate()
+            stat = _Stat.model_validate_json(stdout)
```
xjules, line 98: Shall we?
try:
```python
try:
     stat = _Stat.model_validate_json(stdout)
except ValueError: ...
```

src/ert/scheduler/local_driver.py
pinkwah: Behaviour changed in new code. `-1` is now the default.

src/ert/scheduler/event.py
pinkwah: I have no clue how to do this...

```py
    # The return code of the job
    returncode: int
```
seems like really garbage documentation

src/ert/scheduler/torque_driver.py
berland: I think a name should be required, as it is what users will see when they run `qstat` manually.

pinkwah: Yes. This is given by the ERT config `JOB_NAME` and is apparently the main reason why we aren't using dask (dask does not support setting job names)

src/ert/scheduler/torque_driver.py
```diff
@@ -0,0 +1,129 @@
...
+        process = await asyncio.create_subprocess_exec(
+            "qsub",
+            "-koe",  # Discard stdout/stderr of job
+            "-rn",  # Don't restart on failure
+            f"-N{name}",  # Set name of job
+            *arg_queue_name,
+            "-",
+            stdin=PIPE,
+            stdout=PIPE,
+            stderr=DEVNULL,
```
berland, line 69: (DEVNULL here is temporary I suppose, until we have more features to deal with failures)

src/ert/scheduler/torque_driver.py
pinkwah: I am not sure how to trigger `poll` iff. `self._jobs` is non-empty. `asyncio.Event` and such are all single-shot.

src/ert/scheduler/torque_driver.py
pinkwah: No. Then `poll()` will never discover that the job has been successfully killed and ERT will hang.

src/ert/scheduler/torque_driver.py
pinkwah: No. If `qstat` returns something we don't understand then this whole driver is broken. We should not ignore errors. (Also it's `pydantic.ValidationError`, not `ValueError`)

src/ert/scheduler/torque_driver.py
pinkwah: yeah I should not do `DEVNULL` here.

tests/integration_tests/scheduler/mock_torque_bin/qsub
pinkwah: Changed to just `bin`

tests/integration_tests/scheduler/mock_torque_bin/runner
pinkwah: It's better for the people who will implement LSF or SLURM to modify this script rather than me guessing what they want in the future. Maybe add an environment variable which determines what sigterm should write to returncode, and this can be set by `qsub`, `bsub` etc.

src/ert/scheduler/event.py
xjules: I don't get this then?
```
FinishedEvent(iens=iens, returncode=254)
```

src/ert/scheduler/torque_driver.py
xjules: Yes, makes sense. Btw, documentation says `ValueError` https://docs.pydantic.dev/latest/api/base_model/#pydantic.main.BaseModel.model_validate_json

pinkwah: You're right. Also, `pydantic.ValidationError` subclasses `ValueError`. I thought they were entirely distinct types, but it makes sense.

https://github.com/pydantic/pydantic-core/blob/758bc51d4bb7e94aae0c8e6f4ce515b2d55cc3a6/python/pydantic_core/_pydantic_core.pyi#L723

src/ert/scheduler/event.py
pinkwah: Forward model of `iens` finished with returncode 254.

xjules: That's the part I understand :) But why 254, since if it means that `iff returncode==254` means failed iens, it should be documented :top:

tests/integration_tests/scheduler/test_torque_driver.py
```diff
@@ -0,0 +1,93 @@
...
+        assert iens == 0
+        assert returncode == 42
+        assert aborted is False
+
+        nonlocal finished_called
+        finished_called = True
+
+    await driver.submit(0, "sh", "-c", "exit 42", cwd=str(tmp_path))
+    await poll(driver, {0}, finished=finished)
+    assert finished_called
```
xjules, line 72: Nice!

src/ert/scheduler/local_driver.py
```diff
@@ -66,18 +68,19 @@ async def _init(
...
-    async def _kill(self, proc: Process) -> None:
+    async def _kill(self, proc: Process) -> int:
         """This method exists to allow for mocking it in tests"""
         try:
             proc.terminate()
             await asyncio.wait_for(proc.wait(), _TERMINATE_TIMEOUT)
         except asyncio.TimeoutError:
             proc.kill()
             await asyncio.wait_for(proc.wait(), _TERMINATE_TIMEOUT)
+        return proc.returncode or -1
```
berland, line 83: so if `returncode == 0` we should return -1 ? (or did you mean `is None`?)

pinkwah: We should return non-zero is my thinking. Otherwise we can end up with `FinishedEvent(iens=iens, returncode=0, aborted=True)`, or:

Task failed successfully

src/ert/scheduler/event.py
pinkwah: Changed to 127 and added a comment as to why (because `sh` uses it for file not found error)

xjules (approved):
Very nice work @pinkwah ! Let's bring this in :rocket:

Issue #6885, opened 2024-01-03
Fix failing unit_tests/ensemble_evaluator/test_ensemble_legacy.py::test_run_and_cancel_legacy_ensemble


xjules:
**Describe the bug**
The following test fails, but only when running it as github workflow:
```python
@pytest.mark.timeout(60)
@pytest.mark.scheduler
    def test_run_and_cancel_legacy_ensemble(
        tmpdir, make_ensemble_builder, monkeypatch, try_queue_and_scheduler
    ):
        num_reals = 2
        custom_port_range = range(1024, 65535)
        with tmpdir.as_cwd():
            ensemble = make_ensemble_builder(
                monkeypatch, tmpdir, num_reals, 2, job_sleep=40
            ).build()
            config = EvaluatorServerConfig(
                custom_port_range=custom_port_range,
                custom_host="127.0.0.1",
                use_token=False,
                generate_cert=False,
            )
    
            evaluator = EnsembleEvaluator(ensemble, config, 0)
    
            evaluator._start_running()
            with Monitor(config) as mon:
                cancel = True
                with contextlib.suppress(
                    ConnectionClosed
                ):  # monitor throws some variant of CC if dispatcher dies
>                   for _ in mon.track():

unit_tests/ensemble_evaluator/test_ensemble_legacy.py:78: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/prog/res/komodo/bleeding-py38-rhel7/root/lib64/python3.8/site-packages/ert/ensemble_evaluator/monitor.py:95: in track
    for message in duplexer.receive():
/prog/res/komodo/bleeding-py38-rhel7/root/lib64/python3.8/site-packages/ert/ensemble_evaluator/sync_ws_duplexer.py:126: in receive
    event = asyncio.run_coroutine_threadsafe(
/opt/rh/rh-python38/root/usr/lib64/python3.8/concurrent/futures/_base.py:439: in result
    self._condition.wait(timeout)
self = <Condition(<unlocked _thread.RLock object owner=0 count=0 at 0x7f1aad32b450>, 0)>
timeout = None

    def wait(self, timeout=None):
        """Wait until notified or until a timeout occurs.
    
        If the calling thread has not acquired the lock when this method is
        called, a RuntimeError is raised.
    
        This method releases the underlying lock, and then blocks until it is
[...]

PR #6908, opened 2024-01-05
Correctly cancel tasks from sync context in Scheduler

xjules:
**Issue**
Improves flakiness of #6885 
Use `asyncio.run_coroutine_threadsafe` when cancelling ensemble in Scheduler.

**Approach**

## When applicable

### DIFF
--- a/src/ert/scheduler/scheduler.py
+++ b/src/ert/scheduler/scheduler.py
@@ -62,6 +62,8 @@ class Scheduler:
         }
 
         self._events: asyncio.Queue[Any] = asyncio.Queue()
+        self._loop: Optional[asyncio.AbstractEventLoop] = None
+
         self._average_job_runtime: float = 0
         self._completed_jobs_num: int = 0
         self.completed_jobs: asyncio.Queue[int] = asyncio.Queue()
@@ -76,6 +78,10 @@ class Scheduler:
         self._ee_token = ee_token
 
     def kill_all_jobs(self) -> None:
+        assert self._loop
+        asyncio.run_coroutine_threadsafe(self.cancel_all_jobs(), self._loop)
+
+    async def cancel_all_jobs(self) -> None:
         self._cancelled = True
         for task in self._tasks.values():
             task.cancel()
@@ -148,6 +154,9 @@ class Scheduler:
         self,
         min_required_realizations: int = 0,
     ) -> str:
+        # We need to store the loop due to when calling
+        # cancel jobs from another thread
+        self._loop = asyncio.get_running_loop()
         async with background_tasks() as cancel_when_execute_is_done:
             cancel_when_execute_is_done(self._publisher())
             cancel_when_execute_is_done(self._process_event_queue())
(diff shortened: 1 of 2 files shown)


berland:
Which issue is solved by this?

jonathan-eq:
> Which issue is solved by this?

We have a hypothesis that this is related to #6885

jonathan-eq (approved):
Looks good!

berland:
I am skeptical that this commit can fix anything that is flaky.

xjules:
> I am skeptical that this commit can fix anything that is flaky.

While debugging yesterday with @jonathan-eq we found out that cancelling / killing the jobs from sync context is defintelly not straighforward. For insance, if the running loop has finsihed (eg. started with `asyncio.run` in simulation_context.py) it will crash. Also there is question whether `scheduler.kill_all_jobs` should be blocking or not? Anyway, I'll put this back to todo as this needs to be discussed.

berland:
Can you rebase into a sensible number of commits, and write as much information as we have into the commit message. I foresee this might come up in future bisections :)

berland:
I am certain that this commit makes #6885  *less flaky* but it does not solve it entirely. Maybe there is another spot where we need to do the same trick.

berland (approved):

src/ert/scheduler/scheduler.py
```diff
@@ -76,6 +78,12 @@ def __init__(
         self._ee_token = ee_token
 
     def kill_all_jobs(self) -> None:
+        assert self._loop.is_running()
+        asyncio.run_coroutine_threadsafe(
+            self.signal_kill_all_jobs(), self._loop
+        ).result()
+
+    async def signal_kill_all_jobs(self) -> None:
```
berland, line 86: Rename to `async def cancel_all_jobs(self)`?

(that will avoid introducing the concept of "signal" in this context).

src/ert/scheduler/scheduler.py
```diff
@@ -76,6 +78,10 @@ def __init__(
         self._ee_token = ee_token
 
     def kill_all_jobs(self) -> None:
+        assert self._loop and self._loop.is_running()
```
berland, line 81: why do we assert it is running? I guess it will only change which exception we get, it will crash nonetheless?

xjules: Yes, it needs to be running otherwise we get exception - regardless. But for the sake of understanding I would keep it there.

berland: I suppose `assert self._loop` is there to satisfy `mypy` and for no other reason. 

Then I question if an assertionerror is more readable than an exception that says that the loop is not running. If my memory serves me right, the exception when that happens is readable.

xjules: > I suppose `assert self._loop` is there to satisfy `mypy` and for no other reason.

yes, correct, but also not to cancel before we are running, `scheduler.execute()`.

> Then I question if an assertionerror is more readable than an exception that says that the loop is not running. If my memory serves me right, the exception when that happens is readable.

Can remove it then :+1: 
wrt. readability I meant that somebody reading the code would understand the cancellation (from another thread) can happen only if the loop is running.

berland:
Go for it!

PR #6909, opened 2024-01-07
Ignore CancelledError when killing

### LINKED ISSUE #6885: Fix failing unit_tests/ensemble_evaluator/test_ensemble_legacy.py::test_run_and_cancel_legacy_ensemble

berland:
As this kill() is called from job.py when a CancelledError is handled, some more exception handling is required here. Solves a flaky test

**Issue**
Resolves #6885 

**Approach**

## When applicable

### DIFF
--- a/src/ert/scheduler/local_driver.py
+++ b/src/ert/scheduler/local_driver.py
@@ -26,7 +26,7 @@ class LocalDriver(Driver):
             self._tasks[iens].cancel()
             await self._tasks[iens]
             del self._tasks[iens]
-        except KeyError:
+        except (KeyError, asyncio.CancelledError):
             return
 
     async def finish(self) -> None:


src/ert/scheduler/local_driver.py
```diff
@@ -26,7 +26,7 @@ async def kill(self, iens: int) -> None:
             self._tasks[iens].cancel()
             await self._tasks[iens]
             del self._tasks[iens]
-        except KeyError:
+        except (KeyError, asyncio.CancelledError):
```
xjules, line 29: We do catch cancellation in `driver._run`, so not sure how it helps here?

berland: This helps when we are called from https://github.com/equinor/ert/blob/9b92e9ad9407fd2b81d91e2e3a1570f64fa4faeb/src/ert/scheduler/job.py#L136

xjules (approved):
Let's put it in! Good job @berland !

PR #6920, opened 2024-01-10
Add explicit logging when Scheduler was cancelled

xjules:
**Issue**
We are logging the same info regardless the Scheduler failed (eg. max_submit reached) or it was cancelled. Therefore we should differentiate between these two states.

**Approach**
Add handler for aborted state.

## When applicable

### DIFF
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -187,12 +187,11 @@ class Job:
         log_info_from_exit_file(Path(self.real.run_arg.runpath) / ERROR_file)
 
     async def _handle_aborted(self) -> None:
-        info_msg = "Scheduler cancelled, stopping job"
-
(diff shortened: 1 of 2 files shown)


src/ert/scheduler/job.py
```diff
@@ -186,12 +186,24 @@ async def _handle_failure(self) -> None:
         logger.error(error_msg)
         log_info_from_exit_file(Path(self.real.run_arg.runpath) / ERROR_file)
 
+    async def _handle_aborted(self) -> None:
+        info_msg = "Scheduler cancelled, stopping job"
```
berland, line 190: Is this a scenario that did not exist in the job_queue code? (since we are introducing a brand new type of error message that the users can see)

xjules: Updated accordingly.

berland (approved):

PR #6924, opened 2024-01-11
Add new workflow for dark storage performance benchmark

### LINKED ISSUE #6945: Make a report dashboard from dark storage benchmarking report
Currently, we have dark storage benchmarking results being published in the gh-pages branch. We should create a workflow that creates a report dashboard based on the output in this branch. The workflow should run on updated to a directory in this branch.

jonathan-eq:
Resolves #6945
Adds workflow for dark storage performance test, and uses a marketplace action for adding the result to a convenient dashboard in the gh-pages branch.

It is triggered when the corresponding workflow running on-prem publishes the result in the gh-branch.

### DIFF
--- a/.github/workflows/dark_storage_benchmark.yml
+++ b/.github/workflows/dark_storage_benchmark.yml
@@ -5,7 +5,8 @@ on:
         - gh-pages
       paths:
         - /dev/dark_storage_performance_benchmark/**
-
+    workflow_dispatch:
+      
 permissions:
     # deployments permission to deploy GitHub pages website
     deployments: write
--- a/.github/workflows/dark_storage_benchmark.yml
+++ b/.github/workflows/dark_storage_benchmark.yml
@@ -4,9 +4,8 @@ on:
       branches:
         - gh-pages
       paths:
-        - /dev/dark_storage_performance_benchmark/**
-    workflow_dispatch:
-      
+        - /dev/dark_storage_performance_benchmark/output.json
+
 permissions:
     # deployments permission to deploy GitHub pages website
     deployments: write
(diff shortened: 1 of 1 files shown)


berland:
Can this be merged with the existing `benchmark.yml` file (it does not seem to scale well if we need a new file for every item we want to benchmark, and that could potentially become many)

jonathan-eq:
> Can this be merged with the existing `benchmark.yml` file (it does not seem to scale well if we need a new file for every item we want to benchmark, and that could potentially become many)

This benchmark can take pretty long (5 rounds takes 22 minutes, and we have previously had the default number of rounds at 25), so I think it should stay in a separate workflow. It is also not going to run on every commit, but as a cron job.

jonathan-eq:
The plan is to merge it to be able to run it with its required permissions, then if it works correctly, I will create a new PR for removing the workflow_dispatch trigger.

.github/workflows/dark_storage_benchmark.yml
```diff
@@ -0,0 +1,78 @@
...
+          id: check_for_commit
+    outputs:
+        git_new_commits: ${{steps.check_for_commit.outputs.git_new_commits}}
+
+  benchmark:
+    name: Run pytest-benchmark dark storage
+    needs: check_for_new_commits
+    if: needs.check_for_new_commits.outputs.git_new_commits == 'true'
+    timeout-minutes: 90
+    runs-on: ubuntu-latest
```
xjules, line 40: As was discussed, do we want to run on onoprem instead?

jonathan-eq: Yes, I will split it back into two workflows.

xjules (approved):
Very good @jonathan-eq ! Could you link here also the issue about recfactoring the the plot? Thanks!

jonathan-eq:
There were too many graphs, so we need to merge them in a way. #6966

PR #6932, opened 2024-01-12
Fix an issue where summary keys were not consistently named

eivindjahren:
Ensure consistent naming of summary keys. This also fixes an issue
where history keys were not handled correctly: e.g.

the history key of "BOPR:1,1,3" was before interpreted to be
"BOPR:1,1,3H" but is now interpreted as "BOPRH:1,1,3".

Whether that always makes sense in all simulators is not confirmed, but
it "BOPR:1,1,3H" is guaranteed to not be found in the summary file. The
keys that did work correctly before, `FIELD`, `OTHER`, `GROUP` and `WELL`
still works the same way.

## When applicable


src/ert/config/ert_config.py
```diff
@@ -651,12 +651,10 @@ def preferred_num_cpu(self) -> int:
...
     def _create_observations(self) -> EnkfObs:
         obs_config_file = self.model_config.obs_config_file
-        obs_time_list: List[datetime] = []
+        obs_time_list: Sequence[datetime] = []
         if self.ensemble_config.refcase is not None:
             refcase = self.ensemble_config.refcase
-            obs_time_list = [refcase.get_start_time()] + [
-                CTime(t).datetime() for t in refcase.alloc_time_vector(True)
-            ]
+            obs_time_list = [refcase[0]] + list(refcase[2])
```
andreas-el, line 657: This is so because of startdate, timemap ?

src/ert/config/ensemble_config.py
```diff
@@ -150,28 +131,32 @@ def make_field(field_list: List[str]) -> Field:
...
-        summary_keys = [item for sublist in summary_list for item in sublist]
+        to_add = list(refcase_keys)
         for key in summary_keys:
-            if "*" in key and refcase:
-                optional_keys.extend(list(refcase.keys(pattern=key)))
+            if "*" in key:
+                for i, rkey in list(enumerate(to_add))[::-1]:
+                    if fnmatch(rkey, key) and rkey != "TIME":
+                        optional_keys.append(rkey)
+                        del to_add[i]
```
andreas-el, line 149: Ok since we are reversing iteration

andreas-el (changes requested):
This is definitely moving in the right direction, but there are some failing tests that needs to be resolved at a minimum.

src/ert/config/ensemble_config.py
eivindjahren: Yup.

andreas-el (approved):
Great! 🚀

PR #6934, opened 2024-01-12
Improve license error retry

eivindjahren:
We have observed that there are two kind of events that leads to license errors

* 100% cpu, which means the server is unresponsive for ~25s
* service restart which takes 2-3 minutes

Therefore, we set the first retry to happen after 1 minute to hoping that the server gets enough time to get off 100% cpu.

The second retry happens between 3 and 6 minutes later, ensuring that there is enough time for the server to restart.

There is also a stagger factor added to avoid all retries happening simultanously. It is set to maximum 25s.

-----------------------------

## When applicable

### DIFF
--- a/src/ert/shared/share/ert/forward-models/res/script/ecl_run.py
+++ b/src/ert/shared/share/ert/forward-models/res/script/ecl_run.py
@@ -9,6 +9,7 @@ import time
 from argparse import ArgumentParser
 from collections import namedtuple
 from contextlib import contextmanager, suppress
+from random import random
 
 from ecl_config import EclrunConfig
 from packaging import version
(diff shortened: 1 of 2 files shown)


src/ert/shared/share/ert/forward-models/res/script/ecl_run.py
```diff
@@ -340,9 +341,15 @@ def execEclipse(self, eclrun_config=None):
             )
             return await_process_tee(process, sys.stdout, log_file)
 
-    LICENSE_FAILURE_SLEEP_SECONDS = 25
+    LICENSE_FAILURE_SLEEP_FACTOR = 60
+    LICENSE_RETRY_STAGGER_FACTOR = 25
 
-    def runEclipse(self, eclrun_config=None, retry=True):
+    def runEclipse(self, eclrun_config=None, retry=2, backoff_sleep=None):
```
berland, line 347: `retry` -> `retries_left: int`

eivindjahren: Fixed

oyvindeide (approved):

Issue #6958, opened 2024-01-18
Implement async version of ensemble evaluator


xjules:
The goal is the refactor `EnsembleEvaluator` to use async purely. 

On the first glance it might be doable. There a few caveats though represented by `BatchingDispatcher` (using threads for communication) and updating SnaphShots in Ensemble.

~~blocked by #7775~~

eivindjahren:
I think closing this issue should resolve https://github.com/equinor/ert/issues/3124 .

Issue #3124, opened 2022-03-20
Migrate all evaluator API to async


jondequinor:
**Describe the bug**
Ensemble Evaluator has a lot of synchronous API. It is very IO-centric, so it should be pure `asyncio`. It used to interface with a lot of sync legacy code, and I think there was some uneasiness on how to deal with it, so a lot of it was made sync.

**Expected behavior**
Developing evaluator stuff should be an exercise in calling sync stuff from async, not calling sync stuff calling async stuff calling sync stuff.

See https://github.com/equinor/ert/issues/2844 and, you know, the code…

eivindjahren:
duplicate of https://github.com/equinor/ert/issues/3124

PR #6994, opened 2024-01-23
Implement async version EnsembleEvaluator

xjules:
There are two functions in base_run_model (run_ensemble_evaluator_async and run_ensemble_evaluator). run_ensemble_evaluator is a sync vrapper for EnsembleEvaluatorAsync.
    Async ensemble evaluator follows a similar pattern to Scheduler; ie. uses a set of background tasks responsible for running the ee server, processing events, sending snapshots updates, etc.
    Async evaluator does not use BatchingDispatcher, although the events are still processed in batches.

Resolves https://github.com/equinor/ert/issues/6958
Resolves https://github.com/equinor/ert/issues/3124

## When applicable


src/ert/ensemble_evaluator/_builder/_ensemble.py
```diff
@@ -111,6 +111,9 @@ def __repr__(self) -> str:
     def evaluate(self, config: "EvaluatorServerConfig") -> None:
         pass
 
+    async def evaluate_async(self, config: "EvaluatorServerConfig") -> None:
```
berland, line 114: Should `Ensemble` be an abstract base class instead?

xjules: Maybe it should. I'd also rename the `_LegacyEnsemble` to `Ensemble` and `Ensemble` could be `BaseEnsemble` or something similar. But it should be another PR.

berland:
Can you document in the PR or the issue why we should add an extra ensemble evaluator now, instead of converting the existing to async when the legacy JobQueue is purged?

xjules:
> Can you document in the PR or the issue why we should add an extra ensemble evaluator now, instead of converting the existing to async when the legacy JobQueue is purged?

This PR is still WIP and therefore there is the `EnsembleEvaluatorAsync` as a new class as it is much easier to find bugs (as of now). Once "confident" that it works I will rename `EnsembleEvaluatorAsync` -> `EnsembleEvaluator` and make corresponding changes - in different PR I guess.

Another thing is that Async version might have a slightly different API, which might require to update lots of tests - again should be part of a new PR mentioned above.

src/ert/ensemble_evaluator/_builder/_legacy.py
```diff
@@ -120,6 +120,30 @@ def evaluate(self, config: EvaluatorServerConfig) -> None:
...
+            partialmethod(
+                self.__class__.send_cloudevent,
+                self._config.dispatch_uri,
+                token=self._config.token,
+                cert=self._config.cert,
+            ),
+        )
+        await self._evaluate_inner(
+            cloudevent_unary_send=getattr(self, ce_unary_send_method_name)
+        )
```
jonathan-eq, line 145: Can all of these lines be replaced with this?
```
await self._evaluate_inner(
   cloudevent_unary_send=partialmethod(
      self.__class__.send_cloudevent,
      self._config.dispatch_uri,
      token=self._config.token,
      cert=self._config.cert,
   )
)
```

src/ert/ensemble_evaluator/evaluator_async.py
```diff
@@ -0,0 +1,471 @@
...
+
+            for func, events in function_to_events_map.items():
+                await func(events)
+
+            self._processing_queue.task_done()
+
+    async def _dispatcher(self) -> None:
+        event_handler = {}
+
+        def set_handler(event_types: Set[str], function: Any) -> None:
```
jonathan-eq, line 131: Can we use the `Callable[...,...]` type for function here instead of `Any`?

src/ert/ensemble_evaluator/evaluator_async.py
```diff
@@ -0,0 +1,471 @@
...
+                elif client_event["type"] == EVTYPE_EE_USER_DONE:
+                    logger.debug(f"Client {websocket.remote_address} signalled done.")
+                    await self._stop()
+
+    @asynccontextmanager
+    async def count_dispatcher(self) -> AsyncIterator[None]:
+        await self._dispatchers_connected.put(None)
+        yield
+        await self._dispatchers_connected.get()
+        self._dispatchers_connected.task_done()
```
jonathan-eq, line 275: It seems we are only putting an object in the queue here, and then getting it after exiting context. Can we use a Future or Event instead, and check if it is set rather than checking if the queue is empty?

src/ert/ensemble_evaluator/evaluator_async.py
```diff
@@ -0,0 +1,471 @@
...
+        self, websocket: WebSocketServerProtocol, path: str
+    ) -> None:
+        async with self.count_dispatcher():
+            try:
+                async for msg in websocket:
+                    try:
+                        event = from_json(msg, data_unmarshaller=evaluator_unmarshaller)
+                    except cloudevents.exceptions.DataUnmarshallerError:
+                        event = from_json(msg, data_unmarshaller=pickle.loads)
+                    if self._get_ens_id(event["source"]) != self.ensemble.id_:
```
jonathan-eq, line 287: why does `self.ensemble.id_` have an underscore_ suffix?

src/ert/ensemble_evaluator/evaluator_async.py
```diff
@@ -0,0 +1,471 @@
...
+
+    @property
+    def config(self) -> EvaluatorServerConfig:
+        return self._config
+
+    @property
+    def ensemble(self) -> Ensemble:
+        return self._ensemble
+
+    async def _create_cloud_event(
```
jonathan-eq, line 206: What in this method is `async`?

xjules: you are right, this can go without async :+1:

src/ert/ensemble_evaluator/evaluator_async.py
xjules: I think that `._id` is taken by python.

src/ert/run_models/base_run_model.py
```diff
@@ -378,11 +378,21 @@ def run_ensemble_evaluator(
...
-        ).run_and_get_successful_realizations()
+        if FeatureScheduler.is_enabled(self.queue_system):
+            asyncio.set_event_loop(new_event_loop())
+            successful_realizations = get_event_loop().run_until_complete(
+                EnsembleEvaluatorAsync(
+                    ensemble,
+                    ee_config,
+                    run_context.iteration,
+                ).run_and_get_successful_realizations()
+            )
```
jonathan-eq, line 389: Do we have to set the event loop here?
ChatGPT says we should avoid altering the event loop in the current thread, and suggests doing this instead:
```
# Create a new event loop
loop = asyncio.new_event_loop()

# Run the coroutine to completion using the new event loop
successful_realizations = loop.run_until_complete(
    EnsembleEvaluatorAsync(
        ensemble,
        ee_config,
        run_context.iteration,
    ).run_and_get_successful_realizations()
)

# Close the loop after use
loop.close()
```

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -12,11 +18,58 @@
...
-from .ensemble_evaluator_utils import send_dispatch_event
+from .ensemble_evaluator_utils import TestEnsemble, send_dispatch_event
 
 
-def test_new_monitor_can_pick_up_where_we_left_off(evaluator):
-    evaluator._start_running()
+@pytest.fixture(name="evaluator_handle")
+def start_ee_in_thread(make_ee_config):
+    msg_queue = queue.Queue()
+    ee_async = None
```
jonathan-eq, line 27: Can we add typing to ee_async?

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -12,11 +18,58 @@
...
+
+        loop = new_event_loop()
+        asyncio.set_event_loop(loop)
+        loop.run_until_complete(start_ee_async())
+        loop.close()
+
+    loop_thread = ErtThread(target=_ee_start)
+    loop_thread.start()
+    yield msg_queue
+    ee_async.stop()
```
jonathan-eq, line 51: Does directly stopping an EnsembleEvaluator that was created and is running in a different thread work?

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -12,11 +18,58 @@
...
 
 
-def test_new_monitor_can_pick_up_where_we_left_off(evaluator):
-    evaluator._start_running()
+@pytest.fixture(name="evaluator_handle")
+def start_ee_in_thread(make_ee_config):
+    msg_queue = queue.Queue()
+    ee_async = None
+
+    def _ee_start():
```
jonathan-eq, line 29: Can we somehow change this function name to indicate that it is creating an event loop, and running the async function in it?
Maybe something like `_run_ee_async_in_new_eventloop`?

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -12,11 +18,58 @@
...
+            run_task = asyncio.create_task(
+                ee_async.run_and_get_successful_realizations()
+            )
+            await ee_async._server_started.wait()
+            msg_queue.put("STARTED")
+            num_realizations = await run_task
+            msg_queue.put(num_realizations)
+
+        loop = new_event_loop()
+        asyncio.set_event_loop(loop)
```
jonathan-eq, line 44: I think we might be able to remove `asyncio.set_event_loop(loop)` completely, as we are directly referencing the new event loop below anyways.

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -12,11 +18,58 @@
...
+def start_ee_in_thread(make_ee_config):
+    msg_queue = queue.Queue()
+    ee_async = None
+
+    def _ee_start():
+        async def start_ee_async():
+            nonlocal ee_async, msg_queue
+            ensemble = TestEnsemble(0, 2, 2, id_="0")
+            ee_async = EnsembleEvaluatorAsync(ensemble, make_ee_config(), 0)
+            msg_queue.put(ee_async)
```
jonathan-eq, line 34: Instead of putting the entire EnsembleEvaluatorAsync in the queue as a message, can we yield it together with the message queue? Maybe the `EnsembleEvaluatorAsync` instance can also be a nonlocal variable, and then we `yield (msg_queue, ee_async)`?

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -12,11 +18,58 @@
...
+@pytest.fixture(params=["EnsembleEvaluatorAsync", "EnsembleEvaluator"])
+def evaluator_to_use(request):
+    ee_class = request.param
+
+    if ee_class == "EnsembleEvaluator":
+        evaluator = request.getfixturevalue("evaluator")
+        evaluator._start_running()
+    elif ee_class == "EnsembleEvaluatorAsync":
+        evaluator_handle = request.getfixturevalue("evaluator_handle")
+        evaluator = evaluator_handle.get()
```
jonathan-eq, line 64: Is there a better way to pass EnsembleEvaluator without putting it as a message in the message queue?

jonathan-eq (changes requested):
Great work @xjules! More stuff async might help with some of the bugs we are seeing👌

src/ert/run_models/base_run_model.py
xjules: You might be right. I'll give it a try.

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
xjules: Good idea, but ... I wanted to differentiate between the created `EnsembleEvaluatorAsync` and `STARTED` - meaning that all the tasks have been started.

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
xjules: The message queue is thread safe and thus is ideal for sending / passing variables from async context to sync one.

src/ert/run_models/base_run_model.py
jonathan-eq: Is this what did not work?

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -12,11 +18,58 @@
...
 
-from .ensemble_evaluator_utils import send_dispatch_event
+from .ensemble_evaluator_utils import TestEnsemble, send_dispatch_event
 
 
-def test_new_monitor_can_pick_up_where_we_left_off(evaluator):
-    evaluator._start_running()
+@pytest.fixture(name="evaluator_handle")
+def start_ee_in_thread(make_ee_config):
+    msg_queue = queue.Queue()
```
jonathan-eq, line 26: Should we use `asyncio.Queue` instead?

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
jonathan-eq: Okay, but can we add typing to line 63 then? Maybe something like 
`evaluator_handle: queue.Queue[Union[str, EnsembleEvaluator]]` would work?

src/ert/ensemble_evaluator/_builder/_legacy.py
```diff
@@ -120,6 +120,30 @@ def evaluate(self, config: EvaluatorServerConfig) -> None:
...
+
```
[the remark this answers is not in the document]
xjules: I've checked the comment, which states there should be an async send could event function, but there isn't. So not sure what that comment refers to.

src/ert/ensemble_evaluator/_builder/_legacy.py
xjules: Unfortunately it din't work so need to keep it as is.

jonathan-eq: Why did it not work?
There is probably a reason why they made it this way, but I am curious to know why.

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
jonathan-eq: Can we add typing to what can go in the queue?

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
xjules: I have moved the fixture to conftest and there are not annotations, so will leave it as is I guess

src/ert/ensemble_evaluator/_builder/_legacy.py
xjules: It complained (when starting the ee loop) that there is no running loop.

xjules: The signature of the function changed as if the partialmehod does not preserve the async property. In this case though we attach partial to a class member, which does some magic apparently.

src/ert/ensemble_evaluator/evaluator_async.py
```diff
@@ -0,0 +1,479 @@
...
+            batch: List[
+                Tuple[Callable[[List[CloudEvent]], Awaitable[None]], CloudEvent]
+            ] = []
+            start_time = asyncio.get_running_loop().time()
+            while (
+                len(batch) < self._max_batch_size
+                and asyncio.get_running_loop().time() - start_time
+                < self._batching_interval
+            ):
+                try:
```
yngve-sk, line 157: I'm a bit confused by this: `handle_dispatch` reads messages from a websocket, which are turned into events put into an async queue. Then the `_dispatcher` task gets those events from that queue and parses them into (event function, args) pairs before putting it into processing queues. I think it makes sense somewhat because it handles websocket->queue->processing buffer, but the naming of the functions or something like that could maybe somehow make this more clear

yngve-sk (commented):
Seems overall sensible afaik. A bit confused as to difference between messages (seems to pertain to the state of the ensemble) and events which seem to go through dispatch, which comes from I'm not sure where, monitor/tracker or something like that? Could maybe be clarified somehow

src/ert/ensemble_evaluator/evaluator_async.py
```diff
@@ -0,0 +1,479 @@
...
+    async def _fm_handler(self, events: List[CloudEvent]) -> None:
+        await self._append_message(self.ensemble.update_snapshot(events))
+
+    async def _started_handler(self, events: List[CloudEvent]) -> None:
+        if self.ensemble.status != ENSEMBLE_STATE_FAILED:
+            await self._append_message(self.ensemble.update_snapshot(events))
+
+    async def _stopped_handler(self, events: List[CloudEvent]) -> None:
+        if self.ensemble.status != ENSEMBLE_STATE_FAILED:
+            self._result = events[0].data  # normal termination
```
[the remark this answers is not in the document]
xjules: There it should be  only a single event yes.

src/ert/ensemble_evaluator/evaluator_async.py
```diff
@@ -0,0 +1,479 @@
...
+        self._loop: Optional[asyncio.AbstractEventLoop] = None
+        self._done: asyncio.Future[bool] = asyncio.Future()
+
+        self._clients: Set[WebSocketServerProtocol] = set()
+        self._dispatchers_connected: asyncio.Queue[None] = asyncio.Queue()
+
+        self._events: asyncio.Queue[CloudEvent] = asyncio.Queue()
+        self._messages: asyncio.Queue[str] = asyncio.Queue()
+
+        self._result = None
```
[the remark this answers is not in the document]
xjules: I can improve the name perhaps, but `_events` are received events while `_messages` and events to be sent.

src/ert/ensemble_evaluator/evaluator_async.py
xjules: > I'm a bit confused by this: `handle_dispatch` reads messages from a websocket, which are turned into events put into an async queue. Then the `_dispatcher` task gets those events from that queue and parses them into (event function, args) pairs before putting it into processing queues. I think it makes sense somewhat because it handles websocket->queue->processing buffer, but the naming of the functions or something like that could maybe somehow make this more clear

That's essentially the async version of batch processing, wherein nothing is blocking processing tasks and all exists in their own queues. Will try to improve the naming though.

xjules:
I have tested LSF cluster with EEAsync and it worked smoothly.
Nevertheless, when cancelling some realizations are shown as running
image

Need to find out if it is related to EEAsync closing server too fast.

src/ert/ensemble_evaluator/evaluator_async.py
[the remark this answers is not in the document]
jonathan-eq: Maybe we can rename `_events`-> `_received_events`?

src/ert/ensemble_evaluator/evaluator_async.py
```diff
@@ -0,0 +1,476 @@
...
+        self._batch_processing_queue: asyncio.Queue[
+            List[Tuple[Callable[[List[CloudEvent]], Awaitable[None]], CloudEvent]]
+        ] = asyncio.Queue()
+        self._max_batch_size: int = 500
+        self._batching_interval: int = 2
+
+    async def _publisher(self) -> None:
+        try:
+            while True:
+                msg = await self._messages_to_send.get()
```
jonathan-eq, line 91: Can we change `msg` to `msg_to_send`? It is a bit confusing because we use `...get()` on `messages_to_send`

jonathan-eq: Can we change msg to msg_to_send here too?

src/ert/ensemble_evaluator/evaluator_async.py
```diff
@@ -0,0 +1,476 @@
...
+                await asyncio.gather(
+                    *[client.send(msg) for client in self._clients],
+                    return_exceptions=True,
+                )
+        except asyncio.CancelledError:
+            # when cancelling the task, make sure to send the rest
+            while not self._messages_to_send.empty():
+                msg = self._messages_to_send.get_nowait()
```
jonathan-eq, line 99: Can we change msg to msg_to_send here too?

src/ert/ensemble_evaluator/evaluator_async.py
```diff
@@ -0,0 +1,476 @@
...
+        )
+        if message:
+            await self._messages_to_send.put(message)
+
+    async def _process_event_buffer(self) -> None:
+        while True:
+            batch = await self._batch_processing_queue.get()
+            function_to_events_map: Dict[
+                Callable[[List[CloudEvent]], Awaitable[None]], List[CloudEvent]
+            ] = OrderedDict()
```
jonathan-eq, line 118: Do we need to preserve insert order here, or can we use a normal dict instead?

xjules:
TODO: fix /private/jparu/workspace/ert/kenv/root/lib64/python3.8/site-packages/ert/ensemble_evaluator/_wait_for_evaluator.py to use new async version.

src/ert/ensemble_evaluator/evaluator_async.py
xjules: I think this is fine in this case, we we are polling the msg from the queue.

src/ert/ensemble_evaluator/evaluator_async.py
xjules: Fixed by using a normal dict.

src/ert/ensemble_evaluator/evaluator_async.py
xjules: Cancellation is now not required as join is called on `_messages_to_send` when closing down the server.

jonathan-eq:
We might need more tests

src/ert/run_models/base_run_model.py
```diff
@@ -539,8 +542,53 @@ async def run_monitor(self, ee_config: EvaluatorServerConfig) -> bool:
 
         return True
 
+    async def run_ensemble_evaluator_async(
+        self, run_context: RunContext, ee_config: EvaluatorServerConfig
+    ) -> List[int]:
+        if not self._end_queue.empty():
+            event_logger.debug("Run model canceled - pre evaluation")
```
jonathan-eq, line 549: Typo in cancelled

src/ert/run_models/base_run_model.py
```diff
@@ -539,8 +542,53 @@ async def run_monitor(self, ee_config: EvaluatorServerConfig) -> bool:
...
+            self._end_queue.get()
```
jonathan-eq, line 550: Why do we get one of the items before returning the empty list?

src/ert/run_models/base_run_model.py
```diff
@@ -539,8 +542,53 @@ async def run_monitor(self, ee_config: EvaluatorServerConfig) -> bool:
...
+        ensemble = self._build_ensemble(run_context)
+        evaluator = EnsembleEvaluatorAsync(
+            ensemble,
+            ee_config,
+            run_context.iteration,
+        )
+        evaluator_task = asyncio.create_task(
+            evaluator.run_and_get_successful_realizations()
+        )
+        if not (await self.run_monitor(ee_config)):
```
jonathan-eq, line 561: Is this only running the monitor, or is it also waiting until experiment is done?

tests/unit_tests/ensemble_evaluator/ensemble_evaluator_utils.py
```diff
@@ -75,6 +76,103 @@ def __init__(self, _iter, reals, jobs, id_):
...
+                job_failed = False
+                await send_dispatch_event_async(
+                    dispatch,
+                    identifiers.EVTYPE_REALIZATION_UNKNOWN,
+                    f"/ert/ensemble/{self.id_}/real/{real}",
+                    f"event-{event_id}",
+                    None,
+                )
+                event_id = event_id + 1
+                for job in range(0, self.jobs):
```
jonathan-eq, line 116: Do we need to specify 0 as the start here?

jonathan-eq (commented):
Close to merge-able :) 
See nitpicking changes.

src/ert/run_models/base_run_model.py
xjules: hm.... You might be right. Need to rethink this a bit

berland:
Code coverage claims that 29 lines are missing coverage, could you check these and see if we need some more testing?

src/ert/ensemble_evaluator/evaluator_async.py
```diff
@@ -0,0 +1,474 @@
...
+            snapshot_update_event.to_dict(),
+        )
+        if message:
+            await self._messages_to_send.put(message)
+
+    async def _process_event_buffer(self) -> None:
+        while True:
+            batch = await self._batch_processing_queue.get()
+            function_to_events_map: Dict[
+                Callable[[List[CloudEvent]], Awaitable[None]], List[CloudEvent]
```
jonathan-eq, line 108: The inner type is used here and on line 82. Can we put it outside the class for reuse and give it a more descriptive name?

src/ert/ensemble_evaluator/evaluator_async.py
```diff
@@ -0,0 +1,474 @@
...
+                event_handler[event_type] = function
+
+        for e_type, f in (
+            (EVGROUP_FM_ALL, self._fm_handler),
+            ({EVTYPE_ENSEMBLE_STARTED}, self._started_handler),
+            ({EVTYPE_ENSEMBLE_STOPPED}, self._stopped_handler),
+            ({EVTYPE_ENSEMBLE_CANCELLED}, self._cancelled_handler),
+            ({EVTYPE_ENSEMBLE_FAILED}, self._failed_handler),
+        ):
+            set_event_handler(e_type, f)
```
jonathan-eq, line 134: Instead of having this as a loop and manually creating the set of item; can we call the function on the items one by one instead?

src/ert/ensemble_evaluator/evaluator_async.py
```diff
@@ -0,0 +1,474 @@
...
+    async def _stopped_handler(self, events: List[CloudEvent]) -> None:
+        if self.ensemble.status != ENSEMBLE_STATE_FAILED:
+            self._result = events[0].data  # normal termination
+            max_memory_usage = -1
+            for job in self.ensemble.snapshot.get_all_forward_models().values():
+                memory_usage = job.max_memory_usage or "-1"
+                if int(memory_usage) > max_memory_usage:
+                    max_memory_usage = int(memory_usage)
+            logger.info(
+                f"Ensemble ran with maximum memory usage for a single realization job: {max_memory_usage}"
```
jonathan-eq, line 171: Hmm, do you think here we should replace `...single realization job` -> `...single realization forward model` or `...forward model step`?

src/ert/ensemble_evaluator/evaluator_async.py
```diff
@@ -0,0 +1,474 @@
...
+            )
+            await self._append_message(self.ensemble.update_snapshot(events))
+
+    async def _cancelled_handler(self, events: List[CloudEvent]) -> None:
+        if self.ensemble.status != ENSEMBLE_STATE_FAILED:
+            await self._append_message(self.ensemble.update_snapshot(events))
+            await self._stop()
+
+    async def _failed_handler(self, events: List[CloudEvent]) -> None:
```
jonathan-eq, line 180: Probably for a different PR, but do you think events has to be of type list, or could/should it be a more general sequence?

src/ert/ensemble_evaluator/evaluator_async.py
xjules: But this is not a biggy, since this runs only once

src/ert/ensemble_evaluator/evaluator_async.py
xjules: Right that makes sense. Will fix...

src/ert/ensemble_evaluator/evaluator_async.py
xjules: What to put it outside the class?

src/ert/ensemble_evaluator/evaluator_async.py
xjules: It could but that is indeed a different PR.

src/ert/ensemble_evaluator/evaluator_async.py
jonathan-eq: We can use it as a type alias, that is how we did with the long composite types in the new drivers. We could also have it as a class variable.

src/ert/ensemble_evaluator/evaluator_async.py
jonathan-eq: I think it would be more readable to hard code it directly
```
            set_event_handler(EVGROUP_FM_ALL, self._fm_handler)
            set_event_handler({EVTYPE_ENSEMBLE_STARTED}, self._started_handler)
            set_event_handler({EVTYPE_ENSEMBLE_STOPPED}, self._stopped_handler)
            set_event_handler({EVTYPE_ENSEMBLE_CANCELLED}, self._cancelled_handler)
            set_event_handler({EVTYPE_ENSEMBLE_FAILED}, self._failed_handler)
```

xjules: You might be right there :+1:

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -11,20 +20,96 @@
...
+        ee.start_running()
+        yield ee
+        ee.stop()
+    elif ee_class == "EnsembleEvaluatorAsync":
+        ee_async = EnsembleEvaluatorAsync(ensemble, make_ee_config(), 0)
+        run_task = asyncio.create_task(ee_async.run_and_get_successful_realizations())
+        await ee_async._server_started.wait()
+        yield ee_async
+        await ee_async._stop()
+        await run_task
```
jonathan-eq, line 98: Very nice!

src/ert/ensemble_evaluator/evaluator_async.py
```diff
@@ -0,0 +1,469 @@
...
+                    function_to_events_map[func] = []
+                function_to_events_map[func].append(event)
+
+            for func, events in function_to_events_map.items():
+                await func(events)
+
+            self._batch_processing_queue.task_done()
+
+    async def _batch_events_into_buffer(self) -> None:
+        event_handler = {}
```
jonathan-eq, line 121: Can we add typing `Dict[str, EVENT_HANDLER]` here?

xjules: it's not really a string tbh, more like `set[str]`

xjules: done

eivindjahren (approved):
Amazing!

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -364,53 +375,35 @@ async def evaluator_server(self) -> None:
...
-    def _stop(self) -> None:
-        if not self._done.done():
-            self._done.set_result(None)
+    async def _stop(self) -> None:
+        self._server_done.set()
 
     def stop(self) -> None:
-        self._loop.call_soon_threadsafe(self._stop)
-        self._ws_thread.join()
+        assert self._loop
```
eivindjahren, line 402: this method is no longer called in the tests.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -171,22 +187,16 @@ def _failed_handler(self, events: List[CloudEvent]) -> None:
...
-    async def _send_snapshot_update(
-        self, snapshot_update_event: PartialSnapshot
-    ) -> None:
-        message = self._create_cloud_message(
-            EVTYPE_EE_SNAPSHOT_UPDATE,
-            snapshot_update_event.to_dict(),
-        )
-        await self._send_message(message)
+    @property
+    def config(self) -> EvaluatorServerConfig:
```
eivindjahren, line 194: This property is unused in the tests.

src/ert/ensemble_evaluator/evaluator.py
xjules: It can be removed :+1:

src/ert/ensemble_evaluator/evaluator.py
xjules: checking if it is alright to remove this too.

PR #7008, opened 2024-01-23
Scheduler: Pick last MAX_RUNNING rather than first

pinkwah:


src/ert/config/queue_config.py
```diff
@@ -108,10 +108,11 @@ def create_local_copy(self) -> QueueConfig:
 
     @property
     def max_running(self) -> int:
+        max_running = 0
         for key, val in self.queue_options.get(self.queue_system, []):
             if key == "MAX_RUNNING":
-                return int(val)
-        return 0
+                max_running = int(val)
+        return max_running
```
xjules, line 115: I think `return 0` was correct as this implies all realizations and we should have (in Scheduler) 
`if max_running==0 max_running=len(self._jobs)`

src/ert/shared/share/ert/site-config
```diff
@@ -8,6 +8,3 @@ INSTALL_JOB_DIRECTORY forward-models/res
 INSTALL_JOB_DIRECTORY forward-models/shell
 INSTALL_JOB_DIRECTORY forward-models/templating
 INSTALL_JOB_DIRECTORY forward-models/old_style
-
```
xjules, line 11: Let's not remove it.

xjules (approved):
Splendid! :rocket:

PR #7039, opened 2024-01-29
Check scheduler loop is still running before killing all the jobs.

DanSava:
**Issue**
Resolves: https://github.com/equinor/everest/issues/2191

**Approach**
Check scheduler loop is still running before killing all the jobs. 

## When applicable

### DIFF
--- a/src/ert/scheduler/scheduler.py
+++ b/src/ert/scheduler/scheduler.py
@@ -102,7 +102,10 @@ class Scheduler:
 
     def kill_all_jobs(self) -> None:
         assert self._loop
-        asyncio.run_coroutine_threadsafe(self.cancel_all_jobs(), self._loop)
+        # Checking that the loop is running is required because everest is closing the
+        # simulation context whenever an optimization simulation batch is done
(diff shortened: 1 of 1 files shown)


xjules (approved):
Nice catch @DanSava! 🚀

src/ert/scheduler/scheduler.py
```diff
@@ -102,7 +102,10 @@ def __init__(
 
     def kill_all_jobs(self) -> None:
         assert self._loop
-        asyncio.run_coroutine_threadsafe(self.cancel_all_jobs(), self._loop)
+        # Checking that the loop is running is required because everest is closing the
+        # simulation context every an optimization simulation batch is done
```
andreas-el, line 106: I understand the comment, but did you mean `whenever` instead of `every` ? This is nitpick but the there is something not quite right with this.

DanSava: I wanted to write `every time` but I think `whenever` is better :)

andreas-el (approved):
Brilliant! 🚄

Issue #7048, opened 2024-01-30
Rename TorqueDriver to something accurate


pinkwah:
The new Scheduler TorqueDriver does not conform to what the Torque API specifies. It is neither PBS. The differences are as follows:

- When using `qstat`, we need to add `-x` for the system to output finished jobs. In other systems (SLURM Torque) this is the default and `-x` is not a valid option.
- When using `qstat`, we use `-Fjson` to format the output as JSON. This is not an option in SLURM Torque.
- When a job finishes, `qstat` outputs its `job_status` as `F`. This is documented nowhere.

We think we're using OpenPBS, but the documentation for it is scarce. We should match the documentation and rename the class to `OpenPBSDriver`.

Alternatively, we can rename the driver to `AzurePBSDriver` or `EquinorPBSDriver` to signify that there is some odd stuff going on.

PR #7061, opened 2024-01-31
Rename TorqueDriver to OpenPBSDriver

pinkwah:
This more accurately represents the HPC system that we are targetting.

Resolves: #7048


xjules (approved):
Nice! :+1:

PR #7111, opened 2024-02-05
Remove benchmarking files not in use

jonathan-eq:


berland:
How to assert that these are not in use? But I have enough faith, so 🥓

berland (approved):

jonathan-eq:
I am pretty sure that I added them back when I first introduced the dark storage benchmarking workflow

PR #7131, opened 2024-02-07
Refactor to Native Python BlockFs migration

pinkwah:


src/ert/storage/migration/block_fs.py
```diff
@@ -26,14 +40,203 @@
...
 
+    DType = TypeVar("DType", np.float32, np.float64)
+
+
+_SIZEOF_I32 = 4
+_SIZEOF_I64 = 8
+_SIZEOF_F64 = 8
+
+
+class Kind(IntEnum):
```
eivindjahren, line 59: So maybe `_Kind` ? And that goes for all other identifiers that aren't exported.

eivindjahren (approved):
:+1:

PR #7152, opened 2024-02-09
Add backup path in torque mock binaries

### LINKED ISSUE #7138: Running poly with Scheduler and mocked Torque queue hangs
It should be possible to manipulate `$PATH` and then run the poly_example with the new scheduler and the mocked scripts that translates Torque to local queue:

Something like
```
export PATH=<something>/tests/integration_tests/scheduler/bin:$PATH
ert ensemble_experiment --enable-scheduler poly_torque.ert
```

(where `poly_torque.ert` is just `poly.ert` with `QUEUE_SYSTEM TORQUE`)

This currently hangs with realizations just pending.

jonathan-eq:
Before this commit, the torque mock binaries put the jobs in the root directory if PYTEST_TMP_PATH was not set. This commit sets the cwd as backup instead.

**Issue**
Resolves #7138


tests/integration_tests/scheduler/bin/qstat.py
```diff
@@ -25,7 +25,11 @@ def main() -> None:
     args = parse_args()
     assert args.F == "json", "Mock qstat must have -Fjson"
 
-    jobs_path = Path(os.environ["PYTEST_TMP_PATH"]) / "mock_jobs"
```
berland, line 28: I think you can use `os.environ.get("PYTEST_TMP_PATH", ".")` here.

berland:
If this is successful, we should have an integration test that runs the poly-case with this mock.

berland:
> If this is successful, we should have an integration test that runs the poly-case with this mock.

I think there are more traps, but this PR seems to work on its own.

berland (approved):

PR #7153, opened 2024-02-09
Always say job has started if an event is received

berland:
If the queue system is very quick and the polling time period is long enough, the driver will never observe the job while it is in its running state.


berland:
Currently lacking an idea for how to neatly test this.

jonathan-eq (approved):
LGTM!

PR #7178, opened 2024-02-13
Remove cwd argument from python drivers

### LINKED ISSUE #7147: OpenPBSDriver - consider leaving submission script in runpath
The script submitted to the queue system:
https://github.com/equinor/ert/blob/5072a4eb4c17a20d45efa81df191c5728518d5fe/src/ert/scheduler/openpbs_driver.py#L53-L56

is kept only in memory. When things fail, this makes it more difficult to reproduce. The C-driver leaves this script in the runpath and makes manual resubmission possible f.ex. `qstat -f -x` gives more traceability when run through the C-driver:

```
    Submit_arguments = -k oe -l nodes=1:ppn=1 -q permanent -N poly_5 -r n /priv
        ate/havb/projects/ert/test-data/poly_example/poly_out/realization-5/ite
        r-0/qsub_script.sh
```
while the Python PBS driver only gives this information:
```
    Submit_arguments = -koe -rn -Npoly_1 -q permanent -
```

berland:
Whether to change directory to anything is the responsibility of the submitted script, the driver will not any longer guarantee it.

This makes it possible to unify the PBS and the LSF driver slightly.

The LSF driver still wants to write to lsf_info.json, thus it needs to know the runpath.

**Issue**
Resolves #7147  (it does not do exactly as suggested in the issue, but it will make it reproducible from the logs by going directly to job_dispatch as for LSF)

**Approach**

## When applicable

### DIFF
--- a/src/ert/scheduler/lsf_driver.py
+++ b/src/ert/scheduler/lsf_driver.py
@@ -87,21 +87,18 @@ class LsfDriver(Driver):
         self._retry_sleep_period = 3
 
     async def submit(
-        self, iens: int, executable: str, /, *args: str, cwd: str, name: str = "dummy"
+        self,
+        iens: int,
+        executable: str,
+        /,
+        *args: str,
+        name: str = "dummy",
+        runpath: Optional[str] = None,
     ) -> None:
-        script = executable
         arg_queue_name = ["-q", self._queue_name] if self._queue_name else []
 
         bsub_with_args: List[str] = (
-            [str(self._bsub_cmd)]
-            + arg_queue_name
-            + [
-                "-J",
-                name,
-                script,
-                # skipping args for now, it might not work?
-                cwd,  # assuming job_dispatch.py will handle cwd
-            ]
+            [str(self._bsub_cmd)] + arg_queue_name + ["-J", name, executable, *args]
         )
         logger.debug(f"Submitting to LSF with command {shlex.join(bsub_with_args)}")
         process = await asyncio.create_subprocess_exec(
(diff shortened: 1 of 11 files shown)


berland:
Tested with komodoenv from bleeding, and this passes the poly case with real PBS cluster.

src/ert/scheduler/openpbs_driver.py
```diff
@@ -51,31 +51,33 @@ def __init__(self, *, queue_name: Optional[str] = None) -> None:
...
-        self, iens: int, executable: str, /, *args: str, cwd: str, name: str = "dummy"
+        self,
+        iens: int,
+        executable: str,
+        /,
+        *args: str,
+        name: str = "dummy",
+        runpath: Optional[str] = None,
     ) -> None:
-        script = (
```
xjules, line 56: This is a higher "math" for me :smile:  Why this wrapper script is not required anymore?

berland: There are two ways to submit to `qsub`. Either pipe your script via stdin (current code) or supply a script name and args after `--`. This PR does the latter, and instead of putting a script there, we jump directly to job_dispatch.py.

src/ert/scheduler/lsf_driver.py
```diff
@@ -87,21 +87,18 @@ def __init__(
...
 
     async def submit(
-        self, iens: int, executable: str, /, *args: str, cwd: str, name: str = "dummy"
+        self,
+        iens: int,
+        executable: str,
+        /,
+        *args: str,
+        name: str = "dummy",
+        runpath: Optional[str] = None,
```
xjules, line 96: `runpath` refers to `job_runpath`. Just making sure we use it correctly as it is quite overloaded term.

berland: hmm, I think runpath is always one of the directories `realization-*/iter-*`, where does it mean anything else?

xjules: Yes, I double-checked it is `realization-*/iter-*` so it's good :+1:

xjules (approved):
Very good job @berland ! Let's bring it in. :rocket:

Issue #7186, opened 2024-02-14
Harden Python LSF driver towards faulty bsub/bjobs/bkill


berland:
* Write tests that can mock faulty return codes and garbage output on stdout/stderr from the LSF commands bsub, bjobs, bkill
* Assert that ERT will handle these errors in some way. Identify scenarios that ERT should recognize and do a retry, error hard when relevant.
* Fall back to bhist: #7208

PR #7188, opened 2024-02-14
Tune error handling in Python LSF driver for bsub

berland:
No known error scenarios for which the bsub command should be retried is currently known.

Add tests for the behaviour in error conditions

**Issue**
Partly solves #7186 

**Approach**
Add tests and add simple checks for wrong output from bsub.

## When applicable


src/ert/scheduler/lsf_driver.py
```diff
@@ -107,19 +107,26 @@ async def submit(
...
+            and "is submitted to " in stdout_decoded
+            and "queue" in stdout_decoded
+        ):
+            raise RuntimeError(f"Could not understand {stdout_decoded} from bsub")
+
+        job_id = (
+            stdout.decode("utf-8").strip().replace("<", "").replace(">", "").split()[1]
+        )
+        if not (job_id.isdigit() and int(job_id) > 0):
+            raise RuntimeError(f"job_id {job_id} is not a positive number")
```
pinkwah, line 129: `import re`

```suggestion
        match = re.search(b"Job <([0-9]+)> is submitted to .+ queue")
        if match is None:
            raise RuntimeError(f"Could not understand {repr(stdout.decode())} from bsub")

        job_id = int(match[1])
```

If you want to explicitly check for a negative value, you can the regex to `b"Job <(-?[0-9]+])> is submitted to .+ queue"` and then check whether `int(match[1])` is negative.

tests/integration_tests/scheduler/test_lsf_driver.py
```diff
@@ -153,3 +157,59 @@ async def finished(iens, returncode, aborted):
...
     await driver.submit(0, f"exit {actual_returncode}")
     await poll(driver, {0}, finished=finished)
+
+
+@pytest.mark.parametrize(
+    "bsub_script, expectation",
+    [
+        pytest.param(
+            "echo 'Job <444> is submitted to default queue'",
+            does_not_raise(),
```
pinkwah, line 167: Cool trick. I've had to write some tests where I parametrise on whether it throws or not and this is a neat solution.

src/ert/scheduler/lsf_driver.py
berland: Great. Not sure why I wanted to avoid regexp in the first place, this is neater. Skipping treating negative number specifically, and for simplicity I will allow a zero job id. I have no reference to say that a zero is not permitted.

berland:
tests passing again. semeio is failing, but there is a pr for that too.

src/ert/scheduler/lsf_driver.py
```diff
@@ -107,19 +108,24 @@ async def submit(
...
-                .split()[1]
-            )
-        except IndexError as err:
+        if process.returncode:
             logger.error(
-                f"Command \"{' '.join(bsub_with_args)}\" failed with error message: {stderr.decode()}"
+                f"Command \"{' '.join(bsub_with_args)}\" failed with "
+                f"returncode {process.returncode} and error message: {stderr.decode() or '<empty>'}"
+            )
+            raise RuntimeError(stderr.decode())
```
xjules, line 116: Shall we do `RuntimeError(f"Command \"{' '.join(bsub_with_args)}\" failed with "
                f"returncode {process.returncode} and error message: {stderr.decode() or '<empty>'}")`?

berland: I guess my approach makes the error message end up in potentially two log files, while going to RuntimeError directly makes it go only one place (maybe not predetermined where it ends). 

I guess my thought was that it is nice when debugging to have this in the `jobqueue-` log file because that is a file that you most likely are looking at in any case.

src/ert/scheduler/lsf_driver.py
```diff
@@ -107,19 +108,24 @@ async def submit(
...
+
+        try:
+            stdout_decoded = stdout.decode()
+        except UnicodeDecodeError as err:
+            raise RuntimeError("Could not understand binary output from bsub") from err
+
+        match = re.search("Job <([0-9]+)> is submitted to .+ queue", stdout_decoded)
+        if match is None:
+            raise RuntimeError(
+                f"Could not understand '{repr(stdout.decode())}' from bsub"
```
pinkwah, line 126: Nitpick: If you're going to have `stdout_decoded`, you should use it here as well.

```suggestion
                f"Could not understand '{repr(stdout_decoded)}' from bsub"
```

pinkwah (approved):
LGTM with a nitpick

src/ert/scheduler/lsf_driver.py
berland: nit picked.

PR #7192, opened 2024-02-14
Fix muted exceptions in queue drivers

### LINKED ISSUE #7142: Exceptions in queue drivers are muted
Adding say `1/0` in line 90 in

https://github.com/equinor/ert/blob/main/src/ert/scheduler/openpbs_driver.py#L85-L92

will make the test (`test_openpbs_driver.py`) hang. The exception is never observed.

This makes queue driver development harder than it should.

jonathan-eq:
**Issue**
Resolves #7142

### DIFF
--- a/src/clib/lib/job_queue/torque_driver.cpp
+++ b/src/clib/lib/job_queue/torque_driver.cpp
@@ -671,6 +671,7 @@ void *torque_driver_submit_job(void *_driver, std::string submit_cmd,
 static job_status_type
 torque_driver_get_qstat_status(torque_driver_type *driver,
                                const char *jobnr_char) {
+    1/0
     constexpr int OUTPUT_FILE_SIZE = 32;
     char tmp_std_file[OUTPUT_FILE_SIZE];
     strncpy(tmp_std_file, "/tmp/ert-qstat-std-XXXXXX", OUTPUT_FILE_SIZE);
--- a/src/clib/lib/job_queue/torque_driver.cpp
+++ b/src/clib/lib/job_queue/torque_driver.cpp
@@ -671,7 +671,6 @@ void *torque_driver_submit_job(void *_driver, std::string submit_cmd,
 static job_status_type
 torque_driver_get_qstat_status(torque_driver_type *driver,
                                const char *jobnr_char) {
-    1/0
     constexpr int OUTPUT_FILE_SIZE = 32;
     char tmp_std_file[OUTPUT_FILE_SIZE];
     strncpy(tmp_std_file, "/tmp/ert-qstat-std-XXXXXX", OUTPUT_FILE_SIZE);
--- a/src/ert/scheduler/scheduler.py
+++ b/src/ert/scheduler/scheduler.py
@@ -7,17 +7,10 @@ import os
 import ssl
 import time
 from collections import defaultdict
+from contextlib import suppress
 from dataclasses import asdict
 from pathlib import Path
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Dict,
-    List,
-    MutableMapping,
-    Optional,
-    Sequence,
-)
+from typing import TYPE_CHECKING, Any, Dict, MutableMapping, Optional, Sequence
 
 from pydantic.dataclasses import dataclass
 from websockets import Headers
@@ -208,20 +201,18 @@ class Scheduler:
             self._tasks[iens] = asyncio.create_task(job(start, sem, self._max_submit))
         start.set()
 
-        async def gather_jobs() -> List[BaseException | None]:
-            return asyncio.gather(*self._tasks.values(), return_exceptions=True)
-
         try:
-            job_results = (await asyncio.gather(gather_jobs(), *long_lived_tasks))[0]
+            job_results = (
+                await asyncio.gather(
+                    asyncio.gather(*self._tasks.values(), return_exceptions=True),
+                    *long_lived_tasks,
+                )
+            )[0]
         except Exception as e:
-            print("DEBUG we failed correctly*******⋅⋅⋅!!!!!!!!!!!!!!!!!!!!!!")
             for job_task in self._tasks.values():
                 job_task.cancel()
-                try:
+                with suppress(RuntimeError):
                     await job_task
-                except RuntimeError:
-                    logger.error("Cannot kill job!")
-                    continue
             logger.error(str(e))
             raise e
         finally:
@@ -233,19 +224,6 @@ class Scheduler:
             if isinstance(result, Exception):
                 logger.error(result)
                 raise result
-
-        # job_results: Optional[List[BaseException | None]] = None
-        # try:
-        #     job_results = (await asyncio.gather(gather_jobs(), *long_lived_tasks))[0]
-        # except asyncio.CancelledError:
-        #     for job_task in self._tasks.values():
-        #         job_task.cancel()
-        # for result in job_results or []:
-        #     if isinstance(result, asyncio.CancelledError):
-        #         continue
-        #     if isinstance(result, Exception):
-        #         logger.error(result)
-        #         raise result
 
         await self.driver.finish()
 
--- a/src/ert/scheduler/scheduler.py
+++ b/src/ert/scheduler/scheduler.py
@@ -9,13 +9,20 @@ import time
 from collections import defaultdict
 from dataclasses import asdict
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Dict, MutableMapping, Optional, Sequence
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    List,
+    MutableMapping,
+    Optional,
+    Sequence,
+)
 
 from pydantic.dataclasses import dataclass
 from websockets import Headers
 from websockets.client import connect
 
-from ert.async_utils import background_tasks
 from ert.constant_filenames import CERT_FILE
 from ert.job_queue.queue import EVTYPE_ENSEMBLE_CANCELLED, EVTYPE_ENSEMBLE_STOPPED
 from ert.scheduler.driver import Driver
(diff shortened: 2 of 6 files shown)


src/ert/async_utils.py
```diff
@@ -29,18 +33,14 @@ async def background_tasks() -> AsyncGenerator[Any, Any]:
     tasks: MutableSequence[asyncio.Task[Any]] = []
 
     def add(coro: Coroutine[Any, Any, Any]) -> None:
-        tasks.append(asyncio.create_task(coro))
+        tasks.append(_create_task(asyncio.get_event_loop(), coro))
```
berland, line 36: sure this line is necessary to change?

tests/unit_tests/scheduler/test_scheduler.py
```diff
@@ -444,3 +445,14 @@ async def wait():
         for start, next_start in zip(run_start_times[:-1], run_start_times[1:])
     ]
     assert min(deltas) >= submit_sleep * 0.8
+
+
+@pytest.mark.timeout(5)
+async def test_exceptions_are_propagated(realization):
+    driver = Mock()
+    driver.poll = lambda: 1 / 0
```
berland, line 453: do the same with _process_event_queue and _publisher?

src/ert/async_utils.py
jonathan-eq: No, you are right, I will undo it.

tests/unit_tests/scheduler/test_scheduler.py
jonathan-eq: Yes

jonathan-eq:
Now something else is hanging. I am working on debugging `tests/integration_tests/test_cli.py::test_that_stop_on_fail_workflow_jobs_stop_ert`. It seems like EE and the tracker thread are hanging after the experiment has exited. Also, the job status is set to finished and not failed, which seems odd...

src/ert/scheduler/scheduler.py
```diff
@@ -197,14 +200,10 @@ async def execute(
                 self._tasks[iens] = asyncio.create_task(
                     job(start, sem, self._max_submit)
                 )
-
             start.set()
-            results = await asyncio.gather(
-                *self._tasks.values(), return_exceptions=True
+            await asyncio.gather(
+                *self._tasks.values(), *background_task_list, return_exceptions=False
```
berland, line 205: I don't think you want to wait for background_task_list here. These are just waiting to be cancelled and will not stop by themselves.

src/ert/async_utils.py
```diff
@@ -74,6 +76,6 @@ def _done_callback(task: asyncio.Task[_T_co]) -> None:
         if (exc := task.exception()) is None:
             return
 
-        logger.error(f"Exception occurred during {task.get_name()}", exc_info=exc)
```
berland, line 77: The reason this logging message could never be seen originally is because `ert.async_utils` is missing in `logger.conf`

src/ert/scheduler/scheduler.py
jonathan-eq: We have to somehow wait for them here too, in case any of them fail. Otherwise, we wont check if any of the background tasks raise exception until we exit the context.  That is probably the cause of the hanging. We are stuck here because the poll task raises exception, and we dont check for that until after the (now never-ending) jobs finish.

berland:
This PR will take down ERT nicely if an `1/0` is in  `_publisher` or `driver.poll()` but if it is in `_process_event_queue` it will hang.

jonathan-eq:
> This PR will take down ERT nicely if an `1/0` is in `_publisher` or `driver.poll()` but if it is in `_process_event_queue` it will hang.

Strange, I have a unit test for _process_event_queue, so it should behave the same.

src/ert/async_utils.py
```diff
@@ -26,22 +26,26 @@ async def background_tasks() -> AsyncGenerator[Any, Any]:
...
-                continue
-            if isinstance(exc, BaseException):
-                logger.error(str(exc), exc_info=exc)
-        tasks.clear()
+        try:
+            for exc in await gather_task:
+                if not isinstance(exc, BaseException):
+                    continue
+                raise exc
+        except asyncio.CancelledError:
```
berland, line 45: Thinking out loud: Do we risk stumbling on a CancelledError before we have gone through all tasks, and then miss an exception?

jonathan-eq: I assume that if we stumble upon an exception after cancellederror, it would be due to something not working because something else was cancelled.

jonathan-eq: No, you're right! What if a task fails, cancels the other tasks, but the cancellation of the other tasks finish before the exception is propagated? I guess this is how we have to do it, otherwise something could be hanging.

src/ert/async_utils.py
```diff
@@ -26,22 +26,26 @@ async def background_tasks() -> AsyncGenerator[Any, Any]:
...
         for t in tasks:
             t.cancel()
-        for exc in await asyncio.gather(*tasks, return_exceptions=True):
-            if isinstance(exc, asyncio.CancelledError):
-                continue
-            if isinstance(exc, BaseException):
-                logger.error(str(exc), exc_info=exc)
-        tasks.clear()
+        try:
+            for exc in await gather_task:
```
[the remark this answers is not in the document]
jonathan-eq: The idea behind it was to collect all the tasks, but I am not 100% sure if that is what it is doing.

berland:
Testing this now gives no crash when running ensemble_experiment on poly-case with a 1/0 in localdriver.poll(). It logs the traceback, but no hint is given to user that something is actually broken.

jonathan-eq:
> Testing this now gives no crash when running ensemble_experiment on poly-case with a 1/0 in localdriver.poll(). It logs the traceback, but no hint is given to user that something is actually broken.

I added another integration test for OpenPBS driver, and it seems to be passing. It will still fail gracefully, but the exception will be logged. Maybe we should print it too.

src/ert/async_utils.py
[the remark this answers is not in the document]
berland: I just don't understand why/if it makes a difference to declare it on line 30 and keeping the refernence in "gather_task", you are not using it later.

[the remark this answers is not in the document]
jonathan-eq: Yes, it does. The intention was that the first asyncio.gather (without awaiting it) starts the task, while I await for it to finish in the finally-block. However, I think that the main change in the commit is that I changed from using asyncio.Task to our _create_task function. I assume this was meant to be done implicitly, but I do not see how that was supposed to happen.

src/ert/scheduler/scheduler.py
```diff
@@ -181,32 +189,49 @@ async def execute(
         # We need to store the loop due to when calling
         # cancel jobs from another thread
         self._loop = asyncio.get_running_loop()
-        async with background_tasks() as cancel_when_execute_is_done:
-            cancel_when_execute_is_done(self._publisher())
-            cancel_when_execute_is_done(self._process_event_queue())
-            cancel_when_execute_is_done(self.driver.poll())
-            if min_required_realizations > 0:
-                cancel_when_execute_is_done(
+        ll_tasks = [
```
berland, line 192: is "ll" short for something? Better spell it out..

jonathan-eq: Yes, long-lived. I will change it make it more obvious.

src/ert/scheduler/scheduler.py
```diff
@@ -181,32 +189,49 @@ async def execute(
...
+                )
+            finally:
+                for ll_task in long_lived_tasks:
+                    ll_task.cancel()
+
+        job_results = []
+        try:
+            job_results = (await asyncio.gather(gather_jobs(), *long_lived_tasks))[0]
+        except asyncio.CancelledError:
+            for ll_task in long_lived_tasks:
```
xjules, line 225: Correct if I am wrong, but in case of cancellation `gather_jobs` will take care of cancelling the background tasks or?

src/ert/scheduler/scheduler.py
```diff
@@ -181,32 +189,49 @@ async def execute(
...
-            await self.driver.finish()
+            long_lived_tasks.append(asyncio.create_task(self._update_avg_job_runtime()))
+
+        start = asyncio.Event()
+        sem = asyncio.BoundedSemaphore(self._max_running or len(self._jobs))
+        for iens, job in self._jobs.items():
+            self._tasks[iens] = asyncio.create_task(job(start, sem, self._max_submit))
+        start.set()
+
+        async def gather_jobs() -> List[Union[Any, BaseException]]:
```
pinkwah, line 212: Seems silly to say union of `Any` and `BaseException`, as `BaseException` itself is part of `Any`. However, the callables for each task return `None`, so it shoud be:

```suggestion
        async def gather_jobs() -> List[Optional[BaseException]]:
```

Or, using modern notation:

```suggestion
        async def gather_jobs() -> list[BaseException | None]:
```

src/ert/scheduler/scheduler.py
```diff
@@ -181,32 +189,49 @@ async def execute(
...
+            try:
+                return await asyncio.gather(
+                    *self._tasks.values(), return_exceptions=True
+                )
+            finally:
+                for ll_task in long_lived_tasks:
+                    ll_task.cancel()
+
+        job_results = []
```
pinkwah, line 221: Needs type?

```suggestion
        job_results: list[BaseException | None] = []
```

tests/integration_tests/scheduler/test_openpbs_driver.py
```diff
@@ -107,3 +108,19 @@ def test_openpbs_driver_with_poly_example():
         "--enable-scheduler",
         "poly.ert",
     )
+
+
+@pytest.mark.timeout(180)
+@pytest.mark.integration_test
+@pytest.mark.usefixtures("copy_poly_case")
+def test_openpbs_driver_with_poly_example_submit_fails(monkeypatch, caplog):
```
pinkwah, line 116: What OpenPBS driver functionality is this testing?

src/ert/scheduler/scheduler.py
jonathan-eq: Yes, job.\__call__() returns None, so youre right. I will change it

tests/integration_tests/scheduler/test_openpbs_driver.py
jonathan-eq: It was supposed to test that an exception in OpenPBS submit would crash ert, and propagate the error message to the user. This should maybe be a scheduler test instead of a openpbs driver test.

berland:
It will hang for me if I run poly.ert towards Torque with mocked qsub and a `1/0` in `poll()` of openpbs-driver.

berland:
> It will hang for me if I run poly.ert towards Torque with mocked qsub and a `1/0` in `poll()` of openpbs-driver.

It hangs, but the exception is logged `logs/ee-log.*`

berland:
Have a look at the comment from codecov, it displays five added lines that are never triggered by the tests.

jonathan-eq:
Apparently, when we do 1/0 in the poll function of the jobqueue version, it also hangs so this is not a new problem.

src/clib/lib/job_queue/torque_driver.cpp
```diff
@@ -671,6 +671,7 @@ void *torque_driver_submit_job(void *_driver, std::string submit_cmd,
 static job_status_type
 torque_driver_get_qstat_status(torque_driver_type *driver,
                                const char *jobnr_char) {
+    1/0
```
pinkwah, line 674: This is not an exception but it might raise `SIGFPE` POSIX signal. Division by zero is undefined behaviour (UB), so "anything" can happen. In practice this is likely to be completely optimised out, so you don't get any error.

See: https://godbolt.org/z/n1zfxWEcd and notice how the compiler optimises this to `xor eax, eax; ret`, which just means `return 0` in assembly.

If you want to check that exceptions are caught, use eg. `throw py::runtime_error("Hello from C++");`

src/ert/scheduler/scheduler.py
```diff
@@ -181,32 +188,49 @@ async def execute(
...
-            await self.driver.finish()
+            long_lived_tasks.append(asyncio.create_task(self._update_avg_job_runtime()))
+
+        start = asyncio.Event()
+        sem = asyncio.BoundedSemaphore(self._max_running or len(self._jobs))
+        for iens, job in self._jobs.items():
+            self._tasks[iens] = asyncio.create_task(job(start, sem, self._max_submit))
+        start.set()
+
+        async def gather_jobs() -> List[BaseException | None]:
```
pinkwah, line 211: Nitpick: If we are using non-Python 3.8 type hinting features like `|`, we might as well use the "new" `list` subscript operator:

```suggestion
        async def gather_jobs() -> list[BaseException | None]:
```

tests/integration_tests/scheduler/test_lsf_driver.py
```diff
@@ -53,6 +53,7 @@ async def test_submit(tmp_path):
     assert (tmp_path / "test").read_text(encoding="utf-8") == "test\n"
 
 
+@pytest.mark.timeout(5)
```
pinkwah, line 56: We don't need timeout for integration tests. Consider that `bsub` on the real cluster might take a few minutes to actually work.

tests/integration_tests/scheduler/test_openpbs_driver.py
```diff
@@ -107,3 +108,39 @@ def test_openpbs_driver_with_poly_example():
...
     )
+
+
+@pytest.mark.timeout(180)
+@pytest.mark.integration_test
+@pytest.mark.usefixtures("copy_poly_case")
+def test_openpbs_driver_with_poly_example_failing_submit_fails_ert_and_propagates_exception_to_user(
+    monkeypatch, caplog
+):
+    monkeypatch.setattr(OpenPBSDriver, "submit", lambda *args, **kwargs: 1 / 0)
```
pinkwah, line 119: Consider moving this test to `unit_tests`. This test cannot use the actual OpenPBS cluster so this should instead use `MockDriver`.

Also, consider not using `1 / 0` but throwing an error like `raise RuntimeError("Testing")`

tests/integration_tests/scheduler/test_openpbs_driver.py
```diff
@@ -107,3 +108,39 @@ def test_openpbs_driver_with_poly_example():
...
+    assert "ZeroDivisionError" in caplog.text
+
+
+@pytest.mark.timeout(60)
+@pytest.mark.integration_test
+@pytest.mark.usefixtures("copy_poly_case")
+def test_openpbs_driver_with_poly_example_failing_poll_fails_ert_and_propagates_exception_to_user(
+    monkeypatch, caplog
+):
+    monkeypatch.setattr(OpenPBSDriver, "poll", lambda *args, **kwargs: 1 / 0)
```
pinkwah, line 137: Ditto... Don't actually submit jobs to the cluster if all we care about is that `poll` should fail.

xjules:
Superseded by https://github.com/equinor/ert/pull/7385

PR #7201, opened 2024-02-16
Add queue options for new OpenPBS Driver

### LINKED ISSUE #7101: Implement TORQUE options in OpenPBSDriver
The following QUEUE_OPTION TORQUEs:
1. #7222 
2. `NUM_NODES`, `NUM_CPUS_PER_NODE`
3. `KEEP_QSUB_OUTPUT`. First, check whether output of `job_dispatch.py` is logged anywhere. If it is, then don't implement this option.

jonathan-eq:
**Issue**
Resolves #7101


berland:
Why is not the full PR merge template in use here?

jonathan-eq:
I lost it, I will find it in a different PR.

tests/integration_tests/scheduler/test_openpbs_driver.py
```diff
@@ -96,6 +99,29 @@ async def finished(iens, returncode, aborted):
...
 
+@pytest.mark.timeout(10)
+async def test_keep_qsub_output_option(tmp_path, caplog):
+    driver = OpenPBSDriver(keep_qsub_output=True)
+    with caplog.at_level(logging.DEBUG):
+        await driver.submit(0, "sleep 2")
+    path_regex = re.compile(r"(/[^/\s]+)+")
+    job_id_regex = re.compile(r"(\w+\.\w+)")
+
+    paths_found = [
```
berland, line 110: I don't think asserting on log statements will suffice. What if someone changes the code but not the log message? Then the test will still pass, but the functionality will be broken.

src/ert/scheduler/openpbs_driver.py
```diff
@@ -43,10 +45,23 @@ class _Stat(BaseModel):
 class OpenPBSDriver(Driver):
     """Driver targetting OpenPBS (https://github.com/openpbs/openpbs) / PBS Pro"""
 
-    def __init__(self, *, queue_name: Optional[str] = None) -> None:
+    def __init__(
+        self,
+        *,
+        keep_qsub_output: Optional[bool] = False,
+        num_nodes: Optional[int] = None,
```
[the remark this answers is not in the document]
jonathan-eq: I split out the smallest changes to a separate PR: https://github.com/equinor/ert/pull/7205

tests/integration_tests/scheduler/test_openpbs_driver.py
jonathan-eq: I am not sure how to test this. Should I mock the random function, and assert that the expected path is created and that the file-content is what I expect?

src/ert/scheduler/openpbs_driver.py
```diff
@@ -61,13 +76,23 @@ async def submit(
     ) -> None:
 
         arg_queue_name = ["-q", self._queue_name] if self._queue_name else []
-
+        arg_memory_per_job = (
+            [f":mem={self._memory_per_job}"] if self._memory_per_job else []
+        )
+        arg_num_nodes = [f"nodes={self._num_nodes}"] if self._num_nodes else []
+        arg_num_cpus_per_node = (
+            [f":ppn={self._num_cpus_per_node}"] if self._num_cpus_per_node else []
```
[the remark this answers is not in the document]
jonathan-eq: Moved this to a different PR, and added explicit quotes around the strings

berland:
For testing, you can mock the behaviour of the real `qsub` in https://github.com/equinor/ert/blob/main/tests/integration_tests/scheduler/bin/qsub

tests/integration_tests/scheduler/test_openpbs_driver.py
berland: You want to assert the presence of some specifically named files (the job id is typically included in the filename). You can have a job sent to the cluster that echoes a magic string to stdout and another magic string to stderr, and then you assert this is present. You will need to figure out where the real qsub puts these files (I think it is eager to pollute `$HOME` 😢 ) and then get the mocked qsub to do the same.

PR #7202, opened 2024-02-16
Add tests for LSF driver kills

berland:
The driver will never retry killing but if it does not succeed the error messages are logged.

Killing twice will give a logged error if the first is successful.

**Issue**
Adds to  #7186 

**Approach**

## When applicable

### DIFF
--- a/src/ert/scheduler/lsf_driver.py
+++ b/src/ert/scheduler/lsf_driver.py
@@ -135,12 +135,12 @@ class LsfDriver(Driver):
 
     async def kill(self, iens: int) -> None:
         if iens not in self._iens2jobid:
-            logger.error(f"LSF driver does not know of realization {iens}")
+            logger.error(f"LSF kill failed due to missing jobid for realization {iens}")
             return
 
         job_id = self._iens2jobid[iens]
 
-        logger.info(f"Killing realization {iens} with LSF-id {job_id}")
+        logger.debug(f"Killing realization {iens} with LSF-id {job_id}")
         process = await asyncio.create_subprocess_exec(
             self._bkill_cmd,
             job_id,
@@ -166,8 +166,6 @@ class LsfDriver(Driver):
                 f"{stdout_decoded} {stderr.decode()}"
             )
             return
-
-        del self._iens2jobid[iens]
 
     async def poll(self) -> None:
         while True:
--- a/src/ert/scheduler/lsf_driver.py
+++ b/src/ert/scheduler/lsf_driver.py
@@ -134,14 +134,40 @@ class LsfDriver(Driver):
         self._iens2jobid[iens] = job_id
 
     async def kill(self, iens: int) -> None:
-        try:
-            job_id = self._iens2jobid[iens]
-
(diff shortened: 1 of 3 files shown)


src/ert/scheduler/lsf_driver.py
```diff
@@ -134,15 +134,41 @@ async def submit(
         self._iens2jobid[iens] = job_id
 
     async def kill(self, iens: int) -> None:
```
xjules, line 136: I see that now it would be handy to have a return value (bool) from submit and kill cmd.

src/ert/scheduler/lsf_driver.py
```diff
@@ -134,15 +134,41 @@ async def submit(
...
+        if iens not in self._iens2jobid:
+            logger.error(f"LSF driver does not know of realization {iens}")
```
xjules, line 138: Just to be uniform: `"LSF kill failed to missing realization {iens}"`

src/ert/scheduler/lsf_driver.py
```diff
@@ -134,15 +134,41 @@ async def submit(
...
+            return
+
+        job_id = self._iens2jobid[iens]
+
+        logger.info(f"Killing realization {iens} with LSF-id {job_id}")
```
xjules, line 143: logger.debug

src/ert/scheduler/lsf_driver.py
```diff
@@ -134,15 +134,41 @@ async def submit(
...
-            await proc.wait()
-        except KeyError:
+        if not re.match(f"Job <{job_id}> is being terminated", stdout_decoded):
+            logger.error(
+                "LSF kill failed with error message "
+                f"{stdout_decoded} {stderr.decode()}"
+            )
             return
 
+        del self._iens2jobid[iens]
```
xjules, line 170: Why is this solved differently in OpenPBSDriver?

berland: being unified

src/ert/scheduler/lsf_driver.py
```diff
@@ -134,13 +134,37 @@ async def submit(
...
+        if process.returncode:
+            logger.error(
+                f"LSF kill failed with returncode {process.returncode} "
+                f"and error message {stderr.decode()}"
+            )
+            return
         try:
-            job_id = self._iens2jobid[iens]
+            stdout_decoded = stdout.decode()
+        except UnicodeDecodeError:
```
xjules, line 159: I thought we agreed to let the python code crash by itself - maybe don't remember it correctly :)

berland: We want to crash hard when we have bugs in our code, but if bkill is suddenly misbehaving (giving binary output e.g.), we might not want to crash.

berland: This is replicating the behaviour in the lines below where we just log an error if we cannot find the correct string in bkill's output. By this `except` we are just treating erroneous binary output the same way as erroneous string output.

xjules (approved):
I think it looks good now! Just squash commits and :rocket:

PR #7213, opened 2024-02-19
Fix external_ert_script does not fail on error

### LINKED ISSUE #7212: Fix tests/integration_tests/test_cli.py::test_that_stop_on_fail_workflow_jobs_stop_ert
**Describe the bug**
The test checks that the ert workflow fails, but it passes for the incorrect reasons. The script ran is 
```
 
#!/bin/bash
ekho helo wordl
STOP_ON_FAIL=False
```
and the idea seems like it sees misspelled `echo` and fails on that, but it is actually failing on the empty line before the shebang.
```
RuntimeError: Workflow job failjob failed with error: [Errno 8] Exec format error: '/private/var/folders/7h/phfnnlw16x952zg3m4vgct_c0000gp/T/pytest-of-JONAK/pytest-394/poly_example5/test_data/failing_script_w_stop.sh'
    File "/Users/JONAK/Documents/FMU/SCOUT/ert/src/ert/config/ert_script.py", line 99, in initializeAndRun
      return self.run(*arguments)
    File "/Users/JONAK/Documents/FMU/SCOUT/ert/src/ert/config/external_ert_script.py", line 27, in run
      self.__job = Popen(command, stdout=PIPE, stderr=PIPE)
    File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/subprocess.py", line 971, in __init__
      self._execute_child(args, executable, preexec_fn, close_fds,
    File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/subprocess.py", line 1863, in _execute_child
      raise child_exception_type(errno_num, err_msg, err_filename)
  OSError: [Errno 8] Exec format error: '/private/var/folders/7h/phfnnlw16x952zg3m4vgct_c0000gp/T/pytest-of-JONAK/pytest-394/poly_example5/test_data/failing_script_w_stop.sh'
  ```
  Fixing this made 
**To reproduce**
1. `pip install ert`
2. `ert gui my_config.ert`
4. …

**Expected behaviour**

**Screenshots**

**Environment**
 - OS: [e.g. RHEL7]
 - ERT/Komodo release: [e.g. 2.18 or 2020.12.00]
 - Python version

**Additional context**

### PR DESCRIPTION
**Pull request by jonathan-eq**
**Issue**
Resolves #7212

**Approach**
The script supplied by the user is now copies over to a temporary file, and we add the `set -e` flag to the line after the shebang to ensure it will stop on failure.

## When applicable

### DIFF
--- a/docs/reference/workflows/configuring_jobs.rst
+++ b/docs/reference/workflows/configuring_jobs.rst
@@ -64,7 +64,19 @@ For example, if a job is defined as follows:
     EXECUTABLE script.sh
     STOP_ON_FAIL TRUE                   -- Tell the job to stop ert on failure
 
-STOP_ON_FAIL can also be specified within the internal (python) or external (executable) job script.
+Shell scripts (Bash) must, in addition to having `STOP_ON_FAIL TRUE` in ert config, set the "exit immediately" flag (`set -e`) for the error to propagate. Otherwise, even if errors occur within the script, it will still be "successful".
+
+::
+    set -e
+
+For example, this bash external job script will stop on failure
+
+::
+    #!/usr/bin/env bash
+    set -e
+    ekho 'Hello World' # misspelled command fails
+
+
 For example, this internal job script will stop on failure
 
 ::
@@ -76,15 +88,6 @@ For example, this internal job script will stop on failure
         def run(self):
             assert False, "failure"
     """
-
-As will external .sh executables if they contain the line STOP_ON_FAIL=TRUE:
-
-::
-
-    #!/bin/bash
-    STOP_ON_FAIL=True #
-    ekho helo wordl
-
 
 Configuring the arguments
 -------------------------
--- a/src/ert/config/external_ert_script.py
+++ b/src/ert/config/external_ert_script.py
@@ -1,11 +1,7 @@
 from __future__ import annotations
 
 import codecs
-import os
-import stat
 import sys
-import tempfile
-from pathlib import Path
 from subprocess import PIPE, Popen
 from typing import TYPE_CHECKING, Any, Optional
 
@@ -24,39 +20,22 @@ class ExternalErtScript(ErtScript):
         self.__job: Optional[Popen[bytes]] = None
 
     def run(self, *args: Any) -> None:
-        script_content = (
-            Path(self.__executable).read_text(encoding="utf-8").splitlines()
-        )
-        tmp_script_path: Optional[str] = None
-        if self.__executable[-2:] == "sh":
-            script_content.insert(1, "set -e")
-        try:
-            with tempfile.NamedTemporaryFile(
-                mode="w+", encoding="utf-8", delete=False
-            ) as tmp_file:
-                tmp_file.write("\n".join(script_content))
-                tmp_file.flush()
-                tmp_script_path = tmp_file.name
-            os.chmod(tmp_script_path, os.stat(tmp_script_path).st_mode | stat.S_IEXEC)
-            command = [tmp_script_path]
-            command.extend([str(arg) for arg in args])
+        command = [self.__executable]
+        command.extend([str(arg) for arg in args])
 
-            # we take care to terminate the process in cancel()
-            self.__job = Popen(command, stdout=PIPE, stderr=PIPE)
+        # we take care to terminate the process in cancel()
+        self.__job = Popen(command, stdout=PIPE, stderr=PIPE)
 
-            # The job will complete before stdout and stderr is returned
-            stdoutdata, stderrdata = self.__job.communicate()
+        # The job will complete before stdout and stderr is returned
+        stdoutdata, stderrdata = self.__job.communicate()
 
-            self._stdoutdata = codecs.decode(stdoutdata, "utf8", "replace")
-            self._stderrdata = codecs.decode(stderrdata, "utf8", "replace")
+        self._stdoutdata = codecs.decode(stdoutdata, "utf8", "replace")
+        self._stderrdata = codecs.decode(stderrdata, "utf8", "replace")
 
-            sys.stdout.write(self._stdoutdata)
+        sys.stdout.write(self._stdoutdata)
 
-            if self.__job.returncode != 0:
-                raise RuntimeError(self._stderrdata)
-        finally:
-            if tmp_script_path is not None:
-                os.remove(tmp_script_path)
+        if self.__job.returncode != 0:
+            raise RuntimeError(self._stderrdata)
 
     def cancel(self) -> Any:
         super().cancel()
--- a/src/ert/config/external_ert_script.py
+++ b/src/ert/config/external_ert_script.py
@@ -1,7 +1,11 @@
 from __future__ import annotations
 
 import codecs
+import os
+import stat
 import sys
+import tempfile
+from pathlib import Path
 from subprocess import PIPE, Popen
 from typing import TYPE_CHECKING, Any, Optional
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by jonathan-eq** [comment]
Ideally, we would run the script with `set -e`, but it is difficult to set it outside of the script and have it affect the commands inside.

**Comment by jonathan-eq** [comment]
This will mark a workflow as failed if it writes anything to stderr. Would something silly in a workflow (like a deprecation warning) cause the user's workflow to fail? Maybe @berland knows

**Comment by berland** [comment]
It would be a breaking change to introduce a hard error on non-empty stderr from workflow jobs (assuming this is accepted today).

File: src/ert/config/external_ert_script.py [hunk 1]
```diff
@@ -20,22 +24,39 @@ def __init__(self, ert: EnKFMain, storage: StorageAccessor, executable: str):
...
+            self._stderrdata = codecs.decode(stderrdata, "utf8", "replace")
 
-        sys.stdout.write(self._stdoutdata)
+            sys.stdout.write(self._stdoutdata)
 
-        if self.__job.returncode != 0:
-            raise RuntimeError(self._stderrdata)
+            if self.__job.returncode != 0:
+                raise RuntimeError(self._stderrdata)
+        finally:
```
berland, line 57: This cleanup is a bonus fix, or is it related to the fix of the bug?

tests/integration_tests/test_cli.py
```diff
@@ -393,7 +393,7 @@ def test_that_setenv_sets_environment_variables_in_jobs(setenv_config):
             "sh",
             dedent(
                 """\
-                    #!/bin/bash
+                   #!/bin/bash
```
berland, line 396: This is The Fix ?

jonathan-eq: That is why the test was broken. It failed on this instead of the expected failure..

src/ert/config/external_ert_script.py
jonathan-eq: It is part of the fix. I create a copy of the script file, add the `set -e`-option , run it, then delete it at the end.

berland: Got it. 

What is  self.__executable is a binary file (I do not know if that is disallowed already), could we break the functionality for that with this?

jonathan-eq: I check the file extension (or rather if the filename ends with sh - this could maybe cause problems), but maybe there is a better way. @xjules and I tried to set the flag in subprocess.Popen, but that was not possible.

src/ert/config/external_ert_script.py
```diff
@@ -20,22 +24,39 @@ def __init__(self, ert: EnKFMain, storage: StorageAccessor, executable: str):
         self.__job: Optional[Popen[bytes]] = None
 
     def run(self, *args: Any) -> None:
-        command = [self.__executable]
-        command.extend([str(arg) for arg in args])
+        script_content = (
+            Path(self.__executable).read_text(encoding="utf-8").splitlines()
+        )
+        tmp_script_path: Optional[str] = None
+        if self.__executable[-2:] == "sh":
```
berland, line 31: Is this a file extension like  "failing_script.sh" that you are testing, or is it /bin/bash or /bin/sh? Testing on file extension is not robust enough.

jonathan-eq: It is file extension. Should I check the shebang instead? It could be `/bin/env sh`, `/bin/env bash`, `/bin/bash`, `/bin/sh` or any other shell I guess. Maybe I can check that it is not python instead, but that could also be a lot of different shebangs.

jonathan-eq: What about checking `if "sh" in SHEBANG`?

berland: testing on the shebang is maybe sufficient yes. I would try to make a regexp  requiring either a space or a slash, then an optional "ba" or "c" or "tc", and then "sh" at the end.

pinkwah (commented):
I think we should remove the offending tests rather than try to hack a fix. Whoever wrote this doesn't know how shell scripts work (and that's fine we can't all be perfect).

docs/reference/workflows/configuring_jobs.rst
```diff
@@ -64,7 +64,19 @@ For example, if a job is defined as follows:
...
-STOP_ON_FAIL can also be specified within the internal (python) or external (executable) job script.
+In case of an external sh job script, you have to manually set the stop on fail option (-e)
+
+::
+    set -e
+
+For example, this bash external job script will stop on failure
+
+::
+    #/usr/bin/env bash
```
xjules, line 75: Hm, shouldn't it be: `#!/bin/bash` only?

jonathan-eq: It can be both, but I will change it anyways.

docs/reference/workflows/configuring_jobs.rst
```diff
@@ -64,7 +64,19 @@ For example, if a job is defined as follows:
...
+
+::
+    set -e
+
+For example, this bash external job script will stop on failure
+
+::
+    #!/bin/bash
+    set -e
+    ekho 'Hello World' # misspelled command fails
```
pinkwah, line 77: Explain that we still need `STOP_ON_FAIL TRUE` in ert config.

```suggestion
Shell scripts (Bash) must, in addition to having `STOP_ON_FAIL TRUE` in ert config, set the "exit immediately" flag (`set -e`) for the error to propagate. Otherwise, even if errors occur within the script, it will still be "successful".

::
    #!/usr/bin/env bash
    set -e
    ekho 'Hello World' # misspelled command fails
```

src/ert/config/external_ert_script.py
```diff
@@ -20,22 +24,39 @@ def __init__(self, ert: EnKFMain, storage: StorageAccessor, executable: str):
...
+        try:
+            with tempfile.NamedTemporaryFile(
+                mode="w+", encoding="utf-8", delete=False
+            ) as tmp_file:
+                tmp_file.write("\n".join(script_content))
+                tmp_file.flush()
+                tmp_script_path = tmp_file.name
+            os.chmod(tmp_script_path, os.stat(tmp_script_path).st_mode | stat.S_IEXEC)
+            command = [tmp_script_path]
+            command.extend([str(arg) for arg in args])
```
pinkwah, line 42: Remove this. We don't need to handle script in a special way.

```suggestion
```

If the intention is to execute the program even though it doesn't have the execute bit set (silly legacy), you can open the file in binary mode, read the first 2 characters and check if they are `b"#!"`. If they are, read the full line and prepend the contents to the command. Note that this is easy because the first two bytes _must_ be `#!` for this to be a valid script that the OS can execute. Ie, it can't be ` #!` or `\n#!` as you've discovered.

Eg, if the file is a Python script:
```py
#!/usr/bin/env python
# code here
```

You can do this by:
```py
import shlex

def get_script_command(path: Path) -> Optional[List[str]]:
    with open(path, "rb") as f:
        if f.read(2) != b"#!":
            return None  # not a script file
        return shlex.split(f.readline().decode())
```

tests/integration_tests/test_cli.py
```diff
@@ -298,164 +298,150 @@ def test_that_setenv_sets_environment_variables_in_jobs(setenv_config):
...
+                        def run(self):
+                            assert False, "failure"
+
                 """
             ),
-            "failing_ert_script.py",
-            """
-from ert import ErtScript
-class AScript(ErtScript):
-    stop_on_fail = True
```
pinkwah, line 439: Note that we do want to keep this `stop_on_fail = True` within `ErtScript`s functionality.

pinkwah (changes requested):

PR #7218, opened 2024-02-20
Set default pytest timeout of 240s

### LINKED ISSUE #7217: Add default timeout for tests
**Describe the requested improvement**
It would be great if pytest would time out instead of a github action runner timing out. This way, we would know what test is hanging. It would be fairly easy to add something along the lines of

```
if item.get_marker('timeout') is None:
            item.add_marker(pytest.mark.timeout(60))
```
in conftest.

**Expected behaviour**
We should know what test is hanging.

jonathan-eq:
**Issue**
Resolves #7217 

## When applicable

### DIFF
--- a/.github/workflows/test_ert.yml
+++ b/.github/workflows/test_ert.yml
@@ -11,7 +11,6 @@ jobs:
 
   tests-ert:
     name: Run ert tests
-    timeout-minutes: 30
     runs-on: ${{ inputs.os }}
 
     steps:
@@ -23,6 +22,7 @@ jobs:
 
     - uses: actions/setup-python@v4
       id: setup_python
+      timeout-minutes: 5
       with:
         python-version: ${{ inputs.python-version }}
         cache: "pip"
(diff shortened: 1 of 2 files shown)


berland:
I would claim this is not setting a "default" timeout since any test not having the marker will not time out?

Since this requires an explicit marker on every test, what about just having a custom timeout for each test that could time out (then I am thinking about many tests that I know could have only say 5 seconds timeout)?

jonathan-eq:
> I would claim this is not setting a "default" timeout since any test not having the marker will not time out?
> 
> Since this requires an explicit marker on every test, what about just having a custom timeout for each test that could time out (then I am thinking about many tests that I know could have only say 5 seconds timeout)?

I think this would add the marker to any test not already having the marker.
I will try setting it in pyproject instead, just to make sure.

The idea was to set an upper limit for all tests, and make sure pytest will timeout the test before github actions times out pytest. This way we will know what test is hanging. The timeout period can be anything less than the workflow timeout, but we should be sure the potentially hanging test is actually hanging (so maybe 5-7 minutes?)

.github/workflows/test_ert.yml
```diff
@@ -23,6 +22,7 @@ jobs:
 
     - uses: actions/setup-python@v4
       id: setup_python
+      timeout-minutes: 30
```
berland, line 25: At this level, we can go much lower than 30 minutes :)

berland (approved):
Lets go!

PR #7229, opened 2024-02-21
Log when max_runtime kicks in

### LINKED ISSUE #7226: No logging when jobs are killed by MAX_RUNTIME
**Describe the bug**
There is noe mention in the logs when jobs are killed by MAX_RUNTIME. 

**To reproduce**
1. `pip install ert`
2. Add `MAX_RUNTIME` 20 in the poly config
3. `ert gui poly.ert`
4. Add sleep > 20 in `poly_eval.py`
5. Run experiment (Smoother)
6. Observe no mention of killed jobs

**Expected behaviour**
Logging when jobs are killed due to MAX_RUNTIME.

**Screenshots**
N/A

**Environment**
 - OS: macos
 - ERT/Komodo release: ert-8.4.4.dev44+g1e6fea5bb.d20240221
 - Python version: 3.10

**Additional context**
N/A

berland:
Redoing for the scheduler what job_queue already does. The log message is slightly altered compared to jobqueue.

**Issue**
Resolves #7226 

**Approach**
🗒️ 

See https://github.com/equinor/ert/blob/main/src/ert/job_queue/job_queue_node.py#L230C3-L243C14 for old logging code.

## When applicable

### DIFF
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -176,7 +176,9 @@ class Job:
         )
         assert self._scheduler._events is not None
         await self._scheduler._events.put(to_json(timeout_event))
-
+        logger.error(
+            f"Realization {self.iens} stopped due to MAX_RUNTIME={self.real.max_runtime} seconds"
+        )
         self.returncode.cancel()  # Triggers CancelledError
 
     async def _handle_failure(self) -> None:
--- a/src/ert/job_queue/job_queue_node.py
+++ b/src/ert/job_queue/job_queue_node.py
@@ -233,8 +233,8 @@ class JobQueueNode(BaseCClass):  # type: ignore
         # check before we log.
         if self._tried_killing == 1:
             logger.error(
-                f"MAX_RUNTIME reached in run path {self.run_path}. Runtime: "
-                f"{self.runtime} (max runtime: {self._max_runtime})"
+                f"Realization {self.run_arg.iens} stopped due to "
+                f"MAX_RUNTIME={self._max_runtime} seconds. "
             )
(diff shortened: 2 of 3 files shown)


src/ert/scheduler/job.py
```diff
@@ -176,7 +176,9 @@ async def _max_runtime_task(self) -> None:
         )
         assert self._scheduler._events is not None
         await self._scheduler._events.put(to_json(timeout_event))
-
+        logger.error(
+            f"Realization {self.iens} stopped due to MAX_RUNTIME={self.real.max_runtime} seconds"
```
jonathan-eq, line 180: Shouldn't this be the same message as you would get when using job queue? I think it would be great to have a seamless transition between jobqueue and scheduler, including the log messages. Maybe we should update the JobQueue to match this one.

jonathan-eq (changes requested):
It is weird there were no tests failing for the missing log statement when we ran with the using_queue_and_scheduler fixture

src/ert/scheduler/job.py
berland: Yes, I found the job_queue message a little unclear. Can update it.

berland:
> It is weird there were no tests failing for the missing log statement when we ran with the using_queue_and_scheduler fixture

There were no tests for that particular log message from job_queue. There is only a test for the tracker/snapshot.

src/ert/scheduler/job.py
jonathan-eq: Yes, I agree

jonathan-eq (approved):
LGTM!

PR #7230, opened 2024-02-21
Add missing space in ConfigWarning msg

berland:
**Issue**
Resolves missing space in log message:

`...forward model to fail.To silence this...`


jonathan-eq (approved):
Good fix!

PR #7238, opened 2024-02-22
Support NUM_NODES and NUM_CPUS_PER_NODE in OpenPBS driver

### LINKED ISSUE #7232: Add support for NUM_NODES and NUM_CPUS_PER_NODE in OpenPBS driver
Port behaviour from torque_driver.cpp

berland:
**Issue**
Resolves #7232 

**Approach**
Just do it.

## When applicable

### DIFF
--- a/tests/unit_tests/scheduler/test_openpbs_driver.py
+++ b/tests/unit_tests/scheduler/test_openpbs_driver.py
@@ -1,7 +1,8 @@
 import os
+import shlex
 import stat
 from pathlib import Path
-from typing import List
+from typing import Dict, List
 
 import pytest
 from hypothesis import given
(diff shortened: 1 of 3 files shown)


tests/unit_tests/scheduler/test_openpbs_driver.py
```diff
@@ -40,3 +60,66 @@ async def test_no_validation_of_memory_per_job():
...
+    await driver.submit(0, "sleep")
+    resources = parse_resource_string(
+        Path("captured_qsub_args").read_text(encoding="utf-8")
+    )
+    assert "ppn" not in resources
+
+
+@given(words, st.integers(min_value=0), st.integers(min_value=0))
+@pytest.mark.usefixtures("capturing_qsub")
+async def test_full_resource_string(memory_per_job, num_nodes, num_cpus_per_node):
```
xjules, line 108: Don't fully understand the meaning of this test. I guess we want to test that specifying all the options are present in the resulting strings. Do we want to make sure that empty or 0 values are ok? Isn't this the test that should be caught by the config validation?

berland: I want to ensure that whatever the resource builder in the driver is doing, we can assert that the result is correct and parsable as colonseparated list of key=value pairs.

The "0" part of the test is confusing, thus the comment. I am just using it to be able to send `None` arguments to the driver.

pinkwah (approved):
LGTM

Issue #7235, opened 2024-02-22
Support or deprecate CLUSTER_LABEL and JOB_PREFIX in openpbs-driver


berland:
There are two string options that are quite trivial to add to the OpenPBS driver that the current torque_driver.cpp supports:

* CLUSTER_LABEL (goes into the resource string)
* JOB_PREFIX (prepended to the job name when submitting)

These are not found in the logs for the last three months, but 3 months might not be representative. 

Either implement support, or deprecate them.

xjules:
As both options were ported to openpbs closing this issue.

PR #7243, opened 2024-02-22
Support (in new PBS driver) and deprecate JOB_PREFIX

berland:
JOB_PREFIX is not mentioned in the docs and no usage is known in the logs. It should also be redundant.

**Issue**
Solve 50% of #7235 

**Approach**
Port behaviour from C-driver (torque_driver.cpp) to OpenPBS driver but also deprecate it for both drivers (C-version and Python version) in the config parser.

## When applicable

### DIFF
--- a/src/ert/config/parsing/config_schema_deprecations.py
+++ b/src/ert/config/parsing/config_schema_deprecations.py
@@ -44,6 +44,14 @@ deprecated_keywords_list = [
         for kw in ["DEFINE", "DATA_KW"]
     ],
     DeprecationInfo(
+        keyword="QUEUE_OPTION",
+        message=(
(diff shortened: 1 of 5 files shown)


berland:
Yeah, but it was so simple to implement support for it so we can just do it and remove it entirely in say 6 months.

xjules:
> Yeah, but it was so simple to implement support for it so we can just do it and remove it entirely in say 6 months.

That's when the new lsf_driver hits the users :D

berland:
This PR only applies to PBS

xjules:
> This PR only applies to PBS

Yes, I meant PBS - but PBS can hit them next month indeed.

PR #7253, opened 2024-02-23
Replace gstools with scipy

dafeda:
To get a more realistic test, we want to run
localization on more parameter.
gstools is very slow so generating many parameters would take a lot of time.
Instead, we generate a gaussian field using scipy which is less flexible but much faster.

## When applicable

### DIFF
--- a/tests/unit_tests/analysis/test_es_update.py
+++ b/tests/unit_tests/analysis/test_es_update.py
@@ -3,13 +3,13 @@ import re
 from functools import partial
 from pathlib import Path
 
-import gstools as gs
 import numpy as np
 import pytest
 import scipy as sp
 import xarray as xr
 import xtgeo
 from iterative_ensemble_smoother import steplength_exponential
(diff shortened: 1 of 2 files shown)


tests/unit_tests/analysis/test_es_update.py
```diff
@@ -509,16 +509,23 @@ def g(X):
...
-    model = gs.Exponential(dim=2, var=2, len_scale=8)
-    fields = []
-    seed = gs.random.MasterRNG(20170519)
-    for _ in range(num_ensemble):
-        srf = gs.SRF(model, seed=seed())
-        field = srf.structured([np.arange(num_grid_cells), np.arange(num_grid_cells)])
-        field = field.astype(np.float32)
-        fields.append(field)
-    X = np.vstack([field.flatten() for field in fields]).T
+    # Initialize an array to hold all realizations
```
dafeda, line 512:
```suggestion

```

PR #7277, opened 2024-02-26
Remove Codecov integration

pinkwah:
Our coverage is fairly good nowadays, and considering the failure rate of codecov.io (500 HTTP errors) we can get rid of it.


eivindjahren (approved):
:+1:

pinkwah:
We have agreed to try codecov for a bit, so I am shelving this PR. But, Codecov, you have been warned *shakes fist*.

PR #7284, opened 2024-02-27
Forward exceptions from threads to main thread

pinkwah:
Attempt 2


xjules (approved):
Looks like the performance is sharp again! Let's rock

PR #7293, opened 2024-02-28
Log scheduler/job_queue orchestrator when created

xjules:
**Issue**
Make sure that the info on orchestrator is logged correctly~~and as a debug.~~

**Approach**
Inject logging when creating JobQueue / Scheduler.

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/_builder/_legacy.py
+++ b/src/ert/ensemble_evaluator/_builder/_legacy.py
@@ -204,6 +204,7 @@ class LegacyEnsemble(Ensemble):
                     ee_cert=self._config.cert,
                     ee_token=self._config.token,
                 )
+                scheduler_logger.info("Experiment ran on ORCHESTRATOR: scheduler")
             else:
                 queue = JobQueue(
                     self._queue_config,
@@ -214,6 +215,7 @@ class LegacyEnsemble(Ensemble):
                     ee_token=self._config.token,
                     on_timeout=on_timeout,
                 )
+                scheduler_logger.info("Experiment ran on ORCHESTRATOR: job_queue")
             self._job_queue = queue
 
             await cloudevent_unary_send(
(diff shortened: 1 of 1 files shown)


berland:
This changes the logged string making it harder to search for occurences before and after this commit, but that is maybe ok?

xjules:
> This changes the logged string making it harder to search for occurences before and after this commit, but that is maybe ok?

I'll change `JobQueue` to `job_queue` or you meant something else?

berland:
> I'll change `JobQueue` to `job_queue` 
:+1:

src/ert/ensemble_evaluator/_builder/_legacy.py
```diff
@@ -234,17 +236,14 @@ async def _evaluate_inner(  # pylint: disable=too-many-branches
             )
             result = identifiers.EVTYPE_ENSEMBLE_FAILED
 
-        orchestrator = "scheduler"
```
berland, line 237: `Scheduler` vs `scheduler`

src/ert/ensemble_evaluator/_builder/_legacy.py
```diff
@@ -214,6 +215,7 @@ async def _evaluate_inner(  # pylint: disable=too-many-branches
                     ee_token=self._config.token,
                     on_timeout=on_timeout,
                 )
+                scheduler_logger.debug("Experiment ran on ORCHESTRATOR: job_queue")
```
pinkwah, line 218: I concur with Håvard, and I also think this should be JobQueue as that's what we've been calling it.

pinkwah:
I think it should be `INFO` btw. This is information that is useful to us logged to the azure server and not debug information. There should only be 1 line per experiment so it's fine.

src/ert/ensemble_evaluator/_builder/_legacy.py
xjules: Then we need to change to big-poly workflow, which is fine I guess. Will do.

berland: Just keep the strings as they were prior to this PR. Not worth the effort to change it.

PR #7299, opened 2024-02-29
Let LSF driver fall back to bhist for not-found jobs in bjobs

### LINKED ISSUE #7208: Support bhist when bjobs fails in LSF
This entry point in the C driver needs porting to the Python version:
https://github.com/equinor/ert/blob/9496957720b5cfc7d2924f380d8d92c7d89cb0de/src/clib/lib/job_queue/lsf_driver.cpp#L713C1-L727C14

berland:
**Issue**
Resolves #7208 

**Approach**
Replicate the functionality of the C-driver, but with a different approach. We rely that the polling will 
happen repeatedly and will only provide JobStates from bhist polling when bhist has been polled earlier.

## When applicable

### DIFF
--- a/src/ert/scheduler/lsf_driver.py
+++ b/src/ert/scheduler/lsf_driver.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import itertools
 import json
 import logging
 import re
@@ -9,9 +10,11 @@ import shutil
 import stat
 import subprocess
 import tempfile
+import time
 from pathlib import Path
 from typing import (
     Dict,
+    Iterable,
     List,
     Literal,
     Mapping,
@@ -129,6 +132,36 @@ def parse_resource_requirement_string(
     return " ".join(resource_requirements)
 
 
+def parse_bhist(bhist_output: str) -> Dict[str, Dict[str, int]]:
+    data: Dict[str, Dict[str, int]] = {}
+    for line in bhist_output.splitlines():
+        if line.startswith("Summary of time"):
+            assert "in seconds" in line
+        if not line or not line[0].isdigit():
+            continue
+        tokens = line.split()
+        try:
+            # The bhist output has data in 10 columns in fixed positions,
+            # with spaces possible in field 3. Since `split()` is used
+            # to parse the output, we branch on the number of tokens found.
+            if len(tokens) > 10:
+                data[tokens[0]] = {
+                    "pending_seconds": int(tokens[-7]),
+                    "running_seconds": int(tokens[-5]),
+                }
+            elif len(tokens) >= 6 and tokens[0] and tokens[3] and tokens[5]:
+                data[tokens[0]] = {
+                    "pending_seconds": int(tokens[3]),
+                    "running_seconds": int(tokens[5]),
+                }
+            else:
+                logger.warning(f'bhist parser could not parse "{line}"')
+        except ValueError as err:
+            logger.warning(f'bhist parser could not parse "{line}", "{err}"')
+            continue
+    return data
+
+
 class LsfDriver(Driver):
     def __init__(
         self,
@@ -138,6 +171,7 @@ class LsfDriver(Driver):
         bsub_cmd: Optional[str] = None,
         bjobs_cmd: Optional[str] = None,
         bkill_cmd: Optional[str] = None,
+        bhist_cmd: Optional[str] = None,
     ) -> None:
         super().__init__()
 
@@ -157,6 +191,11 @@ class LsfDriver(Driver):
         self._bsub_retries = 10
 
         self._poll_period = _POLL_PERIOD
+
+        self._bhist_cmd = Path(bhist_cmd or shutil.which("bhist") or "bhist")
+        self._bhist_cache: Optional[Dict[str, Dict[str, int]]] = None
+        self._bhist_required_cache_age: float = 4
+        self._bhist_cache_timestamp: float = time.time()
 
     async def submit(
         self,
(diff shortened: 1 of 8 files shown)


src/ert/scheduler/lsf_driver.py
```diff
@@ -33,6 +34,7 @@
 
 logger = logging.getLogger(__name__)
 
+logger.setLevel(logging.DEBUG)
```
berland, line 37: remove..

src/ert/scheduler/lsf_driver.py
```diff
@@ -236,15 +273,26 @@ async def poll(self) -> None:
...
-            stat = _Stat(**parse_bjobs(stdout.decode(errors="ignore")))
-            for job_id, job in stat.jobs.items():
-                await self._process_job_update(job_id, job)
-            missing_in_bjobs_output = set(self._jobs) - set(stat.jobs.keys())
+            bjobs_states = _Stat(**parse_bjobs(stdout.decode(errors="ignore")))
+            missing_in_bjobs_output = set(self._jobs) - set(bjobs_states.jobs.keys())
             if missing_in_bjobs_output:
-                logger.warning(
-                    f"bjobs did not give status for job_ids {missing_in_bjobs_output}"
+                logger.debug(f"bhist is needed for job ids: {missing_in_bjobs_output}")
```
xjules, line 279: "bhist is used for job ids"

src/ert/scheduler/lsf_driver.py
```diff
@@ -236,15 +273,26 @@ async def poll(self) -> None:
...
+
+            bhist_states = await self._poll_once_by_bhist(missing_in_bjobs_output)
```
xjules, line 281: hm, can the if statement be moved out from the function?:
```python
if not missing_job_ids:
```
Since the function title is misleading then.

src/ert/scheduler/lsf_driver.py
```diff
@@ -283,5 +331,56 @@ async def _process_job_update(self, job_id: str, new_state: AnyJob) -> None:
                 del self._iens2jobid[iens]
             await self.event_queue.put(event)
 
+    async def _poll_once_by_bhist(self, missing_job_ids: Iterable[str]) -> _Stat:
```
xjules, line 334: When I look at the old `lsf_driver.cpp` there we poll immediately two times in a row, I guess here it will be tackled by the `def poll(self)` function. `self._bhist_cache_timestamp` is used to determine whether the bhist_cache is not too old, but it is set only in the constructor. Shouldn't it be set everytime we call it?

src/ert/scheduler/lsf_driver.py
```diff
@@ -98,13 +101,42 @@ def parse_bjobs(bjobs_output: str) -> Dict[str, Dict[str, Dict[str, str]]]:
...
+def parse_bhist(bhist_output: str) -> Dict[str, Dict[str, int]]:
+    data: Dict[str, Dict[str, int]] = {}
+    for line in bhist_output.splitlines():
+        if line.startswith("Summary of time"):
+            assert "in seconds" in line
+        if not line or not line[0].isdigit():
+            continue
+        tokens = line.split()
+        try:
+            if len(tokens) > 10:
```
jonathan-eq, line 113: Can we have an example of tokens where it is larger than 10 here? Maybe as a comment above, or would that look off?
Something like that will help readability, and explain the magic numbers `-7`, `-5` below.

tests/integration_tests/scheduler/test_lsf_driver.py
```diff
@@ -100,3 +118,12 @@ async def finished(iens, returncode):
...
     await poll(driver, {0}, finished=finished)
+
+
+async def test_polling_bhist_fallback(not_found_bjobs):
+    driver = LsfDriver()
+    Path("mock_jobs").mkdir()
+    Path("mock_jobs/pendingtimemillis").write_text("100")
+    driver._poll_period = 0.01
+    await driver.submit(0, "sh", "-c", "sleep 1")
+    await poll(driver, {0})
```
xjules, line 129: Given my comment [above](https://github.com/equinor/ert/pull/7299/files#r1540741183) we would be able to assert that `_poll_once_by_bhist` was called once or ?

src/ert/scheduler/lsf_driver.py
```diff
@@ -98,13 +101,42 @@ def parse_bjobs(bjobs_output: str) -> Dict[str, Dict[str, Dict[str, str]]]:
...
+
+
 class LsfDriver(Driver):
     def __init__(
         self,
         queue_name: Optional[str] = None,
         bsub_cmd: Optional[str] = None,
         bjobs_cmd: Optional[str] = None,
         bkill_cmd: Optional[str] = None,
+        bhist_cmd: Optional[str] = None,
```
jonathan-eq, line 139: Is this driver option propagated correctly from ert config through scheduler's create_driver function?
There is a unit test for this `tests/unit_tests/scheduler/test_scheduler.py::test_scheduler_create_lsf_driver`.

src/ert/scheduler/lsf_driver.py
```diff
@@ -98,13 +101,42 @@ def parse_bjobs(bjobs_output: str) -> Dict[str, Dict[str, Dict[str, str]]]:
...
+            continue
+        tokens = line.split()
+        try:
+            if len(tokens) > 10:
+                # (there are spaces in the job name, field 3)
+                data[tokens[0]] = {
+                    "pending_seconds": int(tokens[-7]),
+                    "running_seconds": int(tokens[-5]),
+                }
+            elif len(tokens) >= 6 and tokens[0] and tokens[3] and tokens[5]:
```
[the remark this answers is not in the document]
jonathan-eq: (the order of this comment and the one below should be flipped)

xjules (changes requested):
Looks good @berland ! Just had a small comment about the way you call `_poll_once_by_bhist`.

src/ert/scheduler/lsf_driver.py
```diff
@@ -283,5 +331,56 @@ async def _process_job_update(self, job_id: str, new_state: AnyJob) -> None:
...
+            return _Stat(**{"jobs": {}})
+
+        process = await asyncio.create_subprocess_exec(
+            self._bhist_cmd,
+            *[str(job_id) for job_id in missing_job_ids],
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+        )
+        stdout, stderr = await process.communicate()
+        if process.returncode:
```
jonathan-eq, line 348: Is this correct?
`process.returncode` is None only when the process has not terminated yet. 
The code in this block will be executed also on success, not only on failure.

src/ert/scheduler/lsf_driver.py
berland: Added a commit now that tries this, it might look slightly better.

src/ert/scheduler/lsf_driver.py
berland: ouch, good catch, that would be catastrophic..

berland: Fixed in dfc1bd2

src/ert/scheduler/lsf_driver.py
berland: There is one already, renaming it to make it clearer.

berland: See 20c328f

tests/integration_tests/scheduler/test_lsf_driver.py
berland: Hmm, I think I want to keep that as an implementation detail, this test should not care as long as `await poll` is working fine.

src/ert/scheduler/lsf_driver.py
berland: Since we await for `process.communicate()` we will not get here until it is finished, and then we are testing for non-zero-ness for the `int`.

src/ert/scheduler/lsf_driver.py
jonathan-eq: Perfect!

xjules (approved):
Looks very good now @berland ! :rocket:

PR #7302, opened 2024-02-29
Add support for KEEP_QSUB_OUTPUT for OpenPBSDriver

### LINKED ISSUE #7233: Add support for KEEP_QSUB_OUTPUT in OpenPBS driver
First, check whether output of job_dispatch.py is logged anywhere. If it is, then don't implement this option.

berland:
This is implemented differently compared to the legacy driver, which injects the '-k' option to 'qsub'. Controlling using ~~-j oe and~~ -o and -e seems more stable

**Issue**
Resolves #7233 

**Approach**

## When applicable

### DIFF
--- a/tests/unit_tests/scheduler/test_openpbs_driver.py
+++ b/tests/unit_tests/scheduler/test_openpbs_driver.py
@@ -327,3 +327,33 @@ async def test_that_qdel_will_retry_and_succeed(
     assert "TRIED" in Path(bin_path / "script_try").read_text(encoding="utf-8")
     assert "qdel executed" in Path(bin_path / "qdel_output").read_text(encoding="utf-8")
     assert error_msg in Path(bin_path / "qdel_error").read_text(encoding="utf-8")
+
+
+@pytest.mark.usefixtures("capturing_qsub")
+@pytest.mark.parametrize(
+    "queue_config_string_for_keep_qsub, expectedkeep",
+    [
+        ("", False),  # Driver default
+        # Testing strings accepted by queue_config.py:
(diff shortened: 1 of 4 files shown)


tests/unit_tests/scheduler/test_openpbs_driver.py
```diff
@@ -327,3 +327,33 @@ async def test_that_qdel_will_retry_and_succeed(
     assert "TRIED" in Path(bin_path / "script_try").read_text(encoding="utf-8")
     assert "qdel executed" in Path(bin_path / "qdel_output").read_text(encoding="utf-8")
     assert error_msg in Path(bin_path / "qdel_error").read_text(encoding="utf-8")
+
+
+@pytest.mark.usefixtures("capturing_qsub")
+@pytest.mark.parametrize(
+    "queue_config_string_for_keep_qsub, expectedkeep",
+    [
+        ("", False),  # Driver default
```
xjules, line 336: I like that the test, but essentially all strings that are not in `["1", "True", "TRUE", "T"]` will be False, which sort of makes all the False cases insignificant.

xjules: Suggestions: either make explicit that `["1", "True", "TRUE", "T"]` and `["0", "False", "FALSE", "F"]` is false, or just make one case that is false?

berland: Any other False scenario will be caught by the config validation.

xjules (approved):
Approved! :rocket:

Issue #7316, opened 2024-02-29
Flaky test_submit 


sondreso:
**Describe the bug**

On the third retry, test_submit failed (See this issue for previous runs: https://github.com/equinor/ert/issues/7310)

```
integration_tests/scheduler/test_lsf_driver.py::test_submit 
[gw2] FAILED integration_tests/scheduler/test_lsf_driver.py::test_submit 
```

```
=================================== FAILURES ===================================
_________________________________ test_submit __________________________________
[gw2] linux -- Python 3.8.18 /tmp/f_scout_ci/actions-runner-01/_temp/test-kenv/root/bin/python

item = <Coroutine test_submit>

    @pytest.hookimpl(wrapper=True, tryfirst=True)
    def pytest_runtest_call(item):
>       result = yield

../../test-kenv/root/lib64/python3.8/site-packages/pytestqt/plugin.py:189: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../test-kenv/root/lib64/python3.8/site-packages/pytest_asyncio/plugin.py:440: in runtest
    super().runtest()
../../test-kenv/root/lib64/python3.8/site-packages/pytest_asyncio/plugin.py:907: in inner
    _loop.run_until_complete(task)
/opt/rh/rh-python38/root/usr/lib64/python3.8/asyncio/base_events.py:603: in run_until_complete
    self.run_forever()
/opt/rh/rh-python38/root/usr/lib64/python3.8/asyncio/base_events.py:570: in run_forever
    self._run_once()
/opt/rh/rh-python38/root/usr/lib64/python3.8/asyncio/base_events.py:1823: in _run_once
    event_list = self._selector.select(timeout)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <selectors.EpollSelector object at 0x7fe8306e9ee0>, timeout = -1

    def select(self, timeout=None):
        if timeout is None:
            timeout = -1
        elif timeout <= 0:
            timeout = 0
        else:
            # epoll_wait() has a resolution of 1 millisecond, round away
            # from zero to wait *at least* timeout seconds.
            timeout = math.ceil(timeout * 1e3) * 1e-3
    
[...]

PR #7318, opened 2024-02-29
Combine OpenPBS & LSF integration tests

pinkwah:
Resolves: #7316 (hopefully)

There is a lot of overlap between the two drivers, so we combine some of the generic integration tests into a single parameterised pytest.

### DIFF
--- a/tests/integration_tests/scheduler/conftest.py
+++ b/tests/integration_tests/scheduler/conftest.py
@@ -1,0 +1,13 @@
+from __future__ import annotations
+
+import os
+import sys
+from pathlib import Path
+
+
+def mock_bin(monkeypatch, tmp_path):
+    bin_path = Path(__file__).parent / "bin"
+
+    monkeypatch.setenv("PATH", f"{bin_path}:{os.environ['PATH']}")
+    monkeypatch.setenv("PYTEST_TMP_PATH", str(tmp_path))
+    monkeypatch.setenv("PYTHON", sys.executable)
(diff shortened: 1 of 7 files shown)


tests/integration_tests/scheduler/conftest.py
```diff
@@ -0,0 +1,56 @@
...
+@pytest.fixture
+def mock_lsf(pytestconfig, monkeypatch, tmp_path):
+    if pytestconfig.getoption("lsf"):
+        # User provided --lsf, which means we should use the actual LSF
+        # cluster without mocking anything.""
+        return
+    mock_bin(monkeypatch, tmp_path)
+
+
+async def poll(driver: Driver, expected: set[int], *, started=None, finished=None):
```
berland, line 32: There is a version of this one in unit_tests/test_lsf_driver.py as well. Maybe that one can also use this common function.

pinkwah: Are you sure? I don't see anything there.

berland: It is well hidden far down... 
https://github.com/equinor/ert/blob/f26a69942191141f1d29998f2fb4010655096502/tests/unit_tests/scheduler/test_lsf_driver.py#L268

pinkwah: Oh, my branch was outdated. Will move `poll` to `tests.utils`.

berland: Maybe also add a docstring explaining its intended usage.

berland (approved):
💯

PR #7329, opened 2024-03-01
Fix ert ci ctests for rhel8

larsevj:

## When applicable

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -22,7 +22,7 @@ jobs:
           submodules: true
           lfs: true
 
-      - uses: actions/setup-python@v4
+      - uses: actions/setup-python@v5
         id: setup_python
         with:
           python-version: "3.10"
--- a/.github/workflows/build-wheels-macos.yml
+++ b/.github/workflows/build-wheels-macos.yml
@@ -19,7 +19,7 @@ jobs:
       with:
         fetch-depth: 0
 
-    - uses: actions/setup-python@v4
+    - uses: actions/setup-python@v5
       id: setup_python
       with:
         python-version: ${{ inputs.python-version }}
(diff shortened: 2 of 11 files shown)


ci/run_ert_ctests.sh
```diff
@@ -23,12 +23,14 @@ set -x
...
 	build_ert_clib
-	build_ert_dev
 	run_ctest
 }
 
 enable_environment () {
-	source /opt/rh/devtoolset-8/enable
+	if [[ -f /opt/rh/devtoolset-8/enable ]] ; then
+		source /opt/rh/devtoolset-8/enable
+	fi
```
andreas-el, line 32: Can and should this have been bumped to `devtoolset-9` ? This has been the practice the last years.

ci/run_ert_ctests.sh
```diff
@@ -58,11 +66,6 @@ build_ert_clib () {
 	popd
 }
 
-build_ert_dev () {
-	enable_environment
-	pip install "${ERT_SOURCE_ROOT}[dev]"
```
andreas-el, line 63: I don't immediately understand why this was removed.
Is it because `build_ert_clib` uses the build script and that implicitly sets up ert using `.[dev]` and more?

andreas-el (changes requested):
Good work here, but I'd like a quick explanation on one item.

ci/run_ert_ctests.sh
andreas-el: That would be the only advantage. And even so we should try to get it to run using the newer `gcc-toolset-13` instead once we get to RHEL8.
A sidenote would be that we are quite close to removing the C-code altogether, so this might not be relevant in the too distant future.

ci/run_ert_ctests.sh
andreas-el: Great. Seems like no reason to keep it. 👍

andreas-el (approved):
Good work! 🚤

PR #7330, opened 2024-03-01
Use PyQt6 on Python 3.9+

pinkwah:
Start migrating over to Qt6 by enabling PyQt6 on Python 3.9+. PyQt6 does not support RHEL7, but it does RHEL8 so this is a way for us to do this upgrade.

Note: On RHEL8, we need to install the package `xcb-util-cursor` for the GUI to start.


sondreso:
Closing this due to inactivity, feel free to reopen in the future!

PR #7333, opened 2024-03-04
Try pytest retry

### LINKED ISSUE #7332: Alternatives to flaky?
Using flaky as it is now will require us to pin pytest to `<8.1.0`. Flaky is kind of a stale project and imports from the private namespace of pytest, if there is a less stale alternative we should use that instead. Maybe https://pypi.org/project/pytest-retry ?

Related: https://github.com/box/flaky/issues/198

yngve-sk:
**Issue**
Resolves #7332


yngve-sk:
Requires python >= 3.9

PR #7338, opened 2024-03-04
Add columns to Manage Cases overview

pinkwah:
Skjermbilde fra 2024-03-04 09-50-29

This adds a new Python package called `humanize` for pretty-printing dates. I felt like ISO-8601 were too difficult to parse when what you are likely to want is "Is this the experiment that I started now or the one I started yesterday?". You can hover over to get a tooltip. Although in the future we should override the default Qt tooltip because it takes forever to show up.


frode-aarstad (approved):
Looks nice 🚀

PR #7347, opened 2024-03-04
Remove usages of Reader/Accessor alias

pinkwah:
At one point there was a difference between Reader and Accessors. Now they are the same class, with aliasing kept for compatibility. This is the first step to remove the distinction in the Ert codebase.

The exact commands run were:
```sh
find src -type f -name "*.py" -exec sed -i "s,\(Storage\|Experiment\|Ensemble\)\(Reader\|Accessor\),\1,g" '{}' \;
find tests -type f -name "*.py" -exec sed -i "s,\(Storage\|Experiment\|Ensemble\)\(Reader\|Accessor\),\1,g" '{}' \;
```
...and excepting `src/ert/storage/__init__.py`


frode-aarstad (approved):
LGTM

PR #7367, opened 2024-03-06
Make qt error message for plot fail resizable+selectable

yngve-sk:
QMessageBox is per definition not resizable (possible but PITA to achieve), so using QDialog+modal instead

This (resizable, selectable & with copy button)
Mar-06-2024 11-43-02

Instead of this
Screenshot 2024-03-06 at 09 12 32


eivindjahren (approved):
:technologist:

Issue #7325, opened 2024-03-01
Exiting state in PBS driver must emit a FinishedEvent


berland:
Integration testing has revealed that jobs linger in the "Exiting" state for many minutes before entering "Finished" state. The driver will wait until the Finished state before it deems a realization done.

There is a belief that the Exit_status might not be available in the "Exiting" state, but sample testing reveal that it is. Not sure if that is a timing issue, and that there might be situations where Exit_status is not ready in the Exiting state.

https://github.com/equinor/ert/blob/f26a69942191141f1d29998f2fb4010655096502/src/ert/scheduler/openpbs_driver.py#L48C1-L52C1

berland:
The legacy C driver treats the E state equal to F state.

pinkwah:
I think we can assume that E and F have the same state. I have no good reason for only checking one of them.

berland:
Yes, assume they are the same, and error hard if Exit_status is not present (then fix that problem if it occurs).

PR #7371, opened 2024-03-06
OpenPBS: Treat 'E' state the same as 'F'

pinkwah:
Both E and F mean that the process has finished. The difference seems that jobs in E state might still be using the queue. This distinction is of no concern to us.

Resolves: #7325


xjules (approved):
Nice! :rocket:

berland:
I think you want to poke these dicts slightly too:
https://github.com/equinor/ert/blob/a0849763ae795000ac2e15ddeeffa1fa300ec1e6/src/ert/scheduler/openpbs_driver.py#L233-L234

pinkwah:
Changed it to save the Pydantic model instead of the "job status" letter. This means that `E` and `F` are the same `FinishedJob` and so the processing returns early.

berland:
Tested on s034-lcatop01 with es_mda on poly_example ✅

PR #7372, opened 2024-03-06
Make sure only the main thread sets a new signal handler

DanSava:
**Issue**
Resolves https://github.com/equinor/everest/issues/2227

**Approach**
Make sure only the main thread sets a new signal handler

## When applicable

### DIFF
--- a/src/_ert/threading.py
+++ b/src/_ert/threading.py
@@ -68,4 +68,5 @@ def _handler(signum: int, frametype: FrameType | None) -> None:
     raise current_exception
 
 
-signal.signal(signal.SIGUSR1, _handler)
+if threading.current_thread() is threading.main_thread():
+    signal.signal(signal.SIGUSR1, _handler)


pinkwah:
I don't know whether this is a sufficient fix. If an error occurs within a `ErtThread`, it'll take down the entire `pytest` process which is bad.

```
zom@be-lx136139 ~> python3.8
Python 3.8.17 (default, Aug 10 2023, 12:50:17) 
[GCC 8.5.0 20210514 (Red Hat 8.5.0-20)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os, signal
>>> os.kill(os.getpid(), signal.SIGUSR1)
fish: Job 1, 'python3.8' terminated by signal SIGUSR1 (User defined signal 1)
```

In general, this sounds like Everest's `pytest` is prone to crashing. I think a fair question to ask is whether the behaviour that ErtThread brings to Everest pytest is desirable. If it is, then this is not the way, and Everest should do whatever it is Ert does. If it isn't, we should maybe change it so that only Ert enables this functionality.

DanSava:
I am not sure what Ert is doing that everest is not but I can tell you that this issue happens only if we are running the Everest tests in parallel with `pytest -n #` once I remove the `-n #` from the pytest command the tests are running just fine. 

Adding the fix from the PR will solve the issue for the concurrent test running in everest, but I agree that this fix might be incomplete given that I did not spend a lot of time trying to figure out how this ErtThread is impacting everest.

pinkwah:
I made another PR here, where this functionality is opt-in: https://github.com/equinor/ert/pull/7374

My understanding is that this works in Ert and not Everest because we have a top-level `conftest.py` that eventually imports this file. Pytest evaluates all `conftest.py`s, then it makes a threadpool to execute the tests. If there is a `conftest.py` that imports `ErtThread`, then this module is pre-loaded when the threads start going, so it works in Ert.

pinkwah:
I'm closing this as I just merged the other PR. Thanks for the PR though. I don't think I would've done it now otherwise.

PR #7373, opened 2024-03-06
Do not retry when we get 'Job has finished' from qdel

### LINKED ISSUE #7369: Killing simulations cause qdel retries on PBS

xjules:
**Issue**
Resolves #7369 

**Approach**
Do not retry when qdel failed with `Job has finished`

## When applicable

### DIFF
--- a/src/ert/scheduler/openpbs_driver.py
+++ b/src/ert/scheduler/openpbs_driver.py
@@ -3,7 +3,16 @@ from __future__ import annotations
 import asyncio
 import logging
 import shlex
-from typing import List, Literal, Mapping, MutableMapping, Optional, Tuple, Union
+from typing import (
+    Iterable,
+    List,
+    Literal,
+    Mapping,
+    MutableMapping,
+    Optional,
+    Tuple,
+    Union,
+)
 
 from pydantic import BaseModel, Field
 from typing_extensions import Annotated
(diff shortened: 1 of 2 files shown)


tests/unit_tests/scheduler/test_openpbs_driver.py
```diff
@@ -326,7 +326,13 @@ async def test_that_qdel_will_retry_and_succeed(
     driver._iens2jobid[0] = 111
     await driver.kill(0)
     assert "TRIED" in Path(bin_path / "script_try").read_text(encoding="utf-8")
-    assert "qdel executed" in Path(bin_path / "qdel_output").read_text(encoding="utf-8")
+    if exit_code == QDEL_JOB_HAS_FINISHED:
+        # the job has been already qdel-ed so no need to retry
+        assert not os.path.exists(bin_path / "qdel_output")
+    else:
+        assert "qdel executed" in Path(bin_path / "qdel_output").read_text(
```
pinkwah, line 333: Nitpick
```suggestion
        assert "qdel executed" in (bin_path / "qdel_output").read_text(
```

tests/unit_tests/scheduler/test_openpbs_driver.py
```diff
@@ -326,7 +326,13 @@ async def test_that_qdel_will_retry_and_succeed(
...
+            encoding="utf-8"
```
pinkwah, line 334: Nitpick: Is encoding required? Is the default not `"utf-8"` already?

src/ert/scheduler/openpbs_driver.py
```diff
@@ -36,13 +36,14 @@
...
 
-QSUB_EXIT_CODES = [
+QSUB_EXIT_CODES_TO_RETRY = [
     QSUB_INVALID_CREDENTIAL,
     QSUB_PREMATURE_END_OF_MESSAGE,
     QSUB_CONNECTION_REFUSED,
 ]
 
-QDEL_EXIT_CODES = [QDEL_REQUEST_INVALID, QDEL_JOB_HAS_FINISHED]
+QDEL_EXIT_CODES_TO_RETRY = [QDEL_REQUEST_INVALID]
```
pinkwah, line 45: Nitpick: the rest of the Scheduler code uses "returncode" (from `asyncio.process.Process`'s `returncode` property), not "exit_code".

Also, perhaps this should be inlined with the `qdel` code?

pinkwah (approved):
Approving with nitpicks

src/ert/scheduler/openpbs_driver.py
xjules: I've removed the groups and inject them directly as individual return_codes :+1:

tests/unit_tests/scheduler/test_openpbs_driver.py
berland: We should use the encoding yes, utf-8 is default on our computers yes, but lets be kind to the world.

See https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/unspecified-encoding.html

berland: This was enforced earlier when we had  pylint in CI, now I see that our codebase has deterioted on this after the switch to ruff 😢

pinkwah: From the PEP it sounds like it only is a problem on Windows. It's what we in the activism business called praxis in our fight against the Microsoft monopoly!

But, joking aside, it is reasonable to specify `encoding="utf-8"` I guess. I always thought it was defined to be as such in Python 3, now that `str` is unicode.

PR #7374, opened 2024-03-06
Make ErtThread signal and re-raising opt-in

pinkwah:
This commit does the following:
1. Adds a function `set_signal_handler` that must be called before re-raising works
2. Adds a kwarg `should_raise` that must be True for re-raising to work
3. Workflow jobs don't reraise, only log


eivindjahren (changes requested):
I think you do not want to change the commit `test-data/block_storage` points to.

eivindjahren (approved):

PR #7376, opened 2024-03-06
Avoid "qstat -f" for non-finished jobs in PBS

### LINKED ISSUE #7368: PBS driver makes a too heavy `qstat` call
Calling `qstat` with `-f` (f for "full") on every job in an experiment is too heavy for the PBS cluster. 

We need to resort to parse the default `qstat` output. This will not give us the Exit_status, so some other measure needs to be taken to get the Exit_status for a job when it is finished. One way to do that is to do the qstat call with -f whenever a realization is finished.

berland:
qstat -f (f for "full format") is a heavy operation for the PBS cluster, and we only use it to obtain the Exit status for the job.

This will change the polling into using qstat -f only for jobs that are already marked as finished, through polling with the default qstat output.

**Issue**
Resolves #7368 

**Approach**
see commit

## When applicable

### DIFF
--- a/src/ert/scheduler/openpbs_driver.py
+++ b/src/ert/scheduler/openpbs_driver.py
@@ -4,14 +4,17 @@ import asyncio
 import logging
 import shlex
 from typing import (
+    Dict,
     Iterable,
     List,
     Literal,
     Mapping,
     MutableMapping,
     Optional,
+    Set,
     Tuple,
     Union,
+    get_args,
 )
 
 from pydantic import BaseModel, Field
@@ -43,6 +46,7 @@ QSUB_PREMATURE_END_OF_MESSAGE = 183
 QSUB_CONNECTION_REFUSED = 162
 QDEL_JOB_HAS_FINISHED = 35
 QDEL_REQUEST_INVALID = 168
+QSTAT_UNKNOWN_JOB_ID = 153
 
 
 class IgnoredJobstates(BaseModel):
@@ -59,7 +63,7 @@ class RunningJob(BaseModel):
 
 class FinishedJob(BaseModel):
     job_state: Literal["E", "F"]
-    returncode: Annotated[int, Field(alias="Exit_status")]
+    returncode: Annotated[Optional[int], Field(alias="Exit_status")] = None
 
 
 AnyJob = Annotated[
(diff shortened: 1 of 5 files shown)


src/ert/scheduler/openpbs_driver.py
```diff
@@ -206,6 +229,20 @@ async def submit(
         logger.debug(f"Realization {iens} accepted by PBS, got id {job_id_}")
         self._jobs[job_id_] = (iens, QueuedJob())
         self._iens2jobid[iens] = job_id_
+        self._non_finished_job_ids.add(job_id_)
+
+    def _expand_truncated_jobids(
+        self, truncated_jobs: Dict[str, Dict[str, Dict[str, str]]]
+    ) -> Dict[str, Dict[str, Dict[str, str]]]:
+        """Job ids gotten through normal qstat are truncated at length 16"""
```
pinkwah, line 237: Could this pose an issue?

src/ert/scheduler/openpbs_driver.py
```diff
@@ -226,25 +263,64 @@ async def kill(self, iens: int) -> None:
...
+                    stderr=asyncio.subprocess.PIPE,
+                )
+                stdout, stderr = await process.communicate()
+                if process.returncode not in {0, QSTAT_UNKNOWN_JOB_ID}:
+                    # Any unknown job ids will yield QSTAT_UNKNOWN_JOB_ID, but
+                    # results for other job ids on stdout can be assumed valid.
+                    await asyncio.sleep(_POLL_PERIOD)
+                    continue
+                if process.returncode == QSTAT_UNKNOWN_JOB_ID:
+                    logger.debug(
```
pinkwah, line 285: Shouldn't this be a `logger.error`? Ert shouldn't ask qstat for jobs that it doesn't know about.

tests/unit_tests/scheduler/test_openpbs_driver.py
```diff
@@ -191,6 +199,100 @@ async def test_cluster_label():
...
+        pytest.param(
+            "echo 'qstat: Unknown Job Id 1' >&2; exit 1",
+            False,
+            id="empty_cluster_specific_id",
+        ),
+        pytest.param(
+            "echo '1 foo someuser 0 R normal'",
+            True,
+            id="missing_header_is_accepted",  # (debatable)
+        ),
```
pinkwah, line 238: Eh, we can remove this.

pinkwah (approved):
Good stuff! This review was sponsored by Nitpicks :tm:

src/ert/scheduler/openpbs_driver.py
berland: If so, I suppose it is fixable by cluster infrastructure, I suppose it would be in a situation where there is a long string prepended to the id, but I don't think that string can be set by the user.

src/ert/scheduler/openpbs_driver.py
berland: I think I chose debug because of the volume of logging that this potentially could give (worst case is every realization every second second)

tests/unit_tests/scheduler/test_openpbs_driver.py
berland: ✂️

src/ert/scheduler/openpbs_driver.py
berland: And I am not convinced upfront that it always would be Ert that is at fault when/if this happens.

PR #7385, opened 2024-03-07
Treat exceptions of long lived tasks (scheduling_tasks) in scheduler

xjules:
**Issue**
Resolves #7192
In addition, this handles exception in long lived tasks (scheduling_tasks), which previously were not handled.

**Approach**
Treat exception in long lived tasks by cancelling the jobs correctly. Added tests.

## When applicable


src/ert/scheduler/scheduler.py
```diff
@@ -15,7 +16,6 @@
 from websockets import Headers
 from websockets.client import connect
 
-from ert.async_utils import background_tasks
```
berland, line 18: This function in `ert.async_utils` is only used here, so it should probably be purged from async_utils , so this PR does not leave dead code lying around.

src/ert/scheduler/scheduler.py
```diff
@@ -181,32 +181,52 @@ async def execute(
         # We need to store the loop due to when calling
         # cancel jobs from another thread
         self._loop = asyncio.get_running_loop()
-        async with background_tasks() as cancel_when_execute_is_done:
-            cancel_when_execute_is_done(self._publisher())
-            cancel_when_execute_is_done(self._process_event_queue())
-            cancel_when_execute_is_done(self.driver.poll())
-            if min_required_realizations > 0:
-                cancel_when_execute_is_done(
+        long_lived_tasks = [
```
berland, line 184: We should call this something more precise, as these tasks are roughly just as longlived as the other ( = realization) tasks.

berland: "event_handler_tasks" ?

berland: hmm, that is maybe not a good name given we also add stop_long_running_jobs to the list.

xjules: `long_lived_tasks` is semantically correct, but it does not say too much - I agree. `scheduler_maintenance_tasks`?

src/ert/scheduler/scheduler.py
```diff
@@ -181,32 +181,52 @@ async def execute(
...
+        async def gather_jobs() -> asyncio.futures.Future[list[BaseException | None]]:
+            try:
+                return asyncio.gather(*self._tasks.values(), return_exceptions=True)
+            finally:
+                for ll_task in long_lived_tasks:
+                    ll_task.cancel()
+
+        job_results: Optional[list[BaseException | None]] = None
+        try:
+            job_results = (await asyncio.gather(gather_jobs(), *long_lived_tasks))[0]
```
berland, line 213: gather(gather(gather())) 🤯

berland:
`ert ensemble_experiment poly_torque.ert `
currently hangs with mocked `qsub` and friends.

xjules:
> `ert ensemble_experiment poly_torque.ert ` currently hangs with mocked `qsub` and friends.

Yes, there was a bug, but should be good now.

src/ert/scheduler/scheduler.py
xjules: Comments added

src/ert/scheduler/scheduler.py
jonathan-eq: What about just scheduler_tasks? The other tasks are job_tasks, which are dependent on the scheduler through the scheduler_tasks.

xjules:
> If I have a `1/0` in poll() then `ert ensemble_experiment poly_torque.ert` still hangs (one realization is set to failed in the monitor before it hangs). Also `grep -i zero logs/*` yields no hit which I suppose it should?

Yes, if poll() fails immediately (or sooner that submit was called) that it hangs.
It shows here:
https://github.com/equinor/ert/actions/runs/8195586367/job/22414187726?pr=7385 
Trying to figure out why.

xjules:
> > If I have a `1/0` in poll() then `ert ensemble_experiment poly_torque.ert` still hangs (one realization is set to failed in the monitor before it hangs). Also `grep -i zero logs/*` yields no hit which I suppose it should?
> 
> Yes, if poll() fails immediately (or sooner that submit was called) that it hangs. It shows here: https://github.com/equinor/ert/actions/runs/8195586367/job/22414187726?pr=7385 Trying to figure out why.

It was `await self.returncode` in job cancellation handling. We can just cancel right away I think.

berland:
I tried with a `1/0` in pbs drivers submit(). ERT stops with all realizations failed, but there is no mention of the exception on either stdout or in logs/

xjules:
> I tried with a `1/0` in pbs drivers submit(). ERT stops with all realizations failed, but there is no mention of the exception on either stdout or in logs/

There is a corresponding issue, which should handle various muted exceptions from job (including submit calls). https://github.com/equinor/ert/issues/7408

src/ert/scheduler/scheduler.py
```diff
@@ -181,32 +181,70 @@ async def execute(
...
+            scheduling_tasks.append(asyncio.create_task(self._update_avg_job_runtime()))
+
+        start = asyncio.Event()
+        sem = asyncio.BoundedSemaphore(self._max_running or len(self._jobs))
+        for iens, job in self._jobs.items():
+            self._tasks[iens] = asyncio.create_task(job(start, sem, self._max_submit))
+        start.set()
+
+        async def gather_realization_jobs() -> list[BaseException | None]:
+            """This handles realization jobs. It makes sure that all the tasks
```
berland, line 205: "This handles realization jobs" is not needed as it now is in the function name.

src/ert/scheduler/scheduler.py
```diff
@@ -181,32 +181,70 @@ async def execute(
...
+            are completed, where afterwards we cancel scheduling_tasks.
+            It returns a future object, which keeps the results of running tasks.
+            """
+            try:
+                return await asyncio.gather(
+                    *self._tasks.values(), return_exceptions=True
+                )
+            finally:
+                for ll_task in scheduling_tasks:
```
berland, line 214: `ll_task` -> `scheduling_task`

tests/integration_tests/scheduler/test_openpbs_driver.py
```diff
@@ -26,3 +30,43 @@ def test_openpbs_driver_with_poly_example():
         "--enable-scheduler",
         "poly.ert",
     )
+
+
+@pytest.mark.timeout(180)
```
berland, line 35: 3 minutes is maybe longer than necessary?

berland: Or you could just use the default timeout if you do not want to tune it.

tests/integration_tests/scheduler/test_openpbs_driver.py
```diff
@@ -26,3 +30,43 @@ def test_openpbs_driver_with_poly_example():
...
+    with pytest.raises(ErtCliError):
+        run_cli(
+            ENSEMBLE_EXPERIMENT_MODE,
+            "--enable-scheduler",
+            "poly.ert",
+        )
+    assert "ZeroDivisionError" in caplog.text
+
+
+@pytest.mark.timeout(60)
```
berland, line 53: ditto

src/ert/scheduler/scheduler.py
xjules: yes, recommended by Prof. @pinkwah ! :smile:

src/ert/scheduler/scheduler.py
pinkwah: `ensemble_tasks`?

src/ert/scheduler/scheduler.py
pinkwah: - Builtin `list[T]` type is new in Python 3.9, with `typing.List` being deprecated
- Union syntax `X | Y` (equivalent to `Union[X, Y]`) is new in Python 3.10

Because we use `from __future__ import annotations`, these expressions are never computed during runtime, so this works in Python 3.8. Mypy evaluates the types separately, and uses the latest Python standard regardless of which CPython version you're running `mypy` from.

The only place this cannot be used (until we deprecate Python 3.9) is where runtime information is important, such as dataclasses:
```py
@dataclass
class Foo:
    mylist: List[Union[int, str]]  # _must_ use Python3.8-compatible syntax

    def do_something(self) -> list[int | str]:  # Not interpreted by Python, so we can use Python3.12 syntax
        ...
```

It might cause confusion but my current take is that 80% of our code is _just_ type hinting, and the new syntax is easier to understand at a glance, so it's worth.

src/ert/scheduler/job.py
```diff
@@ -117,8 +116,7 @@ async def _submit_and_run_once(self, sem: asyncio.BoundedSemaphore) -> None:
         except asyncio.CancelledError:
             await self._send(State.ABORTING)
             await self.driver.kill(self.iens)
-            with suppress(asyncio.CancelledError):
-                await self.returncode
+            self.returncode.cancel()
```
pinkwah, line 119: The reason we did `await self.returncode` is to ensure that the job is dead before we say `State.ABORTED`. Is it reasonable for this to be fire-and-forget?

src/ert/scheduler/scheduler.py
```diff
@@ -181,32 +181,69 @@ async def execute(
...
-                *self._tasks.values(), return_exceptions=True
             )
-            for result in results:
-                if isinstance(result, Exception):
-                    logger.error(result)
-
-            await self.driver.finish()
+            scheduling_tasks.append(asyncio.create_task(self._update_avg_job_runtime()))
+
+        start = asyncio.Event()
```
pinkwah, line 198: Small thing but maybe if you're already touching this code, remove this unnecessary `start` event.

src/ert/scheduler/scheduler.py
```diff
@@ -181,32 +181,69 @@ async def execute(
...
+                for scheduling_task in scheduling_tasks:
+                    scheduling_task.cancel()
+
+        job_results: Optional[list[BaseException | None]] = None
+        try:
+            # we gather collectively results of realization tasks and scheduling_tasks
+            # If Exception/CancellationError were to happen we know it's coming
+            # from scheduling_tasks only and thus we need to cancel realization tasks.
+            # Otherwise potential exceptions from realization tasks
+            # are stored in job_results and handled in else branch
```
pinkwah, line 222: "we gather collectively results..."?

```suggestion
            # `gather_realization_jobs` does not raise; rather, each job task's result
            # is collected into the returning list. If an exception occurs, it must
            # necessarily come from the scheduling tasks. Exceptions from the
            # job tasks are handled in the `else` branch after the evaluation
            # has stopped
```

tests/integration_tests/scheduler/test_openpbs_driver.py
```diff
@@ -26,3 +30,43 @@ def test_openpbs_driver_with_poly_example():
...
     )
+
+
+@pytest.mark.timeout(30)
+@pytest.mark.integration_test
+@pytest.mark.usefixtures("copy_poly_case")
+def test_openpbs_driver_with_poly_example_failing_submit_fails_ert_and_propagates_exception_to_user(
+    monkeypatch, caplog
+):
+    monkeypatch.setattr(OpenPBSDriver, "submit", lambda *args, **kwargs: 1 / 0)
```
pinkwah, line 41: Don't use `1 / 0` to trigger an error in tests. Use `raise RuntimeError` or something where it's obvious that the error is intended by the test.

tests/integration_tests/scheduler/test_openpbs_driver.py
```diff
@@ -26,3 +30,43 @@ def test_openpbs_driver_with_poly_example():
...
+
+@pytest.mark.timeout(30)
+@pytest.mark.integration_test
+@pytest.mark.usefixtures("copy_poly_case")
+def test_openpbs_driver_with_poly_example_failing_poll_fails_ert_and_propagates_exception_to_user(
+    monkeypatch, caplog
+):
+    async def mock_poll(*args, **kwargs):
+        await asyncio.sleep(2.0)
+        raise ZeroDivisionError("Status polling fails")
```
pinkwah, line 61: This is quite silly. Replace with `RuntimeError` for sanity?

tests/unit_tests/scheduler/test_scheduler.py
```diff
@@ -468,3 +469,46 @@ async def wait():
         for start, next_start in zip(run_start_times[:-1], run_start_times[1:])
     ]
     assert min(deltas) >= submit_sleep * 0.8
+
+
+@pytest.mark.timeout(5)
+async def test_driver_poll_exceptions_are_propagated(realization):
+    driver = Mock()
+    driver.poll = lambda: 1 / 0
```
pinkwah, line 477: Ditto wrt. error type. Something like `RuntimeError("Test poll() raises exception")` and match on that entire string further down so that this test doesn't pass if we every _actually_ divide by zero scheduler.

pinkwah (changes requested):
I've mentioned that exception messages should be very obvious that they come from tests before and I felt like there was agreement in that regard. Please fix.

Also English syntax errors in comment.

I'm not too bothered about variable naming.

Otherwise LGTM

src/ert/scheduler/job.py
xjules: Yes, there is. If an exception in Polling then `await self.returncode` will hang (while we cal `job_task.cancel()`) which is treated here.

pinkwah (approved):

xjules:
Don't merge yet! test_scheduler is WIP

src/ert/scheduler/scheduler.py
xjules: Apparently, after removing the "global" start this test becomes flaky: `test_new_monitor_can_pick_up_where_we_left_off` I'll put it back and create an issue on it instead.

PR #7392, opened 2024-03-07
Support old convention for naming "FieldConfig" "Field"

### LINKED ISSUE #7388: Cannot open storage created with ert==8.4
Opening storage created with ert 8.4 fails with
```
>           params[data["name"]] = _KNOWN_PARAMETER_TYPES[param_type](**data)
E           TypeError: GenKwConfig.__init__() missing 1 required positional argument: 'update'
```

Can be reproduced with the following test:
https://github.com/equinor/ert/blob/dd13514d59d1920742a9180cb88109740a8d5d6c/tests/integration_tests/test_storage_migration.py#L11

Solving this issue should remove the xfail from that test.

When there is no GenKW you get the following error:

```
    @cached_property
    def parameter_configuration(self) -> Dict[str, ParameterConfig]:
        params = {}
        for data in self.parameter_info.values():
            param_type = data.pop("_ert_kind")
            if "update" not in data:
                data["update"] = True
>           params[data["name"]] = _KNOWN_PARAMETER_TYPES[param_type](**data)
E           KeyError: 'Field'

/src/ert/storage/local_experiment.py:195: KeyError
```

yngve-sk:
**Issue**
Resolves #7388

### DIFF
--- a/src/ert/storage/local_experiment.py
+++ b/src/ert/storage/local_experiment.py
@@ -37,6 +37,7 @@ _KNOWN_PARAMETER_TYPES = {
     GenKwConfig.__name__: GenKwConfig,
     SurfaceConfig.__name__: SurfaceConfig,
     FieldConfig.__name__: FieldConfig,
+    "Field": FieldConfig,
     ExtParamConfig.__name__: ExtParamConfig,
 }
 


src/ert/storage/local_experiment.py
```diff
@@ -37,6 +37,7 @@
     GenKwConfig.__name__: GenKwConfig,
     SurfaceConfig.__name__: SurfaceConfig,
     FieldConfig.__name__: FieldConfig,
+    "Field": FieldConfig,
```
eivindjahren, line 40: Also need to ensure that "Field" is written to storage instead of "FieldConfig".

yngve-sk: I don't think that's necessary?
If it writes FieldConfig, FieldConfig lookup will give ...`FieldConfig`.
If it has written (with old storage version), Field, it will also give `FieldConfig`

eivindjahren: I think over time, trying to support older versions in this way is going to get hairy. I would rather we continued to migrate rather than keep aliases.

PR #7396, opened 2024-03-08
Use job script on disk for LSF driver and use stdin for PBS

### LINKED ISSUE #7327: Integration test for PBS will never work with current driver
There are reasons to believe that this integration test:

https://github.com/equinor/ert/blob/f26a69942191141f1d29998f2fb4010655096502/tests/integration_tests/scheduler/test_openpbs_driver.py#L53

will never work with the current driver. 

The current driver uses a command like "qsub -- myexecutable myargs"

and there seems to be issues with how the ">" is handled (maybe as a redirection to the qsub command. Some attempts with escaping has not been successful

If  "myexecutable myargs" are injected into a shell/bash script, and then executed, then it will work.

Also note that having "echo" as an executable will also not work. It must be "/bin/echo".

berland:
Let PBS and LSF make explicit job scripts for test robustness 

For OpenPBS the explicit job script only exists in memory and
is provided to qsub via stdin.

For LSF, the bsub command seemingly cannot read stdin, and the
job script is placed on runpath with a unique filename that
will be left behind.

This enables more complex jobs (e.g. redirection) submitted to the
drivers submit() function.

Align Python LSF driver with C-driver in supplying the runpath as the
last argument. This is needed to pass a legacy test. Note that legacy
LSF and legacy Torque drivers differ in this behaviour, the latter does
not supply runpath.

The type of the runpath object supplied is changed from str to
pathlib.Path

**Issue**
Resolves #7327 

**Approach**

## When applicable

### DIFF
--- a/ci/testkomodo.sh
+++ b/ci/testkomodo.sh
@@ -59,8 +59,13 @@ start_tests () {
 
     # Using presence of "bsub" in PATH to detect onprem vs azure
     if which bsub >/dev/null && basetemp=$(mktemp -d -p ~/pytest-tmp); then
-        pytest -v --lsf --basetemp="$basetemp" integration_tests/scheduler/test_lsf_driver.py && \
-        rm -rf "$basetemp"
+        pytest -v --lsf --basetemp="$basetemp" integration_tests/scheduler && \
+            rm -rf "$basetemp"
+    fi
+    if ! which bsub 2>/dev/null && basetemp=$(mktemp -d -p ~/pytest-tmp); then
+        export _ERT_TESTS_DEFAULT_QUEUE_NAME=permanent
+        pytest -v --openpbs --basetemp="$basetemp" integration_tests/scheduler && \
+            rm -rf "$basetemp"
     fi
     popd
 
--- a/src/ert/scheduler/lsf_driver.py
+++ b/src/ert/scheduler/lsf_driver.py
@@ -6,6 +6,8 @@ import logging
 import re
 import shlex
 import shutil
+import stat
+import tempfile
 from pathlib import Path
 from typing import (
     Dict,
@@ -123,12 +125,36 @@ class LsfDriver(Driver):
         /,
         *args: str,
         name: str = "dummy",
-        runpath: Optional[str] = None,
+        runpath: Optional[Path] = None,
     ) -> None:
+        if runpath is None:
+            runpath = Path.cwd()
+
         arg_queue_name = ["-q", self._queue_name] if self._queue_name else []
 
+        script = (
+            "#!/usr/bin/env bash\n"
+            f"cd {shlex.quote(str(runpath))}\n"
+            f"exec -a {shlex.quote(executable)} {executable} {shlex.join(args)}\n"
+        )
+        script_path: Optional[Path] = None
+        with tempfile.NamedTemporaryFile(
+            dir=runpath,
+            prefix=".lsf_submit_",
+            suffix=".sh",
+            mode="w",
+            encoding="utf-8",
+            delete=False,
+        ) as script_handle:
+            script_handle.write(script)
+            script_path = Path(script_handle.name)
+        assert script_path is not None
+        script_path.chmod(script_path.stat().st_mode | stat.S_IEXEC)
+
         bsub_with_args: List[str] = (
-            [str(self._bsub_cmd)] + arg_queue_name + ["-J", name, executable, *args]
+            [str(self._bsub_cmd)]
+            + arg_queue_name
+            + ["-J", name, str(script_path), str(runpath)]
         )
         logger.debug(f"Submitting to LSF with command {shlex.join(bsub_with_args)}")
         process = await asyncio.create_subprocess_exec(
@@ -153,10 +179,9 @@ class LsfDriver(Driver):
         job_id = match[1]
         logger.info(f"Realization {iens} accepted by LSF, got id {job_id}")
 
-        if runpath is not None:
-            (Path(runpath) / LSF_INFO_JSON_FILENAME).write_text(
-                json.dumps({"job_id": job_id}), encoding="utf-8"
-            )
+        (Path(runpath) / LSF_INFO_JSON_FILENAME).write_text(
+            json.dumps({"job_id": job_id}), encoding="utf-8"
+        )
         self._jobs[job_id] = (iens, QueuedJob(job_state="PEND"))
         self._iens2jobid[iens] = job_id
 
--- a/src/ert/scheduler/openpbs_driver.py
+++ b/src/ert/scheduler/openpbs_driver.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import logging
 import shlex
+from pathlib import Path
 from typing import (
     Dict,
     Iterable,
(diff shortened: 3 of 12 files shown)


berland:
Onprem testing:
`pytest --basetemp=/private/havb/pytest-tmp --lsf tests/integration_tests/scheduler  -sxv
`
Azure testing:
```
export _ERT_TESTS_DEFAULT_QUEUE_NAME=permanent
pytest --basetemp=/private/havb/pytest-tmp --openpbs tests/integration_tests/scheduler  -sxv

```

berland:
Some clearification:

* We do not like leaving temporary files, especially in runpath, where it can become a de-facto API
* We want the drivers to behave similarly

Currently, the on-prem /global/bin/bsub cannot take the job script on stdin, which leaves the two goals above impossible to attain.

berland:
Integration tests with --openpbs on Azure and --lsf onprem has been manually verified (error there will not be visible until next komodo build)

ci/testkomodo.sh
```diff
@@ -59,8 +59,13 @@ start_tests () {
 
     # Using presence of "bsub" in PATH to detect onprem vs azure
     if which bsub >/dev/null && basetemp=$(mktemp -d -p ~/pytest-tmp); then
-        pytest -v --lsf --basetemp="$basetemp" integration_tests/scheduler/test_lsf_driver.py && \
-        rm -rf "$basetemp"
+        pytest -v --lsf --basetemp="$basetemp" integration_tests/scheduler && \
+            rm -rf "$basetemp"
+    fi
+    if ! which bsub 2>/dev/null && basetemp=$(mktemp -d -p ~/pytest-tmp); then
+        export _ERT_TESTS_DEFAULT_QUEUE_NAME=permanent
```
xjules, line 66: So if `bsub` is not available we automatically g for openpbs? Should we check that `qsub` is available?

src/ert/scheduler/lsf_driver.py
```diff
@@ -123,12 +125,36 @@ async def submit(
...
     ) -> None:
+        if runpath is None:
+            runpath = Path.cwd()
+
         arg_queue_name = ["-q", self._queue_name] if self._queue_name else []
 
+        script = (
+            "#!/usr/bin/env bash\n"
+            f"cd {shlex.quote(str(runpath))}\n"
+            f"exec -a {shlex.quote(executable)} {executable} {shlex.join(args)}\n"
```
xjules, line 138: So this script is to make sure ">" in `executable` is handled correctly?

src/ert/scheduler/lsf_driver.py
```diff
@@ -123,12 +125,36 @@ async def submit(
...
+            script_handle.write(script)
+            script_path = Path(script_handle.name)
+        assert script_path is not None
+        script_path.chmod(script_path.stat().st_mode | stat.S_IEXEC)
+
         bsub_with_args: List[str] = (
-            [str(self._bsub_cmd)] + arg_queue_name + ["-J", name, executable, *args]
+            [str(self._bsub_cmd)]
+            + arg_queue_name
+            + ["-J", name, str(script_path), str(runpath)]
```
xjules, line 157: I guess this should be only:
`["-J", name, str(script_path)]`

src/ert/scheduler/openpbs_driver.py
```diff
@@ -202,13 +205,24 @@ async def submit(
...
         arg_keep_qsub_output = (
             [] if self._keep_qsub_output else "-o /dev/null -e /dev/null".split()
         )
 
+        script = (
+            "#!/usr/bin/env bash\n"
+            f"cd {shlex.quote(str(runpath))}\n"
+            f"exec -a {shlex.quote(executable)} {executable} {shlex.join(args)}\n"
+        )
+        script_path = runpath / "job_script.sh"
```
xjules, line 223: Why not to use the same `tempfile.NamedTemporaryFile` since when `runpath=None` this will not work or ?

src/ert/scheduler/lsf_driver.py
berland: Yes, it will guarantee that regardless of any bugs in args handling in qsub/bsub.

src/ert/scheduler/openpbs_driver.py
berland: Good catch, it is not used any longer after changing this PR to use stdin

src/ert/scheduler/lsf_driver.py
berland: hmm, but then again it seems odd to have it introduced in this PR.. 👀

berland: Explained in commit message:
    Align Python LSF driver with C-driver in supplying the runpath as the
    last argument. This is needed to pass a legacy test. Note that legacy
    LSF and legacy Torque drivers differ in this behaviour, the latter does
    not supply runpath.

tests/integration_tests/scheduler/test_generic_driver.py
```diff
@@ -1,43 +1,58 @@
...
 
     # It's not possible to dynamically choose a pytest fixture in a fixture, so
     # we copy some code here
     if class_ is OpenPBSDriver and pytestconfig.getoption("openpbs"):
         # User provided --openpbs, which means we should use the actual OpenPBS
         # cluster without mocking anything.
-        pass
+        if str(tmp_path).startswith("/tmp"):
+            print(
+                "Please use --basetemp option to pytest, PBS tests needs a shared disk"
```
xjules, line 27:
```suggestion
                "`Please use --basetemp option to pytest, the real PBS cluster needs a shared disk`"
```

tests/integration_tests/scheduler/test_openpbs_driver.py
```diff
@@ -19,12 +20,14 @@ def mock_openpbs(pytestconfig, monkeypatch, tmp_path):
...
 
-@pytest.mark.timeout(30)
 @pytest.mark.integration_test
 @pytest.mark.usefixtures("copy_poly_case")
 def test_openpbs_driver_with_poly_example():
+    queue_name = os.getenv("_ERT_TESTS_DEFAULT_QUEUE_NAME")
     with open("poly.ert", mode="a+", encoding="utf-8") as f:
-        f.write("QUEUE_SYSTEM TORQUE\nNUM_REALIZATIONS 2")
+        f.write("QUEUE_SYSTEM TORQUE\nNUM_REALIZATIONS 2\n")
+        if queue_name is not None:
```
xjules, line 29:
```suggestion
        if (queue_name: = os.getenv("_ERT_TESTS_DEFAULT_QUEUE_NAME")) is not None:
```

xjules: Get one less line also for the other tests and queue_name is used only in the if statement.

berland: fixing!

tests/integration_tests/scheduler/test_openpbs_driver.py
```diff
@@ -19,12 +20,19 @@ def mock_openpbs(pytestconfig, monkeypatch, tmp_path):
     mock_bin(monkeypatch, tmp_path)
 
 
-@pytest.mark.timeout(30)
+@pytest.fixture()
```
xjules, line 23: Nice!

tests/integration_tests/scheduler/test_openpbs_driver.py
berland: used a fixture as well.

xjules (approved):
Nice job @berland :+1: :rocket:

PR #7407, opened 2024-03-08
Add more tests of storage migration

eivindjahren:
Adds test for #7401 and #7403 

## When applicable


tests/integration_tests/test_storage_migration.py
```diff
@@ -39,6 +41,39 @@ def test_that_storage_matches(tmp_path, source_root, snapshot, monkeypatch):
...
+        # "5.0.8",
+        # "5.0.7",
+        # "5.0.6",
+        # "5.0.5",
+        # "5.0.4",
+        # "5.0.2",
+        # "5.0.1",
+        # "5.0.0",
     ]:
         print(ert_version)
```
DanSava, line 78: Is this print required?

DanSava (approved):
LGTM :medal_military:

tests/integration_tests/test_storage_migration.py
```diff
@@ -59,3 +95,19 @@ def test_that_storage_matches(tmp_path, source_root, snapshot, monkeypatch):
...
+
+            summary_data = ensemble.load_responses(
+                "summary",
+                tuple(ensemble.get_realization_list_with_responses("summary")),
+            )
+            snapshot.assert_match(
+                summary_data.to_dataframe().astype(np.float32).to_csv(),
+                "summary_data",
+            )
+            snapshot.assert_match_dir(
```
yngve-sk, line 107: I think this should be made somehow to not enforce the file structure, but just that the contents of the observations are parsed in correctly. Enforcing the directory structure of observations to be the same will make it impossible to change it away from one observation `.nc` file per observation specification, which currently slows down things quite a bit

tests/integration_tests/test_storage_migration.py
```diff
@@ -59,3 +95,19 @@ def test_that_storage_matches(tmp_path, source_root, snapshot, monkeypatch):
             snapshot.assert_match(
                 str(experiment.response_configuration) + "\n", "responses"
             )
+
+            summary_data = ensemble.load_responses(
+                "summary",
+                tuple(ensemble.get_realization_list_with_responses("summary")),
+            )
+            snapshot.assert_match(
+                summary_data.to_dataframe().astype(np.float32).to_csv(),
```
yngve-sk, line 104: Is the ordering the summary data meant to be locked in? If not, could you add a sorting to the dataframe to make the test insensitive to row order? `.transform(np.sort)` or something liek that. With the way summary data is used now, AFAIK the order does not matter, only the contents

yngve-sk (changes requested):
overall looks good, but some minor comments

tests/integration_tests/test_storage_migration.py
eivindjahren: It only enforces that the dictionary remains the same. Any change to this can just fix up the snapshot. It does not match against any `.nc` files just the values in the dictionary.

yngve-sk (approved):
LGTM

PR #7420, opened 2024-03-11
Have OpenPBS driver use qstat -w option

### LINKED ISSUE #7404: use qstat -w for wide format
The job id is truncated to 17 or 22 characters without the -w option to qstat.

With -w, it is always 30. And 30 characters for a job id ought to be enough for everybody.

We should use -w as it will give us predictable output.

jonathan-eq:
**Issue**
Resolves #7404 
Using the -w option removes the need for the _expand_truncated_jobids method, as the job id is not truncated anymore. 

**Approach**
Added the -w flag in both the driver and the mocked qstat script.

## When applicable

### DIFF
--- a/src/ert/scheduler/openpbs_driver.py
+++ b/src/ert/scheduler/openpbs_driver.py
@@ -304,7 +304,6 @@ class OpenPBSDriver(Driver):
                     "qstat",
                     "-fx",
                     "-Fjson",
-                    "-w",  # wide format
                     *self._finished_job_ids,
                     stdout=asyncio.subprocess.PIPE,
                     stderr=asyncio.subprocess.PIPE,
--- a/src/ert/scheduler/openpbs_driver.py
+++ b/src/ert/scheduler/openpbs_driver.py
@@ -248,19 +248,6 @@ class OpenPBSDriver(Driver):
         self._iens2jobid[iens] = job_id_
         self._non_finished_job_ids.add(job_id_)
 
-    def _expand_truncated_jobids(
-        self, truncated_jobs: Dict[str, Dict[str, Dict[str, str]]]
-    ) -> Dict[str, Dict[str, Dict[str, str]]]:
-        """Job ids gotten through normal qstat are truncated at length 16"""
-        data = {}
-        for truncated_job_id, job_state in truncated_jobs["Jobs"].items():
-            full_job_ids = [
-                job_id for job_id in self._jobs if job_id.startswith(truncated_job_id)
-            ]
-            assert len(full_job_ids) == 1
-            data[full_job_ids[0]] = job_state
-        return {"Jobs": data}
-
     async def kill(self, iens: int) -> None:
         if iens not in self._iens2jobid:
             logger.error(f"PBS kill failed due to missing jobid for realization {iens}")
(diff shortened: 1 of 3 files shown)


tests/unit_tests/scheduler/test_openpbs_driver.py
```diff
@@ -203,20 +203,22 @@ async def test_cluster_label():
...
+"Job id            Name             User              Time Use S Queue\n----------------  ---------------- ----------------  -------- - -----"
+formatter = "%-30s %-15s %-15s %-8s %-1s %-5s"
 
 
 @pytest.mark.parametrize(
     "qstat_script, started_expected",
     [
         pytest.param(
-            f"echo '{QSTAT_HEADER}1 foo someuser 0 R normal'; exit 0",
+            f"echo '{QSTAT_HEADER}';printf '{formatter}' 1 foo someuser 0 R normal; exit 0",
```
jonathan-eq, line 214: Is there a better way of having it formatted the same way as the actual qstat command with -w set?

src/ert/scheduler/openpbs_driver.py
```diff
@@ -303,6 +287,7 @@ async def poll(self) -> None:
                     "qstat",
                     "-fx",
                     "-Fjson",
+                    "-w",  # wide format
```
berland, line 290: This is not relevant here when we have the json format, it will not be truncated.

tests/integration_tests/scheduler/bin/qstat.py
```diff
@@ -67,8 +67,9 @@ def main() -> None:
             user = "mock"
             time = "00:00:00"
             queue = "mocked"
+
             print(
-                f"{job:16.16}  {name:16.16} {user:16.16}  {time:8.8} {state:1.1} {queue:5.5}"
+                f"{job:30.30}  {name:15.15} {user:15.15}  {time:8.8} {state:1.1} {queue:5.5}"
```
berland, line 72: so "wide" only applies to job-name, but name and user will be shortened.. 😞

jonathan-eq: But maybe that is okay, do we even use those fields for anything?

berland: It is okay yes :) To be solved later if a problem appears.

tests/unit_tests/scheduler/test_openpbs_driver.py
```diff
@@ -203,20 +203,22 @@ async def test_cluster_label():
     "Job id            Name             User              Time Use S Queue\n"
     "----------------  ---------------- ----------------  -------- - -----\n"
 )
+"Job id            Name             User              Time Use S Queue\n----------------  ---------------- ----------------  -------- - -----"
```
berland, line 206: The header on line 203 and 204 is the short header, isn't it? Should we include that in the test or should we skip it?

jonathan-eq: I dont think it matters as it is being skipped anyways.

tests/unit_tests/scheduler/test_openpbs_driver.py
berland: Since QSTAT_HEADER is the short header, this test is generating an impossible qstat output, isn't it? (our code probably handles that fine, but it is not an important feature to support, so we don't need to test that weird combination - just change QSTAT_HEADER to the wide format).

jonathan-eq: I ran `qstat -w` while having an active job, and copied the header directly from there.

berland (approved):
Great!

Issue #7408, opened 2024-03-08
Implement a test which correctly handles exception when job.__call__ fails


xjules:
This should include capturing also driver related exceptions.

PR #7421, opened 2024-03-11
Have scheduler handle exceptions in Job.__call__()

jonathan-eq:
**Issue**
Resolves [#7408](https://github.com/equinor/ert/issues/7408)
Potentially resolves https://github.com/equinor/ert/issues/7453

**Approach**
Adds a unit test `test_that_job__call__exceptions_are_propagated`, and makes `scheduler.execute()` handle exceptions in `job.__call__()`.

## When applicable

### DIFF
--- a/src/ert/scheduler/scheduler.py
+++ b/src/ert/scheduler/scheduler.py
@@ -6,11 +6,20 @@ import logging
 import os
 import ssl
 import time
+import traceback
 from collections import defaultdict
 from contextlib import suppress
 from dataclasses import asdict
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Dict, MutableMapping, Optional, Sequence
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Iterable,
+    MutableMapping,
+    Optional,
+    Sequence,
+)
 
 from pydantic.dataclasses import dataclass
 from websockets import Headers
@@ -72,7 +81,7 @@ class Scheduler:
         ee_token: Optional[str] = None,
     ) -> None:
         self.driver = driver
-        self._tasks: MutableMapping[int, asyncio.Task[None]] = {}
+        self._job_tasks: MutableMapping[int, asyncio.Task[None]] = {}
 
         self.submit_sleep_state: Optional[SubmitSleeper] = None
         if submit_sleep > 0:
@@ -107,8 +116,17 @@ class Scheduler:
 
     async def cancel_all_jobs(self) -> None:
         self._cancelled = True
-        for task in self._tasks.values():
+        await self._cancel_job_tasks()
+
+    async def _cancel_job_tasks(self) -> None:
+        for task in self._job_tasks.values():
             task.cancel()
+        with suppress(asyncio.TimeoutError):
+            await asyncio.wait(
+                self._job_tasks.values(),
+                timeout=1.0,
+                return_when=asyncio.ALL_COMPLETED,
+            )
 
     async def _update_avg_job_runtime(self) -> None:
         while True:
@@ -124,7 +142,7 @@ class Scheduler:
     ) -> None:
         while True:
             if self._completed_jobs_num >= minimum_required_realizations:
-                for iens, task in self._tasks.items():
+                for iens, task in self._job_tasks.items():
                     if (
                         self._jobs[iens].running_duration
                         > long_running_factor * self._average_job_runtime
@@ -138,7 +156,7 @@ class Scheduler:
         self._jobs[realization.iens] = Job(self, realization)
 
     def is_active(self) -> bool:
-        return any(not task.done() for task in self._tasks.values())
+        return any(not task.done() for task in self._job_tasks.values())
 
     def count_states(self) -> Dict[JobState, int]:
         counts: Dict[JobState, int] = defaultdict(int)
@@ -174,6 +192,44 @@ class Scheduler:
         for job in self._jobs.values():
             self._update_jobs_json(job.iens, job.real.run_arg.runpath)
 
+    async def _monitor_and_handle_tasks(
+        self, scheduling_tasks: list[asyncio.Task[None]]
+    ) -> None:
+        pending: Iterable[asyncio.Task[None]] = (
+            list(self._job_tasks.values()) + scheduling_tasks
+        )
+
+        while True:
+            done, pending = await asyncio.wait(
+                pending,
+                return_when=asyncio.FIRST_COMPLETED,
+            )
+            for task in done:
+                if task.cancelled():
+                    continue
+                if task_exception := task.exception():
+                    exc_traceback = "".join(
+                        traceback.format_exception(
+                            None, task_exception, task_exception.__traceback__
+                        )
+                    )
+                    logger.error(
+                        (
+                            f"Exception in scheduler task: {task_exception}\n"
+                            f"Traceback: {exc_traceback}"
+                        )
+                    )
+                    if task in scheduling_tasks:
+                        await self._cancel_job_tasks()
+                        raise task_exception
+            if not self.is_active():
+                for task in self._job_tasks.values():
+                    if task.cancelled():
+                        continue
+                    if task_exception := task.exception():
+                        raise task_exception
+                return
+
     async def execute(
         self,
         min_required_realizations: int = 0,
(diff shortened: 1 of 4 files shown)


src/ert/scheduler/scheduler.py
```diff
@@ -214,35 +214,36 @@ async def gather_realization_jobs() -> list[BaseException | None]:
...
             # there are two types of tasks and each type is handled differently:
             # -`gather_realization_jobs` does not raise; rather, each job task's result
             # is collected into the returning list. Exceptions from the job tasks are
             # handled in the `else` branch after the evaluation has stopped.
             # -If exception occurs, it must necessarily come from `scheduling tasks`
-            job_results = (
-                await asyncio.gather(gather_realization_jobs(), *scheduling_tasks)
-            )[0]
+
+            await asyncio.gather(gather_realization_jobs(), *scheduling_tasks)
```
jonathan-eq, line 224: This will raise an asyncio.CancelledError because of line 215. We can instead find all the job_results by looking directly at the tasks after gather-gather has finished.

src/ert/scheduler/scheduler.py
```diff
@@ -214,35 +214,36 @@ async def gather_realization_jobs() -> list[BaseException | None]:
...
-                logger.error(str(e))
-                raise e
-        else:
-            for result in job_results or []:
-                if not isinstance(result, asyncio.CancelledError) and isinstance(
-                    result, Exception
-                ):
-                    logger.error(str(result))
-                    raise result
+            if isinstance(e, asyncio.CancelledError):
```
jonathan-eq, line 227: I think the different branches in this except block should be refactored into separate functions.

src/ert/scheduler/scheduler.py
```diff
@@ -214,35 +214,36 @@ async def gather_realization_jobs() -> list[BaseException | None]:
...
+                    ):
+                        logger.error(str(result))
+                        raise result
+            else:
+                for job_task in self._tasks.values():
+                    job_task.cancel()
+                    # suppress potential error during driver.kill
+                    with suppress(Exception):
+                        await job_task
+                for scheduling_task in scheduling_tasks:
```
jonathan-eq, line 241: Do we need to cancel the scheduling_tasks again? I assume it is already done on line 215.

src/ert/scheduler/scheduler.py
```diff
@@ -214,35 +214,36 @@ async def gather_realization_jobs() -> list[BaseException | None]:
...
-                # suppress potential error during driver.kill
-                with suppress(Exception):
-                    await job_task
-            for scheduling_task in scheduling_tasks:
-                scheduling_task.cancel()
-            # Log and re-raise non-cancellation errors.
-            if not isinstance(e, asyncio.CancelledError):
-                logger.error(str(e))
-                raise e
-        else:
```
jonathan-eq, line 239: I think this else statement might be unreachable.

src/ert/scheduler/scheduler.py
```diff
@@ -214,35 +214,37 @@ async def gather_realization_jobs() -> list[BaseException | None]:
...
-            for result in job_results or []:
-                if not isinstance(result, asyncio.CancelledError) and isinstance(
-                    result, Exception
-                ):
-                    logger.error(str(result))
-                    raise result
+            if isinstance(e, asyncio.CancelledError):
+                for result in self._tasks.values() or []:
+                    try:
+                        await result
```
xjules, line 230: Not sure here. I think if cancellation error is triggered this might hang or ?

src/ert/scheduler/scheduler.py
xjules: I think so, yes. When exception happens if one of the `scheduling_tasks` then the rest will still run and thus we need to cancel.

src/ert/scheduler/scheduler.py
jonathan-eq: The cancellation error should be due to tasks finishing, and then finally killing scheduling_tasks. At that point, the job tasks should be done

src/ert/scheduler/scheduler.py
```diff
@@ -211,8 +219,8 @@ async def gather_realization_jobs() -> list[BaseException | None]:
                     *self._tasks.values(), return_exceptions=True
                 )
             finally:
-                for scheduling_task in scheduling_tasks:
-                    scheduling_task.cancel()
+                self._scheduling_tasks_running = False
```
xjules, line 222:
```suggestion
                self._scheduling_tasks_running = False
                self.driver._is_polling = False
                for scheduling_task in scheduling_tasks:
                     try:
                            await asyncion.wait_for(shceduling_task, timeout=2)
                     except asyncio.TimeoutError:
                               scheduling_task.cancel()
```

jonathan-eq: This change should happen in #7438

jonathan-eq:
Also, PRs like this one; should they be labeled `release-notes:skip` or `release-notes:misc`?

xjules:
> Also, PRs like this one; should they be labeled `release-notes:skip` or `release-notes:misc`?

`release-notes:skip`

jonathan-eq:
I will let @xjules take over this PR.

src/ert/scheduler/scheduler.py
```diff
@@ -107,8 +115,17 @@ def kill_all_jobs(self) -> None:
...
     async def cancel_all_jobs(self) -> None:
         self._cancelled = True
+        await self._cancel_tasks()
+
+    async def _cancel_tasks(self) -> None:
         for task in self._tasks.values():
             task.cancel()
+        with suppress(asyncio.TimeoutError):
+            await asyncio.wait(
+                list(self._tasks.values()),
```
jonathan-eq, line 125: Do we need to make it a list, or can we use the view directly?

src/ert/scheduler/scheduler.py
```diff
@@ -174,6 +191,32 @@ def add_dispatch_information_to_jobs_file(self) -> None:
         for job in self._jobs.values():
             self._update_jobs_json(job.iens, job.real.run_arg.runpath)
 
+    async def monitor_tasks(self, scheduling_tasks: list[asyncio.Task[None]]) -> None:
+        pending: Iterable[asyncio.Task[None]] = (
+            list(self._tasks.values()) + scheduling_tasks
+        )
+
+        while True:
+            done, pending = await asyncio.wait(
```
jonathan-eq, line 200: is `pending` used, or can we change it to `_` if it is unused?

src/ert/scheduler/scheduler.py
xjules: You might be right as the only requirement is that it's Iterable

src/ert/scheduler/scheduler.py
xjules: Yes, it's used in `asyncio.wait`

jonathan-eq: done, **pending** = await asyncio.wait(...)
Is this pending used?
We use done further down, but I don't see this pending being used anywhere.

jonathan-eq: Can we use `while len(pending) > 0` instead of `while True`?

xjules: Since `pending` contains also the scheduling_tasks, we can't use `pending == []`. There is `if not self.is_active():` which takes care of stopping this loop.

src/ert/scheduler/scheduler.py
```diff
@@ -107,8 +115,17 @@ def kill_all_jobs(self) -> None:
...
+
+    async def _cancel_tasks(self) -> None:
         for task in self._tasks.values():
             task.cancel()
+        with suppress(asyncio.TimeoutError):
+            await asyncio.wait(
+                self._tasks.values(),
+                timeout=1.0,
+                return_when=asyncio.ALL_COMPLETED,
+            )
```
jonathan-eq, line 128: Here, `_cancel_tasks(...)` is a very general method, and could be fine with just taking in a list of tasks to cancel instead of the class instance. Could we either make this a general function outside the class, or give it a different name? I think `_cancel_tasks` should be called `_cancel_all_jobs` if we keep it as a method.

src/ert/scheduler/scheduler.py
```diff
@@ -202,48 +247,11 @@ async def execute(
...
-            job_results = (
-                await asyncio.gather(gather_realization_jobs(), *scheduling_tasks)
-            )[0]
-        except (asyncio.CancelledError, Exception) as e:
-            for job_task in self._tasks.values():
-                job_task.cancel()
-                # suppress potential error during driver.kill
-                with suppress(Exception):
-                    await job_task
+            await self._monitor_and_handle_tasks(scheduling_tasks)
```
jonathan-eq, line 251: I like the name of this method!

jonathan-eq:
Can we test this manually with poly_example to make sure it works?

src/ert/scheduler/scheduler.py
```diff
@@ -174,6 +191,34 @@ def add_dispatch_information_to_jobs_file(self) -> None:
         for job in self._jobs.values():
             self._update_jobs_json(job.iens, job.real.run_arg.runpath)
 
+    async def _monitor_and_handle_tasks(
+        self, scheduling_tasks: list[asyncio.Task[None]]
+    ) -> None:
+        pending: Iterable[asyncio.Task[None]] = (
+            list(self._tasks.values()) + scheduling_tasks
```
berland, line 198: We should rename self._tasks to something more specific, as it is "one-asyncio-task-per-job" we are talking about, not any asyncio task.

berland: self._job_tasks would probably do

src/ert/scheduler/scheduler.py
```diff
@@ -174,6 +191,34 @@ def add_dispatch_information_to_jobs_file(self) -> None:
...
+        while True:
+            done, pending = await asyncio.wait(
+                pending,
+                return_when=asyncio.FIRST_COMPLETED,
+            )
+            for task in done:
+                if task.cancelled():
+                    continue
+                if task_exception := task.exception():
+                    logger.error(task_exception)
```
berland, line 210: Can we include the traceback also?

It is easy enough to find my "1/0" string when testing and seeing "division by zero" in the logs, but there will be more difficult scenarios to come.

berland:
I found that adding `1/0` to `LocalDriver.__init__` will not bring down Ert (it will hang), but the exception is logged so not not a biggie.

src/ert/scheduler/scheduler.py
xjules: I'd change it to `_cancel_all_jobs` since currently we don't need to generalize this function but if there is a need to we can create one :)

xjules:
> I found that adding `1/0` to `LocalDriver.__init__` will not bring down Ert (it will hang), but the exception is logged so not not a biggie.

This is due to that the scheduler has not even started as we create driver beforehand `driver = create_driver(self._queue_config)` and just inject the instantiated driver into the sceduler.

PR #7428, opened 2024-03-12
Deprecate unused lsf options

### LINKED ISSUE #7203: Deprecate LSF options LSF_LOGIN_SHELL, LSF_SERVER, LSF_RSH_CMD
The remote shell feature is not meant to be ported to the Python LSF driver.

`ert-configurations` currently set `LSF_SERVER` to some ssh-able machine, and thus this is the dominant code path used in the current C driver.

Some komodo release must change this to activate the local shell before we move to the Python driver (that will only use the local shell). This change should be done with a warning, and for users getting into trouble, it should be possible for them to set LSF_SERVER in their config to reinstate the old behaviour.

Definition of done:
* ERT default is to use local shell for bsub/bjobs/bkill.
* Deprecation warnings are emitted while parsing the config, stating that these keywords have no effect and can safely be removed.
* Removal from documentation

The functionality will be gone from ERT whenever the legacy C LSF driver is purged.

jonathan-eq:
**Issue**
Resolves #7203 

**Approach**

## When applicable

### DIFF
--- a/src/ert/config/parsing/config_schema_deprecations.py
+++ b/src/ert/config/parsing/config_schema_deprecations.py
@@ -177,4 +177,20 @@ deprecated_keywords_list = [
         "for the Ensemble Smoother update algorithm. "
         "Please use ENKF_ALPHA and STD_CUTOFF keywords instead.",
     ),
+    DeprecationInfo(
+        keyword="QUEUE_OPTION",
+        message="LSF_LOGIN_SHELL as QUEUE_OPTION to the LSF system will be removed in "
+        "the future, and it is not recommended to use this QUEUE_OPTION. "
+        "It has been used in the past to force the bsub command to use a "
+        "specific shell. The current ERT default is to use local shell.",
+        check=lambda line: "LSF_LOGIN_SHELL" in line,
+    ),
+    DeprecationInfo(
+        keyword="QUEUE_OPTION",
+        message="LSF_RSH_CMD as QUEUE_OPTION to the LSF system will be removed in "
+        "the future, and it is not recommended to use this QUEUE_OPTION. "
+        "It has been used in the past to set the remote shell command. "
+        "The ERT default is to use /usr/bin/ssh.",
+        check=lambda line: "LSF_RSH_CMD" in line,
+    ),
 ]
--- a/src/ert/config/parsing/config_schema_item.py
+++ b/src/ert/config/parsing/config_schema_item.py
@@ -38,7 +38,7 @@ class SchemaItem:
     # A list of item's which must also be set (if this item is set). (can be NULL)
     required_children: List[str] = Field(default_factory=list)
     # Information about the deprecation if deprecated
-    deprecation_info: Optional[DeprecationInfo] = None
+    deprecation_info: List[DeprecationInfo] = Field(default_factory=list)
     # if positive, arguments after this count will be concatenated with a " " between
     join_after: Optional[PositiveInt] = None
     # if true, will accumulate many values set for key, otherwise each entry will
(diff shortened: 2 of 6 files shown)


src/ert/config/parsing/config_schema_deprecations.py
```diff
@@ -177,4 +177,25 @@
         "for the Ensemble Smoother update algorithm. "
         "Please use ENKF_ALPHA and STD_CUTOFF keywords instead.",
     ),
+    DeprecationInfo(
+        keyword="QUEUE_OPTION",
+        message="The LSF_LOGIN_SHELL queue option has been removed and no longer has any "
+        "effect. It has been used in the past to force the bsub command to use a "
```
berland, line 183: It has not been removed yet, it will probably have an effect if LSF_SERVER is configured (via ert-configurations or not) and the c-driver is used. 

I think we want to say that we don't want user to use this, because it will actually disappear in the future (when the C-driver is deleted)

src/ert/config/parsing/config_schema_deprecations.py
```diff
@@ -177,4 +177,25 @@
...
+        "specific shell. The current ERT default is to use local shell.",
+        check=lambda line: "LSF_LOGIN_SHELL" in line,
+    ),
+    DeprecationInfo(
+        keyword="QUEUE_OPTION",
+        message="The LSF_RSH_CMD queue option has been removed and no longer has any "
```
berland, line 189: same as above.

src/ert/config/parsing/config_schema_deprecations.py
```diff
@@ -177,4 +177,25 @@
...
+        "effect. It has been used in the past to set the remote shell command. "
+        "The ERT default is to use /usr/bin/ssh.",
+        check=lambda line: "LSF_RSH_CMD" in line,
+    ),
+    DeprecationInfo(
+        keyword="QUEUE_OPTION",
+        message="The LSF_SERVER queue option has been deprecated as it is already "
+        "being configured by ert-configurations. You should only use this queue option "
```
berland, line 197: We should not refer to `ert-configurations` in the ert source code, as outside of Equinor, "ert-configurations" does not exist.

Just say it is deprecated because it will become a no-op in the future

berland:
Looks good. Have you tested manually that these warnings actually pop out?

berland (dismissed):

berland:
It looks good, but we can hold merge until we know that removal of LSF_SERVER went well in komodo stable.

berland:
> Do you think removing the options from the docs can be done now, even though they are usable in the legacy driver?

For these two I think so yes. Approved by Øyvind.

src/ert/config/parsing/config_schema_item.py
```diff
@@ -38,7 +38,7 @@ class SchemaItem:
     # A list of item's which must also be set (if this item is set). (can be NULL)
     required_children: List[str] = Field(default_factory=list)
     # Information about the deprecation if deprecated
-    deprecation_info: Optional[DeprecationInfo] = None
+    deprecations_info: List[DeprecationInfo] = Field(default_factory=list)
```
xjules, line 41: Why plural?

jonathan-eq: Because a keyword can have multiple deprecations. In this context, QUEUE_OPTION is the keyword, and the deprecations can be both LSF_LOGIN_SHELL and LSF_RSH_CMD.

xjules: Yes, it is just nitpicking, but still think `deprecation_info` was just fine. But we leave it up to you.

xjules (approved):
Very nice job @jonathan-eq ! Could you squash the commits?

PR #7433, opened 2024-03-12
Have lsf_driver specify SIGKILL when bkilling

### LINKED ISSUE #7405: LSF: Specify POSIX signal when `bkill`ing
LSF's `bkill` will send `SIGINT`, `SIGTERM` and `SIGKILL` if the job doesn't finish. This is different from standard terminating behaviour where `SIGTERM` is sent at first, followed by `SIGKILL` if the program is still alive.

`SIGINT` is usually sent by the terminal on Ctrl-C, which makes programs interpret this to be user-driven termination request rather than system. Python in particular exposes this as `KeyboardInterrupt` error, which has caused confusion in the past.

OpenPBS' `qdel`, Slurm's `scancel` and LocalDriver's `kill()` do what is expected (SIGTERM, wait, SIGKILL), so LSF is the odd one out.

https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=reference-bkill

### LINKED ISSUE #7443: bkill uses SIGINT as first signal
**Describe the bug**
The users get confused when they see Keyboard Interrupt as the reason of a failed lsf job, when in reality it is bkill that uses this signal. TORQUE/OpenPBS and slurm skips SIGINT, and submits bkill with SIGTERM as the first signal.

**To reproduce**
1. `pip install ert`
2. `ert gui my_config.ert`
4. …

**Expected behaviour**

**Screenshots**

**Environment**
 - OS: [e.g. RHEL7]
 - ERT/Komodo release: [e.g. 2.18 or 2020.12.00]
 - Python version

**Additional context**

jonathan-eq:
This commit makes the lsf_driver use bkill with the SIGKILL signal instead of the default SIGINT.

**Issue**
Resolves #7443  

**Approach**

## When applicable

### DIFF
--- a/src/clib/lib/job_queue/lsf_driver.cpp
+++ b/src/clib/lib/job_queue/lsf_driver.cpp
@@ -735,15 +735,24 @@ void lsf_driver_kill_job(void *_driver, void *_job) {
         char **argv = (char **)calloc(2, sizeof *argv);
         CHECK_ALLOC(argv);
         argv[0] = driver->remote_lsf_server;
-        argv[1] = saprintf("%s %s", driver->bkill_cmd, job->lsf_jobnr_char);
+        argv[1] = saprintf("%s %s %s", driver->bkill_cmd, "-s SIGKILL",
+                           job->lsf_jobnr_char);
 
         spawn_blocking(driver->rsh_cmd, 2, (const char **)argv, NULL, NULL);
 
         free(argv[1]);
         free(argv);
     } else if (driver->submit_method == LSF_SUBMIT_LOCAL_SHELL) {
-        spawn_blocking(driver->bkill_cmd, 1,
-                       (const char **)&job->lsf_jobnr_char, NULL, NULL);
+        char **argv = (char **)calloc(3, sizeof *argv);
+        CHECK_ALLOC(argv);
+        argv[0] = saprintf("%s", "-s");
+        argv[1] = saprintf("%s", "SIGKILL");
+        argv[2] = saprintf("%s", job->lsf_jobnr_char);
+        spawn_blocking(driver->bkill_cmd, 3, (const char **)argv, NULL, NULL);
+        free(argv[0]);
+        free(argv[1]);
+        free(argv[2]);
+        free(argv);
     }
 }
 
--- a/src/ert/scheduler/lsf_driver.py
+++ b/src/ert/scheduler/lsf_driver.py
@@ -190,6 +190,8 @@ class LsfDriver(Driver):
         logger.debug(f"Killing realization {iens} with LSF-id {job_id}")
         process = await asyncio.create_subprocess_exec(
             self._bkill_cmd,
+            "-s",
+            "SIGKILL",
             job_id,
             stdout=asyncio.subprocess.PIPE,
             stderr=asyncio.subprocess.PIPE,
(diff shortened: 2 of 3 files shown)


berland:
If the mother ERT process is terminated (say an eager user hits ctrl-c), this signal will be propagated and kill this subprocess, we might not want that, we want this script with bkills live happily independent of mother Ert.

I think you want to do as here:
https://github.com/equinor/ert/blob/2b180e98fe0f8ad1e6eda48daa6a9d9d87c6f0ec/src/ert/scheduler/local_driver.py#L70
in order to avoid this.

src/ert/scheduler/lsf_driver.py
```diff
@@ -149,9 +149,12 @@ async def kill(self, iens: int) -> None:
         job_id = self._iens2jobid[iens]
 
         logger.debug(f"Killing realization {iens} with LSF-id {job_id}")
+
+        bkill_cmd_with_signals = f"{self._bkill_cmd} -s SIGTERM {job_id}; sleep 30; {self._bkill_cmd} -s SIGKILL {job_id};"
```
berland, line 153: If we have tests asserting the output of this bkill (I have not checked), these tests will take 30 seconds I suppose?

jonathan-eq: I will put the sleep period as an instance variable. Or do you think it would be better as a class variable? I see we already use both in lsf_driver.

jonathan-eq:
Should I change this in the job queue driver too?

berland:
> Should I change this in the job queue driver too?

Yes I think that makes sense.

src/clib/lib/job_queue/lsf_driver.cpp
```diff
@@ -735,15 +735,18 @@ void lsf_driver_kill_job(void *_driver, void *_job) {
...
         spawn_blocking(driver->rsh_cmd, 2, (const char **)argv, NULL, NULL);
 
         free(argv[1]);
         free(argv);
     } else if (driver->submit_method == LSF_SUBMIT_LOCAL_SHELL) {
-        spawn_blocking(driver->bkill_cmd, 1,
-                       (const char **)&job->lsf_jobnr_char, NULL, NULL);
+        spawn_blocking(
+            driver->bkill_cmd, 1,
+            (const char **)saprintf("%s %s", "-s SIGTERM", job->lsf_jobnr_char),
```
berland, line 748: I am willing to bet that this won't work (do not assume there are automated tests hitting this line).  `saprintf` is probably returning a `const char *` and not `const char **`.

pinkwah:
This is not the correct behaviour. There needs to be a `SIGKILL` after a while, otherwise a hanging job will hang forever taking up resources.

berland:
> This is not the correct behaviour. There needs to be a `SIGKILL` after a while, otherwise a hanging job will hang forever taking up resources.

Do we expect the job to react differently to SIGTERM than SIGINT? The job can choose to ignore both. I am thinking that changing to SIGTERM solves the confusion for the users, and that sending an additional SIGKILL later is a new feature.

jonathan-eq:
I updated this to do SIGKILL instead of SIGTERM.

xjules (approved):
I think it looks good now! Good job @jonathan-eq ! :rocket:

jonathan-eq:
> Have you performed a manual test of the "kill all realizations" button in the GUI with this change? (or some other real-world test of the changed C-code?)

It does indeed NOT work. 
```
 SIGKILL 785865: Illegal signal value
 SIGKILL 785866: Illegal signal value
 SIGKILL 785867: Illegal signal value
 SIGKILL 785868: Illegal signal value
 SIGKILL 785869: Illegal signal value
 SIGKILL 785870: Illegal signal value
 SIGKILL 785871: Illegal signal value
 SIGKILL 785872: Illegal signal value
 SIGKILL 785873: Illegal signal value

```

xjules (approved):
Very good! :rocket:

PR #7441, opened 2024-03-13
Implement LSF_RESOURCE and EXCLUDE_HOST queue option

### LINKED ISSUE #7345: Implement LSF_RESOURCE queue option for the LSF_driver
See the old JobQueue usage.

jonathan-eq:
**Issue**
Resolves #7345 

**Approach**
This PR makes scheduler lsf_driver use QUEUE_OPTION LSF_RESOURCE as qsub resource_request option.
It also adds the EXCLUDE_HOST queue option for LSF, as they were closely related in the legacy C driver.

## When applicable

### DIFF
--- a/src/ert/scheduler/__init__.py
+++ b/src/ert/scheduler/__init__.py
@@ -39,6 +39,7 @@ def create_driver(config: QueueConfig) -> Driver:
             bkill_cmd=queue_config.get("BKILL_CMD"),
             bjobs_cmd=queue_config.get("BJOBS_CMD"),
             queue_name=queue_config.get("LSF_QUEUE"),
+            resource_requirement=queue_config.get("LSF_RESOURCE"),
         )
     else:
         raise NotImplementedError("Only LOCAL, TORQUE and LSF drivers are implemented")
--- a/src/ert/scheduler/__init__.py
+++ b/src/ert/scheduler/__init__.py
@@ -38,6 +38,7 @@ def create_driver(config: QueueConfig) -> Driver:
             bsub_cmd=queue_config.get("BSUB_CMD"),
             bkill_cmd=queue_config.get("BKILL_CMD"),
             bjobs_cmd=queue_config.get("BJOBS_CMD"),
+            exclude_hosts=queue_config.get("EXCLUDE_HOST", "").split(","),
             queue_name=queue_config.get("LSF_QUEUE"),
             resource_requirement=queue_config.get("LSF_RESOURCE"),
         )
--- a/src/ert/scheduler/lsf_driver.py
+++ b/src/ert/scheduler/lsf_driver.py
@@ -16,6 +16,7 @@ from typing import (
     Mapping,
     MutableMapping,
     Optional,
+    Sequence,
     Tuple,
     Union,
     get_args,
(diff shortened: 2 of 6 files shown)


src/ert/scheduler/__init__.py
```diff
@@ -38,6 +38,8 @@ def create_driver(config: QueueConfig) -> Driver:
             bsub_cmd=queue_config.get("BSUB_CMD"),
             bkill_cmd=queue_config.get("BJOBS_CMD"),
             bjobs_cmd=queue_config.get("BJOBS_CMD"),
+            queue_name=queue_config.get("LSF_QUEUE"),
```
berland, line 41: Is it a bug that we did not pass in LSF_QUEUE already? (if so, it should probably be a separate PR, with some testing)

jonathan-eq: It seems like it. I will split this commit out into a new PR, and create an issue for it.

berland:
From `man bsub`:

```
       -R "res_req" [-R "res_req" ...]
                Runs the job on a host that meets the specified resource
                requirements. A resource requirement string describes
                the resources a job needs. LSF uses resource
                requirements to select hosts for job execution. Resource
                requirement strings can be simple (applying to the
                entire job), compound (applying to the specified number
                of slots), or alternative.

```
which means we should probably name the variable in the driver the same, "`resource_requirement`"

src/ert/scheduler/lsf_driver.py
```diff
@@ -150,10 +152,14 @@ async def submit(
             script_path = Path(script_handle.name)
         assert script_path is not None
         script_path.chmod(script_path.stat().st_mode | stat.S_IEXEC)
+        arg_resource_request = (
+            ["-R", f"'{self._resource_request}'"] if self._resource_request else []
```
berland, line 156: Is the single quoting here needed? Should it be responsibility of the user? Can the quotes even be injected via the config?

tests/unit_tests/scheduler/test_lsf_driver.py
```diff
@@ -130,6 +130,15 @@ async def test_submit_with_default_queue():
     assert "-q" not in Path("captured_bsub_args").read_text(encoding="utf-8")
 
 
+@pytest.mark.usefixtures("capturing_bsub")
+async def test_submit_with_resource_request():
```
berland, line 134: I think we can afford an integration test toward the real cluster also on this. Just use some low requirements that seems relevant and write test function similiar to test_submit in test_generic_driver that writes the string test to a file, and then assert that the string can be found.

jonathan-eq:
> From `man bsub`:
> 
> ```
>        -R "res_req" [-R "res_req" ...]
>                 Runs the job on a host that meets the specified resource
>                 requirements. A resource requirement string describes
>                 the resources a job needs. LSF uses resource
>                 requirements to select hosts for job execution. Resource
>                 requirement strings can be simple (applying to the
>                 entire job), compound (applying to the specified number
>                 of slots), or alternative.
> ```
> 
> which means we should probably name the variable in the driver the same, "`resource_requirement`"

It was named resource_request in the old driver, but I think it is a good improvement.

src/ert/scheduler/lsf_driver.py
jonathan-eq: It seems to do so in the old driver. Is LSF_SUBMIT_REMOTE_SHELL the submit method we will be using, or the one we are dropping with the scheduler?
```
char *quoted_resource_request = NULL;
    if (req) {
        if (driver->submit_method == LSF_SUBMIT_REMOTE_SHELL) {
            quoted_resource_request = saprintf("\"%s\"", req);
            free(req);
        } else
            quoted_resource_request = req;
    }
    return quoted_resource_request;
```

berland: We are dropping the remote shell feature yes.

If there can be spaces in resource requests, it should be tested in integration tests.

src/ert/scheduler/__init__.py
```diff
@@ -38,7 +38,9 @@ def create_driver(config: QueueConfig) -> Driver:
             bsub_cmd=queue_config.get("BSUB_CMD"),
             bkill_cmd=queue_config.get("BKILL_CMD"),
             bjobs_cmd=queue_config.get("BJOBS_CMD"),
+            exclude_hosts=queue_config.get("EXCLUDE_HOST", "").split(","),
```
xjules, line 41: Does it work?

jonathan-eq: Yes

xjules (approved):
Looks very nice! Let's ship it :rocket:

PR #7442, opened 2024-03-13
Only build & test using macos runners for tags

andreas-el:
Mitigation efforts after we learned of the pricing and usage of the GitHub macOS-runners.

### DIFF
--- a/.github/workflows/run_ert_test_data_setups.yml
+++ b/.github/workflows/run_ert_test_data_setups.yml
@@ -21,14 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ['3.8', '3.11', '3.12']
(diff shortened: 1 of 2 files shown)


eivindjahren (approved):
:money_with_wings:

.github/workflows/run_ert_test_data_setups.yml
```diff
@@ -22,13 +22,19 @@ jobs:
       matrix:
         python-version: ['3.8', '3.11', '3.12']
         os: [ubuntu-latest, macos-13, macos-latest-xlarge]
+        isPR:
```
eivindjahren, line 25: Since this seems to be problematic, let's just drop macos here all together.

PR #7444, opened 2024-03-13
Document how delete_directory works on symlinks

berland:
**Issue**
Resolves user confusion in https://app.slack.com/client/T02JL00JU/CDPJULWR4

**Approach**
Test and document

## When applicable

### DIFF
--- a/tests/unit_tests/shared/share/test_shell.py
+++ b/tests/unit_tests/shared/share/test_shell.py
@@ -284,6 +284,64 @@ def test_that_delete_directory_does_not_follow_symlinks(shell):
 
 
 @pytest.mark.usefixtures("use_tmpdir")
+def test_that_delete_directory_on_a_symlink_to_file_is_ignored(shell):
+    with open("link_target", "w", encoding="utf-8") as f:
+        f.write("hei")
+    shell.symlink("link_target", "link")
+    shell.delete_directory("link")
+    assert os.path.exists("link_target")
+    assert os.path.exists("link")
+
+
+@pytest.mark.usefixtures("use_tmpdir")
+def test_that_delete_directory_on_a_symlink_to_a_directory_only_deletes_link(shell):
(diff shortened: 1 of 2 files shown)


berland:
@sondreso you might be opinionated here.

It is debatable whether we want to allow the trailing slashes in the argument to DELETE_DIRECTORY, one option would be to just error hard (breaking change!)

berland:
And the tests uncover that this is again behaves different on MacOS.. 🤯

jonathan-eq:
LGTM, but let's see what @sondreso says.

> @sondreso you might be opinionated here.
> 
> It is debatable whether we want to allow the trailing slashes in the argument to DELETE_DIRECTORY, one option would be to just error hard (breaking change!)

sondreso:
I think we should keep the default behavior here and not fail hard if there is a trailing slash. But the section in the documentation should be clearly marked (With a warning box) about the consequences of doing so.

berland:
Documentation now looks like:
image

tests/unit_tests/shared/share/test_shell.py
```diff
@@ -283,6 +283,51 @@ def test_that_delete_directory_does_not_follow_symlinks(shell):
...
+    assert os.path.exists("link_target/file")
+    assert not os.path.exists("link")
+
+
+@pytest.mark.usefixtures("use_tmpdir")
+@pytest.mark.parametrize("trailing", ["", "/"])
+def test_that_delete_directory_on_a_symlink_to_a_directory_is_conditionally_ignored(
+    shell, trailing
+):
+    """A trailing slash turns the world upside down"""
```
xjules, line 312: :laughing: the comment could be more descriptive - maybe to detailed down both scenarios

berland: Written more now

PR #7482, opened 2024-03-20
Have BKILL signal SIGKILL after SIGTERM

### LINKED ISSUE #7405: LSF: Specify POSIX signal when `bkill`ing

jonathan-eq:
**Issue**
Resolves #7405 

**Approach**
This PR makes the LSF drivers spawn a new process to run `bkill -s SIGKILL` in a fire-and-forget manner.

## When applicable

### DIFF
--- a/src/clib/lib/job_queue/lsf_driver.cpp
+++ b/src/clib/lib/job_queue/lsf_driver.cpp
@@ -735,24 +735,42 @@ void lsf_driver_kill_job(void *_driver, void *_job) {
         char **argv = (char **)calloc(2, sizeof *argv);
         CHECK_ALLOC(argv);
         argv[0] = driver->remote_lsf_server;
-        argv[1] = saprintf("%s %s %s", driver->bkill_cmd, "-s SIGKILL",
+        argv[1] = saprintf("%s %s %s", driver->bkill_cmd, "-s SIGTERM",
                            job->lsf_jobnr_char);
-
         spawn_blocking(driver->rsh_cmd, 2, (const char **)argv, NULL, NULL);
-
         free(argv[1]);
         free(argv);
+
+        char **argv2 = (char **)calloc(2, sizeof *argv2);
+        argv2[0] = driver->remote_lsf_server;
+        argv2[1] = saprintf("%s %s %s %s %s", "sleep 30;", driver->bkill_cmd,
+                            "-s", "SIGKILL", job->lsf_jobnr_char);
+        spawn(driver->rsh_cmd, 2, (const char **)argv2, NULL, NULL);
+        free(argv2[0]);
+        free(argv2);
+
     } else if (driver->submit_method == LSF_SUBMIT_LOCAL_SHELL) {
         char **argv = (char **)calloc(3, sizeof *argv);
         CHECK_ALLOC(argv);
         argv[0] = saprintf("%s", "-s");
-        argv[1] = saprintf("%s", "SIGKILL");
+        argv[1] = saprintf("%s", "SIGTERM");
         argv[2] = saprintf("%s", job->lsf_jobnr_char);
         spawn_blocking(driver->bkill_cmd, 3, (const char **)argv, NULL, NULL);
         free(argv[0]);
         free(argv[1]);
         free(argv[2]);
         free(argv);
+
+        char **argv2 = (char **)calloc(2, sizeof *argv2);
+        CHECK_ALLOC(argv2);
+        argv2[0] = saprintf("%s", "-c");
+        argv2[1] = saprintf("%s %s %s %s %s", "sleep 30;", driver->bkill_cmd,
+                            "-s", "SIGKILL", job->lsf_jobnr_char);
+        spawn((const char *)saprintf("%s", "/bin/sh"), 2, (const char **)argv2,
+              "/dev/null", "/dev/null");
+        free(argv2[0]);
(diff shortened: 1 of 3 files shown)


jonathan-eq:
I wanted to use `preexec_fn=os.setpgrp` when spawning the process; to make sure it is not stopped if the main process quits. This was stopped by ruff, but maybe it is not needed.
```
src/ert/scheduler/lsf_driver.py:205:13: PLW1509 `preexec_fn` argument is unsafe when using threads
```

jonathan-eq:
BLOCKED BY #7433

xjules:
Have you actually tested that it works?

berland:
Use `ruff rule PLW1509` on the command line to get a further explanation.

jonathan-eq:
I am not sure if the C part works. It seems to be logging duplicate. I wonder what happens when I pass NULL to the stdout-arg/stderr-arg instead of nullptr or a file. Is it then redirected to stdout?

jonathan-eq:
> ruff rule PLW1509

They actually had a really good description!

src/clib/lib/job_queue/lsf_driver.cpp
```diff
@@ -735,24 +735,41 @@ void lsf_driver_kill_job(void *_driver, void *_job) {
         char **argv = (char **)calloc(2, sizeof *argv);
         CHECK_ALLOC(argv);
         argv[0] = driver->remote_lsf_server;
-        argv[1] = saprintf("%s %s %s", driver->bkill_cmd, "-s SIGKILL",
+        argv[1] = saprintf("%s %s %s", driver->bkill_cmd, "-s SIGTERM",
                            job->lsf_jobnr_char);
-
-        spawn_blocking(driver->rsh_cmd, 2, (const char **)argv, NULL, NULL);
-
```
berland, line 742: I don't get why this `spawn_blocking` is removed, what use is there of  `argv` when you just `free` it?

src/clib/lib/job_queue/lsf_driver.cpp
```diff
@@ -735,24 +735,41 @@ void lsf_driver_kill_job(void *_driver, void *_job) {
...
         free(argv[1]);
         free(argv);
+
+        char **argv2 = (char **)calloc(2, sizeof *argv2);
```
berland, line 743: Maybe spend some keystrokes on saying what `argv2` is for, like "argv_second_kill"

src/clib/lib/job_queue/lsf_driver.cpp
```diff
@@ -735,24 +735,41 @@ void lsf_driver_kill_job(void *_driver, void *_job) {
...
+        argv2[0] = driver->remote_lsf_server;
```
berland, line 744: I have not checked what CHECK_ALLOC is doing, but should it be included here as well?

pinkwah: `CHECK_ALLOC` was something that was introduced because that's what `util_malloc` did. There is no reason for it to exist.

src/clib/lib/job_queue/lsf_driver.cpp
pinkwah: A refactor of this area of code is pending: https://github.com/equinor/ert/pull/7515/files

Technically, it's enough to remove the `char **` from `char **argv2`. But it's possible to improve the readability of this code quite easily by writing C++, so I recommend we ignore this until the aforementioned PR.

src/clib/lib/job_queue/lsf_driver.cpp
jonathan-eq: You are right, this was removed by accident

tests/unit_tests/scheduler/test_lsf_driver.py
```diff
@@ -267,20 +267,34 @@ async def test_kill(
...
+    driver._sleep_time_between_bkills = 0
+
     await driver.kill(iens_to_kill)
+
+    async def wait_for_sigkill_in_file():
+        while True:
+            bkill_args = (
+                Path("bkill_args").read_text(encoding="utf-8").strip().split("\n")
+            )
+            if f"-s SIGKILL {mocked_iens2jobid[iens_to_kill]}" in bkill_args:
```
xjules, line 287: Not sure if this a good way to do it - it might just never return if the correct jobid is not present. What about just assert / return true/false for the jobid after we got SIGGILL in the file?

jonathan-eq: Yes, I will change it.

xjules (approved):
Looks good now! :rocket:

PR #7484, opened 2024-03-20
Add a test for opening empty storage

### LINKED ISSUE #7045: open_storage on existing empty directory attempts to migrate and fails
Doing `open_storage("/tmp/empty_directory", mode="w")` results in

```
WARNING ert.storage.local_storage:local_storage.py:235 Migrating storage at /tmp/empty_directory failed with Unknown storage version
ERROR ert.storage.local_storage:local_storage.py:357 Migrating storage at /tmp/empty_directory failed with Use 'local_storage_set_ert_config' before retrieving the config
```
This exposes two failures:

* Attempting to migrate an empty directory will always fail
* One migration assumes config is available which might not always be the case: `'Use 'local_storage_set_ert_config' before retrieving the config`

This can be reproduced with
```python


def test_open_storage_failure(tmp_path, caplog):
    with open_storage(tmp_path, mode="w") as f:
        _ = f.create_experiment()
    assert len(caplog.messages) == 0
```

frode-aarstad:
**Issue**
Resolves #7045 

**Approach**
There are no errors in the log anymore when opening an empty storage. 

Added a test to test this behavior

## When applicable

### DIFF
--- a/src/ert/storage/local_storage.py
+++ b/src/ert/storage/local_storage.py
@@ -374,6 +374,7 @@ class LocalStorage(BaseMode):
 
 def _storage_version(path: Path) -> Optional[int]:
     if not path.exists():
+        logger.warning(f"Unknown storage version in '{path}'")
         return None
     try:
         with open(path / "index.json", encoding="utf-8") as f:
--- a/tests/unit_tests/storage/test_local_storage.py
+++ b/tests/unit_tests/storage/test_local_storage.py
@@ -511,3 +511,29 @@ class StatefulStorageTest(RuleBasedStateMachine):
 
 
 TestStorage = StatefulStorageTest.TestCase
+
+
+def test_open_storage_write_with_empty_directory(tmp_path, caplog):
+    with open_storage(tmp_path / "storage", mode="w") as storage:
+        _ = storage.create_experiment()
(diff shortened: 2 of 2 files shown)


dafeda:
What do you think is reasonable behaviour if we try opening a storage with an empty directory in read mode?

So, something like this:

```python
    with open_storage(tmp_path / "storage", mode="r") as storage:
        print(storage.experiments)
```

Not sure if this is relevant here though.

frode-aarstad:
> What do you think is reasonable behaviour if we try opening a storage with an empty directory in read mode?
> 
> So, something like this:
> 
> ```python
>     with open_storage(tmp_path / "storage", mode="r") as storage:
>         print(storage.experiments)
> ```
> 
> Not sure if this is relevant here though.

Might be worth making a separate test for read-only. I would expect the behavior to be the same. IE  `assert "Unknown storage version in" in caplog.messages[0]` should be true

dafeda:
> Might be worth making a separate test for read-only. I would expect the behavior to be the same. IE `assert "Unknown storage version in" in caplog.messages[0]` should be true

Seems reasonable to me to test both modes in the same test.
I'm fine either way though.

dafeda (approved):

PR #7490, opened 2024-03-20
Unify returncode values for different drivers when process is killed by a signal

JHolba:
This should fix a failing test in azure bleeding.

**Approach**
Made sure all drivers return 128+signal when killed by a signal.
Updated some error handling to reflect this.

## When applicable


src/ert/scheduler/lsf_driver.py
```diff
@@ -24,10 +24,13 @@
 from pydantic import BaseModel, Field
 from typing_extensions import Annotated
 
-from ert.scheduler.driver import Driver
+from ert.scheduler.driver import SIGNAL_OFFSET, Driver
 from ert.scheduler.event import Event, FinishedEvent, StartedEvent
 
 _POLL_PERIOD = 2.0  # seconds
+LSF_FAILED_JOB: Annotated[int, "Return code we use when lsf reports failed jobs"] = (
+    SIGNAL_OFFSET + 32
```
pinkwah, line 32: Is this considered a signal or is it a returncode?

According to the relevant manpage[^1], signal `32` is the first "real-time" user-defined signal (`SIGRTMIN`), but `glibc` uses the first two for pthreads. This means that `import signal; signal.SIGRTMIN` is `34` on my machine. So, is LSF raising a signal (that is actually used by pthread) or is this unrelated and is just the returncode?

[^1]: https://www.man7.org/linux/man-pages/man7/signal.7.html

src/ert/scheduler/openpbs_driver.py
```diff
@@ -310,14 +310,12 @@ async def _process_job_update(self, job_id: str, new_state: AnyJob) -> None:
...
             assert new_state.returncode is not None
-            aborted = new_state.returncode >= 256
             event = FinishedEvent(
                 iens=iens,
                 returncode=new_state.returncode,
-                aborted=aborted,
             )
 
-            if aborted:
+            if new_state.returncode >= SIGNAL_OFFSET:
```
pinkwah, line 318: Failure is when the returncode is non-zero, not when there is a signal. This should just check whether `new_state.returncode != 0`.

JHolba: thanks! fixed it

pinkwah (approved):

PR #7506, opened 2024-03-22
Fix If -> if in docs

berland:
Also reformat the paragraph for linelengths.

**Issue**
Resolves wrong casing of a single letter in the docs.


xjules (approved):
Goood

PR #7517, opened 2024-03-25
Fix ruff preview option PLC1901

### LINKED ISSUE #7516: Fix ruff preview option PLC1901
**Describe the bug**
Related to #7377 
``````
(py3.10-venv) x86_64 ➜  ert git:(improve-ruff-coverage) ✗ ruff rule PLC1901       
# compare-to-empty-string (PLC1901)

## What it does
Checks for comparisons to empty strings.

## Why is this bad?
An empty string is falsy, so it is unnecessary to compare it to `""`. If
the value can be something else Python considers falsy, such as `None`,
`0`, or another empty container, then the code is not equivalent.

## Known problems
High false positive rate, as the check is context-insensitive and does not
consider the type of the variable being compared ([#4282]).

## Example
```python
x: str = ...

if x == "":
    print("x is empty")
```

Use instead:
```python
x: str = ...

if not x:
    print("x is empty")
```
``````

### PR DESCRIPTION
**Pull request by jonathan-eq**
**Issue**
Resolves #7516

**Approach**
The commit drops empty string comparisons, as the value would be falsy anyways.

## When applicable

### DIFF
--- a/src/ert/gui/ertwidgets/checklist.py
+++ b/src/ert/gui/ertwidgets/checklist.py
@@ -112,14 +112,14 @@ class CheckList(QWidget):
 
         self.filterList(self._search_box.filter())
 
-    def filterList(self, _filter):
+    def filterList(self, _filter: str):
         _filter = _filter.lower()
 
         for index in range(0, self._list.count()):
             item = self._list.item(index)
-            text = str(item.text()).lower()
+            text = item.text().lower()
 
-            if _filter == "" or _filter in text:
+            if not _filter or _filter in text:
                 item.setHidden(False)
             else:
                 item.setHidden(True)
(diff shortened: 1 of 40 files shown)

### CODE REVIEW

File: src/ert/gui/ertwidgets/checklist.py [hunk 1]
```diff
@@ -119,7 +119,7 @@ def filterList(self, _filter):
             item = self._list.item(index)
             text = str(item.text()).lower()
 
-            if _filter == "" or _filter in text:
+            if not _filter or _filter in text:
```
berland, line 122: Here this change will make it trigger on `_filter` being `None` or `0` in Python as there is no type. But since we have called `_lower()` anything else but a string will make it crash beforehand so the change is ok. It would be nice to have `_filter` typed though.

berland:
Can you go through each change and ensure the type is checked through mypy?

src/ert/gui/ertwidgets/checklist.py
jonathan-eq: I think that if we also want to assert that the type is string, we should do isinstance(str,...) instead. I assume the old idea was that it would only check if the filter was set.

jonathan-eq:
Yes, I have gone through and verified manually. I had to add some typing to plot_style.py.

berland (approved):
LGTM!

PR #7519, opened 2024-03-25
Fix ruff preview option PLC0415

### LINKED ISSUE #7518: Fix ruff preview option PLC0415
**Describe the bug**
Related to #7377 

`````
(py3.10-venv) x86_64 ➜  ert git:(improve-ruff-coverage2) ✗ ruff rule PLC0415                 
# import-outside-top-level (PLC0415)

## What it does
Checks for `import` statements outside of a module's top-level scope, such
as within a function or class definition.

## Why is this bad?
[PEP 8] recommends placing imports not only at the top-level of a module,
but at the very top of the file, "just after any module comments and
docstrings, and before module globals and constants."

`import` statements have effects that are global in scope; defining them at
the top level has a number of benefits. For example, it makes it easier to
identify the dependencies of a module, and ensures that any invalid imports
are caught regardless of whether a specific function is called or class is
instantiated.

An import statement would typically be placed within a function only to
avoid a circular dependency, to defer a costly module load, or to avoid
loading a dependency altogether in a certain runtime environment.

## Example
```python
def print_python_version():
    import platform

    print(python.python_version())
```

Use instead:
```python
import platform


def print_python_version():
    print(python.python_version())
```

[PEP 8]: https://peps.python.org/pep-0008/#imports
`````

### PR DESCRIPTION
**Pull request by jonathan-eq**
**Issue**
Resolves #7518

**Approach**

## When applicable

### DIFF
--- a/src/ert/__main__.py
+++ b/src/ert/__main__.py
@@ -28,7 +28,6 @@ from ert.cli import (
 )
 from ert.cli.main import ErtCliError, ErtTimeoutError, run_cli
 from ert.config import ConfigValidationError, ErtConfig, lint_file
-from ert.gui.main import run_gui
 from ert.logging import LOGGING_CONFIG
 from ert.logging._log_util_abort import _log_util_abort
 from ert.namespace import Namespace
@@ -186,6 +185,8 @@ def valid_port_range(user_input: str) -> range:
 
 
 def run_gui_wrapper(args: Namespace, ert_plugin_manager: ErtPluginManager) -> None:
+    from ert.gui.main import run_gui
+
     run_gui(args, ert_plugin_manager)
 
 
--- a/src/ert/__main__.py
+++ b/src/ert/__main__.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
 import argparse
+import locale
 import logging
 import logging.config
 import os
 import re
+import resource
 import sys
 import warnings
 from argparse import ArgumentParser, ArgumentTypeError
@@ -26,6 +28,7 @@ from ert.cli import (
 )
 from ert.cli.main import ErtCliError, ErtTimeoutError, run_cli
 from ert.config import ConfigValidationError, ErtConfig, lint_file
+from ert.gui.main import run_gui
 from ert.logging import LOGGING_CONFIG
 from ert.logging._log_util_abort import _log_util_abort
 from ert.namespace import Namespace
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: src/ert/storage/local_storage.py [hunk 1]
```diff
@@ -300,29 +299,26 @@ def _save_index(self) -> None:
     def _migrate(self, ignore_migration_check: bool) -> None:
         if ignore_migration_check:
             return
+        from ert.storage.migration import (  # noqa: PLC0415
```
berland, line 302: Are we sure this will work in all circumstances? Considering the exisiting "`disable=C0415`" that is now removed.

jonathan-eq: Yes, that was the pylint equivalent, which we no longer use.

src/ert/__main__.py
```diff
@@ -26,6 +28,7 @@
 )
 from ert.cli.main import ErtCliError, ErtTimeoutError, run_cli
 from ert.config import ConfigValidationError, ErtConfig, lint_file
+from ert.gui.main import run_gui
```
eivindjahren, line 31: This will import `ert.gui` when cli is used. We should check that this 1. does not have unintended side-effects and 2. does not slow down startup time for cli.

src/ert/storage/migration/block_fs.py
```diff
@@ -257,7 +257,7 @@ def migrate(path: Path) -> None:
     )
 
 
-def _migrate_case_ignoring_exceptions(storage: Storage, casedir: Path) -> bool:
+def _migrate_case_ignoring_exceptions(storage: "Storage", casedir: Path) -> bool:
```
eivindjahren, line 260: I suggest using `from __future__ import annotations` instead.

src/ert/job_queue/__init__.py
```diff
@@ -65,9 +65,9 @@ def setenv(var: str, value: str) -> None:
 
 
 def _load_lib() -> Any:
-    import ctypes
+    import ctypes  # noqa: PLC0415
```
eivindjahren, line 68: I think the `ctypes` import should be at the top of the file here if possible. `ert._clib` import can be done dynamically.

src/ert/storage/migration/block_fs.py
jonathan-eq: Yes, the circular dependency issue I had originated elsewhere

eivindjahren (approved):
:+1:

PR #7526, opened 2024-03-26
Remove poly integration test on PBS

### LINKED ISSUE #7524: Integration tests on Azure with PBS fails due to komodo on a local disk
The komodoenv used in this test is (on Azure) located on a local disk, and this is not accessible to the compute nodes, and the test that actually tries to run the poly example successfully will thus fail.

https://github.com/equinor/ert/blob/ef492f029be11b71e114b4e9d1acfbc565594aac/ci/testkomodo.sh#L80

berland:
When tested through Komodo, this will not work as it requires the environment and the runpath to be on a shared disk, which the komodo setup currently does not facilitate.

Komodo testing will still effectively run this integration test through bigpoly.

**Issue**
Resolves #7524 

**Approach**
Drop this test. It is not tested for LSF, that is the reason this test file "passes" onprem.

If test is to be run, it requires some more infrastructure in place from Komodo. The easiest way in Komodo to have this pass is to let all komodo integration test run with everything on a shared disk, which will hamper test performance for all tests.

## When applicable

### DIFF
--- a/tests/integration_tests/scheduler/test_openpbs_driver.py
+++ b/tests/integration_tests/scheduler/test_openpbs_driver.py
@@ -27,19 +27,6 @@ def queue_name_config():
     return ""
 
 
-@pytest.mark.integration_test
-@pytest.mark.usefixtures("copy_poly_case")
-def test_openpbs_driver_with_poly_example(queue_name_config):
-    with open("poly.ert", mode="a+", encoding="utf-8") as f:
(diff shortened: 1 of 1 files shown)


berland:
I agree this test will be missed. But I think we should remove it first since it is blocking nightly testing, and then reintroduce it when done properly.

jonathan-eq (approved):
Let's :rock: and roll! 🎸

PR #7529, opened 2024-04-02
Fix ruff preview option PLR6301

### LINKED ISSUE #7528: Fix ruff preview option PLR6301
**Describe the bug**
Related to #7377 
``````
(py3.10-venv) x86_64 ➜  ert git:(improve-ruff-coverage) ✗ ruff rule PLR6301                                            
# no-self-use (PLR6301)

## What it does
Checks for the presence of unused `self` parameter in methods definitions.

## Why is this bad?
Unused `self` parameters are usually a sign of a method that could be
replaced by a function, class method, or static method.

## Example
```python
class Person:
    def greeting(self):
        print("Greetings friend!")
```

Use instead:
```python
def greeting():
    print("Greetings friend!")
```
``````

jonathan-eq:
**Issue**
Resolves #7528 

**Approach**

## When applicable

### DIFF
--- a/src/_ert_job_runner/reporting/file.py
+++ b/src/_ert_job_runner/reporting/file.py
@@ -111,13 +111,15 @@ class File(Reporter):
                 self._dump_ok_file()
         self._dump_status_json()
 
-    def _delete_old_status_files(self):
+    @staticmethod
+    def _delete_old_status_files():
         logger.debug("Deleting old status files")
         cond_unlink(ERROR_file)
         cond_unlink(STATUS_file)
         cond_unlink(OK_file)
 
-    def _write_status_file(self, msg: str) -> None:
+    @staticmethod
+    def _write_status_file(msg: str) -> None:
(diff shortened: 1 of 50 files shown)


berland:
I think you can skip the `staticmethod` decorator on maybe all these functions, check f.ex. https://stackoverflow.com/questions/55080991/does-the-staticmethod-decorator-do-anything

jonathan-eq:
> I think you can skip the `staticmethod` decorator on maybe all these functions, check f.ex. https://stackoverflow.com/questions/55080991/does-the-staticmethod-decorator-do-anything

I chose to use `@staticmethod` to avoid having to rewrite all method calls from `class_instance.method()` to `Class.method()` as the first one would pass `self` as first argument. `@staticmethod` ignores `self` as first argument.

berland:
ah, right, so the diff would actually become larger if staticmethod was not used. 👍🏻

berland (approved):
LGTM 🧹

PR #7534, opened 2024-04-02
Ignore failed test-summary

eivindjahren:
test-summary will fail if junit.xml does not exist, which happens when the tests time out. Since this is only for pretty output, we ignore such errors.


yngve-sk (approved):
Looking good 🚀

PR #7540, opened 2024-04-03
Fix ErtPlugin interface staticmethods

jonathan-eq:
Commit `15ff246` marked interface methods with `@staticmethod` without considering other scripts inheriting the interface.


eivindjahren (approved):
:+1:

PR #7552, opened 2024-04-04
Add --disable-monitor to Ert client runs in integration tests

berland:
This reduces the amount of lines printed to stdout while running through the test-suite.

**Issue**
Resolves  lengthy output when running `pytest -s`

**Approach**
Add command line option to ert command line client.

## When applicable

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -365,6 +365,7 @@ def _run_snake_oil(source_root):
         parser,
         [
             ENSEMBLE_EXPERIMENT_MODE,
+            "--disable-monitor",
             "--current-case",
             "default_0",
             "snake_oil.ert",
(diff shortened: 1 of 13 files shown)


tests/conftest.py
```diff
@@ -362,6 +362,7 @@ def _run_snake_oil(source_root):
         parser,
         [
             ENSEMBLE_EXPERIMENT_MODE,
+            "--disable-monitor",
```
jonathan-eq, line 365: Is it `--disable-monitor`, or `--disable-monitoring`?
I see another integration test uses `...-monitoring`

jonathan-eq (changes requested):

tests/conftest.py
jonathan-eq: I guess we can use both, but should we have it consistent?

berland: thx, it should only be `monitoring`.

berland: `argparse` allows for shortcut on options..

berland: Fixed.

jonathan-eq (approved):
LGTM!

PR #7556, opened 2024-04-04
Avoid calling kill() on each submit in LocalDriver

berland:
This cleanup seems redundant, and will give wrong logs.

Also removes a test that relies on kill() being called upon
each submit, which we should not rely on.

**Issue**
Resolves suspicios kill call.

**Approach**
✂️ 

## When applicable

### DIFF
--- a/src/ert/scheduler/local_driver.py
+++ b/src/ert/scheduler/local_driver.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import logging
 import os
 from asyncio.subprocess import Process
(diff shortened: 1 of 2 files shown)


berland:
@jonathan-eq , you wrote the test that I am removing. Do you think it should be rewritten or should we just delete it?

jonathan-eq:
> @jonathan-eq , you wrote the test that I am removing. Do you think it should be rewritten or should we just delete it?

The test was mostly there to verify that exceptions in scheduler were properly propagated, so it can be deleted now. We have better tests for it.

jonathan-eq (approved):
LGTM!

PR #7562, opened 2024-04-04
Make sure to establish connection and empty the event queue before cancelling publisher 

### LINKED ISSUE #7522: Scheduler is flaky, "Too many realizations failed"
**Describe the bug**
The test `tests/integration_tests/analysis/test_es_update.py::test_update_multiple_param` is very flaky, atleast using scheduler. 
```
FAILED tests/integration_tests/analysis/test_es_update.py::test_update_multiple_param[using_scheduler] - ert.cli.main.ErtCliError: Too many realizations have failed! Number of successful realizations: 9, expected minimal number of successful realizations: 10
```
(https://github.com/equinor/ert/actions/runs/8420486419/job/23055374718?pr=7521)

**Expected behaviour**
Not flaky.

**Environment**
 - OS: Mostly macos-latest
 - ERT/Komodo release: main
 - Python version: Mostly (if not only) 3.12
 - Remote/HPC execution involved: Github Actions runner

**Additional context**

xjules:
**Issue**
Resolves #7522 

## When applicable

### DIFF
--- a/src/ert/scheduler/scheduler.py
+++ b/src/ert/scheduler/scheduler.py
@@ -26,7 +26,11 @@ from websockets import Headers
 from websockets.client import connect
 
 from ert.constant_filenames import CERT_FILE
-from ert.job_queue.queue import EVTYPE_ENSEMBLE_CANCELLED, EVTYPE_ENSEMBLE_STOPPED
+from ert.job_queue.queue import (
+    CLOSE_PUBLISHER_SENTINEL,
+    EVTYPE_ENSEMBLE_CANCELLED,
+    EVTYPE_ENSEMBLE_STOPPED,
+)
 from ert.scheduler.driver import SIGNAL_OFFSET, Driver
 from ert.scheduler.event import FinishedEvent
 from ert.scheduler.job import Job
@@ -106,6 +110,7 @@ class Scheduler:
         self._ens_id = ens_id
         self._ee_cert = ee_cert
         self._ee_token = ee_token
+        self._publisher_done = asyncio.Event()
 
     def kill_all_jobs(self) -> None:
         assert self._loop
@@ -187,7 +192,11 @@ class Scheduler:
         ):
             while True:
                 event = await self._events.get()
+                if event == CLOSE_PUBLISHER_SENTINEL:
+                    self._publisher_done.set()
+                    return
                 await conn.send(event)
+                self._events.task_done()
 
     def add_dispatch_information_to_jobs_file(self) -> None:
         for job in self._jobs.values():
@@ -225,6 +234,9 @@ class Scheduler:
                         raise task_exception
 
             if not self.is_active():
+                if self._ee_uri is not None:
+                    await self._events.put(CLOSE_PUBLISHER_SENTINEL)
+                    await self._publisher_done.wait()
                 for task in self._job_tasks.values():
                     if task.cancelled():
                         continue
--- a/tests/integration_tests/analysis/test_es_update.py
+++ b/tests/integration_tests/analysis/test_es_update.py
@@ -164,7 +164,6 @@ def test_that_surfaces_retain_their_order_when_loaded_and_saved_by_ert():
         )
 
 
-@pytest.mark.skip(reason="Very flaky with scheduler")
 @pytest.mark.integration_test
 @pytest.mark.usefixtures("copy_snake_oil_field", "using_scheduler")
 def test_update_multiple_param():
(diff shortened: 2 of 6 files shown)


src/ert/scheduler/scheduler.py
```diff
@@ -263,6 +263,8 @@ async def execute(
         try:
             await self._monitor_and_handle_tasks(scheduling_tasks)
         finally:
+            while not self._events.empty():
```
berland, line 266: Maybe it is `_monitor_and_handle_tasks` that should not return until this is empty. 
And inside there one might consider if this should be a test inside `is_active()`, that scheduler should consider itself active while there are events pending.

xjules: Let's give it a try then

berland: simulation_context is using `is_active()` so maybe we should put the guard inside there to avoid the same problem popping up there.

berland: I vote for changing `is_active()` to return True whenever the event queue is nonempty.

berland: Getting there, I see we need to mock `websocket.client.connect` in order to test this.

berland: And possibly we should only require the queue to become empty if `self.__ee_uri` is not None.

xjules: I needed to change it back to a version where task handles cancellation itself. Two main reasons:
1) The way task is doing cleanup should be an implementation detail and as such we don't want to create class variables (eg. publisher running) and maintain them.
2) Simulation context handling becomes far more complex just to check publisher running and queue not empty - I mean the tests mostly.
Thus the best option would be to do the cleanup once we cancel the task as this would become an implementation detail of the publisher task.

src/ert/scheduler/scheduler.py
```diff
@@ -228,9 +231,10 @@ async def _monitor_and_handle_tasks(
                         continue
                     if task_exception := task.exception():
                         raise task_exception
-                # make sure to empty report all events before exit
-                while not self._events.empty():
-                    await asyncio.sleep(0)
+                # make sure to report all events before exit
+                if self._publisher_task_active:
```
jonathan-eq: Should we again set `self._publisher_task_active = False` after this if block is executed?

jonathan-eq: I was looking at a specific commit rather than the entire change...

xjules: Now it's gone - I've reverted it back as described in the comment https://github.com/equinor/ert/pull/7562#discussion_r1553047562

.pre-commit-config.yaml
```diff
@@ -16,7 +16,7 @@ repos:
     - id: ruff-format
 
 -   repo: https://github.com/pre-commit/mirrors-clang-format
-    rev: v18.1.1
+    rev: v18.1.3
```
berland, line 19: would be nice to have in a separate commit in case of rollbacks.

src/ert/async_utils.py
```diff
@@ -40,6 +41,16 @@ def _done_callback(task: asyncio.Task[_T_co]) -> None:
         if (exc := task.exception()) is None:
             return
 
-        logger.error(f"Exception occurred during {task.get_name()}", exc_info=exc)
+        # logger.error(f"Exception occurred during {task.get_name()}", exc_info=exc)
```
berland, line 44: rm this line

.pre-commit-config.yaml
xjules: Sorry, left over. Will rm.

tests/integration_tests/analysis/test_es_update.py
```diff
@@ -164,7 +164,6 @@ def sample_prior(nx, ny):
         )
 
 
-@pytest.mark.skip(reason="Very flaky with scheduler")
```
berland, line 167: thank you for not forgetting this! 😄

src/ert/scheduler/scheduler.py
```diff
@@ -215,14 +225,18 @@ async def _monitor_and_handle_tasks(
                     )
                     logger.error(
                         (
-                            f"Exception in scheduler task: {task_exception}\n"
+                            f"Exception in scheduler task {task.get_name()}: {task_exception}\n"
```
berland, line 228: This should perhaps be a separate commit, as it is unrelated to fixing the premature closing of publisher.

xjules: Now it's in two commits. 1) improving exception messages 2) the actual fix

berland (approved):
splendid

PR #7594, opened 2024-04-08
Fix Python3.12 deprecation warning in async_utils.py

### LINKED ISSUE #7588: Flaky integration test test_tracking_integration.py::test_tracking[using_job_queue-ee_failing_poly_smoother]
**Describe the bug**
`FAILED tests/integration_tests/status/test_tracking_integration.py::test_tracking[using_job_queue-ee_failing_poly_smoother] - Failed: Timeout >360.0s` (https://github.com/equinor/ert/actions/runs/8595749537/job/23551154574?pr=7562)

**To reproduce**
1. `pip install ert`
2. `ert gui my_config.ert`
4. …

**Expected behaviour**

**Screenshots**

**Environment**
 - OS: macos-latest
 - ERT/Komodo release: main
 - Python version: 3.12
 - Remote/HPC execution involved: Github Action Runner

**Additional context**

jonathan-eq:
**Issue**
Python3.12 has a deprecation warning in async_utils.py which might actually be a bug.
We expect `asyncio.get_event_loop()` to raise a `RuntimeError` if there is not an event loop in the thread, but this is actually the behavior of `asyncio.get_running_loop()` instead. 
The consequence of this is that the code in the `except`-block is never executed, and the new loop does not have a task factory with our custom done/debug callbacks from b2b2e48. 

**Approach**
The commit in this PR fixes the `DeprecationWarning: There is no current event loop` warning for Python3.12 in the `get_event_loop` function in `async_utils.py`. Prior to this commit, the function used `asyncio.get_event_loop()` which creates a new event loop directly. `asyncio.get_running_loop()` does however raise a RuntimeError when there is no event loop, which is what we want.

## When applicable

### DIFF
--- a/src/ert/async_utils.py
+++ b/src/ert/async_utils.py
@@ -19,7 +19,7 @@ def new_event_loop() -> asyncio.AbstractEventLoop:
 
 def get_event_loop() -> asyncio.AbstractEventLoop:
     try:
-        return asyncio.get_event_loop()
+        return asyncio.get_running_loop()
     except RuntimeError:
         asyncio.set_event_loop(new_event_loop())
         return asyncio.get_event_loop()


src/ert/async_utils.py
```diff
@@ -19,7 +19,7 @@ def new_event_loop() -> asyncio.AbstractEventLoop:
 
 def get_event_loop() -> asyncio.AbstractEventLoop:
     try:
-        return asyncio.get_event_loop()
+        return asyncio.get_running_loop()
```
xjules, line 22: Not sure about this one. Based on the documentation: https://docs.python.org/3/library/asyncio-eventloop.html
`get_running_loop` can cause RuntimeError while get_event_loop will create one.

xjules: It's fine then

xjules (approved):
Gooooood! :rocket:

src/ert/async_utils.py
jonathan-eq: Yes, I assume that was the expected behavior.

PR #7599, opened 2024-04-09
Let mocked bkill support specific kill signal

### LINKED ISSUE #7598: Mocked bkill.py not compatible with current lsf_driver.py
lsf_driver.py adds a `-s` option to the bkill command line, but this is not supported by the mocked bkill.py

When running test_generic_driver.py with `-sxv`, one sees the errors from argparse in the mock, but another error masks this failure and the test passes as a false positive.

berland:
The tests sent a -s option to the mock that failed the mock. Another error (missing timeout) let the kill test pass when it should not.

**Issue**
Resolves #7598 

**Approach**
Add timeout to fail the test, then fix the mock and pass the test.

(since the LSF driver does not return the correct return code, the assert on return code did not otherwise fail the test as it would for PBS driver)

## When applicable

### DIFF
--- a/tests/integration_tests/scheduler/bin/bkill.py
+++ b/tests/integration_tests/scheduler/bin/bkill.py
@@ -7,6 +7,9 @@ from pathlib import Path
 
 def get_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(description="Kill jobs")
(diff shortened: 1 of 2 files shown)


jonathan-eq (approved):
Very good!👌

Issue #5453, opened 2023-05-23
Deprecated items not handled in new parser


yngve-sk:
**Describe the bug**
Items can be specified to be deprecated with the `deprecated` attribute and the `deprecate_msg` attribute. But these attributes are not used to raise errors.

PR #7602, opened 2024-04-09
Have OpenPBS driver use qstat -E option

jonathan-eq:
**Issue**
Resolves [#5453](https://github.com/equinor/komodo-releases/issues/5453)

**Approach**
The commit in this PR might solve the issue by having qstat be more performant. The -E option groups the job ids together, so instead of one connection per job; they will all be executed in the same one.

## When applicable

### DIFF
--- a/src/ert/scheduler/openpbs_driver.py
+++ b/src/ert/scheduler/openpbs_driver.py
@@ -244,6 +244,7 @@ class OpenPBSDriver(Driver):
                     "qstat",
                     "-x",
                     "-w",  # wide format
+                    "-E",  # group job ids
                     *self._non_finished_job_ids,
                     stdout=asyncio.subprocess.PIPE,
                     stderr=asyncio.subprocess.PIPE,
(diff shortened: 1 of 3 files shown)


src/ert/scheduler/openpbs_driver.py
```diff
@@ -244,6 +244,7 @@ async def poll(self) -> None:
                     "qstat",
                     "-x",
                     "-w",  # wide format
+                    "-E",  # group job ids
```
berland, line 247: Can the comment be made more precise, check the manual, I think it is more like group by PBS server. Maybe also comment that it is critical for speed.

berland: (at least the reason for why -E is added should be either in a comment or in the commit message)

jonathan-eq:
@berland Should I close this in favor of https://github.com/equinor/ert/pull/7573?

berland:
Duplicate of #7573

Issue #7587, opened 2024-04-08
Flaky test, test_observation_times::test_small_time_mismatches_are_ignored


berland:
Occured in https://github.com/equinor/ert/actions/runs/8553582418/job/23475634448?pr=7552
```
FAILED tests/integration_tests/test_observation_times.py::test_small_time_mismatches_are_ignored - AssertionError: assert 'Experiment completed' in ''
 +  where '' = <built-in method getvalue of _io.StringIO object at 0x7f3ba8735700>()
 +    where <built-in method getvalue of _io.StringIO object at 0x7f3ba8735700> = <_io.StringIO object at 0x7f3ba8735700>.getvalue
Falsifying example: test_small_time_mismatches_are_ignored(
    tmp_path_factory=TempPathFactory(_given_basetemp=PosixPath('/tmp/pytest-of-runner/pytest-0/popen-gw2'), _trace=<pluggy._tracing.TagTracerSub object at 0x7f3bb12a20a0>, _basetemp=PosixPath('/tmp/pytest-of-runner/pytest-0/popen-gw2'), _retention_count=3, _retention_policy='all'),
    responses_observation={'responses': [(Smspec(
           nx=int32(10),
           ny=int32(4),
           nz=int32(4),
           restarted_from_step=0,
           num_keywords=2,
           restart='RVYDFMNE',
           keywords=['TIME    ', 'FOPR'],
           well_names=[':+:+:+:+', 'FCVHBISY'],
           lgrs=None,
           numlx=None,
           numly=None,
           numlz=None,
           region_numbers=[-32676, int32(106)],
           units=['HOURS   ', 'HOURS'],
           start_date=Date(day=1, month=1, year=1969, hour=0, minutes=0, micro_seconds=0),
           use_names=True,
           intehead=SmspecIntehead(
               unit=UnitSystem.METRIC,
               simulator=Simulator.INTERSECT,
           ),
       ),
       Unsmry(steps=[SummaryStep(seqnum=0, ministeps=[SummaryMiniStep(mini_step=0, params=[0.0, 5.4072517e+16]), SummaryMiniStep(mini_step=1, params=[121488.0, 211106230000000.0])])])),
      (Smspec(
           nx=int32(1),
           ny=int32(7),
           nz=int32(9),
           restarted_from_step=0,
           num_keywords=2,
           restart='WDQTLACB',
           keywords=['TIME    ', 'FOPR'],
           well_names=[':+:+:+:+', 'HELVCVPG'],
[...]

PR #7612, opened 2024-04-09
Reduce flakyness in test_observation_times

berland:
The commit adding --disable-monitor b5a72e3e1572471b927d64c0146fc739283c900c made this test fail in case it ended in the 'less than one second' if-branch.

**Issue**
Resolves partly the flakyness of test_observation_times. There is still residual flakyness as in #7587

**Approach**
Change the assert to the actual output when --disable-monitoring is in use.

## When applicable

### DIFF
--- a/tests/integration_tests/test_observation_times.py
+++ b/tests/integration_tests/test_observation_times.py
@@ -113,12 +113,14 @@ def test_small_time_mismatches_are_ignored(
         stderr = StringIO()
         with redirect_stderr(stderr):
             run_cli(
-                ES_MDA_MODE,
                 "--disable-monitor",
+                ES_MDA_MODE,
                 str(tmp_path / "config.ert"),
                 "--weights=0,1",
             )
-        assert "Experiment completed" in stderr.getvalue()
+        assert (
+            not stderr.getvalue()
+        ), "There should be no errors when mismatches are small"
     else:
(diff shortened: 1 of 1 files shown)


tests/integration_tests/test_observation_times.py
```diff
@@ -113,12 +113,14 @@ def test_small_time_mismatches_are_ignored(
...
-                ES_MDA_MODE,
                 "--disable-monitor",
+                ES_MDA_MODE,
                 str(tmp_path / "config.ert"),
                 "--weights=0,1",
             )
-        assert "Experiment completed" in stderr.getvalue()
+        assert (
+            not stderr.getvalue()
+        ), "There should be no errors when mismatches are small"
```
sondreso, line 123: I don't think this will work, as the monitor will always print to the same stream (no matter failure or success):

https://github.com/equinor/ert/blob/6f49e28aa529c7d18de9513954ee848fa04d7b1b/src/ert/cli/monitor.py#L106-L115

So this will make the test pass, but I'm not sure it will fail if there is an actual error.

sondreso (commented):
I did not realize that there was already a PR for fixing this so I created a separate PR for just removing the `--disable-monitor` here: https://github.com/equinor/ert/pull/7613

Guess we can consider merging that, or fix the above issue.

berland:
Closing in favour of #7613

PR #7616, opened 2024-04-10
Remove start sync event from scheduler job.__call__

### LINKED ISSUE #7413: Remove start sync event from the job __call__
The synchronization event (`start = asyncio.Event()`) made for the purpose of executing job-group start,  which gets fed into `job.__call__` does not seems to be that necessary in the end as `asyncio.BoundedSemaphore` will handle job execution regardless. Therefore we should remove this start event.

jonathan-eq:
**Issue**
Resolves #7413 

**Approach**
The scheduler job has a synchronization event (`start = asyncio.Event()`) with the purpose of executing job-group start,  and was passed as a parameter to `job.__call__`. It does not seems to be that necessary in the end as `asyncio.BoundedSemaphore` will handle job execution regardless. This commit removes this redundant event.

## When applicable

### DIFF
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -125,11 +125,8 @@ class Job:
                 timeout_task.cancel()
             sem.release()
 
-    async def run(
-        self, start: asyncio.Event, sem: asyncio.BoundedSemaphore, max_submit: int = 2
-    ) -> None:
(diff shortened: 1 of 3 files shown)


xjules (approved):
I think it should be doable now, as we do we have a guard on completion.
Some formatting needed though.

PR #7624, opened 2024-04-10
Have async function use event loop from async_utils

### LINKED ISSUE #7619: Async functions are not using event loop configured in async_utils.py
**Describe the bug**
A lot of the async ert code uses `asyncio.get_event_loop()` or `asyncio.get_running_loop()` to get an event loop, but we should be using our own `get_running_loop()` function from `async_utils.py` instead.
The event loop from `async_utils.py` adds a done-callback to every task created, which can greatly help debugging.

**Expected behaviour**
All places where we source an event loop should be using the function from `async_utils`.

jonathan-eq:
**Issue**
Resolves #7619 

**Approach**
Have async function use event loop from async_utils

## When applicable


src/_ert_job_runner/client.py
```diff
@@ -73,7 +75,7 @@ def __init__(
         self._max_retries = max_retries
         self._timeout_multiplier = timeout_multiplier
         self.websocket: Optional[WebSocketClientProtocol] = None
-        self.loop = asyncio.new_event_loop()
+        self.loop = new_event_loop()
```
xjules, line 78: Unfortunately, we'd like to avoid using ert in job_runner.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -67,7 +67,7 @@ def __init__(self, ensemble: Ensemble, config: EvaluatorServerConfig, iter_: int
         self._config: EvaluatorServerConfig = config
         self._ensemble: Ensemble = ensemble
 
-        self._loop = new_event_loop()
+        self._loop = get_running_loop()
```
xjules, line 70: Not sure here as I think evaluator would like to have its own loop, but  it might be just fine :shrug:

src/_ert_job_runner/client.py
jonathan-eq: Even the code in `_ert`?

xjules: As discussed, that is fine in the end. So just ignore my comment above :)

src/ert/scheduler/scheduler.py
```diff
@@ -95,8 +96,8 @@ def __init__(
             real.iens: Job(self, real) for real in (realizations or [])
         }
 
+        self._loop = get_running_loop()
```
xjules, line 99: I'm still wondering whether `get_running_loop` should just crash when there is no loop running - instead of creating one. As this is the basic behavior of asyncio https://docs.python.org/3/library/asyncio-eventloop.html opposite to our version.

jonathan-eq: Yes, but why would we crash instead of handling it by creating a new one?

xjules: Depends what we expect. Do we expect that there is already one? Maybe we don't care :smile: In that case it is fine. Let's just say it's fine then.

PR #7625, opened 2024-04-10
Add test to confirm that the job does not retry when activelly cancelled

### LINKED ISSUE #7500: Change retry submit in scheduler to run unless user cancelled a realization
Currently we retry submit if return code of a job was >128 (a signal killed the job).
It would be better if we kept information about user cancellation in the driver and stopped retries based on that instead.

xjules:
**Issue**
Resolves #7500 

**Approach**
The functionality should be there already. Just made a test that verifies that it works.

## When applicable


tests/unit_tests/scheduler/test_scheduler.py
```diff
@@ -333,6 +333,36 @@ async def kill():
...
+    sch = scheduler.Scheduler(mock_driver(wait=wait), [realization], max_submit=2)
+
+    scheduler_task = asyncio.create_task(sch.execute())
+
+    await kill_me.wait()
+    await sch.cancel_all_jobs()
+
+    await scheduler_task
+    assert scheduler_task.result() == EVTYPE_ENSEMBLE_CANCELLED
+    assert retries == 1
```
berland, line 357: I suggest changing the line to `assert retries == 1, "Job` was resubmitted after killing"` - self-documenting :)

tests/unit_tests/scheduler/test_scheduler.py
```diff
@@ -333,6 +333,36 @@ async def kill():
...
+    assert sch._max_submit == 2
```
berland, line 358: this should not be needed, and is testing an implemention detail.

berland (approved):
👍🏻

PR #7654, opened 2024-04-12
Fix documentation for LSF memory booking

### LINKED ISSUE #7651: Documentation for memory booking on LSF is wrong
The specification of units in the example in the documentation is only valid for LSF10, which we don't yet facilitate.

https://github.com/equinor/ert/blob/4496033be04dab7622780034a19d7ace9902a514/docs/reference/configuration/queue.rst?plain=1#L189

For LSF9 the unit must be omitted, and the default unit is KB (so the LSF9 docs say)

jonathan-eq:
**Issue**
Resolves #7651 

**Approach**
The commit in this PR updates the documentation for the LSF queue option `LSF_RESOURCE`. The old example used units in the resource string, but that is not supported by LSF9. The default unit is KB, so all memory requests will use that.

## When applicable

### DIFF
--- a/docs/reference/configuration/queue.rst
+++ b/docs/reference/configuration/queue.rst
@@ -183,10 +183,13 @@ The following is a list of available LSF configuration options:
   LSF uses resource requirements to select hosts for remote execution and
   job execution. Resource requirement strings can be simple (applying to the
   entire job) or compound (applying to the specified number of slots).
+  The value passed does not use units and depends on the cluster's configuration,
+  so follow up with cluster administrator to find out what the set unit is.
   `Docs. <https://www.ibm.com/support/knowledgecenter/SSWRJV_10.1.0/lsf_admin/res_req_strings_about.html>`__
-  Passed as the ``-R`` option to ``bsub``. For example::
+  Passed as the ``-R`` option to ``bsub``. For example, this will
+  request approximately 15 gigabytes when the default unit is megabytes::
 
-    QUEUE_OPTION LSF LSF_RESOURCE rusage[mem=512MB:swp=1GB]
+    QUEUE_OPTION LSF LSF_RESOURCE rusage[mem=15000]
 
 .. _project_code:
 .. topic:: PROJECT_CODE


docs/reference/configuration/queue.rst
```diff
@@ -183,10 +183,11 @@ The following is a list of available LSF configuration options:
   LSF uses resource requirements to select hosts for remote execution and
   job execution. Resource requirement strings can be simple (applying to the
   entire job) or compound (applying to the specified number of slots).
+  The set unit for memory booking is MB, so all values here will be in MB.
```
berland, line 186: We need to be careful here, because the default is installation dependent. LSF9 has a default at KB, while our onpremise data cluster has reconfigured the default to MB. 

(When we progress to LSF10 (soonish), there will be support for units, and then our documentation becomes correct)

jonathan-eq: As long as the documentation is valid at the time of the user reading this, then that should be fine. Is there a point in future proofing it? When when we go over to LSF10, we have to update the docs again anyways.

berland: The documentation in this PR will not be valid for anyone using Ert outside of Equinors premises..

jonathan-eq: Is anyone using it outside of Equinor's premises? I can mention that this is Equinor specific, but should I also say how it is normally, outside of Equinor's premises?

berland: Ert is public so we cannot make the assumption that it is not used outside.

docs/reference/configuration/queue.rst
```diff
@@ -183,10 +183,12 @@ The following is a list of available LSF configuration options:
   LSF uses resource requirements to select hosts for remote execution and
   job execution. Resource requirement strings can be simple (applying to the
   entire job) or compound (applying to the specified number of slots).
+  The value passed does not use units and depends on the cluster's configuration,
+  so follow up with cluster administrator to find out what the set unit is.
   `Docs. <https://www.ibm.com/support/knowledgecenter/SSWRJV_10.1.0/lsf_admin/res_req_strings_about.html>`__
   Passed as the ``-R`` option to ``bsub``. For example::
 
-    QUEUE_OPTION LSF LSF_RESOURCE rusage[mem=512MB:swp=1GB]
+    QUEUE_OPTION LSF LSF_RESOURCE rusage[mem=512000:swp=1000000]
```
berland, line 191: Lets change this to something relevant at Equinor:
`   
   QUEUE_OPTION LSF LSF_RESOURCE rusage[mem=15000]
`
and state this will be (approx) 15 gigabytes when the default unit is megabytes

berland (approved):
go for it!

PR #7656, opened 2024-04-12
Fix memory being reported without units in GUI

### LINKED ISSUE #7636: Write memory usage in a human friendly format
It can be argued that the memory consumption in this screenshot is not humanly parseable:

image

Suggest to change it to something human friendly format, KB, MB, GB..

jonathan-eq:
**Issue**
Resolves #7636 

**Approach**
Memory will now be reported in with suitable units.
image

## When applicable

### DIFF
--- a/src/ert/gui/model/snapshot.py
+++ b/src/ert/gui/model/snapshot.py
@@ -414,10 +414,10 @@ class SnapshotModel(QAbstractItemModel):
         if role == Qt.DisplayRole:
             _, data_name = COLUMNS[NodeType.REAL][index.column()]
             if data_name in [ids.CURRENT_MEMORY_USAGE, ids.MAX_MEMORY_USAGE]:
-                data = node.data.get(ids.DATA)
-                _bytes = data.get(data_name) if data else None
+                data = node.data
+                _bytes: Optional[str] = data.get(data_name) if data else None
                 if _bytes:
-                    return byte_with_unit(_bytes)
+                    return byte_with_unit(float(_bytes))
 
             if data_name in [ids.STDOUT, ids.STDERR]:
                 return "OPEN" if node.data.get(data_name) else QVariant()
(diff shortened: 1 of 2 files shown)


src/ert/gui/model/snapshot.py
```diff
@@ -414,10 +414,10 @@ def _job_data(index: QModelIndex, node: Node, role: int):
         if role == Qt.DisplayRole:
             _, data_name = COLUMNS[NodeType.REAL][index.column()]
             if data_name in [ids.CURRENT_MEMORY_USAGE, ids.MAX_MEMORY_USAGE]:
-                data = node.data.get(ids.DATA)
+                data = node.data
                 _bytes = data.get(data_name) if data else None
                 if _bytes:
-                    return byte_with_unit(_bytes)
+                    return byte_with_unit(float(_bytes))
```
berland, line 420: what is the type of `_bytes` ? Is this a bug that mypy could have caught?

jonathan-eq: I guess this might have happened when we refactored the snapshot and flattened the data dictionary. The problem was here `data = node.data.get(ids.DATA)`. We check if there is a data section inside of the data section. This would always be None.

berland: But we had numbers there before this PR, just not with units (?)

jonathan-eq: Maybe there is a different rendering somewhere else...
I fixed the initial problem, then I got a TypeError because it tried running `>=`between a float value and a str, so it never got to that point before.

jonathan-eq: Ahh, I see what happened now. This if block does not return anything if _bytes was None. It then hit the default return statement at the end of `if role == Qt.DisplayRole:` ... `return node.data.get(data_name)`

berland (approved):
💯

PR #7660, opened 2024-04-12
Avoid cancelling the returncode on normal failures

berland:
**Issue**
Resolves #7657 

**Approach**
_Short description of the approach_

(Screenshot of new behavior in GUI if applicable)

- [ ] PR title captures the intent of the changes, and is fitting for release notes.
- [ ] Added appropriate release note label
- [ ] Commit history is consistent and clean, in line with the [contribution guidelines](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md).
- [ ] Make sure tests pass locally (after every commit!)

## When applicable
- [ ] **When there are user facing changes**: Updated documentation
- [ ] **New behavior or changes to existing untested code**: Ensured that unit tests are added (See [Ground Rules](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md#ground-rules)).
- [ ] **Large PR**: Prepare changes in small commits for more convenient review
- [ ] **Bug fix**: Add regression test for the bug
- [ ] **Bug fix**: Create Backport PR to latest release

PR #7661, opened 2024-04-12
Fix scheduler LSF driver logging error when killing finished jobs

### LINKED ISSUE #7658: LSF driver tries to kill failed realizations
Observed in the log

```
2024-04-12 12:52:17,533 - ert.scheduler.lsf_driver - LegacyEnsemble - DEBUG - Realization 9 (LSF-id: 7810) failed
2024-04-12 12:52:17,569 - ert.scheduler.lsf_driver - LegacyEnsemble - ERROR - LSF kill failed due to missing jobid for realization 9
```

This is analogous to #7560 

Make sure to unify the behaviour over all three drivers.

jonathan-eq:
**Issue**
Resolves #7658 

**Approach**
The commit in this PR fixes the bug where the scheduler LSF driver would log `LSF kill failed due to missing jobid for realization` when killing jobs that had already finished with a failure.

## When applicable

### DIFF
--- a/src/ert/scheduler/lsf_driver.py
+++ b/src/ert/scheduler/lsf_driver.py
@@ -21,6 +21,7 @@ from typing import (
     MutableMapping,
     Optional,
     Sequence,
+    Set,
     Tuple,
     Union,
     get_args,
@@ -191,6 +192,7 @@ class LsfDriver(Driver):
 
         self._jobs: MutableMapping[str, Tuple[int, AnyJob]] = {}
         self._iens2jobid: MutableMapping[int, str] = {}
+        self._finished_iens: Set[int] = set()
         self._max_attempt: int = 100
         self._sleep_time_between_bkills = 30
         self._sleep_time_between_cmd_retries = 3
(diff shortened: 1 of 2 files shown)


src/ert/scheduler/lsf_driver.py
```diff
@@ -265,6 +267,8 @@ async def submit(
         self._iens2jobid[iens] = job_id
 
     async def kill(self, iens: int) -> None:
+        if iens in self._finished_iens:
+            return
```
berland, line 271: I think we are just covering an underlying problem with this, and should rather ensure we don't call `kill()` when we should not.

jonathan-eq: Maybe we can move the `if iens in self._finished:...` check to before the place `driver.kill(...)` is called.

jonathan-eq: Should the `finished` state be moved to the individual jobs (job.py) instead? That way we can check directly before calling `driver.kill(...)`

berland: https://github.com/equinor/ert/pull/7660 is one possible solution that will fix this.

jonathan-eq: @berland Great! Can I reassign you to this issue?

PR #7672, opened 2024-04-15
Fix LSF driver logging wrong message when killing

### LINKED ISSUE #7664: LSF driver logs wrong message when killing
When ctrl-c-ing ("Terminate experiment") is clicked on an ensemble_experiment (add `time.sleep(10)` to poly_eval.py)  on an LSF run, there are messages in the log:

```
LSF kill failed with stdout: Job <8055> is being signaled
 and stderr: 

```

This means that the `bkill` probably gives a nonzero returncode when it is doing the signalling, and we should not claim this is an error. Also `strip()` the stdout and stderr from the process to have this end up on one line.

After the Ert process is done, we get additional output to the terminal:
`Job <8056>: Job has already finished`
which should be muted.

jonathan-eq:
**Issue**
Resolves #7664 

**Approach**
The commit in this PR fixes some minor logging issues in scheduler lsf driver:
* sigkill fire-and-forget process logged to terminal instead of devnull after sleeping 30s.
* stdout and stderr from bkill was missing `strip()` so logs were split over multiple lines.
* bkill can output `Job <id> is being signaled`, but it should not be interpreted as an error.

## When applicable

### DIFF
--- a/src/ert/scheduler/lsf_driver.py
+++ b/src/ert/scheduler/lsf_driver.py
@@ -285,22 +285,26 @@ class LsfDriver(Driver):
             f"sleep {self._sleep_time_between_bkills}; {self._bkill_cmd} -s SIGKILL {job_id}",
             shell=True,
             start_new_session=True,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
         )
         if process.returncode:
             logger.error(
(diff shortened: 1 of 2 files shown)


berland:
Has this been verified manually to solve all issues mentioned?

berland (approved):

PR #7688, opened 2024-04-17
Make flakiness test for qstat faster

### LINKED ISSUE #7686: test_that_openpbs_driver_ignores_qstat_flakiness is too slow
```
$ pytest tests/*_tests/scheduler/ -sxv -k flakiness
tests/integration_tests/scheduler/test_openpbs_driver.py::test_that_openpbs_driver_ignores_qstat_flakiness[pbs_iff: cannot connect to host\npbs_iff: all reserved ports in use] PASSED
tests/integration_tests/scheduler/test_openpbs_driver.py::test_that_openpbs_driver_ignores_qstat_flakiness[qstat: Invalid credential] PASSED
tests/unit_tests/scheduler/test_openpbs_driver.py::test_that_openpbs_driver_ignores_qstat_flakiness[pbs_iff: cannot connect to host\npbs_iff: all reserved ports in use] PASSED
tests/unit_tests/scheduler/test_openpbs_driver.py::test_that_openpbs_driver_ignores_qstat_flakiness[qstat: Invalid credential] PASSED

======== 4 passed, 182 deselected in 46.86s =================
```

47 seconds is too much for this test.

berland:
Delete the integration test, the unit test should be sufficient.

Let the polling period be configurable for tests, in line with lsf_driver

This cuts the test time from 46 seconds to 1 second

**Issue**
Resolves #7686 

**Approach**
:brain: and ✂️ 

## When applicable

### DIFF
--- a/src/ert/scheduler/openpbs_driver.py
+++ b/src/ert/scheduler/openpbs_driver.py
@@ -126,6 +126,7 @@ class OpenPBSDriver(Driver):
         self._job_prefix = job_prefix
         self._num_pbs_cmd_retries = 10
         self._sleep_time_between_cmd_retries = 2
+        self._poll_period = _POLL_PERIOD
 
         self._jobs: MutableMapping[str, Tuple[int, AnyJob]] = {}
         self._iens2jobid: MutableMapping[int, str] = {}
@@ -236,7 +237,7 @@ class OpenPBSDriver(Driver):
     async def poll(self) -> None:
         while True:
             if not self._jobs:
-                await asyncio.sleep(_POLL_PERIOD)
+                await asyncio.sleep(self._poll_period)
                 continue
 
             if self._non_finished_job_ids:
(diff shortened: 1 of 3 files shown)


jonathan-eq:
In the unit test, should we assert that the error message does not appear in the logs? That's what the integration test here tested, but that is not present in the unit test.

berland:
> In the unit test, should we assert that the error message does not appear in the logs? That's what the integration test here tested, but that is not present in the unit test.

I think I managed to fix this. At least I am able to fail the unit test by injecting a logger.debug() in the drivers polling code that logs stderr.

jonathan-eq:
> In the unit test, should we assert that the error message does not appear in the logs? That's what the integration test here tested, but that is not present in the unit test.

I was wrong, it was being checked...

jonathan-eq (approved):
LGTM!

Issue #7676, opened 2024-04-16
LSF driver tries to kill finished realizations


jonathan-eq:
**Describe the bug**
When ctrl-c-ing ("Terminate experiment") is clicked on an ensemble_experiment (add time.sleep(10) to poly_eval.py) on an LSF run, there are messages in the log:
`Job <8056>: Job has already finished`. 
We should not run bkill on jobs that are already finished.
Very similar to https://github.com/equinor/ert/issues/7658

PR #7690, opened 2024-04-17
Do not cancel the returncode in scheduler

### LINKED ISSUE #7689: Scheduler should not depend on return_code for cancellation
Return codes from jobs through drivers currently determine the code path through scheduler:

https://github.com/equinor/ert/blob/5ba5b091377fd5d5695597a43594c1d231b0dc53/src/ert/scheduler/scheduler.py#L308-L312

This should not be the case. The effect of obtaining cancellation from the returncode is that the job will not be retried, and we currently don't want that under control of the job itself. 

Removing the branching on returncode can be done, pending a fix of the `mock_driver._kill()` function

berland:
Cancellation based on return_code can lead to jobs being able to circumvent resubmission based on return code. This is not functionality we want.

**Issue**
Resolves #7689 

**Approach**
just-do-it

This does not change any resubmission behaviour from before and after this PR on local driver or LSF driver, so this is not a bug fix.

## When applicable

### DIFF
--- a/src/ert/scheduler/scheduler.py
+++ b/src/ert/scheduler/scheduler.py
@@ -32,7 +32,7 @@ from ert.job_queue.queue import (
     EVTYPE_ENSEMBLE_CANCELLED,
     EVTYPE_ENSEMBLE_STOPPED,
 )
-from ert.scheduler.driver import SIGNAL_OFFSET, Driver
+from ert.scheduler.driver import Driver
 from ert.scheduler.event import FinishedEvent
 from ert.scheduler.job import Job
 from ert.scheduler.job import State as JobState
@@ -305,11 +305,12 @@ class Scheduler:
             # Any event implies the job has at least started
             job.started.set()
 
-            if isinstance(event, FinishedEvent):
-                if event.returncode >= SIGNAL_OFFSET:
-                    job.returncode.cancel()
-                else:
-                    job.returncode.set_result(event.returncode)
+            if (
+                isinstance(event, FinishedEvent)
+                and not self._cancelled
+                and not job.returncode.cancelled()
+            ):
+                job.returncode.set_result(event.returncode)
 
     def _update_jobs_json(self, iens: int, runpath: str) -> None:
         cert_path = f"{runpath}/{CERT_FILE}"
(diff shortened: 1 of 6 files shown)


src/ert/scheduler/scheduler.py
```diff
@@ -306,9 +306,7 @@ async def _process_event_queue(self) -> None:
             job.started.set()
 
             if isinstance(event, FinishedEvent):
-                if event.returncode >= SIGNAL_OFFSET:
-                    job.returncode.cancel()
-                else:
+                with suppress(asyncio.InvalidStateError):
```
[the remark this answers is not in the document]
jonathan-eq: No, I think it looks fine👌

jonathan-eq (approved):
Very good! 🦖

src/ert/scheduler/scheduler.py
```diff
@@ -305,11 +305,12 @@ async def _process_event_queue(self) -> None:
...
 
-            if isinstance(event, FinishedEvent):
-                if event.returncode >= SIGNAL_OFFSET:
-                    job.returncode.cancel()
-                else:
-                    job.returncode.set_result(event.returncode)
+            if (
+                isinstance(event, FinishedEvent)
+                and not self._cancelled
+                and not job.returncode.cancelled()
```
xjules, line 311: Nice!

tests/unit_tests/scheduler/test_scheduler.py
```diff
@@ -334,18 +334,27 @@ async def kill():
 
 
 @pytest.mark.timeout(6)
-async def test_that_job_does_not_retry_when_killed(realization, mock_driver):
+async def test_that_job_does_not_retry_when_killed_by_scheduler(
```
xjules, line 337: `killed_by_cancellation` ?

xjules (approved):
Very nice @berland ! Just had a small comment about the test function name. Otherwise very good!

tests/unit_tests/scheduler/test_scheduler.py
berland: "killed_by_cancellation" sounds more like we have cancelled the job task in the test, but rather we are using Schedulers API to kill the job tasks (that Scheduler is using cancellation to do the killing is an implementation detail) (there is another `test_cancel` that steps into internals of the scheduler and issues cancellation on its own).

berland:
> Did this also fix #7676?

No guarantees.

Issue #7642, opened 2024-04-11
No running event loop error observed when doing heavy update


berland:
Observed while running the poly-case on version 9.0.4 on 1000 realizations, during the update step:

```
(2024.04.rc0-py38) [havb@be-linrgsn001:~/projects/ert/test-data/poly_example] main$ ert gui --enable-scheduler poly.ert
Exception ignored in: <coroutine object WebSocketCommonProtocol.close_connection at 0x7f2c8c56c340>
Traceback (most recent call last):
  File "/prog/res/komodo/2024.04.rc0-py38-rhel7/root/lib64/python3.8/site-packages/websockets/legacy/protocol.py", line 1337, in close_connection
    await self.close_transport()
  File "/prog/res/komodo/2024.04.rc0-py38-rhel7/root/lib64/python3.8/site-packages/websockets/legacy/protocol.py", line 1355, in close_transport
    if await self.wait_for_connection_lost():
  File "/prog/res/komodo/2024.04.rc0-py38-rhel7/root/lib64/python3.8/site-packages/websockets/legacy/protocol.py", line 1379, in wait_for_connection_lost
    async with asyncio_timeout(self.close_timeout):
  File "/prog/res/komodo/2024.04.rc0-py38-rhel7/root/lib64/python3.8/site-packages/websockets/legacy/async_timeout.py", line 74, in timeout
    loop = asyncio.get_running_loop()
RuntimeError: no running event loop
```

This is probably the Ensemble Evaluator that is too busy doing the maths for the update from an iteration to the next, and not setting aside time to maintain the websocket connection, and then some bad handling of this problem. 

The poly case progresses fine so this is not an Error, just a Warning. But this should not be shown to users.

berland:
Related to #7275

xjules:
Might be related to this one: https://github.com/equinor/komodo-releases/actions/runs/9105017015/job/25030169047
```
File "/prog/komodo/bleeding-py38-rhel7/root/lib64/python3.8/site-packages/ert/job_queue/queue.py", line 277, in _jobqueue_publisher
    change = await self._changes_to_publish.get()
  File "/opt/rh/rh-python38/root/usr/lib64/python3.8/asyncio/queues.py", line 165, in get
    getter.cancel()  # Just in case getter is not done yet.
  File "/opt/rh/rh-python38/root/usr/lib64/python3.8/asyncio/base_events.py", line 719, in call_soon
    self._check_closed()
[...]

PR #7693, opened 2024-04-18
Suppress RuntimeError in EE websocket

jonathan-eq:
**Issue**
Connected to #7642 

**Approach**
The commit in this PR should fix the issue where the websocket server would raise RuntimeError stating `no running event loop` under heavy load. The fix here is to wrap it in a try-except, and checking if the error message contains `no running event loop`.

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -18,7 +18,7 @@ from typing import (
 
 import cloudevents.exceptions
 import cloudpickle
-import websockets
+import websockets.server
 from cloudevents.conversion import to_json
 from cloudevents.http import CloudEvent, from_json
 from websockets.datastructures import Headers, HeadersLike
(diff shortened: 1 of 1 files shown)


jonathan-eq:
This was the suggested fix by chatgpt. We could also pass in our own logger and filter messages instead, but that would be more messy.

berland:
> It seems to be when we are exiting the websocket context, and it is trying to close the connections. At this point, all communication should be done; so we can ignore this failure. Are all realisations already finished when we get this error?

It was in between iterations, so the realizations for one ensemble was probably finished yes.

jonathan-eq:
> > It seems to be when we are exiting the websocket context, and it is trying to close the connections. At this point, all communication should be done; so we can ignore this failure. Are all realisations already finished when we get this error?
> 
> It was in between iterations, so the realizations for one ensemble was probably finished yes.

Every instance of the ensemble_evaluator creates its own event loop, so they should be independent. Maybe we just get unlucky every now and then?

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -331,59 +331,65 @@ async def process_request(
...
+                    extra_attrs=terminated_attrs,
+                    data_marshaller=cloudpickle.dumps,
                 )
+                if self._clients:
+                    # See note about return_exceptions=True above
+                    await asyncio.gather(
+                        *[client.send(message) for client in self._clients],
+                        return_exceptions=True,
+                    )
+        except RuntimeError as exc:
```
xjules, line 388: Can we check that self._done before we return?

berland:
I think I am uncomfortable with just muting this error like this, we don't know what else we are hiding.

jonathan-eq:
> I think I am uncomfortable with just muting this error like this, we don't know what else we are hiding.

Yes, we should fix the underlying problem instead.

PR #7720, opened 2024-04-22
Test PBS queues before initiating pytest

### LINKED ISSUE #7716: Fix integration tests on PBS cluster
The following tests are sometimes timing out  when running on real cluster. 
```sh
self = <selectors.EpollSelector object at 0x7fdea8bbcc10>, timeout = -1

    def select(self, timeout=None):
        if timeout is None:
            timeout = -1
        elif timeout <= 0:
            timeout = 0
        else:
            # epoll_wait() has a resolution of 1 millisecond, round away
            # from zero to wait *at least* timeout seconds.
            timeout = math.ceil(timeout * 1e3) * 1e-3
    
        # epoll_wait() expects `maxevents` to be greater than zero;
        # we want to make sure that `select()` can be called when no
        # FD is registered.
        max_ev = max(len(self._fd_to_key), 1)
    
        ready = []
        try:
>           fd_event_list = self._selector.poll(timeout, max_ev)
E           Failed: Timeout >360.0s

/usr/lib64/python3.8/selectors.py:468: Failed
=========================== short test summary info ============================
FAILED integration_tests/scheduler/test_generic_driver.py::test_submit[OpenPBSDriver] - Failed: Timeout >360.0s
FAILED integration_tests/scheduler/test_generic_driver.py::test_submit_something_that_fails[OpenPBSDriver] - Failed: Timeout >360.0s
FAILED integration_tests/scheduler/test_generic_driver.py::test_kill[OpenPBSDriver] - Failed: Timeout >360.0s
```
https://github.com/equinor/komodo-releases/actions/runs/8755056479/job/24055175088
https://github.com/equinor/komodo-releases/actions/runs/8755056479/job/24055168487

**Potential solution:**
Catch timeout exceptions and run against mocked driver afterwards.

berland:
**Issue**
Resolves #7716 

**Approach**
Use `qsub` in interactive mode (`-I`) to check that a queue is up an running within a short timeout. If a queue is unable to run the echo command within 20 seconds, the next queue is attempted. If no queue is deemed working, we should fail.

In interactive mode, the qsub command does not return until the job is finished and stdout/stderr is forwarded to your terminal.

## When applicable

### DIFF
--- a/ci/testkomodo.sh
+++ b/ci/testkomodo.sh
@@ -72,11 +72,12 @@ start_tests () {
     fi
     if ! which bsub 2>/dev/null && basetemp=$(mktemp -d -p ~/pytest-tmp); then
         export PATH=$PATH:/opt/pbs/bin
-        if [[ $(uname -r) == *el7* ]] ; then
-            export _ERT_TESTS_DEFAULT_QUEUE_NAME=permanent
-        else
-            export _ERT_TESTS_DEFAULT_QUEUE_NAME=permanent_8
-        fi
+        echo '#!/bin/sh' > queue_tester.sh
+        echo echo I am up >> queue_tester.sh
+        chmod 755 queue_tester.sh
+        for candidate in "permanent permanent_8 hb120 hb120_8 short normal"; do
+            timeout -k 1 20 qsub -I -q $candidate -- queue_tester.sh && export _ERT_TESTS_DEFAULT_QUEUE_NAME=$candidate && break
+        done
         pytest --timeout=3600 -v --openpbs --basetemp="$basetemp" integration_tests/scheduler
         rm -rf "$basetemp" || true
     fi


ci/testkomodo.sh
```diff
@@ -72,11 +72,12 @@ start_tests () {
...
         export PATH=$PATH:/opt/pbs/bin
-        if [[ $(uname -r) == *el7* ]] ; then
-            export _ERT_TESTS_DEFAULT_QUEUE_NAME=permanent
-        else
-            export _ERT_TESTS_DEFAULT_QUEUE_NAME=permanent_8
-        fi
+        echo '#!/bin/sh' > queue_tester.sh
+        echo echo I am up >> queue_tester.sh
+        chmod 755 queue_tester.sh
+        for candidate in "permanent permanent_8 hb120 hb120_8 short normal"; do
```
jonathan-eq, line 78: The old script checked for rhel8. Do we have still have to account for that?

xjules, line 78: shall we try the `short` first?

ci/testkomodo.sh
```diff
@@ -72,11 +72,12 @@ start_tests () {
     fi
     if ! which bsub 2>/dev/null && basetemp=$(mktemp -d -p ~/pytest-tmp); then
         export PATH=$PATH:/opt/pbs/bin
-        if [[ $(uname -r) == *el7* ]] ; then
-            export _ERT_TESTS_DEFAULT_QUEUE_NAME=permanent
-        else
-            export _ERT_TESTS_DEFAULT_QUEUE_NAME=permanent_8
-        fi
+        echo '#!/bin/sh' > queue_tester.sh
```
xjules, line 75: As discussed, it'd be better to have this test in `komodo-releases` and set the "working" queue as env or parameter.

ci/testkomodo.sh
berland: Yes we do. I had a hope tha this test would implicitly autodetect rhel7 vs rhel8 but I was not successful.

berland:
The functionality from this PR is moved to komodo-releases.

ci/testkomodo.sh
berland: The `permanent` queue was set up for testing like this. I do not remember anything about the `short` queue that makes it the first choice.

PR #7725, opened 2024-04-23
Improve scheduler driver logging to always include exit code, stdout, stderr

### LINKED ISSUE #7691: Fix LSF kill logging when matching the stdout
In LSF driver we match `stdout` when the `returncode==0` and we should perhaps do it when `returcode!=0` and we should add filtering of `stderr` also. This should include filtering of the message: 
`Job has already finished`

jonathan-eq:
**Issue**
Resolves #7691 

**Approach**
^

## When applicable

### DIFF
--- a/src/ert/scheduler/driver.py
+++ b/src/ert/scheduler/driver.py
@@ -85,7 +85,15 @@ class Driver(ABC):
             stdout, stderr = await process.communicate(stdin)
 
             assert process.returncode is not None
+            outputs = (
+                f"exit code {process.returncode}, "
+                f'output: "{stdout.decode(errors="ignore").strip() or "<empty>"}", and '
+                f'error: "{stderr.decode(errors="ignore").strip() or "<empty>"}"'
+            )
             if process.returncode == 0:
+                _logger.debug(
+                    f'Command "{shlex.join(cmd_with_args)}" succeeded with {outputs}'
+                )
                 return True, stdout.decode(errors="ignore").strip()
             if exit_on_msgs and any(
                 exit_on_msg in stderr.decode(errors="ignore")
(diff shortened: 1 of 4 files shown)


src/ert/scheduler/driver.py
```diff
@@ -84,25 +84,29 @@ async def _execute_with_retry(
...
 
             assert process.returncode is not None
+            outputs = (
+                f"exit code {process.returncode}, "
+                f'output: "{stdout.decode(errors="ignore").strip() or "<empty>"}", and '
+                f'error: "{stderr.decode(errors="ignore").strip() or "<empty>"}"'
+            )
             if process.returncode == 0:
-                return True, stdout.decode(errors="ignore").strip()
+                return True, outputs
```
xjules, line 93: Apparently this changes behavior of some tests.

jonathan-eq: Yes, I will fix those tests, but I am not sure it is a good idea to test the logging format..

src/ert/scheduler/driver.py
```diff
@@ -84,25 +84,35 @@ async def _execute_with_retry(
             stdout, stderr = await process.communicate(stdin)
 
             assert process.returncode is not None
+            outputs = (
+                f"exit code {process.returncode}, "
+                f'output: "{stdout.decode(errors="ignore").strip() or "<empty>"}", and '
```
xjules, line 89: why this weird formatting? I'd change make `" "` the outer ones. Also do we need `" "` inside the error message?

berland: I think I prefer including quotes (single or double) to wrap the stdout/stderr output to distinguish it from the log from our code.

xjules (approved):
I think it can go now! You need to fix some conflicts though. :rocket:

PR #7743, opened 2024-04-23
Add forward models as plugins

### LINKED ISSUE #7409: Make it possible for plugins to subtype ForwardModel
Currently, making e.g. [pyscal](https://github.com/equinor/semeio/blob/main/src/semeio/forward_models/scripts/fm_pyscal.py) form [semeio](https://github.com/equinor/semeio) available as a forward model in ert means creating a [config file](https://github.com/equinor/semeio/blob/main/src/semeio/forward_models/config/PYSCAL_CONFIG),  and pointing ert to it with the plugin interface method `installable_jobs()` (done [here](https://github.com/equinor/semeio/blob/c397b16cf5a23879c14f5d8fbea4dd3ace25fda5/src/semeio/hook_implementations/forward_models.py#L41)).

From there, it is inserted into the site config and read here: https://github.com/equinor/ert/blob/006499cff7d98620446c45872d8d92985dd512bb/src/ert/config/ert_config.py#L684-L708

This makes it difficult to add additional validation of the setup such as valid version number etc.

In order to remedy this, the suggestion is that the plugin interface is extended so that e.g. fm_pyscal can be implemented as a subtype of `ForwardModel`. The specifics of a forward model read from a config file would have to also be pushed out into subtype of `ForwardModel` e.g. `ConfigFileForwardModel`, so that `ForwardModel` is an abstract baseclass.

In this way the PYSCAL forward model could be implemented like this:

```python
class Pyscal(ForwardModel):
    def __init__(self):
        super().__init__(
            name="PYSCAL",
            executable="fm_pyscal",
        )

    def create_arguments(self, **parameters:str) -> List[str]:
        ...
```

where create_arguments is responsible for handling default values and validating parameters.

yngve-sk:
**Not about preinstalled forward model steps "bleeding" into other tests causing failure**
The preinstalled forward model steps are added as a global singleton `ClassVar` to `ErtConfig`, since plugins installed from imported system modules are inherently global. This does have some pitfalls when it comes to running tests, if a test context imports and thus also installs some plugin forward models to the `ErtConfig`, these will persist as classvars of the `ErtConfig` for subsequent test runs. So for example if we run 
```python
pm = ErtPluginManager(plugins=[dummy_plugins])
    run_cli_with_plugin_manager(
        [TEST_RUN_MODE, "--disable-monitor", str(tmp_path / "test.ert")], pm
    )
```
in one test, then the installed plugins will "show up" in subsequent tests, so this test would fail:
```python
def test_that_include_take_into_account_path():
    ...
    assert list(ert_config.installed_forward_model_steps.keys()) == [
        "job1",
        "job2",
    ]
```
because the `DummyForwardModel` installed in the previous step would also show up. However when we run ert, we only create one config and don't care about creating multiple configs with separate sets of extensions within the same python context. 

**Issue**
Resolves #7409


src/ert/config/ert_config.py
```diff
@@ -182,6 +187,10 @@ def from_dict(cls, config_dict) -> Self:
             installed_forward_model_steps = (
                 cls._installed_forward_model_steps_from_dict(config_dict)
             )
+
+            installed_forward_model_steps.update(
```
eivindjahren, line 191: This means that plugins overrwite user specified which I think should be the other way around.

src/ert/config/ert_config.py
```diff
@@ -441,6 +455,25 @@ def _create_list_of_forward_model_steps_to_run(
             fm_step.arglist = fm_step_description[1:]
             fm_steps.append(fm_step)
 
+        for fm_step in fm_steps:
+            if fm_step.name in cls.PREINSTALLED_FORWARD_MODEL_STEPS:
+                try:
+                    substituted_json = ErtConfig._forward_model_data_to_json(
+                        run_id="N/A",
```
eivindjahren, line 462: Should we use something like `run_id=None` here?

src/ert/config/ert_config.py
```diff
@@ -454,8 +487,31 @@ def forward_model_data_to_json(
         run_id: str,
         iens: int = 0,
         itr: int = 0,
+    ):
+        return self._forward_model_data_to_json(
```
eivindjahren, line 491: Maybe rename the private function to disambiguate.

src/ert/config/forward_model_step.py
```diff
@@ -18,6 +25,25 @@
 logger = logging.getLogger(__name__)
 
 
+class ForwardModelStepJSON(TypedDict):
```
eivindjahren, line 28: Since this becomes part of the public API, we probably need docstrings wherever we can have it in this file.

src/ert/config/forward_model_step.py
```diff
@@ -46,6 +72,46 @@ class ForwardModelStep:
...
+            self.stdout_file = None
+
+        if self.stderr_file is None:
+            self.stderr_file = f"{self.name}.stderr"
+        elif self.stderr_file == "null":
+            self.stderr_file = None
+
+        if self.stdin_file == "null":
+            self.stdin_file = None
+
```
eivindjahren, line 114: We probably don't want to expose the `from_config_file` classmethod. We can probably move this classmethod to `ert_config.py`.

src/ert/gui/main.py
```diff
@@ -96,6 +96,10 @@ def _start_initial_gui_window(
             # Changing current working directory means we need to update
             # the config file to be the base name of the original config
             args.config = os.path.basename(args.config)
+
+            if plugin_manager:
+                ErtConfig.add_preinstalled_fm_steps(plugin_manager.forward_model_steps)
```
eivindjahren, line 101: The main function is never called with a plugin_manager in tests, that is line 101 is not covered.

src/ert/shared/plugins/hook_specifications/forward_model_steps.py
```diff
@@ -0,0 +1,37 @@
...
+    Valid fields:
+    description: RST markdown as a string. Example: "This is a **dummy** description"
+    examples: RST markdown as a string. Example: "This is an example"
+    category: Dot separated list categories (main_category.sub_category)
+              for forward model.
+              Example: "simulator.reservoir". When generating documentation in ERT the
+              main category (category before the first dot) will be used to group
+              the forward models into sections.
+    """
+    pass
```
eivindjahren, line 37: Presence of docstring makes `pass` statement redundant.

src/ert/shared/plugins/hook_specifications/forward_model_steps.py
```diff
@@ -0,0 +1,37 @@
...
+from ert.shared.plugins.plugin_manager import hook_specification
+
+if TYPE_CHECKING:
+    from ert.shared.plugins.plugin_response import PluginResponse
+
+
+@no_type_check
+@hook_specification
+def installable_forward_model_steps() -> PluginResponse[List[Type[ForwardModelStep]]]:
+    pass
```
eivindjahren, line 15: hook_specifications need docstrings as they will be used by plugins.

src/ert/gui/main.py
yngve-sk: Added integration test to cover a cli run where there is a plugin manager with a plugin fm step

src/ert/config/ert_config.py
eivindjahren: Seems like this point isn't resolved yet.

src/ert/config/forward_model_step.py
```diff
@@ -46,69 +141,89 @@ class ForwardModelStep:
...
-            )
-        except IOError as err:
-            raise ConfigValidationError.with_context(str(err), config_file) from err
+    def validate_pre_experiment(self, fm_step_json: ForwardModelStepJSON) -> None:
+        """
+        Raise errors pertaining to the environment not being
+        as the forward model step requires it to be. For example
+        a missing FLOW version.
+        """
+        pass
```
eivindjahren, line 150: pass is redundant when there is a docstring :)

src/ert/shared/plugins/plugin_manager.py
```diff
@@ -80,6 +81,16 @@ def __str__(self) -> str:
...
         return ErtPluginManager._merge_dicts(self.hook.help_links())
 
+    @property
+    def forward_model_steps(
+        self,
+    ) -> List[Type[ForwardModelStep]]:
+        fm_steps_listed = [
+            resp.data for resp in self.hook.installable_forward_model_steps()
+        ]
+        all_fm_steps = [fm_step for fm_steps in fm_steps_listed for fm_step in fm_steps]
```
eivindjahren, line 91:
```
        return [fm_step.data for fm_steps in self.hook.installable_forward_model_steps() for fm_step in fm_steps]
```

tests/integration_tests/test_cli.py
```diff
@@ -24,9 +28,30 @@
     ITERATIVE_ENSEMBLE_SMOOTHER_MODE,
     TEST_RUN_MODE,
 )
+from ert.shared.plugins import ErtPluginManager
 from ert.storage import open_storage
 
-from .run_cli import run_cli
+from ..unit_tests.all.plugins import dummy_plugins
```
eivindjahren, line 34:
```suggestion
from tests.unit_tests.all.plugins import dummy_plugins
```

tests/unit_tests/config/test_forward_model.py
```diff
@@ -305,3 +307,300 @@ def test_that_eclipse100_require_version_field():
...
+
+
+def test_that_plugin_forward_models_are_installed():
+    test_config_contents = dedent(
+        """
+        NUM_REALIZATIONS  1
+        FORWARD_MODEL PluginForwardModel(<arg1>=hello,<arg2>=world,<arg3>=derpyderp)
+        """
+    )
+    with open("test.ert", "w", encoding="utf-8") as fh:
```
eivindjahren, line 319: This will write the config in the root directory of the project. You need to use either the `use_tmpdir` fixture or `tmpdir` or `tmp_path`. I suggest the following:

```python
def test_that_plugin_forward_models_are_installed(tmp_path):
   (tmp_path / "config.ert").write_text(dedent(
        """
        NUM_REALIZATIONS  1
        FORWARD_MODEL PluginForwardModel(<arg1>=hello,<arg2>=world,<arg3>=derpyderp)
        """
        )
    )
```

tests/unit_tests/config/test_forward_model.py
```diff
@@ -305,3 +307,300 @@ def test_that_eclipse100_require_version_field():
...
+
+
+def test_that_plugin_forward_model_validation_failure_propagates():
+    test_config_contents = dedent(
+        """
+        NUM_REALIZATIONS  1
+        FORWARD_MODEL PluginFM(<arg1>=hello,<arg2>=world,<arg3>=derpyderp)
+        """
+    )
+    with open("test.ert", "w", encoding="utf-8") as fh:
```
eivindjahren, line 397: Same problem about putting `test.ert` in the root directory of the project.

tests/unit_tests/config/test_forward_model.py
```diff
@@ -305,3 +307,300 @@ def test_that_eclipse100_require_version_field():
...
+
+
+def test_that_plugin_forward_model_validation_accepts_valid_args():
+    test_config_contents = dedent(
+        """
+        NUM_REALIZATIONS  1
+        FORWARD_MODEL FM(<arg1>=never,<arg2>=world,<arg3>=derpyderp)
+        """
+    )
+    with open("test.ert", "w", encoding="utf-8") as fh:
```
eivindjahren, line 434: Same problem about putting test.ert in the root directory of the project.

tests/unit_tests/config/test_forward_model.py
```diff
@@ -305,3 +307,300 @@ def test_that_eclipse100_require_version_field():
...
+
+def test_that_plugin_forward_model_raises_pre_experiment_validation_error_early():
+    test_config_contents = dedent(
+        """
+        NUM_REALIZATIONS  1
+        FORWARD_MODEL FM1(<arg1>=never,<arg2>=world,<arg3>=derpyderp)
+        FORWARD_MODEL FM2
+        """
+    )
+    with open("test.ert", "w", encoding="utf-8") as fh:
```
eivindjahren, line 520: Same problem about putting test.ert in the root directory of the project.

tests/unit_tests/config/test_forward_model.py
```diff
@@ -305,3 +307,300 @@ def test_that_eclipse100_require_version_field():
...
+
+        DEFINE <yo> dear
+        DEFINE <dawg> good
+        DEFINE <iherdulike> solonius
+        DEFINE <some_var> schmidt
+
+        FORWARD_MODEL FM1(<arg1>=<yo>,<arg2>=<dawg>,<arg3>=<iherdulike>)
+        """
+    )
+    with open("test.ert", "w", encoding="utf-8") as fh:
```
eivindjahren, line 561: Same problem about putting test.ert in the root directory of the project.

tests/unit_tests/config/test_forward_model.py
yngve-sk: ooopsie! fixing

src/ert/shared/plugins/plugin_manager.py
yngve-sk: Can rewrite into something closer to a one-liner, suggestion _ almost_ works but needs this tweak:
```
return [
            fm_step
            for fm_steps in [
                x.data for x in self.hook.installable_forward_model_steps()
            ]
            for fm_step in fm_steps
        ]
``` 
as `data` is fetched out of every entry in `self.hook.installable_forward_model_steps()` which wraps the fm step class in a `PluginResponse`

src/ert/config/ert_config.py
```diff
@@ -530,53 +609,70 @@ def filter_env_dict(self, d):
...
+                ],
+                "environment": substituter.filter_env_dict(fm_step.environment),
+                "exec_env": substituter.filter_env_dict(fm_step.exec_env),
+                "max_running_minutes": fm_step.max_running_minutes,
+            }
+
+            try:
+                if not skip_pre_experiment_validation:
+                    fm_step_json = fm_step.validate_pre_realization_run(fm_step_json)
+            except BaseException as exc:
```
eivindjahren, line 658: Instead of `BaseException` I think we should create a specific exception that forward model validation can raise to indicate failed validation. This is in order to avoid catching things like keyboard interrupt (Ctrl-C) or AttributeError (something is wrong with the plugin implementation):
```
class InvalidForwardModelCall(Exception):
    """ Thrown when the user calls the forward model incorrectly.
          
          Can be subtyped by the implementation of ForwardModelStepPlugin and
          thrown from `validate_pre_realization_run` or `validate_pre_experiment`.
    """

File: src/ert/config/forward_model_step.py [hunk 18]
```diff
@@ -1,25 +1,120 @@
...
 logger = logging.getLogger(__name__)
 
 
+class ForwardModelStepJSON(TypedDict):
+    """
+    A dictionary containing information about how a forward model step should be run
+    on the queue.
+
+    Attributes:
+        name: A string representing the name of the job
```
eivindjahren, line 25: name is obvious so maybe just omit  or at least make the description shorter.

src/ert/config/forward_model_step.py
```diff
@@ -1,25 +1,120 @@
...
+        stdout: File where this forward model step's stdout is written
+        stderr: File where this forward model step's stderr is written
+        stdin: File where this forward model step's stdin is written
+        argList: List of command line arguments to be given to the executable.
+        environment: Dictionary of environment variables to inject into the
+            environment of the forward model step run
+        exec_env: Dictionary of environment variables to inject into the execution
+            environment of RMS
+        max_running_minutes: Maximum runtime in minutes. If the forward model step
+            takes longer than this, the job is requested to be cancelled by the queue
```
eivindjahren, line 45: Instead of "cancelled by the queue driver." I think it should say "cancelled."

src/ert/config/forward_model_step.py
```diff
@@ -46,69 +141,88 @@ class ForwardModelStep:
...
+        if self.stdin_file == "null":
+            self.stdin_file = None
+
+
+class ForwardModelStepPlugin(ForwardModelStep):
+    def __init__(
+        self,
+        name: str,
+        command: List[str],
+        options: Optional[ForwardModelStepOptions] = None,
```
eivindjahren, line 189: `options` here seems like a typical use case for https://typing.readthedocs.io/en/latest/spec/callables.html#unpack-kwargs . I suggest we put in `**kwargs: Unpack[ForwardModelStepOptions]` instead. If a plugin wants to use type the call to their `ForwardModelStepPlugin` we also need to expose `ForwardModelStepOptions` through `ert/__init__.py`.

src/ert/shared/plugins/hook_specifications/forward_model_steps.py
```diff
@@ -0,0 +1,42 @@
...
+
+@no_type_check
+@hook_specification
+def installable_forward_model_steps() -> (
+    PluginResponse[List[Type[ForwardModelStepPlugin]]]
+):
+    """
+    :return: List of forward model step plugins in the form of subclasses of the
+        ForwardModelStepPlugin class
+    :rtype: PluginResponse with data as List[Type[ForwardModelStepPlugin]]
```
eivindjahren, line 20: :rtype: is oldschool, we use typehints now, so just remove it.

src/ert/shared/plugins/hook_specifications/forward_model_steps.py
```diff
@@ -0,0 +1,42 @@
...
+@hook_specification(firstresult=True)
+def forward_model_step_documentation(
+    forward_model_step_name: str, description: str, examples: str, category: str
+) -> PluginResponse[Optional[Dict[str, str]]]:
+    """
+    :return: If forward_model_step_name is from your plugin return
+             dict with documentation fields as keys and corresponding
+             text as value (See below for details), else None.
+
+    Valid fields:
```
eivindjahren, line 34: I don't think this is valid google style docstring: https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html#example-google so it won't be possible to put it into the docs. Which is something we should do in order to document how to make a plugin for third parties.

tests/unit_tests/config/test_forward_model.py
```diff
@@ -305,3 +307,304 @@ def test_that_eclipse100_require_version_field():
...
+                "-c",
+                "good",
+                "solonius",
+                "schmidt",
+                "iter",
+            ]
+
+            # It is in the command, but not in the forward model(...) entry in the
+            # ert config. Thus it is not a "private" arg in that sense.
+            # Note2reviewer: We COULD include references to variables in the
```
eivindjahren, line 597: Maybe remove Note2reviewer and put that in github comments instead :D

yngve-sk: Ohh, I was trying to say that private_args will only include substituted args that are declared in the ert config, but not necessarily all the args in arglist.

So if arglist (in the fm step specification) is `[x,y,z]`, but the forward model invocation in the ert config is this `FORWARD_MODEL FM(<x>=2)`, then `self.private_args` will only contain `<x>`, and the concern was whether developers implementing a forward model would be interested in what `y` and `z` _would_ be substituted to, even though it isn't specified in the forward model step invocation in the ert config.

yngve-sk: So my question/concern was whether it was important to tweak the substitution logic to somehow make the substituted args from the entire arglist available, for validation purposes. If a developer implementing a fm step plugin says that `X` is in the arglist, and in the pre-experiment / pre-realization hook wants to know what it would be substituted with, regardless of whether it is invoked in the ert config fm step. I would assume this is never the case but am not completely sure.

src/ert/config/ert_config.py
yngve-sk: I added it to ert, called it `ForwardModelInvalidStepCall`, if you insist I will go with  `InvalidForwardModelCall` but it is nice to be able to just type `Fo..`, in the IDE and get up all the forwardmodel stuff on autocomplete.

src/ert/config/forward_model_step.py
yngve-sk: Removing

src/ert/shared/plugins/hook_specifications/forward_model_steps.py
yngve-sk: Fixed, the comment now corresponds to the google docstring style

src/ert/config/forward_model_step.py
yngve-sk: Did not know about `Unpack`, so nice, updated semeio plugins accordingly https://github.com/equinor/semeio/pull/613/commits/055f28ca0aafc7cf8e86678675cc29d3f9e1ece5

tests/unit_tests/config/test_forward_model.py
eivindjahren: We looked into this a while ago and we are pretty much unable to check the entire arglist because it seems that some user scripts handle unsubstituted arguments on their own.

src/ert/config/ert_config.py
eivindjahren: :+1: You also need to just catch that exception when you call `validate_pre_realization_run`.

src/ert/shared/plugins/hook_specifications/forward_model_steps.py
eivindjahren: We will create an issue to get the plugin docstrings into the docs after this then.

src/ert/config/forward_model_step.py
```diff
@@ -1,25 +1,126 @@
...
+    max_running_minutes: int
+
+
+class ForwardModelStepOptions(TypedDict, total=False):
+    stdin_file: Optional[str]
+    stdout_file: Optional[str]
+    stderr_file: Optional[str]
+    start_file: Optional[str]
+    target_file: Optional[str]
+    error_file: Optional[str]
```
eivindjahren, line 77: With TypedDict you need to use `NotRequired to indicate that the dictionary does not need to contain the keyword: https://typing.readthedocs.io/en/latest/spec/callables.html#required-and-non-required-keys .

src/ert/config/__init__.py
```diff
@@ -52,6 +57,9 @@
     "ExternalErtScript",
     "Field",
     "ForwardModelStep",
+    "ForwardModelStepPlugin",
+    "ForwardModelStepJSON",
+    "ForwardModelInvalidCallError",
```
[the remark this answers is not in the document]
yngve-sk: Yep, I think `ForwardModelValidationError` is nice and more in-tune with other naming we currently have in ert, and is easier to find/lookup when looking for stuff related to `ForwardModel`. Also personal preference but I think having `Error` postfix is nice, esp for just looking up stuff when developing. Just typing in something like `fmer` would find a fuzzy match on this, but on InvalidForwardModelCall it'd be a bit further away from forward model in a fuzzy-textual manner.

 Also the errors we raise in the validation steps `pre_experiment` and `pre_realization_run` are per definition validation errors. What exactly they validate (state or input arguments to the forward model) is arbitrary, but they are still distinct from errors that arise during the run of the forward model.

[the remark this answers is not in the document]
yngve-sk: I actually think it should be `ForwardModelStepValidationError` just to be explicit about the forward model vs forward model step, in spite of it being a bit more verbose

eivindjahren (approved):
:+1:

PR #7744, opened 2024-04-23
Improve codecov coverage lsf_driver polling and using bhist

jonathan-eq:

## When applicable

### DIFF
--- a/tests/integration_tests/scheduler/test_lsf_driver.py
+++ b/tests/integration_tests/scheduler/test_lsf_driver.py
@@ -1,5 +1,6 @@
 import asyncio
 import json
+import logging
 import os
 import stat
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


berland:
The way I read the existing test code, it _should_ trigger the fallback, but if it does not there might be something wrong with it?

xjules:
It seems that there are still some lines not covered: https://app.codecov.io/gh/equinor/ert/commit/20c62b6965d97318990927be49eeec3acedb4da2/blob/src/ert/scheduler/lsf_driver.py

berland:
So the changes in this commit only serves to verify that the test works as intended? (since bjobs is non-functional, the intention was that the poll function would not work unless bhist is actually working)

jonathan-eq:
> So the changes in this commit only serves to verify that the test works as intended? (since bjobs is non-functional, the intention was that the poll function would not work unless bhist is actually working)

Yes. Maybe the more explicit asserts will have codecov not flag the code as untested.

berland:
> Yes. Maybe the more explicit asserts will have codecov not flag the code as untested.

That would be surprising..

PR #7757, opened 2024-04-24
Implement OpenPBS queue option`QSTAT_OPTIONS`

### LINKED ISSUE #7747: OpenPBS driver lacks options the documentation claims it has
The OpenPBS driver that has replaced the Torque driver (unless  `--disable-scheduler` is used on the command line) lacks some settings that the documentation claims are valid:

* `QSUB_CMD` 
* `QSTAT_CMD`
* `QDEL_CMD`
* `QSTAT_OPTIONS`
* `QUEUE_QUERY_TIMEOUT`

The last option is not as relevant to the improved Python driver and should just be provided a deprecation warning for (that it is ignored).

The first four should either be deprecated, or implemented (which will not be difficult, and would make it more in line with LSF driver).

Documentation should be updated accordingly.

jonathan-eq:
**Issue**
Resolves #7747

**Approach**
Implement OpenPBS queue option`QSTAT_OPTIONS`

## When applicable

### DIFF
--- a/src/ert/config/parsing/config_schema_deprecations.py
+++ b/src/ert/config/parsing/config_schema_deprecations.py
@@ -202,4 +202,11 @@ deprecated_keywords_list = [
         "up on hanging backend (TORQUE/PBS) when submitting jobs or job status querying.",
         check=lambda line: "QUEUE_QUERY_TIMEOUT" in line,
     ),
+    DeprecationInfo(
+        keyword="QUEUE_OPTION",
+        message="QSTAT_OPTIONS as QUEUE_OPTION to the TORQUE system will be ignored "
+        "when using the scheduler, and it is not recommended to use this QUEUE_OPTION. "
+        "It has been used in the past to supply options to the qstat command.",
+        check=lambda line: "QSTAT_OPTIONS" in line,
+    ),
 ]


jonathan-eq:
ERT Docs: ```Options to be supplied to the qstat command. This defaults to -x, which tells the qstat command to include exited processes.```

We also `-E` as default qstat option in scheduler. Should I have this be overridden by the queue_option too? Also, what to do about the docs? The default is still only `-x` in the C implementation.

jonathan-eq:
We use different qstat options on the initial and second qstat querying. Maybe we should deprecate this QUEUE_OPTION instead.

xjules:
> We use different qstat options on the initial and second qstat querying. Maybe we should deprecate this QUEUE_OPTION instead.

I guess there might be a point where QSTAT_OPTIONS might be useful, but let's do it when the time comes.

berland:
I think we can deprecate this, if not we are giving the user the possibility of crashing the driver in arbitrary ways.

berland:
> We also `-E` as default qstat option in scheduler. Should I have this be overridden by the queue_option too? 

The `-E` option "should" have been injected into the legacy driver, but we can choose to not care as it is not used anymore.

berland (approved):
:+1: 🔪

PR #7760, opened 2024-04-24
Reverse ensemble order for plotting

andreas-el:
Response to user opinion:

> User opinion here: I would always opt to arrange the plots so that the latest iteration is on top. We expect to see lower variability as we progress through the iterations, and we are typically more interested in observing the behavior of the posterior. The idea is to have a plot where we see how the posterior converges closer to the observed data compared to the priors, which are usually more spread out in the background of the posterior curves. 

ref.
https://equinor.slack.com/archives/C01MCMWD075/p1713877068266489

Screenshot 2024-04-24 at 14 26 00

## When applicable


dafeda (approved):
Very nice 👍

PR #7762, opened 2024-04-24
Have RealizationDelegate be hover-able

jonathan-eq:
**Issue**
Resolves https://github.com/equinor/scout/issues/1054

**Approach**

## When applicable

### DIFF
--- a/src/ert/gui/model/snapshot.py
+++ b/src/ert/gui/model/snapshot.py
@@ -24,6 +24,7 @@ ProgressRole = Qt.UserRole + 5
 FileRole = Qt.UserRole + 6
 RealIens = Qt.UserRole + 7
 IterNum = Qt.UserRole + 12
+MemoryUsageRole = Qt.UserRole + 13
 
 # Indicates what type the underlying data is
 IsEnsembleRole = Qt.UserRole + 8
@@ -211,9 +212,21 @@ class SnapshotModel(QAbstractItemModel):
                 if job.index:
                     job_node.data[ids.INDEX] = job.index
                 if job.current_memory_usage:
-                    job_node.data[ids.CURRENT_MEMORY_USAGE] = job.current_memory_usage
+                    current_memory_usage = float(job.current_memory_usage)
+                    job_node.data[ids.CURRENT_MEMORY_USAGE] = current_memory_usage
+                    real_node.data[ids.CURRENT_MEMORY_USAGE] = current_memory_usage
+                    self.root.data[ids.CURRENT_MEMORY_USAGE] = current_memory_usage
                 if job.max_memory_usage:
-                    job_node.data[ids.MAX_MEMORY_USAGE] = job.max_memory_usage
+                    max_memory_usage = float(job.max_memory_usage)
+                    job_node.data[ids.MAX_MEMORY_USAGE] = max_memory_usage
+                    real_node.data[ids.MAX_MEMORY_USAGE] = max(
+                        real_node.data.get(ids.MAX_MEMORY_USAGE, -1),
+                        max_memory_usage,
+                    )
+                    self.root.data[ids.MAX_MEMORY_USAGE] = max(
+                        self.root.data.get(ids.MAX_MEMORY_USAGE, -1),
+                        max_memory_usage,
+                    )
 
                 # Errors may be unset as the queue restarts the job
                 job_node.data[ids.ERROR] = job.error if job.error else ""
@@ -394,6 +407,11 @@ class SnapshotModel(QAbstractItemModel):
             return node.data[REAL_STATUS_COLOR]
         if role == StatusRole:
             return node.data[ids.STATUS]
+        if role == MemoryUsageRole:
+            return (
+                node.data.get(ids.CURRENT_MEMORY_USAGE),
+                node.data.get(ids.MAX_MEMORY_USAGE),
+            )
         return QVariant()
 
     @staticmethod
--- a/src/ert/gui/simulation/run_dialog.py
+++ b/src/ert/gui/simulation/run_dialog.py
@@ -27,11 +27,17 @@ from ert.ensemble_evaluator import (
     FullSnapshotEvent,
     SnapshotUpdateEvent,
 )
+from ert.ensemble_evaluator import identifiers as ids
 from ert.gui.ertnotifier import ErtNotifier
 from ert.gui.ertwidgets.message_box import ErtMessageBox
 from ert.gui.model.job_list import JobListProxyModel
 from ert.gui.model.progress_proxy import ProgressProxyModel
-from ert.gui.model.snapshot import FileRole, IterNum, RealIens, SnapshotModel
+from ert.gui.model.snapshot import (
+    FileRole,
+    IterNum,
+    RealIens,
+    SnapshotModel,
+)
 from ert.gui.tools.file import FileDialog
 from ert.gui.tools.plot.plot_tool import PlotTool
 from ert.run_models import (
@@ -42,7 +48,7 @@ from ert.run_models import (
     RunModelUpdateEndEvent,
 )
 from ert.run_models.event import RunModelErrorEvent
-from ert.shared.status.utils import format_running_time
+from ert.shared.status.utils import byte_with_unit, format_running_time
 
 from .queue_emitter import QueueEmitter
 from .view import LegendView, ProgressView, RealizationWidget, UpdateWidget
@@ -122,6 +128,7 @@ class RunDialog(QDialog):
         self._job_view.verticalHeader().setMinimumWidth(20)
 
         self.running_time = QLabel("")
+        self.memory_usage = QLabel("")
 
         self.plot_tool = PlotTool(config_file, self.parent())
         self.plot_button = QPushButton(self.plot_tool.getName())
@@ -150,6 +157,8 @@ class RunDialog(QDialog):
         button_layout = QHBoxLayout()
         button_layout.addWidget(self.processing_animation)
         button_layout.addWidget(self.running_time)
+        button_layout.addStretch()
+        button_layout.addWidget(self.memory_usage)
         button_layout.addStretch()
         button_layout.addWidget(self.show_details_button)
         button_layout.addWidget(self.plot_button)
@@ -339,6 +348,16 @@ class RunDialog(QDialog):
         runtime = self._run_model.get_runtime()
         self.running_time.setText(format_running_time(runtime))
 
+        current_memory_usage = self._snapshot_model.root.data.get(
+            ids.CURRENT_MEMORY_USAGE
+        )
+        maximum_memory_usage = self._snapshot_model.root.data.get(ids.MAX_MEMORY_USAGE)
+
+        if current_memory_usage and maximum_memory_usage:
+            self.memory_usage.setText(
+                f"Memory Usage: [Current: {byte_with_unit(current_memory_usage)}, Max: {byte_with_unit(maximum_memory_usage)}]"
+            )
+
     @Slot(object)
     def _on_event(self, event: object):
         if isinstance(event, EndEvent):
@@ -357,7 +376,6 @@ class RunDialog(QDialog):
                     total_progress=progress, phase_name=event.phase_name
                 )
             )
-
         elif isinstance(event, SnapshotUpdateEvent):
             if event.partial_snapshot is not None:
                 self._snapshot_model._add_partial_snapshot(
--- a/src/ert/gui/simulation/view/realization.py
+++ b/src/ert/gui/simulation/view/realization.py
@@ -1,7 +1,7 @@
 import math
 from typing import Final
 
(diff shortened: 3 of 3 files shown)


jonathan-eq:
image

src/ert/gui/simulation/run_dialog.py
```diff
@@ -339,6 +349,21 @@ def _on_ticker(self):
...
+        )
+
+        if maximum_memory_usage:
+            self.maximum_memory_usage.setText(
+                f"Maximum memory usage: {byte_with_unit(float(maximum_memory_usage))}"
+            )
+
+        if current_memory_usage:
+            self.current_memory_usage.setText(
+                f"Current memory usage: {byte_with_unit(float(current_memory_usage))}"
```
xjules, line 364: Check if both all float already.

jonathan-eq: It did :)

src/ert/gui/simulation/view/realization.py
```diff
@@ -153,3 +170,26 @@ def _paint_inner_grid(painter: QPainter, rect: QRect, colors) -> None:
...
+            model: QAbstractItemModel = view.model()
+            for row in range(model.rowCount()):
+                index = model.index(row, 0)
+                rect = view.visualRect(index)
+                if rect.contains(mouse_pos):
+                    (current_memory_usage, maximum_memory_usage) = index.data(
+                        MemoryUsageRole
+                    )
+                    if maximum_memory_usage:
+                        txt = (
```
xjules, line 188: I would rename this variable to `mem_msg` or something along those lines.

src/ert/gui/simulation/view/realization.py
```diff
@@ -153,3 +170,26 @@ def _paint_inner_grid(painter: QPainter, rect: QRect, colors) -> None:
...
+    def eventFilter(self, watched, event: QEvent):
+        if event.type() == QEvent.Type.ToolTip:
+            mouse_pos = event.pos()
+            parent: RealizationWidget = self.parent()
+            view = parent._real_view
+            model: QAbstractItemModel = view.model()
+            for row in range(model.rowCount()):
+                index = model.index(row, 0)
+                rect = view.visualRect(index)
+                if rect.contains(mouse_pos):
```
xjules, line 183: Is there a better way to get the current item than just loop over all the models?

jonathan-eq: I am not sure, but I will figure it out. Things like this in Qt seem overly complicated.

src/ert/gui/model/snapshot.py
```diff
@@ -211,9 +212,27 @@ def _add_partial_snapshot(self, partial: PartialSnapshot, iter_: int):
...
+                        job.current_memory_usage
+                    )
+                    real_node.data[ids.CURRENT_MEMORY_USAGE] = max(
+                        real_node.data.get(ids.CURRENT_MEMORY_USAGE, -1),
+                        float(job.current_memory_usage),
+                    )
+                    self.root.data[ids.CURRENT_MEMORY_USAGE] = max(
+                        self.root.data.get(ids.CURRENT_MEMORY_USAGE, -1),
+                        float(job.current_memory_usage),
+                    )
```
andreas-el, line 225: I assume the two first are for the different views, but what is the last entry?

src/ert/gui/simulation/view/realization.py
```diff
@@ -153,3 +170,25 @@ def _paint_inner_grid(painter: QPainter, rect: QRect, colors) -> None:
...
     def sizeHint(self, option, index) -> QSize:
         return self._size
+
+    def eventFilter(self, watched, event: QEvent):
+        if event.type() == QEvent.Type.ToolTip:
+            mouse_pos = event.pos()
+            parent: RealizationWidget = self.parent()
+            view = parent._real_view
+            model: QAbstractItemModel = view.model()
+            for row in range(model.rowCount()):
```
andreas-el, line 180: There might be a way around this, could you look into the following;

```
        index = view.indexAt(mouse_pos)
        if index.isValid():
            mem_msg = index.data(MemoryUsageRole)
```

src/ert/gui/model/snapshot.py
andreas-el: Will the current memory usage ever go down with this? If you always select max, won't the current always be the same as maximum?

src/ert/gui/simulation/view/realization.py
```diff
@@ -153,3 +170,25 @@ def _paint_inner_grid(painter: QPainter, rect: QRect, colors) -> None:
...
+                index = model.index(row, 0)
+                rect = view.visualRect(index)
+                if rect.contains(mouse_pos):
+                    mem_msg = index.data(MemoryUsageRole)
+                    if mem_msg:
+                        (current_memory_usage, maximum_memory_usage) = mem_msg
+                        txt = (
+                            f"Current memory usage: {byte_with_unit(current_memory_usage)}\n"
+                            f"Maximum memory usage: {byte_with_unit(maximum_memory_usage)}"
```
andreas-el, line 189: Looks good, but have you played around with justification etc to see if you could align the lines better?
Either justifying or with tabs or something?
How does this look when using RGS onprem?

src/ert/gui/simulation/run_dialog.py
```diff
@@ -339,6 +349,21 @@ def _on_ticker(self):
...
+
+        if maximum_memory_usage:
+            self.maximum_memory_usage.setText(
+                f"Maximum memory usage: {byte_with_unit(maximum_memory_usage)}"
+            )
+
+        if current_memory_usage:
+            self.current_memory_usage.setText(
+                f"Current memory usage: {byte_with_unit(current_memory_usage)}"
+            )
```
andreas-el, line 365: Just a thought, have you considered other formats?
```
Memory Usage: [Current: 1204 kB, Max: 1434 kB]
Memory Usage: [1204 kB / 1434 kB]
```

andreas-el, line 365: This is extremely nitpick, but personally I think the logical order is first handle current, then the maximum. Could consider swapping the order here.

src/ert/gui/model/snapshot.py
jonathan-eq: Yes, you are right. This should only be the case for max memory usage.

src/ert/gui/simulation/run_dialog.py
xjules: This looks more PRO indeed 👍

src/ert/gui/simulation/view/realization.py
jonathan-eq: This is the best I got the alignment. It might be a different font on mac, because the alignment looked better there. 
image

jonathan-eq: I tried right aligning the text instead (not with the amount/unit), but I found that left alignment looked better.

jonathan-eq: This is the same, but with right alignment. Right alignment looked better on mac, and left alignment looked better on rgs node.
image

andreas-el: I think the formatting you have now is pretty close to perfect.
What bugs me a bit is that the tooltip seems to be handled incorrectly, in the way that sometimes you get it, and sometimes you don't. Could there be more events taking place?

jonathan-eq: Yes, it seems to be VERY flaky and random. I will ask chatgpt, maybe he knows a better way.

andreas-el: At least check how many times the hover-thing happens. It might be that you get multiple hits, or some with invalid indices.

src/ert/gui/simulation/view/realization.py
```diff
@@ -153,3 +169,22 @@ def _paint_inner_grid(painter: QPainter, rect: QRect, colors) -> None:
...
+            view = parent._real_view
+            index = view.indexAt(mouse_pos)
+            if index.isValid():
+                mem_msg = index.data(MemoryUsageRole)
+                (current_memory_usage, maximum_memory_usage) = mem_msg
+                if current_memory_usage and maximum_memory_usage:
+                    txt = (
+                        f"{'Current memory usage:':<24} {byte_with_unit(current_memory_usage)}\n"
+                        f"Maximum memory usage: {byte_with_unit(maximum_memory_usage)}"
+                    )
```
andreas-el, line 186:
```suggestion
                    txt = (
                        f"Current memory usage:\t {byte_with_unit(current_memory_usage)}\n"
                        f"Maximum memory usage:\t {byte_with_unit(maximum_memory_usage)}"
                    )
```

This yields perfect alignment on my setup

src/ert/gui/model/snapshot.py
```diff
@@ -211,9 +212,25 @@ def _add_partial_snapshot(self, partial: PartialSnapshot, iter_: int):
...
                 if job.current_memory_usage:
-                    job_node.data[ids.CURRENT_MEMORY_USAGE] = job.current_memory_usage
+                    job_node.data[ids.CURRENT_MEMORY_USAGE] = float(
+                        job.current_memory_usage
+                    )
+                    real_node.data[ids.CURRENT_MEMORY_USAGE] = float(
+                        job.current_memory_usage
+                    )
+                    self.root.data[ids.CURRENT_MEMORY_USAGE] = float(
+                        job.current_memory_usage
```
andreas-el, line 222: We could avoid casting the values three times over?

```suggestion
                    current_memory_usage = float(job.current_memory_usage)

                    job_node.data[ids.CURRENT_MEMORY_USAGE] = current_memory_usage
                    real_node.data[ids.CURRENT_MEMORY_USAGE] = current_memory_usage
                    self.root.data[ids.CURRENT_MEMORY_USAGE] = current_memory_usage
```

src/ert/gui/model/snapshot.py
```diff
@@ -211,9 +212,25 @@ def _add_partial_snapshot(self, partial: PartialSnapshot, iter_: int):
...
                 if job.max_memory_usage:
                     job_node.data[ids.MAX_MEMORY_USAGE] = job.max_memory_usage
+                    real_node.data[ids.MAX_MEMORY_USAGE] = max(
+                        real_node.data.get(ids.MAX_MEMORY_USAGE, -1),
+                        float(job.max_memory_usage),
+                    )
+                    self.root.data[ids.MAX_MEMORY_USAGE] = max(
+                        self.root.data.get(ids.MAX_MEMORY_USAGE, -1),
+                        float(job.max_memory_usage),
+                    )
```
andreas-el, line 233: Also here, and why didn't we cast for the first one?
```suggestion
                    max_memory_usage = float(job.max_memory_usage)
                    job_node.data[ids.MAX_MEMORY_USAGE] = max_memory_usage
                    real_node.data[ids.MAX_MEMORY_USAGE] = max(
                        real_node.data.get(ids.MAX_MEMORY_USAGE, -1), max_memory_usage)
                    self.root.data[ids.MAX_MEMORY_USAGE] = max(
                        self.root.data.get(ids.MAX_MEMORY_USAGE, -1), max_memory_usage)
```

src/ert/gui/simulation/view/realization.py
```diff
@@ -153,3 +162,24 @@ def _paint_inner_grid(painter: QPainter, rect: QRect, colors) -> None:
 
     def sizeHint(self, option, index) -> QSize:
         return self._size
+
+    def eventFilter(self, watched, event: QEvent):
+        if event.type() == QEvent.Type.ToolTip:
+            mouse_pos = event.pos() + QPoint(
+                -2 * self.job_rect_margin, -2 * self.job_rect_margin
```
andreas-el, line 169: Propose to move this `QPoint` as an instance variable, -- we don't need to recreate this object every time this function is called.

src/ert/gui/simulation/view/realization.py
```diff
@@ -153,3 +162,24 @@ def _paint_inner_grid(painter: QPainter, rect: QRect, colors) -> None:
...
+            )
+            parent: RealizationWidget = self.parent()
+            view = parent._real_view
+            index = view.indexAt(mouse_pos)
+            if index.isValid():
+                mem_msg = index.data(MemoryUsageRole)
+                (current_memory_usage, maximum_memory_usage) = mem_msg
```
andreas-el, line 176: Possible to combine? We don't use `men_msg` for anything else?

andreas-el (changes requested):
Consider final touchup

src/ert/gui/model/snapshot.py
jonathan-eq: Good question, I probably missed it.
Good suggestion!

src/ert/gui/simulation/view/realization.py
jonathan-eq: Agreed!

andreas-el (approved):
Mighty fine! ✈️

Issue #7763, opened 2024-04-25
Details view is garbled on realization resubmission, shows negative duration.


berland:
**Describe the bug**
Details view is wrong when realizations are resubmitted.  There is no visual indication to which realization rerun a forward model result belongs to, so failed step from the previous run looks valid during the execution of a rerun. Users find stderr files and get confused. Also, durations become negative.

**To reproduce**
Steps to reproduce the behaviour:
You need two forward models. First one job that does nothing but  `time.sleep(10)`

Then modify poly_eval.py to add some `time.sleep` in order to be able to follow the GUI, and then a code block to make it fail on the first try and then succeed, f.ex:

```
    if not Path("firstrounddone").exists():
        Path("firstrounddone").touch()
        sys.exit(1)

```
**Expected behaviour**
Durations should be correct. Results from forward model runs in the previous run should be removed or sufficiently tagged as such.

**Screenshots**

![Screenshot from 2024-04-25 07-28-25](https://github.com/equinor/ert/assets/306834/a5e39d81-2017-46c9-b9b9-1bf3920f1c85)
![Screenshot from 2024-04-25 07-28-31](https://github.com/equinor/ert/assets/306834/f7367761-d21a-4aff-8ef2-5495907904e2)

**Environment**
 - OS: [e.g. RHEL7]
 - ERT/Komodo release: [e.g. 2.18 or 2020.12.00]
 - Python version
 - Remote/HPC execution involved: [yes|no]

**Additional context**
https://engage.cloud.microsoft/main/org/statoil.com/threads/eyJfdHlwZSI6IlRocmVhZCIsImlkIjoiMjc2Nzg3NTUyNTA4MzEzNiJ9

PR #7764, opened 2024-04-25
Avoid possibility for negative duration in GUI

berland:
There is an underlying problem with the snapshot model not being properly reset when realizations are resubmitted upon failure. This commit makes sure that when invalid end times are picked up from previous runs, they are ignored.

**Issue**
Resolves partly #7763 

**Approach**
Hide the problem under the the carpet 🧹 

## When applicable


pinkwah (approved):

berland:
When a forward model is rerun, this change will avoid negative duration popping up, but it will also show the duration to be zero for the entire course of the forward model step rerun, since we never get to compare with datetime.datetime.now() (because end_time is not None, because it is from previous rerun).

jonathan-eq:
I think the carpet, under which you had swept the problems, has been moved 🧹

PR #7768, opened 2024-04-25
Detect if Date should be used for plotting on x-axis

andreas-el:
**Issue**
Resolves #my_issue

**Approach**

## When applicable


src/ert/gui/tools/plot/plot_window.py
```diff
@@ -226,19 +213,27 @@ def currentPlotChanged(self):
 
             plot_context.log_scale = key_def.log_scale
 
-            plot_widget.updatePlot(plot_context, ensemble_to_data_map, observations)
+            if self._preferred_ensemble_x_axis_format == PlotContext.UNKNOWN_AXIS:
+                for _, data in ensemble_to_data_map.items():
+                    data = data.T
```
yngve-sk, line 218: Why is this transpose needed? For the check below? Not seeing it used elsewhere

yngve-sk (approved):
LGTM w/ one minor comment that should be checked (and resolved if it is something to resolve) before merging.

Also, you can "flip" the x axis by swapping the min and the max, though this seems like a harmless extra functionality and may even be useful for some

src/ert/gui/tools/plot/plot_window.py
andreas-el: This was needed for the check below yes.
I've slightly refactored this to be somewhat more optimised, since we will encounter date-time objects for the most part.

PR #7822, opened 2024-05-02
Simplify how we call and parse bjobs

JHolba:
Use the -o option to specify which columns we are interested in. Add the -noheader option to remove the header from output.

**Approach**
Specifying output when calling bjobs to make the lsf driver logic a bit simpler.

## When applicable

### DIFF
--- a/src/ert/scheduler/lsf_driver.py
+++ b/src/ert/scheduler/lsf_driver.py
@@ -95,17 +95,16 @@ class JobData(BaseModel):
 def parse_bjobs(bjobs_output: str) -> Dict[str, Dict[str, Dict[str, str]]]:
     data: Dict[str, Dict[str, str]] = {}
     for line in bjobs_output.splitlines():
-        if not line or not line[0].isdigit():
-            continue
-        tokens = line.split(maxsplit=3)
-        if len(tokens) >= 3 and tokens[0] and tokens[2]:
-            if tokens[2] not in get_args(JobState):
+        tokens = line.split(sep="^")
+        if len(tokens) == 2:
+            job_id, job_state = tokens
+            if job_state not in get_args(JobState):
                 logger.error(
-                    f"Unknown state {tokens[2]} obtained from "
-                    f"LSF for jobid {tokens[0]}, ignored."
+                    f"Unknown state {job_state} obtained from "
+                    f"LSF for jobid {job_id}, ignored."
                 )
                 continue
-            data[tokens[0]] = {"job_state": tokens[2]}
+            data[job_id] = {"job_state": job_state}
     return {"jobs": data}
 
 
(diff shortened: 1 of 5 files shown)


src/ert/scheduler/lsf_driver.py
```diff
@@ -330,7 +329,9 @@ async def poll(self) -> None:
                 continue
             current_jobids = list(self._jobs.keys())
             process = await asyncio.create_subprocess_exec(
-                self._bjobs_cmd,
+                str(self._bjobs_cmd),
+                "-noheader",
+                "-o \"jobid stat delimiter='^'\"",
```
jonathan-eq, line 334: What is the advantage of using `^` instead of space?

JHolba: it doesn't really help with the current chosen columns, but it doesn't hurt either.
in general I find it nicer to use a delimiter that is unlikely to be used for anything in the text

src/ert/scheduler/lsf_driver.py
```diff
@@ -330,7 +329,9 @@ async def poll(self) -> None:
                 continue
             current_jobids = list(self._jobs.keys())
             process = await asyncio.create_subprocess_exec(
-                self._bjobs_cmd,
+                str(self._bjobs_cmd),
```
jonathan-eq, line 332: Do we have to make this a string?

src/ert/scheduler/lsf_driver.py
jonathan-eq: Would it be better to use triple quotes to avoid having to escape the inner quotes?

tests/integration_tests/job_queue/test_lsf_driver.py
```diff
@@ -176,11 +176,7 @@ def copy_lsf_poly_case(copy_poly_case, tmp_path):
     "mock_start_server",
 )
 @pytest.mark.integration_test
-def test_run_mocked_lsf_queue(request, using_scheduler):
-    if "fail" in request.node.name and using_scheduler:
-        pytest.skip(
-            "Python LSF driver does not support general resubmission on bsub errors"
-        )
```
jonathan-eq, line 183: Does this work now?

src/ert/scheduler/lsf_driver.py
JHolba: it looked a bit out of place when I tried it earlier, but I can change it if you have strong feelings about it :)
both ways were accepted by ruff format

tests/integration_tests/job_queue/test_lsf_driver.py
JHolba: depends on what you mean by "work". I turned the test of for scheduler :sweat_smile:.

src/ert/scheduler/lsf_driver.py
jonathan-eq: Okay, lets leave it for now

tests/integration_tests/job_queue/test_lsf_driver.py
jonathan-eq: Ahhhh, so that's what it did.
Should we not run those tests with scheduler?

src/ert/scheduler/lsf_driver.py
jonathan-eq: We might actually have to.

tests/integration_tests/job_queue/test_lsf_driver.py
JHolba: we have other tests covering the same code

jonathan-eq (approved):
LGTM!

Issue #7759, opened 2024-04-24
Improve ambiguous logging when max_submit is 1


xjules:
Currently even-though `MAX_SUBMIT` is set 1 we log failure with `failed after reaching max submit`. We should provide more detailed explanation in `job.handle_failure`.
Also `job._callback_status_msg` might be empty, which produces empty output:
```sh
Realization: 29 failed after reaching max submit (1):
	
Realization: 44 failed after reaching max submit (1):
	
Realization: 30 failed after reaching max submit (1):
```

Definition of done:
In case that job fails, we should provide all the relevant information coming from the driver and the queue stdout and sterr files.

xjules:
Suggestion what to log:

- exit code of the last successful / failed job
- the last know state

xjules:
Apparently, when mimicking NFS syncing issues, we will not get any logs too: Copy from @eivindjahren message:

---

If `_ert_forward_model_runner crashes` (for instance due to missing jobs.json because of NFS sync issues) then you get no indication of what happened. Just the empty failure message:
```
Realization: 44 failed after reaching max submit (1):
```
You can reproduce it with fault injecting not writing the `jobs.json` file:
```sh
--- a/src/ert/enkf_main.py
+++ b/src/ert/enkf_main.py
@@ -231,7 +231,7 @@ def create_run_path(
                     run_context.iteration,
                 )
 
-                json.dump(forward_model_output, fptr)
+                # json.dump(forward_model_output, fptr)
 
     run_context.runpaths.write_runpath_list(
         [run_context.iteration], run_context.active_realizations
 class LegacyEnsemble(Ensemble):
@@ -226,7 +227,7 @@ async def _evaluate_inner(  # pylint: disable=too-many-branches
                 self.min_required_realizations if self.stop_long_running else 0
             )
 
-            queue.add_dispatch_information_to_jobs_file()
+            # queue.add_dispatch_information_to_jobs_file()
             result = await queue.execute(min_required_realizations)
 
         except Exception:
```

xjules:
The logging might be already fixed by  50a4421. Need to just test.

jonathan-eq:
> The logging might be already fixed by [50a4421](https://github.com/equinor/ert/commit/50a44210a148234c08baa85bcb1ca34f0581d5e0). Need to just test.

It did not fix it

xjules:
What we should do is to "find out" that the job does not run and get the lsf stdout into the logs.

PR #7836, opened 2024-05-05
Add driver related failure info when job fails

xjules:
**Issue**
Resolves https://github.com/equinor/ert/issues/7759

**Approach**
This adds two information explicitly to the logs:
 1) when driver.submit fails (ie. non-zero exit code) the error message will be kept and logged later on in handle_failure in job.py
 2) when job_runner fails, the stdout and stderr files (if provided by the queue), will be read and logged on handle_failure in job.py.

## When applicable

### DIFF
--- a/src/ert/scheduler/driver.py
+++ b/src/ert/scheduler/driver.py
@@ -18,6 +18,7 @@ class Driver(ABC):
 
     def __init__(self, **kwargs: Dict[str, str]) -> None:
         self._event_queue: Optional[asyncio.Queue[Event]] = None
+        self._job_error_message_by_iens: Dict[int, str] = {}
 
     @property
     def event_queue(self) -> asyncio.Queue[Event]:
@@ -60,6 +61,12 @@ class Driver(ABC):
     @abstractmethod
     async def finish(self) -> None:
         """make sure that all the jobs / realizations are complete."""
+
+    def read_stdout_and_stderr_files(
+        self, runpath: str, job_name: str, num_characters_to_read_from_end: int = 300
+    ) -> str:
+        """Each driver should provide some output in case of failure."""
+        return ""
 
     @staticmethod
     async def _execute_with_retry(
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -187,6 +187,13 @@ class Job:
             f"\n\t{self._callback_status_msg}"
         )
 
+        if msg := self.driver._job_error_message_by_iens.get(self.iens, ""):
+            error_msg += f"\nDriver reported: {msg}"
+
+        error_msg += self.driver.read_stdout_and_stderr_files(
+            self.real.run_arg.runpath, self.real.run_arg.job_name
+        )
+
         self.real.run_arg.ensemble_storage.set_failure(
             self.real.run_arg.iens, RealizationStorageState.LOAD_FAILURE, error_msg
         )
--- a/src/ert/scheduler/lsf_driver.py
+++ b/src/ert/scheduler/lsf_driver.py
@@ -267,6 +267,7 @@ class LsfDriver(Driver):
             retry_interval=self._sleep_time_between_cmd_retries,
         )
         if not process_success:
+            self._job_error_message_by_iens[iens] = process_message
             raise RuntimeError(process_message)
 
         match = re.search("Job <([0-9]+)> is submitted to .*queue", process_message)
@@ -300,7 +301,7 @@ class LsfDriver(Driver):
             job_id,
         ]
 
-        process_success, process_message = await self._execute_with_retry(
+        _, process_message = await self._execute_with_retry(
             bkill_with_args,
             retry_codes=(FLAKY_SSH_RETURNCODE,),
             retries=3,
@@ -524,3 +525,26 @@ class LsfDriver(Driver):
 
     async def finish(self) -> None:
         pass
+
+    def read_stdout_and_stderr_files(
+        self, runpath: str, job_name: str, num_characters_to_read_from_end: int = 300
+    ) -> str:
+        error_msg = ""
+        stderr_file = Path(runpath) / (job_name + ".LSF-stderr")
+        if msg := tail_textfile(stderr_file, num_characters_to_read_from_end):
+            error_msg += f"\n    LSF-stderr:\n{msg}"
+        stdout_file = Path(runpath) / (job_name + ".LSF-stdout")
+        if msg := tail_textfile(stdout_file, num_characters_to_read_from_end):
+            error_msg += f"\n    LSF-stdout:\n{msg}"
+        return error_msg
+
+
+def tail_textfile(file_path: Path, num_chars: int) -> str:
+    if not file_path.exists():
+        return f"No output file {file_path}"
+    with open(file_path, encoding="utf-8") as file:
+        file.seek(0, 2)
+        file_end_position = file.tell()
+        seek_position = max(0, file_end_position - num_chars)
+        file.seek(seek_position)
+        return file.read()[-num_chars:]
--- a/tests/integration_tests/scheduler/test_lsf_driver.py
+++ b/tests/integration_tests/scheduler/test_lsf_driver.py
@@ -1,7 +1,9 @@
 import json
 import logging
 import os
+import random
 import stat
+import string
 from pathlib import Path
 
 import pytest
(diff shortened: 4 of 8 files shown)


src/ert/scheduler/lsf_driver.py
```diff
@@ -266,6 +266,7 @@ async def submit(
             retry_interval=self._sleep_time_between_cmd_retries,
         )
         if not process_success:
+            self._job_error_message[iens] = process_message
             raise RuntimeError(process_message)
```
jonathan-eq, line 270: I thought this would be logged regardless. We run this coroutine in an event loop from async_utils.py, so it will log any raised errors once it is done.
But maybe it's good to log it once more just to be sure.

src/ert/scheduler/lsf_driver.py
```diff
@@ -266,6 +266,7 @@ async def submit(
             retry_interval=self._sleep_time_between_cmd_retries,
         )
         if not process_success:
+            self._job_error_message[iens] = process_message
```
jonathan-eq, line 269: Can a job have multiple errors? 
Maybe we should do `self._job_error_message[iens] += process_message` instead?

xjules: Maybe, can include this one.

src/ert/scheduler/lsf_driver.py
xjules: For some reason the `RuntimeError` can't be found in logs.

jonathan-eq: Not under jobqueue logs either?

src/ert/scheduler/lsf_driver.py
xjules: Actually, changed my mind - we do care only about the last error msg, I think.

src/ert/scheduler/driver.py
```diff
@@ -61,6 +63,12 @@ async def poll(self) -> None:
     async def finish(self) -> None:
         """make sure that all the jobs / realizations are complete."""
 
+    def read_output_files(
+        self, runpath: str, job_name: str, num_tail_characters_to_read: int = 300
```
jonathan-eq, line 67: Should we change `num_tail_characters_to_read` to `num_characters_to_read_from_end`?

src/ert/scheduler/job.py
```diff
@@ -187,6 +187,18 @@ async def _handle_failure(self) -> None:
             f"\n\t{self._callback_status_msg}"
         )
 
+        if (
+            self.iens in self.driver._job_error_message
+            and self.driver._job_error_message[self.iens]
+        ):
```
jonathan-eq, line 193: Is it not enough to only check if error message for specific iens exists?
```suggestion
        if (iens_error_msg := self.driver._job_error_message.get(self.iens)):
            error_msg += f"\n\tDriver reported: {iens_error_msg}"

```

src/ert/scheduler/job.py
```diff
@@ -187,6 +187,18 @@ async def _handle_failure(self) -> None:
...
+            error_msg += (
+                f"\n\tDriver reported: {self.driver._job_error_message[self.iens]}"
+            )
```
jonathan-eq, line 196: Is it not enough to only check if error message for specific iens exists?
```suggestion
        if (iens_error_msg := self.driver._job_error_message.get(self.iens)):
            error_msg += f"\n\tDriver reported: {iens_error_msg}"

```

src/ert/scheduler/lsf_driver.py
```diff
@@ -508,3 +509,26 @@ def _build_resource_requirement_arg(self) -> List[str]:
...
+        if msg := get_info_from_text_outfile(stdout_file, num_tail_characters_to_read):
+            error_msg += f"\n\t LSF-out: {msg}"
+        return error_msg
+
+
+def get_info_from_text_outfile(file_path: Path, num_chars: int) -> str:
+    if not file_path.exists():
+        return f"No output files for {file_path}"
+    with open(file_path, encoding="utf-8") as file:
+        file.seek(0, 2)
```
jonathan-eq, line 530: Can we change this to `file.seek(num_chars, 2)` instead? Would seek default to position 0 if trying to go before the start of the file?

tests/unit_tests/scheduler/test_job.py
```diff
@@ -109,6 +109,12 @@ async def test_job_run_sends_expected_events(
         lambda _: LoadResult(forward_model_ok_result, ""),
     )
     scheduler = create_scheduler()
+    monkeypatch.setattr(
+        scheduler.driver,
+        "read_output_files",
+        lambda x, y: "",
```
jonathan-eq, line 115:
```suggestion
        lambda *args: "",
```

tests/unit_tests/scheduler/test_lsf_driver.py
```diff
@@ -263,6 +263,27 @@ async def test_faulty_bsub(monkeypatch, tmp_path, bsub_script, expectation):
...
 
+async def test_faulty_bsub_produces_error_log(monkeypatch, tmp_path):
+    monkeypatch.chdir(tmp_path)
+    bin_path = Path("bin")
+    bin_path.mkdir()
+    _out = "THIS_IS_OUTPUT"
+    _err = "THIS_IS_ERROR"
+    bsub_script = f"echo {_out} && echo {_err} >&2; exit 1"
+
+    monkeypatch.setenv("PATH", f"{bin_path}:{os.environ['PATH']}")
```
jonathan-eq, line 274: Let's move this line to right after `bin_path.mkdir()`

src/ert/scheduler/driver.py
```diff
@@ -18,6 +18,8 @@ class Driver(ABC):
 
     def __init__(self, **kwargs: Dict[str, str]) -> None:
         self._event_queue: Optional[asyncio.Queue[Event]] = None
+        # we will keep the error messages coming from the driver
+        self._job_error_message: Dict[int, str] = {}
```
jonathan-eq, line 22: Maybe the instance variable name can be more descriptive?
```suggestion
        self._job_error_message_by_iens: Dict[int, str] = {}
```

src/ert/scheduler/lsf_driver.py
```diff
@@ -508,3 +509,26 @@ def _build_resource_requirement_arg(self) -> List[str]:
...
+        file_end_position = file.tell()
+        seek_position = max(0, file_end_position - num_chars)
+        file.seek(seek_position)
+        return file.read()[-num_chars:]
```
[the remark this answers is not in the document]
xjules: Yes, the output file can be large and thus it makes sense just seek the end and return `X` characters from the end.

src/ert/scheduler/lsf_driver.py
xjules: When opening in text mode this will not work as `file.seek(x, 2)` requires a byte offset and not in terms of number of characters.

jonathan-eq:
:+1:

berland:
If this PR solves the referred issue, the issue should be rephrased and leftover things to do should be in a new issue?

src/ert/scheduler/job.py
```diff
@@ -187,6 +187,13 @@ async def _handle_failure(self) -> None:
             f"\n\t{self._callback_status_msg}"
         )
 
+        if msg := self.driver._job_error_message_by_iens.get(self.iens, ""):
+            error_msg += f"\n\tDriver reported: {msg}"
```
berland, line 191: I am not sure I like the tab-character here..

tests/integration_tests/scheduler/test_lsf_driver.py
```diff
@@ -68,6 +70,30 @@ async def test_lsf_dumps_stderr_to_file(tmp_path, job_name):
     )
 
 
+def generate_random_text(size):
+    letters = string.ascii_letters
+    return "".join(random.choice(letters) for i in range(size))
+
+
+@pytest.mark.parametrize("tail_chars_to_read", [(5), (50), (500)])
+async def test_lsf_read_output_files(tmp_path, job_name, tail_chars_to_read):
```
berland, line 79: `test_lsf_can_retrieve_stdout_and_stderr`, is that more precise?

src/ert/scheduler/lsf_driver.py
```diff
@@ -508,3 +509,30 @@ def _build_resource_requirement_arg(self) -> List[str]:
 
     async def finish(self) -> None:
         pass
+
+    def read_output_files(
```
berland, line 513: I think "output" is too ambigous here. Include stdout and stderr in function name.

src/ert/scheduler/driver.py
```diff
@@ -61,6 +63,12 @@ async def poll(self) -> None:
     async def finish(self) -> None:
         """make sure that all the jobs / realizations are complete."""
 
+    def read_output_files(
```
berland, line 66: If this is to be used in "case of failure", it should be reflected in the function name. 

The driver could have a specific API for stdout and stderr, perhaps a failure message code can be made driver independent based on stdout/stderr api?

src/ert/scheduler/job.py
jonathan-eq: I don't think we need the empty string in `...get(self.iens, "")` either.

xjules: `if empty string` yields False

xjules: I can remove the tab though :+1:

src/ert/scheduler/lsf_driver.py
xjules: `read_stdout_and_stderr_files`

xjules:
> If this PR solves the referred issue, the issue should be rephrased and leftover things to do should be in a new issue?

I added `definition of done` section to do issue.

tests/integration_tests/scheduler/test_lsf_driver.py
```diff
@@ -68,6 +70,32 @@ async def test_lsf_dumps_stderr_to_file(tmp_path, job_name):
     )
 
 
+def generate_random_text(size):
+    letters = string.ascii_letters
+    return "".join(random.choice(letters) for i in range(size))
+
+
+@pytest.mark.parametrize("tail_chars_to_read", [(5), (50), (500)])
```
berland, line 78: This should be tested for a number higher than 600 also.

src/ert/scheduler/lsf_driver.py
```diff
@@ -524,3 +525,30 @@ def _build_resource_requirement_arg(self) -> List[str]:
...
+
+    def read_stdout_and_stderr_files(
+        self, runpath: str, job_name: str, num_characters_to_read_from_end: int = 300
+    ) -> str:
+        error_msg = ""
+        stderr_file = Path(runpath) / (job_name + ".LSF-stderr")
+        if msg := get_info_from_text_outfile(
+            stderr_file, num_characters_to_read_from_end
+        ):
+            error_msg += f"\n\t LSF-err: {msg}"
```
berland, line 537: tab characters...

xjules: These I kept on purpose but can remove them :+1:

berland: Use 4 or 8 explicit spaces if you want to do formatting. Will lineshift and indents always look good? It might make it more difficult to parse the logs, but it will look good to humans.

src/ert/scheduler/driver.py
```diff
@@ -18,6 +18,8 @@ class Driver(ABC):
 
     def __init__(self, **kwargs: Dict[str, str]) -> None:
         self._event_queue: Optional[asyncio.Queue[Event]] = None
+        # we will keep the error messages coming from the driver
```
berland, line 21: (should not need this comment)

berland (approved):
Maybe fine to merge and then iterate when we have done more testing of error scenarios. Good Job!

xjules:
This is the current log output:
image

Issue #3502, opened 2022-06-10
stderr from job_dispatch is lost when runpath is missing


berland:
**Describe the bug**

If `job_dispatch.py` fails on a cluster node due to runpath not existing, it will fail with `sys.exit` and a "No such directory" message, but this error is not conveyed to the ERT user on the master node. This error goes to stderr in the LSF system, which normally is dumped to the runpath, but that cannot happen as it does not exist.

This situation can arise either due to user errors, or if the filesystem on the cluster is misconfigured or flaky.

**To Reproduce**
1. Set queue driver to `LSF` in `snake_oil.ert`.
2. `mkdir /tmp/my-runpath` on your master node (this constitutes a "user error")
3. Set `RUNPATH /tmp/my-runpath/realization-%d` in `snake_oil.ert`
4. `ert gui snake_oil.ert`
5. Run ensemble experiment, one realization is enough (but not test run, as we need the cluster)
6. Observe all realizations fail, but you will find no hint of what is actually the problem.

**Expected behavior**
stdout/stderr from LSF conveyed to the user in some way.

**Environment**
 - OS: RHEHL7
 - ERT/Komodo Release: 2022.05
 - Python version: 3.8
 - Remote/HPC execution involved: yes

**Additional context**
The cluster job probably gets to this line:

https://github.com/equinor/ert/blob/237024c221a415a9300b658320b68d9fcddc12db/job_runner/cli.py#L57

and then it exits. It would be natural to think that `job_dispatch` should call home in this situation with the message, but that is impossible as it uses information in the runpath to do so, loaded in the following lines.

jondequinor:
But we write a lot of files in the run_path prior to running `job_dispatch.py`. So why isn't this caught earlier?

berland:
> But we write a lot of files in the run_path prior to running `job_dispatch.py`. So why isn't this caught earlier?

These files are written by the master node, which in this scenario has access to the runpath.

jonathan-eq:
Some **potential** solutions:
1. Have LSF return the stderr to the client. Check in the LSF documentation if this is possible at all. This would be the preferred way, as we would able to catch more errors down the line.
2. Verify that the runpath is not a shared disk when running on non local queue, and output warning if it is.

jonathan-eq:
Putting this back to Todo as it is an improvement for mostly us developers, and we have not found a good way of implementing it. We could hard code it to check if runpath starts with /tmp, but that is a bad practice..

berland:
[...]

PR #7865, opened 2024-05-10
Instruct LSF to dump stderr of submitted job to runpath

berland:
This information should not be lost, can be crucial for debugging

**Issue**
Related to #3502 

**Approach**
Use `bsub`s `-e` option.

## When applicable

### DIFF
--- a/tests/integration_tests/scheduler/test_lsf_driver.py
+++ b/tests/integration_tests/scheduler/test_lsf_driver.py
@@ -52,6 +52,22 @@ async def test_lsf_stdout_file(tmp_path, job_name):
     assert "yay" in lsf_stdout
 
 
+async def test_lsf_dumps_stderr_to_file(tmp_path, job_name):
+    os.chdir(tmp_path)
+    driver = LsfDriver()
+    failure_message = "failURE"
+    await driver.submit(0, "sh", "-c", f"echo {failure_message} >&2", name=job_name)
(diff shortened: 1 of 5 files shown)


berland:
This incurs a change in the runpaths file system, and will becomes an API we cannot break. Thus, this should also be approved by @sondreso

tests/integration_tests/scheduler/test_lsf_driver.py
```diff
@@ -52,6 +52,22 @@ async def test_lsf_stdout_file(tmp_path, job_name):
     assert "yay" in lsf_stdout
 
 
+async def test_lsf_dumps_stderr_to_file(tmp_path, job_name):
+    os.chdir(tmp_path)
+    driver = LsfDriver()
+    failure_message = "failURE"
```
xjules, line 58: Is it on purpose or just typo: "failURE" ?

berland: On purpose. It crossed my mind that there is a above-zero probability that something else could write the string "failure" to the file, but exactly zero probability that something else would write "failURE"

sondreso (approved):
I think adding this file to the runpath is necessary in order to provide sufficient debugging information to both users and developers, so I'm fine with this. We should strive to get this information displayed to the user in the GUI as well!

PR #7890, opened 2024-05-14
Refactor Snapshot Metadata to be dataclass instead of dict

### LINKED ISSUE #7883: Use typeddict in `ert.gui.model.snapshot`
The SnapshotModel class has a dict called metadata, which should be a dataclass instead to improve readability.

jonathan-eq:
**Issue**
Resolves #7883 

**Approach**

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/snapshot.py
+++ b/src/ert/ensemble_evaluator/snapshot.py
@@ -1,12 +1,16 @@
+from __future__ import annotations
+
 import datetime
 import re
 import typing
 from collections import defaultdict
+from dataclasses import dataclass, field
 from typing import Any, Dict, Mapping, Optional, Sequence, Tuple, Union
 
 from cloudevents.http import CloudEvent
 from dateutil.parser import parse
 from pydantic import BaseModel, ConfigDict
+from qtpy.QtGui import QColor
 
 from ert.ensemble_evaluator import identifiers as ids
 from ert.ensemble_evaluator import state
@@ -71,6 +75,50 @@ def _filter_nones(some_dict: Mapping[str, Any]) -> Dict[str, Any]:
     return {key: value for key, value in some_dict.items() if value is not None}
 
 
+@dataclass
+class SnapshotMetadata:
+    aggr_job_status_colors: defaultdict[str, dict[str, QColor]] = field(
+        default_factory=lambda: defaultdict(dict)
+    )
+    real_status_colors: dict[str, QColor] = field(default_factory=dict)
+    sorted_real_ids: list[str] = field(default_factory=list)
+    sorted_forward_model_ids: defaultdict[str, list[str]] = field(
+        default_factory=lambda: defaultdict(list)
+    )
+
+    def merge(self, other_metadata: SnapshotMetadata) -> None:
+        if other_metadata.aggr_job_status_colors:
+            self.aggr_job_status_colors.update(
+                _filter_nones(other_metadata.aggr_job_status_colors)
+            )
+        if other_metadata.real_status_colors:
+            self.real_status_colors.update(
+                _filter_nones(other_metadata.real_status_colors)
+            )
+        if other_metadata.sorted_real_ids:
+            self.sorted_real_ids = other_metadata.sorted_real_ids
+        if other_metadata.sorted_forward_model_ids:
+            self.sorted_forward_model_ids.update(
+                _filter_nones(other_metadata.sorted_forward_model_ids)
+            )
+
+    def to_dict(
+        self,
+    ) -> dict[
+        str,
+        list[str]
+        | dict[str, QColor]
+        | defaultdict[str, dict[str, QColor]]
+        | defaultdict[str, list[str]],
+    ]:
+        return {
+            "aggr_job_status_colors": self.aggr_job_status_colors,
+            "real_status_colors": self.real_status_colors,
+            "sorted_real_ids": self.sorted_real_ids,
+            "sorted_forward_model_ids": self.sorted_forward_model_ids,
+        }
+
+
 class PartialSnapshot:
     def __init__(self, snapshot: Optional["Snapshot"] = None) -> None:
         self._realization_states: Dict[
@@ -90,7 +138,7 @@ class PartialSnapshot:
         self._ensemble_state: Optional[str] = None
         # TODO not sure about possible values at this point, as GUI hijacks this one as
         # well
-        self._metadata: Dict[str, Any] = defaultdict(dict)
+        self._metadata = SnapshotMetadata()
 
         self._snapshot = snapshot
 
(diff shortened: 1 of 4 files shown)


berland:
Hmm, why do we have to add 60 more lines than before from changing to dataclass?

jonathan-eq:
> Hmm, why do we have to add 60 more lines than before from changing to dataclass?

Before, we used a plain defaultdict, so the dataclass implementation will take up some lines. 
I also had to add a __post_init__ method to make sure the dicts were default dicts. This was sometimes not the case, and would end up as KeyErrors down the line. I will check if I can put this logic somewhere else, but it might be better to just keep it in the dataclass itself.

berland:
> Before, we used a plain defaultdict, so the dataclass implementation will take up some lines. I also had to add a **post_init** method to make sure the dicts were default dicts. This was sometimes not the case, and would end up as KeyErrors down the line. I will check if I can put this logic somewhere else, but it might be better to just keep it in the dataclass itself.

Could that indicate that the default-dict is a better implementation? I see that default-dicts can be typed, `from typing import DefaultDict`.

jonathan-eq:
> > Before, we used a plain defaultdict, so the dataclass implementation will take up some lines. I also had to add a **post_init** method to make sure the dicts were default dicts. This was sometimes not the case, and would end up as KeyErrors down the line. I will check if I can put this logic somewhere else, but it might be better to just keep it in the dataclass itself.
> 
> Could that indicate that the default-dict is a better implementation? I see that default-dicts can be typed, `from typing import DefaultDict`.

The culprit was `SnapshotMetadata(**data['metadata'])`. I did not check if the value was empty dict, so the defaultdict (default) was never initialized.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -71,6 +75,50 @@ def _filter_nones(some_dict: Mapping[str, Any]) -> Dict[str, Any]:
     return {key: value for key, value in some_dict.items() if value is not None}
 
 
+@dataclass
+class SnapshotMetadata:
+    _aggr_job_status_colors: defaultdict[str, dict[str, QColor]] = field(
+        default_factory=lambda: defaultdict(dict)
+    )
+    _real_status_colors: dict[str, QColor] = field(default_factory=dict)
```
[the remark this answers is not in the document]
jonathan-eq: Agreed, I am unsure what the reason for this was in the old dict key.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -71,6 +75,50 @@ def _filter_nones(some_dict: Mapping[str, Any]) -> Dict[str, Any]:
...
+        default_factory=lambda: defaultdict(dict)
+    )
+    real_status_colors: dict[str, QColor] = field(default_factory=dict)
+    sorted_real_ids: list[str] = field(default_factory=list)
+    sorted_forward_model_ids: defaultdict[str, list[str]] = field(
+        default_factory=lambda: defaultdict(list)
+    )
+
+    def merge(self, other_metadata: SnapshotMetadata) -> None:
+        if other_metadata.aggr_job_status_colors:
```
berland, line 90: Is this `if` necessary if `other_metadata` is also a dataclass where `aggr_job_status_colours` always has a default value?

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -468,7 +516,9 @@ def add_forward_model(
 def _from_nested_dict(data: Mapping[str, Any]) -> PartialSnapshot:
     partial = PartialSnapshot()
     if "metadata" in data:
-        partial._metadata = data["metadata"]
+        partial._metadata = SnapshotMetadata(
+            **{k: v for k, v in data["metadata"].items() if v}
```
berland, line 520: I like to avoid using single letter objects. This reads better if k is replaced with key, and v with value.

src/ert/ensemble_evaluator/snapshot.py
jonathan-eq: I guess for some of those fields, the explicit check was to not update the field if the new provided field was the default (so not set). I will try removing the check and see if anything blows up.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -71,6 +75,50 @@ def _filter_nones(some_dict: Mapping[str, Any]) -> Dict[str, Any]:
...
+        if other_metadata.sorted_forward_model_ids:
+            self.sorted_forward_model_ids.update(
+                _filter_nones(other_metadata.sorted_forward_model_ids)
+            )
+
+    def to_dict(
+        self,
+    ) -> dict[
+        str,
+        list[str]
```
berland, line 103: This is a union of all the types the dataclass can contain. I am unsure of the value of typing here when we already have the types defined in the dataclass object. It might create less future maintenance if we compute the types directly here, there is something with `typing.get_type_hints(SnapshotMetadata)` according to a chatty speculative guy, it migth work.

jonathan-eq: Dynamic type hints proved difficult. It failed with `Call expression not allowed in type expression` when trying this:
```
def to_dict(
        self,
    ) -> dict[
        str,
        get_type_hints(SnapshotMetadata),
    ]:
```

jonathan-eq: I could however use the type hints to simplify the implementation, at the cost of readability.
How about this?
```
def to_dict(
        self,
    ) -> dict[
        str,
        list[str]
        | dict[str, QColor]
        | defaultdict[str, dict[str, QColor]]
        | defaultdict[str, list[str]],
    ]:
        hints = get_type_hints(SnapshotMetadata)
        return {field: getattr(self, field) for field in hints.keys()}
```

berland: nah, lets keep it as is then. It would be nicer maybe to be able to skip this to_dict and rather send the dataclass object as far as possible, but at the end it goes into Qt and it probably needs to be a pure dict.

jonathan-eq: Hopefully, we can skip converting it to a dict altogether in the end, but I think that would have to wait until later in the dataclass refactoring project :)

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -1,12 +1,16 @@
...
 import re
 import typing
 from collections import defaultdict
+from dataclasses import dataclass, field
 from typing import Any, Dict, Mapping, Optional, Sequence, Tuple, Union
 
 from cloudevents.http import CloudEvent
 from dateutil.parser import parse
 from pydantic import BaseModel, ConfigDict
+from qtpy.QtGui import QColor
```
jonathan-eq, line 13: @berland Any strong feelings about importing GUI classes outside of the GUI?
The alternative is changing the type to `Any`. I tried putting the class somewhere else, but that lead to circular import problems.
Maybe a better solution would be to move all the new snapshot/gui dataclasses to a `dataclasses.py` file?

berland: If it is a 3-tuple of `int`s, we can call it that. At least not use QColor if it is not such a type.

berland:
This snapshot class has been the subject of performance optimization earlier. There is a dedicated test code in `tests/performance_tests/test_snapshot.py` for measuring this. Can you do a before-and-after test of the effect of this PR?

jonathan-eq:
Before:
image

After:
image

jonathan-eq:
Superceeded by #8051

PR #7901, opened 2024-05-14
Make sure not to close the evaluator _ws_thread when done is not set yet

xjules:
**Issue**
(Maybe) Resolves https://github.com/equinor/ert/issues/7642

**Approach**
Make sure to wait until done is set.

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -91,6 +91,7 @@ class EnsembleEvaluator:
         self._ws_thread = ErtThread(
             name="ert_ee_run_server", target=self._run_server, args=(self._loop,)
         )
+        self._ws_stopped_event = threading.Event()
 
     @property
     def config(self) -> EvaluatorServerConfig:
@@ -144,6 +145,7 @@ class EnsembleEvaluator:
             )
             send_future.result()
             self._loop.call_soon_threadsafe(self._stop)
+            self._ws_stopped_event.wait()
 
     def _failed_handler(self, events: List[CloudEvent]) -> None:
         if self.ensemble.status not in (
@@ -343,6 +345,8 @@ class EnsembleEvaluator:
             close_timeout=60,
         ):
             await self._done
+            # this event signals done to the main thread
+            self._ws_stopped_event.set()
             if self._dispatchers_connected is not None:
                 logger.debug(
                     f"Got done signal. {self._dispatchers_connected.qsize()} "
@@ -401,6 +405,7 @@ class EnsembleEvaluator:
 
     def stop(self) -> None:
         self._loop.call_soon_threadsafe(self._stop)
+        self._ws_stopped_event.wait()
         self._ws_thread.join()
 
     def _signal_cancel(self) -> None:
@@ -418,6 +423,7 @@ class EnsembleEvaluator:
         else:
             logger.debug("Stopping current ensemble")
             self._loop.call_soon_threadsafe(self._stop)
+            self._ws_stopped_event.wait()
 
     def join(self) -> None:
         self._ws_thread.join()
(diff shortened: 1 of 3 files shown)


src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -144,6 +145,8 @@ def _cancelled_handler(self, events: List[CloudEvent]) -> None:
             )
             send_future.result()
             self._loop.call_soon_threadsafe(self._stop)
+            while not self._done.done():
+                time.sleep(0.1)
```
jonathan-eq, line 149: As this is repeated three times; should we put it in a separate method called `wait_until_done` or `sleep_until_done`?

xjules: Yes, we should.

berland:
With this PR, the test

`fix_loop$ pytest tests/unit_tests/ensemble_evaluator/test_ensemble_legacy.py -k run_and_cancel_legacy_ensemble[using_sch --count=10000 -n 48`

passed. Looks good! 👍🏻

src/ert/ensemble_evaluator/evaluator.py
xjules: I have replaced the `time.sleep` with `threading.Event()`

eivindjahren:
> With this PR, the test
> 
> `fix_loop$ pytest tests/unit_tests/ensemble_evaluator/test_ensemble_legacy.py -k run_and_cancel_legacy_ensemble[using_sch --count=10000 -n 48`
> 
> passed. Looks good! 👍🏻

I played a bit with the repeat numbers and the failing behavior becomes consistent at about ~20 repeats. Each time it takes about 4 seconds to run the test so I think it seems reasonable to add
```
@pytest.repeat(20)
```
to the test so that we get consistent test results rather than flaky behavior. We run with pytest-xdist anyways so this should take neglible additional time (at least in comparison to the `using_scheduler` fixture). This also means adding `pytest-repeat` to the dev dependencies in `pyproject.toml`.

eivindjahren:
We talked about the repeat idea, and came up with the idea of just parametrizing the test:
```python
@pytest.mark.timeout(60)
@pytest.mark.usefixtures("using_scheduler")
@pytest.mark.parametrize("num_reals, num_jobs", product(range(2, 6), range(2, 6)))
def test_run_and_cancel_legacy_ensemble(
    tmpdir, make_ensemble_builder, monkeypatch, num_reals, num_jobs
):
    print(num_reals, num_jobs)
    custom_port_range = range(1024, 65535)
    with tmpdir.as_cwd():
        ensemble = make_ensemble_builder(
            monkeypatch, tmpdir, num_reals, num_jobs, job_sleep=40, max_runtime=1
        ).build()
        config = EvaluatorServerConfig(
            custom_port_range=custom_port_range,
            custom_host="127.0.0.1",
            use_token=False,
            generate_cert=False,
        )

        evaluator = EnsembleEvaluator(ensemble, config, 0)

        evaluator.start_running()

        async def _run_monitor():
            async with Monitor(config) as mon:
                cancel = True
                with contextlib.suppress(
                    ConnectionClosed
                ):  # monitor throws some variant of CC if dispatcher dies
                    async for _ in mon.track():
                        # Cancel the ensemble upon the arrival of the first event
                        if cancel:
                            await mon.signal_cancel()
                            cancel = False

        run_monitor_in_loop(_run_monitor)
        assert evaluator._ensemble.status == state.ENSEMBLE_STATE_CANCELLED

        # realisations should not finish, thus not creating a status-file
        for i in range(num_reals):
            assert not os.path.isfile(f"real_{i}/status.txt")
```

xjules:
Closing this PR as this one is not required anymore it seems. Relates https://github.com/equinor/ert/pull/8000

PR #7915, opened 2024-05-16
Refactor Lsf driver to use dataclasses

### LINKED ISSUE #7879: Refactor object from Pydantic to dataclass
dataclass should be more lightweight, and the weight of Pydantic is probably not needed for this object:

https://github.com/equinor/ert/blob/bd52a3785cfbc8e677e6834b86975c2ab3d2786f/src/ert/scheduler/lsf_driver.py#L89

jonathan-eq:
**Issue**
Resolves #7879 

**Approach**
The commit in this PR  refactors lsf driver from using pydantic classes to dataclasses, as pydantic was overkill for our usage.

## When applicable

### DIFF
--- a/src/ert/scheduler/lsf_driver.py
+++ b/src/ert/scheduler/lsf_driver.py
@@ -11,6 +11,7 @@ import stat
 import subprocess
 import tempfile
 import time
+from dataclasses import dataclass
 from pathlib import Path
 from typing import (
     Dict,
@@ -21,12 +22,11 @@ from typing import (
     MutableMapping,
     Optional,
     Sequence,
+    Type,
     Union,
+    cast,
     get_args,
 )
-
-from pydantic import BaseModel, Field
-from typing_extensions import Annotated
 
 from ert.scheduler.driver import SIGNAL_OFFSET, Driver
 from ert.scheduler.event import Event, FinishedEvent, StartedEvent
@@ -42,34 +42,49 @@ JobState = Literal[
 ]
 
 
-class IgnoredJobstates(BaseModel):
+@dataclass(frozen=True)
+class IgnoredJobstates:
     job_state: Literal["UNKWN"]
 
 
-class FinishedJobSuccess(BaseModel):
+@dataclass(frozen=True)
+class FinishedJobSuccess:
     job_state: Literal["DONE", "PDONE"]
 
 
-class FinishedJobFailure(BaseModel):
+@dataclass(frozen=True)
+class FinishedJobFailure:
     job_state: Literal["EXIT", "ZOMBI"]
 
 
-class QueuedJob(BaseModel):
+@dataclass(frozen=True)
+class QueuedJob:
     job_state: Literal["PEND"]
 
 
-class RunningJob(BaseModel):
+@dataclass(frozen=True)
+class RunningJob:
     job_state: Literal["RUN", "SSUSP", "USUSP", "PSUSP"]
 
 
-AnyJob = Annotated[
-    Union[
-        FinishedJobSuccess, FinishedJobFailure, QueuedJob, RunningJob, IgnoredJobstates
-    ],
-    Field(discriminator="job_state"),
+_JOBSTATE_MAP = {
+    "EXIT": FinishedJobFailure,
+    "DONE": FinishedJobSuccess,
+    "PEND": QueuedJob,
+    "RUN": RunningJob,
+    "ZOMBI": FinishedJobFailure,
+    "PDONE": FinishedJobSuccess,
+    "SSUSP": RunningJob,
+    "USUSP": RunningJob,
+    "PSUSP": RunningJob,
+    "UNKWN": IgnoredJobstates,
+}
+
+AnyJob = Union[
+    FinishedJobSuccess, FinishedJobFailure, QueuedJob, RunningJob, IgnoredJobstates
 ]
 
-_STATE_ORDER: dict[type[BaseModel], int] = {
+_STATE_ORDER: dict[Type[AnyJob], int] = {
     IgnoredJobstates: -1,
     QueuedJob: 0,
     RunningJob: 1,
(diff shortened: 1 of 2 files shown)


berland:
This deviates from the way OpenPBS driver works. Should we keep them separate or also change OpenPBS to use dataclasses?

jonathan-eq:
> This deviates from the way OpenPBS driver works. Should we keep them separate or also change OpenPBS to use dataclasses?

I think it would be best to do them separately to keep the PRs smaller.
I have created a separate issue for the OpenPBS driver (https://github.com/equinor/ert/issues/7938)

berland:
> Yes, but is it not overkill to use Pydantic just to avoid implementing a `from_dict` class method in the OpenPBS driver? Would it be a bad idea to implement `from_dict` or `from_json`, and use the builtin json module instead?

It is a tradeoff we must make a decision on.

jonathan-eq:
> > Yes, but is it not overkill to use Pydantic just to avoid implementing a `from_dict` class method in the OpenPBS driver? Would it be a bad idea to implement `from_dict` or `from_json`, and use the builtin json module instead?
> 
> It is a tradeoff we must make a decision on.

I say we should merge this one, and I will get going on OpenPBS #7938 🚀

src/ert/scheduler/lsf_driver.py
```diff
@@ -105,7 +128,7 @@ def parse_bjobs(bjobs_output: str) -> Dict[str, Dict[str, Dict[str, str]]]:
                 )
                 continue
             data[job_id] = {"job_state": job_state}
```
berland, line 130: Would it simplify the code if we returned the correct Job classes here, and not the string `job_state`?

src/ert/scheduler/lsf_driver.py
```diff
@@ -82,17 +85,37 @@ class RunningJob(BaseModel):
 JOB_ALREADY_FINISHED_BKILL_MSG = "Job has already finished"
 
 
-class _Stat(BaseModel):
-    jobs: Mapping[str, AnyJob]
+def _create_job_class(job_dict: Mapping[str, str]) -> AnyJob:
+    job_state = job_dict["job_state"]
+    if job_state in get_type_hints(FinishedJobSuccess)["job_state"].__args__:
```
berland, line 90: This smells a little bit. Can we get away with just a dictionary from string job-state to the correct class?

jonathan-eq: Yes, it looks way better but Mypy does not like that whatsoever. 
```
src/ert/scheduler/lsf_driver.py:98: error: Argument 1 to "FinishedJobSuccess" has incompatible type "Literal['EXIT', 'DONE', 'PEND', 'RUN', 'ZOMBI', 'PDONE', 'SSUSP', 'USUSP', 'PSUSP', 'UNKWN']"; expected "Literal['DONE', 'PDONE']"  [arg-type]
src/ert/scheduler/lsf_driver.py:98: error: Argument 1 to "FinishedJobFailure" has incompatible type "Literal['EXIT', 'DONE', 'PEND', 'RUN', 'ZOMBI', 'PDONE', 'SSUSP', 'USUSP', 'PSUSP', 'UNKWN']"; expected "Literal['EXIT', 'ZOMBI']"  [arg-type]
src/ert/scheduler/lsf_driver.py:98: error: Argument 1 to "QueuedJob" has incompatible type "Literal['EXIT', 'DONE', 'PEND', 'RUN', 'ZOMBI', 'PDONE', 'SSUSP', 'USUSP', 'PSUSP', 'UNKWN']"; expected "Literal['PEND']"  [arg-type]
src/ert/scheduler/lsf_driver.py:98: error: Argument 1 to "RunningJob" has incompatible type "Literal['EXIT', 'DONE', 'PEND', 'RUN', 'ZOMBI', 'PDONE', 'SSUSP', 'USUSP', 'PSUSP', 'UNKWN']"; expected "Literal['RUN', 'SSUSP', 'USUSP', 'PSUSP']"  [arg-type]
src/ert/scheduler/lsf_driver.py:98: error: Argument 1 to "IgnoredJobstates" has incompatible type "Literal['EXIT', 'DONE', 'PEND', 'RUN', 'ZOMBI', 'PDONE', 'SSUSP', 'USUSP', 'PSUSP', 'UNKWN']"; expected "Literal['UNKWN']"  [arg-type]
```

berland:
Note to self: 

Before this commit, a line like  `foo = FinishedJobFailure("BAAAAH")` would raise a validation exception, after this commit it will not complain at all.

src/ert/scheduler/lsf_driver.py
```diff
@@ -82,18 +97,26 @@ class RunningJob(BaseModel):
 JOB_ALREADY_FINISHED_BKILL_MSG = "Job has already finished"
 
 
-class _Stat(BaseModel):
-    jobs: Mapping[str, AnyJob]
+def _create_job_class(job_state: JobState) -> AnyJob:
```
berland, line 100: Since this is only used inside _parse_job_dict, it can probably be inlined there?

src/ert/scheduler/lsf_driver.py
```diff
@@ -82,18 +97,26 @@ class RunningJob(BaseModel):
...
+@dataclass
+class JobData:
     iens: int
     job_state: AnyJob
     submitted_timestamp: float
 
 
-def parse_bjobs(bjobs_output: str) -> Dict[str, Dict[str, Dict[str, str]]]:
-    data: Dict[str, Dict[str, str]] = {}
+def parse_bjobs(bjobs_output: str) -> Dict[str, str]:
```
berland, line 118: Is it possible to tighten typing here from  `str` to something like `JobState`?

src/ert/scheduler/lsf_driver.py
```diff
@@ -463,9 +485,11 @@ async def _log_bhist_job_summary(self, job_id: str) -> None:
         )
         logger.info(f"Output from bhist -l: {process_message}")
 
-    async def _poll_once_by_bhist(self, missing_job_ids: Iterable[str]) -> _Stat:
+    async def _poll_once_by_bhist(
+        self, missing_job_ids: Iterable[str]
+    ) -> Dict[str, AnyJob]:
         if time.time() - self._bhist_cache_timestamp < self._bhist_required_cache_age:
-            return _Stat(**{"jobs": {}})
+            return _parse_jobs_dict({})
```
berland, line 492: Hmm, not sure why we did not do  `return {}` before to avoid the dummy function call?

jonathan-eq: Should we just return empty dict rather than the function call now?

src/ert/scheduler/lsf_driver.py
jonathan-eq: Yes, but I would need multiple casts in the other parts of the code.

jonathan-eq:
> Note to self:
> 
> Before this commit, a line like `foo = FinishedJobFailure("BAAAAH")` would raise a validation exception, after this commit it will not complain at all.

I think this is fine

src/ert/scheduler/lsf_driver.py
berland: It is arguably equivalent, so I think yes.

berland:
> > Note to self:
> > Before this commit, a line like `foo = FinishedJobFailure("BAAAAH")` would raise a validation exception, after this commit it will not complain at all.
> 
> I think this is fine

You can add it to the commit body as a comment when rebasing.

berland (approved):
Lets ship!

PR #7924, opened 2024-05-16
Let ERT be able to stop experiment when all realizations are pending

### LINKED ISSUE #7871: Terminate experiment does not work when jobs are pending
**Describe the bug**
Terminate experiment does not do anything when jobs are pending.

**To reproduce**
1. Login to an onprem or Azure node.
2. Run ERT in gui mode (I tried fmu-drogon/drogon_design.ert)
3. Maybe choose a queue that is very slow to respond, like `short`, or for LSF use an outrageous memory requirement like `QUEUE_OPTION LSF LSF_RESOURCE rusage[mem=200000]`
4. Run an ensemble experiment
5. While realizations are still in "Pending" state, click "Terminate experiment" and "Yes" in the dialoge box
6. Observe that nothing happens.
7. Observe that if the close-button in the upper right corner of the window is clicked, the same dialoge box appears, and if clicking yes, the run-dialogue window is closed. Realizations are still not killed, and the main Ert window has some errors: "Run Experiment" cannot be clicked, and the main window cannot be closed.

**Expected behaviour**
`qdel`/`bkill` commands should be initiated and Ert should tear down the ensemble experient

**Environment**
 - OS: RHEL7
 - ERT/Komodo release: bleeding as of 2024-05-13
 - Python version: 3.8

berland:
This allows the base_run_model to also being able to check regularly if there are other tasks at hand, like terminating the experiment.

**Issue**
Resolves #7871 

**Approach**
Add heartbeat to `monitor.track()`

Add a test to certify that `monitor.track()` is able to emit `None` events, aka "heartbeats". A better test would be to test that `BaseRunmodel.run_monitor()` would be able to exit at any time, but that looks like a big task. If `run_monitor()` is changed not to request hearbeats, the original bug would reappear but this test will not catch it.

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/monitor.py
+++ b/src/ert/ensemble_evaluator/monitor.py
@@ -101,24 +101,39 @@ class Monitor:
         )
         logger.debug(f"monitor-{self._id} informed server monitor is done")
 
-    async def track(self) -> AsyncGenerator[CloudEvent, None]:
-        timeout: Optional[float] = None
+    async def track(
+        self, heartbeat_interval: Optional[float] = None
+    ) -> AsyncGenerator[Optional[CloudEvent], None]:
+        """Yield events from the internal event queue with optional heartbeats.
+
+        Heartbeats are represented by None being yielded.
+
+        Heartbeats stops being emitted after a CloseTrackerEvent is found."""
+        _heartbeat_interval: Optional[float] = heartbeat_interval
+        closetracker_received: bool = False
         while True:
             try:
-                # We need a timeout in case TERMINATED event get not sent
-                # when signalling cancel or done to evaluator
-                event = await asyncio.wait_for(self._event_queue.get(), timeout)
+                event = await asyncio.wait_for(
+                    self._event_queue.get(), timeout=_heartbeat_interval
+                )
             except asyncio.TimeoutError:
-                logger.error("Evaluator did not send the TERMINATED event!")
-                break
+                if closetracker_received:
+                    logger.error("Evaluator did not send the TERMINATED event!")
+                    break
+                event = None
             if isinstance(event, CloseTrackerEvent):
-                timeout = self._receiver_timeout
+                closetracker_received = True
+                _heartbeat_interval = self._receiver_timeout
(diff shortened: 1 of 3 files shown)


berland:
todo: write test triggering the original bug.

src/ert/ensemble_evaluator/monitor.py
```diff
@@ -101,24 +101,39 @@ async def signal_done(self) -> None:
...
-                logger.error("Evaluator did not send the TERMINATED event!")
-                break
+                if closetracker_received:
+                    logger.error("Evaluator did not send the TERMINATED event!")
+                    break
+                event = None
             if isinstance(event, CloseTrackerEvent):
-                timeout = self._receiver_timeout
+                closetracker_received = True
+                _heartbeat_interval = self._receiver_timeout
```
xjules, line 126: Just wondering if we need the `receiver_timeout` at all when having the heartbeat in the first place.

berland: the heartbeat_interval is `None` by default, so there is not really a heartbeat in place unless you ask for it. 

The receiver timeout is 60 second, while a heartbeat is very short. I thought that this could risk logging the error "Evalutor did not send the TERMINATED event" too often and when it should not.

berland: There is only one consumer of the `track()` function except for tests, so the heartbeat could be there by default. It would require modifications in ca 15 tests though to ignore the None event, but that is doable.

xjules: I see. Then it makes sense to keep it - maybe. Nevertheless the `None` there is only for the sake of tests. It very much depends how trivial this modification of tests would be.

xjules: Let's keep it for now. I'll create an issue if it is doable to remove it.

xjules (approved):
Nice job @berland ! :rocket:

xjules:
Relates to https://github.com/equinor/ert/issues/7993

PR #7927, opened 2024-05-17
Wait for nsf file synchronization

### LINKED ISSUE #7788: Increase reliability over NFS by verifying responses
Currently, a number of errors can occur due to file synchronization issues over NFS:
```
done callback.*Could not find any unified summary file matching case
done callback.*Error reading.*Missing output file
done callback.*Error trying to read FIELD.*
done callback.*No such file or directory.*grdecl.*
Could not find any SUMMARY file.*
Could not find SUMMARY file.*
Error reading GEN_DATA.*
```

This could be caused by files written on the machine running `_ert_forward_model_runner` not being synchronized to the machine running `ok_callback`.

In order to mitigate this, `_ert_forward_model_runner` and `ok_callback` should send a manifest of files over ws.

The general idea is to write which files are to be read by `ok_callback` into jobs.json.
```
"files": {
  "SNAKE_OIL_PARAM": "parameters/snake_oil_parameters.txt",
  "SUMMARY_UNSMRY":  "SNAKE_OIL_FIELD.UNSMRY",
  "SUMMARY_SMSPEC": "SNAKE_OIL_FIELD.SMSPEC",
  "SNAKE_OIL_OPR_DIFF": "snake_oil_opr_diff_0.txt",
  "SNAKE_OIL_WPR_DIFF": "snake_oil_wpr_diff_0.txt",
  "SNAKE_OIL_GPR_DIFF": "snake_oil_gpr_diff_0.txt",
}
```
These correspond to parameters and responses in the config.

When `_ert_forward_model_runner` is done running all the forward model steps, it then sends how to fetch the data of all the files on its end:
```
"files": {
  "SUMMARY_UNSMRY": {
    "type": "file",
    "path": "/nfs/runpath/realization-0/iter-0/SNAKE_OIL_FIELD.UNMRY",
    "md5sum": "cfb779648201fdaef4ef31c6032a660f"
  }
}
```

The md5sum should be redundant as NFS has [close to open consistency](https://stackoverflow.com/questions/76091920/nfs-efs-close-to-open-consistency-rule-not-respected), but the file might take a while to be present.

DanSava:
**Issue**
Resolves #7788

## When applicable


src/_ert_forward_model_runner/reporting/message.py
```diff
@@ -119,3 +119,10 @@ class Exited(Message):
     def __init__(self, job, exit_code):
         super().__init__(job)
         self.exit_code = exit_code
+
+
+class Manifest(Message):
+    def __init__(self, data: Dict[str, Any], run_path: str):
```
eivindjahren, line 125: The word `data` says very little here. I would suggest creating a TypeDict for it in `src/_ert` and calling it `ManifestDict`:

```python

class ManifestDict(TypedDict):
    type: Literal["file"]
    path: str

class Manifest(Message):
    def __init__(self, manifest_dicts: Dict[str, ManifestDict], run_path: str):
        ...
```

src/_ert_forward_model_runner/runner.py
```diff
@@ -23,14 +28,42 @@ def __init__(self, jobs_data):
 
         self._set_environment()
 
+    def crate_manifest(self):
+        if not Path("manifest.json").exists():
+            return
+        with open("manifest.json", mode="r", encoding="utf-8") as f:
+            data = json.load(f)
```
eivindjahren, line 35: Its pretty weird that a function called `create_manifest` expects the file `manifest.json` to exist.

src/_ert_forward_model_runner/runner.py
```diff
@@ -23,14 +28,42 @@ def __init__(self, jobs_data):
...
+        keep_waiting = True
+        while keep_waiting:
+            keep_waiting = False
+            for name, file in data.items():
+                if name not in manifest:
+                    manifest[name] = {
+                        "type": "file",
+                        "path": str(Path(file).absolute()),
+                    }
+                if Path(file).exists() and "md5sum" not in manifest[name]:
```
eivindjahren, line 47: I suggest doing `Path(file)` once:

```python
path = Path(file)
```

and then calling it throughout:

```python
if path.exists()
```

eivindjahren (changes requested):
Just some initial comments to simplify reading to make it easier to look for errors.

src/ert/scheduler/job.py
```diff
@@ -165,7 +166,58 @@ async def _max_runtime_task(self) -> None:
...
+            return
+
+        keep_waiting = True
+        while keep_waiting:
+            keep_waiting = False
+            file_paths = []
+            try:
+                file_paths = [info["path"] for _, info in manifest.items()]
+            except (LookupError, AttributeError) as e:
+                logger.exception(f"Invalid manifest exception {e}")
```
eivindjahren, line 196: LookupError and AttributeError should never happen. Instead of logging it and silently returning, I suggest we instead have robust handling of any exception thrown by `_check_manifest`.

src/ert/scheduler/job.py
```diff
@@ -165,7 +166,58 @@ async def _max_runtime_task(self) -> None:
         )
         self.returncode.cancel()
 
+    async def check_manifest(self, timeout: int = 120) -> None:
+        if self._scheduler.manifest is None:
+            return
```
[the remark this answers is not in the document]
DanSava: It was required because of some tests that are mocking the creation of the runpath and in those cases no manifest file is in the runpath so I either had to modify those tests or add this safeguard here, but I agree it is a bit strange I will refactor and remove it.

src/ert/scheduler/scheduler.py
```diff
@@ -112,6 +117,19 @@ def __init__(
         self._ee_cert = ee_cert
         self._ee_token = ee_token
         self._publisher_done = asyncio.Event()
+        self._consumer_started = asyncio.Event()
+        self.manifest: Optional[Dict[str, Dict[str, Any]]] = None
```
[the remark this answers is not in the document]
DanSava: Yes, it makes more sense to have it initially as a `defaultdict()` but the same reasoning as above lead to this implementation. I will refactor and try to avoid it.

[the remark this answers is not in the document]
eivindjahren: :+1:

src/_ert_forward_model_runner/runner.py
```diff
@@ -23,14 +28,42 @@ def __init__(self, jobs_data):
...
+                    }
+                path_exists = path.exists()
+                if path_exists and "md5sum" not in manifest[name]:
+                    md5sum = hashlib.md5(path.read_bytes()).hexdigest()
+                    manifest[name]["md5sum"] = md5sum
+                elif not path_exists:
+                    timeout -= 1
+                    keep_waiting = timeout > 0
+                    if keep_waiting:
+                        time.sleep(1)
```
xjules, line 54: Isn't sleep(1) too generous? maybe like 0.2 :man_shrugging: ?

src/_ert_forward_model_runner/runner.py
```diff
@@ -23,14 +28,42 @@ def __init__(self, jobs_data):
...
+                    manifest[name] = {
+                        "type": "file",
+                        "path": str(path.absolute()),
+                    }
+                path_exists = path.exists()
+                if path_exists and "md5sum" not in manifest[name]:
+                    md5sum = hashlib.md5(path.read_bytes()).hexdigest()
+                    manifest[name]["md5sum"] = md5sum
+                elif not path_exists:
+                    timeout -= 1
```
xjules, line 51: `timeout` seems more like the number of retries :)

src/_ert_forward_model_runner/runner.py
```diff
@@ -23,14 +28,42 @@ def __init__(self, jobs_data):
 
         self._set_environment()
 
+    def _manifest_dict(self):
+        with open("manifest.json", mode="r", encoding="utf-8") as f:
+            data = json.load(f)
+        timeout = 10
```
xjules, line 34: I'd say that the total timeout should be more than 10 seconds.

src/_ert_forward_model_runner/runner.py
DanSava: All the logic around waiting for the files to be available on the runner side is up for discussion. I added it because I triggered this while creating a test scenario. But I am not sure if this a thing that can happen or not on the runner side or the synk issue is only for the moment when `forward_model_ok` is called.

src/_ert_forward_model_runner/runner.py
DanSava: Removed this logic after a discussion with @eivindjahren

src/_ert_forward_model_runner/runner.py
DanSava: Removed this logic after a discussion with @eivindjahren

src/_ert_forward_model_runner/runner.py
DanSava: Removed this logic after a discussion with @eivindjahren

src/ert/config/ert_config.py
```diff
@@ -471,6 +471,23 @@ def _create_list_of_forward_model_steps_to_run(
     def forward_model_step_name_list(self) -> List[str]:
         return [j.name for j in self.forward_model_steps]
 
+    def manifest_to_json(self, iens: int = 0) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
DanSava: Split the manifest concept into  `maifest` just file names that the `runner` will use to build the `Checksum`  message that contains a `checksum_dict` with file paths and respective checksum if they are available.

src/ert/scheduler/job.py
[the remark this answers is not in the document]
DanSava: Refactor to avoid this ugly logic.

src/_ert_forward_model_runner/runner.py
```diff
@@ -23,14 +27,33 @@ def __init__(self, jobs_data):
...
+        return {
+            name: {"type": "file", "path": str(Path(file).absolute())}
+            for name, file in data.items()
+        }
+
+    def _populate_checksums(self, manifest):
+        for info in manifest.values():
+            path = Path(info["path"])
+            path_exists = path.exists()
+            if path_exists:
```
eivindjahren, line 44:
```suggestion
            if path.exists():
```

tests/unit_tests/forward_model_runner/test_jobmanager.py
```diff
@@ -122,6 +122,31 @@ def test_run_multiple_ok():
         assert os.path.getsize(f"mkdir_err.{dir_number}") == 0
 
 
+@pytest.mark.usefixtures("use_tmpdir")
```
eivindjahren, line 125: There should be 2-3 more tests:

* If the forward model steps do not produce a file specified in the manifest then the user should get a failed realization with a nice error message.
* If a file with the expected checksum is present for the forward model runner but not for scheduler a warning should be displayed followed by an error from the ok callback when there is no file and no error when there is a file ok_callback can load.

tests/integration_tests/test_cli.py
```diff
@@ -75,6 +76,52 @@ def test_that_oom_kills_are_reported():
...
+
+                def _evaluate(coeffs, x):
+                    return coeffs["a"] * x**2 + coeffs["b"] * x + coeffs["c"]
+
+                def run_process():
+                    Path("run_me").write_text(
+                        "import time\\n"
+                        "from pathlib import Path\\n"
+                        "if __name__ == '__main__':\\n"
+                        "\\ttime.sleep(0.5)\\n"
```
eivindjahren, line 102: I think we should really avoid testing this with sleeps as this creates tests that are unreliable and runs long. Instead, we can inject removing or changing the file as separate steps, either as forward models (`MOVE_FILE` or `DELETE_FILE`) or custom hooked workflows.

src/ert/scheduler/job.py
```diff
@@ -165,6 +168,42 @@ async def _max_runtime_task(self) -> None:
...
+                    expected_md5sum == hashlib.md5(file_path.read_bytes()).hexdigest()
+                )
+                if matching:
+                    logger.debug(f"File {file_path} checksum successful.")
+                else:
+                    logger.warning(f"File {file_path} checksum failed.")
+            elif file_path.exists() and expected_md5sum is None:
+                logger.warning(f"Checksum not received for file {file_path}")
+            else:
+                logger.error(f"File {file_path} not syncronized after {timeout}s")
```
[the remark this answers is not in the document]
DanSava: How about:

``` python
for info in checksum.values():
    if "path" not in info:
        continue
             .....
```

eivindjahren:
Looks like one of the tests are flaky:
```
FAILED tests/unit_tests/ensemble_evaluator/test_scheduler.py::test_scheduler_receives_checksum_and_waits_for_disk_sync - RuntimeError: Task <Task pending name='Task-1839' coro=<Event.wait() running at /opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/asyncio/locks.py:309> cb=[_done_callback() at /opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/_ert/async_utils.py:39, _release_waiter(<Future pendi...c6aee8520>()]>)() at /opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/asyncio/tasks.py:429]> got Future <Future pending> attached to a different loop
```

tests/unit_tests/scheduler/test_job.py
```diff
@@ -153,3 +158,101 @@ async def test_job_run_sends_expected_events(
...
+    assert error in log_msgs
+
+    await assert_scheduler_events(
+        scheduler, [State.SUBMITTING, State.PENDING, State.RUNNING, State.COMPLETED]
+    )
+
+
+@pytest.mark.usefixtures("use_tmpdir")
+@pytest.mark.asyncio
+async def test_job_run_logs_md5missmatch_warning(realization: Realization):
```
eivindjahren, line 228: Maybe name it `test_when_checksums_do_not_match_a_warning_is_logged`

tests/unit_tests/scheduler/test_job.py
```diff
@@ -153,3 +158,101 @@ async def test_job_run_sends_expected_events(
...
+
+    assert "Disk syncronization failed for does/not/exist" in log_msgs
+
+    await assert_scheduler_events(
+        scheduler, [State.SUBMITTING, State.PENDING, State.RUNNING, State.COMPLETED]
+    )
+
+
+@pytest.mark.asyncio
+async def test_job_run_logs_manifest_errors(realization: Realization, monkeypatch):
```
eivindjahren, line 195: Maybe name it `test_when_files_in_manifest_are_not_created_an_error_is_logged`.

tests/unit_tests/scheduler/test_job.py
```diff
@@ -153,3 +158,101 @@ async def test_job_run_sends_expected_events(
...
+    with captured_logs(log_msgs, logging.ERROR):
+        job_run_task = asyncio.create_task(job.run(asyncio.Semaphore(), max_submit=1))
+        job.started.set()
+        job.returncode.set_result(0)
+        await job_run_task
+
+    assert error in log_msgs
+
+    await assert_scheduler_events(
+        scheduler, [State.SUBMITTING, State.PENDING, State.RUNNING, State.COMPLETED]
```
eivindjahren, line 222: Shouldn't this be a `State.FAILED`. What do you think @xjules ?

tests/unit_tests/scheduler/test_job.py
```diff
@@ -153,3 +158,101 @@ async def test_job_run_sends_expected_events(
         runpath=Path(realization.run_arg.runpath),
     )
     assert scheduler.driver.submit.call_count == max_submit
+
+
+@pytest.mark.asyncio
+async def test_job_run_logs_disk_sync_issue(realization: Realization, monkeypatch):
```
eivindjahren, line 164: The scenario in this test is not completely clear. Is it that the `_verify_checksum` times out? Maybe name it `test_when_waiting_for_disk_sync_times_out_an_error_is_logged`.

tests/unit_tests/forward_model_runner/test_jobmanager.py
```diff
@@ -123,6 +123,62 @@ def test_run_multiple_ok():
...
+def test_run_multiple_jobs_just_one_manifest():
```
eivindjahren, line 127: Not sure what is being tested here, so some better name of the test would be great.

tests/unit_tests/ensemble_evaluator/test_scheduler.py
```diff
@@ -0,0 +1,71 @@
...
+    tmpdir, make_ensemble_builder, monkeypatch, caplog, run_monitor_in_loop
+):
+    num_reals = 1
+    custom_port_range = range(1024, 65535)
+    done_renaming = asyncio.Event()
+
+    async def rename_and_wait():
+        Path("real_0/job_test_file").rename("real_0/test")
+        done_renaming.set()
+        await asyncio.sleep(5)
```
eivindjahren, line 24: I guess you could give `rename_and_wait`  an event to wait for instead of the sleep which you then set once `"Waiting for disk syncronization"` is logged ?

tests/unit_tests/scheduler/test_job.py
DanSava: `forward_model_ok` is mocked in this tests hence the `State.Completed` final event. If it was not mocked then it would log another error when the file is not available. I can just remove the checking for the states that is not the point of this test.

tests/unit_tests/forward_model_runner/test_jobmanager.py
DanSava: Renamed the tests

eivindjahren (approved):
Looks good now!

src/ert/scheduler/job.py
```diff
@@ -167,6 +170,49 @@ async def _max_runtime_task(self) -> None:
         )
         self.returncode.cancel()
 
+    async def _verify_checksum(self, timeout: int = 120) -> None:
+        # Wait for job runpath to be in the checksum dictionary
+        runpath = self.real.run_arg.runpath
+        while runpath not in self._scheduler.checksum:
+            if timeout <= 0:
```
eivindjahren, line 177: This loop never runs in the tests. Should probably have a look at why that is. Should be something we either don't require or test for.

DanSava: Removing this will assume the checksum information reaches the scheduler once the job is complete. This is the case now but was not true when these checksum messages were batched in the ensemble evaluator `BatchingDispatcher`.

I want to keep it to guard against possible connectivity issues that might delay these messages. I will add a test that hits this loop as well

eivindjahren: :+1:

PR #7944, opened 2024-05-22
Have lsf driver log missing jobid in kill to debug instead of error

### LINKED ISSUE #7907: Killing LSF jobs while submission is in progress can leave the job running on the cluster.
If an ensemble experiment is started, and 'Terminate experiment' is clicked while realizations are being submitted, the following "Error" is displayed to the user:

https://github.com/equinor/ert/blob/401d29f04aef827f27011482f42cf6c6a153e8fb/src/ert/scheduler/lsf_driver.py#L289-L291

image

This means that the killing of the realization failed. When the `submit()` function eventually finished, the realization is likely to be running even if it should have been killed.

jonathan-eq:
**Issue**
Resolves #7907 

**Approach**
🔢 

## When applicable

### DIFF
--- a/src/ert/scheduler/lsf_driver.py
+++ b/src/ert/scheduler/lsf_driver.py
@@ -287,8 +287,15 @@ class LsfDriver(Driver):
 
     async def kill(self, iens: int) -> None:
         if iens not in self._iens2jobid:
-            logger.error(f"LSF kill failed due to missing jobid for realization {iens}")
-            return
+            for _ in range(10):
+                await asyncio.sleep(self._poll_period)
+                if iens in self._iens2jobid:
+                    break
+            else:
+                logger.error(
+                    f"LSF kill failed due to missing jobid for realization {iens}"
+                )
+                return
 
         job_id = self._iens2jobid[iens]
 
--- a/src/ert/scheduler/lsf_driver.py
+++ b/src/ert/scheduler/lsf_driver.py
@@ -87,9 +87,13 @@ class _Stat(BaseModel):
 
 
 class JobData(BaseModel):
-    iens: int
-    job_state: AnyJob
-    submitted_timestamp: float
+    job_id: Optional[str] = None
+    job_state: Optional[AnyJob] = None
+    submitted_timestamp: Optional[float] = None
+    submitted_future: asyncio.Future[bool] = Field(default_factory=asyncio.Future)
+
+    class Config:
+        arbitrary_types_allowed = True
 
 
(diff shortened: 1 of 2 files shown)


berland:
I think this is an actual bug that this code (before and after the PR) hides. If we issue `kill()` right after `submit()`, but before `_execute_with_retry()` finishes, the job will be lingering on the compute cluster 🤔

jonathan-eq:
> I think this is an actual bug that this code (before and after the PR) hides. If we issue `kill()` right after `submit()`, but before `_execute_with_retry()` finishes, the job will be lingering on the compute cluster 🤔

Okay, so as I understand it; we are potentially trying to kill a job before qsub has been called, or has not finished, or we havent processed the qsub output to get jobid. 
Should we add a simple loop and sleep between each attempt, or should we use the full scale `_execute_with_retry(...)` for killing jobs?

src/ert/scheduler/lsf_driver.py
```diff
@@ -204,8 +208,8 @@ def __init__(
         self._bjobs_cmd = Path(bjobs_cmd or shutil.which("bjobs") or "bjobs")
         self._bkill_cmd = Path(bkill_cmd or shutil.which("bkill") or "bkill")
 
-        self._jobs: MutableMapping[str, JobData] = {}
-        self._iens2jobid: MutableMapping[int, str] = {}
+        self._iens2jobdata: MutableMapping[int, JobData] = {}
```
berland, line 211: why change iens from `str` to `int` in this pr?

src/ert/scheduler/lsf_driver.py
```diff
@@ -87,9 +87,13 @@ class _Stat(BaseModel):
 
 
 class JobData(BaseModel):
-    iens: int
-    job_state: AnyJob
-    submitted_timestamp: float
+    job_id: Optional[str] = None
+    job_state: Optional[AnyJob] = None
+    submitted_timestamp: Optional[float] = None
+    submitted_future: asyncio.Future[bool] = Field(default_factory=asyncio.Future)
```
berland, line 93: "submitted" is a little bit ambigous, it could be after submit() is called, or when when bsub is called, or when bsub is finished etc. Maybe `submit_function_finished` or something better?

berland: Think about how this will be for resubmissions.

If it interferes with that, maybe we need a lock mechanism instead, "submit_is_in_progress" ?

berland:
Closing this as the original issue has been reformulated.

PR #7960, opened 2024-05-23
Add Sumo authentication validation

adnejacobsen:
If the config contains a Sumo forward model, raise an error if the user is not authenticated.

I'm not expecting this PR to be merged as is, it's just showing what we would like to achieve and we hope to get some feedback on what our options are when it comes to "custom" validation and authentication. 

Skjermbilde 2024-05-23 kl  12 20 57

### DIFF
--- a/src/ert/config/ert_config.py
+++ b/src/ert/config/ert_config.py
@@ -9,6 +9,7 @@ from datetime import datetime
 from fnmatch import fnmatch
 from os import path
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


sondreso:
This is now possible using the system introduced in https://github.com/equinor/ert/pull/7743 :tada: Closing this PR, but feel free to re-open if there is anything else you would like to see or have questions/comments!

PR #7965, opened 2024-05-24
Move get_machine_name to public API

### LINKED ISSUE #7888: Move get_machine_name to a more public looking place
This function is imported by ert-configurations and should thus not be hidden in the config of the ensemble_evaluator. A more appropriate place is perhaps `ert.shared`.

https://github.com/equinor/ert/blob/a02d71d91e48b8df1ebdcf4326c3ad68f1a1a19a/src/ert/ensemble_evaluator/config.py#L27-L34

berland:
As this function is used outside of ert, it should not be hidden below ensemble_evalutor.config.

port_handler.py serves as a placeholder for the code, but the function but exposed and should be used as ert.shared.get_machine.name()

**Issue**
Resolves #7888 

**Approach**
🚚 

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/config.py
+++ b/src/ert/ensemble_evaluator/config.py
@@ -6,6 +6,7 @@ import socket
 import ssl
 import tempfile
 import typing
+import warnings
 from base64 import b64encode
 from datetime import datetime, timedelta
 from typing import Optional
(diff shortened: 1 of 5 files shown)


src/ert/ensemble_evaluator/config.py
```diff
@@ -15,8 +16,8 @@
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import rsa
 from cryptography.x509.oid import NameOID
-from dns import exception, resolver, reversename
 
+from ert.shared import get_machine_name as ert_shared_get_machine_name
```
xjules, line 20: Can this be only `from ert.shared import get_machine_name as _get_machine_name`?

xjules (approved):
Looks good! Just had a small comment about the name

src/ert/ensemble_evaluator/config.py
berland: Hmm, I think i prefer the explicitness of the long name. Nonetheless, we will change this again in short time when we remove the deprecated (private) API.

PR #7969, opened 2024-05-24
Add logging of ert.shared.plugins to Azure

### LINKED ISSUE #7917: Adding of logging handler happens too late to catch which jobs are used
https://github.com/equinor/ert/blob/010ac789ae7977b260f85bd25e65f93b1cb17920/src/ert/__main__.py#L668

Adding the logger happens inside the ErtPluginContext, but this is too late to catch log messages about which plugins are used.

larsevj:
**Issue**
Resolves #7917 

## When applicable

### DIFF
--- a/src/ert/__main__.py
+++ b/src/ert/__main__.py
@@ -664,8 +664,7 @@ def main() -> None:
 
     FeatureScheduler.set_value(args)
     try:
-        with ErtPluginContext() as context:
-            context.plugin_manager.add_logging_handle_to_root(logging.getLogger())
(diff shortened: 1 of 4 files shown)


andreas-el:
> I am a bit unsure why some log statements are set to debug and others to info, and if this should be modified or just left as is?

I think in general we should either change the levels or the logging, so my take on this is to leave levels as-is for now.
We could open an issue regarding the levels though, if you want to follow up on that. Or discuss this during an appropriate meeting.

andreas-el (approved):
This looks fine to me. 💯

PR #7999, opened 2024-05-28
Compare only obs dataset structure

yngve-sk:
Speeds up slow running, sometimes hanging hypothesis test in `test_local_storage.py`. Note; `dims` holds the dimension names and lengths.

Screenshot 2024-05-28 at 13 57 32

### DIFF
--- a/tests/unit_tests/storage/test_local_storage.py
+++ b/tests/unit_tests/storage/test_local_storage.py
@@ -643,9 +643,9 @@ class StatefulStorageTest(RuleBasedStateMachine):
             list(storage_experiment.response_configuration.values())
             == model_experiment.responses
         )
-        assert model_experiment.observations == pytest.approx(
-            storage_experiment.observations
-        )
+
+        for ds_name, ds in model_experiment.observations.items():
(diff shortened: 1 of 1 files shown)


tests/unit_tests/storage/test_local_storage.py
```diff
@@ -643,9 +643,9 @@ def get_experiment(self, model_experiment: Experiment):
             list(storage_experiment.response_configuration.values())
             == model_experiment.responses
         )
-        assert model_experiment.observations == pytest.approx(
```
oyvindeide, line 646: But it also previously did a full comparison of datasets? How can it then go from ~ 1 minute to ~ 1 hour? Or is it compounding errors?

yngve-sk: Not sure exactly how or why, though reducing data has also made it faster so most likely this is/should not be a problem, ref this profiling shows that the comparison was like ~40% of the time spent, though it is scaled by the amount of data generated. 
Screenshot 2024-05-28 at 13 32 03

PR #8009, opened 2024-05-29
Round export misfit snapshot to 10 decimals

yngve-sk:
Older pandas version produces slightly different results, this nullifies that difference

**Issue**
Resolves https://github.com/equinor/komodo-releases/issues/5851


andreas-el (approved):
💯

PR #8019, opened 2024-05-29
Add lock mechanism while submitting and killing LSF jobs

### LINKED ISSUE #7907: Killing LSF jobs while submission is in progress can leave the job running on the cluster.

berland:
**Issue**
Resolves #7907 

**Approach**
🔒 

## When applicable

### DIFF
--- a/src/ert/scheduler/lsf_driver.py
+++ b/src/ert/scheduler/lsf_driver.py
@@ -217,6 +217,8 @@ class LsfDriver(Driver):
         self._bhist_required_cache_age: float = 4
         self._bhist_cache_timestamp: float = time.time()
 
+        self._submit_locks: MutableMapping[int, asyncio.Lock] = {}
+
     async def submit(
         self,
         iens: int,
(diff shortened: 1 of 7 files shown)


src/ert/scheduler/lsf_driver.py
```diff
@@ -227,6 +229,11 @@ async def submit(
         name: str = "dummy",
         runpath: Optional[Path] = None,
     ) -> None:
+        if iens not in self._submit_locks:
+            self._submit_locks[iens] = asyncio.Lock()
+
+        await self._submit_locks[iens].acquire()
```
xjules, line 235: when looking at the code I would prefer to use the:
```python
async with self._submit_locks[iens]:
 ....
```
which then automatically releases the lock

berland: I sort of agree, but I dislike very long indented blocks changing so many lines of code..

berland:
Testing the lock mechanism is blocked by #8020

src/ert/scheduler/lsf_driver.py
berland: Using async.with now.

xjules: I think it is required to use the `async with Lock` as this one guarantees to release the lock when exception occurs :+1:

berland: fixed.

tests/integration_tests/scheduler/test_lsf_driver.py
```diff
@@ -217,3 +218,47 @@ def mock_poll_once_by_bhist(*args, **kwargs):
...
+    # Allow submit and kill to be interleaved by asyncio by issuing
+    # submit() in its own asyncio Task:
+    asyncio.create_task(
+        driver.submit(
+            # The sleep is the time window in which we can kill the job before
+            # the unwanted finish message appears on disk.
+            0,
+            "sh",
+            "-c",
+            f"sleep 1; touch {tmp_path}/survived",
```
xjules, line 249: Shall we sleep (eg 5 secs) more here? Just to remove potential flakiness....

berland: Added some if's to guard against flakyness. This test is also relevant in interactive pytest runs so then I don't want to sleep for too long.

tests/integration_tests/scheduler/test_lsf_driver.py
```diff
@@ -217,3 +219,60 @@ def mock_poll_once_by_bhist(*args, **kwargs):
...
+
+    async def finished(iens: int, returncode: int):
+        SIGTERM = 15
+        assert iens == 0
+        assert returncode == SIGNAL_OFFSET + SIGTERM
+
+    await poll(driver, {0}, finished=finished)
+
+    assert "ERROR" not in str(caplog.text)
+    await asyncio.sleep(test_grace_time)
```
xjules, line 277: Why do we need to sleep here?

berland: We have in the `poll()` await ensured that the realization has been reported back to the scheduler with  a correct error code. But, if that is erroneous, and the shell script that is running is not killed (it is the mother process of the shell script that is actually killed), then the `survived` file would still appear on disk. So we wait until we know the survived file would have been written and then assert it is not there.

xjules: Right, but that deserves a comment before the `await asyncio.sleep(test_grace_time)` line.

berland: added.

xjules (approved):
Very nice PR @berland ! :rocket:

PR #8031, opened 2024-05-30
Make mocked LSF and OpenPBS systems able to kill

### LINKED ISSUE #8020: Mocked LSF system in integration testing does not allow killing
This integration test function will fail
```python
async def test_kill(tmp_path):
    os.chdir(tmp_path)
    driver = LsfDriver()
    await driver.submit(
            # The sleep is the time window in which we can kill the job before
            # the unwanted finish message appears on disk.
            0, "sh", "-c", f"sleep 1; echo I survived > {tmp_path}/test"
        )
    await driver.kill(0)
    await asyncio.sleep(2)  # Give the script a chance to finish if it is running
    assert not Path("test").exists(), "job output should not have been written"
```

berland:
For LSF it is necessary to manually loop over grandchildren in the mocked runner to pass the added test. For OpenPBS the test will pass without the loop over grandchildren, but it was added nonetheless.

**Issue**
Resolves #8020

**Approach**
Explicitly kill grandchildren in mocked LSF.

## When applicable


tests/integration_tests/scheduler/bin/lsfrunner
```diff
@@ -6,13 +6,17 @@ function handle_sigterm {
...
+        kill -s SIGTERM $grandchild
+    done
+    kill -s SIGTERM $child_pid
     exit 1
 }
 
 trap handle_sigterm SIGTERM
 
 echo "$$" > "${job}.pid"
+cat ${job}.script >> /tmp/mylog
```
xjules, line 19: Is this a left over from testing?

berland: yes! will remove

berland: fixed

tests/integration_tests/scheduler/test_generic_driver.py
```diff
@@ -106,3 +108,39 @@ async def finished(iens, returncode):
...
+        "-c",
+        f"sleep {job_kill_window}; touch {tmp_path}/survived",
+        name="kill_me",
+    )
+    await poll(driver, {0}, started=started)
+
+    # Give the script a chance to finish if it is running
+    await asyncio.sleep(test_grace_time)
+
+    assert not Path("survived").exists(), "Job should have been killed"
```
xjules, line 146: How does this guarantee that the job actually have run; ie. submitted correctly? Maybe like to touch a file `file_start` first (before sleep) and then check that the `file_start` is there while `survived` is not?

berland: This is ensured by `await poll(driver, {0}, started=started)`

xjules (approved):
Very good! :+1:

PR #8047, opened 2024-06-03
RuntimeInfo added to storage

### LINKED ISSUE #8046: Add Runtime information per realization
Provide runtime information, specifically when a job has been running on a cluster and distinguish between each submit (i.e. if a job has been resubmitted I would like to know runtime for each submit).

The runtime information is necessary to determine total runtime for the job (i.e. only the time spent when a job has been submitted, not any time spent loading data)

**Additional context**
Will be used to provide more information related to running a job, for instance if there is cost associated.

lars-petter-hauge:
RuntimeInfo represents information related to a job that has been submitted, it does not include anything afterwards (such as loading results). Resubmitting will not overwrite, rather append for each submit

**Issue**
Resolves #8046

### DIFF
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -2,9 +2,9 @@ from __future__ import annotations
 
 import asyncio
 import logging
-import time
 import uuid
 from contextlib import suppress
+from datetime import datetime
 from enum import Enum
 from pathlib import Path
 from typing import TYPE_CHECKING, List, Optional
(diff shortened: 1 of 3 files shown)


sondreso:
Hi! I think it is nice that we start to store this information about runtime information that has previously been stored temporarily in the client. However, I think it is not optimal that we are not using this information in ERT as well, as this lead to potentially multiple sources of the same information. Is it possible to also use this info in the clients.

sondreso:
Closing this for now since there has not been any activity recently, feel free to re-open in the future!

PR #8052, opened 2024-06-04
Use plugin fm steps in LibresFacade

yngve-sk:
Tested on RGS, adding
`FORWARD_MODEL GENDATA_RFT` to `test-data/poly_example/poly.ert`, then opening a python context and doing
```
LibresFacade.from_config_file("test-data/poly_example/poly.ert")
```
gives the error on bleeding, but works with this PR

**Issue**
Resolves https://github.com/equinor/komodo-releases/issues/5907

### DIFF
--- a/src/ert/libres_facade.py
+++ b/src/ert/libres_facade.py
@@ -32,6 +32,7 @@ from ert.data._measured_data import ObservationError, ResponseError
 from ert.load_status import LoadResult, LoadStatus
 
 from .enkf_main import EnKFMain, ensemble_context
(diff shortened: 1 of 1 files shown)


andreas-el (approved):
Seems to do the trick! 💯

andreas-el:
Did you also try some other forward models while you were at it? Not only `GENDATA_RFT` ?

yngve-sk:
> Did you also try some other forward models while you were at it? Not only `GENDATA_RFT` ?

Good idea, did a double check just now, seems to still work :)

PR #8058, opened 2024-06-04
Merge snapshot and partialsnapshot

### LINKED ISSUE #8054: Merge snapshot and partialsnapshot
**Describe the requested improvement**
It seems like a snapshot is just a partialsnapshot, but with some extra methods connecting it to the ensemble.
Either merge those two, or rename snapshot (`ensemble_state` maybe?) 

Removing the redundant snapshot class will improve readability of this part of the codebase.

jonathan-eq:
**Issue**
Resolves #8054 

**Approach**

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/_ensemble.py
+++ b/src/ert/ensemble_evaluator/_ensemble.py
@@ -162,7 +162,6 @@ class LegacyEnsemble:
                     event, snapshot_mutate_event
                 )
             )
-        # JONAK self.snapshot.merge_event(snapshot_mutate_event)
         if self.snapshot.status is not None and self.status != self.snapshot.status:
             self.status = self._status_tracker.update_state(self.snapshot.status)
         return snapshot_mutate_event
--- a/src/ert/ensemble_evaluator/_ensemble.py
+++ b/src/ert/ensemble_evaluator/_ensemble.py
@@ -31,7 +31,6 @@ from .config import EvaluatorServerConfig
 from .identifiers import EVTYPE_ENSEMBLE_FAILED, EVTYPE_ENSEMBLE_STARTED
 from .snapshot import (
     ForwardModel,
-    PartialSnapshot,
     RealizationSnapshot,
     Snapshot,
     SnapshotDict,
@@ -149,16 +148,21 @@ class LegacyEnsemble:
             metadata=self.metadata,
         )
 
-        return Snapshot(top.model_dump())
+        return Snapshot.from_nested_dict(top.model_dump())
 
     def get_successful_realizations(self) -> List[int]:
         return self.snapshot.get_successful_realizations()
 
-    def update_snapshot(self, events: List[CloudEvent]) -> PartialSnapshot:
-        snapshot_mutate_event = PartialSnapshot(self.snapshot)
+    def update_snapshot(self, events: List[CloudEvent]) -> Snapshot:
+        snapshot_mutate_event = Snapshot()
+
         for event in events:
-            snapshot_mutate_event.from_cloudevent(event)
-        self.snapshot.merge_event(snapshot_mutate_event)
+            snapshot_mutate_event = (
+                self.snapshot.update_from_cloudevent_and_generate_update_snapshot(
+                    event, snapshot_mutate_event
+                )
+            )
+        # JONAK self.snapshot.merge_event(snapshot_mutate_event)
         if self.snapshot.status is not None and self.status != self.snapshot.status:
             self.status = self._status_tracker.update_state(self.snapshot.status)
         return snapshot_mutate_event
--- a/src/ert/ensemble_evaluator/snapshot.py
+++ b/src/ert/ensemble_evaluator/snapshot.py
@@ -235,8 +235,10 @@ class Snapshot:
             ids.EVTYPE_REALIZATION_TIMEOUT,
         }:
             end_time = convert_iso8601_to_datetime(timestamp)
-        self.update_realization(_get_real_id(e_source), status, start_time, end_time)
-
+        update_event_snapshot.update_realization(
+            _get_real_id(e_source), status, start_time, end_time
+        )
+        self._realization_states.update(update_event_snapshot._realization_states)
         if e_type == ids.EVTYPE_REALIZATION_TIMEOUT:
             for (
                 forward_model_id,
@@ -246,16 +248,18 @@ class Snapshot:
                     real_id = _get_real_id(e_source)
                     forward_model_idx = (real_id, forward_model_id)
                     if forward_model_idx not in self._forward_model_states:
-                        self._forward_model_states[forward_model_idx] = {}
-                    self._forward_model_states[forward_model_idx].update(
-                        {
-                            "status": state.FORWARD_MODEL_STATE_FAILURE,
-                            "end_time": end_time,
-                            "error": "The run is cancelled due to "
-                            "reaching MAX_RUNTIME",
-                        }
+                        update_event_snapshot._forward_model_states[
+                            forward_model_idx
+                        ] = {}
+                    update_event_snapshot._forward_model_states[forward_model_idx] = {
+                        "status": state.FORWARD_MODEL_STATE_FAILURE,
+                        "end_time": end_time,
+                        "error": "The run is cancelled due to " "reaching MAX_RUNTIME",
+                    }
+                    self._forward_model_states.update(
+                        update_event_snapshot._forward_model_states
                     )
-        return self
+        return update_event_snapshot
 
     def _handle_forward_model_event(
         self, event: CloudEvent, update_event_snapshot: "Snapshot"
--- a/src/ert/ensemble_evaluator/snapshot.py
+++ b/src/ert/ensemble_evaluator/snapshot.py
@@ -78,6 +78,12 @@ _ENSEMBLE_TYPE_EVENT_TO_STATUS = {
 def convert_iso8601_to_datetime(
     timestamp: Union[datetime, str],
 ) -> datetime:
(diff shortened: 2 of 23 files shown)


sondreso:
This has potentially very large consequences for the amount of data that is transferred and consequently the performance of the application, has this been tested with the big poly case? 🤔

xjules:
> This has potentially very large consequences for the amount of data that is transferred and consequently the performance of the application, has this been tested with the big poly case? 🤔

Could you @jonathan-eq put some comments from your previous discussion with @sondreso? For the reviewing sake that is

jonathan-eq:
> > This has potentially very large consequences for the amount of data that is transferred and consequently the performance of the application, has this been tested with the big poly case? 🤔
> 
> Could you @jonathan-eq put some comments from your previous discussion with @sondreso? For the reviewing sake that is

We have to keep snapshots in both client and ensemble evaluator, so in case a websocket connection fails, we can just send a new copy without losing track.

src/ert/gui/simulation/queue_emitter.py
```diff
@@ -42,10 +42,13 @@ def consume_and_emit(self) -> None:
                 continue
 
             # pre-rendering in this thread to avoid work in main rendering thread
-            if isinstance(event, FullSnapshotEvent) and event.snapshot:
+            if (
+                isinstance(event, FullSnapshotEvent)
+                and event.snapshot
+                or isinstance(event, SnapshotUpdateEvent)
+                and event.snapshot
+            ):
```
jonathan-eq, line 50: Combine this

tests/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -25,7 +25,7 @@ async def test_restarted_jobs_do_not_have_error_msgs(evaluator):
     async with Monitor(config_info) as monitor:
         # first snapshot before any event occurs
         snapshot_event = await monitor._event_queue.get()
-        snapshot = PartialSnapshot()._from_nested_dict(snapshot_event.data)
+        snapshot = Snapshot()._from_nested_dict(snapshot_event.data)
```
jonathan-eq, line 28: Class method^

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -240,179 +208,209 @@ def _merge(self, other: "PartialSnapshot") -> "PartialSnapshot":
             self._forward_model_states[forward_model_id].update(other_fm_data)
         return self
 
-    def from_cloudevent(self, event: CloudEvent) -> "PartialSnapshot":
+    def update_from_cloudevent_and_generate_update_snapshot(
```
berland, line 211: It is not clear to me from the docstring if `self` is updated also. Since we actually return something it seems not. And then we should probably write `..updated_snapshot` and not `update_snapshot`.

jonathan-eq: This function is weird, because it updates itself with the event (inplace?), and also returns the update_snapshot only containing the change. The update snapshot (way smaller than actual snapshot) is the one we send to the GUI to avoid resource starvation.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -240,179 +208,209 @@ def _merge(self, other: "PartialSnapshot") -> "PartialSnapshot":
...
+        self._merge(Snapshot.from_nested_dict(update_as_nested_dict))
 
     def merge_metadata(self, metadata: SnapshotMetadata) -> None:
-        self._my_partial._metadata.update(metadata)
-
-    def to_dict(self) -> Dict[str, Any]:
-        return self._my_partial.to_dict()
+        """only used in gui snapshot model, which only cares about the
+        snapshot's metadata"""
+        self._metadata.update(metadata)
```
berland, line 341: Not returning anything?

berland: if this was something forgotten, was it caught by tests?

src/ert/gui/model/snapshot.py
```diff
@@ -168,8 +180,8 @@ def _add_partial_snapshot(self, partial: PartialSnapshot, iter_: int) -> None:
             logger.debug("no full snapshot yet, ignoring partial")
             return
 
-        job_infos = partial.get_all_forward_models()
```
berland, line 171: oh no, a "job" string that should be "forward_model"

src/ert/gui/model/snapshot.py
```diff
@@ -273,7 +285,13 @@ def _add_partial_snapshot(self, partial: PartialSnapshot, iter_: int) -> None:
 
             return
 
-    def _add_snapshot(self, snapshot: Snapshot, iter_: int) -> None:
+    def _process_full_snapshot(self, snapshot: Snapshot, iter_: int):
+        """Add a new snapshot for an iteration, or completely overwrite the current one.
```
berland, line 289: "or" here makes it look conditional whether it is overwritten. You probably mean it will overwrite if there is something already for a particular iteration. You can maybe say "Set snapshot for a given iteration" and comment (in paranthesis?) that anything existing will be overwritten.

tests/unit_tests/ensemble_evaluator/test_snapshot.py
```diff
@@ -138,15 +136,16 @@ def test_update_forward_models_in_partial_from_multiple_cloudevents(snapshot):
...
-    forward_models = partial.to_dict()["reals"]["0"]["forward_models"]
+    forward_models = snapshot.to_dict()["reals"]["0"]["forward_models"]
     assert forward_models["0"]["status"] == state.FORWARD_MODEL_STATE_FAILURE
     assert forward_models["1"]["status"] == state.FORWARD_MODEL_STATE_FINISHED
 
 
-def test_that_realization_success_message_updates_state(snapshot):
+def test_that_realization_success_message_updates_state(
+    snapshot,
+):  # JONAK: Why use snapshot fixture here?
```
berland, line 146: yeah, why?? :)

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -240,179 +208,209 @@ def _merge(self, other: "PartialSnapshot") -> "PartialSnapshot":
...
+        update_event_snapshot: Optional["Snapshot"] = None,
+    ) -> "Snapshot":
+        """Add the data from the cloudevent, and also generate a new snapshot containing only the updated data.
+
+        Args:
+            event (CloudEvent): The CloudEvent with the data to update.
+            update_event_snapshot (Optional[&quot;Snapshot&quot;], optional): The snapshot which should only have the updated data attached. Returns a new snapshot if None is provided.
+
+        Returns:
+            Snapshot: A snapshot only containing the updated data.
```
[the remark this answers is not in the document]
jonathan-eq: If provided, it will update the input snapshot. This is how it was done before, so we could aggregate all cloudevents into a single snapshot without having to explicitly merge it into a snapshot.
I do not like this, but changing it would change functionality and require alterations to a lot of tests. Should I do it in this PR?

src/ert/ensemble_evaluator/snapshot.py
jonathan-eq: We do not use the returned value anymore

tests/unit_tests/ensemble_evaluator/test_snapshot.py
jonathan-eq: My suspicions were correct, it was not needed :)

src/ert/ensemble_evaluator/snapshot.py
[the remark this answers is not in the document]
berland: No.

JHolba:
This seems to be a bit slower than current main. My bigpoly test (edited to fail halfway through the forward model) goes from 1m1s in main to 1m16s in this branch. The ui freezes seem to be slightly longer too, though I do not have any numbers for that.

jonathan-eq:
> This seems to be a bit slower than current main. My bigpoly test (edited to fail halfway through the forward model) goes from 1m1s in main to 1m16s in this branch. The ui freezes seem to be slightly longer too, though I do not have any numbers for that.

Hmm, I wonder what the bottleneck is. I will investigate!

JHolba:
Are we sending full snapshots every time now? The major reason for the slowdown is the parsing of dates. It went from ~51k date parses in main to ~436k date parses in this pr.

JHolba:
Getting #8310 merged would bring the difference between main and this PR down to 3-5 seconds in my test.
Should still figure out why there is so much extra parsing here though

JHolba:
#8304 may also help with performance

src/ert/ensemble_evaluator/_ensemble.py
```diff
@@ -158,16 +157,21 @@ def _create_snapshot(self) -> Snapshot:
...
+
         for event in events:
-            snapshot_mutate_event.from_cloudevent(event)
-        self.snapshot.merge_event(snapshot_mutate_event)
+            snapshot_mutate_event = (
+                self.snapshot.update_from_cloudevent_and_generate_update_snapshot(
+                    event, snapshot_mutate_event
+                )
+            )
+        # JONAK self.snapshot.merge_event(snapshot_mutate_event)
```
jonathan-eq, line 174: ...merge_event(...) is implicitely called in the ...update_from_cloudevent...() method.

jonathan-eq:
> Are we sending full snapshots every time now? The major reason for the slowdown is the parsing of dates. It went from ~51k date parses in main to ~436k date parses in this pr.

Yes, the update_realization method called in update_from_cloudevent returned self instead of the `update_snapshot`-snapshot, so it would always be a full snapshot. Working on a fix now.

PR #8060, opened 2024-06-04
Fix flaky integration test `test_generic_driver.py::test_kill_actuall…

jonathan-eq:
…y_kills`

**Issue**
Resolves #[5897](https://github.com/equinor/komodo-releases/issues/5897)

**Approach**

## When applicable

### DIFF
--- a/tests/integration_tests/scheduler/test_generic_driver.py
+++ b/tests/integration_tests/scheduler/test_generic_driver.py
@@ -110,6 +110,7 @@ async def test_kill_gives_correct_state(driver: Driver, tmp_path, request):
     assert aborted_called
 
 
+@pytest.mark.flaky(reruns=5)
 async def test_kill_actually_kills(driver: Driver, tmp_path, pytestconfig):
     os.chdir(tmp_path)
 
@@ -117,8 +118,8 @@ async def test_kill_actually_kills(driver: Driver, tmp_path, pytestconfig):
         isinstance(driver, OpenPBSDriver) and pytestconfig.getoption("openpbs")
     ):
         # Allow more time when tested on a real compute cluster to avoid false positives.
-        job_kill_window = 10
-        test_grace_time = 20
+        job_kill_window = 60
+        test_grace_time = 120
     elif sys.platform.startswith("darwin"):
         # Mitigate flakiness on low-power test nodes
         job_kill_window = 5
@@ -127,7 +128,7 @@ async def test_kill_actually_kills(driver: Driver, tmp_path, pytestconfig):
         job_kill_window = 1
         test_grace_time = 2
 
-    async def started(iens):
+    async def kill_job_once_started(iens):
         nonlocal driver
         await driver.kill(iens)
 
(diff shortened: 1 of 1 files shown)


tests/integration_tests/scheduler/test_generic_driver.py
```diff
@@ -127,7 +128,7 @@ async def test_kill_actually_kills(driver: Driver, tmp_path, pytestconfig):
         job_kill_window = 1
```
berland, line 128: I think I have seen flakiness also when --openpbs and --lfs is not set as the pytest option. Then one needs to bump these values as well, but you may just mark the test as flaky and say a couple of reruns is ok.

tests/integration_tests/scheduler/test_generic_driver.py
```diff
@@ -138,9 +140,13 @@ async def started(iens):
         f"sleep {job_kill_window}; touch {tmp_path}/survived",
         name="kill_me",
     )
-    await poll(driver, {0}, started=started)
+    await poll(driver, {0}, started=kill_job_once_started)
 
     # Give the script a chance to finish if it is running
-    await asyncio.sleep(test_grace_time)
+    async def check_if_survived_file_exists():
```
berland, line 146: I don't see how this improves the test? It will always wait  until the grace period is finished anyhow.

tests/integration_tests/scheduler/test_generic_driver.py
```diff
@@ -110,15 +111,16 @@ async def finished(iens, returncode):
...
 async def test_kill_actually_kills(driver: Driver, tmp_path, pytestconfig):
     os.chdir(tmp_path)
 
     if (isinstance(driver, LsfDriver) and pytestconfig.getoption("lsf")) or (
         isinstance(driver, OpenPBSDriver) and pytestconfig.getoption("openpbs")
     ):
         # Allow more time when tested on a real compute cluster to avoid false positives.
-        job_kill_window = 10
-        test_grace_time = 20
+        job_kill_window = 60
```
berland, line 122: Has the need for this incrase been observed?

berland:
It is probably possible to get this test to pass faster by trapping the kill signal (assuming SIGTERM) inside the shell script and write to a file called `killed` and wait for this file to exist. But that can go into another PR.

tests/integration_tests/scheduler/test_generic_driver.py
jonathan-eq: The idea was that if the test is failing after 20 seconds; we won't have to wait the full 120s.

tests/integration_tests/scheduler/test_generic_driver.py
jonathan-eq: I have not observed the flakiness since the creation of this ticket. Maybe it got scared off.

jonathan-eq: We discussed that this could be a possible remedy for the flakiness we saw.

berland: Could be correct _only if_ it is observed in the second part of the integration test where --openpbs or --lsf is used.

tests/integration_tests/scheduler/test_generic_driver.py
jonathan-eq: Okay, will remove 👌

tests/integration_tests/scheduler/test_generic_driver.py
jonathan-eq: ^it was apparently observed on bleeding, and on every build currently.

berland (approved):
:+1:

PR #8066, opened 2024-06-04
New LSF driver should allocate correct number of cpus

xjules:
This backports support for NUM_CPU into Scheduler/LSFDriver

The corresponding commit.

9e4c1448fca5a150811d460af11f0ccafd774dab


andreas-el (approved):
💯

PR #8085, opened 2024-06-05
Move some observation data logic away from StorageInfoWidget

yngve-sk:
to `local_experiment.py`
**Issue**
Resolves https://github.com/equinor/ert/issues/8081

### DIFF
--- a/src/ert/storage/ensure_correct_xr_coordinate_order.py
+++ b/src/ert/storage/ensure_correct_xr_coordinate_order.py
@@ -1,0 +1,24 @@
+import xarray as xr
+
+
+def ensure_correct_coordinate_order(ds: xr.Dataset) -> xr.Dataset:
+    """
+    Ensures correct coordinate order or response/param dataset.
+    Slightly less performant than not doing it, but ensure the
+    correct coordinate order is applied when doing .to_dataframe().
+    It is possible to omit using this and instead pass in the correct
+    dim order when doing .to_dataframe(), which is always the same as
+    the .dims of the first data var of this dataset.
+    """
+    # Just to make the order right when
+    # doing .to_dataframe()
+    # (it seems notoriously hard to tell xarray to just reorder
+    # the dimensions/coordinate labels)
+    data_vars = list(ds.data_vars.keys())
+
+    # We assume only data vars with the same dimensions,
+    # i.e., (realization, *index) for all of them.
+    dim_order_of_first_var = ds[data_vars[0]].dims
+    return ds[[*dim_order_of_first_var, *data_vars]].sortby(
+        dim_order_of_first_var[0]  # "realization" / "realizations"
+    )
--- a/src/ert/gui/ertwidgets/storage_info_widget.py
+++ b/src/ert/gui/ertwidgets/storage_info_widget.py
@@ -179,16 +179,16 @@ class _EnsembleWidget(QWidget):
             return
 
         observation_label = selected.data(0, Qt.ItemDataRole.DisplayRole)
-        observations_dict = self._ensemble.experiment.observations
 
         self._figure.clear()
         ax = self._figure.add_subplot(111)
         ax.set_title(observation_name)
         ax.grid(True)
 
-        observation_ds = observations_dict[observation_name]
-
-        response_name = observation_ds.attrs["response"]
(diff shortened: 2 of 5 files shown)


yngve-sk:
Screenshot 2024-06-06 at 10 25 12

yngve-sk:
Screenshot 2024-06-06 at 11 54 33

src/ert/storage/ensure_correct_xr_coordinate_order.py
```diff
@@ -0,0 +1,24 @@
+import xarray as xr
+
+
+def ensure_correct_coordinate_order(ds: xr.Dataset) -> xr.Dataset:
```
dafeda, line 4: I don't understand this. Do you have an example I can play around with?

yngve-sk: TLDR; `.to_dataframe()` doesn't "preserve" the index, giving a a dataframe with this index
```
>          new_combined_ds.to_dataframe().index
>          names=['name', 'index', 'realization', 'report_step'], length=500)
```
instead of this
```
>           ensure_correct_coordinate_order(new_combined_ds).to_dataframe().index
>           names=['realization', 'name', 'report_step', 'index'], length=500)

```,
it was already done in `local_ensemble.py` but was also needed for observations so it was moved to a separate file.

This can be tested on main, if you set a breakpoint on 
L1334 in `local_ensemble.py`:
```                
new_combined_ds = ensure_correct_coordinate_order(new_combined_ds)
```
then run
`tests/integration_tests/analysis/test_es_update.py test_that_posterior_has_lower_variance_than_prior`,

When it breakpoints, if you inspect in the python terminal:
```
new_combined_ds
```
it gives this:
```
<xarray.Dataset> Size: 4kB
Dimensions:      (name: 1, index: 10, realization: 50, report_step: 1)
Coordinates:
  * name         (name) object 8B 'POLY_RES'
  * index        (index) int32 40B 0 1 2 3 4 5 6 7 8 9
  * realization  (realization) int32 200B 1 2 3 4 5 6 7 ... 44 45 46 47 48 49 50
  * report_step  (report_step) int32 4B 0
Data variables:
    values       (realization, name, report_step, index) float64 4kB 4.324 .....
```

where `values` seemingly has the dimension ` (realization, name, report_step, index)`, this is true for xarray, but if you do a `.to_dataframe()`, you'd expect the index to remain the same, but doing it gives this:
```
                                           values
name     index realization report_step           
POLY_RES 0     1           0             4.324111
               2           0             0.212733
               3           0             4.134436
               4           0             2.696731
               5           0             0.390561
...                                           ...
         9     46          0            43.097676
               47          0            77.599153
               48          0            86.139757
               49          0            46.357514
               50          0            46.666562
[500 rows x 1 columns]
```
Now name, index is ahead of realization, with report_step at the end.

So that function is a workaround this jumbling up that happens for ...reasons. IF you do a `.[...dims]`, select from the xr dataset in exactly that way, the `.to_dataframe()` preserves the order, hence:
```
ensure_correct_coordinate_order(new_combined_ds).to_dataframe()

```
gives this
```
                                           values
realization name     report_step index           
1           POLY_RES 0           0       4.324111
                                 1       5.775136
                                 2       8.519636
                                 3      12.557611
                                 4      17.889061
...                                           ...
50          POLY_RES 0           5      16.626644
                                 6      22.691176
                                 7      29.719339
                                 8      37.711135
                                 9      46.666562
[500 rows x 1 columns]

```

On the PR, running `test_that_observation_getters_from_experiment_match_expected_data` , will fail if you remove the `ensure_correct_coordinate_order` around the returned single obs dataset. I think something with doing the selecting and dropping messes up the order

dafeda:
I updated to the latest `xarray`: `pip install xarray==2024.5.0`, set the breakpoint on line `1334` in `local_ensemble.py` and ran `test_that_posterior_has_lower_variance_than_prior()` as you suggested.

`to_dataframe()` seems now to retain the order:

image

dafeda:
> That was only for python3.8 where latest supported version is `2023.1.0`, for python3.11 we can use the newest xarray, but we still have to support both of them.

Thanks, I guess we just have to live with it.

dafeda (approved):

PR #8093, opened 2024-06-06
Revert "Enable Scheduler for LSF Queue by default"

xjules:
This reverts commit 2e48eeead5d321a50c5dd188e6fcd9bb3d4c511c.

**Issue**
After the several experiments, Scheduler is still WIP and thus we need to revert enabling it by default.


andreas-el (approved):
💯

PR #8094, opened 2024-06-06
Let LSFDriver pick num_cpu from the realization object

### LINKED ISSUE #8095: Python LSF driver does not pick the correct NUM_CPU
**Describe the bug**
The LSF driver should obtain the num_cpu from the function preferred_num_cpu() in order to support automatically obtaining it from the DATA file if not provided in the config. Currently it picks it directly from the config object, where it is assumed it to be possible to be empty or defaulted.

**Expected behaviour**
Keep behaviour from legacy LSF driver.

berland:
**Issue**
Resolves #8095 

**Approach**
Use the Realization object which is pre-populated with the correct cpu count.

## When applicable

### DIFF
--- a/src/ert/config/queue_config.py
+++ b/src/ert/config/queue_config.py
@@ -48,7 +48,6 @@ class QueueConfig:
     job_script: str = shutil.which("job_dispatch.py") or "job_dispatch.py"
     max_submit: int = 1
     submit_sleep: float = 0.0
-    num_cpu: int = 1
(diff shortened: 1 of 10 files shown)


xjules:
It makes sense. Just wondering about two things:
 - Can it realy change from realization to realization?
 - why are we sending it explicitly for the OpenPBS driver?

xjules (approved):
After the clarification I happily approve :smile: :rocket:

PR #8097, opened 2024-06-06
Remove sorting logic from get_observations_and_responses

yngve-sk:
Public API and all, caller is responsible for sorting its own data.


src/ert/storage/local_ensemble.py
```diff
@@ -1214,19 +1158,34 @@ def get_observations_and_responses(
                             f"={response_vals_per_real.shape[1]}"
                         )
 
-                    combined_np_long = np.concatenate(
+                    responses: Dict[str, npt.NDArray[np.float_]] = {
```
dafeda, line 1161: I think we may leverage `pandas` here and do something like this instead:

```python
responses_df = pd.DataFrame(response_vals_per_real, columns=[str(i) for i in range(response_vals_per_real.shape[1])])

combined_df_long = pd.concat([
    pd.DataFrame({
        "obs_name": obs_names_1d,
        "key_index": key_index_1d,
        "observations": obs_vals_1d,
        "errors": std_vals_1d,
    }),
    responses_df.astype('float32')
], axis=1)
```

Does this make sense you think?

src/ert/storage/local_ensemble.py
```diff
@@ -1235,10 +1194,9 @@ def get_observations_and_responses(
             raise KeyError(msg)
 
         # Ensure sorting by obs_name->key_index
-        long_np = np.concatenate(long_nps)
-        sorted_long_np = long_np[np.lexsort((long_np[:, 0], long_np[:, 1]))]
+        long_df = pd.concat(long_dfs).sort_values(by=["obs_name", "key_index"])
```
dafeda, line 1197: Tempting to set the index and verify integrity. Something like:

`long_df.set_index(["obs_name", "key_index"], verify_integrity=True)`

From `pandas` docs:

```
verify_integrity
bool, default False
Check the new index for duplicates. Otherwise defer the check until necessary. Setting to False will improve the performance of this method.
```

src/ert/analysis/_es_update.py
```diff
@@ -175,11 +175,17 @@ def _load_observations_and_responses(
         # Exit early if some observations are pointing to non-existing responses
         raise ErtAnalysisError("No active observations for update step") from e
 
-    responses = observations_and_responses.responses()
-    observations = observations_and_responses.observations()
-    errors = observations_and_responses.errors()
-    observation_keys = observations_and_responses.observation_keys()
-    indexes = observations_and_responses.index()
+    responses = observations_and_responses[
```
dafeda, line 178: If we return `observations_and_response` with an index set, we know that the first two columns are `observations` and `errors` and could get the rest by:

```python
observations_and_responses.iloc[:, 2:]
```

src/ert/analysis/_es_update.py
```diff
@@ -175,11 +175,17 @@ def _load_observations_and_responses(
...
+        [
+            i
+            for i in range(ensemble.ensemble_size + 1)
+            if i in observations_and_responses
+        ]
+    ].to_numpy()
+    observations = observations_and_responses["observations"].to_numpy()
+    errors = observations_and_responses["errors"].to_numpy()
+    observation_keys = observations_and_responses["obs_name"].to_numpy()
```
dafeda, line 187: Should we use observation `keys` or `names`?

yngve-sk: Unfortunately now we are currently using both, but we should def switch over to using only one. `*_name` convention is currently used in datasets, but we could use `*_key` (or `*_name`) uniformly for variable names

yngve-sk: Made an issue on it: https://github.com/equinor/ert/issues/8115

src/ert/analysis/_es_update.py
```diff
@@ -169,17 +170,38 @@ def _load_observations_and_responses(
...
     except KeyError as e:
         # Exit early if some observations are pointing to non-existing responses
         raise ErtAnalysisError("No active observations for update step") from e
 
-    responses = observations_and_responses.responses()
-    observations = observations_and_responses.observations()
-    errors = observations_and_responses.errors()
-    observation_keys = observations_and_responses.observation_keys()
-    indexes = observations_and_responses.index()
+    observations = observations_and_responses_df["observations"].to_numpy()
```
dafeda, line 198: I think we can get this from `observations_and_responses` and not the dataframe to avoid calling `to_numpy()` again.

src/ert/storage/local_ensemble.py
```diff
@@ -10,20 +10,21 @@
...
     Tuple,
+    TypedDict,
     Union,
 )
 from uuid import UUID
 
 import numpy as np
 import pandas as pd
 import xarray as xr
+from numpy._typing import NDArray
```
dafeda, line 27: Why use private `_typing` here?

src/ert/storage/local_ensemble.py
```diff
@@ -60,59 +61,12 @@ class _Failure(BaseModel):
...
-        return self._observations_and_responses[:, 2].astype(float)
-
-    def responses(self) -> npt.NDArray[np.float_]:
-        """
-        Extracts a ndarray with the shape (num_obs, num_reals).
-        Each cell holds the response value corresponding to the observation/realization
-        indicated by the index.
-        """
-        return self._observations_and_responses[:, 4:].astype(float)
+class ObservedResponses(TypedDict):
```
dafeda, line 64: What if we add a `to_dataframe()` method to this?

src/ert/analysis/_es_update.py
yngve-sk: I think we need to do this back&forth because of the sorting by `obs_name` and `key_index`

src/ert/storage/local_ensemble.py
yngve-sk: oops

src/ert/analysis/_es_update.py
```diff
@@ -160,51 +162,45 @@ def _load_observations_and_responses(
...
-        List[ObservationAndResponseSnapshot],
-    ],
+    pd.DataFrame,
+    List[ObservationAndResponseSnapshot],
 ]:
     try:
-        observations_and_responses = ensemble.get_observations_and_responses(
-            [*selected_observations], iens_active_index
+        observed_responses = ensemble.get_observations_and_responses(
+            list(selected_observations), iens_active_index
```
dafeda, line 170: Nitpicking, but is the explicit conversion to list necessary?

src/ert/analysis/_es_update.py
```diff
@@ -267,12 +271,16 @@ def _load_observations_and_responses(
...
         )
 
-    for missing_obs in observation_keys[~obs_mask]:
+    for missing_obs in observed_responses.obs_names[~obs_mask]:
         logger.warning(f"Deactivating observation: {missing_obs}")
 
-    return responses[obs_mask], (
-        observations[obs_mask],
-        scaled_errors[obs_mask],
+    # We sort the S-matrix by observation name to preserve
```
dafeda, line 277:
```suggestion
    # We sort the responses by observation name to preserve
```

src/ert/analysis/_es_update.py
yngve-sk: I think it can be removed, just have to change it to iterable where the arg goes further

PR #8104, opened 2024-06-07
Ensure WAITING state is set on resubmissions

### LINKED ISSUE #8079: Provide correct realization state when resubmitting the forward model
When the job fails and it gets resubmitted, the realization is still displayed as `RUNNING` in the GUI while the job needs to wait (can be hours) until the semaphore will allow the job to run.

berland:
Upon resubmissions, this state did not get set, as it before was only set in __init__, with the result that the GUI showed wrong state (RUNNING) before the realization was submitted.

**Issue**
Resolves #8079 

**Approach**
📯 

Screenshot shows that the realizations that failed in the first run are reset to Waiting (blue border):
image

## When applicable

### DIFF
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -89,6 +89,7 @@ class Job:
         return 0
 
     async def _submit_and_run_once(self, sem: asyncio.BoundedSemaphore) -> None:
+        await self._send(State.WAITING)
         await sem.acquire()
         timeout_task: Optional[asyncio.Task[None]] = None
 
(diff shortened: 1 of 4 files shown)


berland:
image

src/ert/scheduler/job.py
```diff
@@ -89,6 +89,7 @@ def running_duration(self) -> float:
         return 0
 
     async def _submit_and_run_once(self, sem: asyncio.BoundedSemaphore) -> None:
+        await self._send(State.WAITING)
```
xjules, line 92: Much better!

berland: fixed.

jonathan-eq (approved):
LGTM!

Issue #8105, opened 2024-06-07
Wrong state of realizations in Scheduler after terminate experiment


xjules:

JHolba:
this is probably not relevant any more after the changes to the ui

xjules:
> this is probably not relevant any more after the changes to the ui

Yes, closing this issue as haven't been observed since.

PR #8109, opened 2024-06-07
Make sure that scheduler send all events when jobs are finished

xjules:
**Issue**
Improves symptoms of https://github.com/equinor/ert/issues/8105

**Approach**
We include scheduler._events.join() and SENTINEL object to make
sure that the events left in the qeueu will be proccessed once the
jobs are done. To makes sure we won't wait for the queue and
SENTINEL indefinitely this introduces self._queue_timeout attribute.

## When applicable

### DIFF
--- a/src/ert/scheduler/scheduler.py
+++ b/src/ert/scheduler/scheduler.py
@@ -120,6 +120,8 @@ class Scheduler:
         self._ee_cert = ee_cert
         self._ee_token = ee_token
         self._publisher_done = asyncio.Event()
+        # this timeout makes sure we won't wait for the queue and the sentinel indefinitely
+        self._queue_timeout: float = 10.0
         self._consumer_started = asyncio.Event()
         self.checksum: Dict[str, Dict[str, Any]] = {}
         self.checksum_listener: Optional[asyncio.Task[None]] = None
@@ -271,8 +273,8 @@ class Scheduler:
                         event = await self._events.get()
                     if event == CLOSE_PUBLISHER_SENTINEL:
                         self._publisher_done.set()
-                        return
-                    await conn.send(event)
+                    else:
+                        await conn.send(event)
                     event = None
                     self._events.task_done()
             except ConnectionClosed:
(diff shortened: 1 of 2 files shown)


src/ert/scheduler/scheduler.py
```diff
@@ -255,8 +247,7 @@ async def _monitor_and_handle_tasks(
 
             if not self.is_active():
                 if self._ee_uri is not None:
-                    await self._events.put(CLOSE_PUBLISHER_SENTINEL)
-                    await self._publisher_done.wait()
+                    await self._events.join()
```
jonathan-eq, line 250: I think we had this or something similar earlier, and it caused us to end before all events had been processed (in this case, the queue would be empty because the final events had not been added yet)

xjules: `await self._events.join()` would make sure that all the events have been processed. Essentially, `CLOSE_PUBLISHER_SENTINEL` was added only here, which serves the same logic.

berland:
Squash the commits? The explanation under "Approach" should also be in the git commit message.

src/ert/scheduler/scheduler.py
berland: I share Jonathans worry, if there is something to be sent but it it is not sent yet, `self._events()` could be empty prematurely.

berland:
nitpick: First line in git commit message should be less than 50 chars, and the remaining lines should be formatted at 72 chars.

src/ert/scheduler/scheduler.py
```diff
@@ -270,9 +267,6 @@ async def _publisher(self) -> None:
                 while True:
                     if event is None:
                         event = await self._events.get()
-                    if event == CLOSE_PUBLISHER_SENTINEL:
-                        self._publisher_done.set()
```
berland, line 274: You are removing all usage of _publisher_done, so it should probably be removed from the class as well.

xjules: Thanks! I thought I did remove it. It might have got lost in the rebasing.

berland:
It might make sense to keep the publisher sentinel in place, to avoid rare corner cases where all events are "sent" but not yet added to the queue, and then it might exit prematurely.

Also, the commit message should be formatted to less than 50 chars on the first line, and then 72 chars for the body.

PR #8110, opened 2024-06-07
Show ert version/komodo version in GUI

yngve-sk:
Show komodo version and hostname if there is a `KOMODO_VERSION` env var
otherwise, show ert version from `ert.shared.__version__`

### DIFF
--- a/src/ert/gui/find_ert_info.py
+++ b/src/ert/gui/find_ert_info.py
@@ -1,0 +1,12 @@
+import os
+import socket
+
+from ert.shared import __version__
+
+
+def find_ert_info() -> str:
+    is_on_cluster = "KOMODO_RELEASE" in os.environ
+    if is_on_cluster:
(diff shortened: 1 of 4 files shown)


yngve-sk:
Screenshot 2024-06-10 at 07 39 07

On RGS with komodoenv

Screenshot 2024-06-10 at 07 56 59

src/ert/gui/find_ert_info.py
```diff
@@ -0,0 +1,12 @@
+import os
+import socket
+
+from ert.shared import __version__
+
+
+def find_ert_info():
```
andreas-el, line 7: `def find_ert_info() -> str:`

andreas-el:
Would it be nicer if removing the parentheses?

andreas-el (approved):
I support this change. ❤️

PR #8112, opened 2024-06-09
Improve LSFDriver.kill and logging for scheduler

xjules:
The goal of this backport is the improve flakyness of integration tests around LSFDriver.kill. Mainly, we fix a bug when the job is not running yet, but driver.kill wants to kill it.


xjules:
Wrong order of commits, closing this PR.

PR #8117, opened 2024-06-10
Type check ert.gui.plottery

eivindjahren:
There are a couple of things that are ignored:

1. Axis.x_label typehints says `str` and not `Optional[str]`, but the code says otherwise.
2. For some reason the ternary operator for calculating min is not caught correctly by mypy. Either that or there is some other reason why it thinks `current_min` can be `None` in the call to `min`.

## When applicable

### DIFF
--- a/src/ert/gui/plottery/plot_config.py
+++ b/src/ert/gui/plottery/plot_config.py
@@ -11,7 +11,13 @@ class PlotConfig:
     # The plot_settings input argument is an internalisation of the (quite few) plot
     # policy settings which can be set in the configuration file.
 
-    def __init__(self, plot_settings=None, title="Unnamed", x_label=None, y_label=None):
+    def __init__(
+        self,
+        plot_settings: Optional[dict[str, Any]] = None,
+        title: str = "Unnamed",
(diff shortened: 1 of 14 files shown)


src/ert/gui/plottery/plot_config.py
```diff
@@ -80,12 +86,13 @@ def __init__(self, plot_settings=None, title="Unnamed", x_label=None, y_label=No
...
-            self.nextColor()
+            return self.nextColor()
 
         return self._current_color
 
-    def nextColor(self):
-        self._current_color = next(self._line_color_cycle)
+    def nextColor(self) -> str:
+        color = next(self._line_color_cycle)
+        self._current_color = color
```
andreas-el, line 95: This seems unnecessary? Why split?

andreas-el (approved):
Nice addition! 🏅

src/ert/gui/plottery/plot_config.py
eivindjahren: `self._current_color` has type `Optional[str]` and not `str`. However, `next(self._line_color_cycle)` will have type `str`.

PR #8122, opened 2024-06-11
Have snapshot use typeddicts instead of dicts.

jonathan-eq:
**Approach**
✏️ 

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/snapshot.py
+++ b/src/ert/ensemble_evaluator/snapshot.py
@@ -90,25 +90,35 @@ class SnapshotMetadata(TypedDict, total=False):
     sorted_forward_model_ids: DefaultDict[RealId, List[FmStepId]]
 
 
-def _filter_nones(some_dict: Dict[str, Any]) -> Dict[str, Any]:
-    return {key: value for key, value in some_dict.items() if value is not None}
+class RealizationState(TypedDict, total=False):
+    active: bool
+    status: str
+    start_time: datetime.datetime
+    end_time: datetime.datetime
+
+
+class ForwardModelState(TypedDict, total=False):
+    status: str
+    start_time: datetime.datetime
+    end_time: datetime.datetime
+    index: str
+    current_memory_usage: str
+    max_memory_usage: str
+    name: str
+    error: str
+    stdout: str
+    stderr: str
 
 
 class PartialSnapshot:
     def __init__(self, snapshot: Optional["Snapshot"] = None) -> None:
-        self._realization_states: Dict[
-            str, Dict[str, Union[bool, datetime.datetime, str]]
-        ] = defaultdict(dict)
-        """A shallow dictionary of realization states. The key is a string with
-        realization number, pointing to a dict with keys active (bool),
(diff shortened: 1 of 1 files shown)


src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -150,15 +161,50 @@ def update_forward_model(
...
+                forward_model.stderr
+            )
         if self._snapshot:
-            self._snapshot._my_partial._forward_model_states[
-                (real_id, forward_model_id)
-            ].update(forward_model_update)
+            self._snapshot._my_partial.update_forward_model(
+                real_id, forward_model_id, forward_model
+            )
         return self
```
jonathan-eq, line 208: This can be removed if we remove the pydantic class for ForwardModel.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -150,15 +161,50 @@ def update_forward_model(
...
-        forward_model_update = _filter_nones(forward_model.model_dump())
-
-        self._forward_model_states[(real_id, forward_model_id)].update(
-            forward_model_update
-        )
-        if self._snapshot:
-            self._snapshot._my_partial._forward_model_states[
-                (real_id, forward_model_id)
-            ].update(forward_model_update)
+        if forward_model.status is not None:
```
berland, line 164: Can you try to put this in a loop over the typed-dicts contents?

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -399,7 +445,7 @@ def get_successful_realizations(self) -> typing.List[int]:
         return [
             int(real_idx)
             for real_idx, real_data in self._my_partial._realization_states.items()
-            if real_data[ids.STATUS] == state.REALIZATION_STATE_FINISHED
+            if real_data["status"] == state.REALIZATION_STATE_FINISHED
```
berland, line 448: Should we remove these from identifiers.py if we don't need them? (assuming we don't need them because the TypedDict would complain if we set some illegal dict key.)

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -499,14 +545,16 @@ def _from_nested_dict(data: Mapping[str, Any]) -> PartialSnapshot:
...
-        partial._realization_states[real_id] = _filter_nones(
-            {
-                "status": realization_data.get("status"),
-                "active": realization_data.get("active"),
-                "start_time": realization_data.get("start_time"),
-                "end_time": realization_data.get("end_time"),
-            }
-        )
+        partial._realization_states[real_id] = RealizationState()
+        if status := realization_data.get("status"):
```
berland, line 549: Maybe for loop here too.

jonathan-eq:
Will this PR be superseded by https://github.com/equinor/ert/pull/8182? @JHolba

src/ert/ensemble_evaluator/snapshot.py
JHolba: we can use them as long as they are marked Final. typechecker will complain if not. did that as part of #8182

PR #8132, opened 2024-06-12
Simplify PlotLimits

eivindjahren:
The PlotLimits class implements runtime typechecking in a very magical way. The type checking isn't really used for anything so it is more useful to implement it in a such a way that it can be type checked by mypy.

All functionality of the "undostack" setting the plotlimits in the custmization panel seems to still work as expected:

image

The DEPTH_AXIS seems to be unused, and index axis makes the corresponding customization field not accept "-" so the minimum is enforced by the text field:

image

## When applicable

### DIFF
--- a/src/ert/gui/plottery/plot_config.py
+++ b/src/ert/gui/plottery/plot_config.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import itertools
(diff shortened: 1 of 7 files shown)


andreas-el (approved):
❤️ TY

PR #8145, opened 2024-06-13
Add NPY to ruff rules with some numpy 2 support

larsevj:

## When applicable


eivindjahren:
We quite often use np.float64, so I am not sure if this is correctly typed.

eivindjahren:
If it should work with both 32-bit and 64-bit there is `np.floating`.

eivindjahren:
You can use `from __future__ import annotations` to have the types only be interpreted when mypy runs. mypy runs using python 3.12.

eivindjahren (approved):
:+1:

PR #8147, opened 2024-06-13
Fix signature of overrides from QAbstractItemModel

eivindjahren:
To satisfy LSP it should have exactly the same signature as QAbstractItemModel.

## When applicable

### DIFF
--- a/src/ert/gui/ertwidgets/models/storage_model.py
+++ b/src/ert/gui/ertwidgets/models/storage_model.py
@@ -173,7 +173,8 @@ class StorageModel(QAbstractItemModel):
     def columnCount(self, parent: QModelIndex = default_index) -> int:
         return _NUM_COLUMNS
 
-    def rowCount(self, parent: QModelIndex) -> int:
+    @override
+    def rowCount(self, parent: QModelIndex = default_index) -> int:
         if parent.isValid():
             if isinstance(parent.internalPointer(), RealizationModel):
                 return 0
--- a/src/ert/gui/ertwidgets/models/storage_model.py
+++ b/src/ert/gui/ertwidgets/models/storage_model.py
@@ -29,8 +29,6 @@ _COLUMN_TEXT = {
     2: "Type",
 }
 
-default_index = QModelIndex()
-
 
 class RealizationModel:
     def __init__(self, realization: int, parent: Any) -> None:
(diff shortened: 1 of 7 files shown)


src/ert/gui/ertwidgets/models/storage_model.py
```diff
@@ -27,6 +29,8 @@ class _Column(IntEnum):
     2: "Type",
 }
 
+default_index = QModelIndex()
```
andreas-el, line 32: Out of curiosity, is the reason for this the same as with dicts?

src/ert/gui/ertwidgets/models/storage_model.py
```diff
@@ -165,45 +169,59 @@ def _load_storage(self, storage: Storage) -> None:
...
 
-    def parent(self, index: QModelIndex) -> QModelIndex:
-        if not index.isValid():
+    @overload
+    def parent(self, child: QModelIndex) -> QModelIndex: ...
+    @overload
+    def parent(self) -> Optional[QObject]: ...
+    @override
+    def parent(self, child: QModelIndex = default_index) -> Optional[QObject]:
+        if not child.isValid():
```
andreas-el, line 191: Why does this function return Optional[QObject] ?
As far as I can see this returns a QModelIndex on all exits?

src/ert/gui/ertwidgets/models/storage_model.py
eivindjahren: So this explictly sets up the same thing as `parent=QModelIndex()`. There will only be one copy. Maybe we should just call it in the parameter? I remember there was a time where some linting rule said to just not do that.

eivindjahren: I will try to redo this in the `if parent is None:` way and see if mypy complains.

eivindjahren: Yea, this was a mistake, I should have used an Optional here as it is variant.

eivindjahren: Fixed it now.

andreas-el (approved):
Fine work. 👍

src/ert/gui/ertwidgets/models/storage_model.py
eivindjahren: That is true, but the type checker does not like that. I believe this has to do with overloads as a similar situation works correctly when there is no overload involved.

Overloads are somewhat of a special case. Here you would have to figure out that

```python
    def parent(self, child: Optional[QModelIndex] = None) -> QModelIndex:
```

satisfies the following interface:

```python
    @overload
    def parent(self, child: QModelIndex) -> QModelIndex: ...
    @overload
    def parent(self) -> Optional[QObject]: ...
```

which is quite complicated when you take into account all possibilities. mypy will not check that you always get a QModelIndex when the parameter `child` is supplied for instance, but it does check that the signature for the implementation accepts all the parameters in the overload defintions.

PR #8168, opened 2024-06-17
Backports to ert==10.1

xjules:
**Issue**
The main goal is to re-enable scheduler

**Approach**
In addition, it fixes flaky `test_kill_before_submit_is_finished` and decrease the amount of logs coming from websockets by setting to log level to info.


andreas-el (approved):
💯

PR #8175, opened 2024-06-18
Add PROJECT_CODE to LSF and OpenPBS drivers when submitting a job

### LINKED ISSUE #8048: Tag which forward models run on lsf
We would like to enable keeping statistics of which simulator runs on lsf. To do that, if a forward model job is of specific interest (say rms or flow) we would like to put it into the lsf project name (argument `-P`). This can be configured using the `QUEUE_OPTION` keyword like this:

```
QUEUE_OPTION LSF LSF_PROJECT_CODE rms+flow
```

so one way to implement this is to modify the user config in `ert_config.py`

```python
    tags = set(fm[0] for fm in config_dict.get("FORWARD_MODEL", []) if is_interesting(fm[0]))
    config_dict["QUEUE_OTPTION"].append(("LSF", "LSF_PROJECT_CODE", "+".join(tags))
```

DanSava:
**Issue**
Resolves #8048 

## When applicable

### DIFF
--- a/docs/reference/configuration/queue.rst
+++ b/docs/reference/configuration/queue.rst
@@ -194,11 +194,23 @@ The following is a list of available LSF configuration options:
 .. _project_code:
 .. topic:: PROJECT_CODE
 
+  String identifier used to map hardware resource usage to a project or account.
+  The project or account does not have to exist.
+
   Equates to the ``-P`` parameter for e.g. ``bsub``.
-  `See docs. <https://www.ibm.com/support/knowledgecenter/SSWRJV_10.1.0/lsf_command_ref/bsub.__p.1.html>`_
+  `See docs. <https://www.ibm.com/support/knowledgecenter/SSWRJV_10.1.0/lsf_command_ref/bsub.__p.1.html>`__
   For example, to register jobs in the ``foo`` project::
 
     QUEUE_OPTION LSF PROJECT_CODE foo
+
+  If the option is not set in the config file and the forward model section contains
+  any of the following simulator jobs [RMS, FLOW, ECLIPSE100, ECLIPSE300]
+  a default will be set.::
+
+    FORWARD_MODEL RMS <args>
+    FORWARD_MODEL ECLIPSE100 <args>
+
+  This will set the PROJECT_CODE option to ``rms+eclipse100``
 
 .. _exclude_host:
 .. topic:: EXCLUDE_HOST
@@ -355,6 +367,26 @@ The following is a list of all queue-specific configuration options:
 
     QUEUE_OPTION TORQUE QUEUE_QUERY_TIMEOUT 254
 
+.. _torque_project_code:
+.. topic:: PROJECT_CODE
+
+  String identifier used to map hardware resource usage to a project or account.
+  The project or account does not have to exist.
+
+  Equates to the ``-A`` parameter for``qsub``
+  `see docs. <https://www.jlab.org/hpc/PBS/qsub.html>`__
+  For example, to register jobs under the ``foo`` account::
+
+    QUEUE_OPTION TORQUE PROJECT_CODE foo
+
+  If the option is not set in the config file and the forward model section contains
+  any of the following simulator jobs [RMS, FLOW, ECLIPSE100, ECLIPSE300]
+  a default will be set.::
+
+    FORWARD_MODEL RMS <args>
+    FORWARD_MODEL ECLIPSE100 <args>
+
+  This will set the PROJECT_CODE option to ``rms+eclipse100``
 
 .. _slurm-systems:
 
@@ -467,6 +499,27 @@ only the most necessary options have been added.
   submitted to the queue when using the queue system, where ``n`` is a positive
   integer::
 
-    QUEUE_OPTION TORQUE MAX_RUNNING n
+    QUEUE_OPTION SLURM MAX_RUNNING n
 
   If ``n`` is zero (the default), then it is set to the number of realizations.
+
+.. _slurm_project_code:
+.. topic:: PROJECT_CODE
+
+  String identifier used to map hardware resource usage to a project or account.
+  The project or account does not have to exist.
+
+  Equates to the ``-A`` parameter for ``sbatch``
+  `see docs. <https://slurm.schedmd.com/sbatch.html#OPT_account>`__
+  For example, to register jobs under the ``foo`` account::
+
+    QUEUE_OPTION SLURM PROJECT_CODE foo
+
+  If the option is not set in the config file and the forward model section contains a
+  any of the following simulator jobs [RMS, FLOW, ECLIPSE100, ECLIPSE300]
+  a default will be set.::
+
+    FORWARD_MODEL RMS <args>
+    FORWARD_MODEL ECLIPSE100 <args>
+
+  This will set the PROJECT_CODE option to ``rms+eclipse100``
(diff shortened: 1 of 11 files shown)


berland:
Can you add the string "PROJECT_CODE" in some way in the commit message?

src/ert/config/queue_config.py
```diff
@@ -83,6 +83,20 @@ def from_dict(cls, config_dict: ConfigDict) -> QueueConfig:
             ):
                 submit_sleep = float(values[0])
 
+        def _is_simulator(fm_name: str):
```
berland, line 86: Why do we restrict the tagging to selected forward model names? That is not mentioned in the documentation for PROJECT_CODE.

DanSava: As far as I understood the idea is to tag job submissions to each of the drivers with a code that can be used to identify what resources are being used. For tracking purposes. 

I will add this information to the queue documentation.

src/ert/scheduler/__init__.py
```diff
@@ -20,53 +20,48 @@ def create_driver(config: QueueConfig) -> Driver:
     if config.queue_system == QueueSystem.LOCAL:
         return LocalDriver()
     elif config.queue_system == QueueSystem.TORQUE:
-        queue_config = {
-            key: value
-            for key, value in config.queue_options.get(QueueSystem.TORQUE, [])
-        }
-        num_nodes: Optional[str] = queue_config.get("NUM_NODES")
-        num_cpus_per_node: Optional[str] = queue_config.get("NUM_CPUS_PER_NODE")
+        queue_options = config.queue_options_as_dict
```
berland, line 23: Looks fine, but why do we want this specific change?

src/ert/config/queue_config.py
```diff
@@ -18,6 +18,7 @@
 )
 
 GENERIC_QUEUE_OPTIONS: List[str] = ["MAX_RUNNING"]
+PROJECT_CODE = ["PROJECT_CODE"]
```
berland, line 21: Why do we need a new category named the same as its only value?

src/ert/scheduler/__init__.py
DanSava: Just for better understanding and readability of the code. Currently we are creating dictionaries called `queue_config` and they only contain queue_options key-value pairs. I figured, maybe I was wrong here that it would be more understandable to call it queue_options given that is what they represent.

src/ert/config/queue_config.py
DanSava: At first, I wanted to add it to `GENERIC_QUEUE_OPTIONS` but I did not want it to be an option for `VALID_QUEUE_OPTIONS-QueueSystem.LOCAL` so this was what I came up with. This will probably be refactored as part of https://github.com/equinor/ert/issues/8201. As this requires a list of keys that will be valid for all drivers and this PROJECT_CODE falls under this category.

berland: I think it is cleaner to inject it inside each `*_DRIVER_OPTIONS`  then, or we should have a category for things valid for all drivers except the LocalDriver.

tests/unit_tests/config/test_queue_config.py
```diff
@@ -30,6 +30,36 @@ def test_stop_long_running_is_set_from_corresponding_keyword(value):
     assert QueueConfig(stop_long_running=value).stop_long_running == value
 
 
+@pytest.mark.parametrize("queue_system", ["LSF", "TORQUE", "SLURM"])
+def test_project_code_is_set_when_forward_model_contains_simulator(queue_system):
```
berland, line 34: `.._contains_selected_simulators` ?? (since "simulator" is not a well-defined term)

src/ert/scheduler/__init__.py
berland: ah, I think I misread the diff..

docs/reference/configuration/queue.rst
```diff
@@ -355,6 +363,22 @@ The following is a list of all queue-specific configuration options:
...
+  `see docs. <https://www.jlab.org/hpc/PBS/qsub.html>`__
+  For example, to register jobs under the ``foo`` account::
+
+    QUEUE_OPTION TORQUE PROJECT_CODE foo
+
+  If the option is not set in the config file and the forward model section contains a
+  simulator job a default will be set.::
+
+    FORWARD_MODEL RMS <args>
+    FORWARD_MODEL FLOW <args>
```
berland, line 379: ECLIPSE is missing here.

docs/reference/configuration/queue.rst
```diff
@@ -355,6 +363,22 @@ The following is a list of all queue-specific configuration options:
...
+.. topic:: PROJECT_CODE
+
+  Equates to the ``-A`` parameter for``qsub``
+  `see docs. <https://www.jlab.org/hpc/PBS/qsub.html>`__
+  For example, to register jobs under the ``foo`` account::
+
+    QUEUE_OPTION TORQUE PROJECT_CODE foo
+
+  If the option is not set in the config file and the forward model section contains a
+  simulator job a default will be set.::
```
berland, line 376: "contains any of the following simulator jobs"

src/ert/config/queue_config.py
```diff
@@ -112,6 +112,28 @@ def from_dict(cls, config_dict: ConfigDict) -> QueueConfig:
             ):
                 submit_sleep = float(values[0])
 
+        def _is_simulator(fm_name: str):
+            return any(item in fm_name.lower() for item in ["rms", "flow", "eclipse"])
```
berland, line 116: The forward models for Eclipse are usually called ECLIPSE100 and ECLIPSE300. Do we want to do `startswith` instead?

berland: It might be the case for many FLOW runs as well, that they only start with FLOW, but then it is probably a user-defined fm, f.ekx. "FLOW_DAILY".

DanSava: You are right this is a bit ambiguous, it might be better to have a list of known simulators and match on the exact name ["RMS", "FLOW", "ECLIPSE100", "ECLIPSE300"] this way we don't risk adding tags that for jobs that are simulator related

docs/reference/configuration/queue.rst
```diff
@@ -195,11 +195,21 @@ The following is a list of available LSF configuration options:
...
   For example, to register jobs in the ``foo`` project::
 
     QUEUE_OPTION LSF PROJECT_CODE foo
 
+  If the option is not set in the config file and the forward model section contains
+  any of the following simulator jobs [RMS, FLOW, ECLIPSE100, ECLIPSE300]
+  a default will be set.::
+
+    FORWARD_MODEL RMS <args>
+    FORWARD_MODEL FLOW <args>
```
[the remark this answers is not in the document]
berland: Or maybe I effectively said the other thing in my last review 🤔 . But try to have both precise information and relevant example.

docs/reference/configuration/queue.rst
```diff
@@ -355,6 +364,23 @@ The following is a list of all queue-specific configuration options:
 
     QUEUE_OPTION TORQUE QUEUE_QUERY_TIMEOUT 254
 
+.. _torque_project_code:
+.. topic:: PROJECT_CODE
+
```
andreas-el, line 369: This is not your doing, but the `PROJECT_CODE` topic does not have an initial description like other topics have.
Does it make sense to add something simplistic to indicate what `PROJECT_CODE` implies, rather than immediately refer to an external link to man pages for qsub/bsub etc?
```
String identifier used to map hardware resource usage to project and/or account
```

andreas-el (approved):
Solid improvement! 🚀

docs/reference/configuration/queue.rst
DanSava: Sounds like a good thing to add.

PR #8192, opened 2024-06-20
Ert fm plugin docs without port

yngve-sk:
Lets us get in semeio docs while awaiting resolution on the how to add default plugins to ert discussion


eivindjahren (approved):
:+1:

PR #8203, opened 2024-06-21
Move driver option names to queue_config

### LINKED ISSUE #8200: Don't get driver specific config options from the c-drivers
The lines

https://github.com/equinor/ert/blob/761b07012717abe6d567cd9cb7d75130f03b122c/src/ert/config/queue_config.py#L41-L42

must be changed into fixed lists as it is done for the OPENPBS_DRIVER_OPTIONS in the lines above.

These lists of strings do not belong in the driver, as the driver should not define how to the ErtConfig should look like, they should only expose a Python API for the values of these options.

berland:
The drivers should not be responsible for defining how the drivers should be configured in the config system, they should only expose an API

**Issue**
Resolves #8200 

**Approach**
🚚 

## When applicable


xjules (approved):
Makes sense! :rocket:

PR #8212, opened 2024-06-21
Convert dict access of ForwardModel to .get() to avoid segfaults

JHolba:
All fields in ForwardModel are optional. Trying to access a field that is not set while result in errors.

**Approach**
Use .get on forwardmodel objects to avoid errors

## When applicable

### DIFF
--- a/src/ert/gui/model/snapshot.py
+++ b/src/ert/gui/model/snapshot.py
@@ -469,7 +469,7 @@ class SnapshotModel(QAbstractItemModel):
             data_name = COLUMNS[NodeType.REAL][index.column()]
             if data_name in [ids.CURRENT_MEMORY_USAGE, ids.MAX_MEMORY_USAGE]:
                 data = node.data
-                _bytes: Optional[str] = data[data_name]
+                _bytes: Optional[str] = data.get(data_name)
                 if _bytes:
                     return byte_with_unit(float(_bytes))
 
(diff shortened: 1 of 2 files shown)


src/ert/gui/model/snapshot.py
```diff
@@ -221,12 +221,17 @@ def _add_partial_snapshot(self, partial: PartialSnapshot, iter_: int) -> None:
...
+                    and "current_memory_usage" in job
                     and job["current_memory_usage"] is not None
                 ):
                     cur_mem_usage = int(float(job["current_memory_usage"]))
                     real_node.data.current_memory_usage = cur_mem_usage
                     self.root.data.current_memory_usage = cur_mem_usage
-                if "max_memory_usage" in job and job["max_memory_usage"] is not None:
+                if (
+                    "max_memory_usage" in job
+                    and "max_memory_usage" in job
```
berland, line 232: ???

berland: This test is not possible to simplify with get?

JHolba: it is possible to use get safely, but the typechecker does not understand it.

src/ert/gui/model/snapshot.py
```diff
@@ -221,12 +221,17 @@ def _add_partial_snapshot(self, partial: PartialSnapshot, iter_: int) -> None:
                 job_node.data.update(job)
                 if (
                     "current_memory_usage" in job
+                    and "current_memory_usage" in job
```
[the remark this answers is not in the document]
JHolba: it was a brainfart when I double checked that I got all the accesses

berland (approved):

PR #8215, opened 2024-06-24
Add a new configuration keyword for memory pr realization

### LINKED ISSUE #7798: Implement new config key MEMORY_PR_(JOB|FORWARD_MODEL|REALIZATION)
We should propagated specifying the MEMORY requirement to the main ERT config level and thus having this option as a new config key.

berland:
**Issue**
Resolves #7798

**Approach**
Adding a new keyword that is propagated to via `QueueConfig` to `driver.submit()`. Overlaps with Torque queue option MEMORY_PER_JOB which should be deprecated if this PR goes through.

## When applicable


berland:
todo: 
* catch overspecification in config system

berland:
> * catch overspecification in config system

Fixed

> * solve conflict with the auto-inserted memory usage from ert-configurations

Fixed by doing silent replacement for the LSF driver when `mem=` is already present in the resource requirement string.

src/ert/config/parsing/config_schema_deprecations.py
```diff
@@ -224,4 +224,16 @@
         "the future. Replace by NUM_CPU on a single compute node.",
         check=lambda line: "NUM_NODES" in line,
     ),
+    DeprecationInfo(
+        keyword="QUEUE_OPTION",
+        message="MEMORY_PER_JOB as QUEUE_OPTION to Torque is deprecated and will removed in "
```
jonathan-eq, line 229: I see some of the other deprecation warnings output the queue system in all uppercase. I think that looks better as QUEUE_OPTION and the queue option itself are uppercase too.

src/ert/ensemble_evaluator/_ensemble.py
```diff
@@ -393,3 +393,4 @@ class Realization:
     run_arg: "RunArg"
     num_cpu: int
     job_script: str
+    realization_memory: int  # Memory to reserve/book, in bytes
```
jonathan-eq, line 396: Can we write this in a way so that we can drop the comment? Maybe we can include the unit in the property name?`reserved_realization_memory_in_bytes` maybe?

src/ert/scheduler/driver.py
```diff
@@ -46,6 +47,7 @@ async def submit(
           cwd: Working directory.
           name: Name of job as submitted to compute cluster
           num_cpu: Number of CPU-cores to allocate
+          realization_memory: Memory to reserve, in bytes. 0 means no allocation.
```
jonathan-eq, line 50: This can potentially be misinterpreted as `no memory will be allocated`. Can we change this to indicate that the default will be used if nothing is specified?

tests/integration_tests/scheduler/test_lsf_driver.py
```diff
@@ -201,6 +201,38 @@ async def test_submit_with_num_cpu(pytestconfig, job_name):
     assert Path("test").read_text(encoding="utf-8") == "test\n"
 
 
+@pytest.mark.usefixtures("use_tmpdir")
+async def test_submit_with_realization_memory(pytestconfig, job_name):
+    if not pytestconfig.getoption("lsf"):
+        # Mocked LSF driver does not provide bhist.
+        return
```
jonathan-eq, line 208: Should we return here, or do ´pytest.skip()`?

src/ert/config/parsing/config_schema_deprecations.py
berland: Fixed (but only for MEMORY*)

src/ert/ensemble_evaluator/_ensemble.py
berland: I would claim that bytes is the canonical unit when the type is `int`, so the comment about `bytes` can strictly be dropped, but not sure if it makes it better. Since bytes is the canonical unit, it should not be necessary to have it in the variable names, so that puts me into a spagat :)

src/ert/scheduler/driver.py
berland: Yes, "reserve" is a wrong word here. Fixing.

tests/unit_tests/config/test_queue_config.py
```diff
@@ -83,22 +83,130 @@ def test_that_invalid_queue_option_raises_validation_error(
...
+        ("10kb", 10 * 1024),
+        ("10mb", 10 * 1024**2),
+        ("10gb", 10 * 1024**3),
+        ("10Mb", 10 * 1024**2),
+        ("10Gb", 10 * 1024**3),
+        ("10Tb", 10 * 1024**4),
+        ("10Pb", 10 * 1024**5),
+    ],
+)
+def test_realization_memory_unit_support(memory_spec: str, expected_bytes):
```
jonathan-eq, line 120: Should we use the tmp_dir fixture here, or is it in autouse?

tests/unit_tests/config/test_queue_config.py
```diff
@@ -83,22 +83,130 @@ def test_that_invalid_queue_option_raises_validation_error(
...
+        f.write("QUEUE_SYSTEM TORQUE\n")
+        f.write("QUEUE_OPTION TORQUE MEMORY_PER_JOB 20m\n")
+
+    with pytest.raises(ConfigValidationError), pytest.warns(
+        ConfigWarning, match="deprecated"
+    ):
+        ErtConfig.from_file(filename)
+
+
+def test_conflicting_realization_openpbs_memory_per_job_but_slurm_activated_only_warns():
```
jonathan-eq, line 182: Should we warn when using a different queue system?

tests/unit_tests/scheduler/test_job.py
```diff
@@ -180,6 +182,28 @@ async def test_num_cpu_is_propagated_to_driver(realization: Realization):
...
+    job_run_task = asyncio.create_task(job.run(asyncio.Semaphore(), max_submit=1))
+    job.started.set()
+    job.returncode.set_result(0)
+    await job_run_task
+    scheduler.driver.submit.assert_called_with(
+        realization.iens,
+        realization.job_script,
+        realization.run_arg.runpath,
+        num_cpu=1,
+        realization_memory=8 * 1024**2,
```
jonathan-eq, line 204: Can we put the magic number `8*1024**2` in a variable and use it on line 192 too?

tests/unit_tests/scheduler/test_slurm_driver.py
```diff
@@ -85,6 +85,26 @@ async def test_memory_is_set(memory_in_mb):
     )
 
 
+@pytest.mark.usefixtures("capturing_sbatch")
+@given(memory_in_bytes=st.integers(min_value=1))
+async def test_realization_memoryt(memory_in_bytes):
```
jonathan-eq, line 90: Typo `...realization_memoryt`

tests/unit_tests/scheduler/test_slurm_driver.py
```diff
@@ -85,6 +85,26 @@ async def test_memory_is_set(memory_in_mb):
...
+    driver = SlurmDriver(realization_memory=memory_in_bytes)
+    await driver.submit(0, "sleep", name="myjobname")
+    assert f"--mem={memory_in_bytes // 1024**2}M" in Path(
+        "captured_sbatch_args"
+    ).read_text(encoding="utf-8")
+
+
+async def test_overspecified_memory_allocation():
+    # Emptry string for memory is equivalent to zero integer for realiation_memory
```
jonathan-eq, line 99: Typo `Emptry`

tests/unit_tests/scheduler/test_slurm_driver.py
```diff
@@ -85,6 +85,26 @@ async def test_memory_is_set(memory_in_mb):
...
+    SlurmDriver(memory="", realization_memory=1)
+    SlurmDriver(memory="1mb", realization_memory=0)
+    with pytest.raises(ValueError):
+        SlurmDriver(memory="1mb", realization_memory=1)
+    with pytest.raises(ValueError):
+        SlurmDriver(memory="0", realization_memory=1)
```
jonathan-eq, line 105: Should this be input as pytest parameters instead? Here we are testing 4 things at the same time

jonathan-eq (changes requested):
Very good change :) 👍

tests/unit_tests/config/test_queue_config.py
berland: Do you mean to warn when Torque options are specified but Slurm is active? I would say no to that, giving users the opportunity to be flexible with which queue system a config file uses without Ert being too noisy.

tests/unit_tests/config/test_queue_config.py
berland: We should. Fixed.

tests/unit_tests/scheduler/test_job.py
berland: fixed.

tests/unit_tests/scheduler/test_slurm_driver.py
berland: fixed.

tests/unit_tests/scheduler/test_slurm_driver.py
berland: fixed!

tests/unit_tests/config/test_queue_config.py
jonathan-eq: Kinda. This test tests that using the deprecated torque queue option `MEMORY_PER_JOB` warns even when not running on torque. Is this something we want?

berland: ah. 🤔 I think we want to keep this warning, as there is no reason to use MEMORY_PER_JOB even if you don't use Torque. Then you should forward that memory information to your other queue system.

src/ert/config/queue_config.py
```diff
@@ -201,6 +210,30 @@ def _check_num_cpu_requirement(
...
+
+    if realization_memory_str.isdigit():
+        return int(realization_memory_str)
+    multipliers = {
+        "b": 1,
+        "k": 1024,
+        "m": 1024**2,
+        "g": 1024**3,
+        "t": 1024**4,
+        "p": 1024**5,
```
xjules, line 227: :smile: in the far far future `t` and `p` might be hopefully used

tests/unit_tests/scheduler/test_slurm_driver.py
berland: Fixed!

berland:
Big kudos to code coverage tool, it pointed to one uncovered line in lsf_driver, and that line was actually buggy!

src/ert/config/parsing/config_schema_deprecations.py
```diff
@@ -224,4 +224,16 @@
...
+        keyword="QUEUE_OPTION",
+        message="MEMORY_PER_JOB as QUEUE_OPTION to TORQUE is deprecated and will be removed in "
+        "the future. Replace by REALIZATION_MEMORY.",
+        check=lambda line: "MEMORY_PER_JOB" in line,
+    ),
+    DeprecationInfo(
+        keyword="QUEUE_OPTION",
+        message="MEMORY as QUEUE_OPTION to SLURM is deprecated and will be removed in "
+        "the future. Replace by REALIZATION_MEMORY.",
+        check=lambda line: "MEMORY" in line,
```
jonathan-eq, line 237: Will this also issue a deprecation warning for the `REALIZATION_MEMORY` keyword?

src/ert/config/queue_config.py
```diff
@@ -147,9 +147,30 @@
...
+                mem_str_format,
+            )
+            is not None
+        )
+
+
+queue_memory_usage_formats: Mapping[str, QueueMemoryStringFormat] = {
+    "SLURM": QueueMemoryStringFormat(suffixes=["", "K", "M", "G", "T"]),
+    "TORQUE": QueueMemoryStringFormat(suffixes=["kb", "mb", "gb", "KB", "MB", "GB"]),
+}
```
jonathan-eq, line 167: Are the queue system formats documented somewhere in the ert docs, or should the user check with the queue system docs directly?

src/ert/config/queue_config.py
```diff
@@ -342,14 +377,24 @@ def throw_error_or_warning(
 
 
 def _validate_queue_driver_settings(
-    queue_system_options: List[Tuple[str, str]], queue_type: str, throw_error: bool
+    queue_system_options: List[Tuple[str, str]],
+    queue_type: str,
+    throw_error: bool,
+    realization_memory: int,
```
jonathan-eq, line 383: Do we need `realization_memory` as a separate parameter, or can we only input queue_system_options and parse it here instead/also?

src/ert/scheduler/lsf_driver.py
```diff
@@ -127,14 +127,55 @@ def parse_bjobs(bjobs_output: str) -> Dict[str, JobState]:
...
+    realization_memory argument will override.
+
+    Args:
+        exclude_hosts: List of hostnames to exclude
+        realization_memory: Memory amount to reserve in bytes
+
+    Returns:
+        Resource specification string to be added to -R option to bsub.
+
+    """
```
jonathan-eq, line 149: Very good!

src/ert/scheduler/lsf_driver.py
```diff
@@ -127,14 +127,55 @@ def parse_bjobs(bjobs_output: str) -> Dict[str, JobState]:
 
 
 def build_resource_requirement_string(
-    exclude_hosts: Sequence[str], resource_requirement: Optional[str]
+    exclude_hosts: Sequence[str],
+    realization_memory: int,
+    resource_requirement: str = "",
```
jonathan-eq, line 132: Should this be `Optional[str]` as we are providing a default value?

tests/unit_tests/scheduler/test_slurm_driver.py
jonathan-eq: I am not sure if fix was pushed

src/ert/config/parsing/config_schema_deprecations.py
berland: No, you would have to write `QUEUE_OPTION * REALIZATION_MEMORY 50mb` for this to kick in, but that is in any case invalid.

jonathan-eq: Right, it is not a queue_option but a general keyword.

src/ert/config/queue_config.py
berland: The existing Torque and Slurm docs for Ert mention units sufficiently yes. At least sufficiently for deprecated parameters. 

The new docs for REALIZATION_MEMORY uses `Gb` in its example. I think that is enough for any relevant usage of the keyword.

src/ert/config/queue_config.py
berland: I think it is okay for this private function, as realization_memory is not part of the tuple of strings constituting the queue_system_options. But I will reorder the arguments, making realization_memory the second arg.

Maybe we can do something about it later when refactoring queue_system_options from a list of tuples to e.g. a dataclass.

src/ert/scheduler/lsf_driver.py
berland: Optional implies that we allow `None` as a value, which we don't here (if you supply `None`, it will crash).

jonathan-eq (approved):
LGTM!

PR #8225, opened 2024-06-24
Add blue color and underscore to STDOUT and STDERR in job view

### LINKED ISSUE #4371: Make OPEN buttons (stdout/stderror for jobs in detailed progress) more obviously clickable
The `OPEN` "buttons" under `stdout` and `stderror` are clickable and will open the output from those files. This is not obvious for the user. It has been tried to make those into buttons before, but performance was an issue. We must consider that some users can have 300 jobs and some users run 10 000 realizations.

Making them underlined or adding an icon may be viable options

Screenshot 2022-11-23 at 10 32 11

frode-aarstad:
**Issue**
Resolves #4371 

**Approach**
Adding qt buttons to the job list is non trivial and would require a larger rewrite. This solution is simpler while adding some feedback to the user that these columns are clickable.

image

## When applicable

### DIFF
--- a/src/ert/gui/model/snapshot.py
+++ b/src/ert/gui/model/snapshot.py
@@ -347,12 +347,14 @@ class SnapshotModel(QAbstractItemModel):
         return self.createIndex(parent_item.row(), 0, parent_item)
 
     @override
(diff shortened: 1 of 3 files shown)


dafeda (approved):
Looks good 👍

Issue #1871, opened 2021-07-14
Remove columns from models that are not displayed in a table


jondequinor:
**Describe the bug**
A relic from qt-refactoring in the GUI is that some levels in the snapshot tree model, have colums. E.g. realizations have three columns, iter has two, etc.

These levels are not shown in a tabular fashion. Reals are shown as a single item list (with various [roles](https://doc.qt.io/qt-5/model-view-programming.html#item-roles)), iterations are represented by a [`QTabWidget`](https://doc.qt.io/qt-5/qtabwidget.html). Only jobs are displayed in a table.

The downside is that columns cannot be assumed to be 1, which is easier, more accurate, and has the potential for a performance increase.

PR #8230, opened 2024-06-25
Simplify Snapshot model

andreas-el:
Resolves https://github.com/equinor/ert/issues/1871

**Issue**

**Approach**

## When applicable

### DIFF
--- a/src/ert/gui/model/node.py
+++ b/src/ert/gui/model/node.py
@@ -49,7 +49,6 @@ class _Node(ABC):
 
 @dataclass
 class RootNodeData:
-    current_memory_usage: Optional[int] = None
     max_memory_usage: Optional[int] = None
 
 
@@ -74,8 +73,6 @@ class IterNodeData:
     sorted_forward_model_step_ids_by_realization_id: dict[str, list[str]] = field(
         default_factory=dict
     )
-    current_memory_usage: Optional[int] = None
-    max_memory_usage: Optional[int] = None
 
 
 @dataclass
--- a/src/ert/gui/model/node.py
+++ b/src/ert/gui/model/node.py
@@ -61,7 +61,6 @@ class RootNode(_Node):
 class IterNodeData:
     index: Optional[str] = None
     status: Optional[str] = None
-    sorted_realization_ids: list[str] = field(default_factory=list)
 
 
 @dataclass
--- a/src/ert/gui/model/node.py
+++ b/src/ert/gui/model/node.py
@@ -40,15 +40,10 @@ class _Node(ABC):
 
 
 @dataclass
-class RootNodeData:
-    max_memory_usage: Optional[int] = None
-
-
-@dataclass
 class RootNode(_Node):
     parent: None = field(default=None, init=False)
     children: dict[int, IterNode] = field(default_factory=dict)
-    data: RootNodeData = field(default_factory=RootNodeData)
+    max_memory_usage: Optional[int] = None
 
     def add_child(self, node: IterNode, node_id: Optional[int] = None) -> None:
         node.parent = self
--- a/src/ert/gui/model/node.py
+++ b/src/ert/gui/model/node.py
@@ -26,7 +26,6 @@ class _Node(ABC):
     def add_child(
         self,
         node: IterNode | RealNode | ForwardModelStepNode,
-        node_id: Optional[int] = None,
     ) -> None:
         pass
 
@@ -42,11 +41,9 @@ class RootNode(_Node):
     children: dict[int, IterNode] = field(default_factory=dict)
     max_memory_usage: Optional[int] = None
 
-    def add_child(self, node: IterNode, node_id: Optional[int] = None) -> None:
+    def add_child(self, node: IterNode) -> None:
         node.parent = self
-        if node_id is None:
-            node_id = node.id_
-        self.children[node_id] = node
+        self.children[node.id_] = node
 
 
 @dataclass
@@ -61,16 +58,13 @@ class IterNode(_Node):
     data: IterNodeData = field(default_factory=IterNodeData)
     children: dict[str, RealNode] = field(default_factory=dict)
 
-    def add_child(self, node: RealNode, node_id: Optional[int] = None) -> None:
+    def add_child(self, node: RealNode) -> None:
         node.parent = self
-        if node_id is None:
-            node_id = node.id_
-        self.children[str(node_id)] = node
+        self.children[str(node.id_)] = node
 
 
 @dataclass
 class RealNodeData:
-    index: Optional[str] = None
     status: Optional[str] = None
     active: Optional[bool] = False
     forward_model_step_status_color_by_id: dict[str, QColor] = field(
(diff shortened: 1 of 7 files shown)


andreas-el:
I think the commits you've added make sense. Let's keep them in.

src/ert/gui/model/node.py
```diff
@@ -2,105 +2,68 @@
...
         children = "no " if len(self.children) == 0 else f"{len(self.children)} "
         return f"Node<{type(self).__name__}>@{self.id_} with {parent}parent and {children}children"
 
     @abstractmethod
-    def add_child(
-        self,
-        node: IterNode | RealNode | ForwardModelStepNode,
-        node_id: Optional[int] = None,
-    ) -> None:
+    def add_child(self, node: _Node) -> None:
```
jonathan-eq, line 26: Is it worth changing the node type from `IterNode | RealNode | ForwardModelStepNode` to `_Node`, if we have to cast the types in every method implementation?

src/ert/gui/model/snapshot.py
```diff
@@ -43,21 +42,17 @@
...
-}
+JOB_COLUMNS: Sequence[str] = [
+    ids.NAME,
+    ids.ERROR,
+    ids.STATUS,
+    DURATION,  # Duration is based on two data fields, not coming directly from ert
+    ids.STDOUT,
+    ids.STDERR,
+    ids.CURRENT_MEMORY_USAGE,
+    ids.MAX_MEMORY_USAGE,
```
jonathan-eq, line 53: Is it possible to remove all the constants from identifiers (ids) and have them here instead?

src/ert/gui/model/snapshot.py
```diff
@@ -177,17 +172,8 @@ def _add_partial_snapshot(self, partial: PartialSnapshot, iter_: int) -> None:
...
-                iter_node.row(), iter_index.column(), QModelIndex()
-            )
-            stack.callback(self.dataChanged.emit, iter_index, iter_index_bottom_right)
-
             reals_changed: List[int] = []
 
-            for idx in job_infos:
-                real_id = idx[0]
-                if real_id not in reals:
-                    reals[real_id] = RealizationSnapshot()
```
jonathan-eq, line 190: Was this loop causing us to count every forward model twice? Once from snapshot.get_all_forward_models() and second from snapshot.reals?

src/ert/gui/model/node.py
JHolba: it will be more or less the same. the signature will be incorrect either way. each node type only accepts one type of child, but you should not change the signature of an overridden method.

src/ert/gui/model/snapshot.py
```diff
@@ -256,19 +242,15 @@ def _add_partial_snapshot(self, partial: PartialSnapshot, iter_: int) -> None:
...
             data=IterNodeData(
                 status=snapshot.status,
-                sorted_realization_ids=metadata["sorted_real_ids"],
-                sorted_forward_model_step_ids_by_realization_id=metadata[
-                    "sorted_forward_model_ids"
-                ],
             ),
         )
-        for real_id in snapshot_tree.data.sorted_realization_ids:
+        for real_id in metadata["sorted_real_ids"]:
```
jonathan-eq, line 253: Good change, helps readability :)

src/ert/gui/model/snapshot.py
```diff
@@ -300,26 +282,14 @@ def _add_snapshot(self, snapshot: Snapshot, iter_: int) -> None:
...
-            elif isinstance(parent_node, IterNode):
-                count = len(COLUMNS[NodeType.ITER])
-            elif isinstance(parent_node, RealNode):
-                count = len(COLUMNS[NodeType.REAL])
-            else:
-                count = len(COLUMNS[NodeType.JOB])
-        return count
+        if parent and isinstance(parent.internalPointer(), RealNode):
+            return JOB_COLUMN_SIZE
+        return 1
```
jonathan-eq, line 292: Why do we return 1?

src/ert/gui/model/snapshot.py
```diff
@@ -438,22 +408,22 @@ def _job_data(
         node_id = str(node.id_)
 
         if role == Qt.ItemDataRole.FontRole:
-            data_name = COLUMNS[NodeType.REAL][index.column()]
+            data_name = JOB_COLUMNS[index.column()]
```
jonathan-eq, line 411: This is beautiful 👌

src/ert/gui/simulation/run_dialog.py
```diff
@@ -403,15 +408,15 @@ def _on_event(self, event: object) -> None:
             self._show_done_button()
         elif isinstance(event, FullSnapshotEvent):
             if event.snapshot is not None:
-                self._snapshot_model._add_snapshot(event.snapshot, event.iteration)
+                self._snapshot_model._add_snapshot(event.snapshot, str(event.iteration))
```
jonathan-eq, line 411: Is it possible to change the event type to have iteration as a string instead of int, so we dont have to parse it every time? Would it break something somewhere else?

src/ert/gui/model/snapshot.py
andreas-el: The indexing was made to fit arrays/tables, but this isn't a table, so specifying something other than 1 will cause either objects not to be drawn, or returning too many columns.

tests/unit_tests/gui/model/test_job_list.py
```diff
@@ -14,17 +14,16 @@
...
 from .gui_models_utils import partial_snapshot
 
 
 def _id_to_col(identifier):
-    for col, field in enumerate(COLUMNS[NodeType.REAL]):
+    for col, field in enumerate(JOB_COLUMNS):
         if field == identifier:
             return col
-    raise ValueError(f"{identifier} not a column in {COLUMNS}")
+    raise ValueError(f"{identifier} not a column in {JOB_COLUMNS}")
```
jonathan-eq, line 26:
```suggestion
return JOB_COLUMNS.index(identifier)
```

src/ert/gui/simulation/run_dialog.py
andreas-el: I've proposed to revert the last two commits, and rather have these as a separate PR.
We are snowballing atm, and this type change str vs int makes sense to resolve, just not in this one. Personal opinion.

jonathan-eq (changes requested):
Very good improvement! 
See minor changes 👌

src/ert/gui/simulation/run_dialog.py
jonathan-eq: Agreed

andreas-el: They have now been removed, just mentioning this so no misunderstanding.

src/ert/gui/model/snapshot.py
andreas-el: It is a possibility. I see no use in having this currently, so we decided to remove this code snippet.

src/ert/gui/model/snapshot.py
andreas-el: Not easily no. A lot of these identifiers are used elsewhere. I suspect the net gain is negative from such an impact.

jonathan-eq (approved):
LGTM 🚢 !
Just remember to squash the commits before merging 👌

PR #8234, opened 2024-06-25
Fix restart for ensemble experiment

### LINKED ISSUE #8032: `Restart` experiment (or realization) from GUI does unexpected things
**Describe the bug**

**To reproduce**
1. `pip install ert`
2. Modify `poly_eval.py` such that some realizations fail:

```python
#!/usr/bin/env python3
import json
import random

def _load_coeffs(filename):
    with open(filename, encoding="utf-8") as f:
        return json.load(f)["COEFFS"]


def _evaluate(coeffs, x):
    return coeffs["a"] * x**2 + coeffs["b"] * x + coeffs["c"]


if __name__ == "__main__":
    if random.random() > 0.5:
        raise ValueError("Failed realization.")
    with open("poly.out", "w", encoding="utf-8") as f:
        f.write("\n".join(map(str, output)))

```
3. Run `Ensemble experiment`
About 50% of realizations should fail:

image

4. Hit `Restart`
Notice that only the failed realizations remain:

image

**Expected behaviour**
I expect `Restart` to restart only the failed realizations and keep the successful ones (successful ones should still be visisble in the gui).

frode-aarstad:
**Issue**
Resolves

**Approach**

## When applicable

### DIFF
--- a/src/ert/gui/simulation/run_dialog.py
+++ b/src/ert/gui/simulation/run_dialog.py
@@ -289,7 +289,7 @@ class RunDialog(QDialog):
         elif self.killJobs() != QMessageBox.Yes:
             QCloseEvent.ignore()  # type: ignore
 
-    def run_experiment(self) -> None:
+    def run_experiment(self, restart: bool = False) -> None:
         self._run_model.reset()
         self._snapshot_model.reset()
         self._tab_widget.clear()
@@ -302,6 +302,7 @@ class RunDialog(QDialog):
         def run() -> None:
             self._run_model.start_simulations_thread(
                 evaluator_server_config=evaluator_server_config,
+                restart=restart,
             )
 
         simulation_thread = ErtThread(
@@ -461,8 +462,7 @@ class RunDialog(QDialog):
             self.restart_button.setVisible(False)
             self.kill_button.setVisible(True)
             self.done_button.setVisible(False)
-            self._run_model.restart()
-            self.run_experiment()
+            self.run_experiment(restart=True)
 
     @Slot()
     def toggle_detailed_progress(self) -> None:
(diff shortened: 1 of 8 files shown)


dafeda:
As far as I can tell, this does not fix the problem described in #8032 .
Or am I missing something?

frode-aarstad:
> As far as I can tell, this does not fix the problem described in #8032 . Or am I missing something?

If the problem you are referring to is that you want the successful reals to also show up after you do a restart, I am not sure that is even possible to do without major refactoring

dafeda:
> > As far as I can tell, this does not fix the problem described in #8032 . Or am I missing something?
> 
> If the problem you are referring to is that you want the successful reals to also show up after you do a restart, I am not sure that is even possible to do without major refactoring

That is what issue #8032 describes I guess, so I was checking whether this PR solves that.
Which part does this PR solve?

frode-aarstad:
> > > As far as I can tell, this does not fix the problem described in #8032 . Or am I missing something?
> > 
> > 
> > If the problem you are referring to is that you want the successful reals to also show up after you do a restart, I am not sure that is even possible to do without major refactoring
> 
> That is what issue #8032 describes I guess, so I was checking whether this PR solves that. Which part does this PR solve?

It solves a bug that corrupts the active realizations selected after the second restart. Perhaps I can detatch this PR from the issue and put the parent issue back in todo

dafeda:
> > > > As far as I can tell, this does not fix the problem described in #8032 . Or am I missing something?
> > > 
> > > 
> > > If the problem you are referring to is that you want the successful reals to also show up after you do a restart, I am not sure that is even possible to do without major refactoring
> > 
> > 
> > That is what issue #8032 describes I guess, so I was checking whether this PR solves that. Which part does this PR solve?
> 
> It solves a bug that corrupts the active realizations selected after the second restart. It also makes each restart run in the same ensemble/experiment. Perhaps I can detatch this PR from the issue and put the parent issue back in todo

I think that makes sense as the other issue is still unsolved and I think we do want it solved eventually.

dafeda (approved):

PR #8278, opened 2024-07-02
Elevate generic queue options to global keyword

### LINKED ISSUE #8201: Elevate generic queue options to global keyword in Ert
Options that relate to the queue system but are valid for any queue system should be possible to configure without specifying which queue system it is valid for. 

Currently, in order to set `MAX_RUNNING` on LSF one must do:

```
QUEUE_SYSTEM LSF
QUEUE_OPTION LSF MAX_RUNNING 40
```

and if one later wants to change to Torque, one must change both these lines to bring along `MAX_RUNNING` (system specific values could make sense for max_running, but that is not true for all thinkable generic queue options.

There is currently only one GENERIC_QUEUE_OPTION:

https://github.com/equinor/ert/blob/761b07012717abe6d567cd9cb7d75130f03b122c/src/ert/config/queue_config.py#L21

while there are more queue related configuration items that arguably should be in that list (NUM_CPU, STOP_LONG_RUNNING, JOB_SCRIPT, SUBMIT_SLEEP, MAX_SUBMIT). 

Requiring adding `QUEUE_OPTION something` to these configuration lines is a breaking change, thus we need to something different to accomodate this.

A possibility is to let the config parser recognize GENERIC_QUEUE_OPTIONS similar to global keywords, and apply them to every configured queue. Queue specific options set using todays syntax could override. Another thinkable feature would be to add the possibility of using a wildcard like `QUEUE_OPTION * SOME_OPTION value` but that would be an extra feature.

DanSava:
**Issue**
Resolves #8201 

## When applicable


src/ert/config/parsing/queue_system.py
```diff
@@ -9,19 +9,22 @@ class QueueDriverEnum(BaseCEnum):  # type: ignore
     LOCAL = None
     TORQUE = None
     SLURM = None
+    GENERAL = None
```
andreas-el, line 12: `GENERIC` ?

src/ert/config/parsing/queue_system.py
```diff
@@ -9,19 +9,22 @@ class QueueDriverEnum(BaseCEnum):  # type: ignore
...
 QueueDriverEnum.addEnum("SLURM", 5)
+QueueDriverEnum.addEnum("GENERIC", 6)
 
 
 class QueueSystem(StrEnum):
     LSF = "LSF"
     LOCAL = "LOCAL"
     TORQUE = "TORQUE"
     SLURM = "SLURM"
+    GENERIC = "*"
```
andreas-el, line 27: I saw the issue had a proposal of using `*` as wildcard, but an option could be to specify `GENERIC` instead.
I'm personally not especially fond of wildcards.

tests/unit_tests/config/test_queue_config.py
```diff
@@ -331,3 +332,76 @@ def test_wrong_max_submit_raises_validation_error(max_submit_value, error_msg):
...
+@pytest.mark.usefixtures("use_tmpdir")
+@pytest.mark.parametrize(
+    "queue_system, key, value",
+    [
+        ("LSF", "MAX_RUNNING", 50),
+        ("SLURM", "MAX_RUNNING", 50),
+        ("TORQUE", "MAX_RUNNING", 50),
+        ("LSF", "SUBMIT_SLEEP", 4.2),
+        ("SLURM", "SUBMIT_SLEEP", 4.2),
+        ("TORQUE", "SUBMIT_SLEEP", 4.2),
```
andreas-el, line 346: I assume we are testing negative values somewhere else, and this test focuses only on values being set?
That might be the drivers responsibility I guess. But implementing generic options would require n-implementations of said checks.

src/ert/config/queue_config.py
```diff
@@ -107,57 +180,26 @@ def from_dict(cls, config_dict: ConfigDict) -> QueueConfig:
...
-            ),
-            False,
+        queue_options[selected_queue_system] = _add_generic_queue_options(
+            config_dict, queue_options[selected_queue_system]
         )
-        if selected_queue_system != QueueSystem.LOCAL and not project_code_is_set:
+
+        if "PROJECT_CODE" not in _option_list_to_dict(
+            queue_options[selected_queue_system]
+        ):
```
andreas-el, line 189: I guess PROJECT_CODE won't hurt on when using queue `LOCAL`.

src/ert/config/parsing/queue_system.py
DanSava: Good catch!

src/ert/config/parsing/queue_system.py
DanSava: I think ERT users are used to `*` as wildcards but I have no strong feelings towards wildcards so I can change it to `GENERIC = "GENERIC"`.

tests/unit_tests/config/test_queue_config.py
DanSava: Yes testing that these values are correct is already tests but it might be a good idea to add another test specifically when setting generic queue options.

src/ert/config/queue_config.py
DanSava: That is what I was thinking.

andreas-el (approved):
Solid work! 🥓

PR #8286, opened 2024-07-03
Refactor job_list & real_list

andreas-el:
Refactor several functions in real_list and job_list.
Commits are all standalone, one for each function.

## When applicable

### DIFF
--- a/src/ert/gui/model/job_list.py
+++ b/src/ert/gui/model/job_list.py
@@ -160,13 +160,11 @@ class JobListProxyModel(QAbstractProxyModel):
     def _source_data_changed(
         self, top_left: QModelIndex, bottom_right: QModelIndex, roles: List[int]
     ) -> None:
-        if not self._accept_index(top_left):
-            return
-        proxy_top_left = self.mapFromSource(top_left)
-        proxy_bottom_right = self.mapFromSource(bottom_right)
-        if not proxy_top_left.isValid() or not proxy_bottom_right.isValid():
-            return
-        self.dataChanged.emit(proxy_top_left, proxy_bottom_right, roles)
+        if self._accept_index(top_left):
+            proxy_top_left = self.mapFromSource(top_left)
+            proxy_bottom_right = self.mapFromSource(bottom_right)
+            if all([proxy_top_left.isValid(), proxy_bottom_right.isValid()]):
+                self.dataChanged.emit(proxy_top_left, proxy_bottom_right, roles)
 
     def _accept_index(self, index: QModelIndex) -> bool:
         if not index.internalPointer() or not index.data(IsJobRole):
--- a/src/ert/gui/model/job_list.py
+++ b/src/ert/gui/model/job_list.py
@@ -88,17 +88,16 @@ class JobListProxyModel(QAbstractProxyModel):
         orientation: Qt.Orientation,
         role: int = Qt.ItemDataRole.DisplayRole,
     ) -> Any:
-        if role != Qt.ItemDataRole.DisplayRole:
-            return QVariant()
-        if orientation == Qt.Orientation.Horizontal:
-            header = JOB_COLUMNS[section]
-            if header in [ids.STDOUT, ids.STDERR]:
-                return header.upper()
-            if header in [ids.CURRENT_MEMORY_USAGE, ids.MAX_MEMORY_USAGE]:
-                header = header.replace("_", " ")
-            return header.capitalize()
-        if orientation == Qt.Orientation.Vertical:
-            return section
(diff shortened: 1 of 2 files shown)


src/ert/gui/model/job_list.py
```diff
@@ -129,65 +126,51 @@ def parent(self, child: Optional[QModelIndex] = None) -> Optional[QObject]:
...
-        if not self._accept_index(sourceIndex):
-            return QModelIndex()
-        return self.index(sourceIndex.row(), sourceIndex.column(), QModelIndex())
+    def mapFromSource(self, src_index: QModelIndex) -> QModelIndex:
+        if src_index.isValid():
+            source_node = src_index.internalPointer()
+
+            if source_node and self._accept_index(src_index):
+                return self.index(src_index.row(), src_index.column(), QModelIndex())
+        return QModelIndex()
```
JHolba, line 152: Do we need the added internalPointer check?
It is also checked in _accept_index().
This could be simplified to
`return self.index(src_index.row(), src_index.column(), QModelIndex()) if src_index.isValid() and self._accept_index(src_index) else QModelIndex()`

src/ert/gui/model/job_list.py
```diff
@@ -129,65 +126,51 @@ def parent(self, child: Optional[QModelIndex] = None) -> Optional[QObject]:
...
+        if index.internalPointer() and index.data(IsJobRole):
+            # traverse upwards and check real and iter against parents of this index
+            while index.isValid() and index.internalPointer():
+                if (index.data(IsRealizationRole) and (index.row() != self._real)) or (
+                    index.data(IsEnsembleRole) and (index.row() != self._iter)
+                ):
+                    return False
+                index = index.parent()
+            return True
+        return False
```
JHolba, line 173: I would prefer it with less nesting.
```
if not index.internalPointer() or not index.data(IsJobRole):
    return False
```
at the the top and the rest the same. but up to you. I will approve it this way too

src/ert/gui/model/real_list.py
```diff
@@ -107,91 +90,71 @@ def parent(self, child: Optional[QModelIndex] = None) -> Optional[QObject]:
...
+            iter_index = sm.index(self._iter, 0, QModelIndex())
+            if iter_index.isValid() and sm.hasChildren(iter_index):
+                return sm.index(proxyIndex.row(), proxyIndex.column(), iter_index)
+        return QModelIndex()
+
+    def mapFromSource(self, src_index: QModelIndex) -> QModelIndex:
+        if src_index.isValid():
+            source_node = src_index.internalPointer()
+
+            if source_node and self._accept_index(src_index):
```
JHolba, line 124: could drop the source_node temp variable. it isn't used after this test

src/ert/gui/model/real_list.py
```diff
@@ -107,91 +90,71 @@ def parent(self, child: Optional[QModelIndex] = None) -> Optional[QObject]:
...
-            index = index.parent()
-        return True
+        if index.internalPointer() and index.data(IsRealizationRole):
+            # traverse upwards the tree, checking whether this index is accepted or not
+            while index.isValid() and index.internalPointer():
+                if index.data(IsEnsembleRole) and index.row() != self._iter:
+                    return False
+                index = index.parent()
+            return True
+        return False
```
JHolba, line 160: same as with _accept_index in job_list.py. would prefer less nesting.

src/ert/gui/model/real_list.py
```diff
@@ -25,40 +24,35 @@ def __init__(
     def get_iter(self) -> int:
         return self._iter
 
-    iter_changed = Signal(int)
```
JHolba, line 28: good catch :)

JHolba (changes requested):

src/ert/gui/model/job_list.py
andreas-el: Nice! I have replaced this proposal. 👍

src/ert/gui/model/job_list.py
andreas-el: Let's go with the less nesting part. It's easier to read. 👍

src/ert/gui/model/real_list.py
andreas-el: Same as the other one, let's go with the less nesting. 👍

src/ert/gui/model/real_list.py
andreas-el: Agreed. That made the function more or less identical to the job_list counterpart too. 👍

JHolba (approved):

PR #8288, opened 2024-07-03
Avoid unneeded lambda functions

berland:
@dafeda, I tried to figure out how to test this change in the GUI but I was unable to find out how. Do you know?

## When applicable

### DIFF
--- a/src/ert/gui/ertwidgets/analysismodulevariablespanel.py
+++ b/src/ert/gui/ertwidgets/analysismodulevariablespanel.py
@@ -121,11 +121,7 @@ class AnalysisModuleVariablesPanel(QWidget):
             localization_frame.layout().addWidget(self.local_spinner)
             layout.addRow(localization_frame)
 
-            local_checkbox.stateChanged.connect(
-                lambda localization_is_on: self.local_spinner.setEnabled(
-                    localization_is_on
-                )
-            )
+            local_checkbox.stateChanged.connect(self.local_spinner.setEnabled)
             local_checkbox.setChecked(analysis_module.localization)
 
         self.setLayout(layout)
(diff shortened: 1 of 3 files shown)


berland:
# unnecessary-lambda (PLW0108)

Fix is always available.

## What it does
Checks for `lambda` definitions that consist of a single function call
with the same arguments as the `lambda` itself.

## Why is this bad?
When a `lambda` is used to wrap a function call, and merely propagates
the `lambda` arguments to that function, it can typically be replaced with
the function itself, removing a level of indirection.

## Example
```python
df.apply(lambda x: str(x))
```

Use instead:
```python
df.apply(str)
```

## Fix safety
This rule's fix is marked as unsafe for two primary reasons.

First, the lambda body itself could contain an effect.

For example, replacing `lambda x, y: (func()(x, y))` with `func()` would
lead to a change in behavior, as `func()` would be evaluated eagerly when
defining the lambda, rather than when the lambda is called.

However, even when the lambda body itself is pure, the lambda may
change the argument names, which can lead to a change in behavior when
callers pass arguments by name.

For example, replacing `foo = lambda x, y: func(x, y)` with `foo = func`,
where `func` is defined as `def func(a, b): return a + b`, would be a
breaking change for callers that execute the lambda by passing arguments by
name, as in: `foo(x=1, y=2)`. Since `func` does not define the arguments
`x` and `y`, unlike the lambda, the call would raise a `TypeError`.

src/ert/gui/ertwidgets/analysismodulevariablespanel.py
```diff
@@ -121,11 +121,7 @@ def __init__(self, analysis_module: AnalysisModule, ensemble_size: int):
             localization_frame.layout().addWidget(self.local_spinner)
             layout.addRow(localization_frame)
 
-            local_checkbox.stateChanged.connect(
-                lambda localization_is_on: self.local_spinner.setEnabled(
-                    localization_is_on
-                )
-            )
+            local_checkbox.stateChanged.connect(self.local_spinner.setEnabled)
```
berland, line 124: This change has been verified manually to work in the GUI.

xjules (approved):
Looks fine with me! :rocket:

PR #8289, opened 2024-07-03
Fix typing issues in gui model

JHolba:
**Approach**
Add typing
Add none checks

## When applicable


.mypy.ini
```diff
@@ -38,24 +38,10 @@ ignore_missing_imports = True
...
-ignore_missing_imports = True
-ignore_errors = True
-
-[mypy-ert.gui.model.snapshot]
-ignore_missing_imports = True
-ignore_errors = True
-
 [mypy-ert.gui.ertwidgets.validationsupport]
-ignore_missing_imports = True
 ignore_errors = True
```
jonathan-eq, line 42: Why do we want to keep this one?

JHolba: because it still has typing issues :melting_face:

src/ert/gui/model/snapshot.py
```diff
@@ -145,12 +146,13 @@ def prerender(
             metadata["aggr_job_status_colors"][real_id][forward_model_id] = color
 
         if isSnapshot:
+            snapshot = cast(Snapshot, snapshot)  # mypy/pyright needs this
```
jonathan-eq, line 149: Let's remove the comment 👌

src/ert/gui/model/snapshot.py
```diff
@@ -317,18 +323,14 @@ def parent(self, child: Optional[QModelIndex] = None) -> Optional[QObject]:
         return self.createIndex(parent_item.row(), 0, parent_item)
 
     @override
-    def data(
-        self, index: QModelIndex, role: Qt.ItemDataRole = Qt.ItemDataRole.DisplayRole
-    ) -> Any:
+    def data(self, index: QModelIndex, role: int = Qt.ItemDataRole.DisplayRole) -> Any:
```
jonathan-eq, line 326: Can we use the a union of all the predefined roles from lines 27-36 here instead of int for role?

src/ert/gui/model/snapshot.py
```diff
@@ -403,8 +405,10 @@ def _real_data(_index: QModelIndex, node: RealNode, role: int):
 
     @staticmethod
     def _job_data(
-        index: QModelIndex, node: ForwardModelStepNode, role: Qt.ItemDataRole
-    ):
+        index: QModelIndex,
+        node: ForwardModelStepNode,
+        role: int,  # Qt.ItemDataRole
```
jonathan-eq, line 410: Same here^
Then we would not need the comment

src/ert/gui/model/snapshot.py
```diff
@@ -403,8 +405,10 @@ def _real_data(_index: QModelIndex, node: RealNode, role: int):
...
+    ) -> Any:
```
jonathan-eq, line 411: Is it a lot of work to get rid of the Any type?

jonathan-eq (changes requested):
Very good! 
See minor nitpicking 🥇

src/ert/gui/model/snapshot.py
JHolba: the qt stubs define that parameter as int, so we have to as well

src/ert/gui/model/snapshot.py
JHolba: same as above

src/ert/gui/model/snapshot.py
JHolba: there is no point. SnapshotModel.data() is the only function that calls this and it has to return Any due to inheriting QAbstractItemModel.data().

src/ert/gui/model/snapshot.py
JHolba: done

jonathan-eq (approved):
LGTM!

jonathan-eq:
Resolve conflicts, squash commits, then we are ready to roll! :ship:

PR #8292, opened 2024-07-03
Show ert filename in eventviewer windowtitle

### LINKED ISSUE #4062: Add config filename to event viewer window
I want to differentiate between event viewers from different ERT runs when I have multiple instances of ERT open. 
image

The name of the config file added to the title of the event viewer window, similar to how the main window and simulations window work. 

N/A
**Additional context**
[Internal link to request](https://equinor.slack.com/archives/CDPJULWR4/p1665391577255179?thread_ts=1652179346.489499&cid=CDPJULWR4)

andreas-el:
**Issue**
Resolves #4062 

Screenshot 2024-07-03 at 16 14 08

## When applicable

### DIFF
--- a/src/ert/gui/main.py
+++ b/src/ert/gui/main.py
@@ -267,7 +267,7 @@ def _setup_main_window(
     window.addTool(PluginsTool(plugin_handler, window.notifier, config))
     window.addTool(RunAnalysisTool(config, window.notifier))
     window.addTool(LoadResultsTool(facade, window.notifier))
-    event_viewer = EventViewerTool(log_handler)
+    event_viewer = EventViewerTool(log_handler, config_file)
     window.addTool(event_viewer)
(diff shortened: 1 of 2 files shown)


DanSava (approved):
Good improvement :+1: 

I not sure how common this is but if the user is running multiple instances of ERT from projects with the same config file name the issue will persist. Adding a full path to the event viewer is not really a good idea so I think this is good enough.

andreas-el:
> Good improvement 👍
> 
> I not sure how common this is but if the user is running multiple instances of ERT from projects with the same config file name the issue will persist. Adding a full path to the event viewer is not really a good idea so I think this is good enough.

I thought about that, and came to the conclusion that it was sufficient for now. Seeing that the event viewer now uses the same configfile / string as in the windowtitle of Ert main window.
If we want to add path to be more precise, we should also add it there.

PR #8293, opened 2024-07-03
Fix and enforce usage of encoding when writing text files

berland:
**Issue**
Resolves ruff exception that we dont want to keep.

**Approach**
Add  `encoding="utf-8"`

## When applicable

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -187,7 +187,6 @@ ignore = ["PLW2901",  # redefined-loop-name
           "PLR6201",  # literal-membership
           "PLR0914",  # too-many-locals
           "F841",  # unused-variable
-          "PLW1514",  # unspecified-encoding
           "PLR6104",  # non-augmented-assignment
           "PLR6301",  # no-self-use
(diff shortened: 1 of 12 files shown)


berland:
# unspecified-encoding (PLW1514)

Fix is always available.

## What it does
Checks for uses of `open` and related calls without an explicit `encoding`
argument.

## Why is this bad?
Using `open` in text mode without an explicit encoding can lead to
non-portable code, with differing behavior across platforms.

Instead, consider using the `encoding` parameter to enforce a specific
encoding. [PEP 597] recommends using `locale.getpreferredencoding(False)`
as the default encoding on versions earlier than Python 3.10, and
`encoding="locale"` on Python 3.10 and later.

## Example
```python
open("file.txt")
```

Use instead:
```python
open("file.txt", encoding="utf-8")
```

## References
- [Python documentation: `open`](https://docs.python.org/3/library/functions.html#open)

[PEP 597]: https://peps.python.org/pep-0597/

jonathan-eq (approved):
LGTM!

PR #8294, opened 2024-07-03
Solve ruff collapsible-else-if

berland:
This reduces indentation.

**Issue**
Ruff issue PLR5501

**Approach**
`ruff check --fix .` - no manual edits needed !

## When applicable

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -181,7 +181,6 @@ ignore = ["PLW2901",  # redefined-loop-name
           "PLR2004",  # magic-value-comparison
           "PLR0915",  # too-many-statements
           "PLR0912",  # too-many-branches
-          "PLR5501",  # collapsible-else-if
           "PLR0911",  # too-many-return-statements
           "PLC2701",  # import-private-name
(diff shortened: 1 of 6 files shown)


berland:
# collapsible-else-if (PLR5501)

Fix is sometimes available.

## What it does
Checks for `else` blocks that consist of a single `if` statement.

## Why is this bad?
If an `else` block contains a single `if` statement, it can be collapsed
into an `elif`, thus reducing the indentation level.

## Example
```python
def check_sign(value: int) -> None:
    if value > 0:
        print("Number is positive.")
    else:
        if value < 0:
            print("Number is negative.")
        else:
            print("Number is zero.")
```

Use instead:
```python
def check_sign(value: int) -> None:
    if value > 0:
        print("Number is positive.")
    elif value < 0:
        print("Number is negative.")
    else:
        print("Number is zero.")
```

## References
- [Python documentation: `if` Statements](https://docs.python.org/3/tutorial/controlflow.html#if-statements)

jonathan-eq (approved):
Good improvement! 🎈

PR #8295, opened 2024-07-03
Avoid import outside top-level

berland:
Keeping as exception at one spot due to circular dependency and one spot due to measurable time-loss to do top-level import.

Ignoring the issue in script/build, this file is being deleted soon.

**Issue**
Resolves ruff PLC0415

**Approach**
fix/add exception/comment.

## When applicable

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -185,7 +185,6 @@ ignore = ["PLW2901",  # redefined-loop-name
           "PLR6104",  # non-augmented-assignment
           "PLR6301",  # no-self-use
           "PLW1641",  # eq-without-hash
-          "PLC0415",  # import-outside-toplevel
           "PLR0904",  # too-many-public-methods
           "PLR1702",  # too-many-nested-blocks
           "PLW3201",  # bad-dunder-method-name
--- a/src/ert/__main__.py
+++ b/src/ert/__main__.py
@@ -184,7 +184,8 @@ def valid_port_range(user_input: str) -> range:
 
 
 def run_gui_wrapper(args: Namespace, ert_plugin_manager: ErtPluginManager) -> None:
-    from ert.gui.main import run_gui
+    # Importing ert.gui on-demand saves ~0.5 seconds off `from ert import __main__`
+    from ert.gui.main import run_gui  # noqa: PLC0415
 
     run_gui(args, ert_plugin_manager)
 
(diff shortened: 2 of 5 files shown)


berland:
# import-outside-top-level (PLC0415)

## What it does
Checks for `import` statements outside of a module's top-level scope, such
as within a function or class definition.

## Why is this bad?
[PEP 8] recommends placing imports not only at the top-level of a module,
but at the very top of the file, "just after any module comments and
docstrings, and before module globals and constants."

`import` statements have effects that are global in scope; defining them at
the top level has a number of benefits. For example, it makes it easier to
identify the dependencies of a module, and ensures that any invalid imports
are caught regardless of whether a specific function is called or class is
instantiated.

An import statement would typically be placed within a function only to
avoid a circular dependency, to defer a costly module load, or to avoid
loading a dependency altogether in a certain runtime environment.

## Example
```python
def print_python_version():
    import platform

    print(python.python_version())
```

Use instead:
```python
import platform


def print_python_version():
    print(python.python_version())
```

[PEP 8]: https://peps.python.org/pep-0008/#imports

jonathan-eq (approved):
LGTM!

src/ert/shared/share/ert/workflows/jobs/internal-gui/scripts/csv_export.py
```diff
@@ -6,6 +6,10 @@
 
 from ert import LibresFacade
 from ert.config import CancelPluginException, ErtPlugin
+from ert.gui.ertwidgets.customdialog import CustomDialog
+from ert.gui.ertwidgets.listeditbox import ListEditBox
+from ert.gui.ertwidgets.models.path_model import PathModel
```
eivindjahren, line 11: I think I moved these in at some point and it was done for the same reason as 
```python
def run_gui_wrapper(args: Namespace, ert_plugin_manager: ErtPluginManager) -> None:
    # Importing ert.gui on-demand saves ~0.5 seconds off `from ert import __main__`
    from ert.gui.main import run_gui  # noqa: PLC0415
```
When you run `ert test_run`, gui will not be imported when you place these imports inside `getArguments`.

eivindjahren (changes requested):

src/ert/shared/share/ert/workflows/jobs/internal-gui/scripts/csv_export.py
berland: fixed!

eivindjahren (approved):

PR #8296, opened 2024-07-03
Solve ruff issue non-augmented assignment

berland:
**Issue**
Resolves ruff issue  on non-augmented assignment

**Approach**
`ruff check --unsafe-fixes --fix .`, 👀 

## When applicable

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -182,7 +182,6 @@ ignore = ["PLW2901",  # redefined-loop-name
           "PLR6201",  # literal-membership
           "PLR0914",  # too-many-locals
           "F841",  # unused-variable
-          "PLR6104",  # non-augmented-assignment
           "PLR6301",  # no-self-use
           "PLW1641",  # eq-without-hash
           "PLC0415",  # import-outside-toplevel
--- a/src/_ert_forward_model_runner/io/__init__.py
+++ b/src/_ert_forward_model_runner/io/__init__.py
@@ -22,7 +22,7 @@ def assert_file_executable(fname):
 
     potential_executables = [os.path.abspath(fname)]
     if not os.path.isabs(fname):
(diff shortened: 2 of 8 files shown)


berland:
# non-augmented-assignment (PLR6104)

Fix is always available.

## What it does
Checks for assignments that can be replaced with augmented assignment
statements.

## Why is this bad?
If an assignment statement consists of a binary operation in which one
operand is the same as the assignment target, it can be rewritten as an
augmented assignment. For example, `x = x + 1` can be rewritten as
`x += 1`.

When performing such an operation, augmented assignments are more concise
and idiomatic.

## Known problems
In some cases, this rule will not detect assignments in which the target
is on the right-hand side of a binary operation (e.g., `x = y + x`, as
opposed to `x = x + y`), as such operations are not commutative for
certain data types, like strings.

For example, `x = "prefix-" + x` is not equivalent to `x += "prefix-"`,
while `x = 1 + x` is equivalent to `x += 1`.

If the type of the left-hand side cannot be inferred trivially, the rule
will ignore the assignment.

## Example
```python
x = x + 1
```

Use instead:
```python
x += 1
```

## Fix safety
This rule's fix is marked as unsafe, as augmented assignments have
different semantics when the target is a mutable data type, like a list or
dictionary.

For example, consider the following:

```python
foo = [1]
bar = foo
foo = foo + [2]
assert (foo, bar) == ([1, 2], [1])
```

If the assignment is replaced with an augmented assignment, the update
operation will apply to both `foo` and `bar`, as they refer to the same
object:

```python
foo = [1]
bar = foo
foo += [2]
assert (foo, bar) == ([1, 2], [1, 2])
```

JHolba (approved):
Nice changes.
The comment listing pitfalls was excellent :)

Issue #8300, opened 2024-07-04
Flaky tests/unit_tests/gui/test_restart_ensemble_experiment.py::test_restart_failed_realizations


yngve-sk:
Happens with both scheduler and job queue, a bit randomly. Likely tied to the randomness of the failed realizations in the test itself, which could/should be addressed.

**Describe the bug**

First occurrence @ commit https://github.com/equinor/ert/commit/f9dc451d9dc4348d36b99abfca11fd0470bb95cf
Revert of it still causes failure; https://github.com/equinor/ert/actions/runs/9792643894/job/27039016350

Potential causing commits are likely located near the first occurrence, it could also be some external package change.
Reverting https://github.com/equinor/ert/commit/7d3b096c256ef7cf98f2b81e805c84249b91a999 did not stop the test from failing

PR #8307, opened 2024-07-05
Test only deterministic realization fails

yngve-sk:
**Issue**
Resolves https://github.com/equinor/ert/issues/8300

Can be inspected with `pytest --show-gui tests/unit_tests/gui/test_restart_ensemble_experiment.py -k test_restart_failed_realizations`

Screenshot 2024-07-05 at 12 35 26
Screenshot 2024-07-05 at 12 35 37
Screenshot 2024-07-05 at 12 35 49


andreas-el (approved):
I think this is better as it leaves nothing up to chance. 🌴

PR #8310, opened 2024-07-08
Faster date parsing

JHolba:
datetime parsing while unpacking json in cloudevents is *very* slow currently.

**Approach**
Replaced datetime parsing library with a new one that is included in python 3.11. Using backport package for older python versions. The new one should be a bit more than two orders of magnitude faster.

## When applicable

### DIFF
--- a/src/ert/serialization/_evaluator.py
+++ b/src/ert/serialization/_evaluator.py
@@ -1,5 +1,5 @@
-import datetime
 import json
+from datetime import date, datetime
 from typing import Any
 
 from backports.datetime_fromisoformat import MonkeyPatch  # type: ignore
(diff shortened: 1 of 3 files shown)


JHolba:
the backports package is written by the author of the ciso8601 python package. ciso8601 may be a better parser, but I figured it is better to go for a package that we can phase out when we move to python 3.11. 
https://github.com/closeio/ciso8601

src/ert/serialization/_evaluator.py
```diff
@@ -1,13 +1,15 @@
-import datetime
 import json
+from datetime import date, datetime
 from typing import Any
 
-from dateutil import parser
+from backports.datetime_fromisoformat import MonkeyPatch  # type: ignore
+
+MonkeyPatch.patch_fromisoformat()
```
jonathan-eq, line 7: Are those the only places we do date parsing?

jonathan-eq (approved):
Very good!

src/ert/serialization/_evaluator.py
JHolba: No, but it was the only places that mattered according to the profiling I did

PR #8315, opened 2024-07-11
Fix typing issues in gui code

JHolba:
Some parts of the gui are still ignore in .mypy.ini

**Approach**
Added casts where needed
Changed method params to match overridden method
Added None check
Typing added to function signatures

## When applicable

### DIFF
--- a/src/ert/gui/ertwidgets/analysismodulevariablespanel.py
+++ b/src/ert/gui/ertwidgets/analysismodulevariablespanel.py
@@ -1,4 +1,7 @@
+from __future__ import annotations
+
 from functools import partial
+from typing import Type, cast
 
 from annotated_types import Ge, Gt, Le
 from qtpy.QtCore import Qt
@@ -10,11 +13,17 @@ from qtpy.QtWidgets import (
     QFrame,
     QHBoxLayout,
     QLabel,
+    QLayout,
     QWidget,
 )
 from typing_extensions import get_args
 
-from ert.config.analysis_module import AnalysisModule, IESSettings
+from ert.config.analysis_module import (
+    AnalysisModule,
+    IESSettings,
+    InversionTypeES,
+    InversionTypeIES,
+)
 
 
 class AnalysisModuleVariablesPanel(QWidget):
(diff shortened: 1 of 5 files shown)


src/ert/gui/ertwidgets/analysismodulevariablespanel.py
```diff
@@ -90,7 +101,8 @@ def __init__(self, analysis_module: AnalysisModule, ensemble_size: int):
 
             localization_frame = QFrame()
             localization_frame.setLayout(QHBoxLayout())
-            localization_frame.layout().setContentsMargins(0, 0, 0, 0)
+            lf_layout = cast(QLayout, localization_frame.layout())
```
yngve-sk, line 104: (nitpick) Why cast instead of assert in this case?

src/ert/gui/ertwidgets/analysismodulevariablespanel.py
```diff
@@ -135,7 +147,7 @@ def update_inversion_algorithm(self, text: str) -> None:
         self.truncation_spinner.setEnabled(
             not any(val in text.lower() for val in ["direct", "exact"])
         )
-        self.analysis_module.inversion = text
+        self.analysis_module.inversion = text  # type: ignore[assignment]
```
yngve-sk, line 150: (nitpick) Would consider just adding a setter to AnalysisModule.inversion if it doesn't bloat the code too much, it is a BaseModel and pydantic2 supports it

src/ert/gui/ertwidgets/analysismodulevariablespanel.py
```diff
@@ -170,8 +182,13 @@ def createDoubleSpinBox(
...
 
-    def valueChanged(self, variable_name, variable_type, variable_control):
-        value = None
+    def valueChanged(
+        self,
+        variable_name: str,
+        variable_type: Type[bool] | Type[float],
+        variable_control: QWidget,
+    ) -> None:
+        value: bool | float | None = None
```
yngve-sk, line 191: I think we need to use the old style convention because we support python3.8

src/ert/gui/ertwidgets/analysismodulevariablespanel.py
JHolba: because cast is free and we know that it is not none. plus we can in-line the cast :)

src/ert/gui/ertwidgets/analysismodulevariablespanel.py
JHolba: `from __future__ import annotation`
takes care of that

yngve-sk (approved):
🚀

PR #8328, opened 2024-07-17
Fix phase count in es mda restarts

larsevj:
**Issue**
Resolves #my_issue

**Approach**

## When applicable

### DIFF
--- a/src/ert/run_models/ensemble_experiment.py
+++ b/src/ert/run_models/ensemble_experiment.py
@@ -93,7 +93,7 @@ class EnsembleExperiment(BaseRunModel):
 
         self._evaluate_and_postprocess(run_context, evaluator_server_config)
 
-        self.setPhase(phase_count, "Simulations completed.")
+        self.setPhase(phase_count, "Experiment completed.")
 
         return run_context
 
--- a/src/ert/run_models/evaluate_ensemble.py
+++ b/src/ert/run_models/evaluate_ensemble.py
@@ -75,7 +75,7 @@ class EvaluateEnsemble(BaseRunModel):
         self.setPhaseCount(phase_count)
         self._evaluate_and_postprocess(prior_context, evaluator_server_config)
 
-        self.setPhase(phase_count, "Simulations completed.")
+        self.setPhase(phase_count, "Experiment completed.")
 
         return prior_context
 


src/ert/run_models/multiple_data_assimilation.py
```diff
@@ -191,7 +191,7 @@ def run_experiment(
 
         self.setPhaseName("Post processing...")
 
-        self.setPhase(self.number_of_iterations, "Experiment completed.")
+        self.setPhase(self.phaseCount(), "Experiment completed.")
```
yngve-sk, line 194: Perhaps just write `self.start_iteration + self.number_of_iterations` instead of `self.phaseCount()`, a bit less indirect. In a separate PR it would then maybe make sense to rename `number_of_iterations` to `number_of_iterations_from_start` to make it more explicit.

tests/unit_tests/gui/test_restart_esmda.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+    gui = ensemble_experiment_has_run_no_failure
+
+    experiment_panel = get_child(gui, ExperimentPanel)
+    simulation_mode_combo = get_child(experiment_panel, QComboBox)
+    simulation_mode_combo.setCurrentText(MultipleDataAssimilation.name())
+
+    es_mda_panel = gui.findChild(QWidget, name="ES_MDA_panel")
+    assert es_mda_panel
+    restart_button = es_mda_panel.findChild(QCheckBox, name="restart_checkbox_esmda")
```
yngve-sk, line 24: Would prefer `restart_checkbox` as it is a checkbox and not a button

tests/unit_tests/gui/test_restart_esmda.py
```diff
@@ -0,0 +1,76 @@
...
+    run_dialog = gui.findChild(RunDialog)
+    qtbot.waitUntil(run_dialog.done_button.isVisible, timeout=60000)
+    assert (
+        run_dialog._total_progress_label.text()
+        == "Total progress 100% — Experiment completed."
+    )
+
+    qtbot.mouseClick(run_dialog.done_button, Qt.MouseButton.LeftButton)
+
+    # Now we try to restart from ensemble iter-0_2
```
yngve-sk, line 52: How long does it take to run this part? If not too long, It'd be nice to do a loop over `["iter-0_1","iter-0_2"] to restart from every possible starting point given the default weights. If it is slow, the whole thing could be speeded up by reducing the number of realizations to 2

yngve-sk (approved):
Gogogo. Although we should look into doing a bigger look into removing this phaseCount notion

PR #8329, opened 2024-07-18
Add traceback information to ErtThreadError

### LINKED ISSUE #7856: ErtThread should show traceback on error
Currently, if an exception is thrown by ErtThread (only used in tests), then the thread message is shown, but not the traceback.

DanSava:
**Issue**
Resolves #7856

## When applicable

### DIFF
--- a/src/_ert/threading.py
+++ b/src/_ert/threading.py
@@ -4,6 +4,7 @@ import logging
 import os
 import signal
 import threading
+import traceback
 from threading import Thread as _Thread
 from types import FrameType
 from typing import Any, Callable, Iterable, Optional
@@ -16,17 +17,25 @@ _can_raise = False
 
 
 class ErtThreadError(Exception):
-    def __init__(self, exception: BaseException, thread: _Thread) -> None:
+    def __init__(
+        self, exception: BaseException, thread: _Thread, err_traceback: str
+    ) -> None:
         super().__init__(repr(exception))
         self._exception = exception
         self._thread = thread
+        self._err_traceback = err_traceback
 
     @property
     def exception(self) -> BaseException:
         return self._exception
 
     def __str__(self) -> str:
-        return f"{self._exception} in thread '{self._thread.name}'"
+        return (
+            f"{self._exception} in thread '{self._thread.name}'\n {self._err_traceback}"
+        )
+
+    def __repr__(self) -> str:
+        return str(self)
 
 
 class ErtThread(_Thread):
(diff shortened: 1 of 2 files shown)


src/_ert/threading.py
```diff
@@ -46,14 +55,17 @@ def run(self) -> None:
         try:
             super().run()
         except BaseException as exc:
-            logger.error(str(exc), exc_info=exc)
+            err_traceback = str(traceback.format_exc())
+            logger.error(err_traceback, exc_info=exc)
             if _can_raise and self._should_raise:
-                _raise_on_main_thread(exc)
+                _raise_on_main_thread(exc, err_traceback)
```
yngve-sk, line 61: Is there some special reason for getting the traceback here and passing it as a param instead of using exc.__traceback__ in ErtThreadErrror.__init__ and skipping the param? If it is possible to skip the param and keep the same behavior that'd would be ideal.

yngve-sk (commented):
Overall looks good, I tested it on some integration tests and in pycharm I'm getting the error messages with traceback. For me as a developer this is the most important. However when I run it via the terminal, ert gui / cli, I'm not getting traceback. Do we want it to be there too?

If I raise an `AssertionError("DePR")` from `base_run_model.py` at the end of a postprocess, I get this output in the cli:
```
Experiment failed with the following error: DePR

DePR

```

And in the GUI I get this
Screenshot 2024-07-19 at 13 18 10

src/_ert/threading.py
DanSava: >Is there some special reason for getting the traceback here and passing it as a param instead of using exc.traceback

Not sure what you mean by `exc.traceback` when I tried it I got `{AttributeError}AttributeError("'ValueError' object has no attribute 'traceback'")` there is something internal `exc.__traceback__` and I am not sure how to parse this object and extract the same information as I am getting from `traceback.format_exc()`. It might be possible but I am not sure it will be as straight forward as the current implementation. 

>  If it is possible to skip the param and keep the same behavior that'd would be ideal.

I think it is possible to skip the `err_traceback` param definition in `_raise_on_main_thread` I can try it.

DanSava:
> However when I run it via the terminal, ert gui / cli, I'm not getting traceback. Do we want it to be there too?

I think as long as there is an unhandled exception in an ErtThread that should contain the traceback with the current changes. 

> If I raise an `AssertionError("DePR")` from `base_run_model.py` at the end of a postprocess, I get this output in the cli:

This one was fun trying to investigate what is going on. seem that if you raise an exception in `base_run_model.py` at the end of a postprocess that exception is handled internally in this section https://github.com/equinor/ert/blob/a21015db7a89e01ea6a2ed69ff10cdc87da4eea6/src/ert/run_models/base_run_model.py#L317

This means we don't get to raise a `ErtThreadError`

eivindjahren (approved):
:+1:

PR #8347, opened 2024-07-25
Simplify openpbs job states

eivindjahren:

## When applicable

### DIFF
--- a/src/ert/scheduler/openpbs_driver.py
+++ b/src/ert/scheduler/openpbs_driver.py
@@ -20,7 +20,6 @@ from typing import (
     Type,
     Union,
     cast,
-    get_args,
     get_type_hints,
 )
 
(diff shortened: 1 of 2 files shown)


src/ert/scheduler/openpbs_driver.py
```diff
@@ -30,7 +29,7 @@
 logger = logging.getLogger(__name__)
 
 _POLL_PERIOD = 2.0  # seconds
-JobState = Literal[
+JOB_STATES = [
```
xjules, line 32: The issue is that we have the same pattern; ie. Literal states in other drivers too.

eivindjahren: This was only used to iterate over in the openpbs driver and in the test so it didn't need to be a Literal, just a list. The other ones seem to be also used for typing.

xjules: Seems to be fine though.

xjules (approved):
Let's ship it! :rocket:

PR #8364, opened 2024-07-31
Show worker ID on test start

yngve-sk:
To find out which tests crash succeeding tests


yngve-sk:
Seems to just show [master], might be possible but probably requires very "creative" solution.

Issue #8359, opened 2024-07-29
Realization is not cancelled properly after MAX_RUNTIME


xjules:
Flaky test it seems: 
`FAILED tests/integration_tests/status/test_tracking_integration.py::test_tracking[ee_poly_experiment_cancelled_by_max_runtime]`

It appears that after realization reaches MAX_RUNTIME the cancellation won't happen and the state would be `PENDING` instead of `FAILED`.
```
 assert match.value == expected, (
                f"{msg_start}{str(match.full_path)} value "
                f"({match.value}) is not equal to ({expected})"
            )
E           AssertionError: Snapshot 0 did not match:
E             reals.1.forward_models.0.status value (Pending) is not equal to (Failed)
E           assert 'Pending' == 'Failed'
```
Please see for more details:

https://github.com/equinor/ert/actions/runs/10143243141/job/28044195831

xjules:
Reproduced when `Scheduler._publisher()` is slow - injecting some `asyncio.sleep(5)` should make do. This naturally will prevent all the events to be sent when MAX_RUNTIME will cancel all realizations and thus the realization.STATUS will be either `RUNNING` or `PENDING`.

eivindjahren:
Another failure here: https://github.com/equinor/ert/actions/runs/10193785679/job/28199058436?pr=8362

xjules:
Further updates: `dispatcher_task` that runs `_batch_events_into_buffer` is unable the correctly append the last batch of events due the task being cancelled after the realization is hit by `max_runtime`. Reproduction of the issue is not so straightforward, but injecting `await asyncio.sleep(xx)` into `_batch_events_into_buffer` in the beginning should do.

xjules:
Another failure occurred: https://github.com/equinor/ert/actions/runs/10454681280/job/28947875443?pr=8424

xjules:
Let's see if this PR might help: https://github.com/equinor/ert/pull/8424

xjules:
I haven't seen this error since, so closing this one.

xjules:
Another entry https://github.com/equinor/ert/actions/runs/10683277953/job/29610991386?pr=8629

eivindjahren:
Another failure here: https://github.com/equinor/ert/actions/runs/10719695610/job/29724320771?pr=8648

xjules:
Seems like this issue still persists: https://github.com/equinor/ert/actions/runs/10738432908/job/29782008935

xjules:
[...]

PR #8370, opened 2024-08-01
Make batching task finite by adding CLOSE_EVENT_SENTINEL

### LINKED ISSUE #8070: Presumably finished realization displayed as pending in joblist view
A user has reported the following issue when running ~~scheduler~~ JobQueue.

--- 

Hi.. 

Currently I am experiencing several realizations finished without doing the steps in the configuration and it is labeled as 'finished'.

 

I know sometimes it is common to have some error due to maybe random licensed error, etc. therefore I set max_submit = 3 just to make sure that if something like that happened (failed run), the case will be restarted.

 

However if the label is 'finished' I think the case wont be restarted. in some cases it can be more than 5 realization like that (picture attached)

Image

xjules:
**Issue**
~~Resolves https://github.com/equinor/ert/issues/8359~~
Resolves #8070 

**Approach**
That error might be due to the last batch of events not being processed. Therefore adding SENTINEL object.

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -56,6 +56,8 @@ logger = logging.getLogger(__name__)
 
 EVENT_HANDLER = Callable[[List[CloudEvent]], Awaitable[None]]
 
+CLOSE_EVENT_SENTINEL = object()
+
 
 class EnsembleEvaluator:
     def __init__(self, ensemble: Ensemble, config: EvaluatorServerConfig):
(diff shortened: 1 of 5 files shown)


src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -271,23 +279,10 @@ async def handle_dispatch(self, websocket: WebSocketServerProtocol) -> None:
                             f"ignoring since I am {self.ensemble.id_}"
                         )
                         continue
-                    try:
-                        if event["type"] == EVTYPE_FORWARD_MODEL_CHECKSUM:
-                            await self.forward_checksum(event)
-                        else:
-                            await self._events.put(event)
-                    except BaseException as ex:
```
berland, line 279: This should be a separate commit (you choose if also a different PR), to be able to have a commit message properly explaining why.

xjules: Ok, can remove it from this PR.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -56,6 +56,8 @@
 
 EVENT_HANDLER = Callable[[List[CloudEvent]], Awaitable[None]]
 
+CLOSE_EVENT_SENTINEL = object()
```
berland, line 59: Should we make this into a dummy class, so that we can type the queue more properly than  `Any`

xjules:
Closing this one as this seems not to fix the underlying issue with PENDING realizations.

PR #8393, opened 2024-08-06
Fix waiting cursor in load results manually panel

### LINKED ISSUE #7668: Load results ui elements indicate work is being done when processing has completed
Using 2024.04.01-py38-rhel7 and RGS.

First initialise an ensemble
Select `Manage experiments` -> `Create a new experiment`

At this point I don't think it matters if you initialise the ensemble or not, the results when loading seem to be the same.

Select `Load results manually` -> `Load`

User is presented with this, where the mouse pointer indicate waiting or processing (which is not happening).

Screenshot 2024-04-15 at 08 37 45

frode-aarstad:
**Issue**
Resolves #7668 

**Approach**

## When applicable

### DIFF
--- a/src/ert/gui/tools/load_results/load_results_panel.py
+++ b/src/ert/gui/tools/load_results/load_results_panel.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from qtpy.QtCore import Qt
 from qtpy.QtWidgets import QFormLayout, QMessageBox, QTextEdit, QWidget
 
 from ert.gui.ertnotifier import ErtNotifier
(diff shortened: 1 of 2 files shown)


dafeda:
The same window still seems to pop up but now there's no waiting-cursor. Is this the intention?

image

frode-aarstad:
> The same window still seems to pop up but now there's no waiting-cursor. Is this the intention?
> 

My interpretation of the issue was to remove the waiting cursor. Perhaps @andreas-el can clarify ?

andreas-el:
> > The same window still seems to pop up but now there's no waiting-cursor. Is this the intention?
> 
> My interpretation of the issue was to remove the waiting cursor. Perhaps @andreas-el can clarify ?

In the bug reported, I experienced that the system would try to load ensembles and fail (there were no ensembles to load). 
There was a visual problem with the cursor indicating that more will happen if you wait.
The cursor will make sense as long as the system is actually processing, but should revert back to normal once loading comes to an end.

dafeda (approved):

PR #8396, opened 2024-08-06
Deprecate slurm queue option MEMORY_PER_CPU

### LINKED ISSUE #8199: Deprecate Slurm queue option MEMORY_PER_CPU
This slurm option does not go hand-in-hand with the unification of memory handling in Ert that is supposed to happen in #7798 , and is therefore suggested to be deprecated.

Adding deprecation messages to the config parsing is sufficient, stating that is feature is slated for removal. 

(The other Slurm option MEMORY will have its functionality kept, but will transition into a global Ert keyword)

jonathan-eq:
**Issue**
Resolves #8199 

**Approach**
Add slurm queue option MEMORY_PER_CPU to list in config_schema_deprecations.

## When applicable

### DIFF
--- a/src/ert/config/parsing/config_schema_deprecations.py
+++ b/src/ert/config/parsing/config_schema_deprecations.py
@@ -236,4 +236,10 @@ deprecated_keywords_list = [
         "the future. Replace by REALIZATION_MEMORY.",
         check=lambda line: "MEMORY" in line,
     ),
+    DeprecationInfo(
+        keyword="QUEUE_OPTION",
+        message="MEMORY_PER_CPU as QUEUE_OPTION to SLURM is deprecated and will be removed in "
+        "the future. Use REALIZATION_MEMORY instead.",
+        check=lambda line: "MEMORY_PER_CPU" in line,
+    ),
 ]


berland:
Depends on #8215 .

src/ert/config/parsing/config_schema_deprecations.py
```diff
@@ -224,4 +224,10 @@
         "the future. Replace by NUM_CPU on a single compute node.",
         check=lambda line: "NUM_NODES" in line,
     ),
+    DeprecationInfo(
+        keyword="QUEUE_OPTION",
+        message="MEMORY_PER_CPU as QUEUE_OPTION to SLURM is deprecated and will removed in "
```
berland, line 229: will removed -> will be removed

src/ert/config/parsing/config_schema_deprecations.py
```diff
@@ -224,4 +224,10 @@
...
+        "the future.",
```
berland, line 230: We should probably have some resolution recommendation in the message.

jonathan-eq:
> Depends on #8215 .

Maybe we should move this to that PR as it is directly related

src/ert/config/parsing/config_schema_deprecations.py
jonathan-eq: I added `Use REALIZATION_MEMORY instead`

berland:
Needs conflict resolution, but I think it should go in.

berland (approved):

PR #8397, opened 2024-08-06
Fix typos in RUNPATH_FILE description

hnformentin:
Try to improve readability.

### DIFF
--- a/docs/reference/configuration/keywords.rst
+++ b/docs/reference/configuration/keywords.rst
@@ -494,11 +494,11 @@ RUNPATH_FILE
 ------------
 .. _runpath_file:
 
-When running workflows based on external scripts it is necessary to 'tell' the
-external script in some way or another were all the realisations are located in
+When running workflows based on external scripts, it is necessary to 'tell' the
+external script where all the realisations are located in
 the filesystem. Since the number of realisations can be quite high this will
-easily overflow the commandline buffer; the solution which is used is therefore
-to let ERT write a regular file which looks like this::
+easily overflow the commandline buffer; the solution used is
+to let ERT write a regular file. It looks like this::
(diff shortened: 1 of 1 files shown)


andreas-el (approved):
💯

docs/reference/configuration/keywords.rst
```diff
@@ -494,11 +494,11 @@ RUNPATH_FILE
...
 .. _runpath_file:
 
-When running workflows based on external scripts it is necessary to 'tell' the
-external script in some way or another were all the realisations are located in
+When running workflows based on external scripts, it is necessary to 'tell' the
+external script where all the realisations are located in
 the filesystem. Since the number of realisations can be quite high this will
-easily overflow the commandline buffer; the solution which is used is therefore
-to let ERT write a regular file which looks like this::
+easily overflow the commandline buffer; the solution used is
```
eivindjahren, line 500:
```suggestion
easily overflow the commandline buffer; the solution is
to let ERT write a file which looks like this::
```

docs/reference/configuration/keywords.rst
```diff
@@ -494,11 +494,11 @@ RUNPATH_FILE
...
+to let ERT write a regular file. It looks like this::
```
eivindjahren, line 501:
```suggestion
easily overflow the commandline buffer; the solution is
to let ERT write a file which looks like this::
```

eivindjahren: Jus a small suggestion :)

hnformentin: Thanks Eivind. I would keep as it is just because I am too lazy to make a rebase on web version of Github :-)

PR #8401, opened 2024-08-06
Solve and enforce ruff-unused-variable

berland:
**Issue**
Resolves unused variables in the source.

**Approach**
Replace with `_` to indicate to code readers that something is not to be used.

For the gui function, change the return statement to not return unusused stuff. The function is internal to the file.

## When applicable

### DIFF
--- a/tests/unit_tests/shared/test_port_handler.py
+++ b/tests/unit_tests/shared/test_port_handler.py
@@ -58,7 +58,7 @@ def test_find_available_port(unused_tcp_port):
 
 def test_find_available_port_forced(unused_tcp_port):
     custom_range = range(unused_tcp_port, unused_tcp_port)
-    host, port, sock = port_handler.find_available_port(
+    _, port, sock = port_handler.find_available_port(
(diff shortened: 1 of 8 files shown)


tests/unit_tests/shared/test_port_handler.py
```diff
@@ -335,7 +335,7 @@ def test_reuse_active_live_nok_nok(unused_tcp_port):
         )
 
     with pytest.raises(port_handler.NoPortsInRangeException):
-        _, port, sock = port_handler.find_available_port(
+        _, port, _ = port_handler.find_available_port(
```
jonathan-eq, line 338: I think we should rewrite this port_handler method. Shouldn't `find_available_port` only return the available port? I will create an issue for this.

jonathan-eq (approved):
Good improvement!

tests/unit_tests/shared/test_port_handler.py
jonathan-eq: https://github.com/equinor/ert/issues/8407

PR #8416, opened 2024-08-07
Make forward_model_ok an async function

### LINKED ISSUE #7275: Make callback.py async
When the legacy jobqueue and the associated C-drivers are removed, make `forward_model_ok` `callbacks.py` an async function in order to mitigate the downsides of running data internalization as sync code. Yield to async as often as possible in this file. 

This was done in https://github.com/equinor/ert/commit/a767ad9c19f5488032825e4df38cdcdd5a39db91 with measurable positive effect.

berland:
forward_model_ok() is in any case called in an async context, but it takes a long time to complete, too long to be sensible in an async context making the async loop non-responsive as long as it runs.

By making it async, and including some asyncio.sleep(0) we will effectively split up its sync part into smaller parts and make the async loop more responsive than before.

This commit only mitigates the problem, it does not solve it. A better solution would to run the IO-bound loading parts into async IO, and the CPU-bound tasks should be offloaded to different processes to avoid Pythons GIL.

**Issue**
Resolves #7275 

**Approach**
Adding asyncio.sleep(0)

## When applicable

### DIFF
--- a/tests/unit_tests/scheduler/test_job.py
+++ b/tests/unit_tests/scheduler/test_job.py
@@ -12,7 +12,7 @@ from lxml import etree
 
 import ert
 from ert.ensemble_evaluator import Realization
-from ert.load_status import LoadResult, LoadStatus
+from ert.load_status import LoadStatus
 from ert.run_arg import RunArg
 from ert.run_models.base_run_model import captured_logs
 from ert.scheduler import Scheduler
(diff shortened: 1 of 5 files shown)


tests/unit_tests/scheduler/test_job.py
```diff
@@ -110,22 +110,17 @@ async def test_job_run_sends_expected_events(
...
     monkeypatch.setattr(
         scheduler.driver,
         "read_stdout_and_stderr_files",
         lambda *args: "",
     )
 
-    scheduler.job.forward_model_ok = MagicMock()
-    scheduler.job.forward_model_ok.return_value = LoadResult(
-        forward_model_ok_result, ""
-    )
```
berland, line 128: These lines did nothing, they manipulated the wrong version of the object, so thus deleted.

sondreso (approved):

PR #8417, opened 2024-08-07
Try to mitigate flakiness in driver test

berland:
If the echo command takes more than 0.1 sec, this could yield flakiness, so change the logic to avoid that pitfall.

**Issue**
Could theoretically solve the flakiness in https://github.com/equinor/ert/actions/runs/10284734597/job/28461458673

**Approach**
🧠 

## When applicable


tests/integration_tests/scheduler/test_generic_driver.py
```diff
@@ -123,27 +123,26 @@ async def finished(iens, returncode):
 @pytest.mark.flaky(reruns=5)
```
xjules, line 123: Don't you want to remove the `(reruns=5)` thingy then?

berland: No, there is still residual flakyness.

xjules (approved):
Let's ship it! :+1:

PR #8422, opened 2024-08-08
`find_available_port` is available via `__init__` [Issue #8407]

### LINKED ISSUE #8407: Rename `ert.shared.port_handler` and its functions
**Describe the improvement**
The ert.shared.port_handler module is badly named and should renamed to avoid confusion. This module has a lot of helper functions for dealing with sockets and network connections, but none are for only ports. 

Also, the `find_available_port()` function returns not only the port, but also the host address and the socket. Can we rename this to something like `setup_connection()`, and investigate if the port and host are actually being used directly, or if they can be combined to a url right off the bat?

muddi900:
**Issue**
Resolves #8407

**Approach**
As discussed in the issue, exposing all externally used functions makes renaming the module moot.

### DIFF
--- a/src/ert/shared/__init__.py
+++ b/src/ert/shared/__init__.py
@@ -16,6 +16,6 @@ def ert_share_path() -> str:
     return str(Path(spec_origin).parent.parent / "resources")
 
 
-from .port_handler import get_machine_name
+from .port_handler import get_machine_name, find_available_port
 
-__all__ = ["ert_share_path", "get_machine_name", "__version__"]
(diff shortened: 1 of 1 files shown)


sondreso:
Hi! Thank you for the interest in our project and for coming with this contributions. It appears the issue was already resolved as part of a bigger refactoring effort [here](https://github.com/equinor/ert/pull/8481/files), so I'll unfortunately close this PR. Apologize for this, but you are more than welcome to contribute to this project in the future!

Issue #7721, opened 2024-04-22
Discuss refactoring run_ensemble_evaluator to avoid using websockets to propagate events


xjules:
Based on this 8c16ed9f6455692f9a5e4db2a20b80aa1038ecf1 it should be possible to get the events from the ensemble evaluator directly via the queue instead websockets.

jonathan-eq:
@xjules Should this issue wait for https://github.com/equinor/ert/issues/6958?

jonathan-eq:
Blocked by #6958

PR #8424, opened 2024-08-08
Use queue to exchange events between scheduler and evaluator

xjules:
**Issue**
~~Resolves https://github.com/equinor/ert/issues/7721~~
Resolves https://github.com/equinor/ert/issues/8485

**Approach**
This PR removes websockets based communication between scheduler and ensemble evaluator. It is replaced by two message queues, which are provided in `LegacyEnsemble.evaluate` function by the ensemble evaluator and passed over to scheduler:

 - `scheduler_queue`: responsible for providing `CloudEvent` (representing realization and driver events) for evaluator
 - `manifest_queue`: responsible for providing `CloudEvent` (representing notification manifest checksum Event) for scheduler

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/_ensemble.py
+++ b/src/ert/ensemble_evaluator/_ensemble.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import logging
 import traceback
 import uuid
@@ -29,12 +30,7 @@ from ert.serialization import evaluator_marshaller
 from ._wait_for_evaluator import wait_for_evaluator
 from .config import EvaluatorServerConfig
 from .identifiers import EVTYPE_ENSEMBLE_FAILED, EVTYPE_ENSEMBLE_STARTED
-from .snapshot import (
-    ForwardModel,
-    RealizationSnapshot,
-    Snapshot,
-    SnapshotDict,
-)
+from .snapshot import ForwardModel, RealizationSnapshot, Snapshot, SnapshotDict
 from .state import (
     ENSEMBLE_STATE_CANCELLED,
     ENSEMBLE_STATE_FAILED,
@@ -191,7 +187,12 @@ class LegacyEnsemble:
 
         return event_builder
 
-    async def evaluate(self, config: EvaluatorServerConfig) -> None:
+    async def evaluate(
+        self,
+        config: EvaluatorServerConfig,
+        scheduler_queue: Optional[asyncio.Queue[CloudEvent]] = None,
+        manifest_queue: Optional[asyncio.Queue[CloudEvent]] = None,
+    ) -> None:
         self._config = config
         ce_unary_send_method_name = "_ce_unary_send"
         setattr(
@@ -210,13 +211,17 @@ class LegacyEnsemble:
             cert=self._config.cert,
         )
         await self._evaluate_inner(
-            cloudevent_unary_send=getattr(self, ce_unary_send_method_name)
+            cloudevent_unary_send=getattr(self, ce_unary_send_method_name),
+            scheduler_queue=scheduler_queue,
+            manifest_queue=manifest_queue,
         )
 
     async def _evaluate_inner(  # pylint: disable=too-many-branches
         self,
         cloudevent_unary_send: Callable[[CloudEvent], Awaitable[None]],
         experiment_id: Optional[str] = None,
+        scheduler_queue: Optional[asyncio.Queue[CloudEvent]] = None,
+        manifest_queue: Optional[asyncio.Queue[CloudEvent]] = None,
     ) -> None:
         """
         This (inner) coroutine does the actual work of evaluating the ensemble. It
@@ -242,6 +247,8 @@ class LegacyEnsemble:
             self._scheduler = Scheduler(
                 driver,
                 self.active_reals,
+                manifest_queue,
+                scheduler_queue,
                 max_submit=self._queue_config.max_submit,
                 max_running=self._queue_config.max_running,
                 submit_sleep=self._queue_config.submit_sleep,
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -1,6 +1,7 @@
 import asyncio
 import logging
 import pickle
+import traceback
 from contextlib import asynccontextmanager, contextmanager
 from http import HTTPStatus
 from typing import (
@@ -68,6 +69,7 @@ class EnsembleEvaluator:
 
         self._events: asyncio.Queue[CloudEvent] = asyncio.Queue()
         self._messages_to_send: asyncio.Queue[str] = asyncio.Queue()
+        self._manifest_queue: asyncio.Queue[Any] = asyncio.Queue()
 
         self._result = None
 
@@ -311,9 +313,11 @@ class EnsembleEvaluator:
             },
             {event["run_path"]: event.data},
         )
+        # clients still need to receive events via ws
         await self._messages_to_send.put(
             to_json(forward_event, data_marshaller=evaluator_marshaller).decode()
         )
+        await self._manifest_queue.put(forward_event)
 
     async def connection_handler(
         self, websocket: WebSocketServerProtocol, path: str
@@ -417,10 +421,13 @@ class EnsembleEvaluator:
         ]
         # now we wait for the server to actually start
         await self._server_started.wait()
-        # let's run
+
         self._ee_tasks.append(
             asyncio.create_task(
-                self._ensemble.evaluate(self._config), name="ensemble_task"
+                self._ensemble.evaluate(
+                    self._config, self._events, self._manifest_queue
+                ),
+                name="ensemble_task",
             )
         )
 
@@ -433,7 +440,17 @@ class EnsembleEvaluator:
             )
             for task in done:
                 if task_exception := task.exception():
-                    logger.error((f"Exception in evaluator task: {task_exception}"))
+                    exc_traceback = "".join(
+                        traceback.format_exception(
+                            None, task_exception, task_exception.__traceback__
+                        )
+                    )
+                    logger.error(
+                        (
+                            f"Exception in evaluator task {task.get_name()}: {task_exception}\n"
+                            f"Traceback: {exc_traceback}"
+                        )
+                    )
                     raise task_exception
                 elif task.get_name() == "server_task":
                     return
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -10,7 +10,6 @@ from enum import Enum
 from pathlib import Path
 from typing import TYPE_CHECKING, Dict, List, Optional
 
-from cloudevents.conversion import to_json
 from cloudevents.http import CloudEvent
 from lxml import etree
 
@@ -146,7 +145,7 @@ class Job:
                 break
 
             if self.returncode.result() == 0:
-                if self._scheduler.wait_for_checksum():
+                if self._scheduler._manifest_queue is not None:
                     await self._verify_checksum()
                 await self._handle_finished_forward_model()
                 break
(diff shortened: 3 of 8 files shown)


src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -417,6 +421,9 @@ async def _start_running(self) -> None:
         ]
         # now we wait for the server to actually start
         await self._server_started.wait()
+        # setup message queues
+        self._ensemble.scheduler_queue = self._events
+        self._ensemble.manifest_queue = self._manifest_queue
```
jonathan-eq, line 426: Should this be moved to the constructor?

src/ert/ensemble_evaluator/_ensemble.py
```diff
@@ -125,6 +120,9 @@ def __post_init__(self) -> None:
         else:
             self._status_tracker = _EnsembleStateTracker()
 
+        self.scheduler_queue: Any = None
+        self.manifest_queue: Any = None
```
jonathan-eq, line 124: Do we need two queues, or can we have a general one and specify cloudevent types instead?

src/ert/scheduler/scheduler.py
```diff
@@ -209,79 +187,21 @@ def count_states(self) -> Dict[JobState, int]:
...
-                    else:
-                        await conn.send(event)
-                    event = None
-                    self._events.task_done()
-            except ConnectionClosed:
-                logger.debug("Connection to EnsembleEvalutor went down, reconnecting.")
-                continue
+        while True:
+            event = await self._events.get()
+            await self._ee_queue.put(event)
```
jonathan-eq, line 203: Here we are only moving the events from one queue to the next one. Can we pass `self._ee_queue` to the job/driver instead of `self._events`? Then we could get rid of the `_publisher` task completely

jonathan-eq:
So is the scheduler_queue for communication `scheduler`->`evaluator`, and manifest_queue for `evaluator` -> `scheduler`?

src/ert/ensemble_evaluator/evaluator.py
xjules: No, it can't as these queues are passed by the evaluator.

src/ert/ensemble_evaluator/_ensemble.py
xjules: We need to two queues yes, but can update to use `Queue[CloudEvent]` instead :+1:

src/ert/scheduler/scheduler.py
xjules: Currently it is required, since `_ee_queue` is a consumer a does not need to be set.

xjules:
> So is the scheduler_queue for communication `scheduler`->`evaluator`, and manifest_queue for `evaluator` -> `scheduler`?

Yes, it is described in the PR :smile_cat:

src/ert/ensemble_evaluator/_ensemble.py
```diff
@@ -125,6 +120,9 @@ def __post_init__(self) -> None:
         else:
             self._status_tracker = _EnsembleStateTracker()
 
+        self.scheduler_queue: Any = None
```
berland, line 123: Can't we type this stricter than `Any`

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -1,6 +1,7 @@
 import asyncio
 import logging
 import pickle
+import traceback
```
berland, line 4: debug leftover

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -311,9 +313,11 @@ async def forward_checksum(self, event: CloudEvent) -> None:
             },
             {event["run_path"]: event.data},
         )
+        # currently clients still need to receive events via ws
```
berland, line 316: ?

berland: (what clients?)

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -417,6 +421,9 @@ async def _start_running(self) -> None:
         ]
         # now we wait for the server to actually start
         await self._server_started.wait()
+        # setup message queues
+        self._ensemble.scheduler_queue = self._events
```
berland, line 425: Should we rename `self._events` to something more precise?

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -433,7 +440,17 @@ async def _monitor_and_handle_tasks(self) -> None:
...
+                            None, task_exception, task_exception.__traceback__
+                        )
+                    )
+                    logger.error(
+                        (
+                            f"Exception in evaluator task {task.get_name()}: {task_exception}\n"
+                            f"Traceback: {exc_traceback}"
+                        )
+                    )
                     raise task_exception
```
berland, line 454: tre traceback is not printed by this latter `raise`?

src/ert/scheduler/job.py
```diff
@@ -164,7 +163,7 @@ async def run(self, sem: asyncio.BoundedSemaphore, max_submit: int = 1) -> None:
                 break
 
             if self.returncode.result() == 0:
-                if self._scheduler.wait_for_checksum():
+                if self._scheduler._manifest_queue:
```
berland, line 166: at first sight, this looks like there can be a bug introduced when the manifest queue is not yet populated, then it will skip it?

src/ert/scheduler/scheduler.py
```diff
@@ -92,6 +85,9 @@ def __init__(
         ee_token: Optional[str] = None,
```
berland, line 85: These `ee_*` class variables can be removed I suppose.

src/ert/scheduler/scheduler.py
```diff
@@ -92,6 +85,9 @@ def __init__(
...
     ) -> None:
         self.driver = driver
+        self._ee_queue = ee_queue
```
berland, line 88: I am not overly happy with the `ee` shortcut. Should we just spell out `_ensemble_evaluator_queue instead?

src/ert/scheduler/scheduler.py
```diff
@@ -319,19 +239,9 @@ async def _monitor_and_handle_tasks(
...
-                        )
-                        await asyncio.wait_for(
-                            self._events.join(), timeout=self._queue_timeout
-                        )
-                    except asyncio.TimeoutError:
-                        logger.error(
-                            f"{self._events.qsize()} items left unprocessed in the queue!"
-                        )
+                if self._ee_queue:
+                    # only join queue if there is a consumer
```
berland, line 243: I find "join" to be a strange word for what is happening here, so in a comment I would write something about waiting.

src/ert/ensemble_evaluator/evaluator.py
xjules: traceback is now used to log more info in case something crashed in the evaluator tasks.

src/ert/ensemble_evaluator/evaluator.py
xjules: yes, we can :)

src/ert/ensemble_evaluator/evaluator.py
xjules: Unfortunately no. Have been testing it (crashing tasks) and traceback needs to be there.

src/ert/scheduler/scheduler.py
xjules: yes, but `_update_jobs_json` needs to be moved to (based_run_model), which I thought could be a different PR?

src/ert/scheduler/scheduler.py
xjules: will do :+1:

src/ert/ensemble_evaluator/evaluator.py
xjules: Update: I've kept the `self._events` but made `Ensemble` stateless wrt. message queues. They are sent as attribute in `ensemble.evaluate` function.

src/ert/ensemble_evaluator/evaluator.py
berland: the forward_model_runner is still intending to use websocket for the communication of checksum events. Then it sounds odd to use the word "currently"?

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -417,10 +421,14 @@ async def _start_running(self) -> None:
         ]
         # now we wait for the server to actually start
         await self._server_started.wait()
-        # let's run
+
+        # let's run with message queues set
```
berland, line 425: that comment should be redundant

src/ert/scheduler/job.py
berland: I think I misread the code. Should we write `if self._scheduler._manifest_queue is not None` to be more explicit? (I thought we were testing whether there is anything in the queue)

src/ert/scheduler/scheduler.py
```diff
@@ -208,79 +188,21 @@ def count_states(self) -> Dict[JobState, int]:
         return counts
 
     async def _checksum_consumer(self) -> None:
-        if not self._ee_uri:
+        if not self._manifest_queue:
```
berland, line 191: `if self._manifest_queue is None`

src/ert/scheduler/scheduler.py
```diff
@@ -208,79 +188,21 @@ def count_states(self) -> Dict[JobState, int]:
...
-                )
+        while True:
+            event = await self._manifest_queue.get()
+            if event["type"] == EVTYPE_FORWARD_MODEL_CHECKSUM:
+                self.checksum.update(event.data)
+            self._manifest_queue.task_done()
 
     async def _publisher(self) -> None:
-        if not self._ee_uri:
+        if not self._ensemble_evaluator_queue:
```
berland, line 200: check for None-ness

src/ert/scheduler/scheduler.py
```diff
@@ -318,19 +240,10 @@ async def _monitor_and_handle_tasks(
...
-                            self._publisher_done.wait(), timeout=self._queue_timeout
-                        )
-                        await asyncio.wait_for(
-                            self._events.join(), timeout=self._queue_timeout
-                        )
-                    except asyncio.TimeoutError:
-                        logger.error(
-                            f"{self._events.qsize()} items left unprocessed in the queue!"
-                        )
+                if self._ensemble_evaluator_queue:
```
berland, line 243: is not None

src/ert/scheduler/scheduler.py
```diff
@@ -342,16 +255,14 @@ async def execute(
...
         min_required_realizations: int = 0,
     ) -> str:
-        listener_task = await self.start_manifest_listener()
         scheduling_tasks = [
             asyncio.create_task(self._publisher(), name="publisher_task"),
             asyncio.create_task(
                 self._process_event_queue(), name="process_event_queue_task"
             ),
             asyncio.create_task(self.driver.poll(), name="poll_task"),
+            asyncio.create_task(self._checksum_consumer(), name="consumer_task"),
```
berland, line 264: name="checksum_consumer_task"

src/ert/scheduler/scheduler.py
xjules: I've checked and queue, even created but empty will be fine with this if.

src/ert/ensemble_evaluator/evaluator.py
xjules: right :smile: that was more emotional comment than helpful really

src/ert/scheduler/scheduler.py
berland: for readability

src/ert/ensemble_evaluator/evaluator.py
xjules: Here, particularly, it has nothing to do with job_dispatch, but it is kept only for the sake of test. As this one only informs the test (described above) that the checksum has been received, since scheduler has not the manifest_queue.

berland (approved):
yay!

Issue #6712, opened 2023-11-29
Implement Pydantic model for queue options


kwinkunks:
We are already validating the strings in the config parser, but it would be nice to:

1. Cast values to their proper types (eg numbers as numbers, since we're in Python now they can be implicitly cast to strings when we compose the `bsub` command or whatever)
2. Apply default values in an obvious and maintainable way.
3. Replace the current list of tuples containing the options with a dataclass (say).

Then we can avoid having to do things like this in `JobQueue` in `queue.py`:

```python
def max_running(self) -> int:
    max_running = 0
    if self.driver.has_option("MAX_RUNNING"):
        max_running = int(self.driver.get_option("MAX_RUNNING"))
    if max_running == 0:
        return len(self._realizations)
    return max_running
```

pinkwah:
Blocked by C JobQueue

PR #8426, opened 2024-08-08
Use a plain dict for queue system options

berland:
This was previously a list of 2-tuples of strings, owing to compatibility with the C drivers which are now purged

**Issue**
Resolves partially #6712 

**Approach**
Switch list of 2-tuples of str with a dict.

## When applicable

### DIFF
--- a/src/ert/config/queue_config.py
+++ b/src/ert/config/queue_config.py
@@ -6,7 +6,7 @@ import shutil
 import warnings
 from collections import defaultdict
 from dataclasses import dataclass, field
-from typing import Any, Dict, List, Mapping, Tuple, no_type_check
+from typing import Any, Dict, List, Mapping, no_type_check
 
 from .parsing import (
     ConfigDict,
@@ -173,9 +173,7 @@ class QueueConfig:
     realization_memory: int = 0
     max_submit: int = 1
     queue_system: QueueSystem = QueueSystem.LOCAL
-    queue_options: Dict[QueueSystem, List[Tuple[str, str]]] = field(
-        default_factory=dict
-    )
+    queue_options: Dict[QueueSystem, Dict[str, str]] = field(default_factory=dict)
     stop_long_running: bool = False
 
     @no_type_check
@@ -192,7 +190,7 @@ class QueueConfig:
         )
         max_submit: int = config_dict.get(ConfigKeys.MAX_SUBMIT, 1)
         stop_long_running = config_dict.get(ConfigKeys.STOP_LONG_RUNNING, False)
-        queue_options: Dict[QueueSystem, List[Tuple[str, str]]] = defaultdict(list)
+        queue_options: Dict[QueueSystem, Dict[str, str]] = defaultdict(dict)
         for queue_system, option_name, *values in config_dict.get("QUEUE_OPTION", []):
             if queue_system == QueueSystem.GENERIC:
                 queue_system = selected_queue_system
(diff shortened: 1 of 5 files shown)


src/ert/config/queue_config.py
```diff
@@ -202,25 +200,29 @@ def from_dict(cls, config_dict: ConfigDict) -> QueueConfig:
...
 
-            queue_options[queue_system].append(
-                (option_name, values[0] if values else "")
-            )
+            value = values[0] if values else ""
+            if (
+                option_name in queue_options[queue_system]
+                and queue_options[queue_system][option_name] != value
+            ):
+                logging.info(
```
jonathan-eq, line 208: Should we also log if there are duplicate options with the same value?

src/ert/config/queue_config.py
```diff
@@ -245,88 +247,52 @@ def create_local_copy(self) -> QueueConfig:
         )
 
     @property
-    def max_running(self) -> int:
-        max_running = 0
-        for key, val in self.queue_options.get(self.queue_system, []):
-            if key == "MAX_RUNNING":
-                max_running = int(val)
-        return max_running
+    def selected_queue_options(self) -> Dict[str, str]:
```
jonathan-eq, line 250: Is it worth creating a type alias for `QueueOption`?
`QueueOption = Literal['QUEUE_SYSTEM', 'REALIZATION_MEMORY', 'MAX_SUBMIT']`

src/ert/config/queue_config.py
```diff
@@ -245,88 +247,52 @@ def create_local_copy(self) -> QueueConfig:
...
-    if not num_nodes_str and not num_cpus_per_node_str:
-        # NUM_CPU will take precedence when deprecated options are not set
+    if (
+        "NUM_NODES" not in queue_system_options
+        and "NUM_CPUS_PER_NODE" not in queue_system_options
+    ):
         return
-    num_nodes = int(num_nodes_str) if num_nodes_str else 1
-    num_cpus_per_node = int(num_cpus_per_node_str) if num_cpus_per_node_str else 1
+    num_nodes = int(queue_system_options.get("NUM_NODES", 1) or "1")
```
jonathan-eq, line 294: Do we need the extra `... or "1"` here?

src/ert/config/queue_config.py
```diff
@@ -245,88 +247,52 @@ def create_local_copy(self) -> QueueConfig:
...
+    num_cpus_per_node = int(queue_system_options.get("NUM_CPUS_PER_NODE", 1) or "1")
```
jonathan-eq, line 295: Same here

src/ert/config/queue_config.py
```diff
@@ -449,11 +415,10 @@ def _validate_queue_driver_settings(
 
 @no_type_check
```
jonathan-eq, line 416: Is it a lot of work to get rid of the `@no_type_check` decorator?

jonathan-eq (changes requested):
Very nice! See minor nitpicking 🔍

src/ert/config/queue_config.py
berland: Hmm, does not sound critical, it is more relevant for the users? (anyway, not in this pr)

src/ert/config/queue_config.py
berland: Yes, because it is (currently) valid configuriation to supply an empty string for NUM_NODES

src/ert/config/queue_config.py
berland: same answer..

src/ert/config/queue_config.py
berland: I think so (and it seems so..) :(

src/ert/config/queue_config.py
jonathan-eq: Why is that valid?

jonathan-eq (approved):
LGTM!

src/ert/config/queue_config.py
berland: Not sure, could be a legacy from the C-driver. We can change it in the future (but separate PR).

berland: Maybe all I know is that it will fail current tests if support for empty strings is removed.

PR #8439, opened 2024-08-12
Upgrade pre-commits, ruff 0.5.0 -> 0.6.1

larsevj:
Some complaints about ASYNC rules in the new ruff version, I could not see that any of them would be usefull to us so ignored them, but I do not have much experience with async so I might have missed something.

## When applicable


src/ert/ensemble_evaluator/_wait_for_evaluator.py
```diff
@@ -42,7 +42,7 @@ async def wait_for_evaluator(
     token: Optional[str] = None,
     cert: Optional[Union[str, bytes]] = None,
     healthcheck_endpoint: str = "/healthcheck",
-    timeout: Optional[float] = None,
+    timeout: Optional[float] = None,  # noqa: ASYNC109
```
berland, line 45: Can you add this exception in `pyproject.toml` instead, makes it easer to know that it is something to fix.

berland: No, just keep them as is, they all seems warranted enough.

PR #8446, opened 2024-08-13
Mitigate specific LSF test flakyness

berland:
On local hardware the test is more aggressive to allow faster test times, implying some flakyness when local hardware is bogged down in tasks. That residual flakyness we can solve better with reruns instead of increasing the wait times in the test.

**Issue**
Resolves  locally observed flakyness.

**Approach**
Rerun ♻️ 

## When applicable

### DIFF
--- a/tests/integration_tests/scheduler/test_lsf_driver.py
+++ b/tests/integration_tests/scheduler/test_lsf_driver.py
@@ -254,12 +254,22 @@ async def test_polling_bhist_fallback(not_found_bjobs, caplog, job_name):
     await poll(driver, {0})
     assert "bhist is used" in caplog.text
     assert bhist_called
-    assert job_id in driver._bhist_cache
+    assert driver._bhist_cache and job_id in driver._bhist_cache
 
 
-async def test_kill_before_submit_is_finished(
+@pytest.mark.flaky(rerun=10)
+async def test_that_kill_before_submit_is_finished_works(
     tmp_path, monkeypatch, caplog, pytestconfig
 ):
+    """This test asserts that it is possible to issue a kill command
+    to a realization right after it has been submitted (as in driver.submit()).
+
+    The bug intended to catch is if the driver gives up on killing before submission
+    is not done yet, it is important not to let the realization through in that scenario.
+
+    The design of the test alludes to much more flakyness than what is probable in reality,
+    thus reruns are allowed to make this pass.
+    """
     os.chdir(tmp_path)
 
     if pytestconfig.getoption("lsf"):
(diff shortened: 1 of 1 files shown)


tests/integration_tests/scheduler/test_lsf_driver.py
```diff
@@ -257,6 +257,7 @@ def mock_poll_once_by_bhist(*args, **kwargs):
     assert job_id in driver._bhist_cache
 
 
+@pytest.mark.flaky(reruns=5)
```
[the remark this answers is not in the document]
berland: We could add the marker programmatically in `conftest.py` where we check for test function name and avoid `darwin` and `--lsf`. Not sure that extra test code noise is worth it.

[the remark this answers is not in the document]
sondreso: Agreed, just a bit worried that we might introduce flakyness in the actual code en then the tests will not catch it :thinking:

tests/integration_tests/scheduler/test_lsf_driver.py
```diff
@@ -271,6 +272,7 @@ async def test_kill_before_submit_is_finished(
         job_kill_window = 5
         test_grace_time = 10
     else:
+        # More aggressive on local iron, fixing residual flakyness with reruns
```
xjules, line 275: As discussed let's try to replace `await asyncio.sleep(0.01)` with the actual lock being set; ie. `self._submit_locks[0]`

berland: Turns out it that is not enough to avoid flakyness

berland:
The flakyness is easily reproducible by adding  `@pytest.mark.repeat(100)` to the offending test, and then running: `pytest tests/integration_tests/ -sxv -n 12 `.

Awaiting the submit locks does not help, as there are more sources of flakiness. E.g. if the execution of the kill itself is too slow, the first process will survive no matter how.  Can't even remove flakyness by increasing job_kill_window and test_grace_time.

berland:
Added a new proposal which is not pretty, but works.

tests/integration_tests/scheduler/test_lsf_driver.py
```diff
@@ -271,8 +281,8 @@ async def test_kill_before_submit_is_finished(
         job_kill_window = 5
         test_grace_time = 10
     else:
-        job_kill_window = 1
-        test_grace_time = 2
+        job_kill_window = 2
```
xjules, line 284: That is not really a fix :smile:

xjules (approved):
Alight! Let's see how long this will last :rocket:

PR #8454, opened 2024-08-13
Remove SnapshotDict

### LINKED ISSUE #1869: `Snapshot` and `SnapshotDict` makes for a confusing API
**Describe the bug**
Basically, a `Snapshot` can be updated, a `SnapshotDict` cannot. Interacting with a `Snapshot` is hard (basically a bunch of `PMap`s. Interacting with `SnapshotDict` is somewhat easier, but you still need to use a lot of `.items()` and `.values()` and so on.

**Expected behavior**
Both the consumer and the evaluator needs to keep track of state in a `Snapshot`, so maybe this should be the primary class to represent the snapshot, but then it needs to be easier to interact with.

If a `SnapshotDict` became as able as the `Snapshot`, maybe _it_ could be the primary representation, but then maybe the use of `pydantic` needs to be deprecated, since the GUI struggles with it? And we don't really use any of the `pydantic` API, so perhaps it was a premature optimization?

jonathan-eq:
**Issue**
Resolves #1869 

**Approach**

## When applicable

### DIFF
--- a/src/ert/cli/monitor.py
+++ b/src/ert/cli/monitor.py
@@ -106,7 +106,7 @@ class Monitor:
         failed_jobs: Dict[Optional[str], int] = {}
         for snapshot in self._snapshots.values():
             for real in snapshot.reals.values():
-                for job in real.forward_models.values():
+                for job in real["forward_models"].values():
                     if job.get(ids.STATUS) == FORWARD_MODEL_STATE_FAILURE:
                         err = job.get(ids.ERROR)
                         result = failed_jobs.get(err, 0)
@@ -120,7 +120,7 @@ class Monitor:
         total_count = len(latest_snapshot.reals)
         aggregate = latest_snapshot.aggregate_real_states()
         for state_ in ALL_REALIZATION_STATES:
-            count = aggregate.get(state_, 0)
+            count = aggregate[state_]
             _countstring = f"{count}/{total_count}"
             out = (
                 f"{self._colorize(self.dot, color=REAL_STATE_TO_COLOR[state_])}"
--- a/src/ert/ensemble_evaluator/_ensemble.py
+++ b/src/ert/ensemble_evaluator/_ensemble.py
@@ -25,7 +25,11 @@ from ert.scheduler import Scheduler, create_driver
 
 from ._wait_for_evaluator import wait_for_evaluator
 from .config import EvaluatorServerConfig
-from .snapshot import ForwardModel, RealizationSnapshot, Snapshot, SnapshotDict
+from .snapshot import (
+    ForwardModel,
+    RealizationSnapshot,
+    Snapshot,
+)
 from .state import (
     ENSEMBLE_STATE_CANCELLED,
     ENSEMBLE_STATE_FAILED,
@@ -120,25 +124,20 @@ class LegacyEnsemble:
         return list(filter(lambda real: real.active, self.reals))
 
     def _create_snapshot(self) -> Snapshot:
-        reals: Dict[str, RealizationSnapshot] = {}
+        snapshot = Snapshot()
+        snapshot._ensemble_state = ENSEMBLE_STATE_UNKNOWN
         for real in self.active_reals:
-            reals[str(real.iens)] = RealizationSnapshot(
-                active=True,
-                status=REALIZATION_STATE_WAITING,
+            realization = RealizationSnapshot(
+                active=True, status=REALIZATION_STATE_WAITING, forward_models={}
             )
             for index, forward_model in enumerate(real.forward_models):
-                reals[str(real.iens)].forward_models[str(index)] = ForwardModel(
+                realization["forward_models"][str(index)] = ForwardModel(
                     status=FORWARD_MODEL_STATE_START,
                     index=str(index),
                     name=forward_model.name,
                 )
-        top = SnapshotDict(
-            reals=reals,
-            status=ENSEMBLE_STATE_UNKNOWN,
-            metadata=self.metadata,
-        )
-
-        return Snapshot.from_nested_dict(top.model_dump())
+            snapshot.add_realization(str(real.iens), realization)
+        return snapshot
 
     def get_successful_realizations(self) -> List[int]:
         return self.snapshot.get_successful_realizations()
--- a/src/ert/ensemble_evaluator/snapshot.py
+++ b/src/ert/ensemble_evaluator/snapshot.py
@@ -4,19 +4,19 @@ from collections import defaultdict
 from datetime import datetime
 from typing import (
     Any,
+    Counter,
     DefaultDict,
     Dict,
     List,
     Mapping,
     Optional,
-    Sequence,
     Tuple,
+    TypeVar,
     Union,
     cast,
     get_args,
 )
 
-from pydantic import BaseModel
 from qtpy.QtGui import QColor
 from typing_extensions import TypedDict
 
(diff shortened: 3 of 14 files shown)


src/ert/cli/monitor.py
```diff
@@ -120,7 +120,7 @@ def _get_legends(self) -> str:
         total_count = len(latest_snapshot.reals)
         aggregate = latest_snapshot.aggregate_real_states()
         for state_ in ALL_REALIZATION_STATES:
-            count = aggregate.get(state_, 0)
+            count = aggregate[state_]
```
jonathan-eq, line 123: This is now a Counter and not a dict

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -397,68 +393,41 @@ class ForwardModel(TypedDict, total=False):
...
+        end_time=realization_data.get("end_time"),
+    )
+
+    realization_snapshot["forward_models"] = {}
+
+    for fm_id, fm_data in realization_data.get("forward_models", {}).items():
+        realization_snapshot["forward_models"][fm_id] = fm_data
+        if "current_memory_usage" in realization_snapshot["forward_models"][fm_id]:
+            realization_snapshot["forward_models"][fm_id]["current_memory_usage"] = str(
+                realization_snapshot["forward_models"][fm_id]["current_memory_usage"]
```
jonathan-eq, line 420: There are tests passing `...memory_usage` as ints, and for some reason we are expecting it to be strings. This caused pydantic to blow up.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -397,68 +393,41 @@ class ForwardModel(TypedDict, total=False):
...
-                    ids.NAME: name,
-                    ids.STDOUT: stdout,
-                    ids.STDERR: stderr,
-                    ids.CURRENT_MEMORY_USAGE: current_memory_usage,
-                    ids.MAX_MEMORY_USAGE: max_memory_usage,
-                }
-            )
-        )
-        return self
+def _filter_nones(input: T) -> T:
```
jonathan-eq, line 432: This method removes the None fields from the TypedDict

tests/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+                start_time=start_time,
+                end_time=end_time,
+                name=name,
+                stdout=stdout,
+                stderr=stderr,
+                current_memory_usage=current_memory_usage,
+                max_memory_usage=max_memory_usage,
+            )
+        )
+        return self
```
jonathan-eq, line 69: This class should go, but it is heavily used in tests. I am happy with moving it to `tests/__init__.py`.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -107,17 +104,16 @@ class Snapshot:
...
-        realization number, pointing to a dict with keys active (bool),
-        start_time (datetime), end_time (datetime) and status (str)."""
-
-        self._forward_model_states: DefaultDict[Tuple[str, str], ForwardModel] = (
-            defaultdict(ForwardModel)  # type: ignore
-        )
+        self._realization_states: DefaultDict[
+            RealId,
+            RealizationSnapshot,
+        ] = defaultdict(RealizationSnapshot)  # type: ignore
```
berland, line 110: check if the `type: ignore` can be dropped.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -219,38 +214,39 @@ def get_forward_model_status_for_all_reals(
...
-    def get_job(self, real_id: str, forward_model_id: str) -> "ForwardModel":
+    def get_job(self, real_id: RealId, forward_model_id: FmStepId) -> "ForwardModel":
         return self._forward_model_states[(real_id, forward_model_id)].copy()
 
     def get_successful_realizations(self) -> typing.List[int]:
         return [
             int(real_idx)
             for real_idx, real_data in self._realization_states.items()
-            if real_data[ids.STATUS] == state.REALIZATION_STATE_FINISHED
+            if real_data["status"] == state.REALIZATION_STATE_FINISHED
```
berland, line 238: why changing from `ids.STATUS` TO `"status"`?

jonathan-eq: I think we should move away from using `ids` altogether in most of the codebase. The benefits are already given to us for free with typed dicts.

src/ert/ensemble_evaluator/_ensemble.py
```diff
@@ -130,25 +129,20 @@ def active_reals(self) -> Sequence[Realization]:
...
-                reals[str(real.iens)].forward_models[str(index)] = ForwardModel(
+                realization["forward_models"][str(index)] = ForwardModel(
                     status=FORWARD_MODEL_STATE_START,
                     index=str(index),
                     name=forward_model.name,
                 )
-        top = SnapshotDict(
-            reals=reals,
-            status=ENSEMBLE_STATE_UNKNOWN,
-            metadata=self.metadata,
```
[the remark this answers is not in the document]
jonathan-eq: It was initialized as empty dict all the way back in base_run_model.py `BaseRunModel._build_ensemble()`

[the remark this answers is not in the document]
berland: I don't quite get it (have not studied the code). In your diff it seems like you are no longer passing on `status` and `metadata`. Is this still kept through some existing code?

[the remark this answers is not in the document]
jonathan-eq: I am setting status directly on the snapshot at the begining of the `_create_snapshot` method, and snapshot metadata already defaults to empty dict, so we don't have to explicitly pass the empty dict from _ensemble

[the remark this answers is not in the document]
berland: ah

src/ert/ensemble_evaluator/snapshot.py
berland: can it?

JHolba: it can't. mypy has problems with defaultdict taking typeddicts as argument. you can write defaultdict(lambda: RealizationSnapshot()), but then pyright will give you warnings that you should write defaultdict(RealizationSnapshot)...

berland (approved):
Fine!

jonathan-eq:
> Fine!

I will wait for @JHolba to run profiling on this and main. If it is not any different, I will revert to using basemodel for RealizationSnapshot 👌

JHolba:
> > Fine!
> 
> I will wait for @JHolba to run profiling on this and main. If it is not any different, I will revert to using basemodel for RealizationSnapshot 👌

Didn't notice any difference in performance between this and main

jonathan-eq:
> > > Fine!
> > 
> > 
> > I will wait for @JHolba to run profiling on this and main. If it is not any different, I will revert to using basemodel for RealizationSnapshot 👌
> 
> Didn't notice any difference in performance between this and main

Okay, I will try removing the extra migration to TypedDict and see if something hits the fan 🪭

jonathan-eq:
> > > > Fine!
> > > 
> > > 
> > > I will wait for @JHolba to run profiling on this and main. If it is not any different, I will revert to using basemodel for RealizationSnapshot 👌
> > 
> > 
> > Didn't notice any difference in performance between this and main
> 
> Okay, I will try removing the extra migration to TypedDict and see if something hits the fan 🪭

Something hit the fan, and it is not looking good  🚫💐

PR #8470, opened 2024-08-15
Add script for generating bigpoly test-data

jonathan-eq:
**Approach**
✏️ 

## When applicable

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,6 @@ src/ert/shared/version.py
 /compile_commands.json
 *_pb2.py
 *_pb2.pyi
+/test-data/bigpoly/*
+!/test-data/bigpoly/make_bigpoly.sh
+!/test-data/bigpoly/generate_eclsum.py
--- a/test-data/bigpoly/generate_eclsum.py
+++ b/test-data/bigpoly/generate_eclsum.py
@@ -1,0 +1,460 @@
+import os
+import sys
+from dataclasses import astuple, dataclass
+from datetime import datetime, timedelta
+from enum import Enum, unique
+from typing import Any, List, Optional, Tuple
+
+try:
+    import resfo as ecl_data_io
+except ImportError:
+    import ecl_data_io
+
+import hypothesis.strategies as st
+import numpy as np
+from hypothesis import HealthCheck, assume, given, settings
+from hypothesis.extra.numpy import from_dtype
+from pydantic import PositiveInt, conint
+from typing_extensions import Self
+
+UNIT_NAMES = st.sampled_from(
+    ["SM3/DAY", "BARSA", "SM3/SM3", "FRACTION", "DAYS", "YEARS", "SM3", "SECONDS"]
+)
+
+NAMES = st.text(
+    min_size=8, max_size=8, alphabet=st.characters(min_codepoint=65, max_codepoint=90)
+)
+
+
+@unique
+class UnitSystem(Enum):
+    METRIC = 1
+    FIELD = 2
+    LAB = 3
+
+    def to_ecl(self):
+        return self.value
+
+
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -34,3 +34,6 @@ src/ert/shared/version.py
 /compile_commands.json
 *_pb2.py
 *_pb2.pyi
+/test-data/bigpoly/*
+!/test-data/bigpoly/make_bigpoly.sh
+!/test-data/bigpoly/generate_eclsum.py
```
berland, line 39: do we need these lines here?

jonathan-eq: I think we should add them for normal poly_example too. It is very easy to commit logs and poly_out by mistake

berland:
Should we have a PR in gkomodo-releases that demonstrates that it is able to build on this PR?

It should get the grdecl script from Ert and generate the remaining queue system tests based on the files dumped by this script.

xjules:
> Should we have a PR in gkomodo-releases that demonstrates that it is able to build on this PR?
> 
> It should get the grdecl script from Ert and generate the remaining queue system tests based on the files dumped by this script.

Yes, I wanted to suggest the same. Also to have a small description how to run / build big poly or is it so simple to run the `make_bigpoly` should be enough?

jonathan-eq:
> > Should we have a PR in gkomodo-releases that demonstrates that it is able to build on this PR?
> > It should get the grdecl script from Ert and generate the remaining queue system tests based on the files dumped by this script.
> 
> Yes, I wanted to suggest the same. Also to have a small description how to run / build big poly or is it so simple to run the `make_bigpoly` should be enough?

I want to add a file `modify_bigpoly.sh` in komodo-releases/
The script will checkout the bigpoly script from ert, run it, then append the other queue_options for cluster

xjules:
After giving `big-poly` some thoughts I've realized that this case / test should remain only in komodo IMHO. Semantically it only serves as a testing detail related to our production environment and as such it does not provide any value to potential users of ert. Any thoughts @sondreso @berland

jonathan-eq:
> After giving `big-poly` some thoughts I've realized that this case / test should remain only in komodo IMHO. Semantically it only serves as a testing detail related to our production environment and as such it does not provide any value to potential users of ert. Any thoughts @sondreso @berland

When refactoring snapshots, it was very useful to have a heavy run, so we had to manually clone over bigpoly from komodo-releases. This was more difficult than it had to be, and most of the bigpoly files created by the script are not for running on cluster anyways. 

Instead of manually cloning the script from komodo-releases and removing the cluster side of it, i think it is better to have the essential part of the script in Ert, and a smaller script for bigpoly_cluster in komodo.

xjules:
Suggestion: Let's create a new issue, which would aim to have such a case in `test-data`, where we should give it a better name (than big-poly) and make a readme, which describes utilization of such a case; ie. stress-testing the messaging system and internationalization.

berland:
It probably belongs in `tests/performance_tests`

xjules:
> It probably belongs in `tests/performance_tests` as a `pytest` test.

do you mean to move the entire big-poly there?

berland:
> > It probably belongs in `tests/performance_tests` as a `pytest` test.
> do you mean to move the entire big-poly there?

Probably a more carefully written test setup that in some way stresses the message system without touching implementation details, with and without some mocked cpu-intensive internalization.

jonathan-eq:
> It probably belongs in `tests/performance_tests` as a `pytest` test.

I don't think so. I mainly use bigpoly for manually testing the GUI. If we want a pytest performance test for it, we can create a fixture and use it the same way we use `copy_poly_example`. It is easier to do it this way than the other way around.

PR #8475, opened 2024-08-15
Refactor queue_options to use Pydantic dataclasses

### LINKED ISSUE #6712: Implement Pydantic model for queue options

berland:
**Issue**
Resolves #6712 

**Approach**
Rewrite

## When applicable

### DIFF
--- a/src/ert/config/queue_config.py
+++ b/src/ert/config/queue_config.py
@@ -3,9 +3,11 @@ from __future__ import annotations
 import logging
 import re
 import shutil
-from collections import defaultdict
-from dataclasses import dataclass, field
-from typing import Any, Dict, List, Mapping, no_type_check
+from dataclasses import asdict, dataclass, field, fields
+from typing import Any, Dict, List, Mapping, Optional, no_type_check
+
+import pydantic
+from typing_extensions import Annotated
 
 from .parsing import (
     ConfigDict,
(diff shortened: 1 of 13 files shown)


src/ert/scheduler/__init__.py
```diff
@@ -22,50 +24,39 @@ def create_driver(config: QueueConfig) -> Driver:
...
-            num_cpus_per_node=int(num_cpus_per_node) if num_cpus_per_node else None,
-            cluster_label=queue_options.get("CLUSTER_LABEL"),
-            job_prefix=queue_options.get("JOB_PREFIX"),
-            project_code=queue_options.get("PROJECT_CODE"),
-        )
+        queue_dict = asdict(config.queue_options)
+        queue_dict["queue_name"] = queue_dict.pop("queue")
+        queue_dict.pop("max_running")
+        queue_dict.pop("submit_sleep")
+        return OpenPBSDriver(**queue_dict)
```
berland, line 31: Should it be a helper function in the TorqueOptions class to provide this argument dictionary for all drivers?

src/ert/config/queue_config.py
```diff
@@ -17,123 +19,167 @@
...
-        "PARTITION",
-        "INCLUDE_HOST",
-        "EXCLUDE_HOST",
-        "PROJECT_CODE",
-    ],
-    "TORQUE": [
-        "QSUB_CMD",
-        "QSTAT_CMD",
-        "QDEL_CMD",
-        "QSTAT_OPTIONS",
```
berland, line 92: This has been removed in the PR.

src/ert/config/queue_config.py
```diff
@@ -17,123 +19,167 @@
...
-        "NUM_CPUS_PER_NODE",
-        "MAX_RUNNING",
-    ],
-    "LOCAL": ["MAX_RUNNING"],
-    "GENERIC": ["MAX_RUNNING"],
-}
-queue_positive_number_options: Mapping[str, List[str]] = {
-    "LSF": ["SUBMIT_SLEEP"],
-    "SLURM": ["SUBMIT_SLEEP", "SQUEUE_TIMEOUT", "MAX_RUNTIME"],
-    "TORQUE": ["SUBMIT_SLEEP", "QUEUE_QUERY_TIMEOUT"],
```
berland, line 119: queue_query_timeout is missing in PR

src/ert/scheduler/__init__.py
berland: Now included as a bonus commit (to be squashed)

src/ert/config/queue_config.py
berland: added

src/ert/config/queue_config.py
berland: added.

src/ert/config/queue_config.py
```diff
@@ -17,123 +19,210 @@
...
-    "MEMORY",
-    "MEMORY_PER_CPU",
-    "PARTITION",
-    "PROJECT_CODE",
-    "SBATCH",
-    "SCANCEL",
-    "SCONTROL",
-    "SQUEUE",
-    "SQUEUE_TIMEOUT",
+_all_queue_systems = [
```
berland, line 22: Unneeded, can just iterate over QueueSystem

src/ert/scheduler/slurm_driver.py
```diff
@@ -73,8 +73,8 @@ def __init__(
         realization_memory: Optional[int] = 0,
         queue_name: Optional[str] = None,
         memory_per_cpu: Optional[str] = None,
-        max_runtime: Optional[str] = None,
-        squeue_timeout: Optional[str] = None,
+        max_runtime: Optional[float] = None,
+        squeue_timeout: float = 2,
```
xjules, line 77: Is this value ever to be changed? ie. why to expose this to the users?

src/ert/config/queue_config.py
```diff
@@ -17,123 +19,159 @@
...
+                    error["input"],
+                    is_selected_queue_system,
+                )
+            return None
+
+    def add_global_queue_options(self, config_dict: ConfigDict) -> None:
+        for generic_option in fields(QueueOptions):
+            if (
+                generic_value := config_dict.get(generic_option.name.upper(), None)  # type: ignore
+            ) and self.__dict__[generic_option.name] == generic_option.default:
```
xjules, line 62: what is this other constrain: `self.__dict__[generic_option.name] == generic_option.default:`?

src/ert/config/queue_config.py
```diff
@@ -17,123 +19,159 @@
...
+    keep_qsub_output: bool = False
+
+    qstat_options: Optional[str] = pydantic.Field(default=None, deprecated=True)
+    queue_query_timeout: Optional[str] = pydantic.Field(default=None, deprecated=True)
+
+    @property
+    def driver_options(self) -> Dict[str, Any]:
+        driver_dict = asdict(self)
+        driver_dict["queue_name"] = driver_dict.pop("queue")
+        driver_dict.pop("max_running")
```
xjules, line 127: Why is `MAX_RUNNING` and `SUMBIT_SLEEP` removed?

src/ert/config/queue_config.py
```diff
@@ -151,18 +189,68 @@ def validate(self, mem_str_format: str) -> bool:
...
+            )
+        processed_options[option_name] = value
+
+
+def _raise_for_defaulted_invalid_options(queue_config_list: List[List[str]]) -> None:
+    # Invalid options names with no values (i.e. defaulted) are not passed to
+    # the validation system, thus we neeed to catch them expliclitly
+    for queue_system, option_name, *_ in queue_config_list:
+        if option_name not in valid_options[queue_system]:
+            raise ConfigValidationError(
```
xjules, line 226: I guess this won't be an issue when running `LOCAL_QUEUE`; ie. when they won't remove settings for the sake of testing but they will change `QUEUE_SYSTEM` to local.

src/ert/scheduler/slurm_driver.py
berland: We keep it only for backwards compatibility.

src/ert/config/queue_config.py
```diff
@@ -180,42 +268,82 @@ def from_dict(cls, config_dict: ConfigDict) -> QueueConfig:
...
-            selected_queue_system, queue_options, config_dict, realization_memory
-        )
+                queue_options.project_code = "+".join(tags)
+
+        if selected_queue_system == QueueSystem.TORQUE:
+            _check_num_cpu_requirement(
+                config_dict.get("NUM_CPU", 1), queue_options, _raw_queue_options
+            )
+
+        for _queue_vals in _all_validated_queue_options.values():
```
xjules, line 316: All these ifs can be solved by one `for loop` I guess :smile:

tests/unit_tests/config/test_queue_config.py
```diff
@@ -276,18 +281,18 @@ def test_initializing_empty_config_queue_options_resets_to_default_value(
...
 
 
 @pytest.mark.usefixtures("use_tmpdir")
 @pytest.mark.parametrize(
     "queue_system, queue_option, queue_value, err_msg",
     [
-        ("SLURM", "SQUEUE_TIMEOUT", "5a", "is not a valid integer or float"),
-        ("TORQUE", "NUM_NODES", "3.5", "is not a valid positive integer"),
+        ("SLURM", "SQUEUE_TIMEOUT", "5a", "should be a valid number"),
+        ("TORQUE", "NUM_NODES", "3.5", "should be a valid integer"),
```
xjules, line 295: `should be a valid positive integer`

tests/unit_tests/config/test_queue_config.py
```diff
@@ -524,22 +528,34 @@ def test_wrong_generic_queue_option_raises_validation_error(queue_system, key, v
...
     )
     with pytest.raises(ConfigValidationError, match=error_msg):
         ErtConfig.from_file("config.ert")
+
+
+@pytest.mark.parametrize(
+    "queue_system",
+    (QueueSystem.LSF, QueueSystem.TORQUE, QueueSystem.LOCAL, QueueSystem.SLURM),
+)
+def test_driver_initialization_from_defaults(queue_system):
```
xjules, line 553: should there be an assert or something in the end?

src/ert/config/queue_config.py
```diff
@@ -18,123 +20,159 @@
...
+
+@pydantic.dataclasses.dataclass
+class LocalOptions(QueueOptions):
+    @property
+    def driver_options(self) -> Dict[str, Any]:
+        return {}
+
+
+@pydantic.dataclasses.dataclass
+class LsfOptions(QueueOptions):
```
xjules, line 88: Just thinking (aloud) here. Should we use the full names as class name; ie. `LsfQueueOptions`? Mainly `LocalOptions` sounds quite strange tbh.

src/ert/config/queue_config.py
berland: It is to satisfy `test_global_config_key_does_not_overwrite_queue_options`

xjules (changes requested):
Very nice job @berland ! Looks good, had a small comment on naming the `*Options` class and one test that requires some input. Otherwise :+1:

src/ert/config/queue_config.py
berland: It is to fail hard whenever you have e.g. `QUEUE_OPTION LSF BLABLABLAOPTION foovalue` in your config.

src/ert/config/queue_config.py
berland: from line 316 and onwards? Not so trivial I think?

tests/unit_tests/config/test_queue_config.py
berland: The intention of the test was just to ensure that we can initialize without errors. If there are any spelling errors in the parameter names to the drivers init() this test will fail, also if we provide too many parameters. Testing that the arguments are processed correctly inside the drivers should be the responsibility of the drivers, not the config tests.

src/ert/config/queue_config.py
berland: Good suggestion. Will fix.

src/ert/config/queue_config.py
xjules: Yes, it is possible but not really pretty. Let's keep it.

xjules (approved):
Very good! :rocket:

PR #8494, opened 2024-08-16
Better handling of faulty Eclipse versions

### LINKED ISSUE #8331: Better handling of faulty Eclipse versions
**Describe the bug**
If you currently specify an Eclipse version that does not exist on disk, with a komodo release the ERT forward models will fail in eclipse with the following error message:

```
Realization: 0 failed after reaching max submit (1):

job ECLIPSE100 failed with: 'Process exited with status code 1'
	stderr file: '/private/levje/hm_analysis/realization-0/iter-0/ECLIPSE100.stderr.20',
	its contents:
  File "/private/levje/ert/src/ert/shared/share/ert/forward-models/res/script/ecl100.py", line 9, in <module>
    run(config, [arg for arg in sys.argv[1:] if len(arg) > 0])
  File "/private/levje/ert/src/ert/shared/share/ert/forward-models/res/script/ecl_run.py", line 544, in run
    sim = config.sim(version=options.version)
  File "/private/levje/ert/src/ert/shared/share/ert/forward-models/res/script/ecl_config.py", line 152, in sim
    return self._get_sim(version, Keys.scalar)
  File "/private/levje/ert/src/ert/shared/share/ert/forward-models/res/script/ecl_config.py", line 135, in _get_sim
    binaries: Dict[str, str] = self._config[Keys.versions][version][exe_type]
KeyError: 'versions'
```
There are multiple issues, or possible improvements here. One is that we do not verify that the version exists during parsing of the config file. Second we should probably give suggestions for what versions exists. Third is that currently if the version does not exist in `eclrun` it falls back to using our implementation for `ecl_run` and this is not compatible with the config file from ert-configurations. 
Some of these issues are however Equinor specific and I believe we would need to figure out in what capacity we want to support a Eclipse forward model that can be used by anyone, or if our internal forward model step should be targeted for internal use.

**To reproduce**
Run a model with Eclipse as a forward model step with Komodo.

**Expected behaviour**
Should provide better feedback for users, and possibly verify this in the parsing (or ert_config.py). Maybe make our ecl100 forward model step less generic, and instead encourage external users to configure their own forward model step, which would give them better controll as well. 

**Environment**
 - OS: RHEL8
 - ERT/Komodo release: Bleeding/Stable
 - Python version: 3.8/3.11

**Additional context**
See this related issue: https://github.com/equinor/ert/issues/4241

DanSava:
**Issue**
Resolves #8331 

Screenshot from 2024-08-16 12-44-28

## When applicable

### DIFF
--- a/src/ert/plugins/hook_implementations/forward_model_steps.py
+++ b/src/ert/plugins/hook_implementations/forward_model_steps.py
@@ -275,6 +275,17 @@ class Eclipse300(ForwardModelStepPlugin):
             ],
             default_mapping={"<NUM_CPU>": 1, "<OPTS>": "", "<VERSION>": "version"},
         )
+    def validate_pre_experiment(self, fm_step_json: ForwardModelStepJSON) -> None:
+        if "<VERSION>" not in self.private_args:
+            raise ForwardModelStepValidationError(
+                "Forward model step ECLIPSE300 must be given a VERSION argument"
+            )
+
+        _validate_ecl_version(
+            version=self.private_args["<VERSION>"],
+            sim_name="e300",
+            ecl_site_config="ECL300_SITE_CONFIG",
+        )
 
     @staticmethod
     def documentation() -> Optional[ForwardModelStepDocumentation]:
@@ -612,10 +623,10 @@ def installable_forward_model_steps() -> List[Type[ForwardModelStepPlugin]]:
     return [*_UpperCaseFMSteps, *_LowerCaseFMSteps]
 
 
-def _validate_ecl_version(version, sim_name, ecl_site_config):
-    def _get_eclrun_env(config: dict) -> dict:
+def _validate_ecl_version(version: str, sim_name: str, ecl_site_config: str) -> None:
+    def _get_eclrun_env(config):
         if "eclrun_env" in config:
-            env: dict = os.environ.copy()
+            env = os.environ.copy()
             eclrun_env = config["eclrun_env"]
             for key, value in eclrun_env.copy().items():
                 if key == "PATH":
@@ -625,7 +636,7 @@ def _validate_ecl_version(version, sim_name, ecl_site_config):
             return env
         return {}
 
-    def _get_available_eclrun_versions(eclrun_env, sim_name) -> List[str]:
+    def _get_available_eclrun_versions(eclrun_env, sim_name):
         try:
             return (
                 subprocess.check_output(
@@ -639,7 +650,7 @@ def _validate_ecl_version(version, sim_name, ecl_site_config):
         except subprocess.CalledProcessError:
             return []
 
-    def _get_ecl_site_config(ecl_config: str) -> dict:
+    def _get_ecl_site_config(ecl_config):
         site_config = ErtConfig.read_site_config()
         config_file = next(
             v for k, v in site_config.get("SETENV", []) if k == ecl_config
--- a/src/ert/plugins/hook_implementations/forward_model_steps.py
+++ b/src/ert/plugins/hook_implementations/forward_model_steps.py
@@ -623,9 +623,9 @@ def installable_forward_model_steps() -> List[Type[ForwardModelStepPlugin]]:
 
 
 def _get_eclrun_env(config: Dict[str, Any]) -> Dict[str, str]:
-    env_path = os.getenv("PATH")
-    ecl_path = config.get("eclrun_env", {}).get("PATH", "")
-    if ecl_path:
+    env_path = os.getenv("PATH", "")
+    ecl_path = config.get("eclrun_env", {}).get("PATH")
+    if ecl_path is not None:
         return {"PATH": env_path + os.pathsep + ecl_path}
     return {"PATH": env_path}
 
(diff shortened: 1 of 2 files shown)


src/ert/plugins/hook_implementations/forward_model_steps.py
```diff
@@ -598,3 +620,57 @@ def documentation() -> Optional[ForwardModelStepDocumentation]:
...
+        return {}
+    with open(config_file, encoding="utf-8") as f:
+        try:
+            return yaml.safe_load(f)
+        except yaml.YAMLError as e:
+            raise ValueError(f"Failed parse: {config_file} as yaml") from e
+
+
+def _validate_ecl_version(version: str, sim_name: str) -> None:
+    ecl_version = 100 if sim_name == "eclipse" else 300
```
yngve-sk, line 667: Do we need both `sim_name` and `version`. If not, this is always only for eclipse anyway, so we could just give a version which is typed Literal["100", "300"] (for now).

src/ert/plugins/hook_implementations/forward_model_steps.py
```diff
@@ -598,3 +620,57 @@ def documentation() -> Optional[ForwardModelStepDocumentation]:
 @plugin(name="ert")
 def installable_forward_model_steps() -> List[Type[ForwardModelStepPlugin]]:
     return [*_UpperCaseFMSteps, *_LowerCaseFMSteps]
+
+
+def _get_eclrun_env(config: Dict[str, Any]) -> Dict[str, str]:
+    env_path = os.getenv("PATH", "")
+    ecl_path = config.get("eclrun_env", {}).get("PATH")
```
yngve-sk, line 627: Why take in the entire config dict? Could take in only `ecl_path` as an arg

src/ert/plugins/hook_implementations/forward_model_steps.py
```diff
@@ -598,3 +620,57 @@ def documentation() -> Optional[ForwardModelStepDocumentation]:
...
+            return yaml.safe_load(f)
+        except yaml.YAMLError as e:
+            raise ValueError(f"Failed parse: {config_file} as yaml") from e
+
+
+def _validate_ecl_version(version: str, sim_name: str) -> None:
+    ecl_version = 100 if sim_name == "eclipse" else 300
+    if shutil.which("eclrun") is not None:
+        ecl_config = _get_ecl_config(sim_name)
+        ecl_env = _get_eclrun_env(ecl_config)
```
yngve-sk, line 670: This seem to only ever contain `PATH`? In which case it could be simplified a bit as it seems to be made for something more generic at the moment

yngve-sk (commented):
Overall idea looks good and the tests seem sensible, I think the functions for ecl version checking could be simplified a bit, maybe even put into fewer/one functions.

src/ert/plugins/hook_implementations/forward_model_steps.py
DanSava: Yes, we can take in only the `ecl_path` as an arg, that will simplify it a bit.

src/ert/plugins/hook_implementations/forward_model_steps.py
DanSava: The version here is the one provided in the user config file i.e. `2022.9` and I think I need the simulator name `eclipse` or `e300` to get the available releases when calling `eclrun`

src/ert/plugins/hook_implementations/forward_model_steps.py
```diff
@@ -264,6 +280,22 @@ def __init__(self) -> None:
             default_mapping={"<NUM_CPU>": 1, "<OPTS>": "", "<VERSION>": "version"},
         )
 
+    def validate_pre_experiment(self, fm_step_json: ForwardModelStepJSON) -> None:
```
andreas-el, line 283: Is `fm_step_json` used here?

src/ert/plugins/hook_implementations/forward_model_steps.py
```diff
@@ -264,6 +280,22 @@ def __init__(self) -> None:
...
+        if shutil.which("eclrun") is not None:
+            pm = ErtPluginManager()
+            ecl_env = _ecl_env_path(pm.get_ecl300_config_path())
+            version = self.private_args["<VERSION>"]
+            available_versions = _available_eclrun_versions(ecl_env, "e300")
+            if version not in available_versions:
+                raise ForwardModelStepValidationError(
+                    f"Unavailable ECLIPSE300 version {version} current supported "
+                    f"versions {available_versions}"
+                )
```
andreas-el, line 297: Could this be one function taking in ecl_env and two strings, "e300" and "ECLIPSE300" ?
In the previous case it would be ecl_env, "eclipse" and "ECLIPSE100" ?

src/ert/plugins/hook_implementations/forward_model_steps.py
DanSava: No, it is not used.

src/ert/plugins/hook_implementations/forward_model_steps.py
DanSava: After some discussion with @yngve-sk and some iterations, we decided to move all the logic inside one function `_available_eclrun_versions` that will only take the simulator name() as an argument.

yngve-sk (approved):
LGTM🚀 Thanks for your patience with all the comments

Issue #8493, opened 2024-08-16
Stop auto-switching to currently running iteration / update-step when users clicks on a realization-circle


dafeda:
Currently, ert will automatically switch to the currently running iteration / udpate-step:

https://github.com/user-attachments/assets/d041d5d0-69f7-4cfe-a6b2-463a615b7213

If the user clicks on a realization-circle, it could indicate that the user wants to take control and automatic switching should be turned off.

PR #8500, opened 2024-08-16
Allow user to inhibit tab changes during simulation

andreas-el:
**Issue**
Resolves https://github.com/equinor/ert/issues/8493

Ert will only change tab index if the last index is the current active one.

https://github.com/user-attachments/assets/b900c307-6b8b-4efa-9c06-95b628bf32cb

## When applicable


dafeda (approved):
Looks great I think!
@eivindjahren are you OK with this even though it does not turn off automatic switching when clicking on one of the circles?

eivindjahren:
The only problem I can see is that if the user is on the very last tab and she has opened a realization to look at the log, once the update step starts then the tab changes. Not sure what can be done about that, but this is probably better than nothing.

PR #8502, opened 2024-08-16
Logging improvements

sondreso:
**Issue**
We are producing many log files for different modules, and some logs are not propagated correctly 

**Approach**
Have one log file for ERT, all modules propagate logs, handlers filter on log level according to their needs. 
Also, set `disable_existing_loggers: False` as the default behavior is to disable all loggers created using `logging.getLogger`. Otherwise we would need to explicitly enable them again in the logger config, making for a very verbose config. 

## When applicable

### DIFF
--- a/src/ert/__main__.py
+++ b/src/ert/__main__.py
@@ -42,6 +42,8 @@ from ert.validation import (
     ValidationStatus,
 )
 
+logger = logging.getLogger(__name__)
+
 
 def run_ert_storage(args: Namespace, _: Optional[ErtPluginManager] = None) -> None:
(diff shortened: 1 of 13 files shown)


eivindjahren (approved):
𝄠Understandable, understandable 𝅘𝅥𝅮
 𝄫 Yes, it's perfectly understandable :trumpet: 
 :violin:  Comprehensible, Comprehensible :postal_horn: 
:musical_keyboard:   Not a bit reprehensible :woman_dancing: 
:dancers:   It's so defensible! :dancers:

PR #8513, opened 2024-08-19
Refactor `Storage` to better represent the hierarchy `Storage`->`Experiments`->`Ensembles`

dafeda:
We have talked about making the concept of `Experiment` more central in ert.
For example, e want to be able to select and delete experiments from the GUI and have a nice API to use via Jupyter Notebooks.

As a step towards this, I propose a refactor of the storage that better reflects the hierarchy of `Storage`->`Experiments`->`Ensembles`.

Main ideas:
- `Storage` is responsible for creating and deleting `Experiments`
- `Experiment` is responsible for creating and (eventually perhaps) deleting `Ensembles`
- `Ensemble` has methods for reading parameters, responses etc.

## When applicable

### DIFF
--- a/src/ert/gui/ertnotifier.py
+++ b/src/ert/gui/ertnotifier.py
@@ -1,8 +1,8 @@
-from typing import Optional
+from typing import List, Optional
 
 from qtpy.QtCore import QObject, Signal, Slot
 
-from ert.storage import Ensemble, Storage
+from ert.storage import LocalEnsemble, LocalStorage
 
 
 class ErtNotifier(QObject):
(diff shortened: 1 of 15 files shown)


src/ert/gui/ertnotifier.py
```diff
@@ -1,28 +1,30 @@
...
+    current_experiment_changed = Signal(object, name="currentExperimentChanged")
     current_ensemble_changed = Signal(object, name="currentEnsembleChanged")
 
     def __init__(self, config_file: str):
         QObject.__init__(self)
         self._config_file = config_file
-        self._storage: Optional[Storage] = None
-        self._current_ensemble: Optional[Ensemble] = None
+        self._storage: Optional[LocalStorage] = None
+        self._current_experiment: Optional[LocalExperiment] = None
```
yngve-sk, line 18: For this and `current_ensemble`, store a direct reference to the object instead of an ID (+ some `@property` fns that fetch from self._storage based on this), would be a bit more decoupled than storing a direct reference

PR #8534, opened 2024-08-21
Fix broken ert test_run

### LINKED ISSUE #8530: ert test_run does not work for non-LOCAL queue systems.
When ert is run on the following:

```
QUEUE_SYSTEM LSF
NUM_REALIZATIONS 1
```

a test_run will crash:

```
$ ert test_run poly.ert
    --> Running forecast for iteration: 0

    1/1 |                                                             |   0% Running time: 2 seconds

    ■ Waiting           1/1
    ■ Pending           0/1
    ■ Running           0/1
    ■ Failed            0/1
    ■ Finished          0/1
    ■ Unknown           0/1

Experiment failed with the following error: Number of active realizations (0) is less than the specified MIN_REALIZATIONS(1)
Unexpected exception in ensemble: 
  File "/data/projects/ert/src/ert/ensemble_evaluator/_ensemble.py", line 246, in _evaluate_inner
    driver = create_driver(self._queue_config)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Unexpected exception in ensemble: 
  File "/data/projects/ert/src/ert/scheduler/__init__.py", line 23, in create_driver
    return LocalDriver(**config.queue_options.driver_options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Unexpected exception in ensemble: 
TypeError: LocalDriver.__init__() got an unexpected keyword argument 'project_code'
```

This is a regression from #8475

berland:
The commit 4013764 introduced a regression such that `ert test_run` would fail on any configs that sets a queue system other than "local".

This was due to the function `create_local_copy()` not having been updated to yield the correct QueueOptions object after refactoring.

**Issue**
Resolves #8530

**Approach**
:brain:

## When applicable

### DIFF
--- a/src/ert/config/queue_config.py
+++ b/src/ert/config/queue_config.py
@@ -254,6 +254,7 @@ class QueueConfig:
     max_submit: int = 1
     queue_system: QueueSystem = QueueSystem.LOCAL
     queue_options: QueueOptions = field(default_factory=QueueOptions)
+    queue_options_test_run: QueueOptions = field(default_factory=LocalQueueOptions)
     stop_long_running: bool = False
 
     @no_type_check
@@ -297,6 +298,7 @@ class QueueConfig:
         )
 
         queue_options = _all_validated_queue_options[selected_queue_system]
+        queue_options_test_run = _all_validated_queue_options[QueueSystem.LOCAL]
         queue_options.add_global_queue_options(config_dict)
 
         if queue_options.project_code is None:
@@ -351,6 +353,7 @@ class QueueConfig:
             max_submit,
             selected_queue_system,
             queue_options,
+            queue_options_test_run,
             stop_long_running=stop_long_running,
         )
 
@@ -360,7 +363,8 @@ class QueueConfig:
             self.realization_memory,
             self.max_submit,
             QueueSystem.LOCAL,
-            self.queue_options,
+            self.queue_options_test_run,
+            self.queue_options_test_run,
             stop_long_running=self.stop_long_running,
         )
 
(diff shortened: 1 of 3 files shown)


src/ert/config/queue_config.py
```diff
@@ -360,7 +363,8 @@ def create_local_copy(self) -> QueueConfig:
             self.realization_memory,
             self.max_submit,
             QueueSystem.LOCAL,
-            self.queue_options,
+            self.queue_options_test_run,
```
xjules, line 366: This should be just `self.queue_options` I guess.

eivindjahren (commented):
Could we maybe have a regression test here where you have a config that uses LSF specific queue options, The queue system is set to LSF, and then you run `ert test_run`? Could probably create something like `test_bad_config_error_message`.

berland:
> Could we maybe have a regression test here where you have a config that uses LSF specific queue options, The queue system is set to LSF, and then you run `ert test_run`? Could probably create something similar to `test_bad_config_error_message`.

Added.

eivindjahren (approved):
:+1:

tests/integration_tests/test_cli.py
```diff
@@ -41,6 +41,13 @@ def test_bad_config_error_message(tmp_path):
         run_cli(TEST_RUN_MODE, "--disable-monitor", str(tmp_path / "test.ert"))
 
 
+def test_test_run_on_lsf_configuration_works_with_no_errors(tmp_path):
+    (tmp_path / "test.ert").write_text(
+        "NUM_REALIZATIONS 1\nQUEUE_SYSTEM LSF", encoding="utf-8"
```
eivindjahren, line 46: Maybe just add one `QUEUE_OPTION LSF PROJECT PROJECT` to make sure we populate some incompatible options.

berland: The options were not really part of the problem in this bug. There are other tests asserting which errors or warnings arise when incompatible options are supplied, I think they suffice.

PR #8554, opened 2024-08-23
Combine gendata config directly

yngve-sk:
Just testing

### DIFF
--- a/src/ert/config/ensemble_config.py
+++ b/src/ert/config/ensemble_config.py
@@ -4,32 +4,28 @@ import logging
 import os
 from collections import Counter
 from dataclasses import dataclass, field
-from datetime import datetime
 from typing import (
-    Any,
     Dict,
     List,
     Optional,
-    Sequence,
     Union,
     no_type_check,
     overload,
 )
 
-import numpy as np
-import numpy.typing as npt
-
 from ert.field_utils import get_shape
 
-from ._read_summary import read_summary
 from .field import Field
 from .gen_data_config import GenDataConfig
 from .gen_kw_config import GenKwConfig
 from .parameter_config import ParameterConfig
 from .parsing import ConfigDict, ConfigKeys, ConfigValidationError
+from .refcase import Refcase
 from .response_config import ResponseConfig
 from .summary_config import SummaryConfig
 from .surface_config import SurfaceConfig
+
+_KNOWN_RESPONSE_TYPES = [SummaryConfig, GenDataConfig]
 
 logger = logging.getLogger(__name__)
 
@@ -50,28 +46,6 @@ def _get_abs_path(file: Optional[str]) -> Optional[str]:
     return file
 
 
-@dataclass(eq=False)
-class Refcase:
-    start_date: datetime
-    keys: List[str]
-    dates: Sequence[datetime]
-    values: npt.NDArray[Any]
-
-    def __eq__(self, other: object) -> bool:
-        if not isinstance(other, Refcase):
-            return False
-        return bool(
-            self.start_date == other.start_date
-            and self.keys == other.keys
-            and self.dates == other.dates
-            and np.all(self.values == other.values)
-        )
-
-    @property
-    def all_dates(self) -> List[datetime]:
-        return [self.start_date] + list(self.dates)
-
-
 @dataclass
 class EnsembleConfig:
     grid_file: Optional[str] = None
@@ -82,15 +56,17 @@ class EnsembleConfig:
 
     def __post_init__(self) -> None:
         self._check_for_duplicate_names(
-            list(self.parameter_configs.values()), list(self.response_configs.values())
+            [p.name for p in self.parameter_configs.values()],
+            [key for config in self.response_configs.values() for key in config.keys],
         )
+
         self.grid_file = _get_abs_path(self.grid_file)
 
     @staticmethod
     def _check_for_duplicate_names(
-        parameter_list: List[ParameterConfig], gen_data_list: List[ResponseConfig]
+        parameter_list: List[str], gen_data_list: List[str]
     ) -> None:
-        names_counter = Counter(g.name for g in parameter_list + gen_data_list)
+        names_counter = Counter(g for g in parameter_list + gen_data_list)
         duplicate_names = [n for n, c in names_counter.items() if c > 1]
         if duplicate_names:
             raise ConfigValidationError.with_context(
@@ -104,8 +80,6 @@ class EnsembleConfig:
     @classmethod
     def from_dict(cls, config_dict: ConfigDict) -> EnsembleConfig:
         grid_file_path = config_dict.get(ConfigKeys.GRID)
-        refcase_file_path = config_dict.get(ConfigKeys.REFCASE)
-        gen_data_list = config_dict.get(ConfigKeys.GEN_DATA, [])
         gen_kw_list = config_dict.get(ConfigKeys.GEN_KW, [])
         surface_list = config_dict.get(ConfigKeys.SURFACE, [])
         field_list = config_dict.get(ConfigKeys.FIELD, [])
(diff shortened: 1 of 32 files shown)


src/ert/storage/local_experiment.py
```diff
@@ -269,13 +269,17 @@ def parameter_configuration(self) -> Dict[str, ParameterConfig]:
...
-    @cached_property
+    @property
     def response_configuration(self) -> Dict[str, ResponseConfig]:
-        params = {}
+        responses = {}
         for data in self.response_info.values():
-            param_type = data.pop("_ert_kind")
-            params[data["name"]] = _KNOWN_RESPONSE_TYPES[param_type](**data)
-        return params
+            ert_kind = data.pop("_ert_kind")
```
oyvindeide, line 276: Maybe drop `_ert_kind` and just use `type`?

src/ert/config/summary_config.py
```diff
@@ -45,3 +44,14 @@ def read_from_file(self, run_path: str, iens: int) -> xr.Dataset:
             coords={"time": time_map, "name": keys},
         )
         return ds.drop_duplicates("time")
+
+    @property
+    def cardinality(self) -> Literal["one_per_key", "one_per_realization"]:
```
oyvindeide, line 49: Not used?

src/ert/run_models/base_run_model.py
```diff
@@ -330,6 +331,9 @@ def start_simulations_thread(
         finally:
             self._clean_env_context()
             self.stop_time = int(time.time())
+            if failed:
```
oyvindeide, line 334: Unrelated?

src/ert/storage/local_ensemble.py
```diff
@@ -609,14 +616,27 @@ def load_responses(self, key: str, realizations: Tuple[int]) -> xr.Dataset:
...
+        elif key not in self.experiment.response_key_to_response_type:
             raise ValueError(f"{key} is not a response")
+        else:
+            response_type = self.experiment.response_key_to_response_type[key]
+            select_key = True
+
         loaded = []
         for realization in realizations:
-            input_path = self._realization_dir(realization) / f"{key}.nc"
+            input_path = self._realization_dir(realization) / f"{response_type}.nc"
```
oyvindeide, line 630: Regarding our offline discussion about new response types (CSV_RESPONSE), would we require something else here?

yngve-sk: I think so yes, not 100% on exactly what. But we can set the response_type (or some other thing to identify it) dynamically when loading data from runpath, as then we will know the primary key of the data, which is part of the dynamic response type. Should be possible to make it fit later on

src/ert/config/summary_config.py
yngve-sk: removed

src/ert/run_models/base_run_model.py
yngve-sk: removed

src/ert/storage/local_experiment.py
yngve-sk: Hmm could, by `type` you mean `response_type`? In the case of a "dynamic" `response_type` with `CSV_RESPONSE` / `CSV_OBSERVATION` we would need some way to identify the class, and `_ert_kind` is the already existing way to do this. If we know for sure we can lock in response type (which is currently used to identify names of tables of responses in storage) we could go for it. Maybe leave this for a separate PR?

src/ert/config/ensemble_config.py
```diff
@@ -132,48 +103,24 @@ def make_field(field_list: List[str]) -> Field:
...
-                SummaryConfig(
-                    name="summary",
-                    input_file=eclbase,
-                    keys=[i for val in summary_keys for i in val],
-                    refcase=time_map,
-                )
-            )
+        response_configs: List[ResponseConfig] = []
+
+        for config_cls in responses_index.values():
```
oyvindeide, line 114: How would this work in our hypothetical `CSV_RESPONSE` example?

src/ert/config/ert_config.py
```diff
@@ -542,17 +544,24 @@ def manifest_to_json(self, iens: int = 0, iter: int = 0) -> Dict[str, Any]:
...
-                input_file = input_file.replace("<IENS>", str(iens))
-                manifest[f"{name}_UNSMRY"] = f"{input_file}.UNSMRY"
-                manifest[f"{name}_SMSPEC"] = f"{input_file}.SMSPEC"
-            if isinstance(respons_config, GenDataConfig):
-                if respons_config.report_steps:
-                    for step in respons_config.report_steps:
+        if "summary" in self.ensemble_config.response_configs:
+            smry_config = self.ensemble_config.response_configs["summary"]
+            input_file = smry_config.input_files[0].replace("<IENS>", str(iens))
+            manifest["summary_UNSMRY"] = f"{input_file}.UNSMRY"
```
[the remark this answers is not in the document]
yngve-sk: Can be done. I think we should consider, instead of replacing `input_files`, introducing another `@property` where the config tells the outside world what files it is expecting. If we store it "directly", for gen data we'll have to map it back to report steps in read_from_file. If we serialize a list of input_files where the report step is at the end we might get some edge cases where there is a _0 but not from a report step and just a file name, and we can't know how.

Manifest would then look something like this:
```
        for config in self.ensemble_config.response_configs.values():
            for expected_runpath_file in config.expected_input_files:
                manifest[f"{config.response_type}_{expected_runpath_file}"] = (
                    expected_runpath_file
                )
```

src/ert/config/ensemble_config.py
yngve-sk: It would instantiate the config itself, then the config object would later read something from the runpath when invoked from `callsbacks.py` and save it to storage. At this point we'd be able to find out what the columns are etc etc.

yngve-sk: Here is a non-working proof of concept example. We would have to make it possible to read_from_file to return multiple datasets as we wouldn't necessarily want to join all the outputs of the (combined) CsvResponseConfig, as they may have different columns
```
import csv
from dataclasses import dataclass
from typing import List, Optional

import xarray as xr
from typing_extensions import Self
from .parsing import ConfigDict
from .response_config import ResponseConfig
from .responses_index import responses_index


@dataclass
class CsvResponseConfig(ResponseConfig):
    name: str = "csv_response"

    @classmethod
    def from_config_dict(cls, config_dict: ConfigDict) -> Optional[Self]:
        csv_response_list = config_dict.get("CSV_RESPONSE", [])  # type: ignore

        for rs_config in csv_response_list:
            pass

        return cls(
            name="csv_response",
            keys="any",
            input_files=["something.csv", "something2.csv"],
            report_steps_list=report_steps,
        )

    def read_from_file(self, run_path: str, _: int) -> List[xr.Dataset]:

        datasets = []
        for fp in self.input_files:
            with open(fp) as file:
                header, content = csv.reader(file)
                # ...
                datasets.append((fp,))

        return xr.concat(datasets)

    @property
    def response_type(self) -> str:
        return "csv_response"


responses_index.add_response_type(CsvResponseConfig)
```

src/ert/simulator/batch_simulator.py
```diff
@@ -102,10 +102,24 @@ def callback(*args, **kwargs):
...
+        if "gen_data" not in ens_config:
             ens_config.addNode(
-                GenDataConfig(name=key, input_file=f"{key}_%d", report_steps=[0])
+                GenDataConfig(
+                    keys=results,
+                    input_files=[f"{k}_%d" for k in results],
+                    report_steps_list=[[0] for _ in results],
+                )
             )
+        else:
```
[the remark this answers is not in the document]
yngve-sk: I think maybe it can't... On 2nd thought I guess there is always gen data as there are always controls in Everest? Can remove it in that case.

src/ert/config/response_config.py
```diff
@@ -1,15 +1,19 @@
...
+from typing_extensions import Self
 
 from ert.config.parameter_config import CustomDict
+from ert.config.parsing import ConfigDict
 
 
 @dataclasses.dataclass
 class ResponseConfig(ABC):
     name: str
+    input_files: List[str] = dataclasses.field(default_factory=list)
```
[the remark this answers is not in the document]
yngve-sk: Addressed in response to same previous comment

src/ert/config/ert_config.py
```diff
@@ -933,7 +943,8 @@ def _create_observations(
                 else:
                     config_errors.append(
                         ErrorInfo(
-                            message=f"Unknown ObservationType {type(values)} for {obs_name}"
+                            message=f"Unknown ObservationType {type(values)} for"
```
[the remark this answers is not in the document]
yngve-sk: Will remove

src/ert/simulator/batch_simulator.py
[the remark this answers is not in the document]
yngve-sk: 👍 Not able to run everest without controls, so there will always be gendata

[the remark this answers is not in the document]
yngve-sk: Edit: Seems gen data can be empty, the ert config parsed from everest2res does not include `GEN_DATA`, it is added "manually" here, so in this case we do need to explicitly initialize it, and probably later in a separate smaller PR, refactor it into `everest2res`

src/ert/config/ert_config.py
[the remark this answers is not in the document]
oyvindeide: That looks a lot cleaner and reduces the need of intricate knowledge of each config type

src/ert/simulator/batch_simulator.py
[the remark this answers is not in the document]
oyvindeide: Great if you create an issue then so we dont forget

src/ert/config/ensemble_config.py
oyvindeide: Right, would that be preferential to having multiple instances of `CSVResponseConfig`, one for each primary key?

yngve-sk: I think they can both do the same and are pretty equal, so not strongly leaning towards the one or the other, just consistency which would be to have "combined" response configs now. Each "instance" would be embedded in here.
Example ERT config:
```
CSV_RESPONSE resp1 key:auto
CSV_RESPONSE resp2 key:auto
```

The primary  keys could be the same, or they could not. If we do not have dynamic primary keys and only declare beforehand, it would make more sense, but we could still just encode it "per instance".

src/ert/simulator/batch_simulator.py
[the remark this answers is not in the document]
yngve-sk: Done: https://github.com/equinor/everest/issues/106

src/ert/config/ensemble_config.py
oyvindeide: I think we should stick to declaring it up front, as that makes validation so much easier, but perhaps there are cases in the future where we dont want that.

src/ert/config/refcase.py
```diff
@@ -0,0 +1,54 @@
...
+import numpy.typing as npt
+
+from ._read_summary import read_summary
+from .parsing.config_dict import ConfigDict
+from .parsing.config_errors import ConfigValidationError
+from .parsing.config_keywords import ConfigKeys
+
+
+@dataclass(eq=False)
+class Refcase:
```
oyvindeide, line 20: Why did this move?

src/ert/config/responses_index.py
```diff
@@ -0,0 +1,44 @@
+from typing import Dict, Iterable, Tuple, Type
+
+from .response_config import ResponseConfig
+
+
+class _ResponsesIndex:
```
oyvindeide, line 6: Can you put everything related to this into a separate commit?

src/ert/config/refcase.py
yngve-sk: To avoid circular import. Was in `EnsembleConfig`, which indirectly instantiates/imports `SummaryConfig`. Moving some `Refcase` logic to `SummaryConfig` makes `SummaryConfig` have to import it, and it can't be from `EnsembleConfig`.

oyvindeide:
Also, remember labels, etc. on the PR

PR #8558, opened 2024-08-23
Remove unused infer iteration from rft csv export

dafeda:

## When applicable

### DIFF
--- a/src/ert/resources/workflows/jobs/internal-gui/scripts/gen_data_rft_export.py
+++ b/src/ert/resources/workflows/jobs/internal-gui/scripts/gen_data_rft_export.py
@@ -1,6 +1,5 @@
 import contextlib
 import os
-import re
 
 import numpy
 import pandas as pd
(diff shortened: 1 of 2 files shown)


src/ert/resources/workflows/jobs/internal-gui/scripts/gen_data_rft_export.py
```diff
@@ -119,8 +115,7 @@ def run(
         output_file = workflow_args[0]
         trajectory_path = workflow_args[1]
         ensemble_list = None if len(workflow_args) < 3 else workflow_args[2]
-        _ = True if len(workflow_args) < 4 else workflow_args[3]
-        drop_const_cols = False if len(workflow_args) < 5 else workflow_args[4]
+        drop_const_cols = False if len(workflow_args) < 4 else workflow_args[3]
```
eivindjahren, line 118: This ends up being a breaking change for the workflow, which is used quite a bit. We should be a bit careful here.

src/ert/resources/workflows/jobs/internal-gui/scripts/gen_data_rft_export.py
```diff
@@ -284,7 +249,7 @@ def getArguments(self, parent, storage):
                     output_path_model.getPath(),
                     trajectory_model.getPath(),
                     ensemble_list,
-                    infer_iteration_check.isChecked(),
+                    True,
```
eivindjahren, line 252: Think you want to remove this from the list now.

dafeda: Done, thanks.
`drop_const_cols` was read from `workflow_args` which a string, so it would be truthy even when "False".
I modified the test so that it checks the drop constant columns checkbox and produces the same results as before.

eivindjahren (approved):
:+1:

PR #8566, opened 2024-08-23
Use Lock mechanism for forward_model_ok

xjules:
**Issue**
Resolves https://github.com/equinor/ert/issues/8070

**Approach**
Make sure that we will not run more than 1 internalization job at a time. 

## When applicable

### DIFF
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -147,7 +147,8 @@ class Job:
             if self.returncode.result() == 0:
                 if self._scheduler._manifest_queue is not None:
                     await self._verify_checksum()
-                await self._handle_finished_forward_model()
+                async with self._scheduler._forward_model_ok_lock:
+                    await self._handle_finished_forward_model()
                 break
 
             if attempt < max_submit - 1:
--- a/src/ert/scheduler/scheduler.py
+++ b/src/ert/scheduler/scheduler.py
@@ -106,6 +106,10 @@ class Scheduler:
         self._completed_jobs_num: int = 0
         self.completed_jobs: asyncio.Queue[int] = asyncio.Queue()
 
+        # this lock is to assure that no more than 1 task
+        # does internalization at a time
+        self._forward_model_ok_lock: asyncio.Lock = asyncio.Lock()
+
         self._cancelled = False
         if max_submit < 0:
             raise ValueError(


src/ert/scheduler/scheduler.py
```diff
@@ -275,6 +275,7 @@ async def execute(
             scheduling_tasks.append(asyncio.create_task(self._update_avg_job_runtime()))
 
         sem = asyncio.BoundedSemaphore(self._max_running or len(self._jobs))
+        self._fmok_sem = asyncio.BoundedSemaphore(value=10)
```
berland, line 278: `value=1` should be enough for everybody

xjules: Actually, when it is only `1` I can use lock instead :+1:

src/ert/scheduler/scheduler.py
```diff
@@ -275,6 +276,9 @@ async def execute(
             scheduling_tasks.append(asyncio.create_task(self._update_avg_job_runtime()))
 
         sem = asyncio.BoundedSemaphore(self._max_running or len(self._jobs))
+        # this semaphore is to assure that no more than 1 task
+        # does internationalization at a time
```
berland, line 280: internalization (not being international today)

xjules:
Tested on bigpoly (300 realz, max_running 200) both on LSF and LOCAL queues and it worked.

src/ert/scheduler/job.py
```diff
@@ -223,17 +223,25 @@ async def _verify_checksum(self, timeout: int = 120) -> None:  # noqa: ASYNC109
...
-            self._callback_status_msg = status_msg
-        else:
-            self._callback_status_msg += f"\nstatus from done callback: {status_msg}"
-
-        if callback_status == LoadStatus.LOAD_SUCCESSFUL:
-            await self._send(JobState.COMPLETED)
-        else:
-            assert callback_status == LoadStatus.LOAD_FAILURE
-            await self._send(JobState.FAILED)
+        if self._scheduler._fmok_lock:
```
berland, line 226: You can do this with `async with self._scheduler._fmok_lock:`

src/ert/scheduler/scheduler.py
```diff
@@ -105,6 +105,7 @@ def __init__(
         self._average_job_runtime: float = 0
         self._completed_jobs_num: int = 0
         self.completed_jobs: asyncio.Queue[int] = asyncio.Queue()
+        self._fmok_lock: Optional[asyncio.Lock] = None
```
berland, line 108: Does this need to be `Optional`, you can set it to `asyncio.Lock()` immediately I suppose (and allowing the `async with` syntax.

berland:
Recommend to use a longer name for this thing rather than `fmok`.

src/ert/scheduler/job.py
xjules: No, I can't unfortunately. This would need to be mocked for many tests as fmok_lock can be `None`

src/ert/scheduler/scheduler.py
xjules: Yes, can try it, but that would alter some tests - can see how many this would be if way too many than can revert.

xjules:
> Recommend to use a longer name for this thing rather than `fmok`.

Any suggestion? `internalization_lock`?

xjules:
> > Recommend to use a longer name for this thing rather than `fmok`.
> 
> Any suggestion? `internalization_lock`?

changed to `self._forward_model_ok_lock`

berland (approved):
nice and terse!

PR #8569, opened 2024-08-26
Add gendata output file for each report step to the output files manifest

DanSava:
**Issue**
The output checksum checking is done only if the report step matches the iteration number and not for all report steps each iteration. 

## When applicable

### DIFF
--- a/src/ert/config/ert_config.py
+++ b/src/ert/config/ert_config.py
@@ -549,8 +549,9 @@ class ErtConfig:
                 manifest[f"{name}_UNSMRY"] = f"{input_file}.UNSMRY"
                 manifest[f"{name}_SMSPEC"] = f"{input_file}.SMSPEC"
             if isinstance(respons_config, GenDataConfig):
-                if respons_config.report_steps and iens in respons_config.report_steps:
-                    manifest[name] = input_file.replace("%d", str(iens))
+                if respons_config.report_steps:
+                    for step in respons_config.report_steps:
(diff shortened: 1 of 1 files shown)


src/ert/config/ert_config.py
```diff
@@ -549,8 +549,9 @@ def manifest_to_json(self, iens: int = 0, iter: int = 0) -> Dict[str, Any]:
                 manifest[f"{name}_UNSMRY"] = f"{input_file}.UNSMRY"
                 manifest[f"{name}_SMSPEC"] = f"{input_file}.SMSPEC"
             if isinstance(respons_config, GenDataConfig):
-                if respons_config.report_steps and iens in respons_config.report_steps:
-                    manifest[name] = input_file.replace("%d", str(iens))
+                if respons_config.report_steps:
+                    for step in respons_config.report_steps:
+                        manifest[f"{name}_{step}"] = input_file.replace("%d", str(step))
```
yngve-sk, line 554: Does the key-name of this manifest dict matter? Seems a bit arbitrary and not 1-to-1 corresponding with the expected filename?

DanSava: It does not matter too much. This was a suggestion/requirement for the manifest structure to look like this, I guess the idea was to be able to match response name to output file. As you have noticed this does not aligh properly every time.

yngve-sk: OK so a separate issue from this then. Otherwise this looks good

yngve-sk (approved):

PR #8579, opened 2024-08-28
Combine rss and oom_score checker in fm_runner

JHolba:
Calling process.children is noticeably slow when the running quick forward model steps.
Combined the rss and oom_score checker so that children only gets called once.

## When applicable

### DIFF
--- a/src/_ert_forward_model_runner/job.py
+++ b/src/_ert_forward_model_runner/job.py
@@ -10,7 +10,7 @@ import time
 from datetime import datetime as dt
 from pathlib import Path
 from subprocess import Popen, run
-from typing import Optional
+from typing import Optional, Tuple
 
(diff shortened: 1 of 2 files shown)


src/_ert_forward_model_runner/job.py
```diff
@@ -179,7 +179,7 @@ def ensure_file_handles_closed():
 
         max_memory_usage = 0
         while exit_code is None:
-            memory_rss = _get_rss_for_processtree(process)
+            (oom_score, memory_rss) = _get_oom_score_and_rss_for_processtree(process)
```
berland, line 182: memory_rss is the primary target, and oom_score is merely a bonus (really, it is a maybe-nice-to-have, maybe it should be the other way around.

JHolba: swapped them and changed name of the function to reflect it

berland (approved):
go!

PR #8593, opened 2024-08-29
Create asyncio.Lock and provide forward_model_ok_lock when when running scheduler.execute and provide it dir…

xjules:
…ectly to job.run

**Issue**
Resolves #my_issue

**Approach**

## When applicable


xjules:
Note to self: update the commit message :smile:

src/ert/scheduler/job.py
```diff
@@ -136,7 +136,12 @@ async def _submit_and_run_once(self, sem: asyncio.BoundedSemaphore) -> None:
                 timeout_task.cancel()
             sem.release()
 
-    async def run(self, sem: asyncio.BoundedSemaphore, max_submit: int = 1) -> None:
+    async def run(
+        self,
+        sem: asyncio.BoundedSemaphore,
+        forward_model_ok_lock: asyncio.Lock,
```
jonathan-eq, line 142: Should we provide a default lock here? Then we would not have to change any of the unit tests.
```suggestion
        forward_model_ok_lock: asyncio.Lock = asyncio.Lock(),
```

xjules: I'd keep it as is, since this emphasizes that `Lock` needs to be created before hand.

PR #8599, opened 2024-08-30
Catch FileNotFoundError in driver._execute_with_retry

### LINKED ISSUE #8563: Catch FileNotFoundError in driver._execute_with_retry
The utility function for running shell commands should probably catch potential FileNotFoundErrors and return immediately (skipping retries).

When run on a command not in path, an exception will be raised and it is better caught.

https://github.com/equinor/ert/blob/357dcef3ad39d4a9ef7b3f7c9ab107d0d9ca32fd/src/ert/scheduler/driver.py#L93

jonathan-eq:
**Issue**
Resolves #8563

**Approach**
This commit makes the drivers handle FileNotFound exceptions in `driver._execute_with_retry(...)` and `driver.poll()` for the various drivers. The driver will now return early and avoid retrying invalid shell commands.

## When applicable

### DIFF
--- a/src/ert/scheduler/driver.py
+++ b/src/ert/scheduler/driver.py
@@ -90,12 +90,16 @@ class Driver(ABC):
         error_message: Optional[str] = None
 
         for _ in range(total_attempts):
-            process = await asyncio.create_subprocess_exec(
-                *cmd_with_args,
-                stdin=asyncio.subprocess.PIPE if stdin else None,
-                stdout=asyncio.subprocess.PIPE,
-                stderr=asyncio.subprocess.PIPE,
-            )
+            try:
+                process = await asyncio.create_subprocess_exec(
+                    *cmd_with_args,
+                    stdin=asyncio.subprocess.PIPE if stdin else None,
+                    stdout=asyncio.subprocess.PIPE,
+                    stderr=asyncio.subprocess.PIPE,
+                )
+            except FileNotFoundError as e:
+                return (False, str(e))
+
             stdout, stderr = await process.communicate(stdin)
 
             assert process.returncode is not None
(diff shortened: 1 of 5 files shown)


src/ert/scheduler/driver.py
```diff
@@ -90,12 +90,16 @@ async def _execute_with_retry(
...
 
         for _ in range(total_attempts):
-            process = await asyncio.create_subprocess_exec(
-                *cmd_with_args,
-                stdin=asyncio.subprocess.PIPE if stdin else None,
-                stdout=asyncio.subprocess.PIPE,
-                stderr=asyncio.subprocess.PIPE,
-            )
+            try:
+                process = await asyncio.create_subprocess_exec(
```
xjules, line 94: It seems that `create_subprocess_exec` will have the same treatment in the driver. Could we make this as an utility function, which we catch `FileNotFoundError` and return `False`?

tests/integration_tests/scheduler/test_generic_driver.py
```diff
@@ -212,3 +213,33 @@ async def test_num_cpu_sets_env_variables(driver: Driver, tmp_path, job_name):
...
+    caplog.set_level(logging.DEBUG)
+    invalid_cmd = ["/usr/bin/foo", "bar"]
+    driver._bjobs_cmd = invalid_cmd
+    driver._qstat_cmd = invalid_cmd
+    driver._squeue = invalid_cmd
+    driver._jobs = {"foo": "bar"}
+    driver._non_finished_job_ids = ["foo"]
+    await driver.poll()
+
+    assert "retry" not in str(caplog.text)
```
xjules, line 244: Don't understand this assert, maybe a comment that explains it (the one above too).

src/ert/scheduler/driver.py
jonathan-eq: I don't quite understand this. I could create a helper function that returns either a process or False, but I do not see the point of that

xjules (approved):
Nice one @jonathan-eq ! :rocket:

PR #8601, opened 2024-09-02
Backport scaling unused fix

yngve-sk:
Bugfix


oyvindeide (approved):
LGTM! Suggest squashing commits though, so the test accompanies the bug-fix

PR #8619, opened 2024-09-03
Fix general queue options not being added bug

### LINKED ISSUE #8605: Possible behaviour change for general queue options
User reports indicate that we have introduced a behaviour change in the configuration for the combination of:

```
QUEUE_OPTION LOCAL MAX_RUNNING 12
MAX_RUNNING 2
```

where now the first line has precedence over the general option. Since the first line can be (and currently is) present in site-config, this leads to the users experiencing that their `MAX_RUNNING 2` configuration line has no effect.

jonathan-eq:
**Issue**
Resolves #8605

**Approach**
🔧 

## When applicable

### DIFF
--- a/src/ert/config/ert_config.py
+++ b/src/ert/config/ert_config.py
@@ -159,7 +159,7 @@ class ErtConfig:
         Warnings will be issued with :python:`warnings.warn(category=ConfigWarning)`
         when the user should be notified with non-fatal configuration problems.
         """
-        user_config_dict = cls.read_user_config(user_config_file)
+        user_config_dict = cls.read_user_config_and_apply_site_config(user_config_file)
         config_dir = path.abspath(path.dirname(user_config_file))
         cls._log_config_file(user_config_file)
         cls._log_config_dict(user_config_dict)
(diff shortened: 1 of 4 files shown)


jonathan-eq:
The bug is occurring for the keywords max_running, submit_sleep, and project code.

jonathan-eq:
> project code

Unsure if project code is affected

src/ert/config/ert_config.py
```diff
@@ -372,12 +372,37 @@ def read_site_config(cls) -> ConfigDict:
...
+    ) -> ConfigDict:
+        site_config_dict = cls.read_site_config()
+        user_config_dict = cls.read_user_config(user_config_file)
+
+        for keyword, value in site_config_dict.items():
+            if keyword == "QUEUE_OPTION":
+                filtered_queue_options = []
+                for queue_option in value:
+                    if (
+                        "MAX_RUNNING" in queue_option
```
berland, line 389: This adds a maintenance problem when we add more queue options later. Can this be automated so we don't need to specifically test for MAX_RUNNING and SUBMIT_SLEEP here?

jonathan-eq: I am not sure, as those are the only options that can be both global and queue specific.

jonathan-eq: I can test if the string is outside of `QUEUE_OPTION`, but I wonder if this can cause other bugs which will be difficult to troubleshoot.

src/ert/config/ert_config.py
```diff
@@ -372,12 +372,32 @@ def read_site_config(cls) -> ConfigDict:
...
+        cls, user_config_file: str
+    ) -> ConfigDict:
+        site_config_dict = cls.read_site_config()
+        user_config_dict = cls.read_user_config(user_config_file)
+
+        for keyword, value in site_config_dict.items():
+            if keyword == "QUEUE_OPTION":
+                filtered_queue_options = []
+                for queue_option in value:
+                    option_name = queue_option[1]
```
jonathan-eq, line 388: I am not a fan of this. Maybe we can change QUEUE_OPTIONS to be `Dict[str, Dict[str, str]` instead of `List[List[str]]`?
Something like ```{
"LSF": {"bsub_cmd": "/path/to/cmd"}
}```

We do this somewhere later on, maybe it can be moved up here instead?
I assume this is due to legacy reasons @berland

jonathan-eq: It seems this would require a larger refactoring of ConfigDict, which will not be of priority right now 🌧️

berland: They are List[List[str]] initially so we can reason about the order of settings before we interpret them into a Dict.

src/ert/config/ert_config.py
```diff
@@ -372,12 +372,32 @@ def read_site_config(cls) -> ConfigDict:
...
+                    if option_name in user_config_dict:
+                        continue
+                    filtered_queue_options.append(queue_option)
+                user_config_dict["QUEUE_OPTION"] = (
```
xjules, line 392: I'm confused by this one. I thought we want to do it other way around; ie. overwriting the site_config with user config?

xjules: Resolved by discussion :)

tests/unit_tests/config/test_ert_config.py
```diff
@@ -1644,3 +1645,74 @@ def test_that_empty_params_file_gives_reasonable_error(tmpdir, param_config):
...
+        pytest.param(
+            """
+        QUEUE_OPTION TORQUE MAX_RUNNING 6
+        MAX_RUNNING 2
+        """,
+            id="queue_option_max_running",
+        ),
+    ],
+)
+def test_queue_config_general_max_running_takes_precedence_over_queue_option(
```
xjules, line 1671: Hm, is it really so? I thought that the queue_option takes precedence ... always

xjules:
Looks very good @jonathan-eq ! Could you squash the commits and improve the commit message?

xjules (approved):
Nice work! Let's bring it in :rocket:

PR #8623, opened 2024-09-03
Delete get_ensemble_by_name from `LocalStorage`

dafeda:
Ensemble names are only unique within a single experiment.

## Breaking Change

- Will cause `semeio` to fail, so this PR will need to be merged: https://github.com/equinor/semeio/pull/640
-  Will cause `everest to fail


DanSava:
Are experiment names unique?

dafeda:
> Are experiment names unique?

Yes they are. One reason why want this is that it makes it easier to read plots since we use `experiment_name:ensemble_name`.

src/ert/storage/migration/block_fs.py
```diff
@@ -314,6 +314,7 @@ def migrate_case(storage: Storage, path: Path, stack: ExitStack) -> None:
         parameters=parameter_configs,
         responses=response_configs,
         observations=ert_config.observations,
+        name="migrate-case",
```
yngve-sk, line 317: Should this be made unique or is it ok for cases to be migrated (from blockfs in this case so maybe no that many) w/ the same experiment name? Logging statement above is
`    logger.info(f"Migrating case '{path.name}'")`

DanSava:
> > Are experiment names unique?
> 
> Yes they are. One reason why want this is that it makes it easier to read plots since we use `experiment_name:ensemble_name`.

I could not find any logic in ERT that enforces the experiment name to be unique, but it might be that I did not look hard enough. 

If the experiment name is unique and this is checked when a new experiment is created then to make Everest work with these changes we would need to update:

*  `ert/src/ert/simulator/batch_simulator.py`

```python
    parameters=self.ert_config.ensemble_config.parameter_configuration,
    responses=self.ert_config.ensemble_config.response_configuration,
)
```
to 
```python
    parameters=self.ert_config.ensemble_config.parameter_configuration,
    responses=self.ert_config.ensemble_config.response_configuration,
    name=f"experiment_{case_name}"
)
```

yngve-sk (approved):
Recommend double checking wrt uniqueness of experiment name in some edge cases, otherwise LGTM

dafeda:
> > > Are experiment names unique?
> > 
> > 
> > Yes they are. One reason why want this is that it makes it easier to read plots since we use `experiment_name:ensemble_name`.
> 
> I could not find any logic in ERT that enforces the experiment name to be unique, but it might be that I did not look hard enough.
> 
> If the experiment name is unique and this is checked when a new experiment is created then to make Everest work with these changes we would need to update:
> 
> * `ert/src/ert/simulator/batch_simulator.py`
> 
> ```python
> experiment = storage.create_experiment(
>     parameters=self.ert_config.ensemble_config.parameter_configuration,
>     responses=self.ert_config.ensemble_config.response_configuration,
> )
> ```
> 
> to
> 
> ```python
> experiment = storage.create_experiment(
>     parameters=self.ert_config.ensemble_config.parameter_configuration,
>     responses=self.ert_config.ensemble_config.response_configuration,
>     name=f"experiment_{case_name}"
> )
> ```

You are right I think, the validation is only in the front-end. @larsevj is working on making sure it's not possible to create experiments with non-unique names I think.

src/ert/storage/migration/block_fs.py
dafeda: Valid point, but this not change behavior as the current code does this when no name is specified:

`name = datetime.today().strftime("%Y-%m-%d")`

Will leave as is for now. OK?

PR #8643, opened 2024-09-05
Get ensembles by id in dark storage

### LINKED ISSUE #8641: Plots are overlapping w/ overlapping ensemble names

If we have N similarly named ensembles and only one of them has a response, the response will be plotted for all the ensembles, as dark storage currently looks up data by ensemble name. Solution is to make dark storage ensemble lookup logic use ensemble id and not name

_No response_

### steps to reproduce

_No response_

- [ ] python 3.8
- [ ] python 3.9
- [ ] python 3.10
- [ ] python 3.11
- [ ] python 3.12
- [ ] macosx
- [ ] rhel7
- [ ] rhel8
- [ ] local queue
- [ ] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

yngve-sk:
**Issue**
Resolves #8641

### DIFF
--- a/src/ert/gui/tools/plot/plot_api.py
+++ b/src/ert/gui/tools/plot/plot_api.py
@@ -149,7 +149,7 @@ class PlotApi:
 
         return list(all_keys.values())
 
-    def data_for_key(self, ensemble_name: str, key: str) -> pd.DataFrame:
+    def data_for_key(self, ensemble_id: str, key: str) -> pd.DataFrame:
         """Returns a pandas DataFrame with the datapoints for a given key for a given
         ensemble. The row index is the realization number, and the columns are an index
         over the indexes/dates"""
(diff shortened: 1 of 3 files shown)


src/ert/gui/tools/plot/plot_api.py
```diff
@@ -157,7 +157,7 @@ def data_for_key(self, ensemble_name: str, key: str) -> pd.DataFrame:
         if key.startswith("LOG10_"):
             key = key[6:]
 
-        ensemble = self._get_ensemble(ensemble_name)
+        ensemble = self._get_ensemble_by_id(ensemble_name)
```
DanSava, line 160: Can you also change the function signature for `data_for_key`?

DanSava (approved):
Looks good one small comment!

src/ert/gui/tools/plot/plot_api.py
yngve-sk: `data_for_key` takes in the entire ensemble at the moment, and has 8 usages, so I think that could be a refactor for another PR as this is just targeted towards fixing the bug w/ overlapping plots.

PR #8661, opened 2024-09-06
Include experiment name in ensemble selector

### LINKED ISSUE #8614: Impossible to distinguish between ensembles to restart from in es_mda

There is nothing stopping you from running two separate experiments where you set the same target ensemble format in both experiments. When doing a restart run in es_mda it will then be impossible to distinguish between the ensembles that stems from different experiments. Note that in the plotter you can see the difference as then the ensembles are prefixed with the experiment name.

image

image

Would expect to tell the difference across experiments, similar to the plotter.

### steps to reproduce

```Python
ert gui poly.ert
run es_mda
run es_mda again with Target ensemble set to same as before
Check restart run, and press the restart from dropdown
```

- [ ] python 3.8
- [ ] python 3.9
- [ ] python 3.10
- [X] python 3.11
- [ ] python 3.12
- [ ] macosx
- [ ] rhel7
- [X] rhel8
- [X] local queue
- [ ] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

larsevj:
**Issue**
Resolves #8614 

Results in the following:

image
Ignore validation error.

## When applicable

### DIFF
--- a/src/ert/gui/ertwidgets/ensembleselector.py
+++ b/src/ert/gui/ertwidgets/ensembleselector.py
@@ -67,7 +67,9 @@ class EnsembleSelector(QComboBox):
             self.setEnabled(True)
 
         for ensemble in self._ensemble_list():
-            self.addItem(ensemble.name, userData=ensemble)
+            self.addItem(
+                f"{ensemble.experiment.name} : {ensemble.name}", userData=ensemble
+            )
 
         current_index = self.findData(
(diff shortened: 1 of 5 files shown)


jonathan-eq:
Do you have a screenshot of how it looks with this commit?

jonathan-eq (approved):
Looks excellent 🌵! Remember to squash commits before merging

PR #8682, opened 2024-09-11
Make copy_directories error message readable

### LINKED ISSUE #8660: COPY_DIRECTORY produces malformed error message
COPY_FILE can produce the following malformed error message:

```
COPY_DIRECTORY failed with the following error: ['<', 'D', 'i', 'r', 'E', 'n', 't', 'r', 'y', ' ', "'", 's', 'e', 'i', 's', 'm', 'i', 'c', 'l', 'a', 'y', 'o', 'u', 't', '.', 's', 'e', 'g', 'y', "'", '>', ' ', 'a', 'n', 'd', ' ', "'", '/', 's', 'c', 'r', 'a', 't', 'c', 'h', '/', 'f', 'm', 'u', '/',..., ' ', 'a', 'r', 'e', ' ', 't', 'h', 'e', ' ', 's', 'a', 'm', 'e', ' ', 'f', 'i', 'l', 'e']
```

andreas-el:
**Issue**
Resolves #8660 

The error would only occur if copying a folder containing a file.

## When applicable

### DIFF
--- a/src/ert/resources/shell_scripts/copy_directory.py
+++ b/src/ert/resources/shell_scripts/copy_directory.py
@@ -32,4 +32,6 @@ if __name__ == "__main__":
     try:
         copy_directory(src_path, target_path)
     except IOError as e:
-        sys.exit(f"COPY_DIRECTORY failed with the following error: {e}")
+        sys.exit(
+            f"COPY_DIRECTORY failed with the following error: {''.join(e.args[0])}"
+        )
--- a/tests/unit_tests/shared/share/test_shell.py
+++ b/tests/unit_tests/shared/share/test_shell.py
@@ -400,6 +400,17 @@ def test_copy_directory_error(shell):
 
     assert b"existing directory" in shell.copy_directory("hei", "target").stderr
 
+    empty_dir = "emptytestdir"
+    if not os.path.exists(empty_dir):
+        os.makedirs(empty_dir)
+
+    file_path = os.path.join(empty_dir, "file")
+
+    with open(file_path, "w", encoding="utf-8") as f:
+        f.write("some_text")
+
+    assert b"are the same" in shell.copy_directory(empty_dir, ".").stderr
+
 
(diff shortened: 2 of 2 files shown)


src/ert/resources/shell_scripts/copy_directory.py
```diff
@@ -32,4 +32,6 @@ def copy_directory(src_path, target_path):
     try:
         copy_directory(src_path, target_path)
     except IOError as e:
-        sys.exit(f"COPY_DIRECTORY failed with the following error: {e}")
+        sys.exit(
+            f"COPY_DIRECTORY failed with the following error: {''.join(e.args[0])}"
```
berland, line 36: This is a little bit disturbing that we cannot do `{e}` on an IOError exception. Did you understand why this was needed?

andreas-el: No, not other than I can guarantee that this comes directly from
```
shutil.copytree(src_path, target_path, dirs_exist_ok=True)
```
So this is some built-in weird behaviour it seems.

It is connected to this `dirs_exist_ok` argument, and when I change that to its default (`False`), the exception becomes more normal again; `FileExistsError(17, 'File exists')`

berland: 👍🏻

berland (approved):
go

src/ert/resources/shell_scripts/copy_directory.py
eivindjahren: I think the problem may be upstream.

eivindjahren: Do we have tests that the other IOErrors are not mangled by this change?

andreas-el:
This issue is also reported elsewhere, but not merged yet (as of time of writing)
https://github.com/python/cpython/issues/102931

PR #8696, opened 2024-09-13
Log contents of custom forward model steps

berland:
**Issue**
Resolves https://github.com/equinor/scout/issues/1132

**Approach**

## When applicable

### DIFF
--- a/src/ert/config/ert_config.py
+++ b/src/ert/config/ert_config.py
@@ -349,6 +349,14 @@ class ErtConfig:
 
         logger.info(f"Content of the config_dict: {tmp_dict}")
 
+    @classmethod
+    def _log_custom_forward_model_steps(cls, user_config: ConfigDict) -> None:
+        for fm_step, fm_step_filename in user_config.get(ConfigKeys.INSTALL_JOB, []):
+            fm_configuration = Path(fm_step_filename).read_text(encoding="utf-8")
+            logger.info(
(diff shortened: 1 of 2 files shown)


src/ert/config/ert_config.py
```diff
@@ -349,6 +350,18 @@ def _log_config_dict(cls, content_dict: Dict[str, Any]) -> None:
...
         logger.info(f"Content of the config_dict: {tmp_dict}")
 
+    @classmethod
+    def _log_custom_forward_model_steps(cls, config_file: str) -> None:
+        user_config = lark_parse(
+            file=config_file,
+            schema=init_user_config_schema(),
+        )
+        for fm_step, fm_step_filename in user_config.get(ConfigKeys.INSTALL_JOB, []):
+            fm_configuration = Path(fm_step_filename).read_text(encoding="utf-8")
```
eivindjahren, line 360: This should probably be logged after the `ErtConfig` is parsed to avoid 1) parsing the config file twice and 2) avoid raising `FileNotFoundError` instead of `ConfigurationValidationError`.

eivindjahren (changes requested):

src/ert/config/ert_config.py
berland: Fixed (methinks)

eivindjahren: Just checked and bad permissions for the config file and non-existent config file is handled by `__main__.py`.

PR #8704, opened 2024-09-16
Remove unused index_lists

yngve-sk:
Not used outside of tests AFAIK. Did not find any instances of initialized `MeasuredData` with `index_lists` argument


oyvindeide (approved):
LGTM!

PR #8734, opened 2024-09-18
Add ForwardModelStepValidationWarning class

jonathan-eq:
**Issue**
Resolves #my_issue

**Approach**

      --exec 'pytest tests/unit_tests -n logical -m "not integration_test"'`)

## When applicable


src/ert/config/ert_config.py
```diff
@@ -688,6 +691,8 @@ def handle_default(fm_step: ForwardModelStep, arg: str) -> str:
                         f"forward model step {fm_step.name}: {str(exc)}"
                     ).set_context(fm_step.name)
                 )
+            except ForwardModelStepValidationWarning as exc:
```
eivindjahren, line 694: Instead of being thrown we would like this to be a warning category so that we can

```python
import warnings

warnings.warn(ForwardModelStepWarning("message"))
```

It would then be enough for ForwardModelStepWarning to be a subtype of ConfigWarning:

```
class ForwardModelStepWarning(ConfigWarning)
```

eivindjahren (changes requested):

src/ert/config/ert_config.py
jonathan-eq: Yes, I think thats a good idea

eivindjahren:
Looks, good, but should probably be added to the plugin documentation.

eivindjahren (approved):
:+1:

PR #8735, opened 2024-09-18
Have ForwardModelRunner validate executable_file earlier

### LINKED ISSUE #8729: Missing validation when using ForwardModelStepPlugin causes ert to hang

Installed a forward model step using ForwardModelStepPlugin. Tested without having the executable file available. Expected ert to fail:
Screenshot 2024-09-18 at 09 59 02

If I install using INSTALL_JOB in the config file it is checked and ert stops with a "could not find executable".

If I try to run it through the gui the progress stops at Pending (the error message appears in the terminal that starts ert)

Ert should stop on failure.

### steps to reproduce

```Python
Create dummy ForwardModelStep hook with a non-existing file as executable and run.
```

- [ ] python 3.8
- [ ] python 3.9
- [ ] python 3.10
- [ ] python 3.11
- [ ] python 3.12
- [ ] macosx
- [ ] rhel7
- [ ] rhel8
- [x] local queue
- [ ] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

jonathan-eq:
This commit fixes the issue where the ForwardModelRunner on the job_runner side would crash when validating the existence of a ForwardModelStep executable; causing the drivers to hang while waiting for a never coming return code from JobRunner.

**Issue**
Resolves #8729

**Approach**
⛄ 

image

      --exec 'pytest tests/unit_tests -n logical -m "not integration_test"'`)

## When applicable

### DIFF
--- a/src/_ert/forward_model_runner/io/__init__.py
+++ b/src/_ert/forward_model_runner/io/__init__.py
@@ -6,8 +6,8 @@ def cond_unlink(file):
         os.unlink(file)
 
 
-def assert_file_executable(fname):
-    """The function raises an IOError if the given file is either not a file or
+def check_executable(fname) -> str:
+    """The function returns an error message if the given file is either not a file or
     not an executable.
 
     If the given file name is an absolute path, its functionality is straight
@@ -17,7 +17,7 @@ def assert_file_executable(fname):
 
     """
     if not fname:
-        raise IOError("No executable provided!")
+        return "No executable provided!"
     fname = os.path.expanduser(fname)
 
     potential_executables = [os.path.abspath(fname)]
@@ -28,7 +28,8 @@ def assert_file_executable(fname):
         ]
 
     if not any(map(os.path.isfile, potential_executables)):
-        raise IOError(f"{fname} is not a file!")
+        return f"{fname} is not a file!"
 
     if not any(os.access(fn, os.X_OK) for fn in potential_executables):
-        raise IOError(f"{fname} is not an executable!")
+        return f"{fname} is not an executable!"
+    return ""
--- a/src/_ert/forward_model_runner/job.py
+++ b/src/_ert/forward_model_runner/job.py
@@ -14,7 +14,7 @@ from typing import Optional, Tuple
 
 from psutil import AccessDenied, NoSuchProcess, Process, TimeoutExpired, ZombieProcess
 
-from .io import assert_file_executable
+from .io import check_executable
 from .reporting.message import (
     Exited,
     ProcessTreeStatus,
@@ -97,10 +97,7 @@ class Job:
 
         yield start_message
 
-        executable = self.job_data.get("executable")
-        assert_file_executable(executable)
-
-        arg_list = [executable]
+        arg_list = [self.job_data.get("executable")]
         if self.job_data.get("argList"):
             arg_list += self.job_data["argList"]
 
@@ -313,6 +310,9 @@ class Job:
             self.job_data.get("error_file")
         ):
             os.unlink(self.job_data.get("error_file"))
+
+        if executable_error := check_executable(self.job_data.get("executable")):
+            errors.append(str(executable_error))
 
         return errors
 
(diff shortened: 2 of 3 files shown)


yngve-sk:
Is there some reason why this couldn't/shouldn't be done in the config parsing stage, and on the job_runner side? It could be added in `ert_config.py` somewhere, in this segment and then also show that executable in the GUI before going further
```
        for fm_step in fm_steps:
            if fm_step.name in cls.PREINSTALLED_FORWARD_MODEL_STEPS:
                try:
                    substituted_json = cls._create_forward_model_json(
                        run_id=None,
                        context=substitution_list,
                        forward_model_steps=[fm_step],
                        skip_pre_experiment_validation=True,
                    )
                    job_json = substituted_json["jobList"][0]
                    fm_step.validate_pre_experiment(job_json)
                except ForwardModelStepValidationError as err:
                    errors.append(
                        ConfigValidationError.with_context(
                            f"Forward model step pre-experiment validation failed: {err!s}",
                            context=fm_step.name,
                        ),
                    )
```
really we want users to add their own, but validating that an executable exists is pretty elementary. Also an edge case; if we expect the executable to be created by a preceding forward model (or maybe we want to disallow that, not sure).

jonathan-eq:
> Is there some reason why this couldn't/shouldn't be done in the config parsing stage, and on the job_runner side? It could be added in `ert_config.py` somewhere, in this segment and then also show that executable in the GUI before going further
> 
> ```
>         for fm_step in fm_steps:
>             if fm_step.name in cls.PREINSTALLED_FORWARD_MODEL_STEPS:
>                 try:
>                     substituted_json = cls._create_forward_model_json(
>                         run_id=None,
>                         context=substitution_list,
>                         forward_model_steps=[fm_step],
>                         skip_pre_experiment_validation=True,
>                     )
>                     job_json = substituted_json["jobList"][0]
>                     fm_step.validate_pre_experiment(job_json)
>                 except ForwardModelStepValidationError as err:
>                     errors.append(
>                         ConfigValidationError.with_context(
>                             f"Forward model step pre-experiment validation failed: {err!s}",
>                             context=fm_step.name,
>                         ),
>                     )
> ```
> 
> really we want users to add their own, but validating that an executable exists is pretty elementary. Also an edge case; if we expect the executable to be created by a preceding forward model (or maybe we want to disallow that, not sure).

Yes, but it is unclear if this is ert's responsibility, or if the forward model step should have already validated itself. 
We had a discussion about it, but it is nonetheless a separate PR

yngve-sk:
> model step should have already validated itself.
> We had a discussion about it, but it is nonetheless a separate PR

I see. Just some context for this: In the ForwardModelStepPlugin we did want users to do their own validation, so we omitted providing default validations for argument types etc. But at the same time we may have just "forgotten" to add the most basic validation (that the executable exists). Non-plugin fms validate this and other things, so it might just make sense to do it there.

src/_ert/forward_model_runner/io/__init__.py
```diff
@@ -6,7 +6,7 @@ def cond_unlink(file):
         os.unlink(file)
 
 
-def assert_file_executable(fname):
+def check_executable(fname) -> str:
     """The function raises an IOError if the given file is either not a file or
```
xjules, line 10: The comment is wrong then ....

xjules:
> > model step should have already validated itself.
> > We had a discussion about it, but it is nonetheless a separate PR
> 
> I see. Just some context for this: In the ForwardModelStepPlugin we did want users to do their own validation, so we omitted providing default validations for argument types etc. But at the same time we may have just "forgotten" to add the most basic validation (that the executable exists). Non-plugin fms validate this and other things, so it might just make sense to do it there.

Actually I agree having the basic `file exists` validation makes sense while all FMplugin specific validation should be up to the user.

jonathan-eq:
> As suggested by @yngve-sk, let's add the "basic" file-exist validation.

Can we have this as a separate issue?
Yngve mentioned that the executable could be created by a preceding forward model, so we would emit a warning at most

xjules:
> > As suggested by @yngve-sk, let's add the "basic" file-exist validation.
> 
> Can we have this as a separate issue? Yngve mentioned that the executable could be created by a preceding forward model, so we would emit a warning at most

Ok, then create one (as you have more insight ;)) and mention it also in this PR

jonathan-eq:
https://github.com/equinor/ert/issues/8779

xjules (approved):
Very nice! I think it should do :rocket:

PR #8737, opened 2024-09-19
Stop merging dark storage observations

yngve-sk:
(Tested w/ Webviz ert, looks the same in Observation Analyzer and tests pass)
Current behavior concatenates all observations into one and then gets one "observation name" from the list of all "observation names" which barely makes sense.

This data goes AFAIK only to `src/ert/gui/plottery/plots/observations.py:_plotObservations`, from `PlotWindow.updatePlot -> PlotApi.observations_for_key -> record.py:get_record_observations`, where it does not even directly use the name.

Behavior in plotter seems similar.
Screenshot 2024-09-19 at 10 37 05
Screenshot 2024-09-19 at 10 39 09

### DIFF
--- a/src/ert/dark_storage/common.py
+++ b/src/ert/dark_storage/common.py
@@ -208,16 +208,6 @@ def get_observations_for_obs_keys(
     return observations
 
 
-def get_observation_name(ensemble: Ensemble, observation_keys: List[str]) -> str:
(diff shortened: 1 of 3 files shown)


DanSava (approved):
Look good :mechanic:

PR #8790, opened 2024-09-24
Avoid bsub retrying on known error messages

### LINKED ISSUE #8774: LSF driver should not retry submission on "No such queue"
In the logs, an exception has been observed:

`ete_2409/realization-99/iter-0" failed after 10 attempts with exit code 255, output: "<empty>", and error: "rhel7: No such queue. Job not submitted."`

This is weird. It should only do 10 attempts on the flaky ssh returncode which is 255, but that error code should not be associated with the message "No such queue". 

When "No such queue" is obtained, it should end up with no submission retries for the realization.

berland:
Contrary to earlier belief in the codebase, returncode 255 does not only mean flaky ssh connection. In order to no retry on known error scenarios, we must detect them explicitly by string matcing.

**Issue**
Resolves #8774 

**Approach**
Provoke different errors, note the error code and the error message, and ensure these scenarios are correctly handled through test code.

## When applicable

### DIFF
--- a/src/ert/scheduler/driver.py
+++ b/src/ert/scheduler/driver.py
@@ -83,7 +83,8 @@ class Driver(ABC):
         total_attempts: int = 1,
         retry_interval: float = 1.0,
         driverlogger: Optional[logging.Logger] = None,
-        exit_on_msgs: Iterable[str] = (),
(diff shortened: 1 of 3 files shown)


eivindjahren (approved):
:+1:

PR #8802, opened 2024-09-26
Fix an issue where aborted processes could corrupt storage

### LINKED ISSUE #8799: Implement transactional writes to storage
When doing

```python
with open("file.txt", "w") as f:
    pass
```

The "file.txt" is created

This also applies to "file2.txt" in the following coe

```python
with open("file2.txt", "w") as f:
    raise ValueError()
```

This way of writing files is used throughout `src/ert/storage`

https://github.com/equinor/ert/blob/21af32915792c103bab4bb6708e1cf861719aea8/src/ert/storage/local_experiment.py#L132-L133

https://github.com/equinor/ert/blob/21af32915792c103bab4bb6708e1cf861719aea8/src/ert/storage/local_experiment.py#L138-L139

https://github.com/equinor/ert/blob/21af32915792c103bab4bb6708e1cf861719aea8/src/ert/storage/local_experiment.py#L147-L149

https://github.com/equinor/ert/blob/21af32915792c103bab4bb6708e1cf861719aea8/src/ert/storage/local_experiment.py#L221-L222

https://github.com/equinor/ert/blob/21af32915792c103bab4bb6708e1cf861719aea8/src/ert/storage/local_experiment.py#L246-L247

And many more.

This causes concurrency  and corruption problems such as https://github.com/equinor/ert/issues/8775

Instead, we need some sort of mechanism to prevent inconsistent files from being read. One way is to create a lock file before writing and than cleaning it up after. This means that one will have to consider what to do when a lock file is present whenever reading and writing, so quite complicated.

A better solution is to rely on OS file system renames being atomic.

eivindjahren:
**Issue**
Resolves #8799 

## When applicable


tests/ert/unit_tests/storage/test_local_storage.py
```diff
@@ -483,6 +483,60 @@ def fields(draw, egrid, num_fields=small_ints) -> List[Field]:
...
 
 
+@pytest.mark.usefixtures("use_tmpdir")
+@given(st.binary())
+def test_write_transaction(data):
+    with open_storage(".", "w") as storage:
+        filepath = Path("./file.txt")
+        storage._write_transaction(filepath, data)
+
+        assert filepath.read_bytes() == data
```
berland, line 493: deindent one level to make sure the context manager does not delete it (which I assume it is not meant to do)?

berland: ah, that context manager is probably irrelevant to _write_transaction.

berland (approved):
Nice!

Issue #999, opened 2020-09-01
[GUI] Expose compute node hostname in simulation details


berland:
Here it could be nice to expose the hostname where the simulation is running:

![image](https://user-images.githubusercontent.com/306834/91828204-1b66b780-ec40-11ea-8fca-810210b985c9.png)

Currently, this information is determined from the first line in `STATUS`

berland:
While at it, perhaps more realization metadata could be contemplated; RUNPATH,  job queue name, LSF log files,

PR #8824, opened 2024-09-30
Propagate and show lsf exec_hosts in run_dialog label

andreas-el:
**Issue**
Resolves https://github.com/equinor/ert/issues/999

Screenshot 2024-10-04 at 15 11 09

## When applicable

### DIFF
--- a/src/_ert/events.py
+++ b/src/_ert/events.py
@@ -110,6 +110,7 @@ class RealizationBaseEvent(BaseEvent):
     real: str
     ensemble: Union[str, None] = None
     queue_event_type: Union[str, None] = None
+    exec_hosts: Union[str, None] = None
 
 
 class RealizationPending(RealizationBaseEvent):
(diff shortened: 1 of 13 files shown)


berland:
Should we do an attempt to obtain a FQDN when printing the hostnames?

andreas-el:
> Should we do an attempt to obtain a FQDN when printing the hostnames?

We could.
Currently bjobs will output like this;
```
(rhel8-bleed-py38-env + bleeding-py38-rhel8) [andrli@st-lintgx0264 komodo]$ bjobs -noheader -o "jobid stat exec_host" 568264
568264 RUN 4*st-rsb17-17-04
```

Do note that `4*st-rsb17-17-04` is not a valid hostname, but I thought this would be useful to see how many slots were assigned on that specific node.

berland:
> Do note that `4*st-rsb17-17-04` is not a valid hostname, but I thought this would be useful to see how many slots were assigned on that specific node.

Right. Maybe just better print the string from bjobs to avoid having to parse something we don't really control.

andreas-el:
There was an additional comment on the original issue;

> While at it, perhaps more realization metadata could be contemplated; RUNPATH, job queue name, LSF log files,

In my opinion the most valuable of these might be the `job_queue_name` ?
The process of adding this will be identical to what was done for `exec_hosts`, even so for the tests, so this should be very low hanging fruit atm.

berland (approved):
💯

PR #8832, opened 2024-10-01
Turn unexpected fm errors into warnings

### LINKED ISSUE #8830: When validating forward model step plugins, other exceptions should be caught and shown as warnings
If a forward model step plugin has a bug in its validation an raises any other exception than `ForwardModelValidationError` then this should be shown as a warning to the user and logged so that we can give a bug report to the responsible team.

https://github.com/equinor/ert/blob/e92f558cf14e9c9823b557b3df05caefef49576f/src/ert/config/ert_config.py#L595-L610

yngve-sk:
**Issue**
Resolves #8830

### DIFF
--- a/src/ert/config/ert_config.py
+++ b/src/ert/config/ert_config.py
@@ -609,9 +609,9 @@ class ErtConfig:
                         ),
                     )
                 except Exception as e:  # type: ignore
-                    ConfigWarning.warn(str(e), context=fm_step.name)
-                    logger.warning(
-                        f"Unexpected plugin forward model exception:" f"{e!s}"
+                    ConfigWarning.warn(
+                        f"Unexpected plugin forward model exception: " f"{e!s}",
+                        context=fm_step.name,
                     )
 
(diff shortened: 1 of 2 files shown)


src/ert/config/ert_config.py
```diff
@@ -608,6 +608,11 @@ def _create_list_of_forward_model_steps_to_run(
                             context=fm_step.name,
                         ),
                     )
+                except Exception as e:  # type: ignore
+                    ConfigWarning.warn(str(e), context=fm_step.name)
```
eivindjahren, line 612: I think both the log and the warning should have the text `Unexpected exception in forward model validation of {fm_step.name}: ` as all `ConfigWarning`s are logged in the `Suggestor` we don't need an additional log statement.

eivindjahren (changes requested):
Just a smalll adjustment to the message shown, otherwise it is good.

eivindjahren (approved):
:+1:

PR #8853, opened 2024-10-03
Return with SIGTERM in mocked kill

berland:
**Issue**
Resolves test failure in https://github.com/equinor/komodo-releases/actions/runs/11153714108/job/31002348945

**Approach**
When the SIGTERM signal is trapped, return SIGTERM as the error code instead of -1 (which is translated to 255 in the log above). The code passed a couple of trials before failing because the test often triggers 143 as an error code.

## When applicable


andreas-el (approved):
This is the correct value. 🔪

PR #8873, opened 2024-10-04
Resolve flakyness in test_kill_before_submit

berland:
Sometimes the job actually starts on a compute node and sometimes it does not before it is removed/killed. In the the latter case, code previous to this commit will fail as it asserts that the job script has actually started to run.

**Issue**
Resolves flakyness.

**Approach**
Don't wait for file when it should not be there.

## When applicable


xjules (approved):
Nice! 🚀

PR #8899, opened 2024-10-07
Refactor BatchSimulator

### LINKED ISSUE #8753: BatchSimulator is "injecting" stuff into ert config object
This 
```
        ens_config = self.ert_config.ensemble_config
        for control_name, variables in controls.items():
            ens_config.addNode(
                ExtParamConfig(
                    name=control_name,
                    input_keys=variables,
                    output_file=control_name + ".json",
                )
            )

        for key in results:
            ens_config.addNode(GenDataConfig(name=key, input_file=f"{key}"))
```

could and should be done in `everest_to_ert` where the ert config is created. 

Can be "recreated" by running `test_math_func` and breakpointing in Simulator.py

verveerpj:
**Issue**
Resolves #8753

We probably will replace the `BatchSimulator `in the future with other ERT functionality, so this PR might not be entirely necessary. Nevertheless, I think the refactored code is clearer which may be useful when we do the work to replace `BatchSimulator`.

**Approach**
- Add some `GEN_DATA` keys to the generated ert config
-  Add some code to handle `ExtParamConfig` parameter configs

**Note**
I handle `ExtParamConfig` by parsing them in `ensemble_config.py`. Currently it seems that `ExtParamConfig` does not have a corresponding keyword, which would make sense if we assume this is only used in the ert configs generated by everest. For this reason I just drop the data to be added in the generated ert config and generate the corresponding `ExtParamConfig` objects in `ExtParamConfig.from_dict`. We should discuss if that is an acceptable solution.

One consequence of this PR is that a lot of tests break that assume that the `BatchSimulator` can be initialized from an ERT config file. That is not possible anymore, presumably the `GEN_DATA` could be added to ERT config files, but the `ExtParamConfig` data cannot. Rather than changing the tests, I chose to patch the BatchSimulator class in the file to recover the old behavior, so that the batch simulator itself can still be tested.

## When applicable

### DIFF
--- a/tests/ert/unit_tests/simulator/test_batch_sim.py
+++ b/tests/ert/unit_tests/simulator/test_batch_sim.py
@@ -4,7 +4,7 @@ import time
 
 import pytest
 
-from ert.config import ErtConfig
+from ert.config import ErtConfig, ExtParamConfig, GenDataConfig
 from ert.scheduler import JobState
 from ert.simulator import BatchContext, BatchSimulator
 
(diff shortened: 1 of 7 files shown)


yngve-sk:
I think overall it makes sense, the `GEN_DATA` part I would approve. But I'm not quite sure about putting `ExtParam` into `EnsembleConfig`. Depends a bit on how we want to set things up, and I think (not totally sure on this) that if we can make the `ExtParam` do its thing without touching the `EnsembleConfig` that is most ideal, though it also isn't doing any harm by being in there (as far as I can see right now). Might become more clear when we refactor the detached server, perhaps it could be initialized separately there. I'm not totally sure, might want to discuss that part some more.

verveerpj:
>But I'm not quite sure about putting some of the `ExtParam` logic into `EnsembleConfig`. Depends a bit on how we want to set things up, and I think (not totally sure on this) that if we can make the `ExtParam` do its thing without touching the `EnsembleConfig` that is most ideal, though it also isn't doing any harm by being in there (as far as I can see right now). Might become more clear when we refactor the detached server, perhaps it could be initialized separately there. I'm not totally sure, might want to discuss that part some more.

I agree, I was also not completely comfortable with it. I have now pushed a commit that undoes the changes to `EnsembleConfig`. this means that it is now necessary to inject some things after making the ERT config. But I am now doing that in the Simulator class instead of the BatchSimulator class and that helps to keep most of the refactored code. Let me know what you think.

tests/ert/unit_tests/simulator/test_batch_sim.py
```diff
@@ -35,6 +35,51 @@ def batch_sim_example(setup_case):
     return setup_case("batch_sim", "batch_sim.ert")
 
 
+# TODO: The batch simulator was recently refactored. It now requires an ERT
```
yngve-sk, line 38: Should probably remove the `TODO: ` but leave the rest of the comment in place

PR #8922, opened 2024-10-09
Remove broadcasting of Eclipse stdout to multiple files

### LINKED ISSUE #8920: Stop broadcasting Eclipse stdout to multiple files
There is code in Ert to broadcast stdout from Eclipse both to the Ert standard stdout location (which the GUI reads) and also to a file next to the DATA file called either LOG or OUT (depending on the Eclipse version).

This extra file (LOG/OUT) seems to be the result of trying to be consistent with what `eclrun` (from SLB) does. Ert only uses eclrun now, and should not do anything about this matter. If eclrun wants to copy stdout to a dedicated file, it is free to do so without Ert interfering. 

The current `eclrun` seems not to write stdout to any dedicated file at all.

Refer to https://github.com/equinor/ert/issues/1911

berland:
Such broadcasting should be handled downstream by eclrun

**Issue**
Resolves #8920 

**Approach**
✂️ 

## When applicable


berland:
Manually tested onprem with ert-configurations installed on test-data/ert/flow_example with flow.ert modified to use ECLIPSE100. 

Ensemble experiment runs through, stdout is provided in ECLIPSE100.stdout.1 (which is the one read by the GUI) and there is no OUT or LOG file. That means that `eclrun` has changed behaviour somewhere between 2021 and now and we will follow suit.

eivindjahren (approved):

eivindjahren:
If this works as intended with ecl versions after 2021, I see no problem with it :)

PR #8935, opened 2024-10-10
Fix run_dialog second progress bar hanging

### LINKED ISSUE #8912: On experiment failure the second progress bar is stuck in indeterminate state

image

Experiment failure should be shown as all failed

Image

### steps to reproduce

```bash
cd test-data/ert/poly_example
mkdir poly_out
chmod 111 poly_out
ert gui poly.ert
```
Then run an ensemble_experiment. It should result in an experiment failure with the above image.

jonathan-eq:
This commit fixes the issue where an experiment failure would leave the second progress bar stuck in indeterminate state. This commit makes it so that second progress bar is invisible as long as there are no realizations or if all realizations are finished.

**Issue**
Resolves #8912

**Approach**

## When applicable

### DIFF
--- a/src/ert/gui/simulation/view/progress_widget.py
+++ b/src/ert/gui/simulation/view/progress_widget.py
@@ -74,7 +74,6 @@ class ProgressWidget(QFrame):
     def repaint_components(self) -> None:
         if self._realization_count > 0:
             full_width = self.width()
-            self.stop_waiting_progress_bar()
 
             for state, label in self._progress_label_map.items():
                 label.setVisible(True)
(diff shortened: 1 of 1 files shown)


src/ert/gui/simulation/view/progress_widget.py
```diff
@@ -90,9 +89,16 @@ def repaint_components(self) -> None:
     def stop_waiting_progress_bar(self) -> None:
         self._waiting_progress_bar.setVisible(False)
 
+    def start_waiting_progress_bar(self) -> None:
```
xjules, line 92: Is it reasonable to write a simple test for it? Ie. `self._wait_progress_bar active` or not?

jonathan-eq: I am not sure; it hanging or not will most likely go unnoticed by the users anyways. I can add a small unit test, but I do not think we should put too much effort into it.

xjules: Hm, I see. Currently it is not tested at all. Then I guess I'll leave it up to you :)

xjules (approved):
Nice job @jonathan-eq ! :guitar:

PR #8937, opened 2024-10-10
Initialize ert storage outside Simulator.start

yngve-sk:
Experiments need all response configs / parameter configs in place before being initialized, and if we are to get the experiment in place before running the simulator, the logic needs to be reordered for this

* Makes it possible to get experiment before invoking the optimizer, avoids having to use callbacks, useful for setting up storage of optimization results
* Makes ERT storage / experiment initialization more explicit, less callbacks

### DIFF
--- a/src/everest/suite.py
+++ b/src/everest/suite.py
@@ -393,18 +393,10 @@ class _EverestWorkflow(object):
         ert_config = everest_to_ert_config(self._config)
 
         with open_storage(ert_config.ens_path, mode="w") as storage:
-            # Initialize the Everest simulator:
-            experiment = storage.create_experiment(
-                name=f"EnOpt@{datetime.datetime.now().strftime('%Y-%m-%d@%H:%M:%S')}",
-                parameters=ert_config.ensemble_config.parameter_configuration,
-                responses=ert_config.ensemble_config.response_configuration,
-            )
-
             simulator = Simulator(
                 self.config,
                 ert_config,
                 storage,
-                experiment,
                 callback=self._simulation_callback,
             )
 
--- a/src/everest/suite.py
+++ b/src/everest/suite.py
@@ -19,10 +19,12 @@ from ropt.plan import OptimizationPlanRunner
 from seba_sqlite import SqliteStorage
 
 from ert.resources import all_shell_script_fm_steps
+from ert.storage import open_storage
 from everest.config import EverestConfig
 from everest.optimizer.everest2ropt import everest2ropt
 from everest.plugins.site_config_env import PluginSiteConfigEnv
 from everest.simulator import Simulator
+from everest.simulator.everest_to_ert import everest_to_ert_config
 from everest.strings import EVEREST, SIMULATOR_END, SIMULATOR_START, SIMULATOR_UPDATE
 from everest.util import makedirs_if_needed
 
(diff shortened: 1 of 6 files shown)


DanSava (dismissed):
LGTM

src/everest/suite.py
```diff
@@ -389,38 +392,68 @@ def start_optimization(self):
...
+                responses=ert_config.ensemble_config.response_configuration,
+            )
+
+            simulator = Simulator(
+                self.config,
+                ert_config,
+                storage,
+                experiment,
+                callback=self._simulation_callback,
+            )
```
verveerpj, line 426: I don't really like this. This has gotten a lot more complicated now. Before this was a single call to instantiate Simulator, now it exposes a lot of the machinery to run the ensemble. I understand that in the future we may want access to the storage (and experiment?) outside of the simulator object. Can we encapsulate this somehow?

verveerpj (changes requested):

src/everest/suite.py
yngve-sk: Hmmm..  Some encapsulation could be done, but I think we do need to be able to:

1. Run an Everest optimization within the context of one write-access storage
2. Initialize the storage outside of the async callbacks etc
3. For future (3) we will be able to link up ROPT callbacks with an existing experiment / ensemble which it can write optimization results into. Currently ROPT initialization is done in `suite.py`, if we were to move all of this into `Simulator`, we would also have to pack the ROPT stuff in there to achieve a "correct" order of initialization so we can link things up.

Currently, open_storage is called after some async call from ROPT to a callback in ERT, and the experiment was created then. If we are to have one experiment per optimization run, and one ensemble per batch (for which we store some optimization results), we need to access the experiment (which we will write to) when we set up the ROPT observers. These observers need to be added before calling `simulator.start()` currently, and we can not add listeners afterwards. So I would say it makes sense to have the entire optimization within the context of one open storage. 

There are also some internal things in storage that makes it a bit harder to keep it encapsulated (these may be revised away on the storage side):

1. We can only have one storage with write-access open at a time. If each batch is to write its own results, we would need multiple write-storages, or a method that writes optimization results without "write access" (easy to achieve but not that clean).
5. Experiments without any ensembles will not be loaded from the file system, this makes it impossible to initialize an empty experiment first, and then add ensembles to it by re-opening a write storage (the re-opened storage will not load the experiment that was created in the first place, we could create a dummy empty ensemble just to make it load, or add something to storage to make it load "empty" experiments.
6. The experiment needs a response / parameter config, which requires the `ExtParamConfig`s to be added to the ERT config prior to initializing the experiment.

yngve-sk: Hmmm.. We could hold the experiment in the simulator and have the same functionality, will set it up.

src/everest/suite.py
```diff
@@ -389,38 +392,68 @@ def start_optimization(self):
...
+        # configuration key, but only a placeholder for the control definitions.
+        ens_config = ert_config.ensemble_config
+        for control_name, variables in config_dict["EXT_PARAM"].items():
+            ens_config.addNode(
+                ExtParamConfig(
+                    name=control_name,
+                    input_keys=variables,
+                    output_file=control_name + ".json",
+                )
+            )
```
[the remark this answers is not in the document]
yngve-sk: Was this maybe resolved by https://github.com/equinor/ert/pull/8957 ?

verveerpj (approved):
Not sure if you are fully finished, but I am happy with it, so approved.

PR #8954, opened 2024-10-14
Remove redundant/superseded Everest tests

### LINKED ISSUE #8858: Remove everest redundant_tests
**Issue**

Make sure Everest tests marked with `redundant_test` are covered by other tests in `tests/everest/test_math_func.py` and if that is the case remove the tests marked with `redundant_test`.

Also, remove the marker `redundant_test` if no longer needed.

StephanDeHoop:
**Issue**
Resolves #8858 

**Approach**
It seems, after investigating, that both the **test_math_func_multiobj** and **test_math_func_advanced** cover all the functionality of the other tests. The only exception, perhaps, is in the part of the **config_minimal** test where we test if the following parameter is ignored when no constraints are provided in the config file: "constraint_tolerance: 0.1  # To test that it is ignored.". **We could keep the minimal test and remove only the other two for this reason?** But it might seem like a stretch. 

## When applicable


tests/everest/test_math_func.py
```diff
@@ -12,117 +12,12 @@
 from everest.suite import _EverestWorkflow
 from everest.util import makedirs_if_needed
 
-CONFIG_FILE_MINIMAL = "config_minimal.yml"
-CONFIG_FILE_IN_CONSTR = "config_in_constr.yml"
-CONFIG_FILE_OUT_CONSTR = "config_out_constr.yml"
```
verveerpj, line 17: Could you check if these files are used in some other test, and if not, remove them?

verveerpj (commented):
One question, otherwise good.

verveerpj (commented):
You will also need to squash the two commits into a single one.

tests/everest/test_math_func.py
StephanDeHoop: I think they are (only) used in the documentation of Everest as examples for slowly building up the complexity of using Everest (at least the minimal one is for sure, maybe the in_constr and out_constr aren't)

verveerpj: But in the documentation are they explicitly referring to the file location? Because they used to be in an `examples `folder, but that has now become `test-data`, and we should not have anything there that is not used. The documentation should not refer to `test-data`.

StephanDeHoop: Sorry you are right, they are making their own ('everest_config.yml') which happens to look exactly the same as the 'config_minimal.yml'. We can delete them I guess, and also clean up the associated 'jobs' I suppose?

verveerpj: yes, please do.

StephanDeHoop: The job (DISTANCE3) for in- and output constraint was used in other tests. Also, config_minimal was used in quite a lot of other tests too, so we need to keep the config file.

verveerpj (changes requested):
Looks good. Can you squash into a single commit?

verveerpj (approved):

PR #8957, opened 2024-10-14
Initialize ErtConfig in everest_to_res

yngve-sk:
Initialize `ErtConfig` with ExtParams included in `everest_to_res`


verveerpj (approved):
👍

PR #8963, opened 2024-10-14
Move everest workflow logic over to a baserunmodel

### LINKED ISSUE #8983: Move _EverestWorkflow logic to a BaseRunModel
Move the logic without altering behavior

yngve-sk:
Resolves #8983 

**Issue**
Does not alter behavior or use ensemble evaluator, only restructures the code that executes the simulation+optimization to live within a baserunmodel. Subsequent refactorings should/could be done in smaller PRs/increments

### DIFF
--- a/src/ert/run_models/batch_simulator_run_model.py
+++ b/src/ert/run_models/batch_simulator_run_model.py
@@ -1,0 +1,441 @@
+from __future__ import annotations
+
+import copy
+import datetime
+import logging
+import os
+import re
+import shutil
+import threading
+import time
+from functools import partial
+from pathlib import Path
+from queue import SimpleQueue
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    TypedDict,
+)
+
+from ropt.enums import EventType, OptimizerExitCode
+from ropt.plan import OptimizationPlanRunner
+from seba_sqlite import SqliteStorage
+
+from ert.config import ErtConfig
+from ert.storage import Storage
+from everest.config import EverestConfig
+from everest.optimizer.everest2ropt import everest2ropt
+from everest.simulator import Simulator
+from everest.strings import EVEREST, SIMULATOR_END, SIMULATOR_START, SIMULATOR_UPDATE
+
+from ..ensemble_evaluator import EvaluatorServerConfig
+from ..resources import all_shell_script_fm_steps
+from .base_run_model import BaseRunModel, StatusEvents
+
+if TYPE_CHECKING:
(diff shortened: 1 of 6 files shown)


verveerpj (commented):
There are some attributes hanging around in `_EverestWorkflow._init__` that maybe can go:
- `self._monitor_thread `
- `self._simulation_delete_run_path`
- `self._display_all_jobs `
- `self._fm_errors`
- `self._max_batch_num_reached`

src/ert/run_models/batch_simulator_run_model.py
```diff
@@ -0,0 +1,502 @@
...
+            context=ctx,
+            error_callback=self._handle_errors,
+            callback=self._sim_callback,
+            delete_run_path=self._simulation_delete_run_path,
+            display_all_jobs=self._display_all_jobs,
+        )
+        self._monitor_thread.start()
+
+    def _ropt_callback(
+        self, _: Event, optimizer: OptimizationPlanRunner, simulator: Simulator
```
verveerpj, line 429: This break in a next release, since that definition moved in `ropt`, so then we have to change the import. We can make our life a bit easier by not typing that here and removing the import. Maybe leave a TODO to type it again later. Or leave it and will fix it when it breaks...

verveerpj (approved):
This does look good to me, some comments:

- There are some attributes hanging around in `_EverestWorkflow._init__` that maybe can go:
  - `self._monitor_thread `
  - `self._simulation_delete_run_path`
  - `self._display_all_jobs `
  - `self._fm_errors`
  - `self._max_batch_num_reached`
- I wonder if the name `BatchSimulatorRunModel `is correct. It basically runs the optimization, so maybe `OptimizationRunModel` or `EverestRunModel`?

yngve-sk:
> This does look good to me, some comments:
> 
> * There are some attributes hanging around in `_EverestWorkflow._init__` that maybe can go:
>   
>   * `self._monitor_thread `
>   * `self._simulation_delete_run_path`
>   * `self._display_all_jobs `
>   * `self._fm_errors`
>   * `self._max_batch_num_reached`
> * I wonder if the name `BatchSimulatorRunModel `is correct. It basically runs the optimization, so maybe `OptimizationRunModel` or `EverestRunModel`?

I was hesitant about the naming, but I think something like `EverestRunModel` makes more sense so I will call it that

src/ert/run_models/batch_simulator_run_model.py
yngve-sk: I think it should be fine to leave it typed now and just update when it changes, we expect mypy to be happy with this, and updating the import should be OK. We will get mypy failures on PRs / main as soon as ROPT updates

PR #8968, opened 2024-10-15
Use only one fm shell script list

yngve-sk:
Stops imports from `everes.jobs.__init__` which in turn imports entire everest, very prone to circular imports. It is all to import this list of shell script forward model steps, which is already in `ert.resources`.


verveerpj (approved):
LGTM!

PR #8971, opened 2024-10-15
Add runpath validation in gui load_results_manually

### LINKED ISSUE #8691: No validation for uninitialized ensembles in manual update/evaluate experiment

Open ert gui and create a new experiment / ensemble in the manage experiments window, then attempt  to run either manual update or evaluate ensemble. 

image

Should validate and disable run experiment button.

### steps to reproduce

```Python
ert gui poly.ert
Create new experiment / ensemble
Run either manual update or evaluate ensemble
```

- [ ] python 3.8
- [ ] python 3.9
- [ ] python 3.10
- [X] python 3.11
- [ ] python 3.12
- [X] macosx
- [ ] rhel7
- [ ] rhel8
- [X] local queue
- [ ] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

jonathan-eq:
The commit adds runpath validation so the load button in `load_results_manually` will be disabled until the runpath exists, and the user has the read/write/execute permissions.
The runpath text box becomes red, and the user can hover it to see the warning message. This is the same way it is done for the other fields. 

**Approach**
😎 

image
image
image

## When applicable

### DIFF
--- a/src/ert/gui/ertwidgets/multilinestringbox.py
+++ b/src/ert/gui/ertwidgets/multilinestringbox.py
@@ -1,0 +1,112 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, Optional
+
+from qtpy.QtGui import QPalette
+from qtpy.QtWidgets import QTextEdit
+
+from .validationsupport import ValidationSupport
+
+if TYPE_CHECKING:
+    from ert.validation import ArgumentDefinition
+
+    from .models import TextModel
+
+
+class MultiLineStringBox(QTextEdit):
+    """MultiLineStringBox shows a multiline string. The data structure expected and sent to the
+    getter and setter is a multiline string."""
+
+    def __init__(
+        self,
+        model: TextModel,
+        default_string: str = "",
+        placeholder_text: str = "",
+        minimum_width: int = 250,
+        readonly: bool = False,
+    ):
+        QTextEdit.__init__(self)
+        self.setMinimumWidth(minimum_width)
+        self._validation = ValidationSupport(self)
+        self._validator: Optional[ArgumentDefinition] = None
+        self._model = model
+        self._enable_validation = True
+
+        if placeholder_text:
+            self.setPlaceholderText(placeholder_text)
+        self.textChanged.connect(self.stringBoxChanged)
+
+        self.textChanged.connect(self.validateString)
+
+        self._valid_color = self.palette().color(self.backgroundRole())
+        self.setText(default_string)
+
+        self._model.valueChanged.connect(self.modelChanged)
+        self.modelChanged()
+        self.setReadOnly(readonly)
+
+    def validateString(self) -> None:
+        if not self._enable_validation or self._validator is None:
+            return
+
+        string_to_validate = self.toPlainText()
+        if not string_to_validate and self.placeholderText():
+            string_to_validate = self.placeholderText()
+
+        validation_success = self._validator.validate(string_to_validate)
+
+        palette = self.palette()
+        if not validation_success:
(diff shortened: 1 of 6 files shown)


src/ert/validation/runpath_argument.py
```diff
@@ -0,0 +1,28 @@
...
+    MISSING_PERMISSION = "You are missing permissions for the specified runpath."
+
+    def __init__(self, **kwargs: bool) -> None:
+        super().__init__(**kwargs)
+
+    def validate(self, token: str) -> ValidationStatus:
+        parsed_runpath_without_suffix = "/".join(token.split("/")[:-2])
+        validation_status = super().validate(token)
+        try:
+            os.listdir(parsed_runpath_without_suffix)
```
dafeda, line 18: Nitpicking here, but is it necessary to list the dir or  could we use `os.path.isdir` and `os.access` instead?
Something like this:

```python
if not os.path.isdir(parsed_runpath_without_suffix):
    validation_status.setFailed()
    validation_status.addToMessage(RunPathArgument.INVALID_PATH)
elif not os.access(parsed_runpath_without_suffix, os.R_OK):
    validation_status.setFailed()
    validation_status.addToMessage(RunPathArgument.MISSING_PERMISSION)
else:
    validation_status.setValue(token)
```

Should be more efficient and perhaps easier to read.

src/ert/validation/runpath_argument.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class RunPathArgument(ArgumentDefinition):
+    INVALID_PATH = "The specified runpath does not exist."
+    MISSING_PERMISSION = "You are missing permissions for the specified runpath."
+
+    def __init__(self, **kwargs: bool) -> None:
+        super().__init__(**kwargs)
+
+    def validate(self, token: str) -> ValidationStatus:
```
dafeda, line 14: Would it make sense to pass in token as a `Path` and not a string?

jonathan-eq: It is a method we override from the super class `ArgumentDefinition`, so I don't think we can change it without changing all argument validations using the same super class.

dafeda: Ah I see, thanks.

src/ert/validation/runpath_argument.py
jonathan-eq: listdir both checks if the dir exists and if the user has access. I can change it to be more concrete 👌.
Do you think the user only needs read_access, or should I check all rwx?

dafeda: I think users need only have read access to load results manually. Might be wrong though.

jonathan-eq: I tested it manually, we apparently need r-x permissions. Will update it right away

jonathan-eq: Should I test for read permission first, then execute permission? I could have different error messages for those two, or is that excessive?

dafeda: I think a single error message is enough.

src/ert/gui/ertwidgets/multilinestringbox.py
```diff
@@ -0,0 +1,112 @@
...
+
+from .validationsupport import ValidationSupport
+
+if TYPE_CHECKING:
+    from ert.validation import ArgumentDefinition
+
+    from .models import TextModel
+
+
+class MultiLineStringBox(QTextEdit):
```
jonathan-eq, line 16: This new class is a copy of StringBox, but it inherits from QTextEdit rather than QLineEdit.

dafeda (approved):
Looks good 👍

PR #9003, opened 2024-10-21
Use a concat instead of vstack

yngve-sk:
Resolves https://github.com/equinor/ert/issues/9004
Resolves failing PRs etc. Stops vconcat'ing non-empty dataframes onto an initially empty df


xjules (approved):
Looks like it! :rocket:

PR #9022, opened 2024-10-23
Rename job.py to fm_step.py and Job to ForwardModelStep

berland:
**Issue**
Resolves eye itch

**Approach**
`nvim`

## When applicable

### DIFF
--- a/src/_ert/forward_model_runner/__init__.py
+++ b/src/_ert/forward_model_runner/__init__.py
@@ -1,4 +1,4 @@
-"""_ert.forward_model_runner is called by ert to run jobs in the runpath.
+"""_ert.forward_model_runner is called by ert to run forward model steps in the runpath.
 
 It is split into its own package for performance reasons,
 simply importing ert can take several seconds, which is not ideal when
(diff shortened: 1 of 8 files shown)


src/_ert/forward_model_runner/fm_step.py
```diff
@@ -76,7 +76,7 @@ def killed_by_oom(pids: Sequence[int]) -> bool:
     return False
 
 
-class Job:
+class FMstep:
```
yngve-sk, line 79: Maybe a nitpick, but I think the S should be big too?

berland: It probably should. Potentially go all the way and write it as ForwardModelStep? That object name is actually not in use anywhere else in the codebase (although we do use it as a prefix for some other object names)

yngve-sk: ForwardModelStep sounds better to me

eivindjahren (commented):
:+1:

yngve-sk (approved):
LGTM. Final nitpick; would consider renaming the file fm_step.py to forward_model_step.py for consistency, but this is a good change regardless

PR #9033, opened 2024-10-24
Remove src/everest/suite.py

yngve-sk:
Resolved broader issue of main everest control flow logic in `everest.suite` still being there, and not all in the baserunmodel. The current everest.suite setup is very prone to circular imports, as it imports a lot of big parts from `everest`.

Resolves https://github.com/equinor/komodo-releases/issues/6438


eivindjahren (approved):
LGTM! (Remember to squash!)

PR #9043, opened 2024-10-25
Let flaky adaptive tests be rerun

berland:
These adaptive localization tests fail very often on weak hardware, and is hampering the deploy process.

Adding reruns on failure will lead to similar hampering when the test actually starts to fail for the correct reasons, but this we have to live with.

**Issue**
Resolves pain by  🩹  (#8787)

**Approach**
♻️ 

## When applicable

### DIFF
--- a/tests/ert/ui_tests/cli/analysis/test_adaptive_localization.py
+++ b/tests/ert/ui_tests/cli/analysis/test_adaptive_localization.py
@@ -30,6 +30,7 @@ def run_cli_ES_with_case(poly_config):
     return prior_ensemble, posterior_ensemble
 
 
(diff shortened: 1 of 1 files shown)


eivindjahren:
I think we would have to check whether this works, by rerunning the actions until we observe a retry. Currently the tests timeout at 600s. If the failure mechanism (seems to be some sort of connection problem) would propagate to a failure immediately rather than a deadlock, then that would also help.

eivindjahren (commented):
I think if we can show that the test is retried or that this makes the workflow more stable then we can merge the PR. Mind, #8787 should be high priority as clearly something is really wrong on mac.

PR #9045, opened 2024-10-25
Handle empty responses in MeasuredData

yngve-sk:
Issue
Resolves failing semeio tests

Approach
Handle case of missing responses


eivindjahren (changes requested):
We should have regression test for this.

eivindjahren (approved):
:+1:

PR #9050, opened 2024-10-25
Add new hook for specifying config per forward model step

### LINKED ISSUE #9036: Allow pluginsystem to configure general key-value pairs pr forward model step
Some forward model steps require certain configuration parameters to be set. This could be environment variables or PATH prefixes, or any other.

Today, such options can be set through the site-configuration which can add `SETENV` lines to every Ert run, yielding environment variables available when running the forward model.

To avoid leaking configuration between unrelated forward models, it should be possible to specify environment variables to specific forward model steps, distinguished by the forward model step name (which might be run repeatedly).

Having this as a hook also enables other parts of Ert to access this information. This is relevant for e.g. `eclrun` validation which queries itself for available versions.

Open questions:
* Should we only allow upper-case keys, since we want to expose this to each step via environment variables?
* Should we expose only upper-case keys through environment variables? 
* Should we allow values to be any Python type (but `str(value)` when exposed through env)

berland:
**Issue**
Resolves #9036 

**Approach**
➕ 

## When applicable


src/ert/plugins/__init__.py
```diff
@@ -27,6 +27,7 @@ def inner(*args: P.args, **kwargs: P.kwargs) -> Any:
                     "installable_workflow_jobs",
                     "help_links",
                     "installable_forward_model_steps",
+                    "forward_model_configuration",
                     "ecl100_config_path",
                     "ecl300_config_path",
                     "flow_config_path",
```
sondreso, line 33: I guess the goal should be to get rid of these hooks as well? :thinking:

berland: Yes, but that will be in a separate PR.

src/ert/plugins/hook_specifications/forward_model_steps.py
```diff
@@ -18,3 +18,11 @@ def installable_forward_model_steps() -> (
     :return: List of forward model step plugins in the form of subclasses of the
         ForwardModelStepPlugin class
     """
+
+
+@no_type_check
+@hook_specification
+def forward_model_configuration() -> PluginResponse[List[Type[ForwardModelStepPlugin]]]:
```
berland, line 25: is this the correct return type??

src/ert/config/ert_config.py
```diff
@@ -748,6 +759,9 @@ def handle_default(fm_step: ForwardModelStep, arg: str) -> str:
 
         job_list_errors = []
         job_list: List[ForwardModelStepJSON] = []
+        env_pr_fm_step = cls._uppercase_subkeys_and_stringify_subvalues(
+            ErtPluginManager().get_forward_model_configuration()
+        )
```
sondreso, line 764: Could we get this info in the `with_plugins` function instead of instantiating and fetching it directly here? Then we would get consistency where if you use the `with_plugins` version then plugins are loaded, while otherwise you just get a plain ert config.

src/ert/plugins/plugin_manager.py
```diff
@@ -139,6 +140,50 @@ def get_flow_config_path(self) -> Optional[str]:
...
+                if not isinstance(fmstep_name, str):
+                    raise TypeError(
+                        f"{res.plugin_metadata.plugin_name} did not "
+                        "provide dict[str, dict[str, Any]]"
+                    )
+                if not isinstance(fmstep_config, dict):
+                    raise TypeError(
+                        f"{res.plugin_metadata.plugin_name} did not "
+                        "provide dict[str, dict[str, Any]]"
+                    )
```
sondreso, line 168: I think this should either be joined in one if statement or provide a more accurate error message for each case :thinking:

berland: fixed

src/ert/config/ert_config.py
berland: fixed.

sondreso (approved):
Nicely done! :rocket:

PR #9081, opened 2024-10-29
Increment storage version

yngve-sk:
Put polars migration into separate storage version due to 2024.10 ERT being still xarray


andreas-el (approved):
🏖️

Issue #9077, opened 2024-10-29
Only one observation rendered on migrated storage


yngve-sk:
### What happened? (You can include a screenshot if it helps explain)
Only for summary
![Image](https://github.com/user-attachments/assets/b6e5ec40-4556-45c7-9f95-4eadae71a923)
![Image](https://github.com/user-attachments/assets/e61dde1e-d277-41f5-ad12-4d0425f8bd16)

Not gen data
![Image](https://github.com/user-attachments/assets/cdd7b159-3958-4508-a33d-24ab88adbd3c)

### steps to reproduce

1. Run Drogon AHM with 2024.06 release
2. Open with 2024.10 release which will migrate the storage
3. Open plotter, select response with observations, expect many observations but see only one

### Environment where bug has been observed

- [x] python 3.8
- [ ] python 3.9
- [ ] python 3.10
- [ ] python 3.11
- [ ] python 3.12
- [ ] macosx
- [x] rhel7
- [ ] rhel8
- [ ] local queue
- [ ] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

PR #9087, opened 2024-10-29
Make plot_api return all observations for response

yngve-sk:
**Issue**
Resolves [#9077](https://github.com/equinor/ert/issues/9077)

Tested on RGS for stable and bleeding:
Before (left is rendering only one instead of 3 observations):

After:
Screenshot 2024-10-30 at 11 56 24

### DIFF
--- a/src/ert/gui/tools/plot/plot_api.py
+++ b/src/ert/gui/tools/plot/plot_api.py
@@ -205,29 +205,37 @@ class PlotApi:
                     timeout=self._timeout,
                 )
                 self._check_response(response)
+
+                observations = response.json()
+                observations_dfs = []
+
                 if not response.json():
                     continue
                 try:
-                    obs = response.json()[0]
+                    response.json()[0]
                 except (KeyError, IndexError, JSONDecodeError) as e:
                     raise httpx.RequestError(
                         f"Observation schema might have changed key={key},  ensemble_name={ensemble.name}, e={e}"
                     ) from e
-                try:
-                    int(obs["x_axis"][0])
-                    key_index = [int(v) for v in obs["x_axis"]]
-                except ValueError:
-                    key_index = [pd.Timestamp(v) for v in obs["x_axis"]]
 
-                data_struct = {
-                    "STD": obs["errors"],
-                    "OBS": obs["values"],
-                    "key_index": key_index,
-                }
+                for obs in observations:
+                    try:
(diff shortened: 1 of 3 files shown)


src/ert/gui/tools/plot/plot_api.py
```diff
@@ -205,29 +205,37 @@ def observations_for_key(self, ensemble_ids: List[str], key: str) -> pd.DataFram
                     timeout=self._timeout,
                 )
                 self._check_response(response)
+
+                observations = response.json()
+                observations_dfs = []
+
                 if not response.json():
```
DanSava, line 212: Why not replace `if not response.json():` with `if not observations:` if you already declared `observations = response.json()` on ln 209

src/ert/gui/tools/plot/plot_api.py
```diff
@@ -205,29 +205,37 @@ def observations_for_key(self, ensemble_ids: List[str], key: str) -> pd.DataFram
...
                     continue
                 try:
-                    obs = response.json()[0]
+                    response.json()[0]
```
DanSava, line 215: Similar why not replace `response.json()[0]` with `observations[0]`

frode-aarstad (approved):
LGTM

src/ert/gui/tools/plot/plot_api.py
```diff
@@ -205,29 +205,37 @@ def observations_for_key(self, ensemble_ids: List[str], key: str) -> pd.DataFram
...
                 except (KeyError, IndexError, JSONDecodeError) as e:
                     raise httpx.RequestError(
                         f"Observation schema might have changed key={key},  ensemble_name={ensemble.name}, e={e}"
                     ) from e
```
DanSava, line 219: It is unclear to me what the purpose of this section is now.

DanSava (approved):
Looks ok as is but with some structure/style comments :racehorse:

src/ert/gui/tools/plot/plot_api.py
yngve-sk: I'm not sure either, but am also not confident to just remove it here as part of this bugfix, although I think this entire module, PlotAPI/Dark storage could be simplified alot

DanSava: I agree with keeping it for now

PR #9137, opened 2024-11-02
Prevent copying mount point data when setting install_data to a non existing directory

### LINKED ISSUE #9071: When setting install_data to a non existing directory, root partition is being copied instead

Related to Everest and ERT, when no variable is defined in Everest, the whole root directory (`/`) is being copied to each simulation directory.

```yaml
install_data:
  - link: False
    source: r{{ does_not_exist }}/
    target: tmp
```

An error e.g.:  "does_not_exist variable does not exist." 

### steps to reproduce

Use in Everest config:

```yaml
install_data:
  - link: False
    source: r{{ does_not_exist }}/
    target: tmp
```

- [ ] python 3.8
- [ ] python 3.9
- [ ] python 3.10
- [x] python 3.11
- [ ] python 3.12
- [ ] macosx
- [ ] rhel7
- [ ] rhel8
- [x] local queue
- [ ] lsf queue
- [x] slurm queue
- [ ] openPBS queue

AugustoMagalhaes:
**Issue**
Resolves #9071

**Approach**
Includes additional validation for source path in EverestConfig to avoid copying data from root folder ("/") when install_data template render an invalid string. I'm not sure wether to add this validation actually, I've included in *check_path_exists* function in validation_utils.py but in the first commit there's a glimpse of how it could be if the validation was in *InstallDataConfig* as a pydantic validation of the *source* property itself. Just let me know which one you guys think fits better (or a new one maybe) and I modify it later.

## When applicable

### DIFF
--- a/src/everest/config/install_data_config.py
+++ b/src/everest/config/install_data_config.py
@@ -1,3 +1,4 @@
+import os
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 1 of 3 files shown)


DanSava (commented):
Requires a rebase on `main` but otherwise looks good

DanSava (approved):
Thanks for the contribution :1st_place_medal:

PR #9158, opened 2024-11-05
Smaller simplifications of everest baserunmodel

yngve-sk:


src/ert/run_models/everest_run_model.py
```diff
@@ -399,15 +399,7 @@ def run_experiment(
         )
 
         # Initialize the ropt optimizer:
-        optimizer = self._configure_optimizer(simulator)
-
-        # Before each batch evaluation we check if we should abort:
-        optimizer.add_observer(
```
StephanDeHoop, line 405: So functionally the only difference is that now we add an observer when we create the optimizer (at the end) instead of outside that method?

StephanDeHoop (approved):
Looks good to me, just have one comment/question

src/ert/run_models/everest_run_model.py
yngve-sk: Yep. It was always created after initializing, just not within that function.

PR #9163, opened 2024-11-06
Limit the use of EverestConfig in function signatures in everest csv exporting

### LINKED ISSUE #9162: Limit the use of EverestConfig everest export

DanSava:
**Issue**
Resolves #9162

## When applicable

### DIFF
--- a/src/everest/__init__.py
+++ b/src/everest/__init__.py
@@ -15,7 +15,7 @@ except ImportError:
 from everest import detached, docs, jobs, templates, util
 from everest.bin.utils import export_to_csv, export_with_progress
 from everest.config_keys import ConfigKeys
-from everest.export import MetaDataColumnNames, export, filter_data, validate_export
+from everest.export import MetaDataColumnNames, filter_data
 
 __author__ = "Equinor ASA and TNO"
 __all__ = [
@@ -23,7 +23,6 @@ __all__ = [
     "MetaDataColumnNames",
     "detached",
     "docs",
-    "export",
     "export_to_csv",
     "export_with_progress",
     "filter_data",
@@ -31,5 +30,4 @@ __all__ = [
     "load",
     "templates",
     "util",
-    "validate_export",
 ]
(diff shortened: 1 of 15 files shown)


yngve-sk (commented):
The refactor in itself looks good but I have some things I'm not totally sure about. For the validation you got rid of the direct everest config and this is nice, and using only `batches` and `keywords` from `self`. 

 For the `export_data`, if I understood correctly, this cannot be put within the export config, because the export config may be null even if you want to export, but if you make this a staticmethod of ExportConfig (or keep it in its original place). I also think if you make `model.data_file` and `.output_dir` explicit arguments you can drop having to reference an EverestConfig as is done through the `self` reference there.

I'm not sure if I see why the export logic should be _partially_ moved to the config? They still refer to some stuff under `export.py`, and the pattern leans towards making it easy to suddenly get circular imports, so if they are to be moved to the configs I'd at least say do it fully and get rid the entire `export.py` file, or keep it there. In ERT we do have some logic in the configs (maybe too much) but it sort of makes sense to keep the exporting logic in the export config. Would try to get it all into the ExportConfig and make the things that should run without an instance a `@staticmethod`

DanSava:
> For the `export_data`, if I understood correctly, this cannot be put within the export config, because the export config may be null even if you want to export, but if you make this a staticmethod of ExportConfig (or keep it in its original place). I also think if you make `model.data_file` and `.output_dir` explicit arguments you can drop having to reference an EverestConfig as is done through the `self` reference there.
> 
> I'm not sure if I see why the export logic should be _partially_ moved to the config? They still refer to some stuff under `export.py`, and the pattern leans towards making it easy to suddenly get circular imports, so if they are to be moved to the configs I'd at least say do it fully and get rid the entire `export.py` file, or keep it there. In ERT we do have some logic in the configs (maybe too much) but it sort of makes sense to keep the exporting logic in the export config. Would try to get it all into the ExportConfig and make the things that should run without an instance a `@staticmethod`

I see your point I will just move it back to `export.py` and make the args explicit.

yngve-sk (approved):
Nice! Behavior is the same and the logic is more clear 🚀

PR #9173, opened 2024-11-11
Replace websockets with zmq

xjules:
**Issue**

POC: testing feasibility of zmq.

Implementing push-pull pattern.
Server (ensemble evaluator) is doing the pull while each client is doing the push.

**Approach**

## When applicable


src/_ert/forward_model_runner/client.py
```diff
@@ -55,84 +51,83 @@ def __init__(
         cert: Optional[Union[str, bytes]] = None,
         max_retries: Optional[int] = None,
         timeout_multiplier: Optional[int] = None,
+        dealer_name: str | None = None,
```
jonathan-eq, line 54: Too keep it the same style as the other parameters
```suggestion
        dealer_name: Optional[str] = None,
```

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -205,43 +223,41 @@ async def _failed_handler(self, events: Sequence[EnsembleFailed]) -> None:
...
+                            if not self._clients_connected:
+                                self._clients_empty.set()
+                        else:
+                            event = event_from_json(raw_msg)
+                            if type(event) is EEUserCancel:
+                                logger.debug("Client asked to cancel.")
+                                self._signal_cancel()
+                            elif type(event) is EEUserDone:
+                                logger.debug("Client signalled done.")
+                                self.stop()
```
jonathan-eq, line 256: Do we have to ACK messages from the monitor/client, or is it self acknowledging when it sees the EE stopping?

src/_ert/forward_model_runner/client.py
xjules: I wanted to be "modern", but will make it the same then :)

src/ert/ensemble_evaluator/evaluator.py
xjules: I've decided not to do acknowledgements between the client and evaluator, since it is dispatchers that are problematic. But time will tell...

jonathan-eq:
All tests passing 😻

src/_ert/forward_model_runner/cli.py
```diff
@@ -22,19 +22,14 @@ def _setup_reporters(
     ens_id,
     dispatch_url,
     ee_token=None,
-    ee_cert_path=None,
```
jonathan-eq, line 25: Do we not need the certificate for TLS/SSL anymore?

src/_ert/forward_model_runner/client.py
```diff
@@ -26,107 +22,139 @@ class ClientConnectionClosedOK(Exception):
...
+    DEFAULT_ACK_TIMEOUT = 5
+    _receiver_task: asyncio.Task[None] | None
 
     def __enter__(self) -> Self:
+        self.loop.run_until_complete(self.__aenter__())
         return self
 
+    def term(self) -> None:
+        self.socket.close()
+        self.context.term()
```
jonathan-eq, line 35: Do we need all the contexts?

src/_ert/forward_model_runner/client.py
```diff
@@ -26,107 +22,139 @@ class ClientConnectionClosedOK(Exception):
...
 class Client:
-    DEFAULT_MAX_RETRIES = 10
-    DEFAULT_TIMEOUT_MULTIPLIER = 5
-    CONNECTION_TIMEOUT = 60
+    DEFAULT_MAX_RETRIES = 5
+    DEFAULT_ACK_TIMEOUT = 5
+    _receiver_task: asyncio.Task[None] | None
 
     def __enter__(self) -> Self:
+        self.loop.run_until_complete(self.__aenter__())
```
jonathan-eq, line 30: Maybe we can remove all of the synced wrappers of the async methods?

src/_ert/forward_model_runner/client.py
```diff
@@ -26,107 +22,139 @@ class ClientConnectionClosedOK(Exception):
...
         self, exc_type: Any, exc_value: Any, exc_traceback: Any
     ) -> None:
-        if self.websocket is not None:
-            await self.websocket.close()
+        try:
+            await self._send("DISCONNECT")
+        except ClientConnectionError:
+            logger.error("No ack for dealer disconnection. Connection is down!")
+        finally:
+            self.socket.disconnect(self.url)
```
jonathan-eq, line 53: Why do we have to specify the url here?

src/_ert/forward_model_runner/client.py
```diff
@@ -26,107 +22,139 @@ class ClientConnectionClosedOK(Exception):
...
+            await self._term_receiver_task()
+            self.term()
+
+    async def _term_receiver_task(self) -> None:
+        if self._receiver_task and not self._receiver_task.done():
+            self._receiver_task.cancel()
+            await asyncio.gather(self._receiver_task, return_exceptions=True)
```
jonathan-eq, line 60: Do we have to gather this cancelled task, or can we await it (and suppress asyncio.cancellationerror if it is raised)?

src/_ert/forward_model_runner/client.py
```diff
@@ -26,107 +22,139 @@ class ClientConnectionClosedOK(Exception):
...
-    CONNECTION_TIMEOUT = 60
+    DEFAULT_MAX_RETRIES = 5
+    DEFAULT_ACK_TIMEOUT = 5
+    _receiver_task: asyncio.Task[None] | None
 
     def __enter__(self) -> Self:
+        self.loop.run_until_complete(self.__aenter__())
         return self
 
+    def term(self) -> None:
```
jonathan-eq, line 33: I think we should rearrange the order of the methods, it is a bit difficult to read

src/_ert/forward_model_runner/client.py
jonathan-eq: After Eivind's typing PR, maybe we should have your original version after all. That way it wont be a rebase conflict right before you try merging it 😆

src/_ert/forward_model_runner/client.py
```diff
@@ -26,107 +22,139 @@ class ClientConnectionClosedOK(Exception):
...
-        if timeout_multiplier is None:
-            timeout_multiplier = self.DEFAULT_TIMEOUT_MULTIPLIER
-        if url is None:
-            raise ValueError("url was None")
+        self._ack_timeout = ack_timeout or self.DEFAULT_ACK_TIMEOUT
         self.url = url
         self.token = token
-        self._additional_headers = Headers()
+
+        # Set up ZeroMQ context and socke
```
jonathan-eq, line 74: I think this comment can be removed

src/_ert/forward_model_runner/client.py
```diff
@@ -26,107 +22,139 @@ class ClientConnectionClosedOK(Exception):
...
+        self._ack_event: asyncio.Event = asyncio.Event()
+        self.context = zmq.asyncio.Context()
+        self.socket = self.context.socket(zmq.DEALER)
+        self.socket.setsockopt(zmq.LINGER, 0)
```
jonathan-eq, line 78: What does this option do?

src/_ert/forward_model_runner/client.py
```diff
@@ -26,107 +22,139 @@ class ClientConnectionClosedOK(Exception):
...
+        if dealer_name is None:
+            self.dealer_id = f"dispatch-{uuid.uuid4().hex[:8]}"
+        else:
+            self.dealer_id = dealer_name
```
jonathan-eq, line 82:
```suggestion
self._dealer_id = dealer_name or f"dispatch-{uuid.uuid4().hex[:8]}"
```

src/_ert/forward_model_runner/client.py
```diff
@@ -26,107 +22,139 @@ class ClientConnectionClosedOK(Exception):
...
+            raise
+
+    def send(self, message: str, retries: int | None = None) -> None:
+        self.loop.run_until_complete(self._send(message, retries))
+
+    async def process_message(self, msg: str) -> None:
+        pass
+
+    async def _receiver(self) -> None:
+        while True:
```
jonathan-eq, line 112: Should this support creating new sockets if it fails, or can we change this to
```suggestion
        while not self.socket.closed():
```
?

src/_ert/forward_model_runner/client.py
```diff
@@ -26,107 +22,139 @@ class ClientConnectionClosedOK(Exception):
...
+                self.socket.connect(self.url)
+            except asyncio.CancelledError:
+                self.term()
+                raise
+
+            retries -= 1
+            if retries > 0:
+                logger.info(f"Retrying... ({retries} attempts left)")
+                await asyncio.sleep(backoff)
+                backoff = min(backoff * 2, 10)  # Exponential backoff
```
jonathan-eq, line 157: Should we add some randomness to the backoff period?

src/_ert/forward_model_runner/client.py
```diff
@@ -26,107 +22,139 @@ class ClientConnectionClosedOK(Exception):
...
+
+    async def process_message(self, msg: str) -> None:
+        pass
+
+    async def _receiver(self) -> None:
+        while True:
+            try:
+                _, raw_msg = await self.socket.recv_multipart()
+                if raw_msg == b"ACK":
+                    self._ack_event.set()
```
jonathan-eq, line 116: We discussed whether this event and the receiver task was even neccessary. Since the socket handles both incoming events and "ACK" on outbound messages, it seems we need both. Maybe this can be refactored in the future. Very good work!

src/_ert/forward_model_runner/client.py
```diff
@@ -26,107 +22,139 @@ class ClientConnectionClosedOK(Exception):
...
-            self.url,
-            ssl=self._ssl_context,
-            additional_headers=self._additional_headers,
-            open_timeout=self.CONNECTION_TIMEOUT,
-            ping_timeout=self.CONNECTION_TIMEOUT,
-            ping_interval=self.CONNECTION_TIMEOUT,
-            close_timeout=self.CONNECTION_TIMEOUT,
-        )
+    async def _send(self, message: str, retries: int | None = None) -> None:
+        self._ack_event.clear()
```
jonathan-eq, line 127: Can we end in a corner case where a forward model step goes from running to finished very quickly, so we try sending a Running and Finished at almost the same time? In this case, will both `send()`-coroutines return when the first "ack" is received?

jonathan-eq: In this case, should every message have its own instance of an ack event (that includes an id) that we store in a dict in the client?

src/_ert/forward_model_runner/client.py
```diff
@@ -26,107 +22,139 @@ class ClientConnectionClosedOK(Exception):
...
-            except ConnectionClosedError as exception:
-                if retry == self._max_retries:
-                    error_msg = (
-                        f"Not been able to send the event"
-                        f" to {self.url}! Max retries reached!"
-                        f" Exception from {type(exception)}: {exception!s}"
+                    return
+                except TimeoutError:
+                    logger.warning(
+                        f"{self.dealer_id} failed to get acknowledgment on the {message}. Resending."
```
jonathan-eq, line 141: What will happen if the event shows up after it times out? Will this cause it to hang, until maxretries is reached?
msg0 goes out, no one replies
msg1 goes out, ack0 returns
we return.             ack1 returns

jonathan-eq: No wait. There is nothing to be left hanging. Getting an extra ACK after everything finished is handled. Very nice!

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -78,53 +75,46 @@ def __init__(self, evaluator_url, token=None, cert_path=None):
...
+                                break
+                        if start_time and (time.time() - start_time) > 2:
+                            break
+                        await client._send(event_to_json(event), self._max_retries)
+                        event = None
+                    except asyncio.CancelledError:
+                        return
+                    except ClientConnectionError as exc:
+                        logger.error(f"Failed to send event: {exc}")
+
```
jonathan-eq, line 113: Can we move this nested async function up a level and call it `async def _publisher(self,...)` instead? Probably easier to test.

src/_ert/forward_model_runner/client.py
xjules: Maybe. I think the synced version if now only used in tests.

src/_ert/forward_model_runner/client.py
xjules: Can do :+1:

src/ert/ensemble_evaluator/_ensemble.py
```diff
@@ -198,11 +191,10 @@ async def send_event(
         url: str,
         event: Event,
         token: str | None = None,
-        cert: str | bytes | None = None,
         retries: int = 10,
     ) -> None:
-        async with Client(url, token, cert, max_retries=retries) as client:
-            await client._send(event_to_json(event))
+        async with Client(url, token) as client:
+            await client._send(event_to_json(event), retries)
```
jonathan-eq, line 197: Why do we want to move the `retries`-parameter from the client constructor to the send method?

src/ert/ensemble_evaluator/_ensemble.py
```diff
@@ -116,6 +108,7 @@ def __post_init__(self) -> None:
         self._config: EvaluatorServerConfig | None = None
         self.snapshot: EnsembleSnapshot = self._create_snapshot()
         self.status = self.snapshot.status
+        self._client: Client | None = None
```
jonathan-eq, line 111: Is this ever set?

src/_ert/forward_model_runner/client.py
xjules: a socket can be connected via different protocols; ie. can boost multiple connect(...)

src/_ert/forward_model_runner/client.py
xjules: Since we capture `asyncio.CancelledError` we can just do await. :+1

src/_ert/forward_model_runner/client.py
xjules: Yes, it already back as Optional got removed.

src/_ert/forward_model_runner/client.py
xjules: Hm, maybe I can put a comment there. It is to prevent queuing messages and instead they are sent right away.

src/_ert/forward_model_runner/client.py
```diff
@@ -26,107 +22,139 @@ class ClientConnectionClosedOK(Exception):
...
+        except ClientConnectionError:
+            await self._term_receiver_task()
+            self.term()
+            raise
+
+    def send(self, message: str, retries: int | None = None) -> None:
+        self.loop.run_until_complete(self._send(message, retries))
+
+    async def process_message(self, msg: str) -> None:
+        pass
```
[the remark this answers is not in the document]
xjules: hm, no really sure to be honest. Maybe I can raise exception if a message (other than `ACK`) is to arrive to the dispatcher.

src/_ert/forward_model_runner/client.py
xjules: The issue here is that it will most by blocked here:
```python
_, raw_msg = await self.socket.recv_multipart()
```
Or if being closed then this will be triggered:
```python
except zmq.ZMQError as exc:
```
which will then try to reconnect. I will think about that.

src/ert/ensemble_evaluator/_ensemble.py
xjules: Two reasons.
1) Making it more stateless
2) I'd like to different retries for connect (which should happen fast) and for a default message sending

src/_ert/forward_model_runner/reporting/event.py
xjules: hm, maybe. 
Additionally this the reason for having all the report tests synced - your fw async PR would help in that matter. Also this is what makes reporter tests difficult to with async zmq sockets.

src/ert/ensemble_evaluator/_ensemble.py
xjules: No, correcto! taken it away.

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
+import asyncio
 import logging
 import queue
-import threading
-from datetime import datetime, timedelta
+import time
```
jonathan-eq, line 6: Can we remove this import and use `asyncio.time()` instead?

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -32,7 +28,7 @@
     Start,
 )
 from _ert.forward_model_runner.reporting.statemachine import StateMachine
-from _ert.threading import ErtThread
+from _ert.threading import ErtThread, threading
```
jonathan-eq, line 31: Should we re-import threading from `_ert.threading` or import it directly?

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -78,53 +75,46 @@ def __init__(self, evaluator_url, token=None, cert_path=None):
...
+                url=self._evaluator_url,
+                token=self._token,
+                ack_timeout=self._ack_timeout,
+            ) as client:
+                event = None
+                start_time = None
+                while True:
+                    try:
+                        if self._done.is_set() and start_time is None:
+                            start_time = time.time()
```
jonathan-eq, line 100: What is start_time? Is that the start of closing the event_publisher?

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -78,53 +75,46 @@ def __init__(self, evaluator_url, token=None, cert_path=None):
...
+                        if event is None:
+                            event = self._event_queue.get()
```
jonathan-eq, line 102: What would happen if we changed this from queue.Queue to asyncio.Queue? Would it fail due to no running event loop outside of the ErtThread?

src/ert/ensemble_evaluator/config.py
```diff
@@ -95,80 +94,38 @@ def _generate_certificate(
...
-        self.dispatch_uri = f"{self.url}/dispatch"
-        if generate_cert:
-            cert, key, pw = _generate_certificate(host)
-        else:
-            cert, key, pw = None, None, None
-        self.cert = cert
-        self._key: bytes | None = key
-        self._key_pw = pw
-
-        self.token = _generate_authentication() if use_token else None
```
jonathan-eq, line 145: Maybe we can delete this function if it is not used elsewhere

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -82,14 +65,20 @@ def __init__(self, ensemble: Ensemble, config: EvaluatorServerConfig):
         self._max_batch_size: int = 500
         self._batching_interval: float = 2.0
         self._complete_batch: asyncio.Event = asyncio.Event()
+        self._server_started: asyncio.Event = asyncio.Event()
+        self._clients_connected: set[bytes] = set()
```
jonathan-eq, line 69: I think `self._connected_clients` would be a better name. `clients_connected` sounds more like a boolean

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -82,14 +65,20 @@ def __init__(self, ensemble: Ensemble, config: EvaluatorServerConfig):
...
+        self._clients_empty: asyncio.Event = asyncio.Event()
+        self._dispatchers_connected: set[bytes] = set()
```
jonathan-eq, line 71: Same here, `self._dispatchers_connected` -> `self._connected_dispatchers`.

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: Also, that is client? Is it the Client we use for sending/receiving, or is it referring to the monitor for UI?

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -82,14 +65,20 @@ def __init__(self, ensemble: Ensemble, config: EvaluatorServerConfig):
...
+        self._dispatchers_empty: asyncio.Event = asyncio.Event()
 
     async def _publisher(self) -> None:
+        await self._server_started.wait()
```
jonathan-eq, line 75: Do we need this here?

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -196,140 +185,137 @@ async def _failed_handler(self, events: Sequence[EnsembleFailed]) -> None:
...
-    def store_client(self, websocket: ServerConnection) -> Generator[None, None, None]:
-        self._clients.add(websocket)
-        yield
-        self._clients.remove(websocket)
-
-    async def handle_client(self, websocket: ServerConnection) -> None:
-        with self.store_client(websocket):
+    async def handle_client(self, dealer: bytes, frame: bytes) -> None:
+        raw_msg = frame.decode("utf-8")
+        if raw_msg == "CONNECT":
```
jonathan-eq, line 190: Should we have the `CONNECT` string as a constant in client.py, and call it CONNECT_MSG?

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -196,140 +185,137 @@ async def _failed_handler(self, events: Sequence[EnsembleFailed]) -> None:
...
-                # Dispatchers may close the connection abruptly in the case of
-                #  * flaky network (then the dispatcher will try to reconnect)
-                #  * job being killed due to MAX_RUNTIME
-                #  * job being killed by user
-                logger.error(
-                    f"a dispatcher abruptly closed a websocket: {connection_error!s}"
+            await self._router_socket.send_multipart(
+                [dealer, b"", event_to_json(event).encode("utf-8")]
+            )
+        elif raw_msg == "DISCONNECT":
```
jonathan-eq, line 201: Same here, maybe it can be a constant in client.py `DISCONNECT_MSG`

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -196,140 +185,137 @@ async def _failed_handler(self, events: Sequence[EnsembleFailed]) -> None:
...
+            self._clients_connected.discard(dealer)
```
jonathan-eq, line 202: Using discard instead of remove is very good!

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -196,140 +185,137 @@ async def _failed_handler(self, events: Sequence[EnsembleFailed]) -> None:
...
+                await asyncio.wait_for(self._clients_empty.wait(), timeout=5)
+            except TimeoutError:
+                logger.warning(
+                    "Not all clients were disconnected when closing zmq server!"
+                )
+            logger.debug("Async server exiting.")
+        finally:
+            try:
+                self._router_socket.close()
+                zmq_context.destroy()
```
jonathan-eq, line 315:
```suggestion
                zmq_context.destroy()
```
The destroy method should close all associated sockets automatically

src/ert/ensemble_evaluator/monitor.py
```diff
@@ -116,45 +90,3 @@ async def track(
...
-                self._connected.set_result(None)
-                async for raw_msg in self._connection:
-                    event = event_from_json(raw_msg)
-                    await self._event_queue.put(event)
-            except (ConnectionRefusedError, ConnectionClosed, ClientError) as exc:
-                self._connection = None
-                self._connected = asyncio.Future()
-                logger.debug(
-                    f"Monitor connection to EnsembleEvaluator went down, reconnecting: {exc}"
-                )
```
jonathan-eq, line 160: This removed half of the monitor :D
Very good!!

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: Maybe we can rename it to monitors

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: Actually, dispatcher is a horrible name. Can we change it to fm_step_runners?

src/ert/ensemble_evaluator/evaluator_connection_info.py
```diff
@@ -5,18 +5,5 @@
 class EvaluatorConnectionInfo:
     """Read only server-info"""
 
-    url: str
-    cert: str | bytes | None = None
+    router_uri: str
```
jonathan-eq, line 8: Should we include `router` in this instance variable? This requires the reader to know the dealer/router pattern, and couples to the implementation very tightly.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -82,14 +65,20 @@ def __init__(self, ensemble: Ensemble, config: EvaluatorServerConfig):
         self._max_batch_size: int = 500
         self._batching_interval: float = 2.0
         self._complete_batch: asyncio.Event = asyncio.Event()
+        self._server_started: asyncio.Event = asyncio.Event()
+        self._clients_connected: set[bytes] = set()
+        self._clients_empty: asyncio.Event = asyncio.Event()
```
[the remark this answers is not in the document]
xjules: The benefit of having it as an event, that we can do `asyncio.wait_for(self._clients_empty.wait(), timeout=...)`. Otherwise, before closing evaluator I'd need to while loop and do the timeout manually.

src/ert/ensemble_evaluator/evaluator.py
xjules: Good suggestion! I thought about that and now it makes sense :+1:

src/ert/ensemble_evaluator/evaluator.py
xjules: Hm. I was trying to preserve the naming conventions we have before (which was clients), but maybe having it explicit as `Monitor` would make more sense. Then I would change `handle_client` -> `handle_monitor` as well.

src/ert/ensemble_evaluator/evaluator.py
xjules: I think this one is mainly for the tests, which not always run the full evaluation.

src/ert/ensemble_evaluator/evaluator.py
xjules: Now `ACK` works on both dispatcher & monitor

src/ert/ensemble_evaluator/evaluator.py
xjules: Done!

src/ert/ensemble_evaluator/evaluator_connection_info.py
xjules: This knowledge about router-dealer needs to be propagated regardless. I'd like to keep it explicit that this is to connect to router and nothing else :)

jonathan-eq:
@xjules Seems codspeed is very impressed by your changes 🤣

tests/ert/conftest.py
```diff
@@ -475,8 +475,8 @@ class MockESConfig(EvaluatorServerConfig):
         def __init__(self, *args, **kwargs):
             if "use_token" not in kwargs:
                 kwargs["use_token"] = False
-            if "generate_cert" not in kwargs:
-                kwargs["generate_cert"] = False
+            if sys.platform != "linux":
+                kwargs["use_ipc_protocol"] = True
```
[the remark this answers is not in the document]
xjules: This is meant only for MacOS GH Action runners.

src/_ert/forward_model_runner/client.py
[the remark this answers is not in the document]
xjules:
```python
async def process_message(self, msg: str) -> None:
        raise NotImplementedError("Only monitor can receive messages!")
```

src/_ert/forward_model_runner/client.py
xjules: I have made some exploration and the issue that the reporter tests are still fully synced and thus we need to preserve both (async and sync) versions alive.

src/_ert/forward_model_runner/client.py
xjules: In case we will not get `ACK` for a message, then it is guaranteed that no further message is being sent unless we received `ACK` correctly. If the Finished is handled then we have a gracious timeout in Event reporter which tries to send the remaining the messages.

src/_ert/forward_model_runner/client.py
xjules: Is there a reason to do that?

src/_ert/forward_model_runner/reporting/event.py
xjules: Yes, it can be replaced. Will do :+1:

src/_ert/forward_model_runner/reporting/event.py
xjules: Yes, that's the `start_time` of closing `event_publisher`. Maybe I can rename it `start_closing_time`?

src/_ert/forward_model_runner/reporting/event.py
xjules: I will look if it can be refactored, but fear that it wouldn't work and reporter calls reporter.report from synced context.

src/ert/ensemble_evaluator/config.py
xjules: I have removed that function and more ....

src/_ert/forward_model_runner/reporting/event.py
xjules: :shrug:

jonathan-eq (changes requested):

src/_ert/forward_model_runner/client.py
```diff
@@ -1,18 +1,12 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
JHolba: it is needed until python 3.14

src/_ert/forward_model_runner/client.py
```diff
@@ -21,112 +15,132 @@ class ClientConnectionError(Exception):
...
-        return self
-
-    async def __aexit__(
-        self, exc_type: Any, exc_value: Any, exc_traceback: Any
-    ) -> None:
-        if self.websocket is not None:
-            await self.websocket.close()
+    DEFAULT_MAX_RETRIES = 5
+    DEFAULT_ACK_TIMEOUT = 5
+    _receiver_task: asyncio.Task[None] | None
```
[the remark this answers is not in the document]
jonathan-eq: ^the receiver_task**

src/_ert/forward_model_runner/client.py
```diff
@@ -21,112 +15,132 @@ class ClientConnectionError(Exception):
...
+                logger.debug(
+                    f"{self.dealer_id} connection to evaluator went down, reconnecting: {exc}"
+                )
+            except asyncio.CancelledError:
+                self.term()
+                raise
+
+            retries -= 1
+            if retries > 0:
+                logger.info(f"Retrying... ({retries} attempts left)")
```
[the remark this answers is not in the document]
xjules: Yes, can do, although it is getting logged in the end if we fail.

src/_ert/forward_model_runner/client.py
```diff
@@ -21,112 +15,132 @@ class ClientConnectionError(Exception):
...
                 )
-                raise ClientConnectionClosedOK(error_msg) from exception
-            except (TimeoutError, InvalidHandshake, InvalidURI, OSError) as exception:
-                if retry == self._max_retries:
-                    error_msg = (
-                        f"Not able to establish the "
-                        f"websocket connection {self.url}! Max retries reached!"
-                        " Check for firewall issues."
-                        f" Exception from {type(exception)}: {exception!s}"
+                await asyncio.sleep(1)
```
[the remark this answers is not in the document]
xjules: Ideally this should never happen - only if the socket is forcibly closed. So I thought to give it some slack to recover. Not sure though, maybe we don't need it. The idea is to yield, since if the socket is closed `await asyncio.sleep(1)` allows other tasks to be run.

src/_ert/forward_model_runner/client.py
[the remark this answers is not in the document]
xjules: Yes, I put it there for a reason that I don't remember now :) Will try to move it to constructor as see if it works.

src/_ert/forward_model_runner/client.py
```diff
@@ -21,112 +15,132 @@ class ClientConnectionError(Exception):
     pass
 
 
-class ClientConnectionClosedOK(Exception):
-    pass
+CONNECT_MSG = "CONNECT"
+DISCONNECT_MSG = "DISCONNECT"
+ACK_MSG = b"ACK"
```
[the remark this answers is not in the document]
xjules: hm, will try.

src/_ert/forward_model_runner/client.py
```diff
@@ -21,112 +15,132 @@ class ClientConnectionError(Exception):
...
-
-    async def __aenter__(self) -> "Client":
-        return self
-
-    async def __aexit__(
-        self, exc_type: Any, exc_value: Any, exc_traceback: Any
-    ) -> None:
-        if self.websocket is not None:
-            await self.websocket.close()
+    DEFAULT_MAX_RETRIES = 5
```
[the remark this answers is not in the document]
xjules: "Lower" did you mean increase? But, I can do 10

src/_ert/forward_model_runner/client.py
```diff
@@ -21,112 +15,132 @@ class ClientConnectionError(Exception):
...
+        self._receiver_task = None
+
+    async def __aenter__(self) -> Self:
+        await self.connect()
+        return self
 
-    async def _send(self, msg: AnyStr) -> None:
-        for retry in range(self._max_retries + 1):
+    async def __aexit__(
+        self, exc_type: Any, exc_value: Any, exc_traceback: Any
```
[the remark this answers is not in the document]
xjules: This needs to be there as `ruff` complains otherwise.

src/_ert/forward_model_runner/client.py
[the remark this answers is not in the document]
jonathan-eq: Yes, we were lowering it in the PR. Lets do 10

src/_ert/forward_model_runner/client.py
```diff
@@ -21,112 +15,132 @@ class ClientConnectionError(Exception):
...
+                self.term()
+                raise
+
+            retries -= 1
+            if retries > 0:
+                logger.info(f"Retrying... ({retries} attempts left)")
+                await asyncio.sleep(backoff)
+                # this call is idempotent
+                self.socket.connect(self.url)
+                backoff = min(backoff * 2, 10)  # Exponential backoff
```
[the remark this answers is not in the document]
xjules: This will be increasing exponentially so I guess it is fine :+1:

src/_ert/forward_model_runner/reporting/event.py
xjules: OK

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -78,53 +75,51 @@ def __init__(self, evaluator_url, token=None, cert_path=None):
...
         self._event_queue: queue.Queue[events.Event | EventSentinel] = queue.Queue()
         self._event_publisher_thread = ErtThread(target=self._event_publisher)
-        self._timeout_timestamp = None
-        self._timestamp_lock = threading.Lock()
-        # seconds to timeout the reporter the thread after Finish() was received
-        self._reporter_timeout = 60
+        self._done = threading.Event()
+        self._ack_timeout = ack_timeout
+        self._max_retries = max_retries
+        self._finished_event_timeout = finished_event_timeout or 60
```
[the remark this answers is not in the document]
xjules: yes, was thinking about it. But could the timeout be ever 0? Let's change it though - just in case :+1:

src/ert/ensemble_evaluator/evaluator.py
xjules: I've decided to keep the names (for the legacy reason). when the dust settles we can rename it.

tests/ert/conftest.py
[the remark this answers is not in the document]
xjules: Yes, but we should test tcp sockets too. Hm, let's see.

tests/ert/unit_tests/shared/test_port_handler.py
```diff
@@ -311,6 +311,7 @@ def test_reuse_active_close_nok_ok(unused_tcp_port):
     assert sock.fileno() != -1
 
 
+# This test is disabled because it is not clear if zmq needs it
```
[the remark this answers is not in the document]
xjules: I've re-enabled it.

tests/ert/unit_tests/ensemble_evaluator/test_monitor.py
```diff
@@ -1,109 +1,158 @@
...
+        async_zmq_server(unused_tcp_port, mock_event_handler)
+    )
+    async with monitor:
+        pass
+    await websocket_server_task
+    dealer, msg = messages[0]
+    assert dealer.startswith("client-")
+    assert msg == CONNECT_MSG
+    dealer, msg = messages[1]
+    assert dealer.startswith("client-")
```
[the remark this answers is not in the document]
xjules: nah, that is not required. The only thing that matters is that it is a monitor.

tests/ert/utils.py
```diff
@@ -62,101 +61,68 @@ def wait_until(func, interval=0.5, timeout=30):
...
-            router_socket = zmq_context.socket(zmq.ROUTER)
-            router_socket.bind(f"tcp://*:{port}")
-            handler_task = asyncio.create_task(_handler(router_socket))
-            await handler_task
-            router_socket.close()
-
-        loop.run_until_complete(_run_server())
-        loop.close()
+    def signal(self, value):
+        self.value = value
```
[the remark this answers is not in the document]
xjules: the signal(value) has more meaning than just "stop". Will put a comment there.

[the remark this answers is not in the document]
xjules: I've added a comment:
```
"""Mock ZMQ server for testing
        signal = 0: normal operation
        signal = 1: don't send ACK and don't receive messages
        signal = 2: don't send ACK, but receive messages
        """
```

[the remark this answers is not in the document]
jonathan-eq: Very good!

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -59,14 +54,16 @@ class Event(Reporter):
...
     _sentinel: Final = EventSentinel()
 
-    def __init__(self, evaluator_url, token=None, cert_path=None):
+    def __init__(
+        self,
+        evaluator_url,
+        token=None,
+        ack_timeout=None,
+        max_retries=None,
+        finished_event_timeout=None,
```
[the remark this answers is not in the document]
xjules: Done

src/_ert/forward_model_runner/client.py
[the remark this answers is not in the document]
xjules: done

src/ert/ensemble_evaluator/evaluator.py
xjules: I would do it in another PR to keep the semantics as it was before.

jonathan-eq (approved):
Very good, let's start the new year without websockets 🥇

PR #9190, opened 2024-11-12
Assume non-LSF host error is flaky

### LINKED ISSUE #9185: Ert crashes with RuntimeError: Request from non-LSF host rejected. Job not submitted.
After the implementation of floating LSF licenses (implying that `bsub` can submit to the LSF cluster directly, and not via `ssh`'ing to some LSF server host, we are seeing some error message in the logs with the message:

`Failed to submit: Request from non-LSF host rejected. Job not submitted.`

This occurs both as a RuntimeError when starting an ensemble, and is then probably a fatal error for the ensemble.

Additionally it can be seen as a response when doing polling (every 2 seconds), then it might spam the user with error message, but still carry on.

berland:
The LSF driver experiences crashes stemming from bsub returning with the error message 'Request from non-LSF host rejected'. There are reasons to believe this is not a permanent error, but some flakyness in the IP infrastructure, and thus should should be categorized as a retriable failure.

The reason for believing this is flakyness is mostly from the fact that the same error is also seen on 'bjobs'-calls. If it was a permanent failure scenario, there would be an enourmous amount of error from these bjobs calls, but there is not.

**Issue**
Resolves #9185 

**Approach**

## When applicable

### DIFF
--- a/src/ert/scheduler/lsf_driver.py
+++ b/src/ert/scheduler/lsf_driver.py
@@ -94,7 +94,14 @@ _STATE_ORDER: dict[Type[AnyJob], int] = {
 LSF_INFO_JSON_FILENAME = "lsf_info.json"
 FLAKY_SSH_RETURNCODE = 255
 JOB_ALREADY_FINISHED_BKILL_MSG = "Job has already finished"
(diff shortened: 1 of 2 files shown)


sondreso (approved):
:ship:

PR #9193, opened 2024-11-12
Give warning if runpath disk space is close to full on ert startup

### LINKED ISSUE #8921: Consider displaying the status of free disk spacefor runpath in the Ert gui
It frequently happens that users run out of disk space during an Ert run. Maybe we could consider displaying information about the current state of free disk space left in the runpath in the Gui, with color encodings related to space left? Primarily before starting a run.

larsevj:
**Issue**
Resolves #8921 

The following is presented if disk is "close" to full:
image

## When applicable

### DIFF
--- a/src/ert/config/model_config.py
+++ b/src/ert/config/model_config.py
@@ -5,13 +5,12 @@ import logging
 import os.path
 import shutil
 from datetime import datetime
-from pathlib import Path
 from typing import no_type_check
 
 from pydantic import field_validator
 from pydantic.dataclasses import dataclass
 
-from ert.shared.status.utils import byte_with_unit
+from ert.shared.status.utils import byte_with_unit, get_mount_directory
 
 from .parsing import (
     ConfigDict,
@@ -94,7 +93,7 @@ class ModelConfig:
             ConfigWarning.warn(msg)
             logger.warning(msg)
         with contextlib.suppress(Exception):
-            mount_dir = _get_mount_directory(runpath_format_string)
+            mount_dir = get_mount_directory(runpath_format_string)
             total_space, used_space, free_space = shutil.disk_usage(mount_dir)
             percentage_used = used_space / total_space
             if (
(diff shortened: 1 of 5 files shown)


src/ert/config/model_config.py
```diff
@@ -37,6 +48,9 @@ def str_to_datetime(date_str: str) -> datetime:
 DEFAULT_JOBNAME_FORMAT = "<CONFIG_FILE>-<IENS>"
 DEFAULT_ECLBASE_FORMAT = "ECLBASE<IENS>"
 
+FULL_DISK_PERCENTAGE_THRESHOLD = 0.97
+MINIMUM_BYTES_LEFT_ON_DISK = 200 * 1024**3  # 200 GB required
```
berland, line 52: Hmm, we will get some complaints from all the Raspberry Pi users out there with a 16Gb memory card.

berland: Could we in some way leave this number for `ert-configurations` to set?

berland: The `0.97` is easier to accept as a general rule in Erts open source code.

berland: Should there be a comment in the source code on this?

tests/ert/unit_tests/config/test_model_config.py
```diff
@@ -61,3 +64,13 @@ def test_that_invalid_time_map_file_raises_config_validation_error(tmpdir):
 
         with pytest.raises(ConfigValidationError, match="Could not read timemap file"):
             _ = ModelConfig.from_dict({ConfigKeys.TIME_MAP: "time_map.txt"})
+
+
+def test_warning_when_full_disk(tmp_path):
```
berland, line 69: Can this test be parameterized over (percentage, freespace, expectation) to also test the conditions where we should not trigger the warning?

berland (approved):
👍🏻

eivindjahren:
Should this check be on startup or on experiment start?

eivindjahren:
> > Should this check be on startup or on experiment start?
> 
> Might be better to be on experiment startup, seems like a lot of users ignore warnings on startup. Maybe we also should check space left on enspath?

Yes, we should check there also, we have seen a lot of errors from missing capacity in enspath. Also, for enspath we could estimate how much space is necessary, but I suggest we keep that for another PR.

eivindjahren:
I updated the description of the issue with a new approach to doing this.

eivindjahren:
This seems ready to squash and merge now?

PR #9228, opened 2024-11-15
Fix walltime computation bug

### LINKED ISSUE #9216: Bug in walltime computation for logged forward model steps
This code is seen compute blatantly wrong walltime durations for some steps:

https://github.com/equinor/ert/blob/91310f99e95e135cc2e63fbce0640033352e9c7b/src/ert/ensemble_evaluator/_ensemble.py#L164

like:
          
`forward_model_job.success COPY_DIRECTORY walltime=4873.97593 cpu_seconds=0.0 current_memory_usage=12288 step_index=12 real=188 ensemble=c43078a7-d31a-42bd-bb14-744xxxx`

berland:
Because we logged finished step events before the current batch of events were processed, there were situations where the snapshot from a previous failed realization was out of date.

This solves that problem. There is still a residual problem that will surface if walltime is to be computed for all steps: If a failure event for a step and its success event from the subsequent run is in the same batch of events, we can get a negative walltime computed. This problem is ignored as the walltime computation is only interesting (currently) for walltimes > 120 seconds, where this cannot happen (max time between batches is 2 seconds).

**Issue**
Resolves #9216 

**Approach**

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/_ensemble.py
+++ b/src/ert/ensemble_evaluator/_ensemble.py
@@ -181,6 +181,14 @@ class LegacyEnsemble:
     def update_snapshot(self, events: Sequence[Event]) -> EnsembleSnapshot:
         snapshot_mutate_event = EnsembleSnapshot()
         for event in events:
+            snapshot_mutate_event = snapshot_mutate_event.update_from_event(
+                event, source_snapshot=self.snapshot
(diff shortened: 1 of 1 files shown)


andreas-el (approved):
Gotta process all! 💯

PR #9299, opened 2024-11-21
Fix progress bar showing finished status for unselected reals

### LINKED ISSUE #9230: Progress bar shows finished status for non existing realizations

Started only 3 realizations on/of an ensemble experiment. Progress bar shows 247 finished realizations which does not exist. 
Image

Progress bar and status only to include the 3 realizations.

### steps to reproduce

```Python

```

- [ ] python 3.10
- [ ] python 3.11
- [ ] python 3.12
- [ ] macosx
- [ ] rhel7
- [x] rhel8
- [ ] local queue
- [x] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

jonathan-eq:
**Issue**
Resolves #9230

**Approach**
This commit fixes the bug where the gui shows the total count of realizations as the ensemble_size rather than the number of realizations actually selected for the run.

image

## When applicable

### DIFF
--- a/src/ert/run_models/base_run_model.py
+++ b/src/ert/run_models/base_run_model.py
@@ -188,7 +188,6 @@ class BaseRunModel(ABC):
         self.minimum_required_realizations = minimum_required_realizations
         self.active_realizations = copy.copy(active_realizations)
         self.start_iteration = start_iteration
-        self.validate_active_realizations_count()
 
     def log_at_startup(self) -> None:
         keys_to_drop = [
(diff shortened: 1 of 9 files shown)


src/ert/run_models/base_run_model.py
```diff
@@ -406,7 +405,9 @@ def get_current_status(self) -> dict[str, int]:
                 for real in all_realizations.values():
                     status[str(real["status"])] += 1
 
-        status["Finished"] += self.active_realizations.count(False)
+        status["Finished"] += self.active_realizations.count(
```
xjules, line 408: It would be nice to have a small comment / explanation here.

jonathan-eq: Yes, either a comment or put it in a separate function and have a unit test on it.
The problem was that when we added all the `self.active_realizations.count(False)`, we also add the realizations that have deselected from the run. 
The idea was to use the the original `'active_realizations'`-list (which was already an instance variable `self_initial_realizations_mask`), and subtract this from the previous number.

src/ert/run_models/base_run_model.py
```diff
@@ -406,9 +405,14 @@ def get_current_status(self) -> dict[str, int]:
                 for real in all_realizations.values():
                     status[str(real["status"])] += 1
 
-        status["Finished"] += self.active_realizations.count(False)
+        status["Finished"] += self._get_number_of_finished_realizations_from_reruns()
         return status
 
+    def _get_number_of_finished_realizations_from_reruns(self) -> int:
```
xjules, line 411: Nice!

xjules (approved):
Nice PR @jonathan-eq ! :rocket:

PR #9375, opened 2024-11-28
Migrate finalized keys for response configs

yngve-sk:
Required with https://github.com/equinor/ert/commit/b5d2b40f66688dd25dbe4dc9822034dd3c9ba442 
otherwise we get migrated response configs that are missing `has_finalized_keys`

### DIFF
--- a/src/ert/storage/migration/to9.py
+++ b/src/ert/storage/migration/to9.py
@@ -1,0 +1,86 @@
+import json
+import os
+from pathlib import Path
+from tempfile import NamedTemporaryFile
+
+import polars
+
+info = "Migrate finalized response keys into configs"
+
+
+def _write_transaction(filename: str | os.PathLike[str], data: bytes) -> None:
+    """
+    Writes the data to the filename as a transaction.
+
+    Guarantees to not leave half-written or empty files on disk if the write
+    fails or the process is killed.
+    """
+
+    Path("./swp").mkdir(parents=True, exist_ok=True)
+    with NamedTemporaryFile(dir="./swp", delete=False) as f:
+        f.write(data)
+        os.rename(f.name, filename)
+
+
+def migrate(path: Path) -> None:
+    for experiment in path.glob("experiments/*"):
+        ensembles = [*path.glob("ensembles/*")]
+
+        with (
+            open(experiment / "index.json", encoding="utf-8") as f_experiment,
+            open(
+                experiment / "responses.json", mode="r", encoding="utf-8"
+            ) as f_responses,
+        ):
+            exp_index = json.load(f_experiment)
+            experiment_id = exp_index["id"]
+
+            responses_config = json.load(f_responses)
+            for response_type, config in responses_config.items():
+                if not config.get("has_finalized_keys"):
+                    # Read a sample response and write the keys
+                    for ens in ensembles:
+                        with open(ens / "index.json", encoding="utf-8") as f_ensemble:
+                            ens_file = json.load(f_ensemble)
+                            if ens_file["experiment_id"] != experiment_id:
+                                continue
+
+                        real_dirs = [*ens.glob("realization-*")]
+
+                        for real_dir in real_dirs:
+                            if (real_dir / f"{response_type}.parquet").exists():
+                                df = polars.scan_parquet(
+                                    real_dir / f"{response_type}.parquet"
+                                )
+                                response_keys = (
+                                    df.select("response_key")
+                                    .unique()
+                                    .collect()
+                                    .to_series()
+                                    .to_list()
+                                )
+                                config["has_finalized_keys"] = True
+                                config["keys"] = sorted(response_keys)
+                                break
+
+                        if config.get("has_finalized_keys"):
+                            break
+
+                # If this is hit, it means no responses were found
+                # for that response type, so we still cannot have "finalized"
+                # keys. We then default it to that of the configs.
+                # At time of writing, it is always True for GenDataConfig
+                # and False for SummaryConfig
+                if "has_finalized_keys" not in config:
+                    config["has_finalized_keys"] = (
+                        config["_ert_kind"] != "SummaryConfig"
+                    )
+
+        _write_transaction(
+            experiment / "responses.json",
+            json.dumps(
+                responses_config,
+                default=str,
+                indent=2,
+            ).encode("utf-8"),
+        )
--- a/tests/ert/unit_tests/storage/migration/test_version_2.py
+++ b/tests/ert/unit_tests/storage/migration/test_version_2.py
@@ -24,10 +24,19 @@ def test_migrate_responses(setup_case, set_ert_config):
         response_info = json.loads(
             (experiment._path / "responses.json").read_text(encoding="utf-8")
         )
-        assert (
-            experiment.response_configuration
-            == ert_config.ensemble_config.response_configs
-        )
+
+        response_config_exp = experiment.response_configuration
+        response_config_ens = ert_config.ensemble_config.response_configs
+
+        # From storage v9 and onwards the response config is mutated
+        # when migrating an existing experiment, because we check that the
+        # keys in response.json are aligned with the dataset.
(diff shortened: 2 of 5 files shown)


oyvindeide (commented):
Have one question, could you also check in some storages to  ert-testdata and add them to the tests?

tests/ert/unit_tests/storage/snapshots/test_storage_migration/test_that_storage_matches/responses
```diff
@@ -1 +1 @@
-{'gen_data': GenDataConfig(name='gen_data', input_files=['gen%d.txt'], keys=['GEN'], has_finalized_keys=True, report_steps_list=[[1]]), 'summary': SummaryConfig(name='summary', input_files=['CASE'], keys=['FOPR', 'RWPR'], has_finalized_keys=False, refcase={})}
```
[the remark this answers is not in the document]
yngve-sk: Not in the dataset in any of the tests. In the ert config the only summary key specified is `FOPR`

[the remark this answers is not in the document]
oyvindeide: We checked this, and the second key is present in the refcase, but not the actual response, so this change is what we want, as responses.json and the dataset are now in sync.

src/ert/storage/migration/to9.py
```diff
@@ -0,0 +1,71 @@
...
+import polars
+
+info = "Migrate finalized response keys into configs"
+
+
+def _migrate_response_configs_wrt_finalized_keys(path: Path) -> None:
+    for experiment in path.glob("experiments/*"):
+        ensembles = path.glob("ensembles/*")
+
+        experiment_id = None
```
oyvindeide, line 14: Why do you set this to `None`?

src/ert/storage/migration/to9.py
```diff
@@ -0,0 +1,71 @@
...
+        with open(experiment / "index.json", encoding="utf-8") as f:
+            exp_index = json.load(f)
+            experiment_id = exp_index["id"]
+
+        responses_config = None
+        with open(experiment / "responses.json", mode="r", encoding="utf-8") as f:
+            responses_config = json.load(f)
+            for response_type, config in responses_config.items():
+                if not config.get("has_finalized_keys"):
```
oyvindeide, line 23:
```suggestion
                if "has_finalized_keys not in config:
```
just a small suggestion I at least think reads better

src/ert/storage/migration/to9.py
```diff
@@ -0,0 +1,71 @@
...
+                                    .to_list()
+                                )
+                                config["has_finalized_keys"] = True
+                                config["keys"] = sorted(response_keys)
+                                break
+
+                        if config["has_finalized_keys"]:
+                            break
+
+                if "has_finalized_keys" not in config:
```
oyvindeide, line 52: How can we hit this, but miss the one above?

src/ert/storage/migration/to9.py
```diff
@@ -0,0 +1,71 @@
...
+                    # At this point in "storage history",
+                    # only gendata and summary response types
+                    # exist, and only summary starts without finalized keys
+                    config["has_finalized_keys"] = (
+                        config["_ert_kind"] != "SummaryConfig"
+                    )
+
+        os.remove(experiment / "responses.json")
```
oyvindeide, line 60: Use transaction?

src/ert/storage/migration/to9.py
```diff
@@ -0,0 +1,71 @@
...
+            json.dump(
+                responses_config,
+                f,
+                default=str,
+                indent=2,
+            )
+
+
+def migrate(path: Path) -> None:
+    _migrate_response_configs_wrt_finalized_keys(path)
```
oyvindeide, line 71: Suggest just in-lining this function instead of just directly calling the private function

tests/ert/unit_tests/storage/migration/test_version_2.py
```diff
@@ -24,10 +24,19 @@ def test_migrate_responses(setup_case, set_ert_config):
...
-            experiment.response_configuration
-            == ert_config.ensemble_config.response_configs
-        )
+
+        response_config_exp = experiment.response_configuration
+        response_config_ens = ert_config.ensemble_config.response_configs
+
+        # From storage v9 and onwards the response config is mutated
+        # when migrating an existing experiment, hence we expect different
+        # keys for summary config in the experiment.
```
oyvindeide, line 33:
```suggestion
        # From storage v9 and onwards the response config is mutated
        # when migrating an existing experiment, because we check that the 
        # keys in response.json are aligned with the dataset.
```

oyvindeide (commented):
Looks good! I have some small suggestions and questions

src/ert/storage/migration/to9.py
yngve-sk: Added a comment explaining why, basically if no responses were found, we want to it to be the default in the response config for that type

src/ert/storage/migration/to9.py
yngve-sk: Done, copypasted in the function from local_storage to keep it from importing anything from ERT

src/ert/storage/migration/to9.py
```diff
@@ -49,22 +66,25 @@ def _migrate_response_configs_wrt_finalized_keys(path: Path) -> None:
                         if config["has_finalized_keys"]:
```
oyvindeide, line 66: This would fail with `KeyError` should it not be the case

src/ert/storage/migration/to9.py
```diff
@@ -0,0 +1,87 @@
...
+            ) as f_responses,
+        ):
+            exp_index = json.load(f_experiment)
+            experiment_id = exp_index["id"]
+
+            responses_config = json.load(f_responses)
+            for response_type, config in responses_config.items():
+                if "has_finalized_keys" not in config:
+                    # Read a sample response and write the keys
+                    for ens in ensembles:
```
oyvindeide, line 42: Perhaps just pick the first one? Think that would simplify the logic somewhat?

src/ert/storage/migration/to9.py
```diff
@@ -0,0 +1,87 @@
...
+                # for that response type, so we still cannot have "finalized"
+                # keys. We then default it to that of the configs.
+                # At time of writing, it is always True for GenDataConfig
+                # and False for SummaryConfig
+                if "has_finalized_keys" not in config:
+                    config["has_finalized_keys"] = (
+                        config["_ert_kind"] != "SummaryConfig"
+                    )
+
+        os.remove(experiment / "responses.json")
```
oyvindeide, line 79: Then this should be removed?

src/ert/storage/migration/to9.py
yngve-sk: I'm not completely sure, wrt edge cases / failed ensembles, unless we can somehow pick the "prior" ensemble always (without the complexity of finding the ensemble with iter=0 which is similar to the complexity of break'ing out)

src/ert/storage/migration/to9.py
yngve-sk: If the transaction just overwrites yes? AFAIK it just writes (maybe that is also an overwrite). Why not keep the removal explicit though? I'm not sure if the transaction is meant to captured file writing failures only, or also file overwriting failures and preserve the to-be-overwritten file?

oyvindeide: It was if the transaction fail, if it would it would be left in an inconsistent state

src/ert/storage/migration/to9.py
yngve-sk: Fixed the logic and added so this case is covered

src/ert/storage/migration/to9.py
yngve-sk: Did not understand at first, but think I see now, if we don't remove it we'll be able to retry doing the migration in this case

oyvindeide (approved):
LGTM!

PR #9385, opened 2024-11-28
Enforce correct dtypes in migration

yngve-sk:
Ensures correctly ported dtypes for port, will otherwise lead to downstream errors. Should be backported to 11.1


StephanDeHoop (approved):
LGTM!

PR #9390, opened 2024-11-28
Ignore erroneous auto-scale configuration, but warn

### LINKED ISSUE #9339: ERT fails in update step if AUTO_SCALE cannot match any observations

User report where there was a mismatch in the name of the observations and the string given to AUTO_SCALE, e.g. observations was named `obs_*` whereas the input for AUTO_SCALE was `OBS_*` . This caused ERT to fail with the following message:
EndEvent(failed=True, msg="cannot concat empty list\nopening handshake failed\nopening handshake failed\nNo observations active for group: ['OBS_*']")

I would except ert to not crash in this case, and ideally we should give a warning that there are no matches during the config parsing to avoid typos and such?

### steps to reproduce

```Python
Add following line to poly.ert: ANALYSIS_SET_VAR OBSERVATIONS AUTO_SCALE OBS_*
run es_mda
```

- [ ] python 3.10
- [x] python 3.11
- [ ] python 3.12
- [ ] macosx
- [ ] rhel7
- [x] rhel8
- [ ] local queue
- [ ] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

berland:
**Issue**
Resolves #9339 

**Approach**
Detect and ignore (but warn)

## When applicable

### DIFF
--- a/src/ert/analysis/_es_update.py
+++ b/src/ert/analysis/_es_update.py
@@ -349,11 +349,19 @@ def _load_observations_and_responses(
                 )
             )
 
-        scaling_factors_df = polars.concat(scaling_factors_dfs)
-        ensemble.save_observation_scaling_factors(scaling_factors_df)
+        if len(scaling_factors_dfs):
+            scaling_factors_df = polars.concat(scaling_factors_dfs)
+            ensemble.save_observation_scaling_factors(scaling_factors_df)
 
-        # Recompute with updated scales
-        scaled_errors = errors * scaling
+            # Recompute with updated scales
+            scaled_errors = errors * scaling
+        else:
+            msg = (
+                f"WARNING: Could not auto-scale the observations {auto_scale_observations}. "
+                f"No match with existing active observations {obs_keys}"
+            )
+            logger.warning(msg)
+            print(msg)
 
     update_snapshot = []
     for (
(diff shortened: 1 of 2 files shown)


berland:
(test missing)

test-data/ert/poly_example/poly.ert
```diff
@@ -6,9 +6,11 @@ RUNPATH poly_out/realization-<IENS>/iter-<ITER>
 OBS_CONFIG observations
 REALIZATION_MEMORY 50mb
 
-NUM_REALIZATIONS 100
+NUM_REALIZATIONS 10
```
eivindjahren, line 9: Not sure if we should change anything in the test-data. Maybe we could instead just edit this in the test.

eivindjahren (changes requested):
Currently edits the test-data in place, this should instead just be handled in the test.

test-data/ert/poly_example/poly.ert
berland: this is just a boo-boo, will fix. Note to self, remove  `git commit -a` from muscle memory.

eivindjahren: No problem, it was just in the ready for review column so thought I would comment on it to be certain.

berland: fixed

eivindjahren (approved):
:100:

PR #9426, opened 2024-12-03
Make corrXY work with multiple GEN_KWs

dafeda:
## Introduction

A while back we introduced a call-back to the `iterative_ensemble_smoother` to allow fetching the cross-correlation matrix between parameters and responses.
This matrix was saved into ert's internal storage as `corr_XY.nc` and the `load_cross_correlations` API end-point was created. 
The idea was to provide users with the raw data and let them analyse and visualise it as they saw fit.
This feature was not clearly communicated to our users and was therefore not used at all.
Also, it turns out that the implementation is buggy in that it overwrites existing data such that only the `GEN_KW` that was specified last in ert's config is available.
Works for the poly-case, but not for configs with multiple `GEN_KW`'s.

## Approach

Save cross correlations to a `parquet` file with the following schema:

image

The `obs_name` is created by appending a suffix to `obs_group`. The suffix is a counter that depends on the number of observations within each group.
Data is stored in ["tidy format"](https://vita.had.co.nz/papers/tidy-data.pdf) to make it easy for users to query the data however they want.

## Testing using the poly-case (see `Plot_correlations.ipynb` in the poly_example test folder)

Note that the poly-case has just one `GEN_KW`.

image

## Testing using the heat equation (see `Plot_correlations.ipynb` in the heat_equation test folder)

To test with multiple `GEN_KW`s, I've added two `GEN_KW`s to the heat-equation, and created a Jupyter Notebook that plots correlations for all iterations. See diff for all plots. Here's an example of one of the parameters at iteration 0:

image

## Testing using Drogon

The FMU-research team has done testing using Drogon.

Here's an example I created using Drogon:

image

## When applicable


xjules:
There are some failing tests though.

src/ert/analysis/_es_update.py
```diff
@@ -474,6 +478,14 @@ def adaptive_localization_progress_callback(
     num_obs = len(observation_values)
 
     smoother_snapshot.update_step_snapshots = update_snapshot
+    # Used as labels for observations in cross-correlation matrix.
+    # Say we have two observation groups "FOPR" and "WOPR" where "FOPR" has
```
xjules, line 482: This is to accommodate having multiple "observed" values under the same observation key? I thought that every observation point needs to have a unique name already or not?

xjules: Oh I see:
```
GENERAL_OBSERVATION MY_OBS_10 {
    DATA       = MY_RESPONSE;
    RESTART    = 10;
    OBS_FILE   = obs_10.txt;
};
```
Now I understand :+1:

src/ert/analysis/_es_update.py
```diff
@@ -474,6 +478,14 @@ def adaptive_localization_progress_callback(
...
+    # 2 responses and "WOPR" has 3.
+    # In this case we create a list [FOPR_0, FOPR_1, WOPR_0, WOPR_1, WOPR_2]
```
oyvindeide, line 484: Would it be possible to use the index for the observations here instead? I guess it is ok for summary observations, though perhaps a bit confusing as this number will not align with `report_step`, but for `GEN_OBS` which already have an int index it would probably not align with that, and could be confusing.

src/ert/storage/local_ensemble.py
```diff
@@ -592,17 +592,28 @@ def save_cross_correlations(
...
+                "obs_group": observation_keys * len(parameter_names),
+                "obs_name": unique_obs_names * len(parameter_names),
+                "value": cross_correlations.ravel(),
+            }
+        )
+
+        file_path = os.path.join(self.mount_point, "corr_XY.parquet")
+        if os.path.exists(file_path):
+            existing_df = pl.read_parquet(file_path)
+            df = pl.concat([existing_df, new_df])
```
oyvindeide, line 613: This might be a bit confusing if we just merge with existing data?

src/ert/analysis/_es_update.py
dafeda: Not sure. Do I remember correctly that the index is a time-stamp or date for summary observations? We could perhaps still use it but perhaps a bit messy.

src/ert/storage/local_ensemble.py
dafeda: What do you find confusing?

oyvindeide: If there is something on disk, we merge with that? Couldnt that potentially be done with a different threshold, and so no longer valid? I might be misunderstanding.

src/ert/analysis/_es_update.py
oyvindeide: Yes, was thinking a bit more about this, and should we consider doing it like we do for the snapshot in the gui showing the scaling factors when doing auto scaling? Those events are automatically written to csv by the client, so will be available to the user.

Something like: https://github.com/equinor/ert/blob/71275280104b46942e41a907a76342e262928aeb/src/ert/analysis/_es_update.py#L239, which alread uses the indices: https://github.com/equinor/ert/blob/71275280104b46942e41a907a76342e262928aeb/src/ert/analysis/_es_update.py#L252

dafeda: As far as I know, these are the two types of indexes we have:

Screenshot 2025-01-08 at 09 02 25

Screenshot 2025-01-08 at 09 03 47

Are you suggesting we use the second part of the string as an index?

src/ert/storage/local_ensemble.py
dafeda: Good point. `self.mount_point` points to a specific ensemble. Can we run the same ensemble with different thresholds?

image

src/ert/analysis/_es_update.py
oyvindeide: > Are you suggesting we use the second part of the string as an index?

Yes, or perhaps send the whole thing as an event instead of saving it in storage. Then everyone gets the benefit right away.

dafeda: Not sure I follow the event thing. Would we then still be able to analyse the data from a Jupyter Notebook?

dafeda:
Closing this for now as it's getting somewhat stale and I think EnIF is a better approach.

PR #9440, opened 2024-12-04
Improve exit code detection in the slurm driver

verveerpj:
**Issue**
The slurm driver is performing very badly in some cases, when many small jobs were submitted. As it turns out, the use of `scontrol` to extract the exit code of finished jobs may fail, since SLURM does not maintain the information of finished jobs indefinitely in memory for use with `scontrol`. If this happens this leads to many retries of `scontrol`, and the exit code is finally not retrieved. In such a case it is possible to fall back on the `sattr` command which is able to retrieve the information for jobs.

This PR adds `sattr` as a fallback in cases where `scontrol` fails. The `scontrol` command is still preferred and tried first, because depending on the configuration of the SLURM cluster it many not be guaranteed that `sattr` can be used. If that is the case, the code falls back to its old behavior of retrying.

**Approach**
- Add `sattr` command to the slurm options.
- If `scontrol` failure is detected, try `sattr`.
- If `sattr` is not available or fails, fall back on the old retry behavior.

## When applicable

### DIFF
--- a/src/ert/scheduler/slurm_driver.py
+++ b/src/ert/scheduler/slurm_driver.py
@@ -67,6 +67,7 @@ class SlurmDriver(Driver):
         include_hosts: str = "",
         squeue_cmd: str = "squeue",
         scontrol_cmd: str = "scontrol",
+        sacct_cmd: str = "sacct",
         scancel_cmd: str = "scancel",
         sbatch_cmd: str = "sbatch",
         user: Optional[str] = None,
@@ -119,6 +120,7 @@ class SlurmDriver(Driver):
         self._squeue = squeue_cmd
 
         self._scontrol = scontrol_cmd
+        self._sacct = sacct_cmd
         self._scontrol_cache_timestamp = 0.0
         self._scontrol_required_cache_age = 30
         self._scontrol_cache: dict[str, ScontrolInfo] = {}
@@ -367,6 +369,18 @@ class SlurmDriver(Driver):
         ) and missing_job_id in self._scontrol_cache:
             return self._scontrol_cache[missing_job_id]
 
+        info = await self._run_scontrol(missing_job_id)
+        if info is None:
+            logger.error("scontrol failed, trying sacct")
+            info = await self._run_sacct(missing_job_id)
+        if info is None:
+            return None
+
+        self._scontrol_cache[missing_job_id] = info
+        self._scontrol_cache_timestamp = time.time()
+        return info
+
+    async def _run_scontrol(self, missing_job_id: str) -> Optional[ScontrolInfo]:
         process = await asyncio.create_subprocess_exec(
             self._scontrol,
             "show",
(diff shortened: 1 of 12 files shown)


berland:
`_poll_once_by_scontrol() is already a fallback when  `squeue` fails in `poll()`. This will imply a fallback to the fallback for the every-2-second polling. That might be fine though. Intentional?

berland:
There are mocked bash/python scripts for the squeue and scontrol commands. Could you maybe write a mocked sacct command and include tests for this scenario? (that is quite nice to have when developing without access to a real slurm cluster)

verveerpj:
> `_poll_once_by_scontrol() is already a fallback when `squeue`fails in`poll()`. This will imply a fallback to the fallback for the every-2-second polling. That might be fine though. Intentional?

Yes, it seems to be there because `squeue` will not report on jobs that have finished, but we need the exit code. `scontrol` will still report on them for a while, but then it also will fail if too much time has passed. Hence `sacct` as a fallback for the fallback.

verveerpj:
> There are mocked bash/python scripts for the squeue and scontrol commands. Could you maybe write a mocked sacct command and include tests for this scenario? (that is quite nice to have when developing without access to a real slurm cluster)

Yes, will do.

verveerpj:
It looks like a test fails on github because sattr is not set up. I have to look into how to do that, or change the test.

tests/ert/unit_tests/scheduler/test_slurm_driver.py
```diff
@@ -450,3 +459,26 @@ async def finished(iens: int, returncode: int):
     assert not Path(
         "survived"
     ).exists(), "The process children of the job should also have been killed"
+
+
```
berland, line 463: Would be nice to also test that we can obtain a zero exit code as well, parametrize with `[("true", 0), )("false", 1)]` ?

src/ert/scheduler/slurm_driver.py
```diff
@@ -367,6 +369,18 @@ async def _poll_once_by_scontrol(
         ) and missing_job_id in self._scontrol_cache:
             return self._scontrol_cache[missing_job_id]
 
+        info = await self._run_scontrol(missing_job_id)
+        if info is None:
+            logger.error("scontrol failed, trying sacct")
```
berland, line 374: This is probably a `logger.warning` since we have a fallback.

src/ert/scheduler/slurm_driver.py
```diff
@@ -384,17 +398,47 @@ async def _poll_once_by_scontrol(
             )
             return None
 
-        info = None
         try:
-            info = _parse_scontrol_output(stdout.decode(errors="ignore"))
+            return _parse_scontrol_output(stdout.decode(errors="ignore"))
         except Exception as err:
             logger.error(
```
berland, line 404: warning here too then (but I see this is not your change)

src/ert/scheduler/slurm_driver.py
```diff
@@ -384,17 +398,47 @@ async def _poll_once_by_scontrol(
...
+                "-j",
+                str(missing_job_id),
+                stdout=asyncio.subprocess.PIPE,
+                stderr=asyncio.subprocess.PIPE,
+            )
+        except FileNotFoundError:
+            logger.error("sacct binary not found")
+            return None
+        stdout, stderr = await process.communicate()
+        if process.returncode:
```
berland, line 427: `sacct` on my ad-hoc slurm cluster always returns exit code 1 and "Slurm accounting storage is disabled", it seems to be correctly handled here 👍🏻

src/ert/scheduler/slurm_driver.py
verveerpj: I agree, I will turn some errors into warnings.

berland (approved):
Very good! You may want to squash commit 5 into one of the others?

verveerpj:
> Very good! You may want to squash commit 5 into one of the others?

Thanks for the review! Yes, I will squash all into the first.

PR #9446, opened 2024-12-04
Fix ForwardModelStep `handle_process_timeout...` timeout

jonathan-eq:
This fixes the bug where some code was unreachable after the refactoring in commit 4dc894ca63687476e091f582df5a42045190f7bd

**Approach**
🔧 

## When applicable

### DIFF
--- a/src/_ert/forward_model_runner/forward_model_step.py
+++ b/src/_ert/forward_model_runner/forward_model_step.py
@@ -175,7 +175,7 @@ class ForwardModelStep:
 
         target_file = self.job_data.get("target_file")
         target_file_mtime: Optional[int] = _get_target_file_ntime(target_file)
-
+        run_start_time = dt.now()
         try:
             proc = Popen(
                 arg_list,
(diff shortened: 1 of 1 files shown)


src/_ert/forward_model_runner/forward_model_step.py
```diff
@@ -217,7 +217,9 @@ def _run(self) -> Generator[Start | Exited | Running | None]:
                 exit_code = process.wait(timeout=self.MEMORY_POLL_PERIOD)
             except TimeoutExpired:
                 potential_exited_msg = (
-                    self.handle_process_timeout_and_create_exited_msg(exit_code, proc)
+                    self.handle_process_timeout_and_create_exited_msg(
```
xjules, line 220: I need some explanation what this function actually does?

jonathan-eq: It handles how we should react on a timeout while polling the process status. If it returns Exited, we yield it and return, otherwise we try again. 
It should fail if the running_time goes above what is configured for the forward_model, but we currently reset the running time every time we timeout, so it never reaches this code path. 
The timeout function should kill the entire process group if it is different than its own process group, then return an exited message with an error attached.

xjules (approved):
Very good! :rocket:

PR #9460, opened 2024-12-05
Fix a bug with skipping two lines after TITLE

eivindjahren:
**Issue**

The behavior of skipping two non-blank lines after TITLE was inherited from the code in resdata. After investigating how this is actually implemented in reservoir simulators, it turns out that one line, regardless of whether it is blank should be skipped. It seems that a blank title might actually cause a crash regardless of what comes after.

## When applicable

### DIFF
--- a/src/ert/config/_get_num_cpu.py
+++ b/src/ert/config/_get_num_cpu.py
@@ -90,9 +90,9 @@ def _get_num_cpu(
 ) -> Optional[int]:
     """Handles reading the lines in the data file and returns the num_cpu
 
-    TITLE keyword requires skipping two non-empty lines
+    TITLE keyword requires skipping one line
 
-    >>> _get_num_cpu(iter(["TITLE", "", "", "PARALLEL", "3 / -- skipped", "PARALLEL", "4 /"]))
+    >>> _get_num_cpu(iter(["TITLE", "PARALLEL EXAMPLE CASE", "PARALLEL", "4 /"]))
     4
 
     PARALLEL takes presedence even when SLAVES comes first:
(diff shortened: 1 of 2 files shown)


src/ert/config/_get_num_cpu.py
```diff
@@ -110,14 +110,10 @@ def _get_num_cpu(
...
-                # Skip two non-blank lines following a TITLE
-                for _ in range(2):
-                    line: list[str] = []
-                    while line == []:
-                        nline = parser.next_line(None)
-                        if nline is None:
-                            break
-                        line = list(nline)
+                # Skip one line following a TITLE
+                for _ in range(1):
```
berland, line 114: Can't this line with `for` just be dropped?

tests/ert/unit_tests/config/test_num_cpu.py
```diff
@@ -43,18 +42,23 @@ def test_num_cpu_from_data_file_used_if_config_num_cpu_not_set(parallelsuffix):
     with open(file=data_file, mode="w", encoding="utf-8") as data_file_hander:
         data_file_hander.write(
             f"""
+RUNSPEC
+--comment
+TITLE
+-- A Test Case --
```
berland, line 48: would suggest not to use commenting symbols in this example line

berland: we should possibly parametrize the title to "", "\n", "-- foo", "foo" etc.

src/ert/config/_get_num_cpu.py
eivindjahren: fixed

tests/ert/unit_tests/config/test_num_cpu.py
```diff
@@ -22,39 +20,50 @@ def test_num_cpu_from_config_preferred():
...
+        ).preferred_num_cpu
+        == config_num_cpu
+    )
 
 
 @pytest.mark.parametrize(
     "parallelsuffix", [("/"), (" /"), (" DISTRIBUTED/"), (" DISTRIBUTED /")]
 )
+@pytest.mark.parametrize(
+    "casetitle", ["CASE", "-- A CASE --", "PARALLEL Tutorial Case"]
```
berland, line 39: Including PARALLEL here is a nice touch! 👍🏻

berland: Out code probably allows an empty line also, should we include it in the tests but comment that ECLIPSE itself actually crashes on it?

tests/ert/unit_tests/config/test_num_cpu.py
eivindjahren: Added a parametrized list of titles as per our conversation.

berland (approved):
yay!

PR #9108, opened 2024-10-31
Replace interface towards eclrun and flow using fmstep_config

berland:
**Issue**
Resolves #8925

- [x] PR title captures the intent of the changes, and is fitting for release notes.
- [x] Added appropriate release note label
- [x] Commit history is consistent and clean, in line with the [contribution guidelines](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md).
- [ ] Make sure unit tests pass locally after every commit (`git rebase -i main
      --exec 'pytest tests/ert/unit_tests -n logical -m "not integration_test"'`)

## When applicable
- [ ] **When there are user facing changes**: Updated documentation
- [ ] **New behavior or changes to existing untested code**: Ensured that unit tests are added (See [Ground Rules](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md#ground-rules)).
- [ ] **Large PR**: Prepare changes in small commits for more convenient review
- [ ] **Bug fix**: Add regression test for the bug
- [ ] **Bug fix**: Create Backport PR to latest release

PR #9479, opened 2024-12-06
Propagate SETENV to forward model validation

### LINKED ISSUE #9474: Eclipse version validation blocks custom ECLX00_SITE_CONFIG
It is supposed to be possible use a custom site-config for Eclipse by adding

`SETENV ECL300_SITE_CONFIG <CONFIG_PATH>/testgrid.yml
`

to the Ert config. However, the version validation code does not honor this and will validate the version towards the system installed site config and then fail if you request a version not installed by default.

https://github.com/equinor/ert/blob/4b34fc94ceab52c8ad34e5059837b7ec26e5dcb9/src/ert/plugins/hook_implementations/forward_model_steps.py#L281-L293

berland:
The settings implied through SETENV can affect how forward model behave, and may also change how validation should be performed.

This is in particular true for the Eclipse forward model steps which try to validate the requested version. If a different configuration of Eclipse is set through

  SETENV ECL100_SITE_CONFIG somefile.yml

(or for ECL300_SITE_CONFIG) this information must be conveyed to the validation code.

**Issue**
Resolves #9474 

## When applicable

### DIFF
--- a/src/ert/plugins/hook_implementations/forward_model_steps.py
+++ b/src/ert/plugins/hook_implementations/forward_model_steps.py
@@ -3,7 +3,7 @@ import shutil
 import subprocess
 from pathlib import Path
 from textwrap import dedent
-from typing import List, Literal, Optional, Type
+from typing import Dict, List, Literal, Optional, Type
 
 import yaml
 
@@ -14,7 +14,6 @@ from ert import (
     ForwardModelStepValidationError,
     plugin,
 )
-from ert.plugins import ErtPluginManager
 
 
 class CarefulCopyFile(ForwardModelStepPlugin):
@@ -220,13 +219,17 @@ class Eclipse100(ForwardModelStepPlugin):
             default_mapping={"<NUM_CPU>": 1, "<OPTS>": ""},
         )
 
-    def validate_pre_experiment(self, _: ForwardModelStepJSON) -> None:
+    def validate_pre_experiment(
+        self, _: ForwardModelStepJSON, env_vars: Dict[str, str]
+    ) -> None:
         if "<VERSION>" not in self.private_args:
             raise ForwardModelStepValidationError(
                 "Forward model step ECLIPSE100 must be given a VERSION argument"
             )
         version = self.private_args["<VERSION>"]
-        available_versions = _available_eclrun_versions(simulator="eclipse")
+        available_versions = _available_eclrun_versions(
+            simulator="eclipse", env_vars=env_vars
+        )
 
         if available_versions and version not in available_versions:
             raise ForwardModelStepValidationError(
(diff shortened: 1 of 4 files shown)


tests/ert/unit_tests/config/test_forward_model.py
```diff
@@ -1011,7 +1035,9 @@ class FMWithAssertionError(ForwardModelStepPlugin):
         def __init__(self):
             super().__init__(name="FMWithAssertionError", command=["the_executable.sh"])
 
-        def validate_pre_experiment(self, fm_step_json: ForwardModelStepJSON) -> None:
+        def validate_pre_experiment(
+            self, fm_step_json: ForwardModelStepJSON, _: dict
```
andreas-el, line 1039: Very nitpick, but could this be `Dict[str, str]` instead of `dict` ?

tests/ert/unit_tests/config/test_forward_model.py
```diff
@@ -1021,7 +1047,9 @@ def __init__(self):
                 command=["the_executable.sh"],
             )
 
-        def validate_pre_experiment(self, fm_step_json: ForwardModelStepJSON) -> None:
+        def validate_pre_experiment(
+            self, fm_step_json: ForwardModelStepJSON, _: dict
```
andreas-el, line 1051: Very nitpick, but could this be Dict[str, str] instead of dict ?

tests/ert/unit_tests/config/test_forward_model.py
berland: fixed.

tests/ert/unit_tests/config/test_forward_model.py
berland: fixed.

sondreso:
Do we want to add this complexity given that we now have https://github.com/equinor/ert/pull/9050 and that ecl config should come through that system instead? :think

berland:
> Do we want to add this complexity given that we now have #9050 and that ecl config should come through that system instead? 🤔

Only if we want to have it available for Ert 11.1 and maybe also Ert 12.0. A similar solution is being implemented in #9108

src/ert/plugins/hook_implementations/forward_model_steps.py
```diff
@@ -278,13 +281,17 @@ def __init__(self) -> None:
             default_mapping={"<NUM_CPU>": 1, "<OPTS>": "", "<VERSION>": "version"},
         )
 
-    def validate_pre_experiment(self, _: ForwardModelStepJSON) -> None:
+    def validate_pre_experiment(
+        self, _: ForwardModelStepJSON, env_vars: Dict[str, str]
```
eivindjahren, line 285: It's a bit unfortunate to break the signature of this function as it is public and used by dependents. I think it would be better to make env_vars part of `ForwardModelStepJSON` and be a non-required field. That way we do not break any dependents.

eivindjahren (changes requested):
I think we should consider adding `env_vars` to `ForwardModelStepJSON` instead in order to not break any dependents.

src/ert/plugins/hook_implementations/forward_model_steps.py
berland: Absolutely agree, this has already been done in #9108

berland:
Closing this, will not spend time shaping this up for a backport unless needed. Will be fixed in #9108.

PR #9495, opened 2024-12-10
Use cached_example in test_everexport.py

yngve-sk:


tests/everest/entry_points/test_everexport.py
```diff
@@ -63,35 +62,32 @@ def validate_export_mock(**_):
...
-def cache_dir(request, monkeypatch):
-    return create_cached_mocked_test_case(request, monkeypatch)
-
-
 @patch("everest.bin.everexport_script.export_with_progress", side_effect=export_mock)
-def test_everexport_entry_run(mocked_func, copy_math_func_test_data_to_tmp):
+def test_everexport_entry_run(_, cached_example):
     """Test running everexport with not flags"""
     # NOTE: there is probably a bug concerning output folders. Everexport
     # seems to assume that the folder where the file will be saved exists.
```
DanSava, line 69: I think this is no longer the case and can be removed

DanSava (approved):
Looking good :man_dancing:

PR #9572, opened 2024-12-17
Solve ruff literal-membership

berland:
`ruff check --fix --unsafe-fixes
`
# literal-membership (PLR6201)

Fix is always available.

## What it does
Checks for membership tests on `list` and `tuple` literals.

## Why is this bad?
When testing for membership in a static sequence, prefer a `set` literal
over a `list` or `tuple`, as Python optimizes `set` membership tests.

## Example
```python
1 in [1, 2, 3]
```

Use instead:
```python
1 in {1, 2, 3}
```

## Fix safety
This rule's fix is marked as unsafe, as the use of a `set` literal will
error at runtime if the sequence contains unhashable elements (like lists
or dictionaries). While Ruff will attempt to infer the hashability of the
elements, it may not always be able to do so.

## References
- [What’s New In Python 3.2](https://docs.python.org/3/whatsnew/3.2.html#optimizations)

## When applicable

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -204,7 +204,6 @@ ignore = [
     "PLR0912", # too-many-branches
     "PLR0911", # too-many-return-statements
     "PLC2701", # import-private-name
-    "PLR6201", # literal-membership
     "PLR0914", # too-many-locals
     "PLR6301", # no-self-use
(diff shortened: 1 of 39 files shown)


eivindjahren (approved):
:memo:

PR #9586, opened 2024-12-18
Use PySide6

JHolba:
Upgrade from qtpy with pyqt5 to PySide6.
Using PySide6 directly should give us better typing.

## When applicable

### DIFF
--- a/src/ert/gui/ertwidgets/searchbox.py
+++ b/src/ert/gui/ertwidgets/searchbox.py
@@ -1,8 +1,8 @@
 from typing import Any
 
-from qtpy.QtCore import Qt, Signal
-from qtpy.QtGui import QColor, QFocusEvent, QKeyEvent
-from qtpy.QtWidgets import QLineEdit
+from PySide6.QtCore import Qt, Signal
+from PySide6.QtGui import QColor, QFocusEvent, QKeyEvent
+from PySide6.QtWidgets import QLineEdit
 
 
 class SearchBox(QLineEdit):
@@ -54,17 +54,17 @@ class SearchBox(QLineEdit):
         if not self.text():
             self.presentSearch()
 
-    def focusInEvent(self, a0: QFocusEvent | None) -> None:
-        QLineEdit.focusInEvent(self, a0)
+    def focusInEvent(self, arg__1: QFocusEvent) -> None:
+        QLineEdit.focusInEvent(self, arg__1)
         self.enterSearch()
 
-    def focusOutEvent(self, a0: QFocusEvent | None) -> None:
-        QLineEdit.focusOutEvent(self, a0)
+    def focusOutEvent(self, arg__1: QFocusEvent) -> None:
+        QLineEdit.focusOutEvent(self, arg__1)
         self.exitSearch()
 
-    def keyPressEvent(self, a0: QKeyEvent | None) -> None:
-        if a0 is not None and a0.key() == Qt.Key.Key_Escape:
+    def keyPressEvent(self, arg__1: QKeyEvent) -> None:
+        if arg__1.key() == Qt.Key.Key_Escape:
             self.clear()
             self.clearFocus()
         else:
-            QLineEdit.keyPressEvent(self, a0)
+            QLineEdit.keyPressEvent(self, arg__1)
--- a/src/ert/gui/model/real_list.py
+++ b/src/ert/gui/model/real_list.py
@@ -1,10 +1,11 @@
 from typing import overload
 
-from qtpy.QtCore import (
+from PySide6.QtCore import (
     QAbstractItemModel,
     QAbstractProxyModel,
     QModelIndex,
     QObject,
+    QPersistentModelIndex,
     Slot,
 )
 from typing_extensions import override
@@ -61,10 +62,14 @@ class RealListModel(QAbstractProxyModel):
         self.endResetModel()
 
     @override
-    def columnCount(self, parent: QModelIndex | None = None) -> int:
+    def columnCount(
+        self, parent: QModelIndex | QPersistentModelIndex | None = None
+    ) -> int:
         return 1
 
-    def rowCount(self, parent: QModelIndex | None = None) -> int:
+    def rowCount(
+        self, parent: QModelIndex | QPersistentModelIndex | None = None
+    ) -> int:
         parent = parent if parent else QModelIndex()
         if not parent.isValid():
             source_model = self.sourceModel()
@@ -75,16 +80,21 @@ class RealListModel(QAbstractProxyModel):
         return 0
 
     @overload
-    def parent(self, child: QModelIndex) -> QModelIndex: ...
+    def parent(self) -> QObject: ...
     @overload
-    def parent(self) -> QObject | None: ...
+    def parent(self, child: QModelIndex | QPersistentModelIndex) -> QModelIndex: ...
     @override
-    def parent(self, child: QModelIndex | None = None) -> QObject | None:
+    def parent(
+        self, child: QModelIndex | QPersistentModelIndex | None = None
+    ) -> QObject | QModelIndex:
         return QModelIndex()
 
     @override
     def index(
-        self, row: int, column: int, parent: QModelIndex | None = None
+        self,
+        row: int,
+        column: int,
+        parent: QModelIndex | QPersistentModelIndex | None = None,
     ) -> QModelIndex:
         parent = parent if parent else QModelIndex()
         if not parent.isValid():
@@ -93,7 +103,9 @@ class RealListModel(QAbstractProxyModel):
         return QModelIndex()
 
     @override
-    def hasChildren(self, parent: QModelIndex | None = None) -> bool:
+    def hasChildren(
+        self, parent: QModelIndex | QPersistentModelIndex | None = None
+    ) -> bool:
         # Reimplemented, since in the source model, the realizations have
         # children (i.e. valid indices.). Realizations do not have children in
         # this model.
@@ -105,7 +117,9 @@ class RealListModel(QAbstractProxyModel):
         return False
 
     @override
-    def mapToSource(self, proxyIndex: QModelIndex) -> QModelIndex:
+    def mapToSource(
+        self, proxyIndex: QModelIndex | QPersistentModelIndex
+    ) -> QModelIndex:
         if proxyIndex.isValid():
             sm = self.sourceModel()
             assert sm is not None
@@ -115,7 +129,9 @@ class RealListModel(QAbstractProxyModel):
         return QModelIndex()
 
     @override
-    def mapFromSource(self, sourceIndex: QModelIndex) -> QModelIndex:
+    def mapFromSource(
+        self, sourceIndex: QModelIndex | QPersistentModelIndex
+    ) -> QModelIndex:
         return (
             self.index(sourceIndex.row(), sourceIndex.column(), QModelIndex())
             if sourceIndex.isValid() and self._accept_index(sourceIndex)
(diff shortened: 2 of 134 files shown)


jonathan-eq:
It fails locally on MacOS with python3.12, but I got it working by modifying:
```
@pytest.fixture
def panel_with_localization_on(qtbot: QtBot):
    def func(settings, ensemble_size):
        widget = AnalysisModuleVariablesPanel(settings, ensemble_size)
        qtbot.addWidget(widget)
        widget.show()
        check_box = widget.findChild(QCheckBox, name="localization")
        qtbot.mouseClick(check_box, Qt.MouseButton.LeftButton)
        return settings, widget

    yield func
```

jonathan-eq:
When it failed it emitted:
```
---------------------------------------------------------------------------------------------------------------------------------------- Captured Qt messages ----------------------------------------------------------------------------------------------------------------------------------------
QtWarningMsg: Populating font family aliases took 77 ms. Replace uses of missing font family "Sans Serif" with one that exists to avoid this cost. 
QtWarningMsg: This plugin does not support propagateSizeHints()
```

JHolba:
> It fails locally on MacOS with python3.12, but I got it working by modifying:
> 
> ```
> @pytest.fixture
> def panel_with_localization_on(qtbot: QtBot):
>     def func(settings, ensemble_size):
>         widget = AnalysisModuleVariablesPanel(settings, ensemble_size)
>         qtbot.addWidget(widget)
>         widget.show()
>         check_box = widget.findChild(QCheckBox, name="localization")
>         qtbot.mouseClick(check_box, Qt.MouseButton.LeftButton)
>         return settings, widget
> 
>     yield func
> ```

Thank you for the help :)

src/ert/gui/ertwidgets/searchbox.py
```diff
@@ -54,17 +54,17 @@ def exitSearch(self) -> None:
...
-    def focusOutEvent(self, a0: QFocusEvent | None) -> None:
-        QLineEdit.focusOutEvent(self, a0)
+    def focusOutEvent(self, arg__1: QFocusEvent) -> None:
+        QLineEdit.focusOutEvent(self, arg__1)
         self.exitSearch()
 
-    def keyPressEvent(self, a0: QKeyEvent | None) -> None:
-        if a0 is not None and a0.key() == Qt.Key.Key_Escape:
+    def keyPressEvent(self, arg__1: QKeyEvent) -> None:
+        if arg__1 is not None and arg__1.key() == Qt.Key.Key_Escape:
```
jonathan-eq, line 66: Why are we checking if it is not None, if the parameter cannot be None?

src/ert/gui/model/real_list.py
```diff
@@ -75,16 +80,21 @@ def rowCount(self, parent: QModelIndex | None = None) -> int:
...
-    def parent(self, child: QModelIndex) -> QModelIndex: ...
+    def parent(self) -> QObject: ...
     @overload
-    def parent(self) -> QObject | None: ...
+    def parent(self, child: QModelIndex | QPersistentModelIndex) -> QModelIndex: ...
     @override
-    def parent(self, child: QModelIndex | None = None) -> QObject | None:
+    def parent(
+        self, child: QModelIndex | QPersistentModelIndex | None = None
+    ) -> QObject | QModelIndex:
```
jonathan-eq, line 89: Do we need `QObject` here?

src/ert/gui/model/snapshot.py
```diff
@@ -276,45 +283,55 @@ def _add_snapshot(self, snapshot: EnsembleSnapshot, iter_: str) -> None:
...
+    ) -> int:
         if parent is None:
             parent = QModelIndex()
         parent_item = self.root if not parent.isValid() else parent.internalPointer()
 
         if parent.column() > 0:
             return 0
 
-        return len(parent_item.children)
+        return len(parent_item.children)  # type: ignore
```
jonathan-eq, line 303: Why does this need `type: ignore`?

src/ert/gui/model/snapshot.py
```diff
@@ -276,45 +283,55 @@ def _add_snapshot(self, snapshot: EnsembleSnapshot, iter_: str) -> None:
...
-    def parent(self, child: QModelIndex) -> QModelIndex: ...
+    def parent(self) -> QObject: ...
     @overload
-    def parent(self) -> QObject | None: ...
+    def parent(self, child: QModelIndex | QPersistentModelIndex) -> QModelIndex: ...
     @override
-    def parent(self, child: QModelIndex | None = None) -> QObject | None:
+    def parent(
+        self, child: QModelIndex | QPersistentModelIndex | None = None
+    ) -> QObject | QModelIndex:
```
jonathan-eq, line 312: Does this also need QObject?
For QModelIndex, is it possible to have the typing include specifically what the index is pointing to?

src/ert/gui/model/snapshot.py
```diff
@@ -276,45 +283,55 @@ def _add_snapshot(self, snapshot: EnsembleSnapshot, iter_: str) -> None:
...
         if child is None or not child.isValid():
             return QModelIndex()
 
-        parent_item = child.internalPointer().parent
+        parent_item = child.internalPointer().parent  # type:ignore
```
jonathan-eq, line 316: Another type:ignore. Maybe we dont need it?

src/ert/gui/model/snapshot.py
```diff
@@ -276,45 +283,55 @@ def _add_snapshot(self, snapshot: EnsembleSnapshot, iter_: str) -> None:
...
 
         return self.createIndex(parent_item.row(), 0, parent_item)
 
     @override
-    def data(self, index: QModelIndex, role: int = Qt.ItemDataRole.DisplayRole) -> Any:
+    def data(
+        self,
+        index: QModelIndex | QPersistentModelIndex,
+        role: int = Qt.ItemDataRole.DisplayRole,
+    ) -> Any:
```
jonathan-eq, line 327: Should we use Any?

src/ert/gui/model/snapshot.py
```diff
@@ -356,10 +373,12 @@ def data(self, index: QModelIndex, role: int = Qt.ItemDataRole.DisplayRole) -> A
...
             return QColor()
 
-        return QVariant()
+        return None
 
     @staticmethod
-    def _real_data(_: QModelIndex, node: RealNode, role: int) -> Any:
+    def _real_data(
+        _: QModelIndex | QPersistentModelIndex, node: RealNode, role: int
+    ) -> Any:
```
jonathan-eq, line 381: Another Any

src/ert/gui/model/snapshot.py
```diff
@@ -445,7 +464,7 @@ def _fm_step_data(
         if role == FileRole:
             data_name = FM_STEP_COLUMNS[index.column()]
             if data_name in {ids.STDOUT, ids.STDERR}:
-                return node.data.get(data_name, QVariant())
+                return node.data.get(data_name, None)
```
jonathan-eq, line 467: Is `None` redundant here?

src/ert/gui/simulation/run_dialog.py
```diff
@@ -307,17 +306,18 @@ def on_snapshot_new_iteration(
     ) -> None:
         if not parent.isValid():
             index = self._snapshot_model.index(start, 0, parent)
+            iteration = cast(Node, index.internalPointer()).id_
```
jonathan-eq, line 309: Should we cast it to Node or IterationNode? I think the point of having Node prefixed with underscore was that it would not be imported elsewhere

src/ert/gui/simulation/view/progress_widget.py
```diff
@@ -2,22 +2,21 @@
...
-    QWidget,
 )
 
 from ert.ensemble_evaluator.state import ENSEMBLE_STATE_FAILED, REAL_STATE_TO_COLOR
 
 
 class ProgressWidget(QFrame):
     def __init__(self) -> None:
-        QWidget.__init__(self)
+        QFrame.__init__(self)
```
jonathan-eq, line 19: Should we change this to `super().__init__()?

src/ert/gui/simulation/view/progress_widget.py
```diff
@@ -110,5 +109,5 @@ def update_progress(self, status: dict[str, int], realization_count: int) -> Non
             self.stop_waiting_progress_bar()
         self.repaint_components()
 
-    def resizeEvent(self, a0: Any, event: Any = None) -> None:
+    def resizeEvent(self, event: Any = None) -> None:
```
jonathan-eq, line 112: Is there a more specific event type we can use?

src/ert/gui/simulation/view/update.py
```diff
@@ -127,7 +127,7 @@ def iteration(self) -> int:
     def _insert_status_message(self, message: str) -> None:
         item = QListWidgetItem()
         item.setText(message)
-        item.setFlags(item.flags() & ~Qt.ItemFlags(Qt.ItemFlag.ItemIsEnabled))
+        item.setFlags(item.flags() & ~Qt.ItemFlag.ItemIsEnabled)
```
jonathan-eq, line 130: Can we use `item.setFlag` instead?

src/ert/gui/suggestor/suggestor.py
```diff
@@ -131,6 +134,7 @@ def __init__(
         self.setStyleSheet(f"background-color: {LIGHT_GREY}; color: black")
         self.__layout.setContentsMargins(32, 47, 32, 16)
         self.__layout.setSpacing(32)
+        self.notifier = notifier
```
jonathan-eq, line 137: Do we need access to the notifier?

jonathan-eq (changes requested):
Very good! I have some questions, see comments :)

src/ert/gui/model/snapshot.py
JHolba: It is an override. It comes from QAbstractItemModel

src/ert/gui/model/snapshot.py
JHolba: it is only used by the data() method which returns any

src/ert/gui/model/snapshot.py
JHolba: both of the parent overloads come from inherited methods. the override must handle them both

src/ert/gui/model/snapshot.py
JHolba: ids.STDOUT and ids.STDERR are both optional in the typeddict. we need the None if they are not defined

src/ert/gui/simulation/view/update.py
JHolba: there is no setFlag method on QListWidgetItem

src/ert/gui/simulation/view/progress_widget.py
JHolba: there is. added it

src/ert/gui/simulation/view/progress_widget.py
JHolba: we should :)

src/ert/gui/model/snapshot.py
JHolba: added a cast to remove the ignore

src/ert/gui/model/snapshot.py
JHolba: added a cast to remove the ignore

src/ert/gui/simulation/run_dialog.py
JHolba: changed to IterationNode. More specific is always better

src/ert/gui/model/snapshot.py
jonathan-eq: Yes, but doesn't the get method return None even without the 2nd parameter?

JHolba: good point :)

src/ert/gui/suggestor/suggestor.py
JHolba: deleted it

JHolba:
going to close this due to segfaults in tests. will use PyQt6 instead #9860.

PR #9589, opened 2024-12-18
Improve handling of invalid experiments in GUI

### LINKED ISSUE #9493: Missing responses.json when opening ert

The following exception was found in the logs: 

```
File "/prog/res/komodo/2024.11.11-py311-rhel8/root/lib64/python3.11/site-packages/ert/__main__.py", line 690, in main
args.func(args, context.plugin_manager)
File "/prog/res/komodo/2024.11.11-py311-rhel8/root/lib64/python3.11/site-packages/ert/__main__.py", line 195, in run_gui_wrapper
run_gui(args, ert_plugin_manager)
File "/prog/res/komodo/2024.11.11-py311-rhel8/root/lib64/python3.11/site-packages/ert/gui/main.py", line 55, in run_gui
window, ens_path = _start_initial_gui_window(args, log_handler, plugin_manager)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/prog/res/komodo/2024.11.11-py311-rhel8/root/lib64/python3.11/site-packages/ert/gui/main.py", line 128, in _start_initial_gui_window
_main_window = _setup_main_window(
^^^^^^^^^^^^^^^^^^^
File "/prog/res/komodo/2024.11.11-py311-rhel8/root/lib64/python3.11/site-packages/ert/gui/main.py", line 169, in _setup_main_window
window.post_init()
File "/prog/res/komodo/2024.11.11-py311-rhel8/root/lib64/python3.11/site-packages/ert/gui/main_window.py", line 231, in post_init
experiment_panel = ExperimentPanel(
^^^^^^^^^^^^^^^^
File "/prog/res/komodo/2024.11.11-py311-rhel8/root/lib64/python3.11/site-packages/ert/gui/simulation/experiment_panel.py", line 152, in __init__
EvaluateEnsemblePanel(ensemble_size, run_path, notifier),
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/prog/res/komodo/2024.11.11-py311-rhel8/root/lib64/python3.11/site-packages/ert/gui/simulation/evaluate_ensemble_panel.py", line 54, in __init__
self._realizations_from_fs()
File "/prog/res/komodo/2024.11.11-py311-rhel8/root/lib64/python3.11/site-packages/ert/gui/simulation/evaluate_ensemble_panel.py", line 88, in _realizations_from_fs
missing_responses = ~ensemble.get_realization_mask_with_responses()
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/prog/res/komodo/2024.11.11-py311-rhel8/root/lib64/python3.11/site-packages/ert/storage/local_ensemble.py", line 251, in get_realization_mask_with_responses
[
File "/prog/res/komodo/2024.11.11-py311-rhel8/root/lib64/python3.11/site-packages/ert/storage/local_ensemble.py", line 252, in <listcomp>
self._responses_exist_for_realization(i, key)
File "/prog/res/komodo/2024.11.11-py311-rhel8/root/lib64/python3.11/site-packages/ert/storage/local_ensemble.py", line 300, in _responses_exist_for_realization
if not self.experiment.response_configuration:
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/prog/res/komodo/2024.11.11-py311-rhel8/root/lib64/python3.11/site-packages/ert/storage/local_experiment.py", line 300, in response_configuration
^^^^^^^^^^^^^^^^^^
File "/prog/res/komodo/2024.11.11-py311-rhel8/root/lib64/python3.11/site-packages/ert/storage/local_experiment.py", line 263, in response_info
raise ValueError(f"{self._responses_file!s} does not exist")
ValueError: responses.json does not exist
```

Link to trace: https://fmu-dashboard-fdeqekb8chekcqhu.neu.grafana.azure.com/d/ae0bpm0169ou8e/traces?from=now-30d&to=now&var-operation_id=352106f845ac2bdc5ffc781da018384c

Not crash

### steps to reproduce

```Python
TBD
```

- [x] python 3.11
- [ ] python 3.12
- [ ] macosx
- [ ] rhel7
- [x] rhel8
- [ ] local queue
- [ ] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

jonathan-eq:
**Issue**
Resolves #9493

**Approach**
The commit in this PR:
* Makes storage re-validate when changing between ert modes (manage experiment, run experiment, plot experiment), and emit storage_changed if validity has changed.
* Disables ensembles with invalid experiments from ensemble_selectors (error is shown as tooltip on hover)
* Filters out invalid experiments from dark storage, so plotter won't attempt to plot it. 
* Reloads storage and re-validates on end of experiment, so ert wont crash if responses.json is deleted mid-run.
 
image
(This is when the responses.json file is deleted mid run. Prior to this PR, ert would crash with `ValueError: responses.json does not exist`)

image
(This is when responses.json is deleted, and we open the plotter. The ensemble exists, but is not given as a option to plot due to it having an invalid experiment)

image
(This is when responses.json is deleted and we open manage-experiment. The experiment is shown, but cannot be selected. It is greyed out, and gives error as a tooltip when hovered)

## When applicable

### DIFF
--- a/src/ert/gui/ertnotifier.py
+++ b/src/ert/gui/ertnotifier.py
@@ -65,3 +65,16 @@ class ErtNotifier(QObject):
     @Slot(bool)
     def set_is_simulation_running(self, is_running: bool) -> None:
         self._is_simulation_running = is_running
+        self.refresh()
+
+    def refresh(self) -> None:
+        if self._storage is None:
+            return
+        self._storage.refresh()
+        self.storage_changed.emit(self._storage)
+
+    def revalidate_storage(self) -> None:
+        if self._storage is None:
+            return
+        if self._storage.check_if_experiment_validity_changed():
+            self.refresh()
--- a/src/ert/gui/ertwidgets/ensembleselector.py
+++ b/src/ert/gui/ertwidgets/ensembleselector.py
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
 from collections.abc import Iterable
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, cast
 
 from PyQt6.QtCore import Qt
 from PyQt6.QtCore import pyqtSignal as Signal
+from PyQt6.QtGui import QStandardItemModel
 from PyQt6.QtWidgets import QComboBox
 
 from ert.gui.ertnotifier import ErtNotifier
@@ -23,6 +24,7 @@ class EnsembleSelector(QComboBox):
         update_ert: bool = True,
         show_only_undefined: bool = False,
         show_only_no_children: bool = False,
+        show_only_with_valid_experiment: bool = False,
     ):
         super().__init__()
         self.notifier = notifier
@@ -37,6 +39,7 @@ class EnsembleSelector(QComboBox):
         # if the ensemble has not been used in an update, as that would
         # invalidate the result
         self._show_only_no_children = show_only_no_children
+        self._show_only_with_valid_experiment = show_only_with_valid_experiment
         self.setSizeAdjustPolicy(QComboBox.SizeAdjustPolicy.AdjustToContents)
 
         self.setEnabled(False)
@@ -69,9 +72,26 @@ class EnsembleSelector(QComboBox):
             self.setEnabled(True)
 
         for ensemble in self._ensemble_list():
+            model = cast(QStandardItemModel, self.model())
             self.addItem(
                 f"{ensemble.experiment.name} : {ensemble.name}", userData=ensemble
             )
+            if (
+                self._show_only_with_valid_experiment
+                and not ensemble.experiment.is_valid()
+            ):
+                index = self.count() - 1
+                model_item = model.item(index)
+                assert model_item is not None
+                new_flags = (
+                    model_item.flags()
+                    & ~Qt.ItemFlag.ItemIsEnabled
+                    & ~Qt.ItemFlag.ItemIsSelectable
+                )
+                model_item.setFlags(new_flags)
+                self.setItemData(
+                    index, "This ensemble is invalid", Qt.ItemDataRole.ToolTipRole
+                )
 
         current_index = self.findData(
             self.notifier.current_ensemble, Qt.ItemDataRole.UserRole
@@ -95,7 +115,10 @@ class EnsembleSelector(QComboBox):
             )
         else:
             ensembles = self.notifier.storage.ensembles
-        ensemble_list = list(ensembles)
+        if self._show_only_with_valid_experiment:
+            ensemble_list = [ens for ens in ensembles if ens.experiment.is_valid()]
+        else:
+            ensemble_list = list(ensembles)
         if self._show_only_no_children:
             parents = [
                 ens.parent for ens in self.notifier.storage.ensembles if ens.parent
--- a/src/ert/gui/main_window.py
+++ b/src/ert/gui/main_window.py
@@ -158,6 +158,7 @@ class ErtMainWindow(QMainWindow):
     def select_central_widget(self) -> None:
         actor = self.sender()
         if actor:
+            self.notifier.revalidate_storage()
             index_name = actor.property("index")
 
             for widget in self.central_panels_map.values():
--- a/src/ert/gui/tools/manage_experiments/manage_experiments_panel.py
+++ b/src/ert/gui/tools/manage_experiments/manage_experiments_panel.py
@@ -75,7 +75,11 @@ class ManageExperimentsPanel(QTabWidget):
 
         ensemble_layout = QHBoxLayout()
         ensemble_label = QLabel("Target ensemble:")
-        ensemble_selector = EnsembleSelector(self.notifier, show_only_undefined=True)
+        ensemble_selector = EnsembleSelector(
+            self.notifier,
+            show_only_undefined=True,
+            show_only_with_valid_experiment=True,
+        )
         ensemble_selector.setMinimumWidth(300)
         ensemble_layout.addWidget(ensemble_label)
         ensemble_layout.addWidget(ensemble_selector)
--- a/src/ert/gui/tools/manage_experiments/storage_info_widget.py
+++ b/src/ert/gui/tools/manage_experiments/storage_info_widget.py
@@ -95,6 +95,8 @@ class _ExperimentWidget(QWidget):
 
     @Slot(Experiment)
     def setExperiment(self, experiment: Experiment) -> None:
+        if not experiment.is_valid():
+            return
         self._experiment = experiment
 
         self._name_label.setText(f"Name: {experiment.name!s}")
--- a/src/ert/gui/tools/manage_experiments/storage_model.py
+++ b/src/ert/gui/tools/manage_experiments/storage_model.py
@@ -26,7 +26,7 @@ _COLUMN_TEXT = {
 
 
 class RealizationModel:
-    def __init__(self, realization: int, parent: Any) -> None:
+    def __init__(self, realization: int, parent: "EnsembleModel") -> None:
         self._parent = parent
         self._name = f"Realization {realization}"
         self._ensemble_id = parent._id
@@ -58,7 +58,7 @@ class RealizationModel:
 
 
 class EnsembleModel:
-    def __init__(self, ensemble: Ensemble, parent: Any):
+    def __init__(self, ensemble: Ensemble, parent: "ExperimentModel"):
         self._parent = parent
         self._name = ensemble.name
         self._id = ensemble.id
@@ -76,7 +76,6 @@ class EnsembleModel:
     def data(self, index: QModelIndex, role: Qt.ItemDataRole) -> Any:
         if not index.isValid():
             return None
-
         col = index.column()
         if role == Qt.ItemDataRole.DisplayRole:
             if col == _Column.NAME:
@@ -86,15 +85,16 @@ class EnsembleModel:
         elif role == Qt.ItemDataRole.ToolTipRole:
             if col == _Column.TIME:
                 return str(self._start_time)
-
         return None
 
 
-class ExperimentModel:
-    def __init__(self, experiment: Experiment, parent: Any):
+class ExperimentModel(QAbstractItemModel):
+    def __init__(self, experiment: Experiment, parent: "StorageModel"):
         self._parent = parent
         self._id = experiment.id
         self._name = experiment.name
+        self._is_valid = experiment.is_valid()
+        self._error = experiment.error_message
         self._experiment_type = experiment.metadata.get("ensemble_type")
         self._children: list[EnsembleModel] = []
 
@@ -106,9 +106,7 @@ class ExperimentModel:
             return self._parent._children.index(self)
         return 0
 
-    def data(
-        self, index: QModelIndex, role: Qt.ItemDataRole = Qt.ItemDataRole.DisplayRole
-    ) -> Any:
+    def data(self, index: QModelIndex, role: int = Qt.ItemDataRole.DisplayRole) -> Any:
         if not index.isValid():
             return None
 
@@ -129,7 +127,9 @@ class ExperimentModel:
                 qapp = QApplication.instance()
                 assert isinstance(qapp, QApplication)
                 return qapp.palette().mid()
-
+        elif role == Qt.ItemDataRole.ToolTipRole:
+            if self._error:
+                return self._error
         return None
 
 
@@ -191,9 +191,8 @@ class StorageModel(QAbstractItemModel):
         child_item = cast(ChildModel, child.internalPointer())
         parentItem = child_item._parent
 
-        if parentItem == self:
+        if isinstance(parentItem, StorageModel):
             return QModelIndex()
-
         return self.createIndex(parentItem.row(), 0, parentItem)
 
     @override
@@ -216,6 +215,28 @@ class StorageModel(QAbstractItemModel):
         return cast(ChildModel | Self, index.internalPointer()).data(
             index, Qt.ItemDataRole(role)
         )
+
+    @override
+    def flags(self, index: QModelIndex) -> Qt.ItemFlag:
+        default_flags = super().flags(index)
+        if not index.isValid():
+            return default_flags
+        item = index.internalPointer()
+        if isinstance(item, ExperimentModel) and not item._is_valid:
+            return default_flags & ~Qt.ItemFlag.ItemIsEnabled
+        return default_flags
+
+    @override
+    def hasChildren(self, parent: QModelIndex | None = None) -> bool:
+        if parent is None or not parent.isValid():
+            return len(self._children) > 0
+
+        flags = self.flags(parent)
+        # hide children if disabled
+        if not (flags & Qt.ItemFlag.ItemIsEnabled):
+            return False
+
+        return super().hasChildren(parent)
 
     @override
     def index(
--- a/src/ert/storage/local_experiment.py
+++ b/src/ert/storage/local_experiment.py
@@ -75,6 +75,26 @@ class LocalExperiment(BaseMode):
         self._index = _Index.model_validate_json(
             (path / "index.json").read_text(encoding="utf-8")
         )
+        self._validate_files()
+
+    def _validate_files(self) -> None:
+        self.valid_parameters = (self._path / self._parameter_file).exists()
+        self.valid_responses = (self._path / self._responses_file).exists()
+        self.valid_metadata = (self._path / self._metadata_file).exists()
+
+    def is_valid(self) -> bool:
+        return self.valid_parameters and self.valid_responses and self.valid_metadata
+
+    @property
+    def error_message(self) -> str:
+        errors = []
+        if not self.valid_parameters:
+            errors.append("Parameter file is missing")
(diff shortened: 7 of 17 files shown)


src/ert/gui/tools/manage_experiments/storage_model.py
```diff
@@ -211,9 +221,30 @@ def headerData(
     def data(self, index: QModelIndex, role: int = Qt.ItemDataRole.DisplayRole) -> Any:
         if not index.isValid():
             return None
-
         return index.internalPointer().data(index, role)
 
+    @override
+    def flags(self, index: QModelIndex) -> Qt.ItemFlag:
```
andreas-el, line 227: I suspect this should return multiple flags instead (the c++ function does that)
`virtual Qt::ItemFlags flags(const QModelIndex &index) const`

src/ert/gui/tools/manage_experiments/storage_model.py
```diff
@@ -211,9 +221,30 @@ def headerData(
...
+        default_flags = super().flags(index)
+        if not index.isValid():
+            return default_flags
+        item = index.internalPointer()
+        if isinstance(item, ExperimentModel) and not item._is_valid:
+            return default_flags & ~Qt.ItemFlag.ItemIsEnabled
+        return default_flags
+
+    @override
+    def hasChildren(self, index):
```
andreas-el, line 237: `bool` return type?

src/ert/gui/tools/manage_experiments/storage_model.py
```diff
@@ -211,9 +221,30 @@ def headerData(
...
         return index.internalPointer().data(index, role)
 
+    @override
+    def flags(self, index: QModelIndex) -> Qt.ItemFlag:
+        default_flags = super().flags(index)
+        if not index.isValid():
+            return default_flags
+        item = index.internalPointer()
+        if isinstance(item, ExperimentModel) and not item._is_valid:
+            return default_flags & ~Qt.ItemFlag.ItemIsEnabled
```
andreas-el, line 233: Could the be resolved if you create a typed variable and apply the flag change?

src/ert/gui/ertwidgets/ensembleselector.py
```diff
@@ -68,9 +72,24 @@ def populate(self) -> None:
...
+            model = self.model()
             self.addItem(
                 f"{ensemble.experiment.name} : {ensemble.name}", userData=ensemble
             )
+            if (
+                self._show_only_with_valid_experiment
+                and not ensemble.experiment.is_valid()
+            ):
+                index = self.count() - 1
+                model_item = model.item(index)
```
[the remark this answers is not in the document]
jonathan-eq: Yes, it works well. Should I use ItemData instead?

src/ert/gui/tools/manage_experiments/storage_model.py
jonathan-eq: I had to wrap the itemflag in ItemFlags, and then mypy was happy :)

src/ert/gui/main_window.py
```diff
@@ -154,6 +154,7 @@ def right_clicked(self) -> None:
     def select_central_widget(self) -> None:
         actor = self.sender()
         if actor:
+            self.notifier.refresh()
```
jonathan-eq, line 157: This refreshes storage to make sure the experiment files (responses, index, metadata, and parameters) are still valid.

jonathan-eq: I updated it so that if will only actually refresh when the validity has changed, when toggling between the panels. This also means that the storage won't be updated if new experiments are added and we toggle between the panels, but that is the same behavior as on main.

jonathan-eq:
This PR contains the initial work towards making this part of the storage more robust. There is still an issue where ert crashes if you have selected an experiment in the manage-experiment panel, tab out and delete responses.json, and tab back in. It seems like the SelectionModel for the QListView tries reselecting the experiment (which is now invalid) before it has been revalidated.

src/ert/gui/tools/manage_experiments/storage_info_widget.py
```diff
@@ -94,6 +94,8 @@ def __init__(self) -> None:
 
     @Slot(Experiment)
     def setExperiment(self, experiment: Experiment) -> None:
+        if not experiment.is_valid():
```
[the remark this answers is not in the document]
jonathan-eq: Yes, when the selected experiment in manage-experiment becomes invalid while still selected. Apparently, when you tab out of ert while an experiment is selected, and tab in again; the same experiment is re-selected, and this signal is emitted. Very fun to debug 🔮

src/ert/gui/tools/manage_experiments/storage_model.py
```diff
@@ -128,7 +129,9 @@ def data(
                 qapp = QApplication.instance()
                 assert isinstance(qapp, QApplication)
                 return qapp.palette().mid()
-
+        elif role == Qt.ItemDataRole.ToolTipRole:
+            if self._error:
```
xjules, line 133: Is is the same error shown twice?

src/ert/gui/tools/manage_experiments/storage_model.py
```diff
@@ -211,9 +214,31 @@ def headerData(
...
+        item = index.internalPointer()
+        if isinstance(item, ExperimentModel) and not item._is_valid:
+            new_flags = default_flags & ~Qt.ItemFlags(Qt.ItemFlag.ItemIsEnabled)
+            return Qt.ItemFlags(new_flags)
+        return default_flags
+
+    @override
+    def hasChildren(self, parent: QModelIndex | None = None) -> bool:
+        if parent is None or not parent.isValid():
+            return True
```
xjules, line 233: shouldn't this be False? At least it sounds like it, but it might be my lack of understanding.

andreas-el:
I will argue that you should alter the title of this PR. The PR does not specifically target the missing file, but rather handling invalid experiments.

tests/ert/ui_tests/gui/test_main_window.py
```diff
@@ -854,3 +856,84 @@ def wait_for_simulation_completed():
...
+                )
+        assert (
+            number_of_found_invalid_experiments
+            == expected_number_of_invalid_experiments
+        )
+        assert number_of_found_valid_experiments == expected_number_of_valid_experiments
+
+    def check_plot_tool(expected_number_of_cases: int) -> None:
+        find_and_click_button("button_Create_plot")
+        # Due to the fact that we create new instances of PlotWindow on tab change, QtBot is defaulting to the first child
```
andreas-el, line 918: If you refer to this;
```
            if index_name == "Create plot":
                if self._plot_window:
                    self._plot_window.close()
                self._plot_window = PlotWindow(self.config_file, self)
```
It should be possible to check if the PlotWindow is open or not. Also, the window should be accessible via;
```
gui._plot_window
```

jonathan-eq:
> I will argue that you should alter the title of this PR. The PR does not specifically target the missing file, but rather handling invalid experiments.

Yes, I was hit by scope creep...

src/ert/storage/local_experiment.py
```diff
@@ -75,6 +75,26 @@ def __init__(
...
         )
+        self._validate_files()
+
+    def _validate_files(self) -> None:
+        self.valid_parameters = (self._path / self._parameter_file).exists()
+        self.valid_responses = (self._path / self._responses_file).exists()
+        self.valid_metadata = (self._path / self._metadata_file).exists()
+
+    def is_valid(self) -> bool:
+        return self.valid_parameters and self.valid_responses and self.valid_metadata
```
andreas-el, line 86: Does it make sense to combine these, and check for existence of files in `is_valid` ?

src/ert/gui/ertnotifier.py
```diff
@@ -63,3 +63,16 @@ def set_current_ensemble(self, ensemble: Ensemble | None = None) -> None:
     @Slot(bool)
     def set_is_simulation_running(self, is_running: bool) -> None:
         self._is_simulation_running = is_running
+        self.refresh()
+
+    def refresh(self) -> None:
+        if self._storage is None:
+            return
+        self._storage.refresh()
+        self.storage_changed.emit(self._storage)
```
andreas-el, line 72: Should this emit regardless of outcome?

src/ert/storage/local_experiment.py
jonathan-eq: Then it becomes more difficult to see if the validity of a specific experiment has changed. I think that is why I chose to have it in two separate steps.

src/ert/gui/ertnotifier.py
jonathan-eq: We only call this method when we know it will be changed when refreshing. Either if the validity of an experiment has changed; or if an experiment has finished/started running

src/ert/gui/ertwidgets/ensembleselector.py
```diff
@@ -22,6 +23,7 @@ def __init__(
         update_ert: bool = True,
         show_only_undefined: bool = False,
         show_only_no_children: bool = False,
+        show_only_with_valid_experiment: bool = False,
```
andreas-el, line 26: Are these flags mutually exclusive? I.e. only one is ever set to true?
This could have been converted to an enum if so.

src/ert/gui/ertwidgets/ensembleselector.py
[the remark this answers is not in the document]
andreas-el: Not sure. I looked into this due to style/linting errors. If the error was resolved, this is probably fine.

src/ert/gui/tools/manage_experiments/manage_experiments_panel.py
```diff
@@ -75,7 +75,11 @@ def _add_initialize_from_scratch_tab(self) -> None:
 
         ensemble_layout = QHBoxLayout()
         ensemble_label = QLabel("Target ensemble:")
-        ensemble_selector = EnsembleSelector(self.notifier, show_only_undefined=True)
+        ensemble_selector = EnsembleSelector(
+            self.notifier,
+            show_only_undefined=True,
+            show_only_with_valid_experiment=True,
```
andreas-el, line 81: Show only valid **and** undefined ?

src/ert/gui/tools/manage_experiments/storage_model.py
andreas-el: A person/item who has no parents, or not any valid parents, are guaranteed to have children?
I'm so confused here.

src/ert/gui/ertwidgets/ensembleselector.py
jonathan-eq: No, they are two completely separate filters

src/ert/gui/tools/manage_experiments/manage_experiments_panel.py
jonathan-eq: They are separate, so it think they should have their own flags.

src/ert/gui/ertwidgets/ensembleselector.py
andreas-el: I think we should spend a couple of minutes looking at this to see if we can alter this. I think this filtering thing have grown outside reasonable scope of multiple booleans.

src/ert/gui/ertnotifier.py
andreas-el: But if you call the refresh function, this will emit regardless. I see that your statement is true when using revalidate_storage, due to the check there that looks at state changes.

src/ert/gui/tools/manage_experiments/storage_model.py
andreas-el: What if you have no experiments? Is this still correct? I think you need to provide some comment or explanation here, since we don't understand why this is so.

src/ert/gui/ertnotifier.py
jonathan-eq: Yes, but we only call it when we already know something has changed, this check is done upstream in both places where it's used 🤔

src/ert/gui/ertwidgets/ensembleselector.py
jonathan-eq: I think we need all those filters, because this component is re-used with a lot of different context

src/ert/gui/tools/manage_experiments/storage_model.py
jonathan-eq: I think hasChildren is there only to check if the tree can be expanded. In our case, the root node (StorageModel) will always be expanded, but the number of rows (ExperimentModels) will determine how much it will be expanded. I can change it to `return len(self._children) > 0`, but hardcoding it to true would probably be fine too.

xjules:
Not sure if it is related but this one hangs: https://github.com/equinor/ert/actions/runs/12924655010/job/36044056095?pr=9589

src/ert/gui/tools/manage_experiments/manage_experiments_panel.py
jonathan-eq: show_only_undefined is for the ensembles in the experiment; whilst the other flag is for the experiment itself.

src/ert/gui/ertwidgets/ensembleselector.py
```diff
@@ -69,9 +72,26 @@ def populate(self) -> None:
             self.setEnabled(True)
 
         for ensemble in self._ensemble_list():
+            model = cast(QStandardItemModel, self.model())
             self.addItem(
                 f"{ensemble.experiment.name} : {ensemble.name}", userData=ensemble
             )
+            if (
+                self._show_only_with_valid_experiment
```
andreas-el, line 80: Do you really need to check for `_show_only_with_valid_experiment` here?
Does it not make sense to just do this for all experiments that are not valid?

src/ert/gui/tools/manage_experiments/storage_model.py
andreas-el: Can you double check this? On first sight It does look like you're returning flags, and not flag?

src/ert/gui/ertwidgets/ensembleselector.py
```diff
@@ -95,7 +115,10 @@ def _ensemble_list(self) -> Iterable[Ensemble]:
             )
         else:
             ensembles = self.notifier.storage.ensembles
-        ensemble_list = list(ensembles)
+        if self._show_only_with_valid_experiment:
+            ensemble_list = [ens for ens in ensembles if ens.experiment.is_valid()]
```
eivindjahren, line 119: We probably want to show ensembles in the ensemble selector if it is invalid, but present it as disabled or bad in some way.

src/ert/storage/local_experiment.py
eivindjahren: Rather than relying on the holder of the object remembering to check whether what you have a valid experiment, it would probably better to have the type enforce this.

```python
def get_experiment(self) -> Experiment | InvalidExperiment:
    ...
```

That way, whenever you call `get_experiment()` you have to check whether you got an InvalidExperiment back:

```python
match storage.get_experiment(key):
    case InvalidExperiment():
      ...
    case Experiment():
      ...
```

In order to avoid having to check for validity changes whenever you get an Experiment, we should probably just open the files or otherwise hold their contents in memory already: https://stackoverflow.com/questions/31099347/what-happens-internally-when-deleting-an-opened-file-in-linux . That way there is no need for `check_if_experiment_validity_changed`.

jonathan-eq:
The PR fails often on runs with the error:
```
  File "/Users/JONAK/Documents/FMU/SCOUT/ert/src/ert/gui/simulation/experiment_panel.py", line 313, in run_experiment
    self._dialog.run_experiment()
  File "/Users/JONAK/Documents/FMU/SCOUT/ert/src/ert/gui/simulation/run_dialog.py", line 386, in run_experiment
    self._notifier.set_is_simulation_running(True)
  File "/Users/JONAK/Documents/FMU/SCOUT/ert/src/ert/gui/ertnotifier.py", line 68, in set_is_simulation_running
    self.refresh()
  File "/Users/JONAK/Documents/FMU/SCOUT/ert/src/ert/gui/ertnotifier.py", line 73, in refresh
    self._storage.refresh()
  File "/Users/JONAK/Documents/FMU/SCOUT/ert/src/ert/storage/local_storage.py", line 132, in refresh
    self._experiments = self._load_experiments()
                        ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/JONAK/Documents/FMU/SCOUT/ert/src/ert/storage/local_storage.py", line 243, in _load_experiments
    for ens in self._ensembles.values():
               ^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: dictionary changed size during iteration
[1]    4052 abort      ert gui poly.ert
```

eivindjahren:
This seems to mostly have to do with users wanting to selectively deleteing experiments and ensembles. We need to implement https://github.com/equinor/ert/issues/7043 first so we can assess the need for handling other invalid storage states.

Issue #9590, opened 2024-12-18
Warning of ill-conditioned matrix only showed in terminal


eivindjahren:
The following warning is shown for the update step sometimes:
```
iterative_ensemble_smoother/esmda_inversion.py:229: Lin
AlgWarning: Ill-conditioned matrix (rcond=1.00742e-17): result may not be accurate.
  T = sp.linalg.solve(C_DD, D - Y, **solver_kwargs)
```

It seems to be neither logged nor displayed in any way in the gui, although it may cause the resulting update to be inaccurate.

This situation may arise from differing magnitude of numbers between realizations. It might be that we want to implement outlier detection to reduce how often this happens.

Can be reproduced with the following setup: https://github.com/eivindjahren/ert/tree/update_err_msgs/test-data/ert/counter_examples/ill-conditioned-matrix-0

Issue #9585, opened 2024-12-18
uninformative error message when writing updated zero values of  logarithmic data


eivindjahren:
![Image](https://github.com/user-attachments/assets/f340bba1-b709-47a8-a951-b72625778243)

Can be reproduced with https://github.com/eivindjahren/ert/tree/update_err_msgs/test-data/ert/counter_examples/math_domain_error_0

traceback of error is:

```
Traceback (most recent call last):
  File "/.../ert/src/ert/gui/simulation/run_dialog.py", line 426, in _on_event
    @Slot(object)

  File "/.../ert/src/_ert/threading.py", line 85, in _handler
    raise current_exception
_ert.threading.ErtThreadError: math domain error in thread 'ert_gui_simulation_thread'
 Traceback (most recent call last):
  File "/.../ert/src/_ert/threading.py", line 57, in run
    super().run()
  File "/.../opentelemetry_.py", line 129, in __wrap_threading_run
    return call_wrapped(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/threading.py", line 1010, in run
    self._target(*self._args, **self._kwargs)
  File "/.../ert/src/ert/gui/simulation/run_dialog.py", line 354, in run
    self._run_model.start_simulations_thread(
  File "/usr/lib/python3.12/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/.../ert/src/ert/run_models/base_run_model.py", line 355, in start_simulati
    raise e
  File "/.../ert/src/ert/run_models/base_run_model.py", line 331, in start_simulati
    self.run_experiment(
  File "/usr/lib/python3.12/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/.../ert/src/ert/run_models/ensemble_smoother.py", line 106, in run_experim
    self._evaluate_and_postprocess(
  File "/.../ert/src/ert/run_models/base_run_model.py", line 693, in _evaluate_and_
    create_run_path(
  File "/.../ert/src/ert/enkf_main.py", line 263, in create_run_path
    _generate_parameter_files(
  File "/.../ert/src/ert/enkf_main.py", line 111, in _generate_parameter_files
    export_values = node.write_to_runpath(Path(run_path), iens, fs)
[...]

Issue #9581, opened 2024-12-18
Uninformative error message in update: "Matrix is singular."


eivindjahren:
![Image](https://github.com/user-attachments/assets/563f25df-2f22-4a17-ae1a-29d5167c09ca)

This behavior can be reproduced with this setup: https://github.com/eivindjahren/ert/tree/update_err_msgs/test-data/ert/counter_examples/matrix-is-singular-0

This is the traceback of the thrown error: 
```

Traceback (most recent call last):
  File "/../ert/src/ert/gui/simulation/run_dialog.py", line 426, in _on_event
    @Slot(object)

  File "/../ert/src/_ert/threading.py", line 85, in _handler
    raise current_exception
_ert.threading.ErtThreadError: Matrix is singular. in thread 'ert_gui_simulation_thread'
 Traceback (most recent call last):
  File "/../ert/src/_ert/threading.py", line 57, in run
    super().run()
  File "/.../opentelemetry/instrumentation/threading/__init__.py", line 129, in __wrap_threading_run
    return call_wrapped(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/threading.py", line 1010, in run
    self._target(*self._args, **self._kwargs)
  File "/../ert/src/ert/gui/simulation/run_dialog.py", line 354, in run
    self._run_model.start_simulations_thread(
  File "/usr/lib/python3.12/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/../ert/src/ert/run_models/base_run_model.py", line 355, in start_simulations_thread
    raise e
  File "/../ert/src/ert/run_models/base_run_model.py", line 331, in start_simulations_thread
    self.run_experiment(
  File "/usr/lib/python3.12/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/../ert/src/ert/run_models/ensemble_smoother.py", line 98, in run_experiment
    posterior = self.update(prior, ensemble_format % 1)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/../ert/src/ert/run_models/base_run_model.py", line 792, in update
    smoother_update(
  File "/../ert/src/ert/analysis/_es_update.py", line 798, in smoother_update
    raise e
[...]

PR #9625, opened 2024-12-27
Fix precision loss issue in calculation of standard deviation.

### LINKED ISSUE #9301: Failure in misfit preprocessor after warning

https://fmu-dashboard-fdeqekb8chekcqhu.neu.grafana.azure.com/d/ae0bpm0169ou8e/traces?from=now-24h&to=now&var-operation_id=9a8b920d1f656a017b3f4e6aa26e1925

User report where ERT failed in the update step due to an error in misfit preprocessor. The following warning was emitted:
```
/prog/res/komodo/2024.11.02-py311-rhel8/root/lib64/python3.11/site-packages/ert/analysis/misfit_preprocessor.py:65: ConstantInputWarning: An input array is constant; the correlation coefficient is not defined. correlation = spearmanr(responses).statistic.
```
and then ERT fails afterwards with:
```
error_msg='y must be 1 or 2 dimensional.
```

For ERT to either skip the auto scaling with a warning, or for ERT to error with a reasonable error message.

### steps to reproduce

Not sure how exactly it was produced by the user, but can "reproduce" it by giving a constant matrix as first input arg to `cluster_responses` in misfit_preprocessor.py, then scipy.stats.spearmanr will produce nans after emitting the warning above, and we do not check for this case.

See: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.spearmanr.html

- [ ] python 3.10
- [x] python 3.11
- [ ] python 3.12
- [ ] macosx
- [ ] rhel7
- [x] rhel8
- [ ] local queue
- [ ] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

eivindjahren:
**Issue**
Resolves #9301

**Approach**

The problem was due to catastrophic precision loss in the standard deviation calculation.
This PR changes the update function so that standard deviation is calculated in 64 bit float
precision. It still remains to validate that this does not use too much memory.

* In `src/ert/analysis/_es_update.py`: Changed the standard deviation calculation to 64-bit precision.
* In `tests/ert/ui_tests/cli/test_update.py`: Added a property test that was used to find the bug, and several others: #9590, #9585, #9581,  #9523, & #9520 .
* In `tests/ert/unit_tests/analysis/test_es_update.py`: Added a new test to handle precision loss in standard deviation calculation.

## When applicable

### DIFF
--- a/src/ert/analysis/_es_update.py
+++ b/src/ert/analysis/_es_update.py
@@ -202,7 +202,7 @@ def _load_observations_and_responses(
 
     # Identifies non-outlier observations based on responses.
     ens_mean = S.mean(axis=1)
-    ens_std = S.std(ddof=0, axis=1)
+    ens_std = S.std(ddof=0, axis=1, dtype=np.float64).astype(np.float32)
     ens_std_mask = ens_std > std_cutoff
     ens_mean_mask = abs(observations - ens_mean) <= alpha * (ens_std + scaled_errors)
     obs_mask = np.logical_and(ens_mean_mask, ens_std_mask)
--- a/tests/ert/ui_tests/cli/test_update.py
+++ b/tests/ert/ui_tests/cli/test_update.py
@@ -1,0 +1,241 @@
+from __future__ import annotations
+
+import os
+import stat
+import warnings
+from pathlib import Path
+
+import hypothesis.strategies as st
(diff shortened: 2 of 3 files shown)


sondreso (approved):
Nice catch!

PR #9635, opened 2025-01-02
Fixed plotter remembering correct tab when switching plot types

### LINKED ISSUE #9451: Plotter not switching to correct plot type when `Std Dev` is available
Run `ES-MDA` using `snake_oil_field.ert` and plot `PORO` which automatically shows `Std Dev`.
Now switch to a `GEN_KW` and then to a `Summary` and notice that the correct tab is not always selected.
Sometimes an `Ensemble` is selected when a `GEN_KW` is chosen, and sometimes `Histogram` is selected when a `Response` is highlighted.
This seems to only be a problem when the `Std Dev` option is available, that is when a `FIELD` is updated.

andreas-el:
**Issue**
Resolves #9451 

**Approach**
This pull request addresses an issue where the plotter was not storing the correct tab when switching between plot types. The approach involves modifying the relevant code to ensure that the correct tab is stored and displayed when the plot type is changed.

## When applicable

### DIFF
--- a/src/ert/gui/tools/plot/plot_window.py
+++ b/src/ert/gui/tools/plot/plot_window.py
@@ -1,6 +1,6 @@
 import logging
 import time
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
 
 import numpy as np
 import pandas as pd
@@ -34,6 +34,10 @@ ENSEMBLE = "Ensemble"
 HISTOGRAM = "Histogram"
 STATISTICS = "Statistics"
 STD_DEV = "Std Dev"
+
+RESPONSE_DEFAULT = 0
+GEN_KW_DEFAULT = 2
+STD_DEV_DEFAULT = 6
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


src/ert/gui/tools/plot/plot_window.py
```diff
@@ -166,6 +178,7 @@ def __init__(self, config_file: str, parent: QWidget | None):
 
     @Slot(int)
     def currentTabChanged(self, index: Any) -> None:
```
berland, line 180: Should `index` be typed to `int` here or will that break stuff? If it breaks stuff, should we add an assert on the type of index before using it?

andreas-el: I definitely should be int, as the slot is connected to the signal;
```
void QTabWidget::currentChanged(int index)
```
Nice catch!

berland:
Reproduced the bug, and failed to reproduce it after this PR, so looks perfect. Bonus issue discovery in #9657.

berland (approved):

PR #9664, opened 2025-01-06
Remove asyncio.run_in_executor() calls and make scheduler cancel() async

### LINKED ISSUE #9307: Cancelling jobs could result in "unexpected error: cannot schedule new futures after shutdown"
No reproducing steps yet, so need to investigate. The exception has been observed raised from https://github.com/equinor/ert/blob/71ec78c7a3d1a00e8b98a0e0a2d48066358eb49d/src/ert/run_models/base_run_model.py#L504

jonathan-eq:
**Issue**
Resolves #9307 

**Approach**
The commit in this PR removes the asyncio.run_in_executor() calls, and runs the methods directly instead.
It also makes the signal_cancel, and cancel methods async/await instead of asyncio.run_coroutine_threadsafe().

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -50,8 +50,6 @@ class EnsembleEvaluator:
         self._config: EvaluatorServerConfig = config
         self._ensemble: Ensemble = ensemble
 
-        self._loop: asyncio.AbstractEventLoop | None = None
-
         self._events: asyncio.Queue[Event] = asyncio.Queue()
         self._events_to_send: asyncio.Queue[Event] = asyncio.Queue()
         self._manifest_queue: asyncio.Queue[Any] = asyncio.Queue()
@@ -190,7 +188,7 @@ class EnsembleEvaluator:
         if len(events) == 0:
             events = [EnsembleFailed(ensemble=self.ensemble.id_)]
         await self._append_message(self.ensemble.update_snapshot(events))
-        self._signal_cancel()  # let ensemble know it should stop
+        await self._signal_cancel()  # let ensemble know it should stop
 
     @property
     def ensemble(self) -> Ensemble:
@@ -216,7 +214,7 @@ class EnsembleEvaluator:
             event = event_from_json(frame.decode("utf-8"))
             if type(event) is EEUserCancel:
                 logger.debug("Client asked to cancel.")
-                self._signal_cancel()
+                await self._signal_cancel()
             elif type(event) is EEUserDone:
                 logger.debug("Client signalled done.")
                 self.stop()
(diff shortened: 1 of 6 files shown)


eivindjahren (changes requested):
This seems to put the application in a bad state in case the condition is ever false.

jonathan-eq:
I was unable to reproduce this with asyncio, but it was possible when using lower level ThreadPoolExecutor. https://superfastpython.com/threadpoolexecutor-shutdown/

src/ert/run_models/base_run_model.py
```diff
@@ -503,12 +503,14 @@ async def run_monitor(
                         EESnapshotUpdate,
                     }:
                         event = cast(EESnapshot | EESnapshotUpdate, event)
-                        await asyncio.get_running_loop().run_in_executor(
-                            None,
-                            self.send_snapshot_event,
-                            event,
-                            iteration,
-                        )
+                        if asyncio.get_running_loop().is_running():
```
xjules, line 506: Hm, this I think is redundant as this should be always true as this runs inside `asyncio.run(...)`
I would recommend instead:
```python
loop = asyncio.get_running_loop()
if loop.is_closed():
   self.send_snapshot_event(event, iteration)
else:
   await asyncio.get_running_loop().run_in_executor(
    None,
    self.send_snapshot_event,
    event,
    iteration,
)
```

xjules:
Maybe it helps if this evaluator function will become async:
```
    def _signal_cancel(self) -> None:
```
Currently we call it directly from async context.

eivindjahren:
> Maybe it helps if this evaluator function will become async:
> 
> ```
>     def _signal_cancel(self) -> None:
> ```
> 
> Currently we call it directly from async context.

So I think we need to reproduce before anything else, as it is, I am not sure whether this is worse or better.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -333,8 +333,7 @@ def _signal_cancel(self) -> None:
         """
         if self._ensemble.cancellable:
             logger.debug("Cancelling current ensemble")
-            assert self._loop is not None
-            self._loop.run_in_executor(None, self._ensemble.cancel)
```
xjules, line 337: Can we remove `self._loop` attribute as well?

src/ert/run_models/base_run_model.py
```diff
@@ -503,12 +503,9 @@ async def run_monitor(
...
                     }:
                         event = cast(EESnapshot | EESnapshotUpdate, event)
-                        await asyncio.get_running_loop().run_in_executor(
-                            None,
-                            self.send_snapshot_event,
-                            event,
-                            iteration,
-                        )
+
+                        self.send_snapshot_event(event, iteration)
```
xjules, line 507: Let's test this first on bigpoly with 400 realizations

jonathan-eq: Tested with 200 realizations, and there was no difference. I think it should be sufficient

xjules (approved):
Ok. Let's merge it then! Good job! :rocket:

src/ert/run_models/base_run_model.py
jonathan-eq: Tested with 400 realization, and it was the same

eivindjahren (approved):

PR #9673, opened 2025-01-08
Add callback system for ert.storage

### LINKED ISSUE #5867: Add callback system for `ert.storage`
This would make it much easier to work with storage, as currently the only way to propagate information to Qt is by the `ertChanged` signal which is far too vague.

We want the following events:
1. Experiment created or removed
2. Ensemble created or removed
3. Parameter added
4. Response added

This can be on the ensemble level. Ie, adding a parameter will call the function `self._ensemble_changed("parameter:add", self, ...)`.

Calling `storage.refresh()` should loop through any differences and call the appropriate callbacks. Maybe there should be support for batch creation and deletion.

The goal is to replace all `ertChanged` in the GUI to be exactly what they need to be. Eg, caselist only cares about new ensembles so that is what it should listen for. Note: in `ert.storage` the API is callbacks, but in Qt it should become a signal. Ie:
```py
storage.register_event_handler("parameter:add", lambda *args: self.parameterAdded.emit(*args))
```

jonathan-eq:
**Issue**
Resolves #5867

**Approach**
Add callback system for ert.storage

## When applicable

### DIFF
--- a/src/ert/gui/ertnotifier.py
+++ b/src/ert/gui/ertnotifier.py
@@ -5,6 +5,8 @@ from ert.storage import Ensemble, Storage
 
 class ErtNotifier(QObject):
     ertChanged = Signal()
+    experimentChanged = Signal()
+    ensembleChanged = Signal()
     storage_changed = Signal(object, name="storageChanged")
     current_ensemble_changed = Signal(object, name="currentEnsembleChanged")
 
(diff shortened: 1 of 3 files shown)


jonathan-eq:
Closing this as I won't work on it anymore. We should cleanup the snapshot_model and proxy models before proceeding.

PR #9686, opened 2025-01-09
Remove test_run_egg_model

yngve-sk:
test_egg_snapshot does close to the same thing and checks results from the same output


DanSava (approved):
Looks good!

PR #9695, opened 2025-01-09
Fix typo in dependencies

HakonSohoel:
**Issue**
Resolves #my_issue

**Approach**

## When applicable

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ dependencies = [
     "openpyxl",  # extra dependency for pandas (excel)
     "opentelemetry-api",
     "opentelemetry-sdk",
-    "opentelemetry.instrumentation.fastapi",
+    "opentelemetry-instrumentation-fastapi",
     "opentelemetry-instrumentation-httpx",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -45,7 +45,7 @@ dependencies = [
     "openpyxl",  # extra dependency for pandas (excel)
     "opentelemetry-api",
     "opentelemetry-sdk",
-    "opentelemetry.instrumentation.fastapi",
+    "opentelemetry-instrumentation-fastapi",
```
berland, line 48: Since this passed initial tests when originally merged, is this particular fastapi pulled in via other packages declaring it as a dependency (?) (if so we can skip it here but need to keep an eye out for Komodo)

HakonSohoel: Seems like pip accepts dot instead of dash, but I still prefer to use dash for consistency 🤓

berland: Then we should merge!

berland (approved):

PR #9712, opened 2025-01-13
Upgrade ruff in pre commit to 0.9.1

larsevj:
Mainly formatting changes, there is also the change to PluginMetadate into a dataclass based on the following rule https://docs.astral.sh/ruff/rules/class-as-data-structure/, but could also ignore this rule if we want to.

## When applicable


berland:
Looks good, but could you reverse the commit order to make every commit passing tests?

src/everest/simulator/__init__.py
```diff
@@ -4,7 +4,7 @@
 JOB_FAILURE = "Failed"
 
 
-DEFAULT_DATA_SUMMARY_KEYS = ("YEAR", "YEARSTCPU", "TCPUDAY", "MONTH", "DAY")
+DEFAULT_DATA_SUMMARY_KEYS = ("YEAR", "YEARS", "TCPU", "TCPUDAY", "MONTH", "DAY")
```
berland, line 7: wow!

berland: Is it worth another issue to track down why this was not caught in tests earlier?

berland:
> > Looks good, but could you reverse the commit order to make every commit passing tests?
> 
> I planned on squashing them, was just unsure if it was best to use `NamedTuple` or `dataclass` for the PluginMetadata class.

NamedTupe is fine

berland (approved):

PR #9780, opened 2025-01-16
Add more snapshots to test_api_summary_snapshot

yngve-sk:
Minimal does not cover case where there are some batches with only function results, only gradients results, etc. Multiobj covers this. Adding advanced for some extra coverage

Batches/ensembles of  multiobj:
```
experiment
Out[2]: <ert.storage.local_experiment.LocalExperiment at 0x17d087050>
experiment.ensembles
Out[3]: <generator object LocalExperiment.ensembles at 0x17d1aaa40>
[e for e in experiment.ensembles]
Out[4]: 
[<ert.storage.local_ensemble.LocalEnsemble at 0x17d1fd070>,
 <ert.storage.local_ensemble.LocalEnsemble at 0x17d1ffb90>,
 <ert.storage.local_ensemble.LocalEnsemble at 0x17d1ff590>]
[e.ensemble_size for e in experiment.ensembles]
Out[5]: [1, 1, 6]
```


StephanDeHoop (approved):
As discussed on Slack, something fishy is going on with the snapshots for these tests (only retrieving the results for the unperturbed controls up until the optimal batch) (the test fixture seems fine as in the results are on disk in the correct tmp folder etc., might be bug in SebaSnapshot/Database or just something weird with the test). 

I leave it up to you to go ahead and merge or not. This will need to be addressed anyway in your big storage PR.

PR #9781, opened 2025-01-16
Remove everest.ConfigKeys

### LINKED ISSUE #9018: Remove the everest ConfigKeys class
The Everest ConfigKeys class (not to be confused with the class of the same name in ERT) serves no purpose, it is essentially a list of constant definitions. That class is not used for anything else, since keyword checking is handled by the pydantic configuration parser. The class should be removed.

verveerpj:
**Issue**
Resolves #9018 

**Approach**
Search and replace...

## When applicable


yngve-sk (commented):
The replace part looks OK to me, but I wonder, there are so many places we use EverestConfig as if it is only a dict, should we maybe also replace those using attribute accessor?

frode-aarstad (approved):
LGTM

Issue #9782, opened 2025-01-16
zmq: Implement ACK or re-connection for Monitor


xjules:
In a rare occasion, the local connection (between monitor and ensemble evaluator) can go down. In such a case, the messages sent by the evaluator will be lost and the GUI won't be updated. In this case we have two options:
1) The messages sent by ensemble evaluator can get `ACK` similar to dispatchers messages
2) When re-connection happens, monitor needs to sent CONNECT_MSG to receive a Full snapshot again.

The symptoms of this behavior are described in this faulty network testing.
https://github.com/equinor/ert/issues/9734

PR #9798, opened 2025-01-17
Do heartbeat on clients

xjules:
**Issue**
Resolves https://github.com/equinor/ert/issues/9782

**Approach**
Heartbeat task sends HEARTBEAT to all the clients (ie. Monitor) at client.HEARTBEAT_TIMEOUT intervals.
Clients do not reply, just process the message. If client detects a longer delay between two heartbeats, the client will send CONNECT to evaluator in addition; ie. getting the connection re-established after a break. This is to simulate re-connection. Each CONNECT_MSG will then trigger sending FullSnapshot from the ensemble evaluator.

Initially HEARTBEAT_TIMEOUT is set to 5 seconds while Monitor accepts 10 seconds at max as a delay. Additionally, initial connection will now undergo same amount of retries as standard messages.

## When applicable

### DIFF
--- a/src/_ert/forward_model_runner/client.py
+++ b/src/_ert/forward_model_runner/client.py
@@ -18,6 +18,8 @@ class ClientConnectionError(Exception):
 CONNECT_MSG = b"CONNECT"
 DISCONNECT_MSG = b"DISCONNECT"
 ACK_MSG = b"ACK"
+HEARTBEAT_MSG = b"BEAT"
+HEARTBEAT_TIMEOUT = 5.0
 
 
 class Client:
@@ -83,7 +85,7 @@ class Client:
         await self._term_receiver_task()
         self._receiver_task = asyncio.create_task(self._receiver())
         try:
-            await self.send(CONNECT_MSG, retries=1)
+            await self.send(CONNECT_MSG)
         except ClientConnectionError:
             await self._term_receiver_task()
             self.term()
@@ -93,11 +95,23 @@ class Client:
         raise NotImplementedError("Only monitor can receive messages!")
 
     async def _receiver(self) -> None:
+        last_heartbeat_time: float | None = None
         while True:
             try:
                 _, raw_msg = await self.socket.recv_multipart()
                 if raw_msg == ACK_MSG:
                     self._ack_event.set()
+                elif raw_msg == HEARTBEAT_MSG:
+                    if (
+                        last_heartbeat_time
+                        and (asyncio.get_running_loop().time() - last_heartbeat_time)
+                        > 2 * HEARTBEAT_TIMEOUT
+                    ):
+                        await self.socket.send_multipart([b"", CONNECT_MSG])
+                        logger.warning(
+                            f"{self.dealer_id} heartbeat failed - reconnecting."
+                        )
+                    last_heartbeat_time = asyncio.get_running_loop().time()
                 else:
                     await self.process_message(raw_msg.decode("utf-8"))
             except zmq.ZMQError as exc:
@@ -144,5 +158,5 @@ class Client:
                 self.socket.connect(self.url)
                 backoff = min(backoff * 2, 10)  # Exponential backoff
         raise ClientConnectionError(
-            f"{self.dealer_id} Failed to send {message!r} after retries!"
+            f"{self.dealer_id} Failed to send {message!r} to {self.url} after retries!"
         )
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -5,6 +5,7 @@ import datetime
 import logging
 import traceback
 from collections.abc import Awaitable, Callable, Iterable, Sequence
+from enum import Enum
 from typing import Any, get_args
 
 import zmq.asyncio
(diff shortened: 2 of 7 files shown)


xjules:
Tested with bigpoly on LSF (400 reals, 200 max running). When cutting down the connection the events were not sent fully; despite realizations marked as finished:

image

tests/ert/unit_tests/ensemble_evaluator/test_ensemble_client.py
```diff
@@ -18,31 +21,54 @@ async def test_invalid_server():
...
+    monkeypatch.setattr(_ert.forward_model_runner.client, "PING_TIMEOUT", 0.1)
+
+    async with (
+        MockZMQServer(unused_tcp_port) as mock_server,
+        Client(url) as client,
+    ):
+        await client.send("start", retries=1)
+
+        await mock_server.do_ping()
+        await asyncio.sleep(1)
```
jonathan-eq, line 66: Do we sleep here to force the monitor to reconnect?

xjules: Yes, that is correct. PING_TIMEOUT is 0.1 so sleeping 1 seconds guarantees re-connection.

src/_ert/forward_model_runner/client.py
```diff
@@ -93,11 +95,23 @@ async def process_message(self, msg: str) -> None:
...
         while True:
             try:
                 _, raw_msg = await self.socket.recv_multipart()
                 if raw_msg == ACK_MSG:
                     self._ack_event.set()
+                elif raw_msg == PING_MSG:
+                    if (
+                        last_ping_time
+                        and (asyncio.get_running_loop().time() - last_ping_time)
+                        > 2 * PING_TIMEOUT
```
jonathan-eq, line 108: I guess the 2* here is for the ping to be received and returned, but do we need it if we are not returning the ping? Maybe in our case it should be added in the `PING_TIMEOUT` directly?

xjules: This `2 * PING_TIMEOUT` is to account for a potential delay, which can happen if the ee is very busy.

src/_ert/forward_model_runner/client.py
```diff
@@ -93,11 +95,23 @@ async def process_message(self, msg: str) -> None:
         raise NotImplementedError("Only monitor can receive messages!")
 
     async def _receiver(self) -> None:
+        last_beat_time: float | None = None
```
jonathan-eq, line 98: Maybe last_heartbeat_time?

jonathan-eq (changes requested):
Change last_beat_time -> last_heartbeat_time, and it's good to go! 🚢

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -342,6 +360,7 @@ async def _start_running(self) -> None:
             raise ValueError("no config for evaluator")
         self._ee_tasks = [
             asyncio.create_task(self._server(), name="server_task"),
+            asyncio.create_task(self._do_heartbeat_clients(), name="beat_task"),
```
jonathan-eq, line 363:
```suggestion
            asyncio.create_task(self._do_heartbeat_clients(), name="heartbeat_task"),
```

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -51,7 +57,7 @@ def __init__(self, ensemble: Ensemble, config: EvaluatorServerConfig):
         self._ensemble: Ensemble = ensemble
 
         self._events: asyncio.Queue[Event] = asyncio.Queue()
-        self._events_to_send: asyncio.Queue[Event] = asyncio.Queue()
+        self._events_to_send: asyncio.Queue[Event | bytes] = asyncio.Queue()
```
sondreso, line 60: Could we be more specific here?

sondreso (approved):
Apart from my small comment this LGTM!

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -45,13 +51,17 @@
 EVENT_HANDLER = Callable[[list[Event]], Awaitable[None]]
 
 
+class HeartbeatEvent(Enum):
+    event = b"BEAT"
```
jonathan-eq, line 55: Rather than hardcoding `b"BEAT"`, can we import it from client?

tests/ert/utils.py
```diff
@@ -64,16 +69,18 @@ def wait_until(func, interval=0.5, timeout=30):
 class MockZMQServer:
     def __init__(self, port, signal=0):
         """Mock ZMQ server for testing
-        signal = 0: normal operation
+        signal = 0: normal operation, receive messages but don't store CONNECT and DISCONNECT messages
         signal = 1: don't send ACK and don't receive messages
         signal = 2: don't send ACK, but receive messages
+        signal = 3: normal operation, and store also CONNECT and DISCONNECT messages
```
jonathan-eq, line 75: Would it be a lot of work to have this signal as 2? I think it would look better to have it sorted by normal operations first, then the special cases

xjules: It would require more test refactoring so keeping this setup for now.

sondreso (approved):

PR #9805, opened 2025-01-20
Refactor control handling in everest config

verveerpj:
**Issue**
The parsing code in `everest2ropt` for the controls is extremely convoluted, due to the nested character of the controls section in the configuration. This PR simplifies the parsing code by creating an intermediate class that stores the controls and its properties in a linear fashion. The relevant code in `everest2ropt` is rewritten accordingly, becoming much more easy to understand.

**Approach**

## When applicable

### DIFF
--- a/src/everest/config/everest_config.py
+++ b/src/everest/config/everest_config.py
@@ -2,7 +2,6 @@ import logging
 import os
 from argparse import ArgumentParser
 from copy import copy
-from functools import cached_property
 from io import StringIO
 from itertools import chain
 from pathlib import Path
@@ -623,25 +622,11 @@ and environment variables are exposed in the form 'os.NAME', for example:
         controls = self.controls or []
         return [control.name for control in controls]
 
-    @cached_property
-    def control_name_tuples(self) -> list[tuple[str, str, int | tuple[str, str]]]:
-        tuples = []
-        for control in self.controls:
-            for variable in control.variables:
-                if isinstance(variable, ControlVariableGuessListConfig):
-                    for index in range(1, len(variable.initial_guess) + 1):
-                        tuples.append((control.name, variable.name, index))
-                elif variable.index is not None:
-                    tuples.append((control.name, variable.name, variable.index))
-                else:
-                    tuples.append((control.name, variable.name))
-        return tuples
-
     @property
     def objective_names(self) -> list[str]:
         return [objective.name for objective in self.objective_functions]
 
-    @cached_property
+    @property
     def constraint_names(self) -> list[str]:
         names: list[str] = []
 
@@ -651,16 +636,12 @@ and environment variables are exposed in the form 'os.NAME', for example:
                 names.append(name if suffix is None else f"{name}:{suffix}")
 
         for constr in self.output_constraints or []:
+            _add_output_constraint(constr.target)
             _add_output_constraint(
-                constr.target,
+                constr.upper_bound, None if constr.lower_bound is None else "upper"
             )
             _add_output_constraint(
-                constr.upper_bound,
-                None if constr.lower_bound is None else "upper",
-            )
-            _add_output_constraint(
-                constr.lower_bound,
-                None if constr.upper_bound is None else "lower",
+                constr.lower_bound, None if constr.upper_bound is None else "lower"
             )
 
         return names
--- a/src/everest/config/utils.py
+++ b/src/everest/config/utils.py
@@ -1,0 +1,117 @@
+from collections.abc import Iterator
+
+from .control_config import ControlConfig
+from .control_variable_config import (
+    ControlVariableConfig,
+    ControlVariableGuessListConfig,
+)
+from .sampler_config import SamplerConfig
+
+
+class FlattenedControls:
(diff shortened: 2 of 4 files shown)


src/everest/optimizer/everest2ropt.py
```diff
@@ -285,14 +139,19 @@ def _parse_objectives(objective_functions: list[ObjectiveFunctionConfig], ropt_c
...
-    formatted_names,
-):
-    if not input_constraints:
+def _parse_input_constraints(ever_config: EverestConfig, ropt_config):
+    if not ever_config.input_constraints:
         return
 
+    formatted_names = [
+        (
+            f"{control_name[0]}.{control_name[1]}-{control_name[2]}"
```
StephanDeHoop, line 148: We have this logic twice (depending on the control_name size), is it worth making a separate class out of it (to have the logic in one place) that has a `__str__` which we could  use here? Maybe not necessary.

StephanDeHoop (approved):
Looks good to me! everest2ropt is indeed much cleaner like this. I have some minor comments that you might want to look at before merging.

src/everest/config/everest_config.py
```diff
@@ -623,25 +622,11 @@ def control_names(self):
...
-                else:
-                    tuples.append((control.name, variable.name))
-        return tuples
-
     @property
     def objective_names(self) -> list[str]:
         return [objective.name for objective in self.objective_functions]
 
-    @cached_property
+    @property
```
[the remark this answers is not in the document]
verveerpj: At some point I ran into an issue where `@cached_property` did not work properly. `EverestConfig` is not immutable and if you change it afterwards by assigning to it the cache is not updated. So that is a bit dangerous. There is currently an effort to make it immutable again, then we could use `@cached_property`.

src/everest/config/utils.py
```diff
@@ -0,0 +1,117 @@
...
+                        if variable.index is None
+                        else (control.name, variable.name, variable.index)
+                    )
+
+                    if variable.sampler is not None:
+                        self._samplers.append(variable.sampler)
+                        var_dict["sampler_idx"] = len(self._samplers) - 1
+                    else:
+                        if control.sampler is not None and control_sampler_idx < 0:
+                            self._samplers.append(control.sampler)
```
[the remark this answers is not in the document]
verveerpj: It is indeed a lot of nesting, I will have a look to reduce that.

src/everest/optimizer/everest2ropt.py
DanSava: Or maybe just have it as a property of everest config.

src/everest/optimizer/everest2ropt.py
```diff
@@ -285,14 +139,19 @@ def _parse_objectives(objective_functions: list[ObjectiveFunctionConfig], ropt_c
         ropt_config["function_transforms"] = transforms
 
 
-def _parse_input_constraints(
-    input_constraints: list[InputConstraintConfig] | None,
-    ropt_config,
-    formatted_names,
-):
-    if not input_constraints:
+def _parse_input_constraints(ever_config: EverestConfig, ropt_config):
```
DanSava, line 142: I remember there was an issue some time back refactoring code to avoid using the full Everest config as function arguments.

src/everest/config/utils.py
```diff
@@ -0,0 +1,117 @@
...
+                            control_sampler_idx = len(self._samplers) - 1
+                        var_dict["sampler_idx"] = control_sampler_idx
```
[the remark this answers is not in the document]
verveerpj: Maybe I misunderstand the question, but it works like this:
- If no variable sampler is given, the control sampler will be used
- If no variable sampler is given and no controls sampler is given, that variable will not be perturbed
- If no samplers are given at all anywhere, a default sampler is used for everything.

src/everest/optimizer/everest2ropt.py
verveerpj: I agree, but I think we should do this in the context of this issue: #9816

src/everest/optimizer/everest2ropt.py
verveerpj: I don't know if it is an issue, but since that seems to be our policy now, I will fix it.

src/everest/optimizer/everest2ropt.py
verveerpj: That issue points out that the formatting is not consistent, and for input constraints it is defined here. When that issue is resolved, it could become a property of the config.

src/everest/config/utils.py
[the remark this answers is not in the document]
StephanDeHoop: Ah I see, thanks :)!

verveerpj:
@StephanDeHoop , @DanSava : did some refactoring, I will hold off a bit before merging in case you want to have a look.

StephanDeHoop:
> @StephanDeHoop , @DanSava : did some refactoring, I will hold off a bit before merging in case you want to have a look.

Looks good, thanks :)!

PR #9813, opened 2025-01-20
Fix storage service not being closed

eivindjahren:
In ec5397136b5122e8f34c11d9e0df672a3d93affe the implementation of closing on parent process death based on `prctl` was replaced by one using `os.ppid` relying on a process being adopted by process 1 whenever parent dies. However, this is not always the case: https://stackoverflow.com/a/40996664/283240 , e.g. on ubuntu 24.04, the process is adopted by `systemd --user`.

Instead we get the parent pid on start up and check that it is still that pid.

Also adds `check-all` to justfile

      --exec 'pytest tests/ert/unit_tests -n auto --hypothesis-profile=fast -m "not integration_test"'`)

## When applicable

### DIFF
--- a/justfile
+++ b/justfile
@@ -11,3 +11,13 @@ snake_oil:
 # execute rapid unittests
 rapid-tests:
     nice pytest -n auto tests/ert/unit_tests --hypothesis-profile=fast -m "not integration_test"
+
+check-all:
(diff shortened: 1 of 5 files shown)


andreas-el (approved):
Let's put this into action! 🌤️

PR #9820, opened 2025-01-21
Add dev documentation on everest vs ert data models

StephanDeHoop:
**Issue**
Resolves #[9619](https://github.com/equinor/ert/issues/9619)

**Approach**
Attempts to clarify the differences between ERT and EVEREST data models. There are things in here that need revision (like the mapping, these equations work but not sure if it makes sense in all cases of optimization meaning gradient vs. derrivative free methods etc.). Also, with Yngve we suggest a couple of name changes: 
- `simulation` -> `ert_realization` (at least in the code, since they are synonymous)
- `geo_realization` -> `model_realization` (or just `realization` towards the user, we are not just concerned with GEO apps?)
- `<GEO_ID>` - > `<STATIC_MODEL_ID>` or `<MODEL_REALIZATION_ID>`

      --exec 'pytest tests/ert/unit_tests -n auto --hypothesis-profile=fast -m "not integration_test"'`)

## When applicable


docs/everest/development.rst
```diff
@@ -52,3 +54,73 @@ long as the optimization process is running.
...
+
+.. math::
+
+	r(g, p) = g * (p<0) + (p + g * P + G) (p>=0),
+
+if `batch` has `unperturbed` and `perturbed controls`, where `r` is the ERT `realization_id` (0, ..., `R` - 1), `g` is the `<GEO_ID>` (0, ..., `G` - 1), `p` is `pertubation_id` (-1, 0, ..., `P` - 1), `R`
+is the total number of ERT `realizations`, `G` is the total number of static `model_realizations`, `P` is the total number of pertubations.
+NOTE: `p = -1` for `unperturbed controls`, and `p = 0, ..., P - 1` for `perturbed controls`.
+**THIS IS MY SUGGESTION AND CURRENTLY NOT HOW IT WORKS AND ONLY VALID FOR GRADIENT BASED OPTIMIZATION ALGORITHMS I GUESS?
+If we don't want `p` to be negative we need to use a flag (e.g., `is_pertubation`)**
```
verveerpj, line 115: This is correct for gradient based algorithms, where we assume there may either a single control vector for function evaluation, a set of perturbed controls for gradient evaluation, or both.

Discrete optimizers will not use perturbed controls, but may pass multiple unperturbed controls if they support parallel evaluation (which they usually do and is very important).

So we do need to support identifying multiple function evaluations. I don't see a use case for supporting multiple gradient evaluations concurrently, but who knows...

docs/everest/development.rst
```diff
@@ -52,3 +54,73 @@ long as the optimization process is running.
...
+
+Another thing to note is that continuity for `realizations` between `ensemble` exists; however, this is not the case for `simulations` in `batches`.
+A `batch` can contain several different configurations (Fig 5) and `simulation 0` for `<GEO_ID> = 0` can be either `unperturbed`
+or `perturbed controls`. `<GEO_ID>` is continuous from one `batch` to the next since they are not changing at all over the course of the optimization.
```
verveerpj, line 119: Note however, that in some cases the optimizer may decide that some realizations are not needed. These will then be skipped, and simulation numbering will be different (since there are less realizations). In such a case, the mapping using the math above will break.

verveerpj (commented):
In principle correct. But some amendments.

docs/everest/development.rst
```diff
@@ -52,3 +54,54 @@ long as the optimization process is running.
    * - POST
      - '/stop'
      - Signal everest optimization run termination. It will be called by the client when the optimization needs to be terminated in the middle of the run
+
+
+EVEREST vs. ERT data models
+===========================
+EVEREST uses ERT for running an experiment, but instead of submitting an `ensemble` (ERT) to the queue we submit
```
yngve-sk, line 61: Would compress the intro a bit more and focus a bit more on technical ERT-side, since Everest mostly uses ERT to submit a forward model, and store its results. I'm not sure if the history matching vs optimization is relevant in this developer-facing context.

The main point I think is that Everest uses ERT by (1) mapping controls to parameters which are forward model inputs, (2) running the forward model and having the results stored by ERT, (3) reading the results of the forward model run and mapping them to objectives/constraints.  In this context, a single realization run corresponds to one run of the forward model, one set of inputs and outputs.

We can say this part without introducing history matching vs optimization, ensemble vs batch etc beforehand.

Then afterwards we can introduce the batch vs ensembles as you do

docs/everest/development.rst
```diff
@@ -52,3 +54,54 @@ long as the optimization process is running.
...
+a `batch` in EVEREST. `Batches` are in principle very similar to `ensembles`, ERT queue system doesn't treat them differently,
+but they have some hierarchical differences in terms of the meaning behind the data.
+ERT history matches `realizations` (i.e., `model parameters`) to data, hence an `ensemble` contains a number of `realizations`.
+EVEREST optimizes a set of `controls` and assumes static (i.e., unchanging) `realizations`.
+In terms of collecting the results of forward model runs, there is a distinction between `unperturbed controls`
```
yngve-sk, line 66: I think a brief summary of the control flow of Everest wrt running batches would be good to lead into this. i.e., what Everest decides to put in a batch and why. AFAIK it is either to evaluate an objective and/or generate a gradient . If only objective, there are no perturbations, if only gradients, there are only perturbations, or there may be both.

docs/everest/development.rst
```diff
@@ -52,3 +54,54 @@ long as the optimization process is running.
...
+
+.. figure:: images/Everest_vs_Ert_02.png
+    :align: center
+    :width: 700px
+    :alt: Additional explanation of Fig 3
+
+    Different meaning of `realization` and `simulation`.
+
+As is evident from the image above, in terms of execution in the queue `realization` (ERT) and `simulation` (EVEREST) are synonymous.
+This means that ERT queue system is agnostic about the meaning of each run only when the data is collected back in EVEREST (`GEN_DATA`) is meaning
```
yngve-sk, line 88: I think wrt my first comment, this info may be stated at the start and thus be redundant here.

docs/everest/development.rst
```diff
@@ -52,3 +54,54 @@ long as the optimization process is running.
...
+of each run attributed.
+The mapping from data models in EVEREST and ERT is done in the `ropt` library, it maps from `realization` (ERT) to `<GEO_ID>` and `pertubation` (EVEREST) and vice versa.
+`Batches` in EVEREST can contain several different configurations depending on the algorithm used. Gradient-based algorithms can have a single function
+evaluation (`unperturbed controls`) per `<GEO_ID>`, a set of `perturbed controls` per `<GEO_ID>` to evaluate the gradient, or both.
+Derivative-free methods can have several function evaluations per `<GEO_ID>` and no `perturbed controls`.
+**NOTE:** the optimizer may decide that some `<GEO_ID>` are not needed, these are then skipped and the mapping from `ropt`
+should reflect this (i.e., less `<GEO_ID>` in the batch results than expected).
+
+Another thing to note is that continuity for `realizations` between `ensemble` exists; however, this is not the case for `simulations` in `batches`.
+A `batch` can contain several different configurations (Fig 5) and `simulation 0` for `<GEO_ID> = 0` can be either `unperturbed`
```
yngve-sk, line 98: I didn't find Fig 5? I'm not exactly sure what this sentence is trying to say? That there are different mappings between (geoid, perturbation) and (ert realization)? I think wrt https://github.com/equinor/ert/pull/9767 , we will receive this mapping from ROPT, so it might be sufficient to say that ROPT will provide that mapping and we will attach that info to the stored forward model results in ERT.

docs/everest/development.rst
```diff
@@ -52,3 +54,59 @@ long as the optimization process is running.
    * - POST
      - '/stop'
      - Signal everest optimization run termination. It will be called by the client when the optimization needs to be terminated in the middle of the run
+
+
+EVEREST vs. ERT data models
+===========================
+EVEREST uses ERT for running an experiment. Whenver an EVEREST `experiment` is ran, `everserver` creates an `EverestRunModel` based on the
+`ErtConfig` (obtained from `EverestConfig`). The `ErtConfig` requires an `EnsembleConfig` where EVEREST `objectives`
+and `constraints` are mapped to ERT `responses` using `GenDataConfig` (stored in `EnsembleConfig.response_config`).
```
yngve-sk, line 63: I think the part about the  `ErtConfig` can be omitted, but saying that constraints and objectives are mapped to `GEN_DATA` within ERT makes sense. You could maybe abstract it one level up and say that constraints and objectives are simply responses from the forward model run by ERT. The details about callsites etc are useful and nice, but if you include them, at least note that they are subject to change, which is why I would opt to write this on a level of detail so that they can change but the text written can still hold true. ERT will always have (1) a thing to submit realizations (forward model evaluations) to the queue, (2) a runpath where forward model reads&writes its own files, and (3) a storage where parameters and responses are stored on a per-realization basis, so these can be used to explain the thing. So when using ERT we create a storage, then run a forward model which in turn populates the storage with parameters (controls) and responses (objs&constraints).

docs/everest/development.rst
```diff
@@ -52,3 +54,59 @@ long as the optimization process is running.
...
+In `everest_to_ert_config` all the EVEREST `controls` are mapped to ERT `parameters` using `ExtParamConfig`
+(stored in `EnsembleConfig.parameter_configs`). Then in `EverestRunModel._forward_model_evaluator` (used as evaluator
+by the external optimization library `ropt`) a local storage is created for the `batch` (i.e., `ensemble`),
+linked to the current `experiment`, and setup up `simulations` for all sets of `controls`.
+
+All the forward models in the `batch` are then evaluated in ERT using `EverestRunModel._evaluate_and_postprocess`
+(and in turn `BaseRunModel.run_workflows`) which uses the `EnsembleEvaluator` for execution. After ERT writes
+to `simulation_results` and stores results in the `LocalEnsemble` (i.e., storage)
```
yngve-sk, line 71: I think this can be shortened down a bit too. These parts you wrote should definitely stay though: (1) the part about `simulation_results` being the folder location of the ERT storage containing the realizations is important. (2) ROPT reading them from there is also good. The other callee/callsite details I think you can remove

docs/everest/development.rst
```diff
@@ -52,3 +54,59 @@ long as the optimization process is running.
...
+these results are then gathered in the forward model evaluator (called by `ropt`) and feedforward into the optimizer using
+`EverestRunModel._gather_simulation_results` (i.e., reading the `.parquet` files generated by ERT). Every set of
+forward model executions for a single set of inputs (`controls`) and outputs (`objectives` / `constraints`)
```
yngve-sk, line 74: `Every forward model evaluation` instead of set would be correct here I think. An ERT realization is a single evaluation of the forward model. Multiple ERT realizations constitute an ensemble.

I'm not sure the "in the ERT queue system" makes sense in this context, it is true that it runs on the queue, but the important part is that an ERT realization represents a run of the forward model, i.e., its parameters and resulting responses.

docs/everest/development.rst
```diff
@@ -52,3 +54,50 @@ long as the optimization process is running.
...
      - '/stop'
      - Signal everest optimization run termination. It will be called by the client when the optimization needs to be terminated in the middle of the run
+
+
+EVEREST vs. ERT data models
+===========================
+EVEREST uses ERT for running an `experiment`. An `experiment` contains several `batches` (i.e., `ensembles`).
+EVEREST `controls` are mapped to ERT `parameters` and serve as input to each forward model evaluation.
+ERT generates `responses` for each forward model evaluation in the `batch`.
+ERT writes these `responses` to storage in the `simulation_results` folder (per `batch` and per `realization`, as defined in the `runpath`).
```
yngve-sk, line 64: ... creating one `ensemble` per `batch`. Each ensemble consists of several `realizations`.

yngve-sk, line 64: I think the as defined in the runpath part can be removed, as ERT writes to storage in a "flat" manner, and the runpath is grouped by geo-realization->[perturbation|result_id]

docs/everest/development.rst
```diff
@@ -52,3 +54,50 @@ long as the optimization process is running.
    * - POST
      - '/stop'
      - Signal everest optimization run termination. It will be called by the client when the optimization needs to be terminated in the middle of the run
+
+
+EVEREST vs. ERT data models
+===========================
+EVEREST uses ERT for running an `experiment`. An `experiment` contains several `batches` (i.e., `ensembles`).
```
yngve-sk, line 61: Append this: An `ensemble`/`batch` contains inputs & computed outputs for several `realizations`, denoted as `simulations` in Everest.

docs/everest/development.rst
```diff
@@ -52,3 +54,50 @@ long as the optimization process is running.
...
+EVEREST `controls` are mapped to ERT `parameters` and serve as input to each forward model evaluation.
+ERT generates `responses` for each forward model evaluation in the `batch`.
+ERT writes these `responses` to storage in the `simulation_results` folder (per `batch` and per `realization`, as defined in the `runpath`).
+These `responses` are mapped to `objectives` and `constraints` in EVEREST and read by `ropt` (i.e., the optimizer).
+To summarize, every forward model evaluation for a single set of inputs/parameters (`controls`) and generated outputs/responses (`objectives` / `constraints`)
+constitutes an ERT `realization` which is denoted in EVEREST as a `simulation`.
+
+For `controls` in EVEREST, there is a distinction between `unperturbed controls` (i.e., current `objective function` value) and
```
yngve-sk, line 69: This part: (i.e., current `objective function` value) , do you mean "current controls"? I also heard this term alot on the Everest workshop. I think it would be good to opt for a definition on first usage, like (controls of the current best solution yielding the best objective value(s)) or something like that. I guess also if it is explained properly earlier it is not too bad, if there is an explanation it would be good to hyperlink to it. If not, would be good to add it somewhere it makes sense, what a "current control" is.

docs/everest/development.rst
```diff
@@ -52,3 +54,50 @@ long as the optimization process is running.
...
      - Signal everest optimization run termination. It will be called by the client when the optimization needs to be terminated in the middle of the run
+
+
+EVEREST vs. ERT data models
+===========================
+EVEREST uses ERT for running an `experiment`. An `experiment` contains several `batches` (i.e., `ensembles`).
+EVEREST `controls` are mapped to ERT `parameters` and serve as input to each forward model evaluation.
+ERT generates `responses` for each forward model evaluation in the `batch`.
+ERT writes these `responses` to storage in the `simulation_results` folder (per `batch` and per `realization`, as defined in the `runpath`).
+These `responses` are mapped to `objectives` and `constraints` in EVEREST and read by `ropt` (i.e., the optimizer).
```
verveerpj, line 65: "read by `ropt`" -> "forwarded to `ropt`", they are not read from file.

docs/everest/development.rst
```diff
@@ -52,3 +54,50 @@ long as the optimization process is running.
...
+To summarize, every forward model evaluation for a single set of inputs/parameters (`controls`) and generated outputs/responses (`objectives` / `constraints`)
+constitutes an ERT `realization` which is denoted in EVEREST as a `simulation`.
+
+For `controls` in EVEREST, there is a distinction between `unperturbed controls` (i.e., current `objective function` value) and
+`perturbed controls` (i.e., required to calculate the `gradient`).
+Furthermore, when performing robust optimization (i.e., having multiple static `geo_realizations` / `model_realizations`, NOTE: not the same as an ERT `realization`) a `batch` contains
```
yngve-sk, line 71: I think it should be mentioned briefly here that a geo_realization / static realization represents one setting of `static` variables. In the case of oil/gas it is geology, it is basically the definition of those things that don't change in that world. If not defined properly somewhere in the docs it should be added, but no need to do it in this PR, just add a paranthesis stating briefly what geo-realization/model realization means)

docs/everest/development.rst
```diff
@@ -52,3 +54,50 @@ long as the optimization process is running.
...
+multiple `geo_realizations` (denoted by `<GEO_ID>`) and each `geo_realization` can contain several `simulations`
+(i.e., forward model run). This is the key differences between the hierarchical data model of EVEREST and ERT (Fig 3).
```
yngve-sk, line 73: Grammar: differences -> difference

docs/everest/development.rst
```diff
@@ -52,3 +54,50 @@ long as the optimization process is running.
...
+**NOTE:** the optimizer may decide that some `<GEO_ID>` are not needed, these are then skipped and the mapping from `ropt`
+should reflect this (i.e., less `<GEO_ID>` in the batch results than expected).
+
+.. figure:: images/Everest_vs_Ert_03.png
+    :align: center
+    :width: 700px
+    :alt: Other `batch` configurations EVEREST
+
+    Three other possible configurations of EVEREST `batches` in the context of gradient-based (i.e., `optpp_q_newton`)
+    and gradient-free (i.e., **WHICH ONE DO WE SUPPORT?**) optimization algorithms.
```
yngve-sk, line 103: I'm not sure which ones we do support, whatever is supported by Dakota / Scipy maybe? But the Q should be answered and/or removed before merging this

verveerpj:
Just a general remark: I am not happy with the use of `geo`, `<GEO>` and `geo realization`. That ties the optimizer unnecessarily to geological applications. Considering that the domain should (at least) be energy / energy transition, we should probably be consistent and talk about model realizations rather than geo realizations.

docs/everest/development.rst
verveerpj: We support `pymoo` and `nomad` optimizers as plugins. SciPy also supports at least one discrete optimizer. I suggest not mentioning any specific algorithms at all. I.e. remove the reference to `optpp_q_newton`, references like that tend to become outdated at some point.

StephanDeHoop:
> Just a general remark: I am not happy with the use of `geo`, `<GEO>` and `geo realization`. That ties the optimizer unnecessarily to geological applications. Considering that the domain should (at least) be energy / energy transition, we should probably be consistent and talk about model realizations rather than geo realizations.

Yes I completely agree, that's why I suggested in the description of the PR to change this name. However, since it is changed in the code yet, I thought we should be consistent with what we currently have until we decide on a better name, like `model_realizations`. If you think it's OK to name it already then I'll do that and make another quick PR doing this name change (at least in the code, we can then think about if we should also rename the actual folder name to where we write the output).

docs/everest/development.rst
```diff
@@ -52,3 +54,53 @@ long as the optimization process is running.
...
+constitutes an ERT `realization` (`simulation` in EVEREST).
+
+For `controls` in EVEREST, there is a distinction between `unperturbed controls` (i.e., `controls` of the current best solution
+yielding the best objective function value(s), see also :ref:`opt-process-label`) and
+`perturbed controls` (i.e., required to calculate the `gradient`). Furthermore, when performing robust optimization a `batch` contains
+multiple `model_realizations` (denoted by `<MODEL_ID>`). A `model_realization` is a set of `static` variables (i.e., not changing
+during the optimization) which affect the response of the underlying optimized model (NOTE: `model_realization` is not the same as an
+ERT `realization`; see also :ref:`robust-optimization-label`). Each `model_realization` can contain several `simulations`
+(i.e., forward model run). This is the key difference between the hierarchical data model of EVEREST and ERT (Fig 3).
+NOTE: `<MODEL_ID>` is inserted (and substituted) in the `run_path` for each `model_realization`.
```
yngve-sk, line 77: `<MODEL_ID>` write a (or `<GEO_ID>` due to legacy reasons) in paranthesis here, as we haven't added the `<MODEL_ID>` thing yet, should be a quick thing to do though. Could you set up an issue for it, and do it if it seems straight forward after getting this PR in?

docs/everest/development.rst
```diff
@@ -52,3 +54,53 @@ long as the optimization process is running.
...
+To summarize, every forward model evaluation for a single set of inputs/parameters (`controls`) and generated outputs/responses (`objectives` / `constraints`)
+constitutes an ERT `realization` (`simulation` in EVEREST).
+
+For `controls` in EVEREST, there is a distinction between `unperturbed controls` (i.e., `controls` of the current best solution
+yielding the best objective function value(s), see also :ref:`opt-process-label`) and
+`perturbed controls` (i.e., required to calculate the `gradient`). Furthermore, when performing robust optimization a `batch` contains
+multiple `model_realizations` (denoted by `<MODEL_ID>`). A `model_realization` is a set of `static` variables (i.e., not changing
+during the optimization) which affect the response of the underlying optimized model (NOTE: `model_realization` is not the same as an
+ERT `realization`; see also :ref:`robust-optimization-label`). Each `model_realization` can contain several `simulations`
+(i.e., forward model run). This is the key difference between the hierarchical data model of EVEREST and ERT (Fig 3).
```
yngve-sk, line 76: grammar: forward model runS

yngve-sk (approved):
LGTM, I left some smaller comments I think you can address before merging

PR #9841, opened 2025-01-22
Have parameter config keep parameters types when using design matrix with categorical data

### LINKED ISSUE #9810: DESIGN_MATRIX: export correct parameter types in parameters.json
We need to make sure that parameters in `parameter.json` will be of the correct type, which now is when exported from a single `GenKW` will be converted into strings in case there was a categorical or string data type.

Check here, where we export data to `parameter.json`
https://github.com/equinor/ert/blob/946862b0d9ecc80f08b49bc01825fdf7108466b4/src/ert/storage/local_experiment.py#L126

Recommendation from Mr GPT:
```python
import json

def convert_to_number(value):
    try:
        # Try converting to integer first
        return int(value)
    except ValueError:
        try:
            # If it fails, try converting to float
            return float(value)
        except ValueError:
            # If all fail, return the original value
            return value

def ensure_types(data):
    if isinstance(data, dict):
        return {key: ensure_types(value) for key, value in data.items()}
    elif isinstance(data, list):
        return [ensure_types(item) for item in data]
    else:
        return convert_to_number(data)

# Example data
data = {
    "key1": "123",
    "key2": "45.67",
    "key3": "not a number",
    "key4": ["10", "20.5", "hello"],
}

# Convert and serialize
processed_data = ensure_types(data)
json_str = json.dumps(processed_data)
print(json_str)

```

Or

```python
class CustomJSONEncoder(json.JSONEncoder):
    def default(self, obj):
        if isinstance(obj, str):
            try:
                return int(obj)
            except ValueError:
                try:
                    return float(obj)
                except ValueError:
                    return obj
        return super().default(obj)

# Serialize with custom encoder
json_str = json.dumps(data, cls=CustomJSONEncoder)
print(json_str)

```

jonathan-eq:
**Issue**
Resolves #9810

**Approach**
This commit makes enkf_main's _value_export_json keep types when dumping the json. Before this commit, all values (also numerical) were turned to strings if used alongside categorical values.

parameters.json in runpath for poly_example with design_matrix after this commit:
```
{
"DESIGN_MATRIX" : {
"a" : 0, 
"category" : "cat1", 
"b" : 1, 
"c" : 2
}
}
```

parameters.json in runpath for poly_example with design_matrix prior to this commit:
```
{
"DESIGN_MATRIX" : {
"a" : "0", 
"category" : "cat1", 
"b" : "1", 
"c" : "2"
}
}
```

## When applicable

### DIFF
--- a/src/ert/enkf_main.py
+++ b/src/ert/enkf_main.py
@@ -65,9 +65,24 @@ def _value_export_json(
     if len(values) == 0:
         return
 
+    def parse_value(value: float | int | str) -> float | int | str:
+        if isinstance(value, float | int):
+            return value
+        try:
+            return int(value)
+        except ValueError:
+            try:
+                return float(value)
+            except ValueError:
+                return value
+
     # Hierarchical
-    json_out: dict[str, float | dict[str, float]] = {
-        key: dict(param_map.items()) for key, param_map in values.items()
+    json_out: dict[str, float | dict[str, float | int | str]] = {
+        key: {
+            inner_key: parse_value(inner_value)
+            for inner_key, inner_value in param_map.items()
+        }
+        for key, param_map in values.items()
     }
 
     # Disallow NaN from being written: ERT produces the parameters and the only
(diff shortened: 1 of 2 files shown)


jonathan-eq:
Will add unittest^

xjules (approved):
Nice work @jonathan-eq ! :rocket:

PR #9857, opened 2025-01-23
Have `forward_model_ok` not run if storage in invalid (missing responses.json)

### LINKED ISSUE #9856: ERT spams exceptions if storage is deleted mid run (missing responses.json)

ERT spams exceptions from forward_model_ok if storage is deleted mid run (missing responses.json)

Exception should only be logged once, and we wouldn't try fm_ok again as it is an error we cannot recover from. 

### steps to reproduce

```Python
Run duplicate poly_eval multiple times in poly.ert.
Run it, and delete entire storage folder mid-run.
```

- [ ] python 3.11
- [x] python 3.12
- [x] macosx
- [ ] rhel7
- [ ] rhel8
- [ ] local queue
- [ ] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

jonathan-eq:
**Issue**
Resolves #9856

**Approach**
This commit makes it so that if a realization fails due to `missing responses.json` (can happen if storage is deleted), we will not run `forward_model_ok` for the next realizations, as it is not something we can recover from. We have this logic to avoid spamming `logger.exception(...)` for something we cannot really handle.

## When applicable

### DIFF
--- a/src/ert/callbacks.py
+++ b/src/ert/callbacks.py
@@ -6,7 +6,7 @@ import time
 from pathlib import Path
 
 from ert.config import InvalidResponseFile
-from ert.storage import Ensemble
+from ert.storage import Ensemble, ErtStorageException
 from ert.storage.realization_storage_state import RealizationStorageState
 
 from .load_status import LoadResult, LoadStatus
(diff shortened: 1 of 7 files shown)


src/ert/callbacks.py
```diff
@@ -96,36 +96,47 @@ async def forward_model_ok(
     realization: int,
     iter: int,
     ensemble: Ensemble,
+    forward_model_ok_permanent_error_future: asyncio.Future[str] | None = None,
```
xjules, line 99: how the Future is usually used is that when you do somewhere:
```python
await forward_model_ok_permanent_error_future
```

and somewhere else you get exception Ex you can propagate it to the future by 
```python
forward_model_ok_permanent_error_future.set_exception(Ex)
```

which then gets triggered by the line above (with await).

xjules:
I think it should be fine but it would be great if somebody else can check it too; e.g. @sondreso

jonathan-eq:
I would like to move the flag up a level and have the jobs not run `forward_model_ok` at all, but that would also require moving the `ensemble.set_failure(
            realization, RealizationStorageState.LOAD_FAILURE, final_result.message
        )`. This would however require changes in libres_fascade too 🤔

src/ert/storage/local_experiment.py
```diff
@@ -253,6 +257,10 @@ def parameter_info(self) -> dict[str, Any]:
     @property
     def response_info(self) -> dict[str, Any]:
         info: dict[str, Any]
+        if not (self.mount_point / self._responses_file).exists():
```
eivindjahren, line 260: This was just changed to give a `FileNotFoundError` because the exception did not show full path. Yes the `responses.json` does not exist, but the entire directory might be wrong:
https://github.com/equinor/ert/commit/76a2ba9f5b61c9d4bf08b42adde4edb2b0c62271

src/ert/storage/local_ensemble.py
```diff
@@ -406,8 +407,10 @@ def get_ensemble_state(self) -> list[set[RealizationStorageState]]:
         states : list of RealizationStorageState
             list of realization states.
         """
-
-        response_configs = self.experiment.response_configuration
+        try:
+            response_configs = self.experiment.response_configuration
```
eivindjahren, line 411: It does not immediately seem fine to ignore a storage exception here.

sondreso (commented):
I think we need to be a bit careful here, as there are some potential undesired consequences of this solution. 

For example, if there is an intermittent disk error (flaky network drive) that causes one of the realizations to fail with an `ErtStorageException`, then all future realizations will not load data even if the disk becomes available again. It essentially turns what was an intermittent error into a permanent one. 

I agree that the logging is a bit excessive in this scenario, but it does also alert us to something that is an actual error. We could try to collect all errors and log only once at the end, but this could also have unintended consequences. In general, it's difficult to separate this unrecoverable-error from something that we can recover from. 

I think the better approach here is to reach out to the users that delete their storage mid run, and try to understand why they are doing it. If we understand that we can make informed improvements and don't solve the wrong problem.

eivindjahren:
This seems to mostly have to do with users wanting to selectively deleteing experiments and ensembles. We need to implement https://github.com/equinor/ert/issues/7043 . After that we can assess the need for reducing the amount of notifications we get.

PR #9864, opened 2025-01-24
Set initial state in gui for forward model steps

larsevj:
Introduces a new state equal to pending which is the initial state of a forward model step. When we receive a fm_step_start event we should instead display running.

This also allows for modifying the logic in fm_step_reporter such that it does not send a running event immediately after a start event. This should reduce the number of events that we receive from jobs that run < 5 seconds.

## When applicable

### DIFF
--- a/tests/ert/unit_tests/forward_model_runner/test_job_dispatch.py
+++ b/tests/ert/unit_tests/forward_model_runner/test_job_dispatch.py
@@ -6,7 +6,6 @@ import json
 import os
 import signal
 import stat
-import subprocess
 import sys
 from subprocess import Popen
 from textwrap import dedent
@@ -116,8 +115,9 @@ else:
     os.wait()  # allow os to clean up zombie processes
 
 
+@pytest.mark.integration_test
 @pytest.mark.usefixtures("use_tmpdir")
-def test_memory_profile_is_logged_as_csv():
+def test_memory_profile_is_logged_as_csv(monkeypatch):
     """This tests that a csv is produced and has basic validity.
     It does not try to verify the validity of the logged RSS values."""
     fm_stepname = "do_nothing"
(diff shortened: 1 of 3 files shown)


src/ert/ensemble_evaluator/state.py
```diff
@@ -46,7 +47,8 @@
 }
 
 FORWARD_MODEL_STATE_TO_COLOR: Final = {
-    str(FORWARD_MODEL_STATE_START): COLOR_PENDING,
+    str(FORWARD_MODEL_STATE_INIT): COLOR_PENDING,
+    str(FORWARD_MODEL_STATE_START): COLOR_RUNNING,
```
jonathan-eq, line 51: Should we use the same color for FORWARD_MODEL_STATE_START and FORWARD_MODEL_STATE_RUNNING? Maybe it would be better to have the colors as it is, and intruduce a new color for INIT? Maybe something like light orange (255, 200, 150)

jonathan-eq (commented):
Very good change! Just a small question, see comment 👍

src/ert/ensemble_evaluator/state.py
jonathan-eq: Clarified this, very good job!

xjules:
Very nice PR @larsevj ! 
Some general comments:
 - I'd suggest to improve the commit messages explaining why new state is introduces and why running msg is postoponed
 - have you checked looked at `tests/ert/unit_tests/forward_model_runner/test_event_reporter.py` if this can be a place to include a test for the new state?

tests/ert/unit_tests/forward_model_runner/test_job_dispatch.py
```diff
@@ -126,7 +127,7 @@ def test_memory_profile_is_logged_as_csv():
     with open(scriptname, "w", encoding="utf-8") as script:
         script.write(
             """#!/bin/sh
-        exit 0
+        sleep 6
```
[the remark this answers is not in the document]
xjules: I would go for mocking the poll interval, yes.

PR #9884, opened 2025-01-28
Make sure check-all runs the same as github actions

eivindjahren:
Also uses gnu-parallel in the check-all command which makes it take ~11m which is about 2-3 minutes faster than running the actions.

A few notes on differences:

* cli tests no longer run with `--benchmark-disable --dist loadgroup` as those do nothing unless you use the benchmark fixture or xdist
* `-n 4` is used for unit tests instead of `-n logical` as that means better usage of cores in `just check-all`, note that this increases the number of process used on mac runners for unit tests from 3 to 4.
* `--show-capture=stderr` is no longer used for unit tests, which was the only place where this was used. This only means that stderr is placed together with the captured output instead of inline.

## When applicable


andreas-el (approved):
💯

PR #9887, opened 2025-01-28
Consolidate optimizer options and backend_options

### LINKED ISSUE #9772: Consolidate the `options` and `backend_options` configuration fields
The `optimization` section of the Everest config has both `options` and `backend_options` fields. The `options` field is used to pass options in the form of plain strings, while `backend_options` expects a dictionary of key/value options. Which one to use depends on the optimizer, i.e. Dakota uses `options`, while SciPy uses `backend_options`. The two options should be consolidated into a single option that accepts both forms.

The Everest configuration is implemented using pydantic, which can easily be used to implement a field that accepts both strings and key/value dicts. The `options` field should be adapted to allow this. The `backend_options` should be retained for now, but deprecated. It may be convenient to just make `backend_options` an alias for the new `options` until it is  removed.

N/A

**Additional context**
The `ropt` configuration already uses this consolidated form, hence the implementation should be straightforward.

yngve-sk:
**Issue**
Resolves #9772 

**Approach**
Pass string to dakota, otherwise pass a dict.

### DIFF
--- a/src/everest/optimizer/everest2ropt.py
+++ b/src/everest/optimizer/everest2ropt.py
@@ -1,3 +1,4 @@
+import logging
 import os
 from typing import Any
 
@@ -14,6 +15,7 @@ from everest.config import (
     OutputConstraintConfig,
 )
 from everest.config.utils import FlattenedControls, control_tuples
+from everest.strings import EVEREST
 
 
 def _parse_controls(ever_controls: list[ControlConfig], ropt_config):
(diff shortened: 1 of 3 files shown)


src/everest/optimizer/everest2ropt.py
```diff
@@ -279,19 +279,9 @@ def _parse_optimization(
...
-    options = ever_opt.options or []
-    backend_options = ever_opt.backend_options or {}
-    if options and backend_options:
-        raise RuntimeError("Only one of 'options' and 'backend_options' allowed.")
+    options = ever_opt.options or ever_opt.backend_options or {}
     # The constraint_tolerance option is only used by Dakota:
-    if backend == "dakota":
-        alg_const_tol = ever_opt.constraint_tolerance or None
-        if has_output_constraints and alg_const_tol is not None:
-            options += [f"constraint_tolerance = {alg_const_tol}"]
```
yngve-sk, line 290: Loss of functionality OK?

verveerpj (changes requested):
Looks good, but needs a deprecation message.

verveerpj (approved):
LGTM!

PR #9898, opened 2025-01-29
Enforce objectives and controls have at least one element in everest config

### LINKED ISSUE #9689: Validate keywords that are not allowed to be empty list using conlist
In the `EverestConfig` we allow required keywords of type `list` to be empty. This is probably to easy testing, but for user doesn't really make sense. We can't run an optimization if we don't define any controls for which we are optimizing (unless I am missing something), same goes for `objective_functions`, if we don't specify an objective function what are we optimizing for?

**TODO**: Refactor this code to enforce required keywords of type `list` to be non-empty to improve the use of validating the `config.yml` (as required in actual use cases). This probably also means fixing associated tests that fail because we now require, for example, len(controls) > 0 (e.g., `EverestConfig.with_defaults()` will need to be changed too).

This relates to the config keywords: `controls` and `objective_functions`

DanSava:
**Issue**
Resolves #9689

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/everest/test_config_validation.py
+++ b/tests/everest/test_config_validation.py
@@ -1025,3 +1025,54 @@ def test_load_file_with_errors(copy_math_func_test_data_to_tmp, capsys):
 
     assert "line: 16, column: 5" in captured.err
     assert "Extra inputs are not permitted (type=extra_forbidden)" in captured.err
+
+
+@pytest.mark.parametrize(
+    ["controls", "objectives", "error_msg"],
+    [
+        (
+            [],
+            [],
+            [
+                "controls\n  List should have at least 1 item after validation, not 0",
+                "objective_functions\n  List should have at least 1 item after validation, not 0",
+            ],
+        ),
+        (
+            [],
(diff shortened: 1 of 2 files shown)


tests/everest/test_config_validation.py
```diff
@@ -1025,3 +1025,56 @@ def test_load_file_with_errors(copy_math_func_test_data_to_tmp, capsys):
 
     assert "line: 16, column: 5" in captured.err
     assert "Extra inputs are not permitted (type=extra_forbidden)" in captured.err
+
+
+@skipif_no_everest_models
+@pytest.mark.everest_models_test
```
yngve-sk, line 1031: I think these don't require everest models installed as they only test that pydantic is validating config as it should?

DanSava: It's just some leftover code from my poorly done copy-pasta job.  It is removed now.

yngve-sk (approved):
Nice🚀

PR #9899, opened 2025-01-29
Deprecate backend options

### LINKED ISSUE #9773: Remove `backend` options from the Everest configuration
In the Everest configuration, both the `optimization` and `samplers` sections have a `backend` option. This is used to specify which plugin/backend should be used to look for algorithms/methods that user specifies with the `algorithm` or `method` keys. The `backend` is not strictly necessary anymore, since `ropt` will search through all its available backends to find an algorithm/method. Further, if desired, the backend can be specified by prefixing the `algorithm`/`method` keyword like this: `scipy/slsqp`. Therefore this keyword can be removed.

The keyword should be deprecated and eventually removed. The documentation should clarify how algorithms/methods are found.

N/A

**Additional context**
Implementation should be straightforward, since `ropt` already uses this approach.

yngve-sk:
**Issue**
Resolves #9773

### DIFF
--- a/src/everest/config/optimization_config.py
+++ b/src/everest/config/optimization_config.py
@@ -1,14 +1,16 @@
+import logging
 from typing import Any
 
 from pydantic import BaseModel, Field, model_validator
 
 from everest.config.cvar_config import CVaRConfig
 from everest.optimizer.utils import get_ropt_plugin_manager
+from everest.strings import EVEREST
 
 
 class OptimizationConfig(BaseModel, extra="forbid"):  # type: ignore
     algorithm: str | None = Field(
-        default="default",
+        default="optpp_q_newton",
         description="""Algorithm used by Everest.  Defaults to
 optpp_q_newton, a quasi-Newton algorithm in Dakota's OPT PP library.
 """,
@@ -33,8 +35,8 @@ iteration.
 (From the Dakota Manual.)""",
     )
     backend: str | None = Field(
-        default="dakota",
-        description="""The optimization backend used.  Defaults to "dakota".
+        default=None,
+        description="""(deprecated) The optimization backend used.".
 
 Currently, backends are included to use Dakota or SciPy ("dakota" and "scipy").
 The Dakota backend is the default, and can be assumed to be installed. The SciPy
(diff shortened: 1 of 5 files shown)


src/everest/config/optimization_config.py
```diff
@@ -8,7 +8,7 @@
 
 class OptimizationConfig(BaseModel, extra="forbid"):  # type: ignore
     algorithm: str | None = Field(
-        default="default",
+        default="dakota/default",
```
verveerpj, line 11: If that works, that's a coincidence :-) Maybe we should say `default=optpp_q_newton" so that we really fix the default of Everest to that.

src/everest/config/sampler_config.py
```diff
@@ -26,7 +26,7 @@ class SamplerConfig(BaseModel):  # type: ignore
 """,
     )
     method: str = Field(
-        default="default",
+        default="dakota/default",
```
verveerpj, line 29: I suggest not giving a default here, because samplers are not specified by default anyway in the main config. If the user does specify one, they better know what method they want to use and should specify it explicitly.

src/everest/optimizer/everest2ropt.py
```diff
@@ -258,9 +267,18 @@ def _parse_optimization(
...
+        message = (
+            "optimization.backend is deprecated. "
+            "The correct backend will be inferred by the algorithm. "
+            "If several backends have an algorithm named A and you want to pick "
+            "a specific backend B, put B/A in optimization.algorithm."
+        )
+        print(message)
+        logging.getLogger(EVEREST).warning(message)
+
+    algorithm = ever_opt.algorithm or "dakota/default"
```
verveerpj, line 280: make this `"optpp_q_newton"`

verveerpj (commented):
Looks good, but some comments.

src/everest/optimizer/everest2ropt.py
```diff
@@ -65,9 +65,18 @@ def _parse_controls(ever_controls: list[ControlConfig], ropt_config):
...
 
     if any(item >= 0 for item in controls.sampler_indices):
+        if any(sampler.backend is not None for sampler in controls.samplers):
+            message = (
+                "sampler.backend is deprecated. "
+                "The correct backend will be inferred by the method. "
+                "If several backends have a method named A and you want to pick "
+                "a specific backend B, put B/A in sampler.method."
+            )
+            print(message)
```
DanSava, line 75: Shouldn't this deprecation message come from `everest/config/sampler_config.py` ?

src/everest/optimizer/everest2ropt.py
```diff
@@ -258,9 +267,18 @@ def _parse_optimization(
...
-    algorithm = ever_opt.algorithm or "default"
-    ropt_optimizer["method"] = f"{backend}/{algorithm}"
+    if ever_opt.backend is not None:
+        message = (
+            "optimization.backend is deprecated. "
+            "The correct backend will be inferred by the algorithm. "
+            "If several backends have an algorithm named A and you want to pick "
+            "a specific backend B, put B/A in optimization.algorithm."
+        )
+        print(message)
```
DanSava, line 277: Shouldn't this be part of `everest/config/optimization_config.py`

verveerpj (approved):
LGTM!

PR #9901, opened 2025-01-29
Remove control type filed

### LINKED ISSUE #9774: Remove or repurpose the `type` field in the controls section of the Everest configuration
The `type` field in the `controls` section of the Everest configuration is not used and can be removed or re-purposed.

We should investigate if this field is indeed not used for anything meaningful. We will likely find that this field is very often specified in existing configuration files, but not used in any way. Based on this information, we can decide if we want to remove it, or repurpose it within the context of this issue: https://github.com/equinor/ert/issues/9537.

We could leave it as is, it would not affect functionality.

DanSava:
**Issue**
Resolves #9774 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/config/everest_config.py
+++ b/src/everest/config/everest_config.py
@@ -452,15 +452,18 @@ and environment variables are exposed in the form 'os.NAME', for example:
         wells = self.wells
         if controls is None or wells is None:
             return self
-        well_names = [w.name for w in wells]
+        well_names = {w.name for w in wells}
         if not well_names:
             return self
         for c in controls:
-            if c.type == "generic_control":
-                continue
-            for v in c.variables:
-                if v.name not in well_names:
-                    raise ValueError("Variable name does not match any well name")
+            variable_names = {v.name for v in c.variables}
+            if len(variable_names.intersection(well_names)) not in {
+                0,
+                len(variable_names),
+            }:
+                raise ValueError(
+                    "Variable names should either all or none match well names"
+                )
 
         return self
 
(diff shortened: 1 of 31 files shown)


src/everest/config/everest_config.py
```diff
@@ -452,15 +452,18 @@ def validate_variable_name_match_well_name(self) -> Self:  # pylint: disable=E02
...
-                if v.name not in well_names:
-                    raise ValueError("Variable name does not match any well name")
+            variable_names = {v.name for v in c.variables}
+            if len(variable_names.intersection(well_names)) not in {
+                0,
+                len(variable_names),
+            }:
+                raise ValueError(
+                    "Variable names should either all or none match well names"
+                )
```
verveerpj, line 466: So this case would not be allowed anymore: A control group where most variable names are arbitrary, but one or more happen to be well names?

If so, it seems an arbitrary limitation, somebody may have a case that breaks it.

DanSava: Yes, that will no longer be allowed. 

We could remove this arbitrary limitation but once the control type filed is removed we can no longer enforce any constraint on the well name and variable names. So as far as can see there we have the following options:

1. We keep the control type. 
2. We remove the control type and this validation 
3. We add this arbitrary limitation that might break some cases where there are `well_controls` and `generic_controls` that happen to have some of the variable names match the well names.

 There could be some other options I am not considering at the moment, but I am open to suggestions.

DanSava:
Given that the control type is used I think we can keep it for now. Closing this pull request until a consensus is reached on handling the control type validation for well controls.

PR #9915, opened 2025-01-30
Add more validation to design matrix

### LINKED ISSUE #8686: Validate that DESIGN_MATRIX is correct upon initialization (num reals vs. num params vs defaults)

jonathan-eq:
**Issue**
Resolves #8686

**Approach**
The commits in this PR add more validation to design matrix

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/design_matrix.py
+++ b/src/ert/config/design_matrix.py
@@ -191,7 +191,7 @@ class DesignMatrix:
 
         if error_list := DesignMatrix._validate_design_matrix(design_matrix_df):
             error_msg = "\n".join(error_list)
-            raise ValueError(f"Design matrix is not valid, error:\n{error_msg}")
+            raise ValueError(f"Design matrix is not valid, error(s):\n{error_msg}")
 
         defaults_to_use = DesignMatrix._read_defaultssheet(
             self.xls_filename, self.default_sheet, design_matrix_df.columns.to_list()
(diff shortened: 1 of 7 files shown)


src/ert/config/design_matrix.py
```diff
@@ -191,7 +194,7 @@ def read_design_matrix(
 
         if error_list := DesignMatrix._validate_design_matrix(design_matrix_df):
             error_msg = "\n".join(error_list)
-            raise ValueError(f"Design matrix is not valid, error:\n{error_msg}")
+            raise ValueError(f"Design matrix is not valid, error(s):\n{error_msg}")
```
xjules, line 197: :)

src/ert/gui/simulation/ensemble_experiment_panel.py
```diff
@@ -88,6 +88,9 @@ def __init__(
             self._active_realizations_field.setText(
                 ActiveRange(design_matrix.active_realizations).rangestring
             )
+            self._active_realizations_field.setValidator(
+                RangeStringArgument(design_matrix.get_num_realizations())
```
xjules, line 92: What does this do?

jonathan-eq: This would not work if the user is missing realizations in their designmatrix. Example: User only has real 0,1,5,6,9 in design matrix. This range string would not be validated correctly in this case.

xjules: Then if we don't have such a custom validator, I'd say let's just check on the actual run whether the active realization mask matches the active realizations in the field box?

jonathan-eq: If we can validate it on the actual run; then it should be a small task to have it validated before enabling the run button. 
Initial count:
image
Validator that counts incorrectly:
image

xjules: Yes, but active_mask=[0,5,9] and you set it to `0-9`?

xjules (changes requested):
Very nice PR @jonathan-eq ! Could you update the commit message that includes all the new items that it adds?

xjules (approved):
:heart_eyes: :rocket:

PR #9923, opened 2025-01-30
Refactor and update for ropt 0.11

verveerpj:
**Issue**
`ropt` has been updated to version 0.11, with a few changes that affect Everest:

- `ropt` does not store names of controls, objectives, constraints and realizations anymore, as it does not need them
- Some functions/classes have been renamed to better reflect their function
- Event handling has been made more generic, changing its interface

These changes are intended to make `ropt` simpler and give more control to Everest over some aspects (such as control name handling). In addition it aims to make Everest less dependent on `ropt` internals.

This PR does the following:
- Implement changes needed to use `ropt==0.11`.
- Remove the output tables generated by `ropt`, since that should be done by Everest.
- Remove some unused functions in the everest data api
- Remove some data from the everest storage that is not needed

      --exec 'just rapid-tests'`)

## When applicable


src/everest/config/simulator_config.py
```diff
@@ -79,17 +79,13 @@ class SimulatorConfig(BaseModel, extra="forbid"):  # type: ignore
     If not specified, a default value of 1 will be used.""",
     )
     enable_cache: bool = Field(
-        default=False,
+        default=True,
```
yngve-sk, line 82: This is a bit of a behavior change, but sensible I think, would it be possible to put it into its own commit?

src/everest/config/everest_config.py
```diff
@@ -621,6 +621,20 @@ def control_names(self):
         controls = self.controls or []
         return [control.name for control in controls]
 
+    @property
+    def formatted_control_names(self) -> list[str]:
```
yngve-sk, line 625: This is just a rename+move of `control_tuples`? own commit would be nice

src/everest/everest_storage.py
```diff
@@ -354,96 +350,82 @@ def read_from_output_dir(self) -> None:
...
-
-        config = event.config
-
-        # Note: We probably do not have to store
-        # all of this information, consider removing.
+    def init(self, everest_config: EverestConfig) -> None:
         self.data.controls = polars.DataFrame(
             {
                 "control_name": polars.Series(
-                    _format_control_names(config.variables.names), dtype=polars.String
```
yngve-sk, line 390: Similar comment/question to removal of these things that really are in the Everest Config: do we want to store the everest config used to create the results? If we don't store it, we would have optimization values but lack some info about the initial setup of the optimization. I guess this exists in ERT storage in the form of parameter values, but some wouldn't.

One option would be to maybe dump the everest config used to create the results in the experiment folder, if this is a relevant use case at all. If we assume that user's don't constantly mutate their everest config and write results into new optimization dirs.

src/everest/everest_storage.py
```diff
@@ -354,96 +350,82 @@ def read_from_output_dir(self) -> None:
...
-                ),
-                "weight": polars.Series(
-                    config.realizations.weights, dtype=polars.Float64
+                    everest_config.model.realizations, dtype=polars.UInt32
                 ),
             }
         )
 
     def _store_function_results(self, results: FunctionResults) -> _EvaluationResults:
+        names = {
```
yngve-sk, line 411: (this comment is also scoped to the same thing in `_store_gradient_results`) This ensures we select only stuff that is actually requested by Everest? Would ROPT ever generate names that are not in this list? Some kind of comment to clarify why we do this would be helpful, also own commit with descriptive comment.

src/everest/optimizer/everest2ropt.py
```diff
@@ -50,15 +48,13 @@ def _parse_controls(ever_controls: list[ControlConfig], ropt_config):
         ]
     indices = [idx for idx, is_enabled in enumerate(controls.enabled) if is_enabled]
     ropt_config["variables"] = {
-        "names": controls.names,
```
yngve-sk, line 53: Oh, this relates to the change in gradient/function results I guess? If so, would be nice to have in the same commit as those

src/everest/optimizer/everest2ropt.py
```diff
@@ -82,66 +78,59 @@ def _parse_controls(ever_controls: list[ControlConfig], ropt_config):
     ]
 
     ropt_config["gradient"]["perturbation_types"] = [
```
yngve-sk, line 80: This part looks a lot simpler, nice! Could we have this part and any related parts (if any?) in its own commit, guessing this is standalone due to a change in ROPT?

src/everest/optimizer/everest2ropt.py
```diff
@@ -82,66 +78,59 @@ def _parse_controls(ever_controls: list[ControlConfig], ropt_config):
...
+        # function estimators in ropt to implement these types. This is done by
+        # supplying a list of estimators and for each objective an index into
         # that list:
         objective_type = objective.type
         if objective_type is None:
             objective_type = "mean"
-        # Find the transform if it exists:
-        transform_idx = next(
+        # Find the estimator if it exists:
+        function_estimator_idx = next(
```
yngve-sk, line 107: Won't `function_estimators` always be empty at this point, so the code is always operating on an empty list or am I missing something?

src/everest/optimizer/everest2ropt.py
```diff
@@ -82,66 +78,59 @@ def _parse_controls(ever_controls: list[ControlConfig], ropt_config):
...
+                if estimator["method"] == objective_type
             ),
             None,
         )
-        # If not, make a new transform:
-        if transform_idx is None:
-            transform_idx = len(transforms)
-            transforms.append({"method": objective_type})
-        transform_indices.append(transform_idx)
+        # If not, make a new estimator:
```
yngve-sk, line 115: Unrelated to this PR but I think this code is a bit too "for-loop"-ey for what it does: create set/list a unique objective types and an index list pointing to them per objective^^

src/everest/optimizer/everest2ropt.py
```diff
@@ -336,9 +322,9 @@ def _parse_optimization(
             # indices to any realization filters that should be applied. In this
             # case, we want all objectives and constraints to refer to the same
             # filter implementing cvar:
-            objective_count = len(ropt_config["objectives"]["names"])
+            objective_count = len(ropt_config["objectives"]["weights"])
```
yngve-sk, line 325: Should go in the stop giving control names to ROPT commit? Similar for the rhs_values change below

src/everest/config/simulator_config.py
verveerpj: Yes, will do. The behavior change is quite minimal, normal use cases will not use the cache anyway. (auto scaling may activate it)

src/everest/config/everest_config.py
verveerpj: No, this formats the control tuples in a standard way.

src/everest/everest_storage.py
verveerpj: `names `just collects the names of controls, objectives etc., so that the generated dataframe contains them, instead of an index. It might be possible/better to add the names to the dataframe afterwards, by replacing indices with names. Then we don't have to depend on `ropt` for that. Do you think that would be doable?

src/everest/optimizer/everest2ropt.py
verveerpj: Yes, they belong together.

src/everest/optimizer/everest2ropt.py
verveerpj: Maybe, will check.

src/everest/optimizer/everest2ropt.py
verveerpj: Yes, the `everest2ropt` code is not an example of python beauty... I have wanted to refactor that, but never got around to it...

src/everest/everest_storage.py
yngve-sk: I'm not fully sure I understand this, so having the names in the dataframe instead of an index? So adding them kind of does the same as we now do with the `.reset_index()`? How would we know where to add which names and so on?

verveerpj: I am doing it right now in an extra commit, so if that works I will push it and you will probably understand.

src/everest/optimizer/everest2ropt.py
verveerpj: Not really worth doing it in one commit, it requires the upgrade `ropt` 0.11, so if I split this up, the first commit will not come through the tests. Keeping them together.

src/everest/everest_storage.py
```diff
@@ -592,6 +617,25 @@ def _store_gradient_results(self, results: GradientResults) -> _GradientResults:
                     ),
                 ]
             )
+            batch_objective_gradient = batch_objective_gradient.with_columns(
```
yngve-sk, line 620: I think some kind of comment somewhere on why we need to do this would be good, it is not apparent that ropt operates only with the indices, and that we have to add them back

verveerpj: Added in the new `_ropt_to_df` method.

src/everest/everest_storage.py
```diff
@@ -328,6 +324,55 @@ def _enforce_dtypes(df: polars.DataFrame) -> polars.DataFrame:
...
+    ) -> polars.DataFrame:
+        df = polars.from_pandas(
+            results.to_dataframe(field, select=values).reset_index(),
+        ).select(select + values)
+
+        # The results from ropt do not contain any names, but indices referring
+        # control names, objective names, etc. The corresponding names can be
+        # retrieved from the everest configuration and were stored in the init
+        # method. Here we replace the indices with those names:
+        replacements = []
```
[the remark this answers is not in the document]
verveerpj: done.

yngve-sk (approved):
Nice! LGTM

Issue #9670, opened 2025-01-08
flaky tests/everest/entry_points/test_config_branch_entry.py::test_config_branch_preserves_config_section_order


eivindjahren:
https://github.com/equinor/ert/actions/runs/12647698686/job/35240588690

The problem seems to be the use of pytest.cache which seemingly can create a race codition.

PR #9928, opened 2025-01-31
Fix everest fixture for caching testcase results

jonathan-eq:
**Issue**
Resolves [#9670](https://github.com/equinor/ert/issues/9670)

**Approach**
pytest-snapshot does not go well with pytest-xdist, so we have to manually set the snapshot path to avoid it using the worker-ids in the path. 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/everest/conftest.py
+++ b/tests/everest/conftest.py
@@ -1,6 +1,7 @@
 import os
 import shutil
 import tempfile
+import time
 import warnings
 from collections.abc import Callable, Iterator
 from copy import deepcopy
@@ -145,11 +146,17 @@ def change_to_tmpdir(tmp_path, monkeypatch):
 
 
 @pytest.fixture
-def cached_example(pytestconfig):
+def cached_example(pytestconfig: pytest.Config):
     cache = pytestconfig.cache
 
     def run_config(test_data_case: str):
-        if cache.get(f"cached_example:{test_data_case}", None) is None:
+        if (
+            cache.get(
+                f"cached_example:{test_data_case}",
+                None,
+            )
+            is None
+        ):
             my_tmpdir = Path(tempfile.mkdtemp())
             config_path = (
                 Path(__file__) / f"../../../test-data/everest/{test_data_case}"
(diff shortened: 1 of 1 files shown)


tests/everest/test_api_snapshots.py
```diff
@@ -74,6 +74,9 @@ def test_api_snapshots(config_file, snapshot, cached_example):
         .strip()
         + "\n"
     )
+    snapshot.snapshot_dir = (
```
yngve-sk, line 77: Is this part necessary? The snapshots are currently already scoped/split without this?

Screenshot 2025-02-05 at 14 43 50

tests/everest/conftest.py
```diff
@@ -157,11 +157,17 @@ def create_evaluator_server_config(run_model):
 
 
 @pytest.fixture
-def cached_example(pytestconfig, evaluator_server_config_generator):
+def cached_example(pytestconfig: pytest.Config, evaluator_server_config_generator):
     cache = pytestconfig.cache
 
     def run_config(test_data_case: str):
-        if cache.get(f"cached_example:{test_data_case}", None) is None:
+        if (
```
yngve-sk, line 164: I think this makes the everest tests take 16m to run (as opposed to currently 6-7mins) as it now has to create one set of complete runs per runner, in this case running sequential might actually be faster/cheaper than parallel. So it would be ideal to be able to share this cache thing between runners and look for another way to avoid the collision. I don't think I understand the problem this solves in full detail, nor whether this is the right way to go about it.

tests/everest/test_api_snapshots.py
jonathan-eq: You are right, it was `pytest-repeat` that changed the snapshot names

tests/everest/conftest.py
eivindjahren: The way this is solved in ert is by using `--dist loadgroup` to put all workflows that require the cached testdata run in the same process.

jonathan-eq:
Superseded by #10041

PR #9942, opened 2025-02-03
Use only dots to delimit controls

### LINKED ISSUE #9816: Make control naming consistent
The variables that are optimized (controls in Everest nomenclature) are specified in a nested fashion: We have controls, which defines groups of variables, which in turn can have multiple components labeled with an integer index. For example (from `config_advanced.yml`):

```yml
controls:
  - name: point
    max: 1.0
    min: -1.0
    initial_guess: 0.25
    perturbation_magnitude: 0.005
    type: generic_control
    variables:
      - name: x
        index: 0
      - name: x
        index: 1
      - name: x
        index: 2
```
 This defines three controls, with the following `control`, `variable`, and `index` fields:

- `point`, `x`, `0`
- `point`, `x`, `1`
- `point`, `x`, `2`

When reporting results, control names are constructed from the three fields as follows `control_variable-index`, so we get:
- `point_x-0`
- `point_x-1`
- `point_x-2`

When specifying input constraints, weights need to be set for a sub-set of controls, so controls are needed, but in this case they follow the form `control.variable-index`. For instance (from `config_advanced.yml`):

```yml
input_constraints:
  - weights:
      point.x-0: 0
      point.x-1: 0
      point.x-2: 1
    upper_bound: 0.4
```

This is inconsistent and potentially confusing for the user.

When a string representation of controls is shown, or required as an input, it should always have the same form. There should be a singe function/method responsible for formatting the string from `control`, `variable`, `index` fields. Preferably this transformation should be invertible, although we should not rely on that in code. If possible, the delimiters used should not be allowed in control and variable names. We should discuss the delimiters to use, but using `.` and `-` would be backwards compatible for specifying input constraints and it would leave the underscore available for control and variable naming.

yngve-sk:
**Issue**
Resolves #9816 

**Approach**
Keep backward compatibility (should be deprecated?) but accept only `.` notation to delimit control names, variable/index etc


src/everest/optimizer/everest2ropt.py
```diff
@@ -107,22 +107,21 @@ def _parse_objectives(objective_functions: list[ObjectiveFunctionConfig], ropt_c
...
-        for control_name in controls.names
-    ]
+    def _get_control_index(name: str):
+        try:
+            matching_index = formatted_control_names.index(name.replace("-", "."))
+            return matching_index
+        except ValueError:
+            pass
+
+        return formatted_control_names_dotdash.index(name)
```
verveerpj, line 124: Eventually this should disappear, since using the dash is deprecated. Maybe a comment?

verveerpj (approved):
LGTM!

verveerpj (approved):
LGTM!

PR #9943, opened 2025-02-03
Correctly calculate cpu_seconds for processtree

### LINKED ISSUE #9941: Logging of cpu-seconds will miss subprocesses that exit before the main process
The polling nature of cpu-seconds counting in

https://github.com/equinor/ert/blob/7c27a9cc3011a7dcd87e7d9242625811a6961949/src/_ert/forward_model_runner/forward_model_step.py#L206

will underestimate the cpu_seconds if the main process for the forward model step launches subprocesses that end earlier than the main process, as the process tree for a process that has ended does not exist in the kernel any longer.

berland:
If a subprocess in the processtree terminates early, its cpu_seconds value will be lost unless we keep all values pr. pid until the end.

Since pids can be reused (although with little probability) we need to detect this through a decreasing cpu_second value for a pid that has already been observed in the processtree.

The test has been modified in order to trigger this scenario.

**Issue**
Resolves #9941 

**Approach**
🗒️ Do the book-keeping necessary.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/_ert/forward_model_runner/forward_model_step.py
+++ b/src/_ert/forward_model_runner/forward_model_step.py
@@ -8,6 +8,7 @@ import signal
 import socket
 import sys
 import time
+import uuid
 from collections.abc import Generator, Sequence
 from datetime import datetime as dt
 from pathlib import Path
@@ -186,8 +187,8 @@ class ForwardModelStep:
         exit_code = None
 
         max_memory_usage = 0
-        max_cpu_seconds = 0
         fm_step_pids = {int(process.pid)}
+        cpu_seconds_pr_pid: dict[str, float] = {}
         while True:
             try:
                 exit_code = process.wait(timeout=self.MEMORY_POLL_PERIOD)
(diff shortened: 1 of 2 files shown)


src/_ert/forward_model_runner/forward_model_step.py
```diff
@@ -203,8 +204,19 @@ def _run(self) -> Generator[Start | Exited | Running]:
                     yield potential_exited_msg
                     return
 
-            (memory_rss, cpu_seconds, oom_score, pids) = _get_processtree_data(process)
-            max_cpu_seconds = max(max_cpu_seconds, cpu_seconds or 0)
+            (memory_rss, cpu_seconds_snapshot, oom_score, pids) = _get_processtree_data(
+                process
+            )
+            for pid, seconds in cpu_seconds_snapshot.items():
```
eivindjahren, line 210: You could extract this part to make it easier to test:
```python
@dataclass
class CpuTimer:
  _cpu_seconds_pr_pid: dict[str, float] = field(default_factory=dict, init=False)

  def update(self, cpu_seconds_snapshot):
    ...

  def total_cpu_seconds(self):
    return sum(self._cpu_seconds_pr_pid.values())
```

Then you can easily test the logic by giving it fake `cpu_seconds_snapshots`.

eivindjahren (approved):
I had a small suggestion about improving the testability of the logic, but I see no need to block merging. The suggested refactoring and adding of testing can be done in a separate PR.

PR #9948, opened 2025-02-03
Log pending time for a job

### LINKED ISSUE #9340: Log pending time for a realization
A nice-to-have log statistic is the time spent Pending in the cluster.

This can be accomplished by a log statement in `src/ert/scheduler/job.py`. 

It should include `self.real.num_cpu` as that is a parameter that probably affects the time spent pending.

berland:
**Issue**
Resolves #9340 

**Approach**
🗒️ 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -109,6 +109,7 @@ class Job:
             if self._scheduler.submit_sleep_state:
                 await self._scheduler.submit_sleep_state.sleep_until_we_can_submit()
             await self._send(JobState.SUBMITTING)
+            submit_time = time.time()
             try:
                 await self.driver.submit(
                     self.real.iens,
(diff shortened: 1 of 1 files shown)


berland:
Open questions:
* Should we log num_cpu and realization_memory as well as done currently in the PR? These are mostly constant for each realization, and can probably also be fetched through the span id.
* If yes to the above, should we include SUBMIT_SLEEP as well?

@andreas-el , you are probably opinionated on how much hassle it is to fetch the data via the span id if we don't include it in each log message.

eivindjahren (approved):
I'll let @andreas-el answer the follow up questions, but this seems ready to go in.

andreas-el:
> Open questions:
> 
> * Should we log num_cpu and realization_memory as well as done currently in the PR? These are mostly constant for each realization, and can probably also be fetched through the span id.
> * If yes to the above, should we include SUBMIT_SLEEP as well?
> 
> @andreas-el , you are probably opinionated on how much hassle it is to fetch the data via the span id if we don't include it in each log message.

I can't see why this won't work nicely. We already look up all kinds of data using the span-id. Go ahead!

PR #9962, opened 2025-02-05
Replace _wait_until with wait_until

xjules:
**Issue**
Utility function _wait_until is used only in a single test function, ~~so moving it there.~~ replacing it with another (similar) one.

**Approach**
:car: 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/forward_model_runner/test_fm_dispatch.py
+++ b/tests/ert/unit_tests/forward_model_runner/test_fm_dispatch.py
@@ -25,15 +25,9 @@ from _ert.forward_model_runner.fm_dispatch import (
 )
 from _ert.forward_model_runner.forward_model_step import killed_by_oom
 from _ert.forward_model_runner.reporting import Event, Interactive, Reporter
-from _ert.forward_model_runner.reporting.message import (
-    Finish,
-    Init,
-    Message,
-)
+from _ert.forward_model_runner.reporting.message import Finish, Init, Message
 from _ert.threading import ErtThread
 from tests.ert.utils import MockZMQServer, wait_until
-
-from .test_event_reporter import _wait_until
 
 
 @pytest.mark.integration_test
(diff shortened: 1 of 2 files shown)


tests/ert/unit_tests/forward_model_runner/test_fm_dispatch.py
```diff
@@ -312,6 +307,12 @@ def test_retry_of_jobs_json_file_read(unused_tcp_port, tmp_path, monkeypatch, ca
         }
     )
 
+    def _wait_until(condition, timeout, fail_msg):
+        start = time.time()
+        while not condition():
+            assert start + timeout > time.time(), fail_msg
+            time.sleep(0.1)
+
```
jonathan-eq, line 315: If it is only used one place, do we need it as a separate function?

xjules: Maybe. Shall I refactor then?

jonathan-eq: I think that would be good yes. `while not condition()` does not read too well without any typing.

xjules: Actually there is another `wait_until` function. I will use that one.

DanSava (approved):
Looks good to me! :rocket:

tests/ert/unit_tests/forward_model_runner/test_fm_dispatch.py
```diff
@@ -313,11 +307,10 @@ def test_retry_of_jobs_json_file_read(unused_tcp_port, tmp_path, monkeypatch, ca
     )
 
     def create_jobs_file_after_lock():
-        _wait_until(
-            lambda: f"Could not find file {FORWARD_MODEL_DESCRIPTION_FILE}, retrying"
-            in caplog.text,
-            2,
-            f"Did not get expected log message from missing {FORWARD_MODEL_DESCRIPTION_FILE}",
+        wait_until(
+            lambda: f"Could not find file {FORWARD_MODEL_DESCRIPTION_FILE}, retrying",
```
jonathan-eq, line 311: I think this lambda is missing `in caplog.txt`, so it is actually not doing anything 🤔

PR #9965, opened 2025-02-05
Replace normalization and auto_normalize keyes with scaling and auto_scale in ObjectiveFunctionConfig

### LINKED ISSUE #9770: Add a scaling option for objectives
In the Everest configuration a normalization factor can be specified. Objectives will be normalized by multiplying with this factor during optimization. Instead of a normalization factor a scaling factor should be used, which divides the objectives. This is more natural, since we tend want to reduce large objective values to a value close to one. It is also more inline with controls and constraints that offer a scaling option.

A scaling option should be added which could be used instead of normalization. The normalization option should be deprecated and in the long run, it should be removed.

N/A

**Additional context**
Internally the `ropt` optimizer already expects scaling rather than normalization, hence the everest code relating to setting up the optimization should become slightly clearer.

DanSava:
**Issue**
Resolves #9770 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable


verveerpj (commented):
The idea is to change from a multiplication to a division factor. This is because we tend to have large objectives and then it is easier to specify the expected maximum value than its inverse. Division is also consistent with the scaling we use with constraints. So, in doc strings multiplication should be replaced with division, and in `everest2ropt` the `1.0 / objective.scaling` should be replaced by `objective.scaling`

Otherwise looks good.

DanSava:
> > Division is also consistent with the scaling we use with constraints. So, in doc strings multiplication should be replaced with division, and in `everest2ropt` the `1.0 / objective.scaling` should be replaced by `objective.scaling`
> 
> I am guessing currently ropt is not designed to handled ropt_config["objectives"]["scales"] = [ 1.0 / objective.scaling] because if I add this change then the output looks like this for `config_multiobj.yml`
> 
> ```
>  "single_objective_values": ['
>    {'
>       "accepted": 1,\n'
>       "batch": 0,\n'
>       "distance_p": -0.33333333,\n'
>       "distance_q": -1.58333333,\n'
>       "objective": -2.33333333\n'
>     },
> ....
>   ]
> ```
> 
> where previously it was
> 
> ```
>  "single_objective_values": ['
>    {'
>       "accepted": 1,
>       "batch": 0,
>       "distance_p": -0.75,
>       "distance_q": -1.58333333,
>       "objective": -2.33333333
>     },
> ....
>   ]
> ```
> 
> which seems a bit more consistent because `-0.75 - 1.5833333 = -2.33333333` and `-0.33333333 - 1.5833333 != -2.33333333`
> 
> So I am a bit confused how both of these can be correct.

Seems there was a place in the everest_data_api where the single objective values where calculated and I had also to update the formulate there to divide by the scaling value and not multiply with the normalization.

verveerpj:
> Seems there was a place in the everest_data_api where the single objective values where calculated and I had also to update the formulate there to divide by the scaling value and not multiply with the normalization.

That seems correct.

verveerpj (approved):
LGTM!

PR #9985, opened 2025-02-05
Reimplement objective scaling

verveerpj:
**Issue**
This PR reimplements objective scaling using the new approach of `ropt `to transform variables and objectives between the  model and optimization domains.

**Approach**
There are three commits:
1. A small commit that uses an improvement in `ropt `to simplify the storage code.
2. A refactoring of the forward model evaluator in the `everest `run model to introduce a clearer boundary between `ropt `and `everest `code.
3. The new implementation of objective scaling.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/run_models/everest_run_model.py
+++ b/src/ert/run_models/everest_run_model.py
@@ -250,19 +250,28 @@ class EverestRunModel(BaseRunModel):
             logging.getLogger(EVEREST).info("User abort requested.")
             optimizer.abort_optimization()
 
-    def _forward_model_evaluator(
-        self, control_values: NDArray[np.float64], evaluator_context: EvaluatorContext
-    ) -> EvaluatorResult:
+    def _run_forward_model(
+        self,
+        control_values: NDArray[np.float64],
+        realizations: list[int],
+        active_control_vectors: list[bool] | None = None,
+    ) -> tuple[NDArray[np.float64], NDArray[np.float64] | None, list[int]]:
         # Reset the current run status:
         self._status = None
 
         # Get cached_results:
-        cached_results = self._get_cached_results(control_values, evaluator_context)
+        cached_results = self._get_cached_results(control_values, realizations)
+
+        # Collect the indices of the controls that must be evaluated in the batch:
+        evaluated_control_indices = [
+            idx
+            for idx in range(control_values.shape[0])
+            if idx not in cached_results
+            and (active_control_vectors is None or active_control_vectors[idx])
+        ]
 
         # Create the batch to run:
-        batch_data = self._init_batch_data(
-            control_values, evaluator_context, cached_results
-        )
+        batch_data = self._init_batch_data(control_values, evaluated_control_indices)
 
         # Initialize a new ensemble in storage:
         assert self._experiment is not None
@@ -274,7 +283,7 @@ class EverestRunModel(BaseRunModel):
             self._setup_sim(sim_id, controls, ensemble)
 
         # Evaluate the batch:
-        run_args = self._get_run_args(ensemble, evaluator_context, batch_data)
+        run_args = self._get_run_args(ensemble, realizations, batch_data)
         self._context_env.update(
             {
                 "_ERT_EXPERIMENT_ID": str(ensemble.experiment_id),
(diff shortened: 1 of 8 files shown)


src/ert/run_models/everest_run_model.py
```diff
@@ -290,33 +322,53 @@ def _forward_model_evaluator(
...
-            evaluator_result.constraints,
+            control_values, realizations, batch_data, objectives, constraints
         )
 
         # Increase the batch ID for the next evaluation:
         self._batch_id += 1
 
-        return evaluator_result
+        sim_ids = np.full(control_values.shape[0], -1, dtype=np.intc)
+        sim_ids[list(batch_data.keys())] = np.arange(len(batch_data), dtype=np.intc)
```
yngve-sk, line 338: I found this sim_ids generation a bit hard to follow. Could you double check this logic and write a comment on the intent? I'm running this for `math_func_advanced` and getting `sim_ids` evaluation to basically just `np.arange(len(batch_data), dtype=np.intc)`. I think also maybe having `batch_data` a bit more clearly defined by name will make this a bit more clear

ref
```
sim_ids = np.full(control_values.shape[0], -1, dtype=np.intc)
sim_ids
Out[22]: 
array([-1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1],
      dtype=int32)
list(batch_data.keys())
Out[23]: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
np.arange(len(batch_data), dtype=np.intc)
Out[24]: 
array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15],
      dtype=int32)
sim_ids[list(batch_data.keys())] = np.arange(len(batch_data), dtype=np.intc)
sim_ids
Out[26]: 
array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15],
      dtype=int32)
```

verveerpj: The `sim_id` are an attempt to track the simulations that you find on disk which are folders called `simulation_#`. So, basically the mapping of the control vector onto the ERT realization, I guess? Normally this would just be a range with a length equal to the number of simulations, i.e. the length of `batch_data`. Normally, that is the number of control vectors, but we may skip control vectors because they may be not active or from cache. So, those do not have a simulation, but the simulation numbers stay continuous. So we have to skip control vectors that were not evaluate, giving them an id of -1, and number the rest consecutively.

Its all messy, and the only reason for this is that we need somehow to report which simulation belongs to a certain control vector. So the simulation number were added to the results generated by `ropt` and reported in the tables that I was writing, so our users could find the simulations back on disk. 

I would like to get rid of it entirely, but we are going to need a way to identify match optimization results with results on disk, so I think we need to think about this at some point...

In the `ropt` results they are called `evaluation_id`. I think you use them also in the storage?

src/ert/run_models/everest_run_model.py
```diff
@@ -290,33 +322,53 @@ def _forward_model_evaluator(
...
+        return objectives, constraints, sim_ids
 
-    def _get_cached_results(
+    def _forward_model_evaluator(
```
yngve-sk, line 341: Pulling out the forward model run logic into a separate  `_run_forward_model` function made this a lot easier to follow.

src/ert/run_models/everest_run_model.py
```diff
@@ -348,15 +400,11 @@ def _add_controls(
                 batch_data_item[control.name] = control_dict
             return batch_data_item
 
-        active = evaluator_context.active
-        realizations = evaluator_context.realizations
         return {
             idx: _add_controls(self._everest_config.controls, control_values[idx, :])
             for idx in range(control_values.shape[0])
```
yngve-sk, line 405: Similar to sim id computation logic, assuming this sort of already holds the sim IDs computed here:

```
        sim_ids = np.full(control_values.shape[0], -1, dtype=np.intc)
        sim_ids[list(batch_data.keys())] = np.arange(len(batch_data), dtype=np.intc)
```
somehow I think this (the copypaste'd part above) can be made a bit more clear/less redundant by reusing some info in the line this comment is is on.

src/ert/run_models/everest_run_model.py
```diff
@@ -483,15 +527,15 @@ def _gather_simulation_results(
...
         self,
         control_values: NDArray[np.float64],
         batch_data: dict[int, Any],
         results: list[dict[str, NDArray[np.float64]]],
         cached_results: dict[int, Any],
-    ) -> EvaluatorResult:
+    ) -> tuple[NDArray[np.float64], NDArray[np.float64] | None]:
         # We minimize the negative of the objectives:
-        objectives = -self._get_simulation_results(
+        objectives = self._get_simulation_results(
```
yngve-sk, line 538: Removal of the negation but not the comment above. If it is a change / moving of negation I think it deserves its own commit or at least a comment to clarify

src/everest/optimizer/opt_model_transforms.py
```diff
@@ -48,7 +48,42 @@ def transform_linear_constraints(
...
+            np.nansum(objectives * self._weights[:, np.newaxis], axis=0)
+        )
+        self._scales[self._auto_scales] *= auto_scales[self._auto_scales]
+
+    @property
+    def has_auto_scale(self) -> bool:
+        return bool(np.any(self._auto_scales))
+
+
+def get_opt_model_transforms(
```
yngve-sk, line 82: Now that it is in Everest is there perhaps a more descriptive name than opt_model_transforms? looks almost like ropt_model_transforms without the R 😅 It is in everest terms, scaling of objectives&constraints from "forward model space" to "optimization space"?

src/ert/run_models/everest_run_model.py
```diff
@@ -290,33 +322,53 @@ def _forward_model_evaluator(
...
+        realizations = [
+            self._everest_config.model.realizations[evaluator_context.realizations[idx]]
+            for idx in range(control_values.shape[0])
+        ]
+        active_control_vectors = [
+            evaluator_context.active is None
+            or bool(evaluator_context.active[evaluator_context.realizations[idx]])
+            for idx in range(control_values.shape[0])
+        ]
+        batch_id = self._batch_id  # Save the batch ID, it will be modified.
```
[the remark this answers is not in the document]
verveerpj: Passing back the batch ID is a similar issue as the simulations id's. It is not strictly necessary but since `ropt` was doing some reporting, it was necessary. It is passed back via the results, and I think the storage uses that? This is something we are going to need to fix in the future.

src/ert/run_models/everest_run_model.py
verveerpj: Yes, I think the sim_id's could be generated here, since they are basically indexes into the batch data. I will look into that.

src/ert/run_models/everest_run_model.py
verveerpj: I will see if I can split this in a separate commit.

tests/everest/snapshots/test_ropt_initialization/test_everest2ropt_snapshot/config_multiobj.yml/ropt_config.json
```diff
@@ -42,7 +42,7 @@
     ],
     "realization_filters": null,
     "scales": [
-      0.6666666667,
+      1.0,
```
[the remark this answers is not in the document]
verveerpj: Yes, this is not used anymore, so now we just set it to 1.0. In a future version of `ropt` this field will disappear.

src/everest/optimizer/opt_model_transforms.py
verveerpj: Indeed, I did in fact mistype it a few times, I will think of something else.

verveerpj: `get_optimization_domain_transforms`

src/ert/run_models/everest_run_model.py
```diff
@@ -290,33 +328,68 @@ def _forward_model_evaluator(
...
+        objectives, constraints, evaluated_control_indices = self._run_forward_model(
+            control_values, realizations, active_control_vectors
+        )
+
+        # The simulation id's are a simple enumeration over the evaluated
+        # forward models. For the evaluated controls they are therefore
+        # implicitly given by there position in the evaluated_control_indices
+        # list. We store for each control vector that id, or -1 if it was not
+        # evaluated:
+        sim_ids = np.fromiter(
```
yngve-sk, line 368: Hmm, it is more understandable I think, and could be revisited a few more times. In the long run we could maybe get rid of the -1 as evaluation id, maybe just use the ERT realization instead of it, since all evaluations are ultimately stored in ERT storage as a realization. This is future work either way

verveerpj: We are going to re-evaluate these ID's anyway. These ID's, and the batch-ID are passed, so you can map a result produced by `ropt` to a runpath on disk. So, if you have the batch and the model realizations, you look for the folder with the right `simulation_#` name, where # is the simulation ID. That is a bit primitive and cumbersome, we should think about improving this, and indeed ultimately there is some mapping to a location in ERT storage.

src/ert/run_models/everest_run_model.py
```diff
@@ -290,33 +328,68 @@ def _forward_model_evaluator(
...
+        # Return the results, together with the indices of the evaluated controls:
+        return objectives, constraints, evaluated_control_indices
 
-    def _get_cached_results(
+    def _forward_model_evaluator(
         self, control_values: NDArray[np.float64], evaluator_context: EvaluatorContext
+    ) -> EvaluatorResult:
+        realizations = [
+            self._everest_config.model.realizations[evaluator_context.realizations[idx]]
+            for idx in range(control_values.shape[0])
```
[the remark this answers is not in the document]
verveerpj: done.

PR #10015, opened 2025-02-07
Some refactorings to EverestRunModel

yngve-sk:


verveerpj (approved):
Looks like a good improvement!

PR #10025, opened 2025-02-10
Add snapshot event for STOP_LONG_RUNNING

### LINKED ISSUE #10016: Realizations killed by STOP_LONG_RUNNING are not properly cancelled
When a realization is killed by STOP_LONG_RUNNING, the GUI will not flag the realization as failed, but in the detailed view for steps, the step is displayed as still running and it is not possible to obtain the logged info message telling why the realization was killed.

This should be made similar to the GUI experience when a realization is killed to due MAX_RUNTIME.

berland:
This makes the effect of STOP_LONG_RUNNING visible in the GUI.

**Issue**
Resolves #10016

**Approach**
Replicate the behaviour and code for MAX_RUNTIME.

stop_long_running

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/snapshot.py
+++ b/src/ert/ensemble_evaluator/snapshot.py
@@ -28,6 +28,7 @@ from _ert.events import (
     RealizationPending,
     RealizationResubmit,
     RealizationRunning,
+    RealizationStoppedLongRunning,
     RealizationSuccess,
     RealizationTimeout,
     RealizationUnknown,
@@ -51,6 +52,7 @@ _FM_TYPE_EVENT_TO_STATUS = {
     RealizationSuccess: state.REALIZATION_STATE_FINISHED,
     RealizationUnknown: state.REALIZATION_STATE_UNKNOWN,
     RealizationTimeout: state.REALIZATION_STATE_FAILED,
+    RealizationStoppedLongRunning: state.REALIZATION_STATE_FAILED,
     RealizationResubmit: state.REALIZATION_STATE_WAITING,  # For consistency since realization will turn to waiting state when resubmitted
     ForwardModelStepStart: state.FORWARD_MODEL_STATE_START,
     ForwardModelStepRunning: state.FORWARD_MODEL_STATE_RUNNING,
@@ -285,6 +287,7 @@ class EnsembleSnapshot:
                 RealizationSuccess,
                 RealizationFailed,
                 RealizationTimeout,
+                RealizationStoppedLongRunning,
             }:
                 end_time = convert_iso8601_to_datetime(timestamp)
             if type(event) is RealizationFailed:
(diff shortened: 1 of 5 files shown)


berland:
Screenshot is outdated with respect to the actual text (check source code).

Also the screenshot displays a wrong Duration. This has also been fixed in the PR.

src/ert/scheduler/scheduler.py
```diff
@@ -142,12 +148,17 @@ async def _stop_long_running_jobs(
                         > long_running_factor * self._average_job_runtime
                         and not task.done()
                     ):
-                        logger.info(
+                        logger.error(
```
xjules, line 151: Is it considered an error? I think that this was "asked" by the user when specifying `STOP_LONG_RUNNING`

tests/ert/unit_tests/scheduler/test_scheduler.py
```diff
@@ -475,8 +482,26 @@ async def wait(iens):
     )
 
     assert await sch.execute(min_required_realizations=5) == Id.ENSEMBLE_SUCCEEDED
+
+    stop_long_running_events_found = 0
+    while not sch._events.empty():
+        event = await sch._events.get()
+        print(event)
```
xjules, line 489: leftover print statement

src/ert/scheduler/scheduler.py
berland: Good question. From the perspective of the scheduler, it is not an error, I agree. From the perspective of the realization (or  job.py), it is an error. For, reasons, the corresponding logging for max_runtime is logged as an error, but this happens inside job.py which you could argue has a different perspective (the realization). 

It makes sense to have the same logleven for stop_long_running and max_runtime, and whether it is logged from scheduler.py or job.py does not matter for the user. 

Maybe warning is a fair compromise?

berland: I have changed both to warning.

tests/ert/unit_tests/scheduler/test_scheduler.py
berland: fixed.

src/ert/scheduler/scheduler.py
xjules: warning sounds good!

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -315,6 +318,24 @@ def update_from_event(
                                 "reaching MAX_RUNTIME",
                             )
                         )
+            elif e_type is RealizationStoppedLongRunning:
```
eivindjahren, line 321: Separate issue but this long list seems like a good candidate for a match-case:

```python
match event:
  ...
  case RealizationStoppedLongRunning(real=real):
    for fm_step_id, fm_step in source_snapshot.get_fm_steps_for_real(real).items():
      if fm_step.get(ids.STATUS) != state.FORWARD_MODEL_STATE_FINISHED:
          fm_idx = (real, fm_step_id)
          if fm_idx not in source_snapshot._fm_step_snapshots:
              self._fm_step_snapshots[fm_idx] = FMStepSnapshot()
          self._fm_step_snapshots[fm_idx].update(
              FMStepSnapshot(
                  status=state.FORWARD_MODEL_STATE_FAILURE,
                  end_time=end_time,
                  error="The run is cancelled due to "
                  "excessive runtime, 25% more than the average "
                  "runtime (check keyword STOP_LONG_RUNNING)",
              )
          )
```

Issue #9381, opened 2024-11-28
Create popup window for warnings post simulation


berland:
There are some situations that warrant a popup window to get the users attention after a run is finished, these are unfortunate situations that affect compute cluster stability which we are not able to detect/validate upfront. 

* Misconfigured NUM_CPU, this currently only logs a warning.
* Too high memory usage compared to REALIZATION_MEMORY.  The maximum is only shown as a number at the bottom of the GUI currently, but it is not related to the configured REALIZATION_MEMORY.
 
See:
* https://github.com/equinor/fmu-ops/issues/14
* https://github.com/equinor/fmu-ops/issues/16
* https://github.com/equinor/ert/issues/8674
* https://github.com/equinor/ert/issues/9590
* https://github.com/equinor/ert/issues/10039

PR #10050, opened 2025-02-12
Extract and log warnings from forward model steps

berland:
Each individual forward model step is run in a separate subprocess and has no relation to Erts root logger. Thus, in order to log warnings, there is no other way than parsing the stdout and stderr files for anything that looks interesting.

It is assumed that each realization will trigger approximately the same set of warnings, so the first realization to be finished will get the honour of exposing its warnings.

**Issue**
Resolves #

      --exec 'just rapid-tests'`)

## When applicable


src/ert/scheduler/job.py
```diff
@@ -349,3 +353,44 @@ def log_info_from_exit_file(exit_file_path: Path) -> None:
...
+
+    This is not a critical task to perform, but it is critical not to crash
+    during this process, thus any Exception is suppressed."""
+
+    def line_contains_warning(line: str) -> bool:
+        return (
+            "Warning:" in line
+            or "FutureWarning" in line
+            or "DeprecationWarning" in line
+            or "WARNING" in line
```
jonathan-eq, line 370:
```suggestion
            "Warning" in line
            or "warning" in line
```
Would just checking this be sufficient, or would we pick up unwanted things?

berland: That is an open question for now :) . Probably no and yes.

berland: I will add some test code that makes it more specific what we want to include and not, based on what I can find in real life examples on scratch.

eivindjahren: Potentially we just want FutureWarnings first. Then we can add more later. The most permissive would be `re.search('warning', line, re.IGNORECASE)`.

berland: In one example runpath..
```
$ grep -i warning *stdout* *stderr* | wc -l
15587
```
(of which 7000 are unique).  I think we must be more specific.

src/ert/scheduler/job.py
```diff
@@ -349,3 +354,53 @@ def log_info_from_exit_file(exit_file_path: Path) -> None:
...
+    during this process."""
+
+    max_length = 2048  # Lines will be truncated in length when logged
+
+    def line_contains_warning(line: str) -> bool:
+        return (
+            "Warning:" in line
+            or "FutureWarning" in line
+            or "DeprecationWarning" in line
+            or "UserWarning" in line
```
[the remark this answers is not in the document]
berland: It depends on our purpose, is this only for getting Future/DeprecationWarnings or do we also want to make a "service" to the users to be able to show warnings in the GUI at a later stage? (https://github.com/equinor/ert/issues/9381 )

eivindjahren (approved):

PR #10052, opened 2025-02-12
Update to ropt 0.14

verveerpj:
**Issue**
This updates Everest to use `ropt` 0.14. That version has significant enhancements on how it handles constraints. In particular, Everest does not need to split two-sided constraints (with both lower and upper bounds) in two separate constraints anymore. This simplifies the corresponding Everest code significantly. Also, reporting and tracking of two-sided constraints in Everest was potentially broken, and now becomes straightforward.

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/optimizer/everest2ropt.py
+++ b/src/everest/optimizer/everest2ropt.py
@@ -2,8 +2,9 @@ import logging
 import os
 from typing import Any
 
+import numpy as np
 from ropt.config.enopt import EnOptConfig, EnOptContext
-from ropt.enums import ConstraintType, PerturbationType, VariableType
+from ropt.enums import PerturbationType, VariableType
 from ropt.transforms import OptModelTransforms
 
 from everest.config import (
@@ -23,13 +24,12 @@ def _parse_controls(controls: FlattenedControls, ropt_config):
         None if type_ is None else VariableType[type_.upper()]
         for type_ in controls.types
     ]
-    indices = [idx for idx, is_enabled in enumerate(controls.enabled) if is_enabled]
     ropt_config["variables"] = {
         "types": None if all(item is None for item in control_types) else control_types,
         "initial_values": controls.initial_guesses,
         "lower_bounds": controls.lower_bounds,
         "upper_bounds": controls.upper_bounds,
-        "indices": indices if indices else None,
+        "mask": controls.enabled,
     }
 
     if "gradients" not in ropt_config:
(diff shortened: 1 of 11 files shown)


src/everest/optimizer/opt_model_transforms.py
```diff
@@ -49,12 +48,34 @@ def transform_magnitudes(self, values: NDArray[np.float64]) -> NDArray[np.float6
...
-        return (
-            coefficients * self._scales,
-            rhs_values - np.matmul(coefficients, self._offsets),
-        )
+        self,
+        coefficients: NDArray[np.float64],
+        lower_bounds: NDArray[np.float64],
+        upper_bounds: NDArray[np.float64],
+    ) -> tuple[NDArray[np.float64], NDArray[np.float64], NDArray[np.float64]]:
+        # The set of linear constraints can be represented by a matrix equation:
```
yngve-sk, line 56: Could you perhaps put this in a `"""` style comment so it is easier to copypaste into a latex parser?

src/everest/optimizer/everest2ropt.py
```diff
@@ -98,15 +98,35 @@ def _parse_objectives(objective_functions: list[ObjectiveFunctionConfig], ropt_c
...
+            )
+            upper_bounds.append(
+                np.inf if constr.upper_bound is None else constr.upper_bound
+            )
+        else:
+            if constr.lower_bound is not None or constr.upper_bound is not None:
+                raise RuntimeError(
+                    "input constraint error: target cannot be combined with bounds"
+                )
+            lower_bounds.append(constr.target)
```
yngve-sk, line 119: Maybe unrelated/self-answering question but does this change the concept of a violation? i.e., are these two things:
1. A variable x being `|d|` away from a target
2. A variable x being `|d|` outside lower & upper bound

Handled exactly the same in ROPT now?

src/everest/optimizer/everest2ropt.py
```diff
@@ -118,76 +138,33 @@ def _get_control_index(name: str):
...
-    rhs_values = []
-    types = []
-
-    def _add_input_constraint(rhs_value, coefficients, constraint_type):
-        if rhs_value is not None:
+    if input_constraints:
+        coefficients_matrix = []
+        for constr in input_constraints:
+            coefficients = [0.0] * len(formatted_control_names)
+            for name, value in constr.weights.items():
```
yngve-sk, line 145: Also maybe not completely related, and this code was not really touched by the PR but here is still a question related to it: It was no immediately obvious to me that the `weights` subfield of the `input_constraints` in the config are used to actually target which variable(s) the constraint applies to. In the case we only use it on one variable, do we still do:
```
input_constraints:
   - weights:
        var_1: 1.0
```
?

src/everest/optimizer/everest2ropt.py
```diff
@@ -118,76 +138,33 @@ def _get_control_index(name: str):
...
-        )
-        _add_output_constraint(
-            lower_bound, ConstraintType.GE, None if upper_bound is None else "lower"
-        )
-
-    ropt_config["nonlinear_constraints"] = {
-        "rhs_values": rhs_values,
-        "types": types,
-    }
+    if output_constraints:
```
yngve-sk, line 162: This looks a lot clearer now. Unrelated Q along similar lines to that above: Here it seems we don't go for doing weighted constraints to multiple objectives, guessing this discrepancy is somehow sensible as there is no `coefficients` is passed as `nonlinear_constraints`?

yngve-sk:
Overall looks good and way more simple than before, could you also add:
* commit message with something like "simplify constraint handling with ropt 0.14" instead of just the update
* a similar (if not the same) description to the commit body as is in the PR message?

src/everest/optimizer/everest2ropt.py
verveerpj: Yes, violations will be handled differently, something like you propose. Since we do not use them yet in Everest, I will redo the way we handle constraint violation in `ropt`, so we can proceed to add that to Everest in the near future.

src/everest/optimizer/everest2ropt.py
verveerpj: It depends on how you see it. In my view it does targets the variables, it just specifies the non-zero coefficients of the set of linear equations. It is a bit cumbersome...

yngve-sk (approved):
LGTM 👍

PR #10054, opened 2025-02-13
Mute FutureWarnings from plugin initialization

berland:
This makes it possible to have modules that emit FutureWarnings upon initialization to be installed as plugins in Ert, without printing these warnings to the user on every Ert startup.

The FutureWarnings should be displayed only when users later actually use deprecated plugins.

**Issue**
Resolves a FutureProblem with res-compat.

**Approach**
Mute.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/plugins/plugin_manager.py
+++ b/src/ert/plugins/plugin_manager.py
@@ -5,6 +5,7 @@ import logging
 import os
 import shutil
 import tempfile
+import warnings
 from argparse import ArgumentParser
 from collections.abc import Callable, Mapping, Sequence
 from itertools import chain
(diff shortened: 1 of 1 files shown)


eivindjahren:
Maybe we just want to log the FuturWarnings that happen on initialization. We let the plugins validate on startup and before starting an experiment so that can be used to give warning when it is used.

src/ert/plugins/plugin_manager.py
```diff
@@ -56,7 +57,9 @@ def __init__(self, plugins: Sequence[object] | None = None) -> None:
         self.add_hookspecs(ert.plugins.hook_specifications)
         if plugins is None:
             self.register(ert.plugins.hook_implementations)
-            self.load_setuptools_entrypoints(_PLUGIN_NAMESPACE)
+            with warnings.catch_warnings():
```
eivindjahren, line 60: I think this needs a small comment about why we are doing this.

berland: fixed.

eivindjahren: Maybe something about the user not being able to action on installed deprecated plugins could be added? Just wanted something more clearly about why and less about what. Other then that we this is good.

eivindjahren (approved):

PR #10061, opened 2025-02-13
Fixups for terminate experiment correctness

xjules:
**Issue**
 - In case that termination  of experiment happens monitor returns immediately and thus killing the entire ensemble task and thus evaluator does not finish. 
 - router_socket.close() should be called only by the server task

**Approach**
Await evaluator task

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -281,7 +281,6 @@ class EnsembleEvaluator:
                 else:
                     logger.error(f"Unexpected error when listening to messages: {e}")
             except asyncio.CancelledError:
-                self._router_socket.close()
                 return
 
     async def forward_checksum(self, event: Event) -> None:
--- a/src/ert/run_models/base_run_model.py
+++ b/src/ert/run_models/base_run_model.py
@@ -20,10 +20,7 @@ from typing import TYPE_CHECKING, Any, Protocol, cast
 import numpy as np
 
 from _ert.events import EESnapshot, EESnapshotUpdate, EETerminated, Event
-from ert.analysis import (
-    ErtAnalysisError,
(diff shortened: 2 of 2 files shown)


src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -281,7 +281,6 @@ async def listen_for_messages(self) -> None:
                 else:
                     logger.error(f"Unexpected error when listening to messages: {e}")
             except asyncio.CancelledError:
-                self._router_socket.close()
```
jonathan-eq, line 284: Very good! Do you know what would happen if we call `self._router_socket.close()` twice?

jonathan-eq (changes requested):
Very good! I have some questions (see comments), and there is a small typo in the first commit message 👍

src/ert/run_models/base_run_model.py
```diff
@@ -600,6 +597,7 @@ async def run_ensemble_evaluator_async(
         )
         await evaluator._server_started
         if not (await self.run_monitor(ee_config, ensemble.iteration)):
+            await evaluator_task
```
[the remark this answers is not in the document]
xjules: It shouldn't hang as monitor is done, meaning that `evaluator.stop()` was initiated. If we cancel it then none of the (scheduler, dispatch) events will be processed anymore.

src/ert/ensemble_evaluator/evaluator.py
xjules: > Very good! Do you know what would happen if we call `self._router_socket.close()` twice?

It is unpredictable,we could get `zmq.error.ZMQError`

jonathan-eq (approved):
LGTM! 🏄

PR #10069, opened 2025-02-14
Fix fm_dispatch not terminating correctly and gui stuck on fm running when terminating

### LINKED ISSUE #10055: Terminate experiment not working properly on localdriver

When pressing terminate experiment, the fm_step tableview still shows the fm as running. Some investigation shows that this is due to the event reporter not being able to close down. This might be related to the method for exiting context of the new zeromq client

The experiment (incl fm steps) is terminated successfully. 

### steps to reproduce

```Python
Run poly.ert
Press terminate experiment
See how the fm tableview never sets the Status to failed (it is stuck at Running)

```

Image

- [ ] python 3.11
- [x] python 3.12
- [x] macosx
- [ ] rhel7
- [ ] rhel8
- [x] local queue
- [ ] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

### LINKED ISSUE #10059: FM steps are not marked as failed in GUI after pressing terminate experiment

When pressing terminate experiment, the fm_step tableview still shows the fm as running. Some investigation shows that this is due to the event reporter not sending an Exited message before getting torn down. 

The fm step is marked as failed, and the duration counter is stopped. 
Image
### steps to reproduce

Run poly.ert
Press terminate experiment
See how the fm tableview never sets the Status to failed (it is stuck at Running)

- [ ] python 3.11
- [x] python 3.12
- [x] macosx
- [ ] rhel7
- [ ] rhel8
- [x] local queue
- [ ] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

### LINKED ISSUE #10060: Fm_dispatch has problems getting ack on DISCONNECT msg when calling terminate

When calling terminate, there seems to be problems when fm_dispatch is trying to send DISCONNECTED msg to EE, in the midst of tearing down. It requires a lot of retries, and with the local driver, it calls SIGKILL before it has managed to shut down gracefully. 
Maybe there is a race condition somewhere, so EE is shutting down before fm_dispatch has(?)

It should not require retries on getting ACK on disconnect. 

### steps to reproduce

```Python

```

- [ ] python 3.11
- [x] python 3.12
- [x] macosx
- [ ] rhel7
- [ ] rhel8
- [x] local queue
- [ ] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

jonathan-eq:
**Issue**
Resolves #10055 
Resolves #10059 
Resolves #10060 

**Approach**
This PR issue where cancelling an ensemble in EE would cause messaging (and therefore teardown) from fm_dispatch to hang, due to ensemble.cancel() being blocking. This causes teardown of fm_dispatch to go from 35-40s -> 1s. 

This PR also makes fm_dispatch send ForwardModelStepFailure with the message "fm_dispatch was terminated" when it is being terminated, instead of sending nothing. 

Before:
image

After:
image

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/_ert/forward_model_runner/fm_dispatch.py
+++ b/src/_ert/forward_model_runner/fm_dispatch.py
@@ -11,6 +11,7 @@ from datetime import datetime
 
 from _ert.forward_model_runner import reporting
 from _ert.forward_model_runner.reporting.message import (
+    Exited,
     Finish,
     Message,
     ProcessTreeStatus,
@@ -25,6 +26,7 @@ FORWARD_MODEL_DESCRIPTION_FILE = "jobs.json"
 # yielding FileNotFoundError due to synchronization issues. This constant
 # determines how long we can wait for synchronization to happen:
 FILE_RETRY_TIME = 30
+FORWARD_MODEL_TERMINATED_MSG = "Forward model was terminated"
 
 logger = logging.getLogger(__name__)
 
@@ -104,7 +106,8 @@ def _read_fm_description_file(retry=True):
 
 
 def _report_all_messages(
-    messages: Generator[Message], reporters: list[reporting.Reporter]
+    messages: Generator[Message],
+    reporters: list[reporting.Reporter],
 ) -> None:
     for msg in messages:
         logger.info(f"Forward model status: {msg}")
@@ -127,16 +130,18 @@ def _report_all_messages(
             _stop_reporters_and_sigkill(reporters)
 
 
-def _stop_reporters_and_sigkill(reporters):
-    _stop_reporters(reporters)
+def _stop_reporters_and_sigkill(reporters, exited_event: Exited | None = None):
+    _stop_reporters(reporters, exited_event)
     pgid = os.getpgid(os.getpid())
     os.killpg(pgid, signal.SIGKILL)
 
 
-def _stop_reporters(reporters: Iterable[reporting.Reporter]) -> None:
+def _stop_reporters(
+    reporters: Iterable[reporting.Reporter], exited_event: Exited | None = None
+) -> None:
     for reporter in reporters:
         if isinstance(reporter, reporting.Event):
-            reporter.stop()
+            reporter.stop(exited_event=exited_event)
 
 
 def sigterm_handler(_signo, _stack_frame):
@@ -189,7 +194,14 @@ def fm_dispatch(args):
     )
 
     fm_runner = ForwardModelRunner(fm_description)
-    signal.signal(signal.SIGTERM, lambda _, __: _stop_reporters_and_sigkill(reporters))
+
+    def sigterm_handler(_signo, _stack_frame):
+        exited_event = Exited(
+            fm_runner._currently_running_step, exit_code=1
+        ).with_error(FORWARD_MODEL_TERMINATED_MSG)
+        _stop_reporters_and_sigkill(reporters, exited_event)
+
+    signal.signal(signal.SIGTERM, sigterm_handler)
     _report_all_messages(fm_runner.run(parsed_args.steps), reporters)
 
 
--- a/src/_ert/forward_model_runner/reporting/event.py
+++ b/src/_ert/forward_model_runner/reporting/event.py
@@ -87,7 +87,9 @@ class Event(Reporter):
             # can be finished but not all the events were sent yet
             self._finished_event_timeout = 600
 
-    def stop(self):
+    def stop(self, exited_event: Exited | None = None):
+        if exited_event:
+            self._statemachine.transition(exited_event)
         self._event_queue.put(Event._sentinel)
         self._done.set()
         if self._event_publisher_thread.is_alive():
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -357,7 +357,12 @@ class EnsembleEvaluator:
         """
         if self._ensemble.cancellable:
             logger.debug("Cancelling current ensemble")
-            await self._ensemble.cancel()
+            self._ee_tasks.append(
+                asyncio.create_task(
+                    self._ensemble.cancel(), name="ensemble_cancellation_task"
+                )
+            )
+
         else:
             logger.debug("Stopping current ensemble")
             self.stop()
@@ -408,6 +413,7 @@ class EnsembleEvaluator:
                 elif task.get_name() in {
                     "ensemble_task",
                     "listener_task",
+                    "ensemble_cancellation_task",
                 }:
                     timeout = self.CLOSE_SERVER_TIMEOUT
                 else:
--- a/tests/ert/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
+++ b/tests/ert/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
@@ -37,6 +37,7 @@ from ert.ensemble_evaluator import (
     FMStepSnapshot,
     Monitor,
 )
+from ert.ensemble_evaluator._ensemble import LegacyEnsemble
 from ert.ensemble_evaluator.evaluator import detect_overspent_cpu
 from ert.ensemble_evaluator.state import (
     ENSEMBLE_STATE_STARTED,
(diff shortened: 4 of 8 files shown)


jonathan-eq:
@andreas-el Do you think the gui change is a good idea? Maybe it should have a better "error" message?

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -408,6 +413,7 @@ async def _monitor_and_handle_tasks(self) -> None:
                 elif task.get_name() in {
                     "ensemble_task",
                     "listener_task",
+                    "ensemble_cancellation_task",
```
xjules, line 416: It's not part of this PR, but wondering whether listener_task should be there :thinking:

tests/ert/unit_tests/forward_model_runner/test_fm_dispatch.py
```diff
@@ -405,3 +407,94 @@ def raises(*args, **kwargs):
 
     _report_all_messages(iter([message1, message2]), [reporter])
     reporter.report.assert_called_once_with(message1)
+
+
+@pytest.mark.integration_test
```
xjules, line 412: Nice test! Shall we also make a full integration test that will confirm the ensemble_cancellation_task working?

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: Hmm, maybe not. Is there a case where it will be the first one to finish?

tests/ert/unit_tests/forward_model_runner/test_fm_dispatch.py
jonathan-eq: How about a EnsembleEvaluator unit test instead?
Something like `test_ensemble_evaluator_signal_cancel_does_not_block`, or `test_ensemble_evaluator__listen_for_messages_handlers_do_not_block`?

jonathan-eq:
This now has the issue that it will overwrite the error message in case of `The run is cancelled due to reaching MAX_RUNTIME`. We can update the error message to contain both the old one and the new one (`fm_dispatch was terminated`), or have it ignore the new error message if it is already set. 

This PR also makes the end_time closer to the actual end time of when the FM step was actually stopped, so it requires the test snapshot to be updated.
What do you think @berland?

src/_ert/forward_model_runner/fm_dispatch.py
```diff
@@ -25,6 +26,7 @@
 # yielding FileNotFoundError due to synchronization issues. This constant
 # determines how long we can wait for synchronization to happen:
 FILE_RETRY_TIME = 30
+FM_DISPATCH_WAS_TERMINATED_MSG = "fm_dispatch was terminated"
```
xjules, line 29: Let's change it to "Forward model was terminated"

tests/ert/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -672,3 +673,54 @@ async def test_snapshot_on_resubmit_is_cleared(evaluator_to_use):
             assert snapshot.get_fm_step("0", "1")["status"] == FORWARD_MODEL_STATE_INIT
 
         await monitor.signal_done()
+
+
+@pytest.mark.integration_test
+async def test_signal_cancel_does_not_cause_evaluator_dispatcher_communication_to_hang(
```
xjules, line 679: Tbh, I like to test setup but missing a self-explanatory assert (as a punchline :smile: )

tests/ert/unit_tests/forward_model_runner/test_fm_dispatch.py
```diff
@@ -405,3 +407,94 @@ def raises(*args, **kwargs):
...
+                await asyncio.sleep(0.5)
+                if any(
+                    msg_type in orjson.loads(msg).get("event_type", "")
+                    for msg in zmq_server.messages
+                ):
+                    return
+
+        # wait for fm running
+        await asyncio.wait_for(wait_for_msg("forward_model_step.running"), timeout=10)
+        p.terminate()
```
xjules, line 494: How safe is it to test process terminate? Maybe @berland can double check?

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -87,7 +87,9 @@ def __init__(
             # can be finished but not all the events were sent yet
             self._finished_event_timeout = 600
 
-    def stop(self):
+    def stop(self, exited_event: Exited | None = None):
+        if exited_event:
+            self._statemachine.transition(exited_event)
```
berland, line 92: Reading the source code of `statemachine.py` I would think you would need to add the Exited state to `_transitions` for the transition to be accepted?

tests/ert/unit_tests/forward_model_runner/test_fm_dispatch.py
berland: should be the same as `os.kill()` (with sigterm I suppose). If it works as intended, I would be happy :)

tests/ert/unit_tests/forward_model_runner/test_fm_dispatch.py
```diff
@@ -405,3 +407,94 @@ def raises(*args, **kwargs):
...
+        f.write(
+            """#!/usr/bin/env python
+import time
+time.sleep(180)"""
+        )
+
+    executable = os.path.realpath("dummy_executable")
+    os.chmod("dummy_executable", stat.S_IRWXU | stat.S_IRWXO | stat.S_IRWXG)
+
+    fm_description = {
```
berland, line 427: I think it should be possible to remove all empty settings here (empty dicts or None). These are `@dataclass` objects now.

src/_ert/forward_model_runner/reporting/event.py
berland: me blind 🙈

tests/ert/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -672,3 +673,54 @@ async def test_snapshot_on_resubmit_is_cleared(evaluator_to_use):
...
+    async with Monitor(url, token) as monitor, Client(url, token=token) as dispatch:
+        event = ForwardModelStepRunning(
+            ensemble=evaluator.ensemble.id_,
+            real="0",
+            fm_step="0",
+            current_memory_usage=1000,
+        )
+        await dispatch.send(event_to_json(event))
+
+        async def try_sending_event_while_monitor_is_cancelling():
```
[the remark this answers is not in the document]
jonathan-eq: Renamed it 👌

src/_ert/forward_model_runner/fm_dispatch.py
```diff
@@ -189,7 +194,14 @@ def fm_dispatch(args):
     )
 
     fm_runner = ForwardModelRunner(fm_description)
-    signal.signal(signal.SIGTERM, lambda _, __: _stop_reporters_and_sigkill(reporters))
+
+    def sigterm_handler(_signo, _stack_frame):
+        exited_event = Exited(
+            fm_runner._currently_running_step, exit_code=1
+        ).with_error(FORWARD_MODEL_TERMINATED_MSG)
```
jonathan-eq, line 201: This might cause a bug if we SIGTERM after the final fm_step has already reported finished, but the event-reporter has not processed that event yet.

jonathan-eq: Will try writing a test for it^

src/ert/ensemble_evaluator/evaluator.py
xjules: listener_task should never finish. It should be only cancelled once server_task is done. I'll write an issue for it as if `listener_task` fails we should try to reconnect.

tests/ert/unit_tests/forward_model_runner/test_fm_dispatch.py
```diff
@@ -405,3 +408,79 @@ def raises(*args, **kwargs):
...
+                ):
+                    return
+
+        # wait for fm running
+        await asyncio.wait_for(wait_for_msg("forward_model_step.running"), timeout=10)
+        p.terminate()
+        # wait for fm_dispatch has been terminated, and sends failure message
+        await asyncio.wait_for(wait_for_msg("forward_model_step.failure"), timeout=10)
+        assert (
+            orjson.loads(zmq_server.messages[-1])["error_msg"]
```
xjules, line 484: do you need to use `orjson.laods(...)`? We do have utility `event_from_json`...

jonathan-eq: Yes, I will use the utility function instead 🙆‍♂️

xjules (approved):
Nice job @jonathan-eq ! :rocket:

src/_ert/forward_model_runner/fm_dispatch.py
jonathan-eq: Difficult test to write for such a marginal cornercase. Will not write a test for it; but we are (hopefully) aware of it in case it ever happens.

PR #10076, opened 2025-02-16
Update ropt 0.15

verveerpj:
**Issue**
Update to ropt 0.15

**Approach**
Following changes are maded:
- Not related to the 0.15 update:
       - Renamed the `init_transforms` method to `init_domain_transforms`.
- Needed for the 0.15 update:
       1. Ropt now requires evaluations for unscaled controls. Everest must scale again before calling forward models until these are adapted to handle unscaled controls.
       2. Small change in the signature of the context object expected by EnOptConfig.model_validate.
       3. The optimization domain transformation classes have renamed methods that more clear convey their intent.
       4. The transformation classes now support transformation of constraint differences/violations.
       5. A weighting has been introduced for linear constraints after scaling.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/optimizer/opt_model_transforms.py
+++ b/src/everest/optimizer/opt_model_transforms.py
@@ -18,6 +18,12 @@ from everest.config.utils import FlattenedControls
 
 
 class ControlScaler(VariableTransform):
+    """Transformation object to define a linear scaling of controls.
+
+    This object defines a linear scaling from lower and upper bounds [lb, ub] in
+    the user domain to a target range in the optimizer domain.
+    """
+
     def __init__(
         self,
         lower_bounds: Sequence[float],
@@ -25,7 +31,19 @@ class ControlScaler(VariableTransform):
         scaled_ranges: Sequence[tuple[float, float]],
         auto_scales: Sequence[bool],
     ) -> None:
-        self._scales = [
+        """Transformation object to define a linear scaling.
+
+        This is implemented by internally representing the transformation from
+        the user to the optimizer domain by a subtraction of an offset and a
+        division by a scaling factor.
+
+         Args:
+             lower_bounds:  Lower bounds in the user domain.
+             upper_bounds:  Upper bounds in the user domain.
+             scaled_ranges: target ranges in the optimizer domain.
+             auto_scales:   Flag indicating if scaling is required.
+        """
+        self._scaling_factors = [
             (ub - lb) / (sr[1] - sr[0]) if au else 1.0
             for au, lb, ub, sr in zip(
                 auto_scales, lower_bounds, upper_bounds, scaled_ranges, strict=True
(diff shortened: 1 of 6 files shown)


src/everest/optimizer/opt_model_transforms.py
```diff
@@ -37,6 +37,7 @@ def __init__(
                 auto_scales, lower_bounds, self._scales, scaled_ranges, strict=True
             )
         ]
+        self._equation_scaling: NDArray[np.float64] | None = None
```
yngve-sk, line 40: Is equation scaling the term usually used for this? I guess within ROPT it is more math-y, but here an equation is a... I'm not totally sure. AFAIK from the comment it is the max weight, so basically it is just there to get all weights into a [0, 1] range, and there is no offset so it is basically a scale. Also, do we now assume that `linear_constraints_to_optimizer` is invoked before all the other dependents on `self._equation_scaling` are called?

verveerpj: Hm, yes. I am not sure if it is indeed a good idea to include this in this PR. There are some scaling issues with linear constraints, but may that needs a different solution. We did not have any scaling for that before, so it should be fine to continue with this. I will drop this commit.

src/everest/optimizer/opt_model_transforms.py
```diff
@@ -79,6 +81,19 @@ def transform_linear_constraints(  # type: ignore
             coefficients = coefficients * self._scales  # noqa: PLR6104
         return coefficients, lower_bounds, upper_bounds
 
+    def bound_constraint_diffs_from_optimizer(
```
yngve-sk, line 84: Is this used now?

yngve-sk:
Mostly looks good. Nitpick terminology question: Is scaling/transform synonymous? I see variables called something with transform then comments talking about scaled/unscaled. If it would make sense I think I would have an easier time understanding "variable X is in optimization domain/space" rather than "unscaled"/"scaled" as it is a bit arbitrary what we define as scaled/unscaled. I guess scaled now means mostly natural numbers, but it could also be argued to be the other way around.

src/everest/optimizer/opt_model_transforms.py
verveerpj: Yes, it is part of the abstract interface. It does nothin here because the way we scale variables does not really affect the value of linear constraints (even they must be adjusted).

yngve-sk: Right, could we remove it for now so it appears in the commit when we start using it?

verveerpj:
> Mostly looks good. Nitpick terminology question: Is scaling/transform synonymous? I see variables called something with transform then comments talking about scaled/unscaled. If it would make sense I think I would have an easier time understanding "variable X is in optimization domain/space" rather than "unscaled"/"scaled" as it is a bit arbitrary what we define as scaled/unscaled. I guess scaled now means mostly natural numbers, but it could also be argued to be the other way around. The `.to|from_optimizer` etc naming makes it very clear what the scale does though, so something more similar to that throughout the codebase/comments would be more readable I think.

I have added documentation to hopefully clarify things better.

src/everest/optimizer/opt_model_transforms.py
verveerpj: I do not consider it a no-opt. I think it is good to be explicit that the result happens to be the same as the input. Guess, it's a question of taste, but I prefer to keep it like this and not complicate the logic in the ropt transformation code.

PR #10079, opened 2025-02-17
Type everest/config

### LINKED ISSUE #9973: Type everest/config
Todo: remove [mypy-everest.config.*] from .mypy.ini

frode-aarstad:
**Issue**
Resolves #9973 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/config/everest_config.py
+++ b/src/everest/config/everest_config.py
@@ -6,8 +6,8 @@ from io import StringIO
 from itertools import chain
 from pathlib import Path
 from typing import (
-    TYPE_CHECKING,
     Annotated,
+    Any,
     Optional,
     Protocol,
     Self,
@@ -23,6 +23,7 @@ from pydantic import (
     field_validator,
     model_validator,
 )
+from pydantic_core import ErrorDetails
 from ruamel.yaml import YAML, YAMLError
 
 from ert.config import ErtConfig
@@ -64,36 +65,33 @@ from .simulator_config import SimulatorConfig, simulator_example
 from .well_config import WellConfig
 from .workflow_config import WorkflowConfig
 
-if TYPE_CHECKING:
-    from pydantic_core import ErrorDetails
 
-
-def _dummy_ert_config():
+def _dummy_ert_config() -> ErtConfig:
     site_config = ErtConfig.read_site_config()
     dummy_config = {"NUM_REALIZATIONS": 1, "ENSPATH": "."}
-    dummy_config.update(site_config)
+    dummy_config.update(site_config)  # type: ignore
     return ErtConfig.with_plugins().from_dict(config_dict=dummy_config)
 
 
-def get_system_installed_jobs():
+def get_system_installed_jobs() -> list[str]:
     """Returns list of all system installed job names"""
     return list(_dummy_ert_config().installed_forward_model_steps.keys())
 
 
 class EverestValidationError(ValueError):
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__()
         self.errors: list[tuple[ErrorDetails, tuple[int, int] | None]] = []
 
     @property
-    def error(self):
+    def error(self) -> list[tuple[ErrorDetails, tuple[int, int] | None]]:
         return self.errors
 
-    def __str__(self):
+    def __str__(self) -> str:
         return f"{self.errors!s}"
 
 
-def _error_loc(error_dict: "ErrorDetails") -> str:
+def _error_loc(error_dict: ErrorDetails) -> str:
     return " -> ".join(
         str(e) for e in error_dict["loc"] if e is not None and e != "__root__"
     )
@@ -118,7 +116,7 @@ class HasName(Protocol):
     name: str
 
 
-class EverestConfig(BaseModel):  # type: ignore
+class EverestConfig(BaseModel):
     controls: Annotated[list[ControlConfig], AfterValidator(unique_items)] = Field(
         description="""Defines a list of controls.
          Controls should have unique names each control defines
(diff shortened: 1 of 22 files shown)


src/everest/config/everest_config.py
```diff
@@ -588,36 +592,36 @@ def simulation_dir(self) -> str | None:
...
 
         cfgdir = self.output_dir
         if cfgdir is None:
             return path
 
-        return os.path.join(cfgdir, path)
+        return os.path.join(cfgdir, path)  # type: ignore
 
-    def _get_output_subdirectory(self, subdirname: str):
+    def _get_output_subdirectory(self, subdirname: str) -> str | None:
```
DanSava, line 604: This cannot be None because `self.output_dir` is never None. I think you can remove 
``` python 
  if self.output_dir is None:
      return None
```

src/everest/config/optimization_config.py
```diff
@@ -199,7 +199,7 @@ class OptimizationConfig(BaseModel, extra="forbid"):  # type: ignore
     )
 
     @model_validator(mode="after")
-    def validate_backend_and_algorithm(self):  # pylint: disable=E0213
+    def validate_backend_and_algorithm(self) -> "OptimizationConfig":
```
DanSava, line 202: You can use `Self` as a return type `from typing import Self`

src/everest/config/sampler_config.py
```diff
@@ -37,9 +37,12 @@ class SamplerConfig(BaseModel):  # type: ignore
         description="""Whether to share perturbations between realizations.
 """,
     )
+    model_config = ConfigDict(
+        extra="forbid",
+    )
 
     @model_validator(mode="after")
-    def validate_backend_and_method(self):  # pylint: disable=E0213
+    def validate_backend_and_method(self) -> "SamplerConfig":
```
DanSava, line 45: You can use `Self` as a return type `from typing import Self`

src/everest/config/sampler_config.py
```diff
@@ -37,9 +37,12 @@ class SamplerConfig(BaseModel):  # type: ignore
         description="""Whether to share perturbations between realizations.
 """,
     )
+    model_config = ConfigDict(
+        extra="forbid",
```
DanSava, line 41: In `extra="forbid",` is the comma `,` really needed? Adding it forces ruff formatting to put everything on a new line.

src/everest/config/validation_utils.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, install_data: list[InstallDataConfig], config_path: Path):
         self._config_dir = str(config_path.parent)
         self._cwd = os.getcwd()
 
-    def __enter__(self):
+    def __enter__(self) -> "InstallDataContext":
```
DanSava, line 32: You can use `Self` as a return type

src/everest/config/validation_utils.py
```diff
@@ -139,13 +139,13 @@ def unique_items(items: Sequence[T]) -> Sequence[T]:
...
 
-def valid_range(range_value: tuple[float, float]):
+def valid_range(range_value: tuple[float, float]) -> tuple[float, float]:
     if range_value[0] >= range_value[1]:
         raise ValueError("scaled_range must be a valid range [a, b], where a < b.")
     return range_value
 
 
-def check_path_valid(path: str):
+def check_path_valid(path: str | None) -> None:
```
DanSava, line 148: Why allow None  if we have code that raises an error if we have anything else than a string? 
``` python
    if not isinstance(path, str):
        raise ValueError("str type expected")
```

src/everest/config/export_config.py
```diff
@@ -36,6 +36,8 @@ class ExportConfig(BaseModel, extra="forbid"):  # type: ignore
 
     @field_validator("csv_output_filepath", mode="before")
     @classmethod
-    def validate_output_file_writable(cls, csv_output_filepath):  # pylint:disable=E0213
+    def validate_output_file_writable(
+        cls, csv_output_filepath: str | None
+    ) -> str | None:
         check_writable_filepath(csv_output_filepath)
```
DanSava, line 42: Why not just return None and not force to pass None to `check_writable_filepath`?

src/everest/config/validation_utils.py
```diff
@@ -161,16 +161,18 @@ def check_path_valid(path: str):
             raise ValueError(str(e)) from e
 
 
-def check_writable_filepath(path: str):
+def check_writable_filepath(path: str | None) -> None:
```
DanSava, line 164: I don't think `path` should be allowed to be None

src/everest/config/export_config.py
frode-aarstad: The subsequent functions raise a Exception on None. Returning None would change behavior

src/everest/config/validation_utils.py
frode-aarstad: I guess in this case a better type for paths is Any

frode-aarstad:
I changed the validation logic a bit to avoid the None issues mentioned in the comments

src/everest/config/export_config.py
```diff
@@ -36,6 +36,8 @@ class ExportConfig(BaseModel, extra="forbid"):  # type: ignore
 
     @field_validator("csv_output_filepath", mode="before")
     @classmethod
-    def validate_output_file_writable(cls, csv_output_filepath):  # pylint:disable=E0213
+    def validate_output_file_writable(cls, csv_output_filepath: str | None) -> str:
+        if csv_output_filepath is None:
```
DanSava, line 40: Now I understand the confusion with this one. The export functionality does not work anymore. It needs to be removed. I will make a PR with that.

DanSava (approved):
Looks good! Thank you for your patience.

PR #10081, opened 2025-02-17
Skip executing empty forward model w/ ensemble evaluator

### LINKED ISSUE #10075: Test test_simulator_cache hangs or is extremely slow

After this PR (reverted since) https://github.com/equinor/ert/pull/10068 
`tests/everest/test_simulator_cache.py::test_simulator_cache` hangs or fails on the PR runs

https://github.com/equinor/ert/actions/runs/13329790996/job/37234298932

The test should just pass.

### steps to reproduce

```Python
`pytest -svv tests/everest/test_simulator_cache.py::test_simulator_cache`
```

- [x] python 3.11
- [x] python 3.12
- [ ] macosx
- [ ] rhel7
- [x] rhel8
- [ ] local queue
- [ ] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

yngve-sk:
Resolves #10075 
Correct everest behavior. If it found the simulation results in cache, it should not have to create a runpath, evaluate the forward model etc. This will make it also skip creating the ensemble in storage, this behavior will/should be revisited in the future as we use ERT storage for the simulation results.

### DIFF
--- a/src/ert/run_models/everest_run_model.py
+++ b/src/ert/run_models/everest_run_model.py
@@ -453,6 +453,46 @@ class EverestRunModel(BaseRunModel):
             and (active_control_vectors is None or active_control_vectors[idx])
         ]
 
+        if not evaluated_control_indices:
+            cached_results = self._get_cached_results(
+                control_values, model_realizations
+            )
+
+            objectives = np.zeros(
+                (
+                    control_values.shape[0],
+                    len(self._everest_config.objective_names),
+                ),
+                dtype=float64,
+            )
+
+            constraints = (
+                np.zeros(
+                    (
+                        control_values.shape[0],
+                        len(self._everest_config.constraint_names),
+                    ),
+                    dtype=float64,
+                )
+                if self._everest_config.constraint_names
+                else None
+            )
+
+            for control_idx, (
+                cached_objectives,
+                cached_constraints,
+            ) in cached_results.items():
+                objectives[control_idx, ...] = cached_objectives
+                if constraints is not None:
+                    assert cached_constraints is not None
+                    constraints[control_idx, ...] = cached_constraints
+
+                # Increase the batch ID for the next evaluation:
+            self._batch_id += 1
+
+            # Return the results, together with the indices of the evaluated controls:
+            return objectives, constraints, evaluated_control_indices
+
         # Create the batch to run:
         sim_controls = self._create_simulation_controls(
             control_values, evaluated_control_indices
--- a/tests/everest/test_simulator_cache.py
+++ b/tests/everest/test_simulator_cache.py
@@ -8,12 +8,12 @@ from everest.config import EverestConfig
 
 @pytest.mark.integration_test
 def test_simulator_cache(copy_math_func_test_data_to_tmp):
-    n_evals = 0
+    n_invocations = 0
 
     def new_call(*args):
-        nonlocal n_evals
+        nonlocal n_invocations
         result = original_call(*args)
-        n_evals += (result.evaluation_ids >= 0).sum()
+        n_invocations += 1
         return result
 
     config = EverestConfig.load_file("config_minimal.yml")
@@ -24,22 +24,22 @@ def test_simulator_cache(copy_math_func_test_data_to_tmp):
     evaluator_server_config = EvaluatorServerConfig()
 
     # Modify the forward model function to track number of calls:
-    original_call = run_model._forward_model_evaluator
-    run_model._forward_model_evaluator = new_call
+    original_call = run_model._evaluate_and_postprocess
+    run_model._evaluate_and_postprocess = new_call
 
     # First run populates the cache:
     run_model.run_experiment(evaluator_server_config)
-    assert n_evals > 0
+    assert n_invocations > 0
     variables1 = list(run_model.result.controls.values())
     assert np.allclose(variables1, [0.5, 0.5, 0.5], atol=0.02)
 
     # Now do another run, where the functions should come from the cache:
-    n_evals = 0
+    n_invocations = 0
 
     # The batch_id was used as a stopping criterion, so it must be reset:
     run_model._batch_id = 0
 
     run_model.run_experiment(evaluator_server_config)
-    assert n_evals == 0
+    assert n_invocations == 0
     variables2 = list(run_model.result.controls.values())
     assert np.array_equal(variables1, variables2)


src/ert/run_models/everest_run_model.py
```diff
@@ -364,94 +364,126 @@ def _run_forward_model(
...
-        )
+            # Add the results from the evaluations to the cache:
+            self._add_results_to_cache(
+                control_values,
+                model_realizations,
+                sim_ids_to_evaluate,
+                objectives,
+                constraints,
+            )
+        else:
```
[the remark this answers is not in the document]
yngve-sk: Hmm yeah that confused me too, thing with the current "mixed" getter is it checked the cache lastly, and it took in a larger set of args than what is needed to get only from the cache, so I would opt for splitting, I'll take another round and try to simplify it

src/ert/run_models/everest_run_model.py
```diff
@@ -405,8 +407,36 @@ def _run_forward_model(
...
+                control_values, sim_ids_to_evaluate, results
+            )
+
+            cached_objectives, cached_constraints = self._get_cached_results(
+                control_values, model_realizations
+            )
+
+            objectives = np.array(
+                [
+                    cached if not np.all(cached == 0) else simulated
```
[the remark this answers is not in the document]
yngve-sk: Ooops! I think I missed that they are cast to nans after, I changed it do use nans instead of 0 to denote "missingness"

src/ert/run_models/everest_run_model.py
```diff
@@ -355,103 +355,181 @@ def _run_forward_model(
...
+            control_values, model_realizations
+        )
 
         # Collect the indices of the controls that must be evaluated in the batch:
-        evaluated_control_indices = [
-            idx
-            for idx in range(control_values.shape[0])
-            if idx not in cached_results
-            and (active_control_vectors is None or active_control_vectors[idx])
+        sim_ids_to_evaluate = [
```
xjules, line 369: I don't like this indentation tbh. Could you instead just do:
```ptyhon
if not sim_ids_to_evaluate:
self._batch_id += 1

        # Return the results, together with the indices of the evaluated controls:
        return objectives, constraints, sim_ids_to_evaluate
```
and then just continue?

src/ert/run_models/everest_run_model.py
```diff
@@ -355,103 +355,181 @@ def _run_forward_model(
...
-            objectives,
-            constraints,
-        )
+            cached_objectives, cached_constraints = self._get_cached_results(
+                control_values, model_realizations
+            )
+
+            objectives = np.array(
+                [
+                    cached if not np.all(np.isnan(cached)) else simulated
```
verveerpj, line 420: I think it is dangerous to rely on a special value to check is something was cached or not. In this case, there could simply be nan values because in the original run (when the cached values were created) the forward model failed for some reason. In that case, the results were set to nan, so that is basically a valid value that carries the information that the forward model failed.

In you current code, this would mean that the cached values are not used, since they are nan, but also the forward model did not run, so the end value would be whatever you used to initialize that objecives/constraints with.

Solutions include:
1. initialize with nan, so in this specific case you end up with nan again
2. not allow failed runs to be stored in cache, so
3. use the original information in the cache structure (a dict mapping id to cached values) which clearly defines what was cached.

Option 1 could be very confusing in the future. Using nan as a flag remains confusing to me.
Option 2 is something we should discuss separately, maybe that is a good idea, but let's not do that in this PR, seems to be a different issue to me.
Option 3 seems safer to me: consult the dict with cached values to see if the  results are actually cached.

yngve-sk: Hmm that makes sense, I'll go for option 3

src/ert/run_models/everest_run_model.py
```diff
@@ -355,40 +355,68 @@ def _run_forward_model(
...
+                )
+                if has_constraints
+                else None
+            )
+
+            # Increase batch ID, preserves current behavior
+            # Should be revised
+            self._batch_id += 1
+            return objectives, constraints, sim_ids_to_evaluate
+
```
verveerpj, line 403: This should work fine. Maybe there is a way to consolidate this piece of code and the one below into a method call, maybe combining it with `_get_cached_results_by_sim_id` or something like that? Now these are big pieces of code, distracting a bit from the flow of the method. And the are very similar, apart from taking defaults from `simulated_constraints`.

src/ert/run_models/everest_run_model.py
```diff
@@ -421,44 +421,86 @@ def _on_before_forward_model_evaluation(
...
+                [cached_results.get(sim) for sim in range(len(cached_results))],
+                dtype=np.float64,
+            )
+        else:
+            n_cached_results = len(cached_results) if cached_results is not None else 0
+            n_sim_results = (
+                len(simulated_results) if simulated_results is not None else 0
+            )
+
+            n_sims = max(n_cached_results, n_sim_results)
```
verveerpj, line 446: I do not understand this. The size of the output matrices should be fixed by the number of control vectors passed to the evaluator. Some entries may be zero, due to the control vectors not being active. Others should be filled in by the evaluations or from cache. I am not seeing how this would work here.

yngve-sk: We could pass the control vectors or num sims explicitly, but this function is basically to get the cached / simulated output, there are cases when both are None, or one is None, as it does the same thing for constraints / objectives. So some `None` checking is needed for this to work the same for the constraints / objectives. We have these cases to cover:

* both are `None`, happens for simulations with no constraints
* cached (first arg) is `None`, simulated is not `None`
* cached is an empty obj, simulated is not `None`
* cached is populated, simulated is None (happens when cache hit) 

I will have another round and resolve some of these cases a bit more upstream so there is less null-logic in here, but it is a function that deals with some non-ness as it is applied the same to the constraints/objectives.

verveerpj: Yes, I see now how that would work, it would work. However, that is not easy to deduce from the code at all...

yngve-sk: I attempted to clarify it some more, some of the null-logic is moved/removed and some tests are making the edge cases more clear.

src/ert/run_models/everest_run_model.py
```diff
@@ -420,44 +420,89 @@ def _on_before_forward_model_evaluation(
...
+    def _consolidate_cache_and_sim_results(
+        num_simulations: int,
+        cached_results: dict[int, Any],
+        simulated_results: NDArray[np.float64] | None = None,
+    ) -> NDArray[np.float64] | None:
+        if simulated_results is None:
+            return np.array(
+                [cached_results.get(sim) for sim in range(num_simulations)],
+                dtype=np.float64,
+            )
```
verveerpj, line 446: This assumes that `cached_results` contains all results and they are the expected results. I do not think that is garantueed, see my comments below at the point where the cached values are retrieved.  If not all values are present, this will fail, if wrong values are present they will end up in the results.

src/ert/run_models/everest_run_model.py
```diff
@@ -420,44 +420,89 @@ def _on_before_forward_model_evaluation(
...
+        else:
+            return np.array(
+                [
+                    cached_results.get(sim, simulated_results[sim])
```
verveerpj, line 450: Same as above, it is assumed that `cached_results` has valid values.

src/ert/run_models/everest_run_model.py
```diff
@@ -420,44 +420,89 @@ def _on_before_forward_model_evaluation(
...
+        active_sim_ids: list[bool] | None = None,
     ) -> tuple[NDArray[np.float64], NDArray[np.float64] | None, list[int]]:
         # Reset the current run status:
         self._status = None
 
         # Get cached_results:
-        cached_results = self._get_cached_results(control_values, model_realizations)
+        cached_objectives, cached_constraints = self._get_cached_results_by_sim_id(
+            control_values, model_realizations
+        )
```
verveerpj, line 468: This does not take `active_sim_ids` into account. This means that the cache may contain values that should not be used. Or maybe they do not contain those values, but since these are also not evaluated, `_consolidate_cache_and_sim_results` might fail if it tries to construct a result from cached results only (there will be too few results). Solution would be not to cache anything that has a `False` entry in `active_sim_ids`. `_consolidate_cache_and_sim_results` would need to be adapted to make sure that the results have the right size and have zero entries in the correspoding places.

src/ert/run_models/everest_run_model.py
```diff
@@ -433,44 +433,89 @@ def _on_before_forward_model_evaluation(
             logging.getLogger(EVEREST).info("User abort requested.")
             optimizer.abort_optimization()
 
+    @staticmethod
+    def _consolidate_cache_and_sim_results(
```
verveerpj, line 437: Maybe it is better to split this in two methods?

src/ert/run_models/everest_run_model.py
verveerpj: I think you can do `control_values[active_sim_ids, :]` to filter out the unwanted values.

verveerpj: need to filter `model_realizations` too, I guess, so maybe not convient.

verveerpj (approved):

yngve-sk:
Dropping renames/simulator cache simplification and saving for other PR, there are some upstream code changes that need to happen before this can be simplified sensibly

PR #10102, opened 2025-02-18
Fix muted exceptions in fm_dispatch due to self SIGKILL

### LINKED ISSUE #9989: Exception is never raised in fm_dispatch due to self-kill
The `os.killpg` kills the fm_dispatch process such that the exception `e` is never raised. If there is a bug in fm_dispatch that triggers an exception, it will not be printed to the terminal. Running `fm_dispatch` in the runpath only gives you the output `Terminated`. 

https://github.com/equinor/ert/blob/68271ad439e6bb116f5df0f453f88abf4dbc0b8a/src/_ert/forward_model_runner/fm_dispatch.py#L194-L199

jonathan-eq:
**Issue**
Resolves #9989

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/_ert/forward_model_runner/fm_dispatch.py
+++ b/src/_ert/forward_model_runner/fm_dispatch.py
@@ -144,11 +144,6 @@ def _stop_reporters(
             reporter.stop(exited_event=exited_event)
 
 
-def sigterm_handler(_signo, _stack_frame):
-    signal.signal(signal.SIGTERM, signal.SIG_DFL)
-    os.kill(0, signal.SIGTERM)
-
-
 def fm_dispatch(args):
     parser = argparse.ArgumentParser(
         description=(
@@ -207,13 +202,14 @@ def fm_dispatch(args):
 
 def main():
     os.nice(19)
-    signal.signal(signal.SIGTERM, sigterm_handler)
     try:
         fm_dispatch(sys.argv)
-    except Exception as e:
+    except Exception as exc:
+        print(f"fm_dispatch failed with {exc=}")
         pgid = os.getpgid(os.getpid())
-        os.killpg(pgid, signal.SIGTERM)
-        raise e
+        os.killpg(
+            pgid, signal.SIGTERM
+        )  # This will trigger the sigterm_handler, which shuts down the reporters and SIGKILLS any remaining processes.
 
 
 if __name__ == "__main__":
--- a/tests/ert/unit_tests/forward_model_runner/test_fm_dispatch.py
+++ b/tests/ert/unit_tests/forward_model_runner/test_fm_dispatch.py
@@ -484,3 +484,20 @@ time.sleep(180)"""
             event_from_json(zmq_server.messages[-1]).error_msg
             == FORWARD_MODEL_TERMINATED_MSG
         )
+
+
+async def test_fm_dispatch_main_signals_sigterm_on_exception(capsys):
+    def mock_fm_dispatch_raises(*args):
+        raise RuntimeError("forward model critical error")
+
+    with (
+        patch("_ert.forward_model_runner.fm_dispatch.os.killpg") as mock_killpg,
+        patch("_ert.forward_model_runner.fm_dispatch.os.getpgid") as mock_getpgid,
+        patch("_ert.forward_model_runner.fm_dispatch.fm_dispatch") as mock_fm_dispatch,
+    ):
+        mock_getpgid.return_value = 17
+        mock_fm_dispatch.side_effect = mock_fm_dispatch_raises
+        _ert.forward_model_runner.fm_dispatch.main()
+    assert "forward model critical error" in capsys.readouterr().out
+
+    mock_killpg.assert_called_with(17, signal.SIGTERM)


src/_ert/forward_model_runner/fm_dispatch.py
```diff
@@ -127,13 +127,7 @@ def _report_all_messages(
...
                     )
         if isinstance(msg, Finish) and not msg.success():
-            _stop_reporters_and_sigkill(reporters)
-
-
-def _stop_reporters_and_sigkill(reporters, exited_event: Exited | None = None):
-    _stop_reporters(reporters, exited_event)
-    pgid = os.getpgid(os.getpid())
-    os.killpg(pgid, signal.SIGKILL)
+            os.kill(os.getpid(), signal.SIGTERM)
```
berland, line 130: the  `g` part of `os.killpg` is quite important. It ensures that all subprocesses of fm_dispatch are taken down as well (no matter what). Otherwise we risk that runaway processes are lingering on compute nodes and consume resources.

jonathan-eq: This was only to trigger the sigterm_handler function, which further stops the reporters, and signals SIGQUIT to the running fm_steps. When they see SIGQUIT, they call process.kill. I tried avoiding SIGKILL, as that would make us unable to reraise the caught exception.

src/_ert/forward_model_runner/fm_dispatch.py
```diff
@@ -210,10 +210,10 @@ def main():
     signal.signal(signal.SIGTERM, sigterm_handler)
     try:
         fm_dispatch(sys.argv)
-    except Exception as e:
+    except Exception as exc:
+        print(f"fm_dispatch failed with {exc=}")
```
xjules, line 214: I guess this is a forgotten print

src/_ert/forward_model_runner/fm_dispatch.py
```diff
@@ -210,10 +210,10 @@ def main():
...
         pgid = os.getpgid(os.getpid())
         os.killpg(pgid, signal.SIGTERM)
```
berland, line 216: Add a comment here that this line kills ourselves and all our descendants (and yielding some exit code)

jonathan-eq: I am not sure if this is the line that directly kills ourselves and all our descendants. This signals SIGTERM, which should be handled by the sigterm handler in `fm_dispatch()` (thats the one that SIGKILLs), but that won't happen if we get an exception before creating that signal handler. Before then, we won't have any descendants, so it should be fine.

src/_ert/forward_model_runner/fm_dispatch.py
```diff
@@ -210,10 +210,12 @@ def main():
...
         fm_dispatch(sys.argv)
-    except Exception as e:
+    except Exception as exc:
+        print(f"fm_dispatch failed with {exc=}")
         pgid = os.getpgid(os.getpid())
-        os.killpg(pgid, signal.SIGTERM)
-        raise e
+        os.killpg(
+            pgid, signal.SIGTERM
+        )  # This will trigger the sigterm_handler, which shuts down the reporters and SIGKILLS any remaining processes.
```
xjules, line 218: Just wondering if you want to write a similar test for this behavior (if already not in place) as this one: 
https://github.com/equinor/ert/blob/47934489e6cceedfe6745348d62c31703aa47cd3/tests/ert/unit_tests/forward_model_runner/test_fm_dispatch.py#L34

src/_ert/forward_model_runner/fm_dispatch.py
xjules: Also shouldn't the exception be logged before we start autokill?

jonathan-eq: Yes, that is what this PR is doing

src/_ert/forward_model_runner/fm_dispatch.py
xjules: That test only covers terminate experiment functionality, while this code is trigger only when job raises exception. at least that's how I understand it.

jonathan-eq: I added a test :)

tests/ert/unit_tests/forward_model_runner/test_fm_dispatch.py
```diff
@@ -484,3 +484,20 @@ async def wait_for_msg(msg_type):
             event_from_json(zmq_server.messages[-1]).error_msg
             == FORWARD_MODEL_TERMINATED_MSG
         )
+
+
+async def test_fm_dispatch_main_signals_sigterm_on_exception(capsys):
+    def mock_fm_dispatch_raises(*args):
+        raise RuntimeError("foo bar")
```
xjules, line 491: Could you use something else than "foo bar"  ; eg. "forward model critical error" or something :)

tests/ert/unit_tests/forward_model_runner/test_fm_dispatch.py
```diff
@@ -484,3 +484,20 @@ async def wait_for_msg(msg_type):
             event_from_json(zmq_server.messages[-1]).error_msg
             == FORWARD_MODEL_TERMINATED_MSG
         )
+
+
+async def test_fm_dispatch_main_signals_sigterm_on_exception(capsys):
```
xjules, line 489: nice test anyway!

xjules (approved):
Nice one @jonathan-eq ! :rocket:

PR #10112, opened 2025-02-20
Move active/cached controls logic out of run_forward_model

yngve-sk:
Moving of some logic related to discussion in 
https://github.com/equinor/ert/pull/10081

### DIFF
--- a/src/ert/run_models/everest_run_model.py
+++ b/src/ert/run_models/everest_run_model.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import dataclasses
 import datetime
 import functools
 import json
@@ -9,7 +10,7 @@ import queue
 import shutil
 from collections import defaultdict
 from collections.abc import Callable
-from enum import IntEnum
+from enum import IntEnum, auto
 from pathlib import Path
 from types import TracebackType
 from typing import TYPE_CHECKING, Any, Protocol
@@ -97,6 +98,23 @@ class EverestExitCode(IntEnum):
     MAX_FUNCTIONS_REACHED = 3
     MAX_BATCH_NUM_REACHED = 4
     USER_ABORT = 5
+
+
+class _EvaluationStatus(IntEnum):
+    TO_SIMULATE = auto()
+    CACHED = auto()
+    INACTIVE = auto()
+
+
+@dataclasses.dataclass
+class _EvaluationInfo:
+    control_vector: NDArray[np.float64]
+    status: _EvaluationStatus
+    flat_index: int
+    simulation_id: int | None
+    model_realization: int
+    objectives: NDArray[np.float64] | None = None
+    constraints: NDArray[np.float64] | None = None
 
 
 class EverestRunModel(BaseRunModel):
@@ -315,7 +333,7 @@ class EverestRunModel(BaseRunModel):
                 )
             )
 
-            objectives, constraints, _ = self._run_forward_model(
+            objectives, constraints = self._run_forward_model(
                 np.repeat(np.expand_dims(control_variables, axis=0), nreal, axis=0),
                 model_realizations,
             )
@@ -435,81 +453,25 @@ class EverestRunModel(BaseRunModel):
         self,
         control_values: NDArray[np.float64],
         model_realizations: list[int],
-        active_control_vectors: list[bool] | None = None,
-    ) -> tuple[NDArray[np.float64], NDArray[np.float64] | None, list[int]]:
+    ) -> tuple[NDArray[np.float64], NDArray[np.float64] | None]:
         # Reset the current run status:
         self._status = None
 
-        # Get cached_results:
-        cached_results = self._get_cached_results(control_values, model_realizations)
-
-        # Collect the indices of the controls that must be evaluated in the batch:
-        evaluated_control_indices = [
-            idx
-            for idx in range(control_values.shape[0])
-            if idx not in cached_results
-            and (active_control_vectors is None or active_control_vectors[idx])
-        ]
-
-        if not evaluated_control_indices:
-            cached_results = self._get_cached_results(
-                control_values, model_realizations
-            )
-
-            objectives = np.zeros(
-                (
-                    control_values.shape[0],
-                    len(self._everest_config.objective_names),
-                ),
-                dtype=float64,
-            )
-
-            constraints = (
-                np.zeros(
-                    (
-                        control_values.shape[0],
-                        len(self._everest_config.constraint_names),
-                    ),
-                    dtype=float64,
-                )
-                if self._everest_config.constraint_names
-                else None
-            )
-
-            for control_idx, (
-                cached_objectives,
-                cached_constraints,
-            ) in cached_results.items():
-                objectives[control_idx, ...] = cached_objectives
-                if constraints is not None:
-                    assert cached_constraints is not None
-                    constraints[control_idx, ...] = cached_constraints
-
-                # Increase the batch ID for the next evaluation:
-            self._batch_id += 1
-
-            # Return the results, together with the indices of the evaluated controls:
-            return objectives, constraints, evaluated_control_indices
-
         # Create the batch to run:
-        sim_controls = self._create_simulation_controls(
-            control_values, evaluated_control_indices
-        )
+        sim_controls = self._create_simulation_controls(control_values)
 
         # Initialize a new ensemble in storage:
         assert self._experiment is not None
         ensemble = self._experiment.create_ensemble(
             name=f"batch_{self._batch_id}",
-            ensemble_size=len(evaluated_control_indices),
+            ensemble_size=control_values.shape[0],
             iteration=self._batch_id,
         )
         for sim_id, controls in enumerate(sim_controls.values()):
             self._setup_sim(sim_id, controls, ensemble)
 
         # Evaluate the batch:
-        run_args = self._get_run_args(
-            ensemble, model_realizations, evaluated_control_indices
-        )
+        run_args = self._get_run_args(ensemble, model_realizations)
         self._context_env.update(
             {
                 "_ERT_EXPERIMENT_ID": str(ensemble.experiment_id),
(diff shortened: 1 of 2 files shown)


src/ert/run_models/everest_run_model.py
```diff
@@ -99,6 +100,23 @@ class EverestExitCode(IntEnum):
     USER_ABORT = 5
 
 
+class _EvaluationStatus(StrEnum):
+    TO_SIMULATE = "TO_SIMULATE"
+    CACHED = "CACHED"
+    INACTIVE = "INACTIVE"
```
verveerpj, line 106: Do we need string enums?

src/ert/run_models/everest_run_model.py
```diff
@@ -542,7 +503,65 @@ def _run_forward_model(
...
+        return objectives, constraints
+
+    @staticmethod
+    def _create_evaluation_infos(
+        control_values: NDArray[np.float64],
+        model_realizations: list[int],
+        active_controls: list[bool],
+        simulator_cache: SimulatorCache,
+        inactive_objective_fill_value: NDArray[np.float64],
+        inactive_constraints_fill_value: NDArray[np.float64],
```
verveerpj, line 515: Might as well hard-code somewhere that these should be zero.

src/ert/run_models/everest_run_model.py
```diff
@@ -542,7 +503,65 @@ def _run_forward_model(
...
+                        model_realization=model_realization,
+                        flat_index=flat_index,
+                        simulation_id=None,
+                        objectives=inactive_objective_fill_value,
+                        constraints=inactive_constraints_fill_value,
+                    )
+                )
+                continue
+
+            cached_result = simulator_cache.find(control_vector)
```
verveerpj, line 538: I think this needs the `model_realization` as an argument, since the cache stores values by realization and control values.

src/ert/run_models/everest_run_model.py
```diff
@@ -566,24 +585,87 @@ def _forward_model_evaluator(
...
+                if c.status == _EvaluationStatus.TO_SIMULATE
+            ],
+            dtype=np.float64,
+        )
+
+        model_realizations_to_simulate = [
+            c.model_realization
+            for c in evaluation_infos
+            if c.status == _EvaluationStatus.TO_SIMULATE
+        ]
```
verveerpj, line 615: Could move this in the `if` below?

src/ert/run_models/everest_run_model.py
```diff
@@ -566,24 +585,87 @@ def _forward_model_evaluator(
...
+                control_values=control_values_to_simulate,
+                model_realizations=model_realizations_to_simulate,
+            )
+        else:
+            sim_objectives = np.array([], dtype=np.float64)
+            sim_constraints = np.array([], dtype=np.float64)
+
+            # Note: removing this causes some tests to fail, but this behavior
+            # should maybe be revised?
+            self._batch_id += 1
```
verveerpj, line 628: The batch id must be increment, even if there were no simulations, otherwise we get very different behavior. But it is indeed now a bit unclear where and when this happens.

src/ert/run_models/everest_run_model.py
```diff
@@ -779,31 +857,19 @@ def _get_objectives_and_constraints(
...
     def _get_simulation_results(
         results: list[dict[str, NDArray[np.float64]]],
         names: list[str],
         controls: NDArray[np.float64],
-        evaluated_control_indices: list[int],
     ) -> NDArray[np.float64]:
         values = np.zeros((controls.shape[0], len(names)), dtype=float64)
         for func_idx, name in enumerate(names):
-            values[evaluated_control_indices, func_idx] = np.fromiter(
+            values[range(len(controls)), func_idx] = np.fromiter(
```
verveerpj, line 872: probably could be `values[:, func_idx] = ...`

src/ert/run_models/everest_run_model.py
```diff
@@ -435,81 +453,25 @@ def _run_forward_model(
...
-            control_values, evaluated_control_indices
-        )
+        sim_controls = self._create_simulation_controls(control_values)
 
         # Initialize a new ensemble in storage:
         assert self._experiment is not None
         ensemble = self._experiment.create_ensemble(
             name=f"batch_{self._batch_id}",
-            ensemble_size=len(evaluated_control_indices),
+            ensemble_size=len(control_values),
```
verveerpj, line 467: prefer: `control_values.shape[0]`, as it is a matrix.

src/ert/run_models/everest_run_model.py
```diff
@@ -435,81 +453,25 @@ def _run_forward_model(
...
             iteration=self._batch_id,
         )
         for sim_id, controls in enumerate(sim_controls.values()):
             self._setup_sim(sim_id, controls, ensemble)
 
         # Evaluate the batch:
-        run_args = self._get_run_args(
-            ensemble, model_realizations, evaluated_control_indices
-        )
+        run_args = self._get_run_args(ensemble, model_realizations, len(control_values))
```
verveerpj, line 474: prefer: `control_values.shape[0]`, as it is a matrix.

src/ert/run_models/everest_run_model.py
```diff
@@ -566,24 +585,87 @@ def _forward_model_evaluator(
...
+            dtype=np.float64,
+        )
+
+        model_realizations_to_simulate = [
+            c.model_realization
+            for c in evaluation_infos
+            if c.status == _EvaluationStatus.TO_SIMULATE
+        ]
+
+        if len(control_values_to_simulate) > 0:
```
verveerpj, line 617: prefer: `control_values_to_simulate.shape[0]`, as it is a matrix.

src/ert/run_models/everest_run_model.py
```diff
@@ -635,7 +716,7 @@ def _create_control_dicts_for_simulation(
             sim_id: _create_control_dicts_for_simulation(
                 self._everest_config.controls, control_values[sim_id, :]
             )
-            for sim_id in controls_to_evaluate
+            for sim_id in range(len(control_values))
```
verveerpj, line 719: prefer: `range(control_values.shape[0])`, as it is a matrix.

src/ert/run_models/everest_run_model.py
```diff
@@ -813,12 +879,11 @@ def _add_results_to_cache(
         self,
         control_values: NDArray[np.float64],
         model_realizations: list[int],
-        evaluated_control_indices: list[int],
         objectives: NDArray[np.float64],
         constraints: NDArray[np.float64] | None,
     ) -> None:
         if self._simulator_cache is not None:
-            for sim_id in evaluated_control_indices:
+            for sim_id in range(len(control_values)):
```
verveerpj, line 886: prefer: `control_values.shape[0]`, as it is a matrix.

src/ert/run_models/everest_run_model.py
```diff
@@ -948,6 +1013,16 @@ def add(
             ),
         )
 
+    def find(
```
verveerpj, line 1016: This will return the first realization, that has the right control values. However, the cached is indexed by realization, because there are multiple realizations with the same controls. So, you probably need to pass the realization you want and first index into the cache before searching. Actually `get` below does that.

verveerpj (commented):
I have a number of comments, and may have pinpointed the reason for getting different values. I will have another closer look tomorrow.

src/ert/run_models/everest_run_model.py
yngve-sk: Hmm not really,  will go for IntEnum as is done elsewhere

src/ert/run_models/everest_run_model.py
yngve-sk: I will pass it through and use .get(), though I think it should be revised in a future PR, since in theory the model realization has no effect on the outputs generated by the forward model run, only the controls themselves, so I think it should be indexed by control values only.

verveerpj: What do you mean? Different models will definitely generate different ouput for the same controls, that is why we run multiple realizations in the first place.

src/ert/run_models/everest_run_model.py
yngve-sk: Seems the last arg could be removed from this function

src/ert/run_models/everest_run_model.py
yngve-sk: Ohh.. I think I missed that then, the controls we feed into ROPT, `control_values` are only the changing controls, but we also give some other static controls tied to the geo_realizations as ext params to ERT?

src/ert/run_models/everest_run_model.py
```diff
@@ -542,7 +503,67 @@ def _run_forward_model(
...
+        model_realizations: list[int],
+        active_controls: list[bool],
+        simulator_cache: SimulatorCache,
+        num_objectives: int,
+        num_constraints: int,
+    ) -> list[_EvaluationInfo]:
+        inactive_objective_fill_value: NDArray[np.float64] = np.zeros(num_objectives)
+        inactive_constraints_fill_value: NDArray[np.float64] = np.zeros(num_constraints)
+        evaluation_infos_oo = []  # object oriented (should be more readable)
+        #  control_infos_do = {}  # data oriented (mostly paradigm within ropt/everest)
```
verveerpj, line 520: Fix this variable name and comments?

src/ert/run_models/everest_run_model.py
verveerpj: Correct!

src/ert/run_models/everest_run_model.py
verveerpj: Any ideas about this? I had a look but I am not sure how to make this clearer. We could leave it for now.

verveerpj (approved):
Still some minor comments, but I think it is looking good.

src/ert/run_models/everest_run_model.py
yngve-sk: I think we should address it some other time, as it will probably change a lot of snapshots/tests if we decide to not increment the batch IDs on full cache hits. Would opt for just not incrementing it I guess?.. Depends on what we want to show in the user also, if we want to show a batch of cache hits as a batch in itself, or just omit it, not really sure

verveerpj: I think we should increment it. Unfortunately, users have gotten used to associate batch numbers with the progress of the optimizer, so if suddenly batches are not reported that could lead to confusion.

yngve-sk: For the GUI I think it will be nice to see that a batch is a cache hit actually, it'll make it more clear what's happening. Like when we do the sampling for autoscaling, the GUI would ideally show that the batch was evaluated instantly due to a cache hit, so at some point we might want to emit some kind of event when that happens, as we don't get any `BaseRunModel` type events that indicate batch progress from those.

PR #10119, opened 2025-02-20
Format logged config dict to enhance readability

### LINKED ISSUE #9315: Log the arguments to each forward model step
When debugging Ert runs, the contents of the config dict is logged, but truncated in size. This prohibits usage and hides important details.

This print of the config dict is also close to unreadable as it is all printed on one long line.

Some fix is needed to make this parseable and readable, calling arguments of forward model steps is interesting f.ex.

HakonSohoel:
Previously some parts of the config dict has been filtered out from the log message to make it fit within the 32768 message size limit of App Insights (among others forward model steps). These parts should be kept, and instead the config dict logging is split into multiple messages when it exceeds the message size limit of App Insights.  

With this commit the config dict log message(s) will look as follows:
```
2025-02-20 09:50:36.539 | Content of the config_dict (part 1/2): {   
  "DEFINE": [     
    [
      "<CONFIG_PATH>",
      "/var/folders/mg/jwjt9mcj02l1cqy_57szt4hc0000gp/T/tmph0lhx8yc"
    ],...

2025-02-20 09:50:36.540 | Content of the config_dict (part 2/2): 
    <Continues where part 1 stopped>
...
```

**Issue**
Resolves #9315

**Approach**
Log config dict as readable json
Split message if it exceeds the message size limit in App Insights in order to keep all the content

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/ert_config.py
+++ b/src/ert/config/ert_config.py
@@ -917,7 +917,15 @@ class ErtConfig:
         # In doing do, the message lenght will typically increase a bit. To Avoid hiting the App Insights' hard limit of message length, the limit is set to 80%
         # of MAX_MESSAGE_LENGTH_APP_INSIGHTS = 32768
         SAFE_MESSAGE_LENGTH_LIMIT = 26214  # <= MAX_MESSAGE_LENGTH_APP_INSIGHTS * 0.8
-        config_dict_content = json.dumps(content_dict, indent=2, cls=ContextBoolEncoder)
+        try:
+            config_dict_content = json.dumps(
+                content_dict, indent=2, cls=ContextBoolEncoder
+            )
+        except Exception as err:
+            config_dict_content = str(content_dict)
+            logger.warning(
+                f"Logging of config dict could not be formatted for enhanced readability. {err}"
+            )
         config_dict_content_length = len(config_dict_content)
         if config_dict_content_length > SAFE_MESSAGE_LENGTH_LIMIT:
             config_sections = _split_string_into_sections(
--- a/src/ert/config/ert_config.py
+++ b/src/ert/config/ert_config.py
@@ -913,16 +913,15 @@ class ErtConfig:
 
     @classmethod
     def _log_config_dict(cls, content_dict: dict[str, Any]) -> None:
-        MAX_MESSAGE_LENGTH_APP_INSIGHTS = 32768
-        max_safe_message_length = int(
-            MAX_MESSAGE_LENGTH_APP_INSIGHTS * 0.8
-        )  # The content of the message is sanitized before beeing sendt to App Insigths to make sure GDPR-rules are not violated.
+        # The content of the message is sanitized before beeing sendt to App Insigths to make sure GDPR-rules are not violated.
         # In doing do, the message lenght will typically increase a bit. To Avoid hiting the App Insights' hard limit of message length, the limit is set to 80%
+        # of MAX_MESSAGE_LENGTH_APP_INSIGHTS = 32768
+        SAFE_MESSAGE_LENGTH_LIMIT = 26214  # <= MAX_MESSAGE_LENGTH_APP_INSIGHTS * 0.8
         config_dict_content = json.dumps(content_dict, indent=2, cls=ContextBoolEncoder)
         config_dict_content_length = len(config_dict_content)
-        if config_dict_content_length > max_safe_message_length:
+        if config_dict_content_length > SAFE_MESSAGE_LENGTH_LIMIT:
             config_sections = _split_string_into_sections(
-                config_dict_content, max_safe_message_length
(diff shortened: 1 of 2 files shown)


src/ert/config/ert_config.py
```diff
@@ -1089,6 +1101,23 @@ def _create_observations(
         return EnkfObs(obs_vectors, obs_time_list)
 
 
+def _split_string_into_sections(input: str, section_length: int) -> list[str]:
```
berland, line 1104: Would it be an idea to split on newlines for readability?

src/ert/config/ert_config.py
```diff
@@ -911,14 +913,24 @@ def _log_config_file(cls, config_file: str, config_file_contents: str) -> None:
...
+        config_dict_content = json.dumps(content_dict, indent=2)
+        config_dict_content_length = len(config_dict_content)
+        if config_dict_content_length > max_safe_message_length:
+            config_sections = _split_string_into_sections(
+                config_dict_content, max_safe_message_length
+            )
+            section_count = len(config_sections)
+            for i, section in enumerate(config_sections):
+                logger.info(
+                    f"Content of the config_dict (part {i + 1}/{section_count}): {section}"
```
berland, line 930: Is it trivial to reassemble the config programmatically? I guess it is, splitting by the first colon and just concatenating the data.

src/ert/config/ert_config.py
```diff
@@ -911,14 +913,24 @@ def _log_config_file(cls, config_file: str, config_file_contents: str) -> None:
...
-        tmp_dict.pop("LOAD_WORKFLOW", None)
-        tmp_dict.pop("LOAD_WORKFLOW_JOB", None)
-        tmp_dict.pop("HOOK_WORKFLOW", None)
-        tmp_dict.pop("WORKFLOW_JOB_DIRECTORY", None)
-
-        logger.info(f"Content of the config_dict: {tmp_dict}")
+        MAX_MESSAGE_LENGTH_APP_INSIGHTS = 32768
+        max_safe_message_length = int(
+            MAX_MESSAGE_LENGTH_APP_INSIGHTS * 0.8
+        )  # The content of the message is sanitized by the FMUAzureLogHandler before beeing sendt to App Insigths to make sure GDPR-rules are not violated.
```
berland, line 919: FMUAzureLogHandler is a private repo so externals reading the source code can not figure out what it is. Just say that we shorten it to allow for some log handler massaging or similar.

HakonSohoel: Good point. Removed mention of FMUAzureLogHandler 👌

src/ert/config/ert_config.py
HakonSohoel: Yes, that should be trivial as you describe 😎

src/ert/config/ert_config.py
HakonSohoel: Would be nice. I suggest to save it for a future improvement if it bothers us when we work with the logs.

src/ert/config/ert_config.py
```diff
@@ -1089,6 +1102,23 @@ def _create_observations(
         return EnkfObs(obs_vectors, obs_time_list)
 
 
+def _split_string_into_sections(input: str, section_length: int) -> list[str]:
```
andreas-el, line 1105: Could you also provide a test or two for this functionality?
The `_split_string_into_sections` shouldn't be too hard, you could also mock the limitation down to something more practical when testing.

src/ert/config/ert_config.py
```diff
@@ -1089,6 +1102,23 @@ def _create_observations(
...
+    """
+    Splits a string into sections of length section_length
+    and returns it as an list.
+    """
```
andreas-el, line 1109: Have you considered using `textwrap.wrap()` for doing this?

Or list comprehension?
```
s = "abcdefghij"
k = 3
chunks = [s[i:i+k] for i in range(0, len(s), k)]
```

The `if section_length < 1` could be `max(section_length, 1)`

HakonSohoel: Yes, list comprehension is a lot nicer 😄

src/ert/config/ert_config.py
HakonSohoel: Sure 👌

andreas-el (changes requested):
Please provide a better descriptive PR title as this might be ending up in the documentation users will read.
Does the PR make the logs more readable? There is no example of how this will result?
As far as I can tell, the config dict is more complete with this PR, and we overcome the limitation set by Azure by sending parts of the content if need be.

Please specify release label also.

Good work! ☀️

andreas-el (approved):
☀️ Great work!

PR #10147, opened 2025-02-25
Add test for the propagation of queue_options from ertconfig to cluster

### LINKED ISSUE #10141: Add E2E tests for queue options
There was an issue where one of the queue options for the slurm driver were never propagated (https://github.com/equinor/ert/issues/10124). We have no realistic end-to-end tests for this, which would have caught this.
The test should run poly_example, and make sure the queue_options are propagated all the way from ert config to the shell command.

jonathan-eq:
**Issue**
Resolves #10141

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/scheduler/bin/bhist.py
+++ b/tests/ert/unit_tests/scheduler/bin/bhist.py
@@ -83,18 +83,16 @@ def main() -> None:
 
     jobs_output: list[Job] = []
     for job in args.jobs:
-        job_name: str = read(jobs_path / f"{job}.name") or "_"
+        job_name: str = read(jobs_path / job / "name") or "_"
         assert job_name is not None
 
-        submit_time_millis: int = int(
-            os.path.getctime(jobs_path / f"{job}.name") * 1000
-        )
+        submit_time_millis: int = int(os.path.getctime(jobs_path / job / "name") * 1000)
         pending_time_millis = int(read(jobs_path / "pendingtimemillis") or 0)
         run_start_time_millis: int = submit_time_millis + pending_time_millis
         end_time_millis: int = int(time.time() * 1000)
-        if (jobs_path / f"{job}.returncode").exists():
+        if (jobs_path / job / "returncode").exists():
             end_time_millis = int(
-                os.path.getctime(jobs_path / f"{job}.returncode") * 1000
+                os.path.getctime(jobs_path / job / "returncode") * 1000
             )
             if not args.l:
                 print("bhist says job is done")
--- a/tests/ert/unit_tests/scheduler/test_lsf_driver.py
+++ b/tests/ert/unit_tests/scheduler/test_lsf_driver.py
@@ -19,6 +19,8 @@ from hypothesis import given
 from hypothesis import strategies as st
 
 from ert.config import QueueConfig
+from ert.config.queue_config import _parse_realization_memory_str
+from ert.mode_definitions import ENSEMBLE_EXPERIMENT_MODE
 from ert.scheduler import LsfDriver, create_driver
 from ert.scheduler.driver import SIGNAL_OFFSET
 from ert.scheduler.lsf_driver import (
@@ -39,6 +41,7 @@ from ert.scheduler.lsf_driver import (
     parse_bjobs,
     parse_bjobs_exec_hosts,
 )
+from tests.ert.ui_tests.cli.run_cli import run_cli
 from tests.ert.utils import poll, wait_until
 
 from .conftest import mock_bin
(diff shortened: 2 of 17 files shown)


JHolba (changes requested):

JHolba (approved):
the new tests are good :)

tests/ert/unit_tests/scheduler/bin/bhist.py
```diff
@@ -83,18 +83,16 @@ def main() -> None:
 
     jobs_output: list[Job] = []
     for job in args.jobs:
-        job_name: str = read(jobs_path / f"{job}.name") or "_"
+        job_name: str = read(jobs_path / job / "name") or "_"
```
berland, line 86: Are these changes related to "Add tests for the propagation"  or does these changes serve some other purpose (was looking for a longer description in the commit message but could not find it).

(two years down the road we will be looking at the git blame for this line and dont get why this was changed based on reading the log message)

jonathan-eq: This was just refactoring. In the cases we run multiple realizations with the mocked binaries, I think it makes sense that each job has its own directory instead of prefixing the files with the job_id.

jonathan-eq: Updated the commit message to include this^

berland: If refactoring, it would make sense to have it in a separate commit.

tests/ert/unit_tests/scheduler/test_lsf_driver.py
```diff
@@ -1364,3 +1369,65 @@ async def finished(iens: int, returncode: int):
     # a controlled fashion:
     if (tmp_path / "trap_handle_installed").exists():
         wait_until((tmp_path / "was_killed").exists, timeout=4)
+
+
+@pytest.mark.integration_test
+@pytest.mark.usefixtures("copy_poly_case")
+def test_queue_options_are_propagated_from_config_to_bsub(monkeypatch):
+    """
+    This end to end test is here to verify that queue_options are correctly propagated all the way from ert config to the cluster.
```
berland, line 1378: (this could be line formatted, I speculate it extends 88 characters)

tests/ert/unit_tests/scheduler/test_lsf_driver.py
```diff
@@ -1364,3 +1369,65 @@ async def finished(iens: int, returncode: int):
...
+                QUEUE_OPTION LSF LSF_QUEUE {expected_queue}
+                QUEUE_OPTION LSF LSF_RESOURCE {expected_resource_string}
+                QUEUE_OPTION LSF PROJECT_CODE {expected_project_code}
+                QUEUE_OPTION LSF EXCLUDE_HOST {expected_excluded_hosts}
+                NUM_REALIZATIONS 1
+                """
+            )
+        )
+    run_cli(ENSEMBLE_EXPERIMENT_MODE, "--disable-monitoring", "poly.ert")
+    mock_jobs_dir = Path(f"{os.environ.get('PYTEST_TMP_PATH')}/mock_jobs")
```
berland, line 1404: it smells a little bit to use this environment variable. Will capturing_bsub solve it or can't you find it from `cwd`?

tests/ert/unit_tests/scheduler/test_openpbs_driver.py
```diff
@@ -606,3 +608,58 @@ def test_openpbs_driver_with_poly_example_failing_poll_fails_ert_and_propagates_
...
+                NUM_REALIZATIONS 1
+                """
+            )
+        )
+    run_cli(ENSEMBLE_EXPERIMENT_MODE, "--disable-monitoring", "poly.ert")
+    mock_jobs_dir = Path(f"{os.environ.get('PYTEST_TMP_PATH')}/mock_jobs")
+    job_dir = next(
+        mock_jobs_dir.iterdir()
+    )  # There is only one realization in this test
+    complete_command_invocation = (job_dir / "complete_command_invocation").read_text(
```
berland, line 644: check `capturing_qsub`

tests/ert/unit_tests/scheduler/test_openpbs_driver.py
```diff
@@ -606,3 +608,58 @@ def test_openpbs_driver_with_poly_example_failing_poll_fails_ert_and_propagates_
...
+
+    args = shlex.split(complete_command_invocation)
+    _ = args.pop(0)  # script path
+    opts, _ = getopt.getopt(args, "N:r:o:e:l:q:A:")
+    parsed_options = dict(opts)
+
+    assert parsed_options.get("-q") == expected_queue
+    assert parsed_options.get("-A") == expected_project_code
+
+    # -l was not parsed correctly by getopt, so we read it manually instead.
```
berland, line 656: Do you really need to parse the command line or just rely on certain strings being present in the command?

tests/ert/unit_tests/scheduler/test_openpbs_driver.py
berland: But `scheduler/bin/bsub.py` is just a mock, so this is not an end-to-end test in any case. 

If the real `bsub` command from LSF gets a breaking change requiring a change in our code, this code will not catch it, so I am unsure if it adds value to this test to use `scheduler/bin/bsub.py` over `capturing_bsub`.

tests/ert/unit_tests/scheduler/test_openpbs_driver.py
jonathan-eq: Yes, you are right. Very good!

tests/ert/unit_tests/scheduler/test_openpbs_driver.py
jonathan-eq: We have some unit tests for this which I could implement for all queue_options like this
```
@pytest.mark.usefixtures("capturing_bsub")
async def test_submit_with_project_code():
    queue_config_dict = {
        "QUEUE_SYSTEM": "LSF",
        "FORWARD_MODEL": [("FLOW",), ("ECLIPSE",), ("RMS",)],
    }
    queue_config = QueueConfig.from_dict(queue_config_dict)
    driver: LsfDriver = create_driver(queue_config.queue_options)
    await driver.submit(0, "sleep")
    assert f"-P {queue_config.queue_options.project_code}" in Path(
        "captured_bsub_args"
    ).read_text(encoding="utf-8")
```
But this part might not be accurate to how we do it in `ensemble.py`, and any breaking changes there would not be picked up by the tests. 
```
queue_config = QueueConfig.from_dict(queue_config_dict)
driver: LsfDriver = create_driver(queue_config.queue_options)
```

tests/ert/unit_tests/scheduler/test_lsf_driver.py
```diff
@@ -223,7 +239,7 @@ async def test_submit_sets_stderr():
 
 
 @pytest.mark.usefixtures("capturing_bsub")
-async def test_submit_with_realization_memory_with_bsub_capture():
+async def test_submit_with_realization_memory_with_bsub_capture():  # JONAK - CAN THIS BE REMOVED?
```
jonathan-eq, line 242: This is a unit test, but we are testing the same thing in an integration test. Should we keep both as the integration_tests are not run with `just rapid-tests`?

xjules:
General note: it looks to be very complicated. I thought something similar to this test?
https://github.com/equinor/ert/blob/7642464403bd509255dad43ccc89dafba1a58fde/tests/ert/unit_tests/test_tracking.py#L147
Which essentially create dedicated models and check that the scheduler got the parameters correctly.

PR #10158, opened 2025-02-26
Better error message when experiment is terminated by user

### LINKED ISSUE #9960: Terminate experiment produces unexpected ValueError

JHolba:
**Issue**
Resolves #9960

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/run_models/test_base_run_model.py
+++ b/tests/ert/unit_tests/run_models/test_base_run_model.py
@@ -403,7 +403,7 @@ async def test_terminate_in_pre_evaluation():
     with pytest.raises(
         UserCancelled, match="Experiment cancelled by user in pre evaluation"
     ):
-        await brm.run_ensemble_evaluator_async(MagicMock(), AsyncMock(), AsyncMock())
+        await brm.run_ensemble_evaluator_async(AsyncMock(), AsyncMock(), AsyncMock())
 
 
 @patch("ert.run_models.base_run_model.EnsembleEvaluator")
@@ -429,4 +429,4 @@ async def test_terminate_in_post_evaluation(evaluator):
         UserCancelled,
         match="Experiment cancelled by user in post evaluation",
     ):
-        await brm.run_ensemble_evaluator_async(MagicMock(), AsyncMock(), AsyncMock())
+        await brm.run_ensemble_evaluator_async(AsyncMock(), AsyncMock(), AsyncMock())


src/ert/run_models/base_run_model.py
```diff
@@ -571,7 +575,7 @@ async def run_monitor(
                             # Allow track() to emit an EndEvent.
                             return False
                     elif type(event) is EETerminated:
-                        logger.debug("got terminator event")
+                        logger.debug("got terminated event")
```
jonathan-eq, line 578: Good catch!

src/ert/run_models/base_run_model.py
```diff
@@ -610,7 +615,7 @@ async def run_ensemble_evaluator_async(
         await evaluator._server_started
         if not (await self.run_monitor(ee_config, ensemble.iteration)):
             await evaluator_task
-            return []
+            return _UserCancelled()
```
jonathan-eq, line 618: Is it a good practice to return Exceptions and raise it somewhere else?
Maybe it would be better to just do the extra logic already here, and raise it directly? 
```suggestion
            self.active_realizations = [False for _ in self.active_realizations]
            raise _UserCancelled()
```

src/ert/run_models/base_run_model.py
```diff
@@ -620,7 +625,8 @@ async def run_ensemble_evaluator_async(
             logger.debug("Run model canceled - post evaluation")
             self._end_queue.get()
             await evaluator_task
-            return []
+            return _UserCancelled()
```
jonathan-eq, line 628: Same here^

src/ert/run_models/base_run_model.py
```diff
@@ -94,6 +94,10 @@ def delete_runpath(run_path: str) -> None:
         shutil.rmtree(run_path)
 
 
+class _UserCancelled(Exception):
```
jonathan-eq, line 97: Do we want this exception to be private, if we intend for it to be raised up to a different module?

jonathan-eq (commented):
I have some questions 🤔

src/ert/run_models/base_run_model.py
xjules: Agree with @jonathan-eq on this one. Besides, it might be that `await evaluator_task` raises first if something fails (?)

JHolba: it now raises instead

JHolba: added a nicer error message if it fails while waiting for evaluator_task

src/ert/run_models/base_run_model.py
JHolba: Made it public. I guess it could be useful, but it was just added to control the exception handling inside base_run_model better

src/ert/run_models/base_run_model.py
```diff
@@ -598,7 +605,8 @@ async def run_ensemble_evaluator_async(
         if not self._end_queue.empty():
             logger.debug("Run model canceled - pre evaluation")
             self._end_queue.get()
-            return []
+            raise UserCancelled("Experiment cancelled by user in pre evaluation")
```
jonathan-eq, line 608: Should we have tests for these different scenarios? (cancelling pre evaluation, during evaluation, and post evaluation)

JHolba: I don't think we need specific tests for that 🤷

jonathan-eq: Have you manually tested that this works?

tests/ert/unit_tests/run_models/test_base_run_model.py
```diff
@@ -23,22 +26,31 @@ def __init__(self, status):
...
+        "status_queue": MagicMock(spec=SimpleQueue),
+        "substitutions": MagicMock(spec=Substitutions),
+        "templates": MagicMock(spec=dict),
+        "hooked_workflows": MagicMock(spec=dict),
+        "active_realizations": MagicMock(spec=list),
+        "log_path": Path(""),
+    }
+    return BaseRunModel(**(default_args | kwargs))
+
+
```
jonathan-eq, line 48: This is great!

jonathan-eq: Should we keep this refactoring as a separate commit? `Cleanup tests in test_base_run_model` or something "in that dur".

tests/ert/unit_tests/run_models/test_base_run_model.py
```diff
@@ -481,23 +386,47 @@ def test_get_number_of_active_realizations_varies_when_rerun_or_new_iteration(
...
+async def test_terminate_in_pre_evaluation():
+    brm = create_base_run_model()
+    brm._end_queue.put("terminate")
+    with pytest.raises(
+        UserCancelled, match="Experiment cancelled by user in pre evaluation"
+    ):
+        await brm.run_ensemble_evaluator_async(MagicMock(), AsyncMock(), AsyncMock())
+
+
+@patch("ert.run_models.base_run_model.EnsembleEvaluator")
```
jonathan-eq, line 409: What does this patch outside the test do?

tests/ert/unit_tests/run_models/test_base_run_model.py
```diff
@@ -481,23 +386,47 @@ def test_get_number_of_active_realizations_varies_when_rerun_or_new_iteration(
...
+async def test_terminate_in_post_evaluation(evaluator):
+    async def mocked_run_and_get_successful_realizations() -> list[int]:
+        return list(range(5))
+
+    evaluator().run_and_get_successful_realizations = (
+        mocked_run_and_get_successful_realizations
+    )
+    evaluator()._server_started = asyncio.Future()
+    evaluator()._server_started.set_result(None)
```
jonathan-eq, line 418: Why do we have to call evaluator when we set the instance variables?

jonathan-eq (commented):
Very good! I just have some minor questions

tests/ert/unit_tests/run_models/test_base_run_model.py
JHolba: it is the EnsembleEvaluator module that we are patching, not an instance of EnsembleEvaluator. not sure why they made the interface that way though :shrug:

tests/ert/unit_tests/run_models/test_base_run_model.py
JHolba: it gives us the EnsembleEvaluator module as seen from  `ert.run_models.base_run_model`. it is given to the test function as the first parameter.

src/ert/run_models/base_run_model.py
JHolba: added 2 tests for this

tests/ert/unit_tests/run_models/test_base_run_model.py
JHolba: I agree that it would have been nicer. it is still fairly easy to read though and only affects the test file. good enough? :)

tests/ert/unit_tests/run_models/test_base_run_model.py
jonathan-eq: Is this the same as having `with patch("ert.run_models.base_run_model.EnsembleEvaluator") as evaluator:` inside the test? 
If we have that inside the test instead, can we use `evaluator.run_and_get_successful_realizations` instead of `evaluator().run_and_get_successful_realizations`?

tests/ert/unit_tests/run_models/test_base_run_model.py
```diff
@@ -481,23 +386,47 @@ def test_get_number_of_active_realizations_varies_when_rerun_or_new_iteration(
...
+
+    async def run_monitor_successfully_but_terminate(
+        self, ee_config, iteration: int
+    ) -> bool:
+        self._end_queue.put("terminate")
+        return True
+
+    brm = create_base_run_model()
+    brm.run_monitor = MethodType(run_monitor_successfully_but_terminate, brm)
```
jonathan-eq, line 427: Is this MethodType the same as a partial method?

tests/ert/unit_tests/run_models/test_base_run_model.py
JHolba: it is the same as using with patch(...), but without the extra indentation.
would still have to use evaluator().run...

tests/ert/unit_tests/run_models/test_base_run_model.py
JHolba: it is a bit like functools.partial, but specialized to bind the self parameter of a method

JHolba: same thing could also be achieved with
```python
with patch.object(
    BaseRunModel, "run_monitor", run_monitor_successfully_but_terminate
):
```
this would change all BaseRunModels created in that context, but that would not matter in this case

PR #10168, opened 2025-02-27
Refactor ropt event interactions

verveerpj:
**Issue**
This PR resolves some issues on how the Everest run model interacts with the `ropt `optimizer (and fixes an unrelated bug). There are three commits:

* Fix a bug in the Everest configuration code.
* Remove the `exit_code` field from the events. There is no need for the exit code to be in the events, and this will be removed from the `ropt` events in the near future.
* Change the way the `ropt` events are forwarded. Currently all events are forwarded blindly, except for events that produce resutls, which are split up in multiple events. First of all, I do not think we should forward them blindly if we don't need them. They may not currently have any meaning in Everest. Secondly, by splitting the results into separate events we are not mirroring `ropt` events, but creating new ones with a slightly different meaning (the `ropt` events send results of a whole batch, the new events send one result per batch.) Therefore I suggest here not to mirror `ropt` events, but create new Everest optimization events and emit those when appropriate.

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/everest/snapshots/test_everest_runmodel_events/test_everest_events/config_advanced.yml/snapshot.json
+++ b/tests/everest/snapshots/test_everest_runmodel_events/test_everest_events/config_advanced.yml/snapshot.json
@@ -3,134 +3,114 @@
     {
       "batch": null,
       "event_type": "EverestStatusEvent",
-      "everest_event": "START_OPTIMIZER_STEP",
-      "exit_code": null
+      "everest_event": "START_OPTIMIZER_STEP"
     },
     {
       "batch": null,
       "event_type": "EverestStatusEvent",
-      "everest_event": "START_EVALUATION",
-      "exit_code": null
+      "everest_event": "START_EVALUATION"
     },
     {
       "batch": 0,
       "event_type": "EverestBatchResultEvent",
       "everest_event": "FINISHED_EVALUATION",
-      "exit_code": null,
       "result_type": "FunctionResult"
     },
     {
       "batch": 0,
       "event_type": "EverestBatchResultEvent",
       "everest_event": "FINISHED_EVALUATION",
-      "exit_code": null,
       "result_type": "GradientResult"
     },
     {
       "batch": null,
       "event_type": "EverestStatusEvent",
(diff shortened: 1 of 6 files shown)


tests/everest/snapshots/test_everest_runmodel_events/test_everest_events/config_advanced.yml/snapshot.json
```diff
@@ -3,114 +3,140 @@
...
     },
     {
       "batch": null,
       "event_type": "EverestStatusEvent",
-      "everest_event": "START_EVALUATION"
+      "everest_event": "START_OPTIMIZER_EVALUATION"
     },
     {
-      "batch": 5,
+      "batch": 6,
```
yngve-sk, line 124: Is this expected?

tests/everest/snapshots/test_everest_runmodel_events/test_everest_events/config_advanced.yml/snapshot.json
```diff
@@ -3,114 +3,140 @@
...
       "result_type": "GradientResult"
     },
     {
       "batch": null,
       "event_type": "EverestStatusEvent",
-      "everest_event": "START_EVALUATION"
+      "everest_event": "START_OPTIMIZER_EVALUATION"
+    },
+    {
+      "batch": 5,
```
yngve-sk, line 101: The order of events seems different now, they no longer seem sequential wrt batch id, expected?

tests/everest/snapshots/test_everest_runmodel_events/test_everest_events/config_advanced.yml/snapshot.json
verveerpj: No, bug, fixed.

tests/everest/snapshots/test_everest_runmodel_events/test_everest_events/config_advanced.yml/snapshot.json
verveerpj: also bug.

PR #10174, opened 2025-02-27
Refactor test_scheduler_receives_checksum_and_waits_for_disk_sync

xjules:
**Issue**
This gets rid of port_range specification and rewrites the rename_and_wait not to throw ASYNC110.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/ensemble_evaluator/test_scheduler.py
+++ b/tests/ert/unit_tests/ensemble_evaluator/test_scheduler.py
@@ -15,12 +15,11 @@ from ert.ensemble_evaluator.config import EvaluatorServerConfig
 async def test_scheduler_receives_checksum_and_waits_for_disk_sync(
     tmpdir, make_ensemble, monkeypatch, caplog
 ):
-    num_reals = 1
-    port_range = (1024, 65535)
-
     async def rename_and_wait():
         Path("real_0/job_test_file").rename("real_0/test")
-        while "Waiting for disk synchronization" not in caplog.messages:  # noqa: ASYNC110
+        for _ in iter(
+            lambda: "Waiting for disk synchronization" not in caplog.messages, False
+        ):
             await asyncio.sleep(0.1)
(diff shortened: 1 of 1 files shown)


tests/ert/unit_tests/ensemble_evaluator/test_scheduler.py
```diff
@@ -15,12 +15,11 @@
...
 ):
-    num_reals = 1
-    port_range = (1024, 65535)
-
     async def rename_and_wait():
         Path("real_0/job_test_file").rename("real_0/test")
-        while "Waiting for disk synchronization" not in caplog.messages:  # noqa: ASYNC110
+        for _ in iter(
+            lambda: "Waiting for disk synchronization" not in caplog.messages, False
+        ):
```
jonathan-eq, line 22: How does this work? I am not seeing what we are iterating over 🤔

xjules: It is identical to:
```python
while "Waiting for disk synchronization" not in caplog.messages:
    # Do something, like sleep
    await asyncio.sleep(0.1)
```

but more functional and thus not triggering ASYNC110

xjules: btw, the second parameter is sentinel and essentially exits when the first parameter==second(False)

jonathan-eq: Instead of two negatives, can we change it to positives? 
```suggestion
          for _ in iter(
	     lambda: "Waiting for disk synchronization" in caplog.messages,
	     True
        ):
```

xjules: That won't work as it exit immediately.  It needs to be negative.

jonathan-eq (approved):
Very good! 🚢

PR #10262, opened 2025-03-11
Add more debugging when submitting everserver

frode-aarstad:
**Issue**
Resolves #my_issue

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/bin/everest_script.py
+++ b/src/everest/bin/everest_script.py
@@ -151,10 +151,22 @@ async def run_everest(options: argparse.Namespace) -> None:
             logger.error(f"Failed to save optimization config: {e}")
 
         logging_level = logging.DEBUG if options.debug else options.config.logging_level
-        await start_server(options.config, logging_level)
+
         print("Waiting for server ...")
-        wait_for_server(options.config.output_dir, timeout=600)  # 10 minutes
+        logger.debug("Submitting everserver")
+        try:
+            await asyncio.wait_for(
+                start_server(options.config, logging_level), timeout=1800
+            )  # 30 minutes
+            logger.debug("Everserver submitted and started")
+        except TimeoutError as e:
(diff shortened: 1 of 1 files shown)


ci/testkomodo.sh
```diff
@@ -92,7 +92,7 @@ run_everest_egg_test() {
     fi
     cat "$CONFIG"
 
-    everest run "$CONFIG"
+    everest run "$CONFIG" --debug
```
DanSava, line 95: You will get many debug messages in the console output, but I guess that is what you want.

src/everest/bin/everest_script.py
```diff
@@ -145,10 +145,22 @@ async def run_everest(options: argparse.Namespace) -> None:
...
 
         logging_level = logging.DEBUG if options.debug else options.config.logging_level
-        await start_server(options.config, logging_level)
+
         print("Waiting for server ...")
-        wait_for_server(options.config.output_dir, timeout=600)  # 10 minutes
+        logger.debug("Submitting everserver")
+        try:
+            await asyncio.wait_for(
+                start_server(options.config, logging_level), timeout=1800
```
DanSava, line 153: I am assuming this only happens for the testing hopefully. 

Is it possible to just use a different queue when running the tests to avoid queue congestion if that is indeed the source of the longer start times?

DanSava (approved):
Looks good!

src/everest/bin/everest_script.py
frode-aarstad: Its always possible that someone starts an everest run on a cluster that has no available slots thus getting put in the queue. We have no way of handling this currently other than wait and timeout after time=X (in this case 1800s).

Ideally the info what we are in a queue should be propagated to the user

PR #10264, opened 2025-03-11
Log when duration of forward model step is negative

### LINKED ISSUE #10150: FMTableView shows negative duration

While running bigpoly, I noticed the duration was off. The steps running were taking 0 seconds (env command), but before the finished event was proccessed, it showed as `-1 day, 23:01:48`

Image

_No response_

### steps to reproduce

```Python

```

- [x] python 3.11
- [ ] python 3.12
- [ ] macosx
- [ ] rhel7
- [x] rhel8
- [ ] local queue
- [x] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

JHolba:
Set negative durations to 0 in simulation status
Will also log warnings when encountering these issues.

**Issue**
Resolves #10150 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/gui/model/snapshot.py
+++ b/src/ert/gui/model/snapshot.py
@@ -69,6 +69,8 @@ _QCOLORS = {
     state.COLOR_NOT_ACTIVE: QColor(*state.COLOR_NOT_ACTIVE),
 }
 
+_warn_once = True
+
 
 def _estimate_duration(
     start_time: datetime, end_time: datetime | None = None
@@ -429,6 +431,16 @@ class SnapshotModel(QAbstractItemModel):
                 )
                 # There is no method for truncating microseconds, so we remove them
                 delta -= timedelta(microseconds=delta.microseconds)
+                if delta < timedelta():
+                    global _warn_once  # noqa: PLW0603
+                    if _warn_once:
+                        _warn_once = False
+                        logger.warning(
+                            "Negative duration in snapshot encountered. "
+                            f"start_time={start_time} end_time={node.data.get(ids.END_TIME)} "
+                            f"delta={delta}"
+                        )
+                    delta = timedelta()
                 return str(delta)
 
             return node.data.get(data_name)


src/ert/gui/model/snapshot.py
```diff
@@ -429,6 +431,16 @@ def _fm_step_data(
                 )
                 # There is no method for truncating microseconds, so we remove them
                 delta -= timedelta(microseconds=delta.microseconds)
+                if delta < timedelta():
```
jonathan-eq, line 434: Is this the same as `if delta < 0`?

src/ert/gui/model/snapshot.py
```diff
@@ -429,6 +431,16 @@ def _fm_step_data(
...
+                    global warn_once  # noqa: PLW0603
+                    if warn_once:
+                        warn_once = False
+                        logger.warning(
+                            "Negative duration in snapshot encountered. "
+                            f"start_time={start_time} end_time={node.data.get(ids.END_TIME)} "
+                            f"delta={delta}"
+                        )
```
jonathan-eq, line 442: @xjules showed me a trick in 2023, and now I am showing it to you =)
```suggestion
                        end_time = node.data.get(ids.END_TIME)
                        logger.warning(
                             "Negative duration in snapshot encountered. "
                             f"{start_time=} {end_time=} {delta=}"
                        )
```

src/ert/gui/model/snapshot.py
JHolba: it is the same as delta < timedelta(seconds=0)

src/ert/gui/model/snapshot.py
JHolba: in this instance it is not that useful. f"{object=} gives object.\_\_repr__ while I want object.\_\_str__
\_\_str__ `'2025-01-01 00:00:00'` 
\_\_repr__ `'datetime.datetime(2025, 1, 1, 0, 0)'`

xjules: I don't get why this shouldn't work @JHolba ?

jonathan-eq (approved):
Very good, but let's wait for @eivindjahren

src/ert/gui/model/snapshot.py
JHolba: it works, but you get different text. I want this representation: `'2025-01-01 00:00:00'`.

you can try this in a repl
```python
from datetime import datetime
start = datetime.now()
print("print using {start=}: " f"{start=}")
print("print using start={start}: " f"start={start}")
```

PR #10265, opened 2025-03-11
Make flowbase definition passthrough to ertconfig

yngve-sk:
**Issue**
Resolves [#10248](https://github.com/equinor/ert/issues/10248)

Passthrough `flowbase` as `ECLBASE` to ERT, and disallow both `flowbase` and `eclbase` in everest config simultaneously.

Discussion point maybe for another issue: We could rename the base directory to something not explicitly flow/eclipse


DanSava (approved):
Looks good :100:

PR #10289, opened 2025-03-12
Add lru cache to _get_machine_name

yngve-sk:
Avoids two calls to it when starting up everserver, it goes over network and may be slow. 
Locally on macbook pro M1 max it takes about 5sec per invocation, so this reduces the startup delay by 5sec


DanSava (approved):
:+1:

PR #10302, opened 2025-03-13
Revise Everest storage writing

### LINKED ISSUE #10235: Everest does not dump results until run is complete
It would be best to have the parquet files dumped after every batch is finished, instead of after the entire optimization is finished.

> I expect users to kill optimization runs at any time, or it to die out of too many reasons (from licenses to unlucky controls) so it's probably better if those files could be produced on the flight

yngve-sk:
**Issue**
Resolves #10235 

**Approach**
Write files instantly, and read them from the file system rather than from in-memory. Should set things up for redirecting some of the dataframe getters to ERT storage.

### DIFF
--- a/src/everest/api/everest_data_api.py
+++ b/src/everest/api/everest_data_api.py
@@ -312,6 +312,9 @@ class EverestDataAPI:
             )
 
         for batch in self._ever_storage.data.batches:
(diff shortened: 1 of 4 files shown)


verveerpj (approved):
LGTM!

PR #10317, opened 2025-03-14
Streamline get_opt_status logic

yngve-sk:


src/everest/detached/__init__.py
```diff
@@ -163,43 +163,29 @@ def get_opt_status(output_folder: str) -> dict[str, Any]:
...
-            b.batch_id
-            for b in storage.data.batches
-            if b.realization_controls is not None and b.batch_objectives is not None
-        ],
+        "batches": [b.batch_id for b in batches],
         "controls": [
-            b.realization_controls.select(control_names).to_dicts()[0]
-            for b in storage.data.batches
-            if b.realization_controls is not None
+            b.realization_controls.select(control_names).to_dicts()[0] for b in batches
```
eivindjahren, line 183: This seems to be a change as `batches` contains those elements where `batch_objectives is not None`. However, before, these lists would also expect that `realization_controls is not None` .

yngve-sk: Yup, I can set it back to make the behavior the same, the logic is still a bit implicit and should be fixed more in a different PR, batch data is not tagged as gradient|function yet, and this is a bit of an ad hoc way of checking it, basically there is an unmentioned/unspecified thing here.

yngve-sk: Fixed now, behavior still same even though the logic looks a tad more stringent by requiring them all to have `realization_controls` and `batch_objectives`, tested by `test_get_opt_status`. Added a "rewrite" to commit to indicate the change

yngve-sk:
included in other PR

PR #10385, opened 2025-03-19
Make cached_example overwrite cache folder

yngve-sk:
If the folder already exists, but the cache entry does not (due to some failure in run_experiment, there is a FileExistsError and you have to clear the cache manually.


DanSava (approved):
Looks good!

PR #10402, opened 2025-03-20
Add inline directive to config validation schema

eivindjahren:
By reading the files first, we avoid replicating logic for handling unicode and adding context in several places. 

This PR only add usage of EXISITING_FILE_INLINE to gen_kw and time_map, which shows how it can be used to 
avoid writing/reading of files in tests. 

The directive can also be used later to avoid the same IO in tests with observations and workflows.

Note: does not avoid writing gen_kw templates yet as that means migrating storage.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/config/test_gen_kw_config.py
+++ b/tests/ert/unit_tests/config/test_gen_kw_config.py
@@ -9,12 +9,11 @@ from lark import Token
 from ert.config import (
     ConfigValidationError,
     ConfigWarning,
-    EnsembleConfig,
     ErtConfig,
     GenKwConfig,
 )
 from ert.config.gen_kw_config import TransformFunctionDefinition
-from ert.config.parsing import ConfigKeys, ContextString
+from ert.config.parsing import ContextString
 from ert.config.parsing.file_context_token import FileContextToken
 from ert.enkf_main import create_run_path, sample_prior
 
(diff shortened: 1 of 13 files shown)


tests/ert/unit_tests/config/test_gen_kw_config.py
```diff
@@ -263,32 +262,32 @@ def test_gen_kw_is_log_or_not(
...
-        config = dedent(
-            """
-        JOBNAME my_name%d
-        NUM_REALIZATIONS 1
-        GEN_KW KW_NAME template.txt kw.txt prior.txt
-        """
-        )
-        with open("config.ert", "w", encoding="utf-8") as fh:
-            fh.writelines(config)
+        template = "MY_KEYWORD <MY_KEYWORD>"
```
andreas-el, line 265: Reason why this wasn't inlined too?

tests/ert/unit_tests/config/test_gen_kw_config.py
```diff
@@ -263,32 +262,32 @@ def test_gen_kw_is_log_or_not(
...
-        with open("prior.txt", "w", encoding="utf-8") as fh:
-            if distribution == "TRUNCATED_NORMAL":
-                fh.writelines(f"MY_KEYWORD {distribution} {mean} {std} -1 1")
-            else:
-                fh.writelines(f"MY_KEYWORD {distribution} {mean} {std}")
+
+        if distribution == "TRUNCATED_NORMAL":
+            distribution_line = f"MY_KEYWORD {distribution} {mean} {std} -1 1"
+        else:
+            distribution_line = f"MY_KEYWORD {distribution} {mean} {std}"
```
andreas-el, line 272: This seems a bit excessive, but might be easier on the eye?

tests/ert/unit_tests/config/test_gen_kw_config.py
eivindjahren: That is because the template file name goes into storage and we don't want to mix in storage migration at this point.

tests/ert/unit_tests/config/test_gen_kw_config.py
eivindjahren: Not sure what you mean, maybe you have an example alternative in mind?

andreas-el:
```
distribution_line = f"MY_KEYWORD {distribution} {mean} {std}"
if distribution == "TRUNCATED_NORMAL":
    distribution_line += " -1 1"
```

or

```
distribution_line = f"MY_KEYWORD {distribution} {mean} {std}{' -1 1' if distribution == 'TRUNCATED_NORMAL' else ''}"
```

tests/ert/unit_tests/config/test_gen_kw_config.py
eivindjahren: Ah, I understand now. Fixed it.

andreas-el (approved):
This is really solid work. 🏖️

tests/ert/unit_tests/config/test_gen_kw_config.py
eivindjahren: Aha, I'll just leave the original in place I think. The alternatives had a bit of stateful logic to them.

PR #10416, opened 2025-03-20
Bugfix wrong src realization for simulator cache hits

yngve-sk:


src/ert/run_models/everest_run_model.py
```diff
@@ -493,6 +493,70 @@ def _run_forward_model(
...
+                right.sort(["model_realization", control_name]),
+                on=control_name,
+                by="model_realization",  # pre-join by model realization
+                tolerance=EPS,  # Same as np.allclose with atol=EPS
+                strategy="nearest",
+                check_sortedness=False,
+                # Ref: https://github.com/pola-rs/polars/issues/21693
+            )
+
+            left = left.filter(pl.col("flat_index").is_not_null())
```
verveerpj, line 545: why is this needed?

src/ert/run_models/everest_run_model.py
```diff
@@ -493,6 +493,70 @@ def _run_forward_model(
...
+
+            left = left.drop([s for s in left.columns if s.endswith("_right")])
+
+            if left.is_empty():
+                break
+
+        return (
+            left.rename({"realization": "simulation_id"}).drop(
```
verveerpj, line 553: It seems you did not do this before, why  now?

src/ert/run_models/everest_run_model.py
```diff
@@ -592,7 +612,7 @@ def _create_evaluation_infos(
 
             if cache_hits_df is not None and not cache_hits_df.is_empty():
                 hit_row = cache_hits_df.filter(
-                    cache_hits_df["realization"] == flat_index
+                    cache_hits_df["flat_index"] == flat_index
```
verveerpj, line 615: This was the actual bug, right?

verveerpj (commented):
some questions

src/ert/run_models/everest_run_model.py
yngve-sk: yep, it was supposed to take the `realization` from the right, but took from the left

src/ert/run_models/everest_run_model.py
yngve-sk: Will put in separate commit, basically we are saying a cache hit is a certain simulation from a different batch, `realization` in terms of ERT, I think `simulation_id` is more correct in that sense, what do you think?

src/ert/run_models/everest_run_model.py
yngve-sk: If the `flat_index` is null, it means there was basically no matching row in the `right` dataframe, which is the controls we are to evaluate. I initially thought this would be handled on its own by the asof join, but apparently not, as we incrementally join by one and one control

verveerpj (approved):

PR #10417, opened 2025-03-20
Fix gui not working with high contrast mode

### LINKED ISSUE #10336: Ert gui does not work with high contrast mode

The colors get screwed up when high contrast mode is enabled.

Image
The easiest fix would be to emit a qmessage with a warning as it is something we do not currently support, but it should be implemented in the long run in case we have users who are visually impaired. 

Ert works with high contrast mode, or the user is made aware that high contrast mode is enabled and might cause issues with the gui. 

### steps to reproduce

```Python
Enable high contrast mode
Run ert
```

- [ ] python 3.11
- [ ] python 3.12
- [ ] macosx
- [ ] rhel7
- [x] rhel8
- [ ] local queue
- [ ] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

### LINKED ISSUE #10419: Almost invisible text in manage experiment gui

There is a very faint text under type (None) because it is white text on light gray background. 
Image

The text should be black.

### steps to reproduce

```Python

```

- [ ] python 3.11
- [ ] python 3.12
- [ ] macosx
- [ ] rhel7
- [x] rhel8
- [ ] local queue
- [ ] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

jonathan-eq:
**Issue**
Resolves #10336 
Resolves #10419 

**Approach**
This commit fixes the issue with high contrast mode by emitting a warning message box stating that there might be unexpected behavior when using high contrast mode, and also changing the gui to be as similar to normal contrast mode as possible.
It also fixes the issue where we use a bad text color for ManageExperiment QTreeView "Type" column. 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/gui/main_window.py
+++ b/src/ert/gui/main_window.py
@@ -8,13 +8,20 @@ from pathlib import Path
 from PyQt6.QtCore import QCoreApplication, QEvent, QSize, Qt
 from PyQt6.QtCore import pyqtSignal as Signal
 from PyQt6.QtCore import pyqtSlot as Slot
-from PyQt6.QtGui import QAction, QCloseEvent, QCursor, QIcon, QMouseEvent
+from PyQt6.QtGui import (
+    QAction,
+    QCloseEvent,
+    QCursor,
+    QIcon,
+    QMouseEvent,
+)
 from PyQt6.QtWidgets import (
     QButtonGroup,
     QFrame,
     QHBoxLayout,
     QMainWindow,
     QMenu,
+    QMessageBox,
     QToolButton,
     QVBoxLayout,
     QWidget,
@@ -27,6 +34,7 @@ from ert.gui.about_dialog import AboutDialog
 from ert.gui.ertnotifier import ErtNotifier
 from ert.gui.find_ert_info import find_ert_info
 from ert.gui.simulation import ExperimentPanel
+from ert.gui.simulation.experiment_panel import is_high_contrast_mode
 from ert.gui.simulation.run_dialog import RunDialog
 from ert.gui.tools.event_viewer import EventViewerTool, GUILogHandler
 from ert.gui.tools.export import ExportTool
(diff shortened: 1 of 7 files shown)


jonathan-eq:
Main: 
Screenshot 2025-03-20 at 15 44 03
Screenshot 2025-03-20 at 15 44 30
Screenshot 2025-03-20 at 15 45 24

Screenshot 2025-03-20 at 15 45 44
Screenshot 2025-03-20 at 15 46 00

jonathan-eq:
Branch - normal mode:
image

jonathan-eq:
Will manually test this with the final commit^

xjules:
Some mypy issue:
```
Error: src/everest/config/simulator_config.py:32: error: "ModelMetaclass" has no attribute "model_fields"  [attr-defined]
```

src/ert/gui/main_window.py
```diff
@@ -109,11 +117,23 @@ def __init__(
...
+        self.setStyleSheet("QWidget { color: black; }")
         if self.is_dark_mode():
             self.side_frame.setStyleSheet("background-color: rgb(64, 64, 64);")
         else:
             self.side_frame.setStyleSheet("background-color: lightgray;")
 
+        if self.is_high_contrast_mode():
+            msg_box = QMessageBox()
+            msg_box.setText(
+                "High contrast mode detected. This is not supported by Ert and features may not work as expected."
```
xjules, line 129: ""High contrast mode detected. This is not supported by Ert and some features may not work as expected."

src/ert/gui/tools/plugins/plugins_tool.py
```diff
@@ -35,6 +35,9 @@ def __init__(
         self.__plugins = {}
 
         self.menu = QMenu("&Plugins")
+        self.menu.setStyleSheet(
+            "QMenu {color: black;} QMenu::item:selected {background-color: rgb(70,140,230); color: white;} QMenu::item:disabled {color: rgb(170,170,170);}"
+        )
```
jonathan-eq, line 40: I am not sure this stylesheet will be necessary if we can correctly set parent as the main_window. There we already set the stylesheet for QMenu.
Maybe it is enough to pass `self` to the QMenu on L39?

src/ert/gui/simulation/experiment_panel.py
```diff
@@ -24,6 +24,13 @@
 from ert.config import QueueSystem
 from ert.ensemble_evaluator import EvaluatorServerConfig
 from ert.gui.ertnotifier import ErtNotifier
+
+
+def is_high_contrast_mode() -> bool:
```
xjules, line 29: Shall this be a utility function in order you can access it also from the main window?

jonathan-eq: Yes, it should

jonathan-eq: I removed the one in main_window and kept this one.

PR #10432, opened 2025-03-21
Fix bug where no summary data leads to everviz not opening

yngve-sk:


oyvindeide (approved):
Great! Would be good to have a regression test though.

PR #10433, opened 2025-03-21
Avoid making assumptions on timestamps for cpu overspending detection

### LINKED ISSUE #10057: NUM_CPU misconfiguration detection can go wrong when realizations are resubmitted
In the logs it has been observed:

`Misconfigured NUM_CPU, forward model step 'ECLIPSE100' for realization 128 spent 144.51000000000002 cpu seconds with wall clock duration 4.9 seconds, a factor of 29.53, while NUM_CPU was 1`

This is for a realization that has been resubmitted, and the resubmission is probably involved in tricking the log message to yield such an outrageous factor.

OperationId: 26cdd2df9f540f04ac5eebd8497682c9

berland:
There is no reason to believe that datetime.datetime.now() can be used in place of actual reported timestamps if one is to claim the user is doing something wrong.

**Issue**
Resolves #10057 

**Approach**
Return early.

      --exec 'just rapid-tests'`)

## When applicable


src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -466,10 +465,13 @@ def detect_overspent_cpu(num_cpu: int, real_id: str, fm_step: FMStepSnapshot) ->
...
-    duration = (
-        (fm_step.get(ids.END_TIME) or now) - (fm_step.get(ids.START_TIME) or now)
-    ).total_seconds()
+    start_time = fm_step.get(ids.START_TIME)
+    if start_time is None:
+        return ""
+    end_time = fm_step.get(ids.END_TIME)
+    if end_time is None:
+        return ""
+    duration = (end_time - start_time).total_seconds()
```
andreas-el, line 474: Something to consider;
```
start_time = fm_step.get(ids.START_TIME)
end_time = fm_step.get(ids.END_TIME)

if not all([start_time, end_time]):
    return ""

duration = (end_time - start_time).total_seconds()
```

berland: I like it, but  `mypy` does not. 🤔

berland: Went for the obvious solution.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -466,10 +465,11 @@ def detect_overspent_cpu(num_cpu: int, real_id: str, fm_step: FMStepSnapshot) ->
     """Produces a message warning about misconfiguration of NUM_CPU if
     so is detected. Returns an empty string if everything is ok."""
     allowed_overspending = 1.05
-    now = datetime.datetime.now()
-    duration = (
-        (fm_step.get(ids.END_TIME) or now) - (fm_step.get(ids.START_TIME) or now)
-    ).total_seconds()
+    start_time = fm_step.get(ids.START_TIME)
```
eivindjahren, line 468: I think it would be a good idea to split this part out for testing purposes, but that does not block this PR.

My suggestion would be the following (using doctest as it is easier in a github comment)

```python
def _fm_duration_seconds(start_time: datetime | None, end_time: datetime | None) -> float | None:
  """
  >>> _fm_duration_seconds(None, datetime(1970, 1, 1))
  None
  >>>  _fm_duration_seconds( _fm_duration_seconds(None, datetime(1970, 1, 1)), None)
  None
  >>>  _fm_duration_seconds( _fm_duration_seconds(datetime(1970, 1, 1), datetime(1970, 1, 1)))
  0.0
  """
  if start_time is None or end_time is None:
    return None
  return (end_time - start_time).total_seconds()


def detect_overspent_cpu(num_cpu: int, real_id: str, fm_step: FMStepSnapshot) -> str:
  ...
  duration = _fm_duration_seconds(fm_step.get(ids.START_TIME), fm_step.get(ids.END_TIME))
  if duration is None or duration <= 0:
      return ""
```

eivindjahren (approved):
Had a small comment, but it is not blocking.

PR #10446, opened 2025-03-24
Rewrite everest rundialog and gui window

frode-aarstad:
**Issue**
Resolves

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/bin/everest_script.py
+++ b/src/everest/bin/everest_script.py
@@ -22,6 +22,7 @@ from everest.detached import (
     wait_for_server,
 )
 from everest.everest_storage import EverestStorage
+from everest.gui.everest_client import EverestClient
 from everest.simulator.everest_to_ert import (
     everest_to_ert_config_dict,
 )
@@ -188,6 +189,11 @@ async def run_everest(options: argparse.Namespace) -> None:
         if options.gui:
             from everest.gui.main import run_gui  # noqa
 
+            client = EverestClient(
+                output_dir=options.config.output_dir,
+                config_file=options.config.config_path,
+            )
+
             monitor_thread = ErtThread(
                 target=run_detached_monitor,
                 name="Everest CLI monitor thread",
@@ -195,7 +201,7 @@ async def run_everest(options: argparse.Namespace) -> None:
                 daemon=True,
             )
             monitor_thread.start()
-            run_gui(options.config.config_path)
+            run_gui(client, options.config.config_path)
             monitor_thread.join()
         else:
             # blocks until the run is finished
--- a/src/everest/gui/everest_client.py
+++ b/src/everest/gui/everest_client.py
@@ -17,6 +17,8 @@ from _ert.threading import ErtThread
 from ert.ensemble_evaluator import EvaluatorServerConfig
 from ert.run_models import BaseRunModelAPI
 from ert.run_models.event import StatusEvents, status_event_from_json
+from everest.config.everest_config import EverestConfig
+from everest.config.server_config import ServerConfig
 from everest.strings import (
     START_EXPERIMENT_ENDPOINT,
     STOP_ENDPOINT,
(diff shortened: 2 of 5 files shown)


src/everest/bin/everest_script.py
```diff
@@ -188,14 +189,19 @@ async def run_everest(options: argparse.Namespace) -> None:
         if options.gui:
             from everest.gui.main import run_gui  # noqa
 
+            client = EverestClient(
+                output_dir=options.config.output_dir,
+                config_file=options.config.config_path,
```
DanSava, line 194: This path is used to reconstruct the everest config object in `client.create_run_model_api` is there a better way of doing this?

src/everest/gui/everest_client.py
```diff
@@ -91,9 +92,17 @@ def passthrough_ws_events() -> None:
 
         return event_queue, monitor_thread
 
-    def create_run_model_api(
-        self, queue_system: str, runpath_format_string: str
-    ) -> BaseRunModelAPI:
+    def create_run_model_api(self) -> BaseRunModelAPI:
```
DanSava, line 95: I don't understand why this needs to change in this way and if you want to refactor this I would argue that it should be moved to EverestMainWindow

src/everest/bin/everest_script.py
frode-aarstad: I guess we could extract the stuff we need from options.config and add that to the EverestClient(). Would that be better?

src/everest/gui/everest_client.py
```diff
@@ -28,20 +30,19 @@
 class EverestClient:
     def __init__(
         self,
-        url: str,
-        cert_file: str,
-        username: str,
-        password: str,
-        ssl_context: ssl.SSLContext,
+        output_dir: str,
```
yngve-sk, line 33: I don't think this change to the signature makes it more clear/testable/explicit. Passing in the output dir and path to config file seems more implicit/convoluted to me at least, vs just passing in the previous args to start up the client. Would maybe opt for having a staticmethod `create` or something like that taking in fewer args that are implicitly unpacked to more args, but would argue for keeping it explicit where these settings are set when we start up the client. The issue was mainly focused on initializing the window without the entire EverestConfig, this refactor is related but does not seem directly necessary to solve that issue to me. Similar comment for the `test_everest_client.py` changes.

src/everest/gui/main.py
```diff
@@ -31,10 +34,28 @@ def run_gui(config_file: str) -> None:
...
+    main_window.show()
+    main_window.activateWindow()
+    main_window.raise_()
+
+    run_model_api = client.create_run_model_api()
+    event_queue, event_monitor_thread = client.setup_event_queue_from_ws_endpoint(
+        refresh_interval=0.02, open_timeout=40, websocket_recv_timeout=1.0
+    )
+
+    run_dialog = RunDialog(
```
yngve-sk, line 48: Lifting some of the logic up is probably OK, but it does not seem related/necessary for solving the issue of initializing the main window using only the output directory instead of the config

src/everest/gui/everest_client.py
frode-aarstad: So the work here should just be to add sim_dir and queue_system_name to the MainWindow _init_?

frode-aarstad: I the everest rundialog the EverestMainWindow?

yngve-sk: Hmm. Issue was a bit unclear on whether it was main window or rundialog now that I re-read it, but mainly the `EverestClient` should be the only way to get the info. It should only need the connection info (which can be derived from the output dir). 

The BaseRunModelAPI currently does have some dependencies on the EverestConfig like queue system / sim_dir for the runpath, but ideally these should be exposed through endpoints so that they aren't needed. The RunDialog itself may not need the info. 

* I think https://github.com/equinor/ert/pull/10447 will remove the `RunDialog` dependency on the queue system, so that can be removed from the API alltogether.
* The simulation_dir / runpath must currently be in the rundialog, I suggest to just create an endpoint for fetching that through the everserver, and point the `EverestClient` to that endpoint.

frode-aarstad: Ill redo this after the the PR you mentioned has been merger

yngve-sk: My initial comment was a bit off, edited that out once I realized 😅 We will still need to have the `queue_system`, just not in the rundialog, but in the scope outside of it, so I think that too should be exposed through an endpoint.

Issue #10457, opened 2025-03-25
Add load_parameter_graph from parameter configs


Blunde1:
See implementation here: https://github.com/equinor/ert/commit/8c960dee956922863457e8add2563b5ea1c5d2b0

PR #10466, opened 2025-03-26
Add Ensemble Information Filter runmodel 

### LINKED ISSUE #10455: EnIF available in ERT GUI
EnIF is not TRL4 as of writing. Usage must be an active choice on the user, with clear information that this is not TRL4

yngve-sk:
TODO:

- [x] Make existing tests pass
- [x] Add performance tests for enif
- [x] Add integration test for enif poly 
- [x] Add snapshot test for enif

**Issue**
Resolves #10455, #10456, #10457, #10458, #10459

Screenshot 2025-03-25 at 12 43 40
Screenshot 2025-03-25 at 12 49 47


src/ert/analysis/_es_update.py
```diff
@@ -152,6 +157,57 @@ def _expand_wildcards(
     return sorted(set(matches))
 
 
+def _create_update_snapshot(
```
Blunde1, line 160: Kvifor er dette del av EnIF PR?

src/ert/analysis/_es_update.py
```diff
@@ -702,3 +733,231 @@ def smoother_update(
...
+        iens_active_index,
+    )
+
+    observations_and_responses = observations_and_responses.sort(
+        by=["observation_key", "index"]
+    )
+
+    S = observations_and_responses.select(
+        observations_and_responses.columns[5:]
+    ).to_numpy()
```
Blunde1, line 823: `columns[5:]` deserve a comment or explanation. There is no "natural" API for selecting `S`?

Blunde1: I think I have seen this before, and asked the same question. I think the answer is that this indeed is the way, but then there should be a comment

src/ert/analysis/_es_update.py
```diff
@@ -702,3 +733,231 @@ def smoother_update(
...
+    observation_values = (
+        observations_and_responses.select("observations").to_numpy().reshape((-1,))
+    )
+    observation_errors = (
+        observations_and_responses.select("std").to_numpy().reshape((-1,))
+    )
+    obs_keys = (
+        observations_and_responses.select("observation_key").to_numpy().reshape((-1,))
+    )
+    indexes = observations_and_responses.select("index").to_numpy().reshape((-1,))
```
Blunde1, line 833: Should we use the below call instead? In new notebooks I have used
```py
def fetch_responses_and_parameters_from_storage(prior):
    # Get responses
    iens_active_index = np.flatnonzero(prior.get_realization_mask_with_parameters())
    selected_observations = prior.experiment.observation_keys
    observations_and_responses = prior.get_observations_and_responses(
        selected_observations,
        iens_active_index,
    )
    S = observations_and_responses.select(
        observations_and_responses.columns[5:]
    ).to_numpy()
    observation_values = (
        observations_and_responses.select("observations").to_numpy().reshape((-1,))
    )
    observation_errors = observations_and_responses.select("std").to_numpy().reshape((-1,))
    observation_names = (
        observations_and_responses.select("observation_key").to_numpy().reshape((-1,))
    )
    indexes = observations_and_responses.select("index").to_numpy().reshape((-1,))
    observation_names_unique = result = [": ".join(t) for t in list(zip(observation_names, indexes))]
    Y = S.T.copy()
    d = observation_values
    d_sd = observation_errors
    
    # Get prior and standardize it
    parameter_groups = list(prior.experiment.parameter_info.keys())
    X_full, param_group_bounds, parameter_names = _all_parameters(
        ensemble=prior,
        iens_active_index=iens_active_index,
        param_groups=parameter_groups,
    )
    scaler_x = StandardScaler()
    U_prior = scaler_x.fit_transform(X_full.T)

    return U_prior, scaler_x, Y, d, d_sd, parameter_groups, iens_active_index, parameter_names, observation_names_unique, param_group_bounds
```

This also loads information that is used for plotting:
**call to regression plots**
```py
observation_index = 1
plot_centered_S_vs_features(
    observation_index, 
    Y.T, 
    U_prior.T, 
    gtmap.H, 
    d, 
    d_sd, 
    param_group_bounds,
    num_cols=3,
    X_post = X_full_post,
    observation_name = observation_names_unique[observation_index],
    parameter_names = parameter_names
)
```

**call to waterfall plot**
```py
create_waterfall_chart(
    K[parameter_index,:],
    innovation_mean,
    param_prior,
    param_posterior,
    parameter_names[parameter_index],
    observation_names_unique,
    nobservations = 2
)
```
The `K` is created from a special version of the Kalman gain used in
```py
K = np.linalg.inv(gtmap.Prec_u.toarray()) @ gtmap.H.T @ gtmap.Prec_residual_noisy()
innovation_mean = d - np.mean(Y, axis=0)
```
- `gtmap.Prec_u` is the posterior precision (sparse)
- `gtmap.H` is the sparse linear map
- `gtmap.Prec_residual_noisy()` returns another sparse precision
we need to store all of these, or the gtmap object itself.

Also 
`innovation_mean = d - np.mean(Y, axis=0)`, but perhaps not necessary? can be re-created after loading data

Do we want to store e.g. `observation_names_unique`, `param_group_bounds` and  `parameter_names`?  Perhaps not, as they may be loaded again from storage...

src/ert/analysis/_es_update.py
```diff
@@ -702,3 +733,231 @@ def smoother_update(
...
+    X_full_scaler = StandardScaler()
+    X_full_scaled = X_full_scaler.fit_transform(X_full.T)
+    print(f"Scaled X_full has shape: {X_full_scaled.shape}")
+
+    # Call fit: Learn sparse linear map only
+    H = linear_boost_ic_regression(
+        U=X_full_scaled,
+        Y=S.T,
+        learning_rate=1.0,
+        effective_dimension=0,
```
Blunde1, line 873: Let's let `graphite-maps` set these values. I think for instance default `learning_rate=0.5` allows to learn slightly more and is more conservative to some extent

Blunde1: The boosting learns in smaller chunks with learning_rate smaller

src/ert/analysis/_es_update.py
```diff
@@ -702,3 +733,231 @@ def smoother_update(
...
+        X_local_scaler = StandardScaler()
+        X_scaled = X_local_scaler.fit_transform(X_local.T)
+
+        graph_u_sub = config_node.load_parameter_graph()
+
+        # This will work for dim(X_scaled) on order O(n^5)
+        Prec_u_sub, *_ = fit_precision_cholesky(X_scaled, graph_u_sub, verbose_level=2)
+
+        # Add to block-diagonal full precision
+        Prec_u = sp.sparse.block_diag((Prec_u, Prec_u_sub), format="csc")
```
Blunde1, line 894: This is good 👍

Blunde1:
Har sett gjennom `_es_update` med litt kommentarar. Hovuddelen er bra.

Lasting av data: Vi må også filtrere ut responsar/observasjonar med NaN. Eg er ganske sikker på at dette blir gjort som preprosessering av data for dei andre oppdaterings algoritmene.

src/ert/analysis/_es_update.py
Blunde1: Standard kode har snippet
```py
ens_mean_mask = abs(observations - ens_mean) <= alpha * (ens_std + scaled_errors)
obs_mask = np.logical_and(ens_mean_mask, ens_std_mask)
```
som fangar NaN i responsar også.

Blunde1: This modification might work?

```python
def fetch_responses_and_parameters_from_storage(prior):
    # Get responses
    iens_active_index = np.flatnonzero(prior.get_realization_mask_with_parameters())
    selected_observations = prior.experiment.observation_keys
    observations_and_responses = prior.get_observations_and_responses(
        selected_observations,
        iens_active_index,
    )
    S = observations_and_responses.select(
        observations_and_responses.columns[5:]
    ).to_numpy()
    observation_values = (
        observations_and_responses.select("observations").to_numpy().reshape((-1,))
    )
    observation_errors = observations_and_responses.select("std").to_numpy().reshape((-1,))
    observation_names = (
        observations_and_responses.select("observation_key").to_numpy().reshape((-1,))
    )
    indexes = observations_and_responses.select("index").to_numpy().reshape((-1,))
    observation_names_unique = np.array([
        f"{k}: {i}" for k, i in zip(observation_names, indexes)
    ])
    Y = S.T.copy()
    d = observation_values
    d_sd = observation_errors

    # NaN detection and removal
    nan_mask = np.any(np.isnan(S), axis=1)
    if np.any(nan_mask):
        obs_with_nan = np.array([
            f"{key}: {idx}" for key, idx in zip(observation_names[nan_mask], indexes[nan_mask])
        ])
        logger.warning("Deactivating observations due to NaNs in responses: %s", obs_with_nan.tolist())
        S = S[~nan_mask]
        d = d[~nan_mask]
        d_sd = d_sd[~nan_mask]
        observation_names_unique = observation_names_unique[~nan_mask]

    # Get prior and standardize it
    parameter_groups = list(prior.experiment.parameter_info.keys())
    X_full, param_group_bounds, parameter_names = _all_parameters(
        ensemble=prior,
        iens_active_index=iens_active_index,
        param_groups=parameter_groups,
    )
    scaler_x = StandardScaler()
    U_prior = scaler_x.fit_transform(X_full.T)

    return U_prior, scaler_x, S.T, d, d_sd, parameter_groups, iens_active_index, parameter_names, observation_names_unique, param_group_bounds
```

src/ert/analysis/_es_update.py
yngve-sk: Snapshots need to be propagated for GUI to work the same, it is currently a side effect of calling `_load_observations_and_responses` (which also has misfit preprocessor and non outlier detection, which we want to skip for EnIF?), I split it out so we can send them from both places without doing a copypaste.

src/ert/analysis/_es_update.py
yngve-sk: I will add a comment here, it is a bit implicit, first 5 columns are 
```
                                "response_key",
                                "index",
                                "observation_key",
                                "observations",
                                "std",
```

src/ert/analysis/_es_update.py
yngve-sk: Hmm, I have some questions wrt this:

**On dropping NaNs**
=================
We have these options
1. Drop all rows where the response is NaN for at least one realization
2. Ignore NaNs by setting them to the mean response value (or something like that), this is less loss of data, an ensemble may have responses for 99 realizations and nans for 1 of them.
3. Drop all realizations where there is at least one NaN (guessing this is not ideal)

AFAIK Your suggestion here is option 1, is that the best thing to do here?

**Storing K**:
`K = np.linalg.inv(gtmap.Prec_u.toarray()) @ gtmap.H.T @ gtmap.Prec_residual_noisy()`
We will store Prec_U and H as far as I know, is it slow/expensive if we just store the residual, and recompute K from that? 

**Outlier detection/mean masking** 
We only want to filter NaNs and drop the rest of the stuff there I guess?

yngve-sk: Wrt code organization / structuring I think the function you pasted in does mostly the same, I think it would be good to not keep two parallel almost-the-same versions of `_load_observations_and_responses` there though, so I would lean towards keeping the logic itself inlined in `enif_update`, rather splitting up `_load_observations_and_responses` to reuse some of that logic within `enif_update`

yngve-sk: `observation_names_unique, param_group_bounds and parameter_names` are already in storage :)

yngve-sk: After short discussion: Will add explicit nan-dropping to ERT and keep the same in the enif. Dropping will:
(with realizations as columns)
1. Drop rows with nan responses
2. Drop realizations with nan responses

Blunde1: Drop observations if there exist NaN responses of it

src/ert/analysis/_es_update.py
```diff
@@ -702,3 +733,229 @@ def smoother_update(
...
+        config_node = source_ensemble.experiment.parameter_configuration[param_group]
+        X_local = _load_param_ensemble_array(
+            source_ensemble, param_group, iens_active_index
+        )
+        X_local_scaler = StandardScaler()
+        X_scaled = X_local_scaler.fit_transform(X_local.T)
+
+        graph_u_sub = config_node.load_parameter_graph()
+
+        # This will work for dim(X_scaled) on order O(n^5)
```
Blunde1, line 888: Will likely add `fit_precision_cholesky_approximate()` to `graphite-maps`. This should scale better, also `O(n^6 or n^7)`

src/ert/config/field.py
```diff
@@ -26,6 +27,60 @@
...
+                if y < py - 1:
+                    G.add_edge(index, index + pz)
+
+                # Connect to the bottom neighbor (x-direction)
+                if x < px - 1:
+                    G.add_edge(index, index + py * pz)
+
+                # Connect to the neighbor in front (z-direction)
+                if z < pz - 1:
+                    G.add_edge(index, index + 1)
```
oyvindeide, line 55:
```suggestion
for x, y, z in itertools.product(range(px), range(py), range(pz)):
    # Flatten the 3D index to a single index
    index = x * py * pz + y * pz + z

    # Connect to the right neighbor (y-direction)
    if y < py - 1:
        G.add_edge(index, index + pz)

    # Connect to the bottom neighbor (x-direction)
    if x < px - 1:
        G.add_edge(index, index + py * pz)

    # Connect to the neighbor in front (z-direction)
    if z < pz - 1:
        G.add_edge(index, index + 1)
```

src/ert/config/surface_config.py
```diff
@@ -20,6 +21,27 @@
     from ert.storage import Ensemble
 
 
+def create_flattened_2d_graph(px: int, py: int) -> nx.Graph[int]:
```
oyvindeide, line 24:
```suggestion
def load_parameter_graph(self) -> nx.Graph:
```

src/ert/config/surface_config.py
```diff
@@ -20,6 +21,27 @@
...
+    """Graph created with nodes numbered from 0 to px*py
+    corresponds to the "vectorization" or flattening of
+    a 2D cube with shape (px,py) in the same way as
+    reshaping such a surface into a one-dimensional array.
+    The indexing scheme used to create the graph reflects
+    this flattening process"""
+
+    G: nx.Graph[int] = nx.Graph()
+    for i in range(px):
+        for j in range(py):
```
oyvindeide, line 34:
```suggestion
    for i, j in itertools.product(range(px), range(py)):
```

tests/ert/unit_tests/config/test_field.py
```diff
@@ -56,6 +59,78 @@ def test_write_to_runpath_produces_the_transformed_field_in_storage(
...
+        )
+        .decode("utf-8")
+        .strip()
+        + "\n"
+    )
+
+
+def test_field_grid_mask_correspondence_all_false(snake_oil_field_config, snapshot):
+    snake_oil_field_config.mask[:, :, :] = False
+    snapshot.assert_match(create_graph_snapshot(snake_oil_field_config), "graph.json")
```
oyvindeide, line 99: Think you can make this a bit easier to read:

```python
def test_field_grid_mask_correspondence_all_false(tmp_path, monkeypatch):
    monkeypatch.chdir(tmp_path)
    field = Field(
        name="some_name",
        forward_init=True,
        nx=2,
        ny=2,
        nz=1,
        file_format=FieldFileFormat.ROFF,
        update=True,
        mask_file=Path("grid_mask.npy"),
    )
    np.save("grid_mask.npy", np.zeros(field.nx * field.ny * field.nz))
    graph = field.load_parameter_graph()
    assert nx.node_link_data(graph)["links"] == [
        {"source": 0, "target": 1},
        {"source": 0, "target": 2},
        {"source": 1, "target": 3},
        {"source": 2, "target": 3},
    ]
```
could move the creation of the Field to a fixture to reuse it, then you can set the mask explicitly. much like you did for surface.

tests/ert/unit_tests/config/test_gen_kw_config.py
```diff
@@ -794,3 +795,27 @@ def test_validation_derrf_distribution(
                 GenKwConfig.from_config_list(config_list)
         else:
             GenKwConfig.from_config_list(config_list)
+
+
+@pytest.mark.parametrize("num_tfs", [0, 1, 3, 8, 15, 40, 100, 1000])
```
oyvindeide, line 800: Not sure you need that many in the parameterize, could potentially do `hypothesis` with a low sample count instead.

src/ert/run_models/base_run_model.py
```diff
@@ -941,21 +946,39 @@ def update(
...
+                enif_update(
+                    prior,
+                    posterior,
+                    update_settings=self._update_settings,
+                    es_settings=self._analysis_settings,
+                    parameters=prior.experiment.update_parameters,
+                    observations=prior.experiment.observation_keys,
+                    global_scaling=weight,
+                    rng=self.rng,
+                    random_seed=self.random_seed,
```
oyvindeide, line 975: Why is this different from the base model? We should use the same for both?

src/ert/analysis/_es_update.py
```diff
@@ -782,3 +804,228 @@ def smoother_update(
...
+        Prec_u=Prec_u,
+        Prec_eps=Prec_eps,
+        H=H,
+    )
+
+    update_indices = gtmap.get_update_indices(
+        neighbor_propagation_order=15, verbose_level=1
+    )
+
+    # Call transport? might have to do some coding here
```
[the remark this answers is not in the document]
yngve-sk: Not sure what it means really. @Blunde1 probably knows.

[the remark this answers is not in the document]
Blunde1: `transport` is called on the EnIF object below. It is called with `iterative=True,` which leads to conjugate gradient sparse **iterative** solver (instead of a direct solver with Cholesky factorization).

I figure this was an old comment together with the next line
`# Perhaps use an iterative solver instead of direct spsolve or similar`

tests/ert/unit_tests/config/test_gen_kw_config.py
yngve-sk: True, will reduce example

src/ert/analysis/_es_update.py
```diff
@@ -18,7 +18,14 @@
...
 import psutil
 import scipy
+import scipy as sp
+from graphite_maps.enif import EnIF  # type: ignore
+from graphite_maps.linear_regression import linear_boost_ic_regression  # type: ignore
+from graphite_maps.precision_estimation import (  # type: ignore
+    fit_precision_cholesky_approximate,
+)
 from iterative_ensemble_smoother.experimental import AdaptiveESMDA
+from sklearn.preprocessing import StandardScaler  # type: ignore
```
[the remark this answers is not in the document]
yngve-sk: None are "directly" available via pip install at the moment, we could add a folder/submodule to the repo as stubs, but not don't think it is worth it
ref
https://github.com/scikit-learn/scikit-learn/discussions/25307

tests/ert/unit_tests/config/test_field.py
yngve-sk: I simplified it some by removing the dependency to snake oil. I think the results could be inlined if we set the Z-dimension to 1, but otherwise it is a bit verbose (could still inline it though). I think we should still have the z dimension larger than 1 in the test case, otherwise we won't have the same test coverage for neighbors in the z-dimension.

src/ert/run_models/base_run_model.py
yngve-sk: The enif update needs the random seed to be detereministic in tests. ref
```
X_full = gtmap.transport(
        X_full_scaled,
        S.T,
        observation_values,
        update_indices=update_indices,
        iterative=True,
        verbose_level=5,
        seed=random_seed,
    )
```

src/ert/run_models/base_run_model.py
```diff
@@ -941,21 +946,39 @@ def update(
...
-                observations=prior.experiment.observation_keys,
-                global_scaling=weight,
-                rng=self.rng,
-                progress_callback=functools.partial(
-                    self.send_smoother_event,
-                    prior.iteration,
-                    prior.id,
-                ),
-            )
+            if not use_enif:
```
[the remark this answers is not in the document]
yngve-sk: Hmm, I think we should consolidate it to one function if we omit passing the random seed, and instead hardcode it to a default in the call to  `gtmap.transport`, mainly used to generate noise AFAIK, so the update should be the same. Changing the hardcoded value will also change the snapshot values, though. 

Passing the rng is also an option but would require changes to `graphite_maps`

Separate runmodel would be OK, mostly copypaste job.

src/ert/config/surface_config.py
```diff
@@ -165,3 +167,25 @@ def load_parameters(
         ds = ensemble.load_parameters(self.name, realizations)
         ensemble_size = len(ds.realizations)
         return ds["values"].values.reshape(ensemble_size, -1).T
+
+    def load_parameter_graph(self) -> nx.Graph[int]:
```
oyvindeide, line 171: Isnt this a special case of: `create_flattened_cube_graph`?

oyvindeide (approved):
LGTM! Remember to squash the fixups

PR #10473, opened 2025-03-26
Add info on cached evaluations to the evaluator result

verveerpj:
**Issue**
This PR adds information on cache hits to the evaluation results that are sent back to `ropt`. These are not needed by Everest, nor are they used by `ropt` directly, but serve as additional information that is reported by TNO tools.

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/run_models/everest_run_model.py
+++ b/src/ert/run_models/everest_run_model.py
@@ -685,13 +685,14 @@ class EverestRunModel(BaseRunModel):
             all_results=all_results,
         )
 
+        cache_hits_dict: list[dict[str, Any]] = []
         if cache_hits_df is not None and not cache_hits_df.is_empty():
             perturbations = (
                 evaluator_context.perturbations
                 if evaluator_context.perturbations is not None
                 else [-1] * len(model_realizations)
             )
-            event_data = (
+            cache_hits_dict = (
                 cache_hits_df.with_columns(
                     pl.col("flat_index")
                     .map_elements(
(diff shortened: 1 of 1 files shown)


src/ert/run_models/everest_run_model.py
```diff
@@ -787,24 +790,38 @@ def _forward_model_evaluator(
...
+        # simulation id. If the results came from the cache, we get the
+        # simulation id from the source batch. We aso store the source batch of
+        # the results, from from the cache info if the results was cached; if
+        # not the current batch is stored as the source. Results that were not
+        # active and therefore were not evaluated receive -1 values.
+        sim_ids = []
+        source_batch_ids = []
+        for ei in evaluation_infos:
+            if ei.simulation_id is None:
+                sim_id, batch_id = next(
```
yngve-sk, line 803: If it is not in the cache hit dict i would raise an error, if it has the status `CACHED` it should always be there I think. So either raise an explicit error or remove the silent failure defaulting to (-1, -1)

verveerpj: There is also the case that a vector of controls is passed, but marked as inactive. In that case it is not evaluated at all. So there is nothing run or cached, and we need a place-holder.

yngve-sk: Forgot about the inactive ones, would prefer adding another if checking if they are flagged as `INACTIVE` instead of defaulting, but I see we do need to indicate the inactive-ness somehow now

PR #10478, opened 2025-03-27
Allow default sheet to be None in Design Matrix

### LINKED ISSUE #10475: DESIGN_MATRIX: make default sheet not required
As default sheet in the design matrix excel sheet can be empty, we don't need to make the presence of default sheet required.

larsevj:
**Issue**
Resolves #10475 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/design_matrix.py
+++ b/src/ert/config/design_matrix.py
@@ -24,7 +24,7 @@ from ert.shared.status.utils import convert_to_numeric
 class DesignMatrix:
     xls_filename: Path
     design_sheet: str
-    default_sheet: str
+    default_sheet: str | None
 
     def __post_init__(self) -> None:
         try:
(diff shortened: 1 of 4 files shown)


src/ert/config/design_matrix.py
```diff
@@ -44,7 +44,7 @@ def from_config_list(cls, config_list: list[str | dict[str, str]]) -> DesignMatr
         filename = Path(cast(str, config_list[0]))
         options = cast(dict[str, str], config_list[1])
         design_sheet = options.get("DESIGN_SHEET", "DesignSheet")
-        default_sheet = options.get("DEFAULT_SHEET", "DefaultSheet")
+        default_sheet = options.get("DEFAULT_SHEET", None)
```
xjules, line 47: So not sure here now. If we specify just the name of XLSL file, by default it does not expect the DEFAULT_SHEET?

xjules: I sort of agree, but essentially what that means is if there is a default sheet it needs to be specified while design sheet not. And thus we can end up with something like this:
`DESIGN_MATRIX design.xlsl DEFAULT_SHEET:some_name`
which probably is fine

xjules (approved):
Just had a small comment, but generally I think it is fine :rocket:

Issue #10492, opened 2025-03-28
excessive warnings in tests


eivindjahren:
The following warnings are shown thousands of times in our tests
```
src/ert/storage/local_ensemble.py:989: UserWarning: Sortedness of columns cannot be checked when 'by' groups provided
  joined = observations_for_type.join_asof(
```

PR #10495, opened 2025-03-28
Sort time data explicitly before asof join

yngve-sk:
**Issue**
Resolves [#10492](https://github.com/equinor/ert/issues/10492)

Performance tests @ `test_obs_and_responses_performance.py` should catch performance regressions


xjules (approved):
Nice! :rocket:

PR #10499, opened 2025-03-28
Extend ruff linting with FURB and a little bit TRY

berland:
**Issue**
Resolves tiny bit of issues that ruff is able to detect.

**Approach**
🔧 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/parsing/context_values.py
+++ b/src/ert/config/parsing/context_values.py
@@ -69,7 +69,7 @@ class ContextFloat(float):
         return new_instance
 
 
-class ContextString(str):
+class ContextString(str):  # noqa: FURB189
     @classmethod
     def from_token(cls, token: FileContextToken) -> "ContextString":
         return cls(val=str(token), token=token, keyword_token=token)
@@ -92,7 +92,7 @@ class ContextString(str):
 T = TypeVar("T")
 
 
-class ContextList(list[T]):
+class ContextList(list[T]):  # noqa: FURB189
     keyword_token: FileContextToken
 
     def __init__(self, token: FileContextToken) -> None:
--- a/tests/ert/unit_tests/gui/plottery/test_plot_limits.py
+++ b/tests/ert/unit_tests/gui/plottery/test_plot_limits.py
@@ -31,8 +31,8 @@ def test_plot_limits():
     plot_limits = PlotLimits()
     limit_names = ["value", "index", "count", "density", "date"]
 
-    positive_floats = [1.0, 1.5, 3.1415, 1e10, 5.2e-7]
-    negative_floats = [-1.0, -1.5, -3.1415, -1e10, -5.2e-7]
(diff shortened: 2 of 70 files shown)


src/ert/config/parsing/context_values.py
```diff
@@ -69,7 +69,7 @@ def __deepcopy__(self, memo):
         return new_instance
 
 
-class ContextString(str):
+class ContextString(str):  # noqa: FURB189
```
eivindjahren, line 72: What is the message for this?

berland: # subclass-builtin (FURB189)

Derived from the **refurb** linter.

Fix is always available.

## What it does
Checks for subclasses of `dict`, `list` or `str`.

## Why is this bad?
Subclassing `dict`, `list`, or `str` objects can be error prone, use the
`UserDict`, `UserList`, and `UserString` objects from the `collections` module
instead.

## Example
```python
class CaseInsensitiveDict(dict): ...
```

Use instead:
```python
from collections import UserDict


class CaseInsensitiveDict(UserDict): ...
```

## Fix safety
This fix is marked as unsafe because `isinstance()` checks for `dict`,
`list`, and `str` types will fail when using the corresponding User class.
If you need to pass custom `dict` or `list` objects to code you don't
control, ignore this check. If you do control the code, consider using
the following type checks instead:

* `dict` -> `collections.abc.MutableMapping`
* `list` -> `collections.abc.MutableSequence`
* `str` -> No such conversion exists

## References

- [Python documentation: `collections`](https://docs.python.org/3/library/collections.html)

berland: But I tried using the UserStuff here, but it did not work, so I just assumed we have to go rogue and subclass the builtins.

tests/ert/unit_tests/gui/plottery/test_plot_limits.py
```diff
@@ -31,8 +32,8 @@ def test_plot_limits():
     plot_limits = PlotLimits()
     limit_names = ["value", "index", "count", "density", "date"]
 
-    positive_floats = [1.0, 1.5, 3.1415, 1e10, 5.2e-7]
-    negative_floats = [-1.0, -1.5, -3.1415, -1e10, -5.2e-7]
+    positive_floats = [1.0, 1.5, math.pi, 1e10, 5.2e-7]
```
eivindjahren, line 35: Seems a bit excessive. I suggest just removing 3.1415 from the test data.

src/ert/config/parsing/context_values.py
eivindjahren: Yea, I guess we would have to try that separately.

eivindjahren (approved):
Seems like it works fine. I had a qustion about silencing the UserStr part, but we can fix that up separately.

PR #10504, opened 2025-03-31
Rewrite everest run gui

### LINKED ISSUE #10278: Rewrite everest rundialog to only require output path and not entire everest config
The Everest GUI should only need to know about the storage, and events etc, all this information exists within the output directory, so that should be enough information for it to monitor the progress of the current Everest run. One key point to remember: The output directory also contains the credentials & connection details to connect to all the everserver endpoints for receiving events etc etc.

For this to happen cleanly, we need to expose the information it currently gets from the config, through endpoints in `everserver.py`.

frode-aarstad:
**Issue**
Resolves 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/run_models/base_run_model.py
+++ b/src/ert/run_models/base_run_model.py
@@ -145,6 +145,7 @@ class BaseRunModelAPI:
     cancel: Callable[[], None]
     get_runtime: Callable[[], int]
     has_failed_realizations: Callable[[], bool]
+    config_file: str
 
 
 class BaseRunModel(ABC):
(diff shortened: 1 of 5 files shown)


yngve-sk:
Hmm. I don't think adding the `config_file` arg to the `BaseRunModelAPI` is a step in the right direction as we want separation between the server and the client. I think the `everserver` should have access to the everest config and mirror the necessary information through some endpoints. The `queue_system` can be removed, and the `config_file` is not necessary either. The client should only require the output directory to connect to the server, and then get whatever else it needs by doing requests to the server. The moving of some of the logic from mainwindow to main is not directly related to the issue and should be a separate PR/commit.

PR #10511, opened 2025-03-31
Move numpy saving/loading params logic to storage

yngve-sk:
Defer to after some discussions relating to https://github.com/equinor/ert/pull/10095


xjules (approved):
This makes a lot of sense! :rocket:

PR #10538, opened 2025-04-03
Add typing to shared_data

### LINKED ISSUE #10516: everserver shared_data is not typed
it should be more descriptive than dict[str, Any]

yngve-sk:
**Issue**
Resolves #10516

### DIFF
--- a/src/everest/detached/jobs/everserver.py
+++ b/src/everest/detached/jobs/everserver.py
@@ -1,5 +1,6 @@
 import argparse
 import asyncio
+import dataclasses
 import datetime
 import json
 import logging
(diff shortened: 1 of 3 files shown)


src/everest/detached/__init__.py
```diff
@@ -324,6 +324,14 @@ def update_everserver_status(
     everserver_status_path: str, status: ServerStatus, message: str | None = None
 ) -> None:
     """Update the everest server status with new status information"""
+
+    if not isinstance(message, str):
+        print(f"message not string: {message}")
+        traceback.print_stack()
+
+        # breakpoint()
```
StephanDeHoop, line 332: This was for debugging purpose? I guess this includes the whole if statement and print afterwards? Since we type that `message: str` something goes clearly wrong, is that why you want to print the stack trace in any case (not just when running in debug verbosity, if that is even a thing for `ever_server`).

src/everest/detached/__init__.py
```diff
@@ -355,6 +363,7 @@ def everserver_status(everserver_status_path: str) -> dict[str, Any]:
     """
     if os.path.exists(everserver_status_path):
         with open(everserver_status_path, encoding="utf-8") as f:
-            return json.load(f, object_hook=ServerStatusEncoder.decode)
+            a = json.load(f, object_hook=ServerStatusEncoder.decode)
```
StephanDeHoop, line 366: Why is this necessary :o?

StephanDeHoop (approved):
Looks good :)! Some minor comments.

src/everest/detached/__init__.py
yngve-sk: Ohh oops, that and what was below was debugging artifacts, will remove!

PR #10553, opened 2025-04-03
Have gui notify user if high contrast mode is enabled

### LINKED ISSUE #10336: Ert gui does not work with high contrast mode

jonathan-eq:
**Issue**
Resolves #10336 

**Approach**
This commit makes the gui emit a QMessageBox on startup, stating that the user has high contrast mode enabled, and that it is not by ert and some features might not work as expected.

image

The gui still looks like this, but the previous fix failed (e37a8389f144ef5f219e74d9aa8f6df1775dca6c), so this is the simplest solution for now. 
image

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/gui/main_window.py
+++ b/src/ert/gui/main_window.py
@@ -4,17 +4,20 @@ import datetime
 import functools
 import webbrowser
 from pathlib import Path
+from typing import cast
 
 from PyQt6.QtCore import QCoreApplication, QEvent, QSize, Qt
 from PyQt6.QtCore import pyqtSignal as Signal
 from PyQt6.QtCore import pyqtSlot as Slot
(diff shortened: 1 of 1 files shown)


xjules (approved):
I think it is fine to just have it so. :rocket:

PR #10559, opened 2025-04-04
Replace QColor with state string in model

andreas-el:
Stores realization and forward_model state as string in snapshot model instead of QColor object.
Performs lookup in static dict when drawing.

      --exec 'just rapid-tests'`)

## When applicable


src/ert/cli/monitor.py
```diff
@@ -115,7 +114,7 @@ def _get_legends(self) -> str:
         latest_snapshot = self._snapshots[max(self._snapshots.keys())]
         total_count = len(latest_snapshot.reals)
         aggregate = latest_snapshot.aggregate_real_states()
-        for state_ in ALL_REALIZATION_STATES:
+        for state_ in REAL_STATE_TO_COLOR:
```
berland, line 117: This looks like it could change the order of states printed to the terminal?

andreas-el: It did. So I changed the order in the dict to match the tests again.

src/ert/gui/model/node.py
```diff
@@ -73,8 +71,8 @@ def __repr__(self) -> str:
 class RealNodeData:
     status: str | None = None
     active: bool | None = False
-    fm_step_status_color_by_id: dict[str, QColor] = field(default_factory=dict)
-    real_status_color: QColor | None = None
+    fm_step_status_color_by_id: dict[str, str] = field(default_factory=dict)
```
berland, line 74: "color" should maybe not be part of the variable name since there is only descriptive strings for states in the dict?

andreas-el: Agreed. I've renamed variables that do not contain color anymore.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -85,10 +84,10 @@ def convert_iso8601_to_datetime(
 
 
 class EnsembleSnapshotMetadata(TypedDict):
-    # contains the QColor used in the GUI for each fm_step
-    aggr_fm_step_status_colors: defaultdict[RealId, dict[FmStepId, QColor]]
-    # contains the QColor used in the GUI for each real
-    real_status_colors: dict[RealId, QColor]
+    # contains state mapped to QColor used in the GUI for each fm_step
```
andreas-el, line 87: Not sure if these should be removed as well.

berland: The comment? Maybe it does not add much value no. It kind of still says that this dict is the map to QColor, but it is only meant to be used for that (?)

berland (approved):

PR #10560, opened 2025-04-04
Remove experiment-type column from Manage experiments

### LINKED ISSUE #10470: ManageExperiments Type is always `None`

Image

These identifiers looks to have been removed in e5468950b6adbc925113fad1fbe58f33879665f0

Expected to see identifiers, or have the column removed if not deemed useful.

### steps to reproduce

```Python
Run any simulation, open manage experiments and see the `None` value in the `Type` column.
```

- [x] python 3.11
- [ ] python 3.12
- [x] macosx
- [ ] rhel7
- [x] rhel8
- [x] local queue
- [x] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

SAKavli:
Ref discussion in #10470
The ForegroundRole conditional statement's purpose is to get a specific color from the palette for the type column when experiment_type is undefined.

**Issue**
Resolves #10470 

**Approach**

As the issue discussion concluded in removing the column, my approach was to:

- Find and remove definition of the Type column
- Remove code which references and depends to the Type property of the class being present

GUI before:

image

GUI after:

image

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/gui/tools/manage_experiments/storage_model.py
+++ b/src/ert/gui/tools/manage_experiments/storage_model.py
@@ -92,7 +92,6 @@ class ExperimentModel:
         self._parent = parent
         self._id = experiment.id
         self._name = experiment.name
-        self._experiment_type = experiment.metadata.get("ensemble_type")
         self._children: list[EnsembleModel] = []
 
     def add_ensemble(self, ensemble: EnsembleModel) -> None:
--- a/src/ert/gui/tools/manage_experiments/storage_model.py
+++ b/src/ert/gui/tools/manage_experiments/storage_model.py
@@ -5,7 +5,6 @@ from uuid import UUID
 import humanize
 from PyQt6.QtCore import QAbstractItemModel, QModelIndex, QObject, Qt
 from PyQt6.QtCore import pyqtSlot as Slot
-from PyQt6.QtWidgets import QApplication
 from typing_extensions import override
 
 from ert.storage import Ensemble, Experiment, Storage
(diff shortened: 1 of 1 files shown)


src/ert/gui/tools/manage_experiments/storage_model.py
```diff
@@ -122,13 +119,6 @@ def data(
                     if self._children
                     else "None"
                 )
-            if col == _Column.TYPE:
-                return self._experiment_type or "None"
```
andreas-el, line 126: Could you check if `self._experiment_type` could also be removed?
You could have removed all references to this.

andreas-el:
> Would like input on labeltype for this issue/PR. Also, should there be a test when a feature is removed, if categorized as bug?

I don't think this warrants a test. Although it was marked as a bug, it does not have any impact other than missing information that we now remove. If this was something we would like to have, adding a test makes sense.

Regarding the labels, perhaps  release-notes:`Maintainance` or `Skip` would be most appropriate.

A title to consider could be (although yours is fine too):
Remove experiment-type column from Manage experiments

src/ert/gui/tools/manage_experiments/storage_model.py
SAKavli: I agree, I have done the following:

Investigated all referenced of ExperimentModel class:
image
Found no references using the Type attribute for anything

Found no references for the property in the repo in general:
image

Tried to run polyexample with the line removed, no issues. 

Safe to remove?

andreas-el: Yes. Go ahead! 👍

andreas-el (approved):
Nice work! 🏖️

PR #10563, opened 2025-04-04
[Backport version-14.1] Have gui notify user if high contrast mode is enabled

larsevj:
# Description
Backport of #10553 to `version-14.1`.


jonathan-eq (approved):
Very good!

PR #10593, opened 2025-04-09
Make EverestClient only need output dir

yngve-sk:
**Issue**
Resolves [#10278](https://github.com/equinor/ert/issues/10278)

**Approach**
Get information about currently running experiment through server endpoints only


src/everest/gui/everest_client.py
```diff
@@ -91,37 +119,33 @@ def passthrough_ws_events() -> None:
...
         def start_fn(
             evaluator_server_config: EvaluatorServerConfig, restart: bool = False
         ) -> None:
             pass
 
         return BaseRunModelAPI(
-            experiment_name="Everest Experiment",
-            runpath_format_string=runpath_format_string,
+            experiment_name=self.config_filename,
+            runpath_format_string=self.simulation_dir,
```
verveerpj, line 130: This is a little bit confusing, is `simulation_dir` actually a format string?

src/everest/gui/everest_client.py
```diff
@@ -91,37 +119,33 @@ def passthrough_ws_events() -> None:
...
             support_restart=False,
             start_simulations_thread=start_fn,
             cancel=self.stop,
-            get_runtime=lambda: -1,  # Not currently shown in Everest gui
+            get_runtime=self.get_runtime,  # Not currently shown in Everest gui
```
verveerpj, line 134: Is it still not show, even if you pass it?

verveerpj (approved):
LGTM, small questions only.

src/everest/gui/everest_client.py
yngve-sk: Ohh, it does show, I forgot to remove the comment, will fix

src/everest/gui/everest_client.py
yngve-sk: Yes, in the ERT rundialog we currently take in the runpath format string to track the disk space, which corresponds to the everest simulation dir (at least per the old behavior which this PR just replicates as server-to-client)

verveerpj: Should it be called `simulation_dir` then?

yngve-sk: Hmm. The BaseRunModelAPI is mainly for the ERT gui, as is the RunDialog, so I think we should stick to the naming convention there. The Everest RunDialog is basically the ERT rundialog monitoring running ERT simulations. In Everest we just call the runpath `simulation_dir`, we should maybe think about consolidating these terms though, but in Everest lingo and wrt the users I feel they feel more at home with "simulations" / "simulations_dir" and less so with runpath?

verveerpj: I don't know, personally I find the simulations terms very weird. It probably comes from running the simulator, but in reality we are funning forward models that do more than just the simulation. Maybe it is something to discuss and we might be better of adjusting to the ERT terminology.

PR #10598, opened 2025-04-10
Fix error racecondition bug

### LINKED ISSUE #9366: Logging of failures when MAX_SUBMIT>1 goes wrong due to invalid XML
Since 

https://github.com/equinor/ert/blob/8f8a329e12fcaf826b5914f98f51a2133cbc6564/src/_ert/forward_model_runner/reporting/file.py#L171

is appending XML contents to a file, this file will not be valid XML if a realization is rerun through MAX_SUBMIT.

The parser of this file:

https://github.com/equinor/ert/blob/8f8a329e12fcaf826b5914f98f51a2133cbc6564/src/ert/scheduler/job.py#L317

will then fail (but still log the contents) parsing the file. 

Example seen in logs: 

```
job failed with invalid XML ERROR file, contents '<error>
  <time>14:18:12</time>
  <job>DESIGN2PARAMS</job>
  <reason>Process exited with status code
...
```

SAKavli:
**Issue**
Resolves #9366 

**Approach**

- Reproducing this error was troublesome and confusing at first.
- We figured out the error was due to a racecondition between different job submissions, where each submission is supposed to delete the ERROR file. However, this is not always done if the first submission is still writing when the new submission starts up. This happens in some cases and is difficult to reproduce.
- Why this happens is because the code appends text to the ERROR message. 
- We found this to be unnecessary, and that it should be changed to (over)writing the ERROR file instead of appending.
- Along the way, we also found an unnecessary alias for open() in append mode, which we found confusing. We chose to remove this in the same go.

### DIFF
--- a/src/_ert/forward_model_runner/reporting/file.py
+++ b/src/_ert/forward_model_runner/reporting/file.py
@@ -169,7 +169,7 @@ class File(Reporter):
 
     @staticmethod
     def _dump_error_file(fm_step, error_msg):
-        with open(ERROR_file, "a", encoding="utf-8") as file:
+        with open(ERROR_file, "w", encoding="utf-8") as file:
             file.write("<error>\n")
             file.write(
                 f"  <time>{time.strftime(TIME_FORMAT, time.localtime())}</time>\n"
--- a/src/_ert/forward_model_runner/reporting/file.py
+++ b/src/_ert/forward_model_runner/reporting/file.py
@@ -1,4 +1,3 @@
-import functools
 import logging
 import os
 import socket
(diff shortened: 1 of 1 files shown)


berland:
The style check is probably solved through rebasing main. 

Also recommend to use pre-commit to avoid style errors being pushed (it costs github actions cpu seconds)

SAKavli:
I have the pre-commit activated, but hadn't pulled in a while. I believe it was a matter of seconds the failing commit was merged into main.
Do you also have some automatic pulling on pre-commits?

berland:
> I have the pre-commit activated, but hadn't pulled in a while. I believe it was a matter of seconds the failing commit was merged into main before i pushed. Is there some some automatic pulling on pre-commits, or is this not a thing you would use?

No, we do not have auto-pull on pre-commit. The style issue sneak into main because it was tested only in its non-updated branch before merging. One of the rare problems we have that would have been caught with a merge-queue.

berland (approved):
splendid

PR #10601, opened 2025-04-10
Refactor max_batch_num handling

verveerpj:
**Issue**
The handling of the `max_batch_num` keyword in Everest was sub-optimal:

1. It "mis-used" the user abort function of `ropt`, which makes it more difficult to distinguish between the two events for aborting optimization.
2. It included the potential extra batch run that is sometimes done at the beginning for auto-scaling purposes, meaning that enabling auto-scaling changes the meaning of `max_batch_num` somewhat.

**Approach**
`ropt` now supports stopping after a given number of batches directly, with a corresponding correct exit code. The custom handling of `max_batch_num` can be removed.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/run_models/everest_run_model.py
+++ b/src/ert/run_models/everest_run_model.py
@@ -331,6 +331,8 @@ class EverestRunModel(BaseRunModel):
             match optimizer_exit_code:
                 case OptimizerExitCode.MAX_FUNCTIONS_REACHED:
                     self._exit_code = EverestExitCode.MAX_FUNCTIONS_REACHED
+                case OptimizerExitCode.MAX_BATCHES_REACHED:
+                    self._exit_code = EverestExitCode.MAX_BATCH_NUM_REACHED
                 case OptimizerExitCode.USER_ABORT:
                     self._exit_code = EverestExitCode.USER_ABORT
                 case OptimizerExitCode.TOO_FEW_REALIZATIONS:
@@ -403,14 +405,6 @@ class EverestRunModel(BaseRunModel):
 
     def _check_for_abort(self) -> bool:
         logging.getLogger(EVEREST).debug("Optimization callback called")
-        if (
-            self._everest_config.optimization is not None
-            and self._everest_config.optimization.max_batch_num is not None
-            and (self._batch_id >= self._everest_config.optimization.max_batch_num)
-        ):
-            self._exit_code = EverestExitCode.MAX_BATCH_NUM_REACHED
-            logging.getLogger(EVEREST).info("Maximum number of batches reached")
-            return True
         if (
             self._opt_callback is not None
             and self._opt_callback() == "stop_optimization"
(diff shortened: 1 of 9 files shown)


tests/everest/test_math_func.py
```diff
@@ -193,7 +193,7 @@ def test_math_func_auto_scaled_constraints(copy_math_func_test_data_to_tmp):
 
     # control number of batches, no need for full convergence:
     config_dict["optimization"]["convergence_tolerance"] = 1e-10
-    config_dict["optimization"]["max_batch_num"] = 3
+    config_dict["optimization"]["max_batch_num"] = 2
```
yngve-sk, line 196: Assuming this is functionally equivalent to the previous behavior? Only possible with this new change, or a drive-by change?

tests/everest/test_simulator_cache.py
```diff
@@ -24,7 +24,7 @@ def new_call(*args):
 
     config = EverestConfig.load_file("config_minimal.yml")
     config_dict = config.model_dump(exclude_none=True)
-    config_dict["optimization"]["max_batch_num"] = 5
+    config_dict["optimization"]["max_batch_num"] = 4
```
yngve-sk, line 27: How does this relate to the change in handling?

src/ert/run_models/everest_run_model.py
```diff
@@ -403,14 +405,6 @@ def _init_domain_transforms(
...
-        if (
-            self._everest_config.optimization is not None
-            and self._everest_config.optimization.max_batch_num is not None
-            and (self._batch_id >= self._everest_config.optimization.max_batch_num)
-        ):
-            self._exit_code = EverestExitCode.MAX_BATCH_NUM_REACHED
-            logging.getLogger(EVEREST).info("Maximum number of batches reached")
-            return True
         if (
             self._opt_callback is not None
```
yngve-sk, line 409: This function is now only called when ROPT decides to abort I take it, and this is only for propagating it into the logs, could you rename the function to something more in line with that? Also I guess it doesn't have to return anything anymore?

tests/everest/test_math_func.py
verveerpj: The change in behavior was actually when we introduced the extra simulations for the auto-scaling. After that, `max_batch_num` did not represent the actual number of batches in the optimization run, but added one extra. This PR reverts to the old behavior, hence one less in the corresponding tests.

tests/everest/test_simulator_cache.py
verveerpj: Same as above.

src/ert/run_models/everest_run_model.py
verveerpj: This function is called before each new evaluation. It still calls the Everest callback that checks if the user requested the optimization to stop. So, the name is still correct, and it still has to return a bool to indicate if an abort was requested.

yngve-sk (approved):
LGTM!

PR #10634, opened 2025-04-15
Remove event_from_dict function and ZMQ client from LegacyEnsemble

jonathan-eq:
**Approach**
This first commit in this PR removes the event_from_dict function and makes the callers create the Events directly instead. This increases the readability and makes it easier to find where the events are actually being created without having to jump through hoops and doing eight spins. 

The second commit makes the LegacyEnsemble put events directly in the evaluator queue (which it has access to) instead of connecting to EE via ZMQ client and sending events. Before this commit, the client it used was also called dispatcher, which caused some confusion as to who was actually creating and sending Ensemble{Started,Cancelled,Stopped}.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/_ensemble.py
+++ b/src/ert/ensemble_evaluator/_ensemble.py
@@ -8,12 +8,14 @@ from dataclasses import dataclass
 from typing import Any
 
 from _ert.events import (
+    EnsembleCancelled,
+    EnsembleFailed,
+    EnsembleStarted,
+    EnsembleSucceeded,
     Event,
     FMEvent,
     ForwardModelStepFailure,
     ForwardModelStepSuccess,
-    Id,
-    event_from_dict,
     event_to_json,
 )
 from _ert.forward_model_runner.client import Client
@@ -192,16 +194,6 @@ class LegacyEnsemble:
         async with Client(self._config.get_uri(), token=self._config.token) as client:
             await client.send(event_to_json(event), retries)
 
-    def generate_event_creator(self) -> Callable[[Id.ENSEMBLE_TYPES], Event]:
-        def event_builder(status: str) -> Event:
-            event = {
-                "event_type": status,
-                "ensemble": self.id_,
-            }
-            return event_from_dict(event)
-
-        return event_builder
-
     async def evaluate(
         self,
         config: EvaluatorServerConfig,
@@ -217,7 +209,6 @@ class LegacyEnsemble:
         the final result of executing all its jobs through a scheduler and driver.
         """
         self._config = config
-        event_creator = self.generate_event_creator()
 
         if not self.id_:
             raise ValueError("Ensemble id not set")
@@ -238,7 +229,7 @@ class LegacyEnsemble:
                 ee_token=self._config.token,
             )
 
-            await self.send_event(event_creator(Id.ENSEMBLE_STARTED))
+            await self.send_event(EnsembleStarted(ensemble=self.id_))
 
             min_required_realizations = (
                 self.min_required_realizations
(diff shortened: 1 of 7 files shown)


src/ert/ensemble_evaluator/_ensemble.py
```diff
@@ -209,6 +198,7 @@ async def evaluate(
         the final result of executing all its jobs through a scheduler and driver.
         """
         self._config = config
+        outbound_event_queue = scheduler_queue or asyncio.Queue()
```
xjules, line 201: We should refactor client, maybe not in this PR. At least to create the issue for it...

src/ert/scheduler/scheduler.py
```diff
@@ -244,7 +243,7 @@ async def _monitor_and_handle_tasks(
     async def execute(
         self,
         min_required_realizations: int = 0,
-    ) -> Id.ENSEMBLE_SUCCEEDED_TYPE | Id.ENSEMBLE_CANCELLED_TYPE:
+    ) -> bool:
```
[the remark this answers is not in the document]
jonathan-eq: I will add a docstring to document this change 👌

src/ert/ensemble_evaluator/_ensemble.py
```diff
@@ -217,7 +198,7 @@ async def evaluate(
         the final result of executing all its jobs through a scheduler and driver.
         """
         self._config = config
-        event_creator = self.generate_event_creator()
+        outbound_event_queue = scheduler_queue or asyncio.Queue()
```
xjules, line 201: So scheduler_queue can be None?

jonathan-eq: Yes, but maybe this is only for testing

xjules: Have you tried the running the tests without asyncio.Queue()?

src/ert/scheduler/job.py
```diff
@@ -310,28 +306,37 @@ async def _handle_aborted(self) -> None:
...
-        elif state == JobState.ABORTED:
-            await self._handle_aborted()
-
-        elif state == JobState.COMPLETED:
-            self._end_time = time.time()
-            await self._scheduler.completed_jobs.put(self.iens)
-
-        msg = event_from_dict(event_dict)
+        self.state = state
+        if event is not None:
```
xjules, line 334: Hm, why this can't be part of the constructors above?

tests/ert/unit_tests/scheduler/test_scheduler.py
```diff
@@ -196,8 +196,8 @@ async def init(*args, **kwargs):
 
     driver = mock_driver(init=init)
     sch = scheduler.Scheduler(driver, [realization], max_submit=5)
-
-    assert await sch.execute() == Id.ENSEMBLE_SUCCEEDED
+    scheduler_finished_successfully = await sch.execute()
+    assert scheduler_finished_successfully
```
xjules, line 200: `assert await sch.execute()` ?

tests/ert/unit_tests/scheduler/test_scheduler.py
```diff
@@ -90,8 +90,8 @@ async def init(iens, *args, **kwargs):
     driver = mock_driver(init=init)
 
     sch = scheduler.Scheduler(driver, [realization])
-
-    assert await sch.execute() == Id.ENSEMBLE_SUCCEEDED
+    scheduler_finished_successfully = await sch.execute()
+    assert scheduler_finished_successfully
```
xjules, line 94: `assert await sch.execute()` ?

jonathan-eq: I changed this to increase readability. `await sch.execute()` does not indicate what is being returned (succeeded or cancelled), but maybe it is enough if I modify the docstring.

src/ert/scheduler/job.py
jonathan-eq: Then I would need to repeat them in all the arms of the match statement

src/ert/ensemble_evaluator/_ensemble.py
jonathan-eq: I think it worked 👌

xjules (approved):
Nice improvement @jonathan-eq ! :rocket:

PR #10672, opened 2025-04-23
Add more debug logging to everest run model

yngve-sk:


src/ert/run_models/everest_run_model.py
```diff
@@ -119,6 +119,9 @@ class _EvaluationInfo:
     constraints: NDArray[np.float64] | None = None
 
 
+logger = logging.getLogger(EVEREST)
```
frode-aarstad, line 122: We should use one way of logging in a file: ie `logger.debug `or `logging.getLogger(EVEREST).debug` either way is fine with me

yngve-sk: Fixed

frode-aarstad (approved):
🚀

PR #10680, opened 2025-04-24
Fixup reference to ObservationStatus.ACTIVE as "active"

yngve-sk:
tiny bugfix


frode-aarstad (approved):

oyvindeide (changes requested):
Are there no tests covering this? What is the consequence?

PR #10696, opened 2025-04-25
Catch and log Python exceptions in QT code

### LINKED ISSUE #10310: Investigate logging exceptions in Qt event loop
We did not get the exception from https://github.com/equinor/ert/issues/10306 into our logs as an exception raised from the Qt event loop results in an abort signal.

We could either trap abort signals and log those as exceptions or we could create a qt event handler and get the exception from there.

berland:
**Issue**
Resolves #10310 

**Approach**
Add a new span for the GUI part of Ert and inject a custom exception hook in Python's `sys` module in order
to propagate span status and logs to opentelemetry in case of an exception occuring in Qt's event thread. 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/gui/main.py
+++ b/src/ert/gui/main.py
@@ -2,10 +2,15 @@ from __future__ import annotations
 
 import logging
 import os
+import sys
+import time
+import traceback
+import types
 from collections import Counter
 from importlib.resources import files
 from signal import SIG_DFL, SIGINT, signal
 
+from opentelemetry.trace import Status, StatusCode
 from PyQt6.QtCore import QDir
 from PyQt6.QtGui import QIcon
 from PyQt6.QtWidgets import QApplication, QWidget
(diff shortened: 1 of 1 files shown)


berland:
Note to self: It takes 1-2 minutes for log messages to appear in Azure, but exceptions take some extra minutes after that (making the span state not necessarily correct at first sight).

berland:
Manual test procedure:

* Inject `1/0` in e.g. `src/ert/gui/tools/workflows/run_workflow_widget.py::__init__()` right before `layout = QFormLayout()`
* Do `touch test-data/ert/poly_example/wf.wf`
* Add `LOAD_WORKFLOW wf.wf` to `test-data/ert/poly_example/poly.ert`
* Run `ert gui poly.ert`, and go to the menu and select "Tools -> Run workflow".
* Wait ~2 minutes and check that the Azure logs are populated with the exception
* Wait ~2 minutes more and check that the AppException is present in Azure logs and that the span state is updated (to be failed).

berland:
Catching SIGSEGV (which can happen in the C/C++ part of Qt) seems to be an order more difficult to handle and log properly, not part of current scope.

andreas-el (approved):
💯

PR #10707, opened 2025-04-28
Fix too many open files

### LINKED ISSUE #9832: Assess "Too many open files"
In light of #9830 Erts behaviour of open files should be assessed. Some questions:
* How does it scale with number of realizations?
* Differences running locally versus on compute cluster

larsevj:
**Issue**
Resolves #9832 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/scheduler/driver.py
+++ b/src/ert/scheduler/driver.py
@@ -117,9 +117,13 @@ class Driver(ABC):
                 )
             except FileNotFoundError as e:
                 return (False, str(e))
-            except OSError:
-                await asyncio.sleep(retry_interval)
-                continue
+            except OSError as e:
+                if e.errno == 24:
+                    # Too many files open error, we try again
+                    await asyncio.sleep(retry_interval)
+                    continue
+                else:
+                    return (False, str(e))
 
             stdout, stderr = await process.communicate(stdin)
 


src/ert/scheduler/job.py
```diff
@@ -98,9 +98,11 @@ def running_duration(self) -> float:
             return time.time() - self._start_time
         return 0
 
-    async def _submit_and_run_once(self, sem: asyncio.BoundedSemaphore) -> None:
+    async def _submit_and_run_once(
+        self, run_sem: asyncio.BoundedSemaphore, kill_sem: asyncio.BoundedSemaphore
```
xjules, line 102: why the name change?

xjules: Oh, to get semantics right :) what about `start_sem` ?

src/ert/scheduler/driver.py
```diff
@@ -117,6 +117,9 @@ async def _execute_with_retry(
                 )
             except FileNotFoundError as e:
                 return (False, str(e))
+            except OSError:
+                await asyncio.sleep(retry_interval)
+                continue
```
jonathan-eq, line 122: Will there be a situation where we get an unrecoverable OSError, and retrying here would not make sense? Should we make it more specific to only too-many-open-files-os-error?

src/ert/scheduler/lsf_driver.py
```diff
@@ -347,86 +347,93 @@ async def submit(
...
-            (Path(runpath) / LSF_INFO_JSON_FILENAME).write_text(
-                json.dumps({"job_id": job_id}), encoding="utf-8"
-            )
-            self._jobs[job_id] = JobData(
-                iens=iens,
-                job_state=QueuedJob(job_state="PEND"),
-                submitted_timestamp=time.time(),
-            )
-            self._iens2jobid[iens] = job_id
+        await asyncio.shield(protected_submit_job())
```
jonathan-eq, line 387: What does the `asyncio.shield` do?

jonathan-eq, line 387: Will this cause hanging somewhere else if we try cancelling? Maybe we can drop the retry or set total_attemps=1 if it cancelled

src/ert/scheduler/scheduler.py
```diff
@@ -276,7 +276,8 @@ async def execute(
                 )
             )
 
-        sem = asyncio.BoundedSemaphore(self._max_running or len(self._jobs))
+        run_sem = asyncio.BoundedSemaphore(self._max_running or len(self._jobs))
+        kill_sem = asyncio.BoundedSemaphore(50)
```
jonathan-eq, line 280: Why is this BoundedSemaphore set to 50?

src/ert/scheduler/driver.py
jonathan-eq: Also, do we have a test for this OSError?

src/ert/scheduler/scheduler.py
larsevj: Just had to put something there, not sure what will be the best option..

src/ert/scheduler/driver.py
larsevj: Have now made it more specific. Will look into making a test for this.

PR #10710, opened 2025-04-29
Merge heatbeat_task and publisher in EE

### LINKED ISSUE #10251: EE heartbeat might not be working as expected

The heartbeat it there to make sure the monitor asks for a new full snapshot from EE in case the connection has slowed down beyond a 5s thresh-hold. The worry here is that if the connection is having issues, we might have lost events. 
This should not be a worry, because our client makes sure that events are delivered by expecting ACK.

The other issue that heartbeats were supposed to fix is that we will notice if a connection is lost. The issue here is that we only evaluate the time since last heartbeat, when we get a new heartbeat. If we lose connection, then get a new heartbeat 30s later; the connection will already have re-established itself.  

We have the heartbeat concept in monitor.py also, but here we use the events themselves as heartbeats, and create empty events in case there are no new events. By doing the same in EE we would get rid of the heartbeat task (if the concept is at all needed anymore). 
@xjules had an idea for this, that we can use ipc connection monitor<-->EE, and tcp connection EE<-->dispatchers, which could speed up things. 

Evaluate whether or not we need this concept

### steps to reproduce

```Python

```

- [ ] python 3.11
- [ ] python 3.12
- [ ] macosx
- [ ] rhel7
- [ ] rhel8
- [ ] local queue
- [ ] lsf queue
- [ ] slurm queue
- [ ] openPBS queue

jonathan-eq:
**Issue**
Resolves #10251 

**Approach**
This commit merges the two tasks, by having publisher directly sending
heatbeat if there is not a new event in the event_to_send queue within
the `HEARTBEAT_TIMEOUT`. This means less messages being sent overall, as
we are now also using events as heartbeats, the same way that is being
done in monitor.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -85,9 +85,10 @@ class EnsembleEvaluator:
         self._dispatchers_empty.set()
 
     async def _publisher(self) -> None:
+        event: Event | _HeartbeatEvent
         while True:
             try:
-                event: Event | _HeartbeatEvent = await asyncio.wait_for(
+                event = await asyncio.wait_for(
                     self._events_to_send.get(), timeout=HEARTBEAT_TIMEOUT
                 )
             except TimeoutError:
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -4,7 +4,6 @@ import asyncio
 import logging
 import traceback
 from collections.abc import Awaitable, Callable, Iterable, Sequence
-from enum import Enum
 from typing import Any, get_args
 
 import zmq.asyncio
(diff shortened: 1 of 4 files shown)


src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -61,7 +63,7 @@ def __init__(self, ensemble: Ensemble, config: EvaluatorServerConfig):
         self._ensemble: Ensemble = ensemble
 
         self._events: asyncio.Queue[Event] = asyncio.Queue()
-        self._events_to_send: asyncio.Queue[Event | HeartbeatEvent] = asyncio.Queue()
+        self._events_to_send: asyncio.Queue[Event] = asyncio.Queue()
```
xjules, line 66: This does not match now since down the line you expect:
```python
event: Event | _HeartbeatEvent = await asyncio.wait_for(
                    self._events_to_send.get(), timeout=HEARTBEAT_TIMEOUT
                )
```

xjules:
Based on the discussion yesterday, is this to be closed then?

PR #10719, opened 2025-04-29
Move export everest csv logic to storage

### LINKED ISSUE #10646: Make everest export csv with all results when experiment is completed
Currently we only export when asked to do so by Everviz through the EverestDataAPI. It would be ideal to have this logic inside Everest Storage, suggested approach:

1. Move `everest_csv` and `export_dataframes` to `EverestStorage` and redirect EverestDataAPI to get it through storage
2. Do one invocation at the end of each experiment so that this csv is always dumped.

Consider a flag for including/excluding summary data (only the last timestep if in summary, creating a date column will introduce a lot of redundancy), or whether it should always be included. If included, put the simulator result columns to the right. Order should be: `[index (batch id , realization, simulation, perturbation etc etc), controls, objectives, constraints, ... simulator results]`.

StephanDeHoop:
**Issue**
(partially) Resolves #10646

**Approach**
Move the logic of exporting the `DataFrames` of all batches to `EverestStorage` and invoke one export at the end of an Everest simulation. Also, added an exception for discrete optimization methods that contain no perturbations in the `export_dataframes` method. Moved the test from api to storage as well. 

**TODO** (in other PR): Allow the option to include summary data in the export.

      --exec 'just rapid-tests'`)

## When applicable


src/ert/run_models/everest_run_model.py
```diff
@@ -342,6 +342,8 @@ def run_experiment(
 
         # Store some final results.
         self.ever_storage.on_optimization_finished()
+        if optimizer_exit_code > OptimizerExitCode.TOO_FEW_REALIZATIONS:
+            _ = self.ever_storage.everest_csv
```
StephanDeHoop, line 346: Only exporting when at least a few realizations are run since `polars` was throwing an exception in `export_dataframes` when the result are empty. Or is it better to catch that exception (now that exception still gets thrown when someone calls `EverestDataApi.everest_csv` (or `export_dataframes` directly).

StephanDeHoop:
Currently the `export_dataframes` will fail for discrete optimizers (see failing everest test) since `perturbation_dfs_to_join` will be empty because discrete optimizer doesn't have pertubations only function evaluations. Will chat with Yngve how to resolve.

src/everest/everest_storage.py
```diff
@@ -1050,3 +1050,188 @@ def _parse_merit_line(merit_values_string: str) -> tuple[int, float] | None:
...
+                    "batch_id",
+                    "realization",
+                    "perturbation",
+                    "simulation_id",
+                ],
+            ),
+            _sort_df(batch_df, ["batch_id", "total_objective_value"]),
+        )
+
+    @property
```
yngve-sk, line 1231: Would remove the `@property` decorator here, it only really needs to be there in the data API for everviz compatibility.

For readability (it is not that obvious that this function also returns the path), would return a `Path` instead of str, if you do 
```
        full_path = self._output_dir / "experiment_results.csv"
```
it will be a `Path` since `self._output_dir` is a Path

yngve-sk:
> Currently the `export_dataframes` will fail for discrete optimizers (see failing everest test) since `perturbation_dfs_to_join` will be empty because discrete optimizer doesn't have pertubations only function evaluations. Will chat with Yngve how to resolve.

Would just add some logic to build the dataframe without the perturbations (as a commit inbetween your current 2 commits). A batch does not necessarily have always have perturbations, so in theory it should already be skipping over perturbations (for some results). I'm not extremely familiar with the discrete optimizer, so would double check with @verveerpj  that this makes sense.

verveerpj:
> > Currently the `export_dataframes` will fail for discrete optimizers (see failing everest test) since `perturbation_dfs_to_join` will be empty because discrete optimizer doesn't have pertubations only function evaluations. Will chat with Yngve how to resolve.
> 
> Would just add some logic to build the dataframe without the perturbations (as a commit inbetween your current 2 commits). A batch does not necessarily have always have perturbations, so in theory it should already be skipping over perturbations (for some results). I'm not extremely familiar with the discrete optimizer, so would double check with @verveerpj that this makes sense.

Batches without perturbations are always possible, gradient-based algorithms may also request just functions evaluations, in which case there will be no perturbations. That was taken into account I assume, but here the issue is likely that there are no perturbations at all because discrete (and continuous non-gradient based) optimizers will never request them.

StephanDeHoop:
> > > Currently the `export_dataframes` will fail for discrete optimizers (see failing everest test) since `perturbation_dfs_to_join` will be empty because discrete optimizer doesn't have pertubations only function evaluations. Will chat with Yngve how to resolve.
> > 
> > 
> > Would just add some logic to build the dataframe without the perturbations (as a commit inbetween your current 2 commits). A batch does not necessarily have always have perturbations, so in theory it should already be skipping over perturbations (for some results). I'm not extremely familiar with the discrete optimizer, so would double check with @verveerpj that this makes sense.
> 
> Batches without perturbations are always possible, gradient-based algorithms may also request just functions evaluations, in which case there will be no perturbations. That was taken into account I assume, but here the issue is likely that there are no perturbations at all because discrete (and continuous non-gradient based) optimizers will never request them.

Sorry, I should have striked-through this comment as I already fixed it.

src/everest/everest_storage.py
```diff
@@ -1050,3 +1050,188 @@ def _parse_merit_line(merit_values_string: str) -> tuple[int, float] | None:
...
+
+        realization_dfs_to_concat = _join_by_batch(
+            realization_dfs_to_join, on=["batch_id", "realization", "simulation_id"]
+        )
+        realization_df = pl.concat(realization_dfs_to_concat, how="diagonal")
+
+        perturbation_dfs_to_concat = _join_by_batch(
+            perturbation_dfs_to_join, on=["batch_id", "realization", "perturbation"]
+        )
+        if perturbation_dfs_to_concat:
```
StephanDeHoop, line 1177: @yngve-sk see my solution here, my reasoning for not making another commit: the test fails without this change so I when I merge this should be in the same commit as moving the logic in my opinion. I should have probably separated it to make it easier to review for you though, my bad!

src/ert/run_models/everest_run_model.py
```diff
@@ -340,6 +340,12 @@ def run_experiment(
 
         # Store some final results.
         self.ever_storage.on_optimization_finished()
+        if (
+            optimizer_exit_code is not OptimizerExitCode.UNKNOWN
+            and optimizer_exit_code is not OptimizerExitCode.TOO_FEW_REALIZATIONS
+            and optimizer_exit_code is not OptimizerExitCode.USER_ABORT
+        ):
+            _ = self.ever_storage.export_everest_opt_results_to_csv
```
yngve-sk, line 348: missing ()

yngve-sk (approved):
LGTM!

Issue #10354, opened 2025-03-18
Consider merging the dispatcher communication responsibility of EE into scheduler


jonathan-eq:
There is now a lot of cross component communication between scheduler and EE as the responsibility of communicating with the running job is split between them. This should be refactored so the responsibility is contained in one component, and we won't have to forward events back and forth through separate queues between EE and scheduler. <img width="923" alt="Image" src="https://github.com/user-attachments/assets/37a5c0c0-a24d-464c-8b51-0213fbf5d434" />

I think moving the zmq router from EE into scheduler would be a good idea, so all events will be collected in scheduler; and we can forward them to EE for merging into the snapshot and emitting the changes to GUI. 

<img width="874" alt="Image" src="https://github.com/user-attachments/assets/e449c2b2-80aa-4320-9119-6ffe8a1dc1a6" />

jonathan-eq:
This might be superseded by the experiment server

PR #10726, opened 2025-04-30
Remove monitor

jonathan-eq:
**Issue**
Resolves #my_issue

**Approach**

      --exec 'just rapid-tests'`)

## When applicable


src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -76,34 +91,91 @@ def __init__(self, ensemble: Ensemble, config: EvaluatorServerConfig):
...
-                    await self._router_socket.send_multipart(
-                        [identity, b"", event_to_json(event).encode("utf-8")]
+            try:
+                event = await asyncio.wait_for(
+                    self._events_to_send.get(), timeout=heartbeat_interval_
+                )
+            except TimeoutError:
+                if closetracker_received:  # THIS SHOULD NOT BE NEEDED ANYMORE
+                    logger.error("Evaluator did not send the TERMINATED event!")
+                    break
```
jonathan-eq, line 131: Do we need this robustness? It was needed before as this part was not internally in EE, but now that it is - we might be able to solely rely on EETerminated.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -76,34 +91,91 @@ def __init__(self, ensemble: Ensemble, config: EvaluatorServerConfig):
...
+                self._events_to_send.task_done()
+                continue
+
+            if isinstance(event, EETerminated):
+                logger.debug("EE inner_publisher received EETerminated. Exiting...")
+                self._events_to_send.task_done()
+                break
+
+            if isinstance(event, ForwardModelStepChecksum):
+                self.forward_event(event)
```
jonathan-eq, line 147: Ideally, we would not forward ForwardModelStepChecksum to BRM, but a test went haywire without it. It is probably possible to mock the receiving part in scheduler instead, but that is for a future PR.

jonathan-eq: I did it :)

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -193,7 +255,6 @@ async def _stopped_handler(self, events: Sequence[EnsembleSucceeded]) -> None:
     async def _cancelled_handler(self, events: Sequence[EnsembleCancelled]) -> None:
         if self.ensemble.status != ENSEMBLE_STATE_FAILED:
             await self._append_message(self.ensemble.update_snapshot(events))
-            self.stop()
```
jonathan-eq, line 196: This is now done when the EnsembleCancelled event is merged into the main snapshot.

src/ert/run_models/base_run_model.py
```diff
@@ -681,6 +626,9 @@ async def run_ensemble_evaluator_async(
 
         return evaluator_task.result()
 
+    async def monitor_evaluator(self, evaluator: EnsembleEvaluator) -> bool:
+        return await evaluator._monitoring_result
```
jonathan-eq, line 630: This is here for unit tests.

src/ert/scheduler/scheduler.py
```diff
@@ -183,9 +183,12 @@ async def _checksum_consumer(self) -> None:
         while True:
             event = await self._manifest_queue.get()
             if type(event) is ForwardModelStepChecksum:
-                self.checksum.update(event.checksums)
+                await self._update_checksum(event.checksums)
             self._manifest_queue.task_done()
 
+    async def _update_checksum(self, checksum: dict[str, dict[str, Any]]) -> None:
+        self.checksum.update(checksum)
```
jonathan-eq, line 191: This is not really async, but had to be for a unit test.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -214,32 +274,16 @@ async def _failed_handler(self, events: Sequence[EnsembleFailed]) -> None:
...
+    async def forward_checksum(self, event: ForwardModelStepChecksum) -> None:
+        await self._manifest_queue.put(event)
+
+    def _put_full_snapshot_in_queue(self) -> None:
+        current_snapshot_dict = self._ensemble.snapshot.to_dict()
+        event = EESnapshot(
+            snapshot=current_snapshot_dict,
+            ensemble=self.ensemble.id_,
+        )
+        self._events_to_send.put_nowait(event)
```
jonathan-eq, line 289: This is because we previously send a fullsnapshot on client `CONNECT` message. Now we just forward it to BRM on startup.

jonathan-eq:
It doesnt work on py311 😭

src/_ert/events.py
```diff
@@ -216,34 +195,23 @@ class EEUserDone(BaseEvent):
...
 EnsembleEvent = EnsembleStarted | EnsembleSucceeded | EnsembleFailed | EnsembleCancelled
 
-EEEvent = EESnapshot | EESnapshotUpdate | EETerminated | EEUserCancel | EEUserDone
+EEEvent = EESnapshot | EESnapshotUpdate
 
-Event = FMEvent | ForwardModelStepChecksum | RealizationEvent | EEEvent | EnsembleEvent
+SnapshotInputEvent = RealizationEvent | EnsembleEvent | FMEvent
 
-DispatchEvent = FMEvent | ForwardModelStepChecksum | RealizationEvent | EnsembleEvent
+DispatcherEvent = FMEvent | ForwardModelStepChecksum
```
xjules, line 202: Why renaming the `DispatchEvent`?

jonathan-eq: It is not a dispatch event, it is an event from dispatcher - DispatcherEvent.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -49,61 +43,131 @@
...
+                    self._monitoring_result.set_exception(
+                        UserCancelled("Experiment cancelled by user during evaluation")
                     )
-            self._events_to_send.task_done()
+
+        if not self._monitoring_result.done():
+            self._monitoring_result.set_result(True)
+
+    async def _monitor_end_queue(self) -> None:
+        while self._end_queue.empty():
```
xjules: This looks like something that could be replaced by `asyncio.Lock() / asyncio.Condition`
https://docs.python.org/3/library/asyncio-sync.html#asyncio.Condition

jonathan-eq: I am not sure, the queue is passed from far back and through BRM, so it is a synced SimpleQueue...

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -76,34 +86,85 @@ def __init__(self, ensemble: Ensemble, config: EvaluatorServerConfig):
...
+            try:
+                event = await asyncio.wait_for(
+                    self._events_to_send.get(), timeout=heartbeat_interval_
+                )
+            except TimeoutError:
+                if closetracker_received:  # THIS SHOULD NOT BE NEEDED ANYMORE
+                    logger.error("Evaluator did not send the TERMINATED event!")
+                    break
+                await asyncio.sleep(0.5)
+                event = None
```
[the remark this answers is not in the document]
jonathan-eq: I moved it to a new task :)

src/ert/scheduler/scheduler.py
```diff
@@ -62,8 +63,8 @@ def __init__(
         self,
         driver: Driver,
         realizations: Sequence[Realization] | None = None,
-        manifest_queue: asyncio.Queue[Event] | None = None,
-        ensemble_evaluator_queue: asyncio.Queue[Event] | None = None,
+        manifest_queue: asyncio.Queue[ForwardModelStepChecksum] | None = None,
```
xjules, line 66: I see that you made a lot of updates to event names.

jonathan-eq:
> I think this requires (also before but now it is definitely a good time) a diagram with different queues; mainly where are they instantiated and how the messages propagate.

I can rename the queues `events_to_scheduler` and `events_to_ee` for documenting this, but I think drawing (mermaid) diagrams would be part of a bigger documentation issue. 
There is an issue for splitting up some of the EE responsibilities (some into BRM, some into Scheduler). I think the diagram should be done in that PR. https://github.com/equinor/ert/issues/10354

xjules:
Blocked by https://github.com/equinor/ert/issues/10855

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -49,61 +43,129 @@
...
+            except Exception as e:
+                logger.exception(f"unexpected error: {e}")
+                # We really don't know what happened...  shut down
+                # the thread and get out of here. The monitor has
+                # been stopped by the ctx-mgr
+                self._monitoring_result.set_result(False)
+                return
+
+    async def _monitor_end_queue(self) -> None:
+        while True:
```
yngve-sk, line 165: Would it be possible to only hold `_end_queue` (consider also renaming, not sure what the current name is trying to communicate) in evaluator only, and not in the `run_model`? Its only use outside of _ is this:
```
    def cancel(self) -> None:
        self._end_queue.put("END")
```
which could be pointed to the evaluator.
The pattern of monitoring the queue of its "parent" seems a bit off

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -214,32 +278,16 @@ async def _failed_handler(self, events: Sequence[EnsembleFailed]) -> None:
...
-        else:
-            event = event_from_json(frame.decode("utf-8"))
-            if type(event) is EEUserCancel:
-                logger.debug("Client asked to cancel.")
-                await self._signal_cancel()
-            elif type(event) is EEUserDone:
-                logger.debug("Client signalled done.")
-                self.stop()
+    async def forward_checksum(self, event: ForwardModelStepChecksum) -> None:
+        await self._manifest_queue.put(event)
```
yngve-sk: Why move this up? 2 lines were removed from it, diff would make it clearer if it kept its original position

src/ert/ensemble_evaluator/evaluator.py
yngve-sk: Do you know why we need to send this for things to work? I see if it is removed, the GUI just hangs, but in theory, it should be OK and still be able to propagate stuff?

Nitpick: 1 usage utility fn, consider moving logic to __init__ 

Also consider inlining the vars
```
self._events_to_send.put_nowait(
    EESnapshot(
        snapshot=self._ensemble.snapshot.to_dict(),
        ensemble=self.ensemble.id_,
    )
)
```

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -397,9 +402,9 @@ def update_from_event(
         elif e_type in get_args(EnsembleEvent):
             event = cast(EnsembleEvent, event)
             self._ensemble_state = _ENSEMBLE_TYPE_EVENT_TO_STATUS[type(event)]
-        elif type(event) is EESnapshotUpdate:
+        elif isinstance(event, EESnapshotUpdate):
```
yngve-sk: Is this change necessary/part of the monitor removal?

src/ert/run_models/run_model.py
```diff
@@ -633,6 +579,9 @@ async def run_ensemble_evaluator_async(
 
         return evaluator_task.result()
 
+    async def monitor_evaluator(self, evaluator: EnsembleEvaluator) -> bool:
+        return await evaluator._monitoring_result
```
yngve-sk: Is this just to wait for the monitor to finish and inspect if it finished successfully? Would consider rename

src/ert/scheduler/scheduler.py
yngve-sk: Would it be possible (and not too extensive) to just rewrite the test to not require this? In general we should not have code that is just for tests

jonathan-eq: Yes, I mocked the outer function, and it still works. Great catch! 💯

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -76,34 +86,86 @@ def __init__(self, ensemble: Ensemble, config: EvaluatorServerConfig) -> None:
...
-                    await self._router_socket.send_multipart(
-                        [identity, b"", event.value]
-                    )
-                else:
-                    await self._router_socket.send_multipart(
-                        [identity, b"", event_to_json(event).encode("utf-8")]
-                    )
-            self._events_to_send.task_done()
+            try:
+                event = await asyncio.wait_for(
```
yngve-sk, line 110: Did something change that made it necessary to wrap this in `wait_for`? See it was the same before but without that

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -76,34 +86,86 @@ def __init__(self, ensemble: Ensemble, config: EvaluatorServerConfig) -> None:
...
-        while True:
-            if self._clients_connected:
-                await self._events_to_send.put(HeartbeatEvent.event)
-                await asyncio.sleep(HEARTBEAT_TIMEOUT)
-            else:
-                await asyncio.sleep(0.1)
+        self._publisher_receiving_timeout: float = 60.0
+        self._monitoring_result: asyncio.Future[bool] = asyncio.Future()
+
+    def forward_event(self, event: EEEvent) -> None:
```
yngve-sk: Suggest inlining / moving into _publisher due to 1 usage

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: The initial snapshot is currently required by the GUI, otherwise it won't accept the SnapshotUpdateEvents later on. The ensemble creates the initial snapshot, propagates it with RealizationSnapshots for the realizations that will be run, and sets all realization states to `REALIZATION_STATE_WAITING` and `ENSEMBLE_STATE_UNKNOWN`. I guess the GUI might specify which realizations it wants to run, but the actual realization that will be run is decided and confirmed by the ensemble.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -76,34 +86,86 @@ def __init__(self, ensemble: Ensemble, config: EvaluatorServerConfig) -> None:
...
+                        )
+
+                        self._monitoring_result.set_exception(
+                            UserCancelled(
+                                "Experiment cancelled by user during evaluation"
+                            )
+                        )
+
+            except TimeoutError:
+                if closetracker_received:  # THIS SHOULD NOT BE NEEDED ANYMORE
```
yngve-sk, line 153: Either create an issue to remove it (just before merging), or try to remove it. Do you have some "proof"/reason why we should not need it? Would we still need to handle the `TimeoutError`?

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -76,34 +86,86 @@ def __init__(self, ensemble: Ensemble, config: EvaluatorServerConfig) -> None:
...
+                    EESnapshotUpdate,
+                }:
+                    self.forward_event(event)
+                    self._events_to_send.task_done()
+                    if event.snapshot.get(ids.STATUS) in {
+                        ENSEMBLE_STATE_STOPPED,
+                        ENSEMBLE_STATE_FAILED,
+                    }:
+                        logger.debug("observed evaluation stopped event, signal done")
+                        await self._events_to_send.put(EventSentinel())
```
yngve-sk, line 138: Is this necessary? Is it handled outside of the evaluator or just for that edge case that might be removed?

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -76,34 +86,86 @@ def __init__(self, ensemble: Ensemble, config: EvaluatorServerConfig) -> None:
...
+                return
+
+    async def _monitor_end_queue(self) -> None:
+        while True:
+            if not self._end_queue.empty():
+                logger.debug("Run model cancelled - during evaluation")
+                self._end_queue.get()
+                await self._signal_cancel()
+                logger.debug("Run model cancelled - during evaluation - cancel sent")
+            await asyncio.sleep(0.5)
```
yngve-sk: Was this sleep there before?

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -443,11 +482,11 @@ async def run_and_get_successful_realizations(self) -> list[int]:
             await self._monitor_and_handle_tasks()
         finally:
             self._server_done.set()
-            self._clients_empty.set()
             self._dispatchers_empty.set()
             for task in self._ee_tasks:
                 if not task.done():
                     task.cancel()
+                    await asyncio.sleep(0)
```
yngve-sk, line 490: Do we need this?

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: It is also used internally in the RunModel in the cases where the user wants to cancel in either pre-evaluation or post-evaluation. I wanted to change it to an asyncio.Event, and just call `.wait()`, but it is created outside of an async context. Maybe I can try with a synced event instead. The only worry I have is that the user might be able to press terminate experiment twice, and having two cancellations is possible in the queue, but not with the event.

jonathan-eq: > Would it be possible to only hold `_end_queue` (consider also renaming, not sure what the current name is trying to communicate) in evaluator only, and not in the `run_model`? Its only use outside of _ is this:
> 
> ```
>     def cancel(self) -> None:
>         self._end_queue.put("END")
> ```
> 
> which could be pointed to the evaluator. The pattern of monitoring the queue of its "parent" seems a bit off

Yes, the pattern is like a DIY implementation of signaling the child. The best would be to call task.cancel, but then we dont cover the cancellation pre-/post-evaluation.

src/ert/ensemble_evaluator/snapshot.py
jonathan-eq: No, I saw it as a general improvement, but maybe it was better beforehand.

tests/ert/unit_tests/ensemble_evaluator/test_scheduler.py
```diff
@@ -66,14 +49,21 @@ def create_manifest_file():
         )
         # actual_md5sum = hashlib.md5(file_path.read_bytes()).hexdigest()
```
yngve-sk, line 51: unrelated to PR but stray comment, feel free to remove in separate commit

tests/ert/unit_tests/ensemble_evaluator/test_monitor.py
```diff
@@ -1,244 +0,0 @@
-import asyncio
```
yngve-sk, line 1: Are some of these test cases still applicable to `Scheduler`, just wondering if we are losing some coverage here

tests/ert/unit_tests/ensemble_evaluator/test_monitor.py
```diff
@@ -1,244 +0,0 @@
...
-    monkeypatch.setattr(Monitor, "DEFAULT_MAX_RETRIES", 0)
-    monitor = Monitor(ee_config.get_uri())
-    monitor._ack_timeout = 0.1
-    with pytest.raises(ClientConnectionError):
-        async with monitor:
-            pass
-
-
-async def test_immediate_stop(unused_tcp_port):
-    connected = False
```
yngve-sk, line 76: applicable/similarly tested by evaluator tests?

tests/ert/unit_tests/ensemble_evaluator/test_monitor.py
```diff
@@ -1,244 +0,0 @@
...
-        await monitor.signal_done()
-    await websocket_server_task
-    assert connected is False
-
-
-@pytest.mark.integration_test
-async def test_unexpected_close_after_connection_successful(
-    monkeypatch, unused_tcp_port
-):
-    monkeypatch.setattr(Monitor, "DEFAULT_MAX_RETRIES", 0)
```
yngve-sk, line 108: applicable/similarly tested by evaluator tests?

src/ert/ensemble_evaluator/evaluator.py
yngve-sk: Okok, sounds like something for a different issue. I think ideally the GUI shouldn't need to "pre-allocate" a snapshot, and would just expand as it goes wrt the realizations it receives.

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: Yes, otherwise it will not cancel. I did a small POC on this issue, and it seems that if you have a co-routine inside a co-routine, and cancel the outer one, the inner one will refuse to cancel and carry on like normal unless you yield.

src/ert/ensemble_evaluator/snapshot.py
yngve-sk: Was just questioning the "relatedness" to the PR, type(..) is more specific than isinstance, isinstance would also cover subclasses (if we need that)

tests/ert/unit_tests/ensemble_evaluator/test_scheduler.py
jonathan-eq: Good catch, will take a note of it

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: No, but we need to yield otherwise the other tasks will be hanging.

tests/ert/unit_tests/ensemble_evaluator/test_monitor.py
jonathan-eq: I don't think so, but I will double check if any of the tests can be moved 👌

src/ert/ensemble_evaluator/evaluator.py
yngve-sk: After some discussion we found that this should be deferred to be future work. Before merging, could you set up an issue for this? I think it will also involve maybe a small/medium refactor to how the evaluator is initialized.

src/ert/ensemble_evaluator/evaluator.py
yngve-sk: Okok, please ad a short comment to the code, explaining why we have the sleep

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: Yes, before it was a heartbeat_task putting events every other second. This removes that task, and handles it directly instead.

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: It is instead of the heartbeat_task. Created an issue for it https://github.com/equinor/ert/issues/11338

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: This is to finish sending all the events that are remaining in the queue before shutting down. This will discard any messages that were added AFTER the EventSentinel was put in the queue, but send the ones prior to it.

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: No, it is to yield and not block the event loop. I think 0.5s sounds like a reasonable polling period.

src/ert/run_models/run_model.py
jonathan-eq: Yes, maybe something like `get_evaluator_result`? Before, it was `run_monitor` I think.

tests/ert/unit_tests/ensemble_evaluator/test_monitor.py
jonathan-eq: This is testing a concept that it is no longer applicable. This was from when the monitor had to tell the evaluator that it was done, after monitor got an ENSEMBLE_STOPPED, through the `.signal_done()` method. The new approach is having evaluator look for `ENSEMBLE_STOPPED` directly, and shutting itself down.

tests/ert/unit_tests/ensemble_evaluator/test_monitor.py
jonathan-eq: There is no longer any zmq connections as we got rid of that layer, so the unit that this was testing is no longer there.

src/ert/ensemble_evaluator/snapshot.py
jonathan-eq: Ahhh, you are right! I will change it back

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: #11340

src/ert/run_models/run_model.py
yngve-sk: The new name is more readable, but actually on 2nd thought, can't we just remove it and go direct? I see there is 1 usage in the test code but surely that can just be rewritten
```
        if not (await self.wait_for_evaluator_to_finish_and_get_result(evaluator)):
```

becomes 
```
   is_successful = await evaluator._monitoring_result
   if not successful:
```

jonathan-eq: It is a standalone function for testing purposes,  but I will look for another workaround

yngve-sk: I think you could just inline its contents to its 2 callsites? since it wraps only a oneliner

jonathan-eq: Without a wrapper, we would have to start mocking primitive asyncio.Future, which seems to be difficult. I will move the oneliner to evaluator, that way we will have it more testable, and we won't expose the evaluator internals too much to run_model.

yngve-sk (approved):
LGTM! Thanks for the patience wrt review comments

PR #10789, opened 2025-05-06
Remove everest caching code and refactor the evaluator function

### LINKED ISSUE #10734: Remove the caching mechanism from Everest
Currently the Everest run model maintains a caching mechanism to prevent running forward models more than once for a given control vector. In practice this was mainly used in the case of auto-scaling of objectives and constraints, where an extra batch is run before starting the optimization. PR #10730 changes how auto-scaling is done, and removes the need for running the extra batch. Although the caching may be useful in specific cases, it will not be used anymore in the current applications. As it complicates handling of results in storage significantly, it is better to remove it for now until we consolidate result handling.

Blocked by #10730

verveerpj:
**Issue**
Resolves #10734 

**Approach**
There are two commits:
- The first removes the caching code
- The second refactors the evaluator function

I recommend reviewing the separately.

      --exec 'just rapid-tests'`)

## When applicable


src/ert/run_models/everest_run_model.py
```diff
@@ -443,192 +443,73 @@ def _run_forward_model(
...
-        control_values_to_simulate = np.array(
-            [
-                c.control_vector
-                for c in evaluation_infos
-                if c.status == _EvaluationStatus.TO_SIMULATE
-            ],
-            dtype=np.float64,
+        # This needs to happen first: Remove inactive control vectors
+        active_control_vectors, control_values, realizations, perturbations = (
+            self._filter_inactive_realizations(
```
yngve-sk, line 453: Do we have to remove the inactive evaluations on the Everest side now that caching is gone?

verveerpj: This is just a different approach than we had before: We first filter away all inactive realizations. Then we can assume that all control vectors are active, so we can just run them all. Finally, the results for the inactive vectors must have zeros, so we expand the result matrices with zero rows in the correct places.

yngve-sk: Hmm,  I think my comment was then based on a misunderstanding of what `evaluator_context.active` really is, not a mapping from control index to bool whether it is active or not I guess?

verveerpj: Its a mapping from realization index to bool, it probably should be named better...

yngve-sk: model realization index?

src/ert/run_models/everest_run_model.py
```diff
@@ -449,376 +443,253 @@ def _run_forward_model(
...
+        #    realizations for each control vector. A numpy vector with a length
+        #    equal to the number of rows of `control_values`
+        # 2. evaluator_context.perturbations: The indices of the perturbations
+        #    for each control vector. A numpy vector with a length equal to the
+        #    number of rows of `control_values`. If an entry is less than zero,
+        #    the corresponding control vector is not a perturbation. If
+        #    evaluator_context.perturbations is `None`, none of the vectors is a
+        #    perturbation.
+        # 3. evaluator_context.active: A boolean numpy vector of length equal to
+        #    the number of realizations. Each entry indicates if the
```
yngve-sk, line 475: model realizations

src/ert/run_models/everest_run_model.py
```diff
@@ -449,376 +443,253 @@ def _run_forward_model(
...
+        #    corresponding realizations is active. If yes, all control vectors
```
yngve-sk, line 476:
```suggestion
        #    corresponding model realization is active. If yes, all control vectors
```

src/ert/run_models/everest_run_model.py
yngve-sk, line 475:
```suggestion
        #    the number of model realizations. Each entry indicates if the
```

src/ert/run_models/everest_run_model.py
```diff
@@ -449,376 +443,253 @@ def _run_forward_model(
...
+        #    with the corresponding realization should be evaluated, the results
```
yngve-sk, line 477:
```suggestion
        #    pertaining to the given model realization should be evaluated, the results
```

src/ert/run_models/everest_run_model.py
```diff
@@ -449,376 +443,253 @@ def _run_forward_model(
...
+        #  models for active control vectors. `control_values_mask` is needed to
+        #  reconstruct the results at the end by inserting rows of zeros for
+        #  inactive control vectors.
+        (
+            control_values_mask,
+            active_control_vectors,
+            realization_indices,
+            perturbation_indices,
+        ) = self._filter_inactive_realizations(
+            evaluator_context.active,
```
yngve-sk, line 521: Just a question: The purpose of this (at the moment) is still only to filter out evaluations pertaining to deactivated model realizations. What is the benefit of having this passed here, as opposed to just passing the control value mask from ROPT, or even just doing the masking on the ROPT side?

```
control_values_mask = active_realizations[realization_indices]
``` 
is the same as
```
evaluator_context.active[evaluator_context.realizations]
```

yngve-sk: Same question as above, but for inserting the zeros, transitively dependent on Q above

src/ert/run_models/everest_run_model.py
```diff
@@ -449,376 +443,253 @@ def _run_forward_model(
...
-                        "batch": "source_batch_id",
-                        "simulation_id": "source_simulation_id",
-                        "flat_index": "target_evaluation_id",
-                    }
-                )
-                .to_dicts()
-            )
-
-            logger.debug(f"Found {len(cache_hits_dict)} cache_hits: {cache_hits_dict}")
+        if active_control_vectors.shape[0] > 0:
```
yngve-sk, line 527: This line is basically saying "If there are any simulations to run", reads slightly implicit. I understand it after reading the doc and having worked on this, but feel just a bool variable would make it more readable. Assigning `active_control_vectors.shape[0]` to a variable (for example `num_simulations` or something like that, and replacing subsequent calls with `num_simulations` or something like that would clarify this a bit.

src/ert/run_models/everest_run_model.py
```diff
@@ -449,376 +443,253 @@ def _run_forward_model(
...
+        self,
+        objectives: NDArray[np.float64],
+        constraints: NDArray[np.float64] | None,
+        realization_indices: NDArray[np.intc],
+        perturbation_indices: NDArray[np.intc] | None,
+    ) -> None:
+        mask = slice(None) if perturbation_indices is None else perturbation_indices < 0
+        non_perturbed_realizations = realization_indices[mask]
+        if non_perturbed_realizations.size > 0:
+            assert self._transforms is not None
```
yngve-sk, line 677: Could we initialize self._transforms on initialization of `EverestRunModel` to avoid the lateinit and nullcheck?

src/everest/everest_storage.py
```diff
@@ -447,7 +447,7 @@ def _enforce_dtypes(df: pl.DataFrame) -> pl.DataFrame:
             "batch_id": pl.UInt32,
             "perturbation": pl.Int32,
             "realization": pl.UInt32,
-            # -1 is used as a value in simulator cache.
+            # -1 is used as a place-holder value.
```
yngve-sk, line 450: (does not have to be done in this PR) this means we don't need Int32 type for `simulation_id` anymore?

src/ert/run_models/everest_run_model.py
verveerpj: This is mostly to keep `ropt`, and potentially also the evaluator function flexible. `ropt` can in principle even pass the active state of realizations on a per-objective/constraint basis (can happen in some applications), so potentially an evaluator could take advantage of that and skip evaluating some objectives/constraints, i.e partially evaluate a realization. Returning a full matrix could also be beneficial if you would like to evaluate forward models, whether they are needed or not for reporting or so. Everest does not really use that, so it is indeed a bit of a burden.

You could do most of  that at the `ropt` side, except for the `sim_ids`, those are unknown to `ropt`, they are just passed through in the results. `ropt` would not know what fill value to use to expand those.

I may be helpful to add some utility methods to the `EvaluatorContext` object to support this.

src/everest/everest_storage.py
verveerpj: The -1 is still needed for inactive control vectors.

src/ert/run_models/everest_run_model.py
```diff
@@ -455,376 +449,201 @@ def _run_forward_model(
...
 
+    def _calculate_auto_scales(
+        self,
+        objectives: NDArray[np.float64],
+        constraints: NDArray[np.float64] | None,
+        realization_indices: NDArray[np.intc],
+        perturbation_indices: NDArray[np.intc] | None,
+    ) -> None:
+        mask = slice(None) if perturbation_indices is None else perturbation_indices < 0
+        non_perturbed_realizations = realization_indices[mask]
```
yngve-sk, line 630: I think it makes more sense to filter out perturbations before invoking this, making it look something like this:
```
def _calculate_auto_scales(
    self,
    objectives: NDArray[np.float64],
    constraints: NDArray[np.float64] | None,
    non_perturbed_realizations: NDArray[np.intc],
) -> None:
    assert isinstance(self._transforms.objectives, ObjectiveScaler)
    if self._transforms.objectives.has_auto_scale:
        self._transforms.objectives.calculate_auto_scales(
            objectives, non_perturbed_realizations
        )
    if self._transforms.nonlinear_constraints:
        assert isinstance(self._transforms.nonlinear_constraints, ConstraintScaler)
        if self._transforms.nonlinear_constraints.has_auto_scale:
            assert constraints is not None
            self._transforms.nonlinear_constraints.calculate_auto_scales(
                constraints, non_perturbed_realizations
            )
```

and change the callsite to something like this:
```
if self._batch_id == 0 and mask := (any(perturbation_indices < 0) if perturbation_indices is not None else slice(None)):
```

this function really only have to deal with non-perturbations, it is no-op if invoked with only perturbations (assuming this case can happen, batch 0 with only perturbations and no function evaluations?)

src/ert/run_models/everest_run_model.py
```diff
@@ -455,376 +449,201 @@ def _run_forward_model(
...
         )
-
-        cache_hits_dict: list[dict[str, Any]] = []
-        if cache_hits_df is not None and not cache_hits_df.is_empty():
-            perturbations = (
+        realization_indices = evaluator_context.filter_inactive_realizations(
+            evaluator_context.realizations
+        )
+        perturbation_indices = (
+            None
```
yngve-sk, line 527: Any benefit to having it as `None` vs all -1s wrt typing? Ref downstream nullchecking

src/ert/run_models/everest_run_model.py
```diff
@@ -455,376 +449,201 @@ def _run_forward_model(
...
+            if constraints is not None:
+                constraints = evaluator_context.insert_inactive_realizations(
+                    constraints
+                )
+            sim_ids = evaluator_context.insert_inactive_realizations(
+                sim_ids, fill_value=-1
+            )
+        else:
+            # Nothing to do, there may only have been inactive control vectors:
+            num_all_simulations = control_values.shape[0]
```
yngve-sk, line 595: Similar logic to evaluator_context.insert_inactive_realizations but just for all the realizations, but it has to be in this `else` still maybe? Don't know the full context here

src/ert/run_models/everest_run_model.py
```diff
@@ -455,376 +449,201 @@ def _run_forward_model(
...
+        # 1. `evaluator_context.realizations`: The indices of the model
+        #    realizations for each control vector. A numpy vector with a length
+        #    equal to the number of rows of `control_values`
+        # 2. `evaluator_context.perturbations`: The indices of the perturbations
+        #    for each control vector. A numpy vector with a length equal to the
+        #    number of rows of `control_values`. If an entry is less than zero,
+        #    the corresponding control vector is not a perturbation. If
+        #    evaluator_context.perturbations is `None`, none of the vectors is a
+        #    perturbation.
+        # 3. `evaluator_context.active`: A boolean numpy vector of length equal to
```
yngve-sk, line 480: This part is now unused in Everest and lives only in ROPT I think, so no need to document the `.active` part I think

src/ert/run_models/everest_run_model.py
verveerpj: Agreed, need to pass the mask then, but that is fine. Will use this to make a fallback for the case of only perturbations.

verveerpj: > Agreed, need to pass the mask then, but that is fine. Will use this to make a fallback for the case of only perturbations.

Actually, I went the other way, I consolidated the whole auto-scaling into a single function, so the logic is contained. The reason is that I wanted to also get rid of the test for batch 0, because, at least in theory, it is possible that batch 0 does nothing (if the else clause is triggered.) In that case you want to try in the next batch.

src/ert/run_models/everest_run_model.py
verveerpj: The else should cover the case that all input control vectors were inactive, which is unlikely, but theoretically could happen.

src/ert/run_models/everest_run_model.py
yngve-sk: Hmm, behavior change wrt batch looks OK, would it be OK to see if upstream typing can be fixed to reduce casting? The contents of the function read pretty well esp without the casting/masking overhead, enough that I would suggest just inlining it and not have the function. There is also no dependency between computing the constraint scales and objective scales, yet they are in one function, which is one more argument just splitting into 2 fns / inlining it.
```
            scaling_simids = perturbation_indices < 0
            if not np.any(scaling_simids):  # If only perturbations exist, just use all.
                scaling_simids = np.ones(perturbation_indices.shape[0], dtype=np.bool_)

            if self._transforms.objectives.needs_auto_scale_calculation:
                self._transforms.objectives.calculate_auto_scales(
                    objectives[scaling_simids, :], realization_indices[scaling_simids]
                )

            if (
                constraint_scales := self._transforms.nonlinear_constraints is not None
            ) and constraint_scales.needs_auto_scale_calculation:
                self._transforms.nonlinear_constraints.calculate_auto_scales(
                    constraints[scaling_simids, :], realization_indices[scaling_simids]
                )
```

verveerpj: I have split in two functions, and I have changed the typing around the transforms.

yngve-sk (approved):
LGTM

Issue #10794, opened 2025-05-07
Use polars to store GenKw parameters


xjules:
On the way to scalars, we should firstly make use of polars when storing genkw parameter data and thus replacing the current xr.Dataset format.

PR #10803, opened 2025-05-07
Use polars to save and load genkw parameters

xjules:
**Issue**
Resolves https://github.com/equinor/ert/issues/10794

**Approach**
Internal data format for GenKw (scalar) parameters changes from xr.Dataset to polars.Dataframe. There are several advantages, which includes storing hybrid data types, faster i/o operations, data handling and manipulation. Additionally it represents a necessary step to implement scalar parameters.

 Each parameter is stored in a column where realizations are the represented as rows. Column realization then defines realizations for which the parameters are defined. 
```python
df = pl.DataFrame({
    "realization": [0, 1],
    "a": [1.0, 4.0],
    "b": [2.0, 5.0],
    "c": [3.0, 6.0],
})
```
One downside is that the load_parameters in LocalEnsemble will now return pl.Dataframe | xr.Dataset. This needs to remain while fields and surfaces rely on xr.Dataset.
In order to determine realization state, the parquet file needs to be open to retrieve the present realizations.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/gen_kw_config.py
+++ b/src/ert/config/gen_kw_config.py
@@ -12,6 +12,7 @@ from typing import TYPE_CHECKING, Any, Self, cast, overload
 import networkx as nx
 import numpy as np
 import pandas as pd
+import polars as pl
 import xarray as xr
 from scipy.stats import norm
 from typing_extensions import TypedDict
@@ -290,7 +291,7 @@ class GenKwConfig(ParameterConfig):
 
     def sample_or_load(
         self, real_nr: int, random_seed: int, ensemble_size: int
-    ) -> xr.Dataset:
+    ) -> pl.DataFrame:
         keys = [e.name for e in self.transform_functions]
         parameter_value = self._sample_value(
             self.name,
@@ -299,11 +300,15 @@ class GenKwConfig(ParameterConfig):
             real_nr,
         )
 
-        return xr.Dataset(
-            {
-                "values": ("names", parameter_value),
-                "names": keys,
-            }
+        parameter_dict = {
+            parameter.name: parameter_value[idx]
+            for idx, parameter in enumerate(self.transform_functions)
+        }
+        parameter_dict["realization"] = real_nr
+        return pl.DataFrame(
+            parameter_dict,
+            schema={tf.name: pl.Float64 for tf in self.transform_functions}
+            | {"realization": pl.Int64},
         )
 
     def load_parameter_graph(self) -> nx.Graph[int]:
@@ -326,32 +331,19 @@ class GenKwConfig(ParameterConfig):
         real_nr: int,
         ensemble: Ensemble,
     ) -> dict[str, dict[str, float | str]]:
-        array = ensemble.load_parameters(self.name, real_nr, transformed=True)["values"]
-        assert isinstance(array, xr.DataArray)
-        if not array.size == len(self.transform_functions):
+        df = ensemble.load_parameters(self.name, real_nr, transformed=True).drop(
+            "realization"
+        )
+
+        assert isinstance(df, pl.DataFrame)
+        if not df.width == len(self.transform_functions):
             raise ValueError(
                 f"The configuration of GEN_KW parameter {self.name}"
-                f" is of size {len(self.transform_functions)}, expected {array.size}"
+                f" has {len(self.transform_functions)} parameters, but ensemble dataset"
+                f" for realization {real_nr} has {df.width} parameters."
             )
 
-        def parse_value(value: float | int | str) -> float | int | str:
-            if isinstance(value, float | int):
-                return value
-            try:
-                return int(value)
-            except ValueError:
-                try:
-                    return float(value)
-                except ValueError:
-                    return value
-
-        data = dict(
-            zip(
-                array["names"].values.tolist(),
-                [parse_value(i) for i in array.values],
-                strict=False,
-            )
-        )
+        data = df.to_dicts()[0]
 
         log10_data: dict[str, float | str] = {
             tf.name: math.log10(data[tf.name])
@@ -370,18 +362,39 @@ class GenKwConfig(ParameterConfig):
         realization: int,
         data: npt.NDArray[np.float64],
     ) -> None:
-        ds = xr.Dataset(
-            {
-                "values": ("names", data),
-                "names": [e.name for e in self.transform_functions],
-            }
+        parameter_dict = {
+            parameter.name: data[idx]
+            for idx, parameter in enumerate(self.transform_functions)
+        }
+        parameter_dict["realization"] = realization
+        ensemble.save_parameters(
+            self.name,
+            realization=None,
+            dataset=pl.DataFrame(
+                parameter_dict,
+                schema={tf.name: pl.Float64 for tf in self.transform_functions}
+                | {"realization": pl.Int64},
+            ),
         )
-        ensemble.save_parameters(self.name, realization, ds)
 
     def load_parameters(
         self, ensemble: Ensemble, realizations: npt.NDArray[np.int_]
     ) -> npt.NDArray[np.float64]:
-        return ensemble.load_parameters(self.name, realizations)["values"].values.T
+        return (
+            ensemble.load_parameters(self.name, realizations)
+            .drop("realization")
+            .to_numpy()
+            .T.copy()
+        )
+
+    def copy_parameters(
+        self,
+        source_ensemble: Ensemble,
+        target_ensemble: Ensemble,
+        realizations: npt.NDArray[np.int_],
+    ) -> None:
+        df = source_ensemble.load_parameters(self.name, realizations)
+        target_ensemble.save_parameters(self.name, realization=None, dataset=df)
 
     def shouldUseLogScale(self, keyword: str) -> bool:
         for tf in self.transform_functions:
@@ -415,6 +428,15 @@ class GenKwConfig(ParameterConfig):
         for index, tf in enumerate(self.transform_functions):
             array[index] = tf.calc_func(array[index], list(tf.parameter_list.values()))
         return array
+
+    def transform_col(self, param_name: str) -> Callable[[float], float]:
+        tf: TransformFunction | None = None
+        for tf in self.transform_functions:
+            if tf.name == param_name:
+                break
+        assert tf is not None, f"Transform function {param_name} not found"
+        arg = list(tf.parameter_list.values())
+        return lambda x: tf.calc_func(x, arg)
 
     @staticmethod
     def _values_from_file(file_name: str, keys: list[str]) -> npt.NDArray[np.double]:
--- a/src/ert/dark_storage/common.py
+++ b/src/ert/dark_storage/common.py
@@ -1,4 +1,3 @@
-import contextlib
 import logging
 import operator
 import os
@@ -6,10 +5,8 @@ from collections.abc import Callable, Iterator
 from typing import Any
 from uuid import UUID
 
-import numpy as np
 import pandas as pd
 import polars as pl
-import xarray as xr
 from polars.exceptions import ColumnNotFoundError
 
 from ert.config import Field, GenDataConfig, GenKwConfig
@@ -129,41 +126,22 @@ def _extract_parameter_group_and_key(key: str) -> tuple[str, str] | tuple[None, 
 
 def data_for_parameter(ensemble: Ensemble, key: str) -> pd.DataFrame:
     group, _ = _extract_parameter_group_and_key(key)
-    parameters = ensemble.experiment.parameter_configuration
-    if group in parameters and isinstance(gen_kw := parameters[group], GenKwConfig):
-        dataframes: list[pd.DataFrame] = []
+    df = ensemble.load_scalars(group)
+    if df.is_empty():
+        return pd.DataFrame()
 
-        with contextlib.suppress(KeyError):
-            try:
-                da = ensemble.load_parameters(group, transformed=True)["values"]
-            except ValueError as err:
-                print(f"Could not load parameter {group}: {err}")
-                return pd.DataFrame()
-
-            assert isinstance(da, xr.DataArray)
-            da["names"] = np.char.add(f"{gen_kw.name}:", da["names"].astype(np.str_))
-            df = da.to_dataframe().unstack(level="names")
-            df.columns = df.columns.droplevel()
-            for parameter in df.columns:
-                if gen_kw.shouldUseLogScale(parameter.split(":")[1]):
-                    df[f"LOG10_{parameter}"] = np.log10(df[parameter])
-            dataframes.append(df)
-        if not dataframes:
-            return pd.DataFrame()
-
-        dataframe = pd.concat(dataframes, axis=1)
-        dataframe.columns.name = None
-        dataframe.index.name = "Realization"
-
-        data = dataframe.sort_index(axis=1)
-        if data.empty or key not in data:
-            return pd.DataFrame()
-        data = data[key].to_frame().dropna()
-        data.columns = pd.Index([0])
-        try:
-            return data.astype(float)
-        except ValueError:
-            return data
+    dataframe = df.to_pandas().set_index("realization")
+    dataframe.columns.name = None
+    dataframe.index.name = "Realization"
+    data = dataframe.sort_index(axis=1)
+    if data.empty or key not in data:
+        return pd.DataFrame()
+    data = data[key].to_frame().dropna()
+    data.columns = pd.Index([0])
+    try:
+        return data.astype(float)
+    except ValueError:
+        return data
 
 
 def _extract_response_type_and_key(
--- a/src/ert/enkf_main.py
+++ b/src/ert/enkf_main.py
@@ -10,7 +10,7 @@ from typing import TYPE_CHECKING, Any
 
 import orjson
 import pandas as pd
-import xarray as xr
+import polars as pl
 
 from ert.substitutions import Substitutions, substitute_runpath_name
 from ert.utils import log_duration
@@ -158,19 +158,16 @@ def save_design_matrix_to_ensemble(
     design_group_name: str = DESIGN_MATRIX_GROUP,
 ) -> None:
     assert not design_matrix_df.empty
-    for realization_nr in active_realizations:
-        row = design_matrix_df.loc[realization_nr]
-        ds = xr.Dataset(
-            {
-                "values": ("names", list(row.values)),
-                "names": list(row.keys()),
-            }
-        )
-        ensemble.save_parameters(
-            design_group_name,
-            realization_nr,
-            ds,
-        )
+    df = design_matrix_df.copy()
+    df["realization"] = df.index
+    pl_df = pl.from_pandas(df.reset_index(drop=True))
+    if not set(active_realizations) <= set(pl_df["realization"].to_list()):
+        raise KeyError("Active realization mask is not in design matrix!")
+    ensemble.save_parameters(
+        design_group_name,
+        realization=None,
+        dataset=pl_df.filter(pl.col("realization").is_in(list(active_realizations))),
+    )
 
 
 @log_duration(
@@ -199,13 +196,29 @@ def sample_prior(
             f"Sampling parameter {config_node.name} "
             f"for realizations {active_realizations}"
         )
-        for realization_nr in active_realizations:
-            ds = config_node.sample_or_load(
-                realization_nr,
-                random_seed=random_seed,
-                ensemble_size=ensemble.ensemble_size,
-            )
-            ensemble.save_parameters(parameter, realization_nr, ds)
+        if isinstance(config_node, GenKwConfig):
+            datasets = [
+                config_node.sample_or_load(
+                    realization_nr,
+                    random_seed=random_seed,
+                    ensemble_size=ensemble.ensemble_size,
+                )
+                for realization_nr in active_realizations
+            ]
+            if datasets:
+                ensemble.save_parameters(
+                    parameter,
+                    realization=None,
+                    dataset=pl.concat(datasets, how="vertical"),
+                )
+        else:
+            for realization_nr in active_realizations:
+                ds = config_node.sample_or_load(
+                    realization_nr,
+                    random_seed=random_seed,
+                    ensemble_size=ensemble.ensemble_size,
+                )
+                ensemble.save_parameters(parameter, realization_nr, ds)
 
     ensemble.refresh_ensemble_state()
 
--- a/src/ert/storage/local_ensemble.py
+++ b/src/ert/storage/local_ensemble.py
@@ -1,13 +1,12 @@
 from __future__ import annotations
 
 import asyncio
-import contextlib
 import logging
 import os
 import time
 from collections.abc import Iterable
 from datetime import datetime
-from functools import cache, lru_cache
+from functools import cache, cached_property, lru_cache
 from multiprocessing.pool import ThreadPool
 from pathlib import Path
 from typing import TYPE_CHECKING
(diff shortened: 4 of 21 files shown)


src/ert/config/gen_kw_config.py
```diff
@@ -279,13 +280,15 @@ def sample_or_load(
...
-        return xr.Dataset(
-            {
-                "values": ("names", parameter_value),
-                "transformed_values": ("names", self.transform(parameter_value)),
-                "names": keys,
-            }
-        )
+        parameter_dict = {}
+        parameter_dict["realization"] = real_nr
+        trasfmored_values = self.transform(parameter_value)
```
yngve-sk: typo

src/ert/config/gen_kw_config.py
```diff
@@ -279,13 +280,15 @@ def sample_or_load(
...
+        for parameter in self.transform_function_definitions:
```
yngve-sk: Can we loop only over transform_functions and simplify this a bit? Do we need to use both the parsed and unparsed?

src/ert/config/gen_kw_config.py
```diff
@@ -307,12 +310,15 @@ def write_to_runpath(
         real_nr: int,
         ensemble: Ensemble,
     ) -> dict[str, dict[str, float | str]]:
-        array = ensemble.load_parameters(self.name, real_nr)["transformed_values"]
-        assert isinstance(array, xr.DataArray)
-        if not array.size == len(self.transform_functions):
+        df = ensemble.load_parameters_pl(self.name, [real_nr]).select(
```
yngve-sk, line 315: Maybe micro optimization, but would consider the idea of using lazy load before filtering, can do scan -> filter out non "*.transformed" -> load instead of loading before filter. (if this is performance intensive, might be for future uses like fields etc)

src/ert/config/gen_kw_config.py
```diff
@@ -307,12 +310,15 @@ def write_to_runpath(
...
+            pl.col("^.*\\.transformed$")
+        )
+
+        assert isinstance(df, pl.DataFrame)
+        if not df.width == len(self.transform_functions):
             raise ValueError(
                 f"The configuration of GEN_KW parameter {self.name}"
-                f" is of size {len(self.transform_functions)}, expected {array.size}"
+                f" is of size {len(self.transform_functions)}, expected {df.width}"
```
yngve-sk, line 323: `df.width` is expected to be `2*len(self.transforms_functions)` due to `.transformed`, so this error message may need some revision?

src/ert/config/gen_kw_config.py
```diff
@@ -366,7 +369,12 @@ def save_parameters(
     def load_parameters(
         self, ensemble: Ensemble, realizations: npt.NDArray[np.int_]
     ) -> npt.NDArray[np.float64]:
-        return ensemble.load_parameters(self.name, realizations)["values"].values.T
+        return (
+            ensemble.load_parameters_pl(self.name, realizations)
+            .select(~pl.col("^.*\\.transformed$"))
```
yngve-sk: Same comment wrt lazy load before filter

src/ert/enkf_main.py
```diff
@@ -158,20 +158,17 @@ def save_design_matrix_to_ensemble(
...
-        )
-        ensemble.save_parameters(
-            design_group_name,
-            realization_nr,
-            ds,
-        )
+    df = design_matrix_df.copy()
+    for col in df.columns:
+        df[f"{col}.transformed"] = df[col]
+    df["realization"] = df.index
```
yngve-sk, line 164: What does this do? Does it assume some existing order of the rows in the design matrix df? Not super obvious I think

src/ert/enkf_main.py
```diff
@@ -200,13 +197,25 @@ def sample_prior(
...
-        for realization_nr in active_realizations:
-            ds = config_node.sample_or_load(
-                realization_nr,
-                random_seed=random_seed,
-                ensemble_size=ensemble.ensemble_size,
-            )
-            ensemble.save_parameters(parameter, realization_nr, ds)
+        if isinstance(config_node, GenKwConfig):
+            datasets = []
+            for realization_nr in active_realizations:
```
yngve-sk, line 202: Style preference: Would use a comprehension over `active_realizations` to create the `datasets` list

src/ert/storage/local_ensemble.py
```diff
@@ -266,6 +266,21 @@ def get_realization_mask_with_responses(self) -> npt.NDArray[np.bool_]:
             ]
         )
 
+    @lru_cache  # noqa: B019
+    def _scalar_exist(self) -> dict[str, list[int]]:
+        genkwn_mask: dict[str, list[int]] = {}
```
yngve-sk, line 271: does the n at the end of `genkw` in `genkwn_mask` mean something?

src/ert/enkf_main.py
xjules: Yes, design matrix (once created as pd.dataframe) contains realizations (defined by the user) and that is set as index.

src/ert/storage/local_ensemble.py
```diff
@@ -544,6 +568,112 @@ def load_parameters(
...
+            existing_realizations = df.get_column("realization").unique()
+            new_data = dataset.filter(
+                ~pl.col("realization").is_in(existing_realizations)
+            )
+            if new_data.height > 0:
+                df = pl.concat([df, new_data], how="vertical")
+            else:
+                return
+            # df = pl.concat([df, dataset], how="vertical")
+            # TODO bug when sampling as it samples N*N time
```
yngve-sk, line 596: still there?

src/ert/storage/local_ensemble.py
```diff
@@ -266,6 +266,21 @@ def get_realization_mask_with_responses(self) -> npt.NDArray[np.bool_]:
             ]
         )
 
+    @lru_cache  # noqa: B019
+    def _scalar_exist(self) -> dict[str, list[int]]:
```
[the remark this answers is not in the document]
yngve-sk: Would move it into get_ensemble_state if it is the only place it is used, then no need to do caching as caching is same as get_ensemble_state caching

src/ert/storage/local_ensemble.py
```diff
@@ -544,6 +568,112 @@ def load_parameters(
...
+                return
+            # df = pl.concat([df, dataset], how="vertical")
+            # TODO bug when sampling as it samples N*N time
+        except KeyError:
+            df = dataset
+
+        group_path = self.mount_point / f"{_escape_filename(group)}.parquet"
+        self._storage._to_parquet_transaction(group_path, df)
+
+    def load_scalars(
```
yngve-sk, line 603: Only used in `load_all_gen_kw_data` and dark storage? Makes me question why it exists alongside `load_parameters_pl`

src/ert/storage/local_ensemble.py
```diff
@@ -564,6 +566,44 @@ def save_parameters_pl(self, group: str, dataset: pl.DataFrame) -> None:
...
+        dataframes = []
+        gen_kws = [
+            config
+            for config in self.experiment.parameter_configuration.values()
+            if isinstance(config, GenKwConfig)
+        ]
+        if group:
+            gen_kws = [config for config in gen_kws if config.name == group]
+        for key in gen_kws:
+            with contextlib.suppress(KeyError):
```
yngve-sk, line 615: Why the suppress?

yngve-sk:
Would consider doing fields and surfaces to polars first. We anyway flatten it to 1D before it goes into the update, so don't really see the benefit of having it as NDarray other than conceptual (physically the xr ndarray is just a 1d array with implicit 3d indexing encoded into the order of its elements)

xjules:
> Would consider doing fields and surfaces to polars first. We anyway flatten it to 1D before it goes into the update, so don't really see the benefit of having it as NDarray other than conceptual (physically the xr ndarray is just a 1d array with implicit 3d indexing encoded into the order of its elements)

The one smaller issue is that it is actually stored as a 3D array: eg.
```python
prior_result = prior.load_parameters("MY_PARAM", list(range(realizations)))[
                "values"
            ]
            posterior_result = posterior.load_parameters(
                "MY_PARAM", list(range(realizations))
            )["values"]

            # check the shape of internal data used in the update
            assert prior_result.shape == (5, NCOL, NROW, NLAY)
            assert posterior_result.shape == (5, NCOL, NROW, NLAY)
```

src/ert/storage/local_ensemble.py
xjules: yes, the bug emerges actually in the update step, so it is not related to the sampling after all.

src/ert/storage/local_ensemble.py
xjules: no :) typo; will remove it

src/ert/storage/local_ensemble.py
xjules: maybe I can put the code in place...

src/ert/config/gen_kw_config.py
xjules: It's been refactored via the load_parameters(..., trasformed=true)

src/ert/enkf_main.py
```diff
@@ -158,20 +158,17 @@ def save_design_matrix_to_ensemble(
...
-            }
-        )
-        ensemble.save_parameters(
-            design_group_name,
-            realization_nr,
-            ds,
-        )
+    df = design_matrix_df.copy()
+    for col in df.columns:
+        df[f"{col}.transformed"] = df[col]
```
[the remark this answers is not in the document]
xjules: Gone with the wind! :)

src/ert/storage/local_ensemble.py
xjules: I'll try to reproduce on main

src/ert/storage/local_ensemble.py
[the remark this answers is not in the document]
yngve-sk: If you did not already, please set up an issue for this

src/ert/config/gen_kw_config.py
```diff
@@ -299,12 +300,12 @@ def sample_or_load(
...
-                "values": ("names", parameter_value),
-                "names": keys,
-            }
-        )
+        parameter_dict = {
+            parameter.name: parameter_value[idx]
+            for idx, parameter in enumerate(self.transform_functions)
+        }
+        parameter_dict["realization"] = real_nr
+        return pl.DataFrame(parameter_dict)
```
yngve-sk: Nothing you have to fix, but maybe future work / just a question: Do we assume that for the numpy endpoints, the order (for one realization) will correspond to the order set here, i.e., `self.transform_functions`? I see the order is set here.

src/ert/config/gen_kw_config.py
```diff
@@ -326,32 +327,19 @@ def write_to_runpath(
...
+        df = ensemble.load_parameters(self.name, real_nr, transformed=True).drop(
+            "realization"
+        )
+
+        assert isinstance(df, pl.DataFrame)
+        if not df.width == len(self.transform_functions):
             raise ValueError(
                 f"The configuration of GEN_KW parameter {self.name}"
-                f" is of size {len(self.transform_functions)}, expected {array.size}"
+                f" is of size {len(self.transform_functions)}, expected {df.width}"
```
yngve-sk: Nitpick: Error message, we expect df.width to correspond to len(self.transform functions), df.width is not the "expected" value.

```suggestion
                f" has {len(self.transform_functions)} parameters, but ensemble dataset for realization {real_nr} has {df.width} parameters."
```

src/ert/config/gen_kw_config.py
```diff
@@ -326,32 +327,19 @@ def write_to_runpath(
...
-
-        data = dict(
-            zip(
-                array["names"].values.tolist(),
-                [parse_value(i) for i in array.values],
-                strict=False,
-            )
-        )
+        data_dict = df.to_dict()
+        data = {key: value[0] for key, value in data_dict.items()}
```
yngve-sk:
```suggestion
        data = df.to_dicts()[0]
```
there's always only one row so I think this can replace the 2 lines above

src/ert/config/gen_kw_config.py
```diff
@@ -416,6 +419,15 @@ def transform(self, array: npt.ArrayLike) -> npt.NDArray[np.float64]:
...
         return array
 
+    def transform_col(self, param_name: str) -> Callable[[float], float]:
+        tf: TransformFunction | None = None
+        for tf in self.transform_functions:
+            if tf.name == param_name:
+                break
+        assert tf is not None, f"Transform function {param_name} not found"
+        arg = list(tf.parameter_list.values())
+        return lambda x: tf.calc_func(x, arg)
```
yngve-sk, line 439: I think there should be an issue to look into converting the transform funcs into pure polars expressions for "native" performance, could you set one up?

src/ert/dark_storage/common.py
```diff
@@ -196,44 +194,22 @@ def data_for_key(
...
-            return data.astype(float)
-        except ValueError:
-            return data
-
-    return pd.DataFrame()
+    df = ensemble.load_scalars(group)
+    if df.is_empty():
+        return pd.DataFrame()
+
+    dataframe = df.to_pandas().set_index("realization")
```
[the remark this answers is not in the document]
yngve-sk: Could you set up an issue for that if you did not already? It has to be removed from `ert-internal-examples`, `semeio`, and some tests/places in ERT. Having genkw as polars will replace it in principle.

src/ert/enkf_main.py
```diff
@@ -158,19 +158,16 @@ def save_design_matrix_to_ensemble(
...
-        ensemble.save_parameters(
-            design_group_name,
-            realization_nr,
-            ds,
-        )
+    df = design_matrix_df.copy()
+    df["realization"] = df.index
+    pl_df = pl.from_pandas(df.reset_index(drop=True))
+    if not set(active_realizations) <= set(pl_df["realization"].to_list()):
+        raise KeyError("Active realization mask is not in design matrix!")
```
yngve-sk, line 165: Is there a test hitting this case?

src/ert/storage/local_ensemble.py
```diff
@@ -541,22 +562,40 @@ def load_parameters(
...
-
+        if group not in self.experiment.parameter_configuration:
+            raise KeyError(f"{group} is not registered to the experiment.")
+        config = self.experiment.parameter_configuration[group]
+        if isinstance(config, GenKwConfig):
+            if realizations is not None and isinstance(realizations, int):
+                realizations = np.array([realizations])
+            group_path = self.mount_point / f"{_escape_filename(group)}.parquet"
+            if not group_path.exists():
+                raise KeyError(
```
yngve-sk: Covered by some test?

src/ert/storage/local_ensemble.py
```diff
@@ -541,22 +562,40 @@ def load_parameters(
...
+                    f"No {group} dataset in storage for ensemble {self.name}"
+                )
+            df = pl.read_parquet(group_path)
+            if realizations is not None:
+                if isinstance(realizations, int):
+                    realizations = np.array([realizations])
+                df = df.filter(pl.col("realization").is_in(realizations))
+                if df.is_empty():
+                    raise IndexError(
```
yngve-sk, line 590: Covered by some test?

src/ert/storage/local_ensemble.py
```diff
@@ -541,22 +562,40 @@ def load_parameters(
         parameters : Dataset
             Loaded xarray Dataset with parameters.
         """
-
+        if group not in self.experiment.parameter_configuration:
+            raise KeyError(f"{group} is not registered to the experiment.")
```
yngve-sk, line 581: Covered by some test?

src/ert/config/gen_kw_config.py
xjules: yes, that's what we need to go for I think. As we do it anyway when doing the es_update while retrieving parameters as numpy.

src/ert/config/gen_kw_config.py
xjules: have you tested this that it works? I'm always getting an error.
```
KeyError                                  Traceback (most recent call last)
Cell In[13], line 1
----> 1 df.drop("realization").to_dict()[0]

KeyError: 0

```

src/ert/config/gen_kw_config.py
xjules: There will be a todo, after this has been merged.

src/ert/enkf_main.py
xjules: This one should never happen as we do the validation upfront, so basically I can remove it. Maybe?

src/ert/storage/local_ensemble.py
xjules: hm, I thought it was, but apparently it is not there :) will fix it.

src/ert/storage/local_ensemble.py
xjules: That one is covered in test_local_storage.py / load_unknown_parameter

src/ert/storage/local_ensemble.py
xjules: test added

src/ert/storage/local_ensemble.py
xjules: test added

src/ert/enkf_main.py
yngve-sk: If you are confident it does not happen, and is validated beforehand, you should remove it. Where is it validated beforehand?

tests/ert/unit_tests/storage/test_parameter_sample_types.py
```diff
@@ -423,6 +423,41 @@ def test_gen_kw(storage, tmpdir, config_str, expected, extra_files, expectation)
...
+        NUM_REALIZATIONS 2
+        GEN_KW KW_NAME prior.txt
+        """
+        )
+        with open("config.ert", mode="w", encoding="utf-8") as fh:
+            fh.writelines(config)
+        with open("prior.txt", mode="w", encoding="utf-8") as fh:
+            fh.writelines("MY_KEYWORD NORMAL 0 1")
+        with expectation:
+            _, fs = create_runpath(storage, "config.ert", active_reals)
```
yngve-sk: I know this is already done throughout and just a copypaste remnant, but why do we call `ensemble` `fs`? would prefer call it `ensemble`

tests/ert/unit_tests/storage/test_parameter_sample_types.py
```diff
@@ -423,6 +423,41 @@ def test_gen_kw(storage, tmpdir, config_str, expected, extra_files, expectation)
...
+                IndexError, match=r"No matching realizations \[0\] found for KW_NAME"
+            ),
+        ),
+    ],
+)
+def test_gen_kw_missing_in_storage(storage, tmpdir, active_reals, expectation):
+    with tmpdir.as_cwd():
+        config = dedent(
+            """
+        JOBNAME my_name%d
```
yngve-sk: is this part `JOBNAME` needed?

xjules: Right, I thought JOBNAME was mandatory (as also present in other tests), but I guess it is not :)

src/ert/enkf_main.py
xjules: It's done in here:
https://github.com/equinor/ert/blob/87d745b02f39b59897ac75e8ff42e614bfb16859/src/ert/gui/simulation/_design_matrix_panel.py#L83

xjules: Update: it is needed and there is a test for it: tests/ert/unit_tests/test_libres_facade.py::test_save_parameters_to_storage_from_design_dataframe

yngve-sk (approved):
LGTM! Remember to squash/rebase

src/ert/storage/migration/to11.py
```diff
@@ -0,0 +1,74 @@
...
+            parameters_json = json.load(fin)
+
+        for ens in ensembles:
+            with open(ens / "index.json", encoding="utf-8") as f:
+                ens_file = json.load(f)
+                if ens_file["experiment_id"] != experiment_id:
+                    continue
+
+            real_dirs = [*ens.glob("realization-*")]
+            for param in parameters_json.values():
```
yngve-sk: suggest rename `param`->`param_config`, as it is really a config that contains several parameters (transformfndefs)

yngve-sk:
Migration itself looks OK but it would be good to add a unit test that tests `to11.py` with some more params, current `test_storage_migration` has only 1 genkw with 1 actual parameter

src/ert/storage/migration/to11.py
```diff
@@ -0,0 +1,74 @@
...
+                            zip(
+                                array["names"].values.tolist(),
+                                [parse_value(i) for i in array.values],
+                                strict=False,
+                            )
+                        )
+                        data_dict["realization"] = realization
+                        records.append(data_dict)
+                    if records:
+                        df = pl.DataFrame(records)
```
yngve-sk: Would be explicit about the dtypes here to match what we want it to be from outside. You can pass in `schema_overrides` here. On second thought, the datatype should be explicitly set when constructing the dataframes elsewhere in the code too.

src/ert/storage/local_ensemble.py
```diff
@@ -770,6 +838,28 @@ def save_parameters(
             a 1d-vector. When saving multiple realizations, dataset must
             have a 'realizations' dimension.
         """
+        if isinstance(dataset, pl.DataFrame):
+            try:
+                df = self.load_parameters(group)
```
yngve-sk: Why load before saving? To keep it merged I guess? I think this can also be done my lazily, like we don't have to load the entire existing dataset to check which realizations are there. (when only genkw probably no big deal, unless invoked many many times)

src/ert/storage/local_ensemble.py
```diff
@@ -770,6 +838,28 @@ def save_parameters(
...
+                assert isinstance(df, pl.DataFrame)
+                existing_realizations = df.get_column("realization").unique()
+                new_data = dataset.filter(
```
yngve-sk, line 850: This is a kind of sneaky assumption but I guess it is true, we never overwrite any parameters, only save unwritten ones. there should at least be a comment in there stating this assumption. It also makes more of an argument for doing lazy loading here before doing the concat etc.

src/ert/storage/local_ensemble.py
```diff
@@ -844,17 +932,23 @@ def calculate_std_dev_for_parameter_group(
         if parameter_group not in self.experiment.parameter_configuration:
             raise ValueError(f"{parameter_group} is not registered to the experiment.")
 
-        ds = self.load_parameters(parameter_group)
-        return ds.std("realizations")["values"].values
+        data = self.load_parameters(parameter_group)
+        if isinstance(data, pl.DataFrame):
+            return data.std().to_numpy().squeeze()
```
yngve-sk: would opt for .flatten() or .reshape(-1) over this (flatten will copy, reshape will not, guess copy is OK here?)

yngve-sk: This also takes the std of the `realization` column, guess we want to drop that first?

src/ert/config/gen_kw_config.py
yngve-sk: Should set dtypes explicitly here

src/ert/storage/local_ensemble.py
```diff
@@ -539,6 +539,27 @@ def _load_dataset(
             datasets = [self._load_single_dataset(group, i) for i in realizations]
         return xr.combine_nested(datasets, concat_dim="realizations")
 
+    def _load_parameters_lazy(
+        self,
+        group: str,
+    ) -> pl.LazyFrame:
+        """
+        Parameters
```
yngve-sk, line 552: Docstring needs a bit more text. Also the `.load_parameters` docstring needs update. Would be ideal to make the docstrings agnostic of implementation details, don't think we need to document parameters/return types either, as they are given by type annotations.

src/ert/storage/migration/to11.py
xjules: I thought about getting the schema first; ie. now it would be great to have config.schema; but since we are storing the correct data types it might be not necessary.

src/ert/storage/local_ensemble.py
xjules: I've added a comment. Now it is with lazy_load

Issue #10800, opened 2025-05-07
Traceback shown when poly_eval crashes


SAKavli:
### What happened? (You can include a screenshot if it helps explain)

Related to #10711 ?
When I run poly.ert with a 1/0 in poly_eval.py, i get this error message:

![Image](https://github.com/user-attachments/assets/8aecb706-2d7c-4929-a3a8-ef341e36429a)

### What did you expect to happen?

Not to see the whole traceback

### Steps to reproduce

```Python
- Add 1/0 in poly_eval.py
- Run ert gui poly.ert
```

### Environment where bug has been observed

- [ ] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

SAKavli:
I have traced this issue back to this part of the code:

<img width="560" alt="Image" src="https://github.com/user-attachments/assets/a49c1603-b856-4236-9e61-9346e57ec379" />

This behaviour originates from this [commit](https://github.com/equinor/ert/commit/34336f0583742c36c5470dbb43e08a205f2d0b67), perhaps you could help me look at this tomorrow, @frode-aarstad ? 😊

PR #10811, opened 2025-05-08
Replace ert message box with suggestor

SAKavli:
**Issue**
Resolves #my_issue

**Approach**

      --exec 'just rapid-tests'`)

## When applicable


SAKavli:
Is fontsize 12 better:
image
Than 16?
image
Have to scroll sideways to read everything ...

berland:
I think keeping the fontsize similar to Erts other GUI elements is fine.

There is no horizontal scrollbar?

SAKavli:
> I think keeping the fontsize similar to Erts other GUI elements is fine.
> 
> There is no horizontal scrollbar?

I didn't notice the horizontal scrollbar was missing, I scrolled horizontally with the scrollwheel, but perhaps there is a quick for for this.

berland:
I found something similar in:
https://github.com/equinor/ert/blob/main/src/ert/gui/tools/plot/plot_window.py#L74-L95

In that GUI code there is a copy button (assuming clipboard related). That might be needed at some point in this suggestor, as this code should probably also utilize the suggestor (maybe later)

SAKavli:
I tried changing this value without any horizontal scrollbar appearing, quite strange. However, in the screenshots, there is an active issue affecting the error message: #10800 
Perhaps this wide messages is not that common, and could be a separate issue should the need arise?
The width of the window was quite delicately defined in the code, I would like to leave it as is if it's not necessary.
image

SAKavli:
Is the final commit (just code improvement spotted by chance) welcome in this PR?

berland:
> Is the final commit (just code improvement spotted by chance) welcome in this PR?

The important bit is to have it in a separate commit. After merge into main, it will be invisible in the git log whether it was in one or two PR's. 

So then this question is mostly related to the review process. Piggy-backing on an existing PR saves time and clicks , but is vulnerable if the original PR gets delayed for whatever reason. It migth pay off here :)

SAKavli:
Have added scrollbar and copy button. I added the copy button by wrapping it together with the icon in a vertical QT layout, which again is added together with the error message in a horizontal QT layout. Only annoying thing now is that the vertical layout is evenly distributing its items, meaning the copy button is sometimes farther down if the horizontal layout is tall.
Is this feasible given this?
image
image

SAKavli:
Should I add tests for this?

berland:
I think the placement of the copy button was a little bit odd, and it even looks like it is part of the error message on my display (and I had to scroll down to see it). Suggestion: Put the Copy button in the upper right of the red box, and only use the icon, not the text "Copy".

src/ert/gui/suggestor/_suggestor_message.py
```diff
@@ -133,9 +145,9 @@ def _expanded_text(self) -> str:
 
     def _text(self, location: str) -> str:
         return (
-            '<div style="font-size: 16px; line-height: 24px;">'
+            '<div style="font-size: 12px; line-height: 24px;">'
             + self._color_bold(self._header)
-            + self._message
+            + f"<pre>{self._message}</pre>"
```
berland, line 150: This changes the font to Courier for all messages, which is not really necessary for strings emitted by say the config parser.

berland:
There is an unfortunate effect probably from `<pre>` that makes the layout too wide when there are two messages to display:

Before this PR:
image
After:
image
(you have to scroll to the far right to see that there is actually a second message (warning) present)

SAKavli:
image

SAKavli:
Have tidied up by squashing. 
Now the word wrapping and suggestor size is "dynamically" scaling with the CARD_WIDTH.
The size of cards is now constant, no matter the length of the message.
Examples:
CARD_WIDTH = 400 (less than 400 doesn't fit nicely anymore, but why would anyone change this to <400?)
image
image

CARD_WIDTH = 800 (What is set in my branch right now)
image@
image

CARD_WIDTH 1600 (unrealistic as well)
image
image

SAKavli:
Thoughts on this solution? Scaling card size simply by the length of the message context given this method. Min size will be 400 pixels, max size will be 900 pixels.
image
image
Window of 2x 400px error cards (msg length 1)
image
Window of 2x 900px error cards (msg length 1200)
image

SAKavli:
If there are more than 1 card, each card will be 500 px, else 900px
image
image

berland:
With the risk of getting into bike-shedding:
* The font size looks too large after the font change. 16 in Courier looks larger than the rest of text in the Ert application (testing on a RHEL8 machine). Size 12 is more in line with the rest of the application.
* I think I still want to move the copy button to the upper right, and drop the text "Copy". The symbol should be sufficient.

berland:
It feels unexpected that the card width will not increase if I make the window wider.

SAKavli:
> In the interest of minimizing scope, the copy button can be taken out and done later in a separate PR, as it was not part of the original issue.

I've managed to move it like you suggested, do you still want it removed? Also text wrapping now follows window size.
image
image

berland:
Keep the error symbol in upper left?

src/ert/gui/main.py
```diff
@@ -192,6 +192,11 @@ def continue_action() -> None:
             validation_messages.deprecations,
             continue_action,
             plugin_manager.get_help_links() if plugin_manager is not None else {},
+            widget_info="""\
```
berland, line 195: This must be supplied on line 162 as well.

berland:
I have noticed that when you have error lines (with erroneous tracebacks) of 100 realizations, scrolling becomes very laggy. The `QLabel` widget is probably not made for this purpose.  But we can consider to postpone this to a separate issue.

SAKavli:
> I have noticed that when you have error lines (with erroneous tracebacks) of 100 realizations, scrolling becomes very laggy. The `QLabel` widget is probably not made for this purpose. But we can consider to postpone this to a separate issue.

Ajjj, I see what you mean. I don't mind looking into it, but perhaps in another issue as you say. It's going to be refreshing to start a new QT PR after working on this one for a week with 27 comments 😅

SAKavli:
> If you provoke a validation error (add e.g. `FORWARD_MODEL FOOBAR`) then the header text is missing.
image

berland (approved):
I say lets go! Makes the GUI much prettier!

src/ert/gui/suggestor/_suggestor_message.py
eivindjahren: This relates to issue https://github.com/equinor/ert/issues/10892

PR #10834, opened 2025-05-12
Do not ignore .github in .gitignore

berland:
This allows ripgrep and IDEs (at least nvim) to also find files in .github

**Issue**
Resolves nvim not able to find strings in github action files.

**Approach**
:brain:

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,8 @@
 .vscode
 ## Eclipse?
 /.project
+# neovim (reversed rule)
+!.github
 
 # pytest ignores
 .pytest_cache


.gitignore
```diff
@@ -13,6 +13,8 @@
 .vscode
 ## Eclipse?
 /.project
+# neovim (reversed rule)
+!.github
```
SAKavli, line 17: I am using ChatGPT to understand more of this.
I read that the .github is not ignored by default unless specified in a .gitignore file. For example, in a global one. (~/.gitignore_global).
Could it be that this is the case on your computer? This would fix it, but as .github is not ignored in the first place, does it make sense to add it here?

berland: My "personal" problem is that `neovim` by default ignores all directories starting with `.`, because the dot indicates they are hidden. I would argue that Github has named the `.github` folder wrongly, since it is a very important folder in which we have code, and not something random that should be hidden by default.

Then there are many workarounds to tell `neovim` that it should not ignore `.github` folder, but editor-specific exceptions should ideally be kept outside the repo. Then the `.gitignore` file is to add files that should also be ignored by `git`, but many editors will parse it by default, in order to hide even more files than the dot-files and dot-dirs. And then it turns out we can use it for the opposite as well, telling both git and any other applications that `.github` is not a directory that should be ignored in any situation.

SAKavli (approved):
Thanks for the insight! Sounds good! 🧠

PR #10858, opened 2025-05-14
Add special result types for everest objectives and constraints to ert storage

### LINKED ISSUE #10790: Create own response types for Everest objectives / constraints
Everest currently uses the GEN_DATA as response type for constraints | objectives. Instead, constraints and objectives should be mapped to their own response type within ERT. This will also make it possible to tell what kind of "everest response", i.e., objective or constraint an ERT response in a clean way.

Proposed solution: Make a simplified version of GenDataConfig without the report steps logic, inherit from that to two classes: `EverestObjectiveResponseConfig` / `EverestConstraintResponseConfig` (or something like that), and map constraints / objectives to those in `everest_to_ert`

DanSava:
**Issue**
Resolves #10790 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable


tests/everest/test_everest_storage.py
```diff
@@ -57,3 +58,62 @@ def _sort_df(df: pl.DataFrame) -> pl.DataFrame:
...
+        assert set(experiment.response_info.keys()) == responses
+        ens_range = tuple(range(ensemble.ensemble_size))
+
+        if gen_data_only:
+            df_gen_data = ensemble.load_responses("gen_data", ens_range)
+            gendata_only_keys = set(df_gen_data["response_key"].unique())
+            assert gendata_only_keys == gen_data_only
+
+        if constraints:
+            df_constraints = ensemble.load_responses("everest_constraints", ens_range)
```
yngve-sk, line 111: Reads a lot better and explicit than the .read `gen_data`, nice

src/everest/config/forward_model_config.py
```diff
@@ -30,7 +36,7 @@ class ForwardModelStepConfig(BaseModelWithContextSupport):
     results: (
         (
             Annotated[
-                SummaryResults | GenDataResults,
+                SummaryResults | GenDataResults | ObjectiveResults,
```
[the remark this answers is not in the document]
DanSava: Constraints/constraint output files are not generated by the forward-model.

src/ert/config/everest_constraints_config.py
```diff
@@ -0,0 +1,126 @@
...
+    name: str = "everest_constraints"
+    has_finalized_keys: bool = True
+
+    @property
+    def expected_input_files(self) -> list[str]:
+        return self.input_files
+
+    @classmethod
+    def from_config_dict(cls, config_dict: ConfigDict) -> Self:
+        data = config_dict.get("EVEREST_GEN_DATA", [])
```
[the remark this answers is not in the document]
DanSava: Good point :+1:

src/everest/config/forward_model_config.py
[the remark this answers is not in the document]
yngve-sk: output constraints are, they are gendata currently.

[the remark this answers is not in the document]
DanSava: Hmm ok, it is not an issue to just add a placeholder class for Constraints also. I am not sure if it will ever be used but at least it will be there if it will be needed.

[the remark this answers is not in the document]
yngve-sk: You can say in an EverestConfig that the result produced by one forward model will be an output constraint. Ex:

```
...
forward_model:
...
  - job: adv_distance3 --point-file point.json --target-file data/r{{ realization}}/target.json --out A
    results:
      file_name: A
      type: output_constraint
```

So if the output constraint in the everestconfig says that `A < 1000`, some forward model has to produce the number  `A`. Or if we just have an output constraint spec directly:
```
output_constraints:
  - name: A
    lower_bound: 0.1
    scale: 0.1
```

that implicitly tells us to parse it. So adding `OutputConstraintResults` will let us explicitly say that a forward model produces a given constraint, the same way we can say it for objective (even though objectives are "specified" through the objectives section independently). 

Whether to have the option to say explicitly that a forward model produces an objective/constraint is up for discussion though I guess. But it should be the same for both objectives and constraints. I think it is more explicit to say _which_ fm produces it than just saying that _some_ fmstep produces it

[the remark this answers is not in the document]
DanSava:
```
forward_model:
...
  - job: adv_distance3 --point-file point.json --target-file data/r{{ realization}}/target.json --out A
    results:
      file_name: A
      type: output_constraint
```

I am assuming this is something you want to have in the future. Maybe it makes sense, but currently the Everest optimization cannot use output constraints unless they are defined in the config file, like you also pointed out:

```
output_constraints:
  - name: A
    lower_bound: 0.1
    scale: 0.1
``` 

As far as I know, the output constraints are computed by ropt and added to the runpath. They are loaded into ert storage because they are considered results; otherwise, ert will not load them from the runpath.  I am struggling to understand how an explicit forward model step would be associated with an output constraint, but maybe there is something that I am missing at the moment.

src/ert/config/everest_objective_config.py
```diff
@@ -0,0 +1,121 @@
...
+            input_files=input_files,
+        )
+
+    def read_from_file(self, run_path: str, iens: int, iter_: int) -> pl.DataFrame:
+        def _read_file(filename: Path) -> pl.DataFrame:
+            try:
+                data = np.loadtxt(filename, ndmin=1)
+            except ValueError as err:
+                raise InvalidResponseFile(str(err)) from err
+            active_information_file = filename.parent / (filename.name + "_active")
```
yngve-sk, line 63: Do we need this part for everest objectives/constraints?

yngve-sk: Pretty sure it is not needed, so it can be removed from these. 

There is some re-use between these and gendata. No need to resolve it in this PR, but could you create an issue for deduping some of this logic? Likely best to do one superclass which gendata and everest constraint / everest objective inherits from.

https://github.com/equinor/semeio/blob/f3020802cdddcc60100bf35aaf8fb8727037544b/src/semeio/forward_models/rft/gendata_rft.py#L50

src/everest/config/forward_model_config.py
```diff
@@ -8,19 +8,29 @@
...
         default="*",
     )
 
 
-class GenDataResults(BaseModelWithContextSupport):
+class ObjectiveResults(ForwardModelResult):
+    type: Literal["objective"] = "objective"
+
+
+class ConstraintResults(ForwardModelResult):
```
yngve-sk, line 28: `ConstraintResult`

src/everest/simulator/everest_to_ert.py
```diff
@@ -389,21 +389,31 @@ def _extract_seed(ever_config: EverestConfig, ert_config: dict[str, Any]) -> Non
...
-    objectives_names = [objective.name for objective in ever_config.objective_functions]
-    constraint_names = [
-        constraint.name for constraint in ever_config.output_constraints
-    ]
+    everest_objectives = []
+    for objective in ever_config.objective_functions:
+        everest_objectives.append(
+            {"name": objective.name, "input_file": objective.name}
+        )
+    if everest_objectives:
```
yngve-sk, line 397: Suggest doing a validation for constraints / objectives:
If there are forward model results with `type: objective` and no corresponding objectives in the `objective_functions` section, we should raise an error imo. Same for constraints.

We COULD validate the other way (but that would be a breaking change and may be deferred to later)

DanSava: This sounds like a config validation and not something that should be done here. I will add an issue for this.

src/ert/config/everest_constraints_config.py
```diff
@@ -0,0 +1,122 @@
...
+            active_information_file = filename.parent / (filename.name + "_active")
+            if active_information_file.exists():
+                try:
+                    active_list = np.loadtxt(active_information_file)
+                except ValueError as err:
+                    raise InvalidResponseFile(str(err)) from err
+                data[active_list == 0] = np.nan
+            return pl.DataFrame(
+                {
+                    "index": pl.Series(np.arange(len(data)), dtype=pl.UInt16),
```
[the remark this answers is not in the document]
DanSava: Yes we do

src/everest/simulator/everest_to_ert.py
DanSava: https://github.com/equinor/ert/issues/10896

tests/everest/test_math_func.py
```diff
@@ -50,16 +50,6 @@ def test_math_func_advanced(cached_example):
 
     point_names = ["x.0", "x.1", "x.2"]
 
-    with open_storage(config.storage_dir) as storage:
-        experiment = next(s for s in storage._experiments.values())
-        ensemble = next(experiment.ensembles)
-        df = ensemble.load_responses("gen_data", tuple(range(ensemble.ensemble_size)))
```
yngve-sk, line 56: I think this logic should still be possible to keep, just swap out "gen_data" with "everest_objectives"

yngve-sk:
I think the logic in `everest_run_model._gather_simulation_results` should be revised so it does not have to loop over constraint / objective names, as each successful realization will hold them.

src/everest/simulator/everest_to_ert.py
```diff
@@ -389,21 +389,31 @@ def _extract_seed(ever_config: EverestConfig, ert_config: dict[str, Any]) -> Non
...
         for fm in (ever_config.forward_model or [])
         if fm.results is not None and fm.results.type == "gen_data"
     ]
-
-    gen_data = ert_config.get(ErtConfigKeys.GEN_DATA, [])
-    for name in set(objectives_names + constraint_names + gen_data_files):
-        gen_data.append((name, {"RESULT_FILE": name}))
-    ert_config[ErtConfigKeys.GEN_DATA] = gen_data
+    ert_config[ErtConfigKeys.GEN_DATA] = [
+        *ert_config.get(ErtConfigKeys.GEN_DATA, []),
```
yngve-sk, line 414: Is this line necessary?

tests/everest/test_math_func.py
DanSava: I removed it from here because I moved this logic into a separate test, `test_everest_data_stored_in_ert_local_storage`

src/everest/simulator/everest_to_ert.py
yngve-sk: Can you rename the issue to something more specific?

tests/everest/test_everest_storage.py
```diff
@@ -57,3 +58,83 @@ def _sort_df(df: pl.DataFrame) -> pl.DataFrame:
...
+    config = EverestConfig.load_file(Path(config_path) / config_file)
+    with open_storage(config.storage_dir, mode="r") as storage:
+        experiment = next(s for s in storage._experiments.values())
+        ensemble = next(experiment.ensembles)
+        assert set(experiment.response_info.keys()) == responses
+        ens_range = tuple(range(ensemble.ensemble_size))
+
+        if gen_data_only:
+            df_gen_data = ensemble.load_responses("gen_data", ens_range)
+            gendata_only_keys = set(df_gen_data["response_key"].unique())
```
yngve-sk, line 107: Small thing: .unique() not needed if `set(...)` is used, same for other occurrences

tests/everest/test_everest_storage.py
```diff
@@ -57,3 +58,83 @@ def _sort_df(df: pl.DataFrame) -> pl.DataFrame:
...
+        df_objectives = ensemble.load_responses("everest_objectives", ens_range)
+        objective_keys = set(df_objectives["response_key"].unique())
+        assert objective_keys == objectives
+
+        local_storage_params = [
+            {name: param_config.input_keys}
+            for name, param_config in experiment.parameter_configuration.items()
+        ]
+        everest_controls = []
+        for control in config.controls:
```
yngve-sk, line 124: Any special reason for inlining the traversal of the nested/unnested control names? I think it is similar to some of the util functions we have in EverestConfig, like `formatted_control_names`

tests/everest/test_math_func.py
yngve-sk: Oh, then it is OK. Did not catch the move in the diff

src/everest/simulator/everest_to_ert.py
yngve-sk: Ahh right, I see it is there from old behavior, so should keep it (for now).

tests/everest/test_everest_storage.py
DanSava: Absolutely no reason, I just did not look for the utility functionality in EverestConfig. I will try to use that

yngve-sk (approved):
LGTM, remember to squash

PR #10912, opened 2025-05-21
Replace format_running_time with precisedelta

timurivanov21:
Replace format_running_time with humanize.precisedelta


andreas-el (approved):
🏖️

PR #10920, opened 2025-05-22
Fix broken test

frode-aarstad:
**Issue**
Resolves #my_issue

**Approach**

      --exec 'just rapid-tests'`)

## When applicable


tests/ert/unit_tests/config/test_ert_config.py
```diff
@@ -443,9 +443,12 @@ def test_that_creating_ert_config_from_dict_is_same_as_from_file(
 ):
     filename = "config.ert"
     with config_generator(tmp_path_factory, filename) as config_values:
-        assert ErtConfig.from_dict(
+        config_from_dict = ErtConfig.from_dict(
             config_values.to_config_dict("config.ert", os.getcwd())
-        ) == ErtConfig.from_file(filename)
+        )
+        config_form_file = ErtConfig.from_file(filename)
```
berland, line 449: form -> from

tests/ert/unit_tests/config/config_dict_generator.py
```diff
@@ -342,7 +342,7 @@ def to_config_dict(self, config_file, cwd, all_defines=True):
                     name,
                     (
                         fn,
-                        f"EXECUTABLE script/{name}.exe\nMIN_ARG 0\nMAX_ARG 1\n",
+                        open(fn, encoding="utf-8").read(),  # noqa SIM115
```
berland, line 345: You can use `pathlib`'s `read_text()` function to avoid adding the ruff exception.

tests/ert/unit_tests/config/test_ert_config.py
```diff
@@ -443,9 +443,12 @@ def test_that_creating_ert_config_from_dict_is_same_as_from_file(
 ):
     filename = "config.ert"
     with config_generator(tmp_path_factory, filename) as config_values:
-        assert ErtConfig.from_dict(
+        config_from_dict = ErtConfig.from_dict(
```
berland, line 446: Cannot see a code change here?

tests/ert/unit_tests/config/config_dict_generator.py
frode-aarstad: `{name}.exe` was never correct. In the failing instance hypothesis generated two jobs with different name that used the job config. The EXECUTABLE should in that case be the same.

When and why hypothesis decided to generate this lately I do not know

berland:
The config generator adds the line 

`QUEUE_OPTION lsf JOB_SCRIPT AAAA
`

that probably trips the test after 42083ce. Is it an existing bug that was uncovered when job_script was moved, or do we need to also update the config_dict_generator?

frode-aarstad:
I mixed up two bugs. This PR relates to https://github.com/equinor/ert/actions/runs/15157126060/job/42614474983?pr=10902

frode-aarstad:
> I fail to see the link between this diff and the presence of ECLBASE in the error in the linked issue. 🤷🏻

The output in the issue is misleading. Its not ECLBASE that is the problem

frode-aarstad:
> When reproducing this failure, the difference between `from_dict.dict()` and `from_file.dict()` seems only to be the order of keys (`NUM_CPU` is pushed forward a notch), and thus I am not confident that this PR is solving the same thing (although it might accidentally trigger different ordering of keys?).
> 
> If different key ordering is the bug 10932, it probably makes sense to make the test independent of order. Then this particular PR is probably solving something else - we should know what it solves and make an issue for it.

Strange when I try to reproduce I get the following for` "root['installed_forward_model_steps']['AAAA']['executable']"`

image

berland:
Right, I see this difference now. 

The follow-up question is then:
* How do we know that this PR actually solves the problem and not covering it up? Preferably there should be an explanation in the commit message.

frode-aarstad:
> Right, I see this difference now.
> 
> The follow-up question is then:
> 
>     * How do we know that this PR actually solves the problem and not covering it up? Preferably there should be an explanation in the commit message.

It does not really solve any problem. From my understanding its just hypothesis that has created a invalid test configuration (configfile) that a user would not. Hence the fix here is to prevent hypothesis to create this type of test config.

berland:
But even if the configuration is "invalid", why will `from_file` be different from `from_dict`?

frode-aarstad:
> But even if the configuration is "invalid", why will `from_file` be different from `from_dict`?

My code change is to reflect how we write to file here:

https://github.com/equinor/ert/blob/0acdc5a6bb11d08c3546845cc05ab444a40e25ae/tests/ert/unit_tests/config/config_dict_generator.py#L710

berland:
Thanks, then it makes sense. So this is a bug in `config_dict_generator.py::to_config_dict()` solving an inconsistency with `generate_files_and_dict()`. Maybe worth to expand the commit message even further.

berland (approved):

PR #10925, opened 2025-05-22
Dedupe obs&responses join logic in MeasuredData

yngve-sk:
Towards removing libresfacade


src/ert/data/_measured_data.py
```diff
@@ -91,75 +90,42 @@ def _get_data(
         observed standard deviation will be named STD.
         """
 
-        observations_by_type = ensemble.experiment.observations
+        selected_response_types = {
+            rt
```
berland, line 94: Can you spell out rt -> response_type and k -> key? Makes it more readable.

src/ert/data/_measured_data.py
```diff
@@ -83,83 +82,51 @@ def is_empty(self) -> bool:
     @staticmethod
     def _get_data(
         ensemble: Ensemble,
-        observation_keys: list[str],
+        observed_response_keys: list[str],
     ) -> pd.DataFrame:
```
berland, line 86: is pd.DataFrame correct typing, guessing it is a polars DataFrame that is being returned? (or is that somehow the same type?) (if I am correct, the follow-up question is why mypy did not catch this)

berland: Ah, I found the `to_pandas()` there. No worries.

yngve-sk: Yep it is correct, mainly for legacy backward compatibility. This is basically converting the polars to a pandas and renaming some columns to fit what is currently required by mainly Drogon (only current dependent AFAIK). Ref  `        pddf = df.to_pandas()[ ...`

src/ert/storage/local_ensemble.py
```diff
@@ -977,6 +977,16 @@ def get_observations_and_responses(
     ) -> pl.DataFrame:
         """Fetches and aligns selected observations with their
         corresponding simulated responses from an ensemble."""
+        known_observations = self.experiment.observation_keys
+        unknown_observations = [
+            obs for obs in selected_observations if obs not in known_observations
+        ]
+
+        if unknown_observations:
+            raise KeyError(
```
berland: We do not enforce ruff rule EM102 (yet?), but I think that rule makes sense. Check `ruff rule EM102` and do as it suggests.

PR #10930, opened 2025-05-22
Catch all warnings raised when doing ensemble update

### LINKED ISSUE #9590: Warning of ill-conditioned matrix only showed in terminal

berland:
This will at least catch scenarios where Scipy encounters an ill-conditioned matrix. With this commit, that warning will be printed to the terminal as before, but also logged through the logging system

**Issue**
Resolves #9590 

**Approach**
🎣 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/analysis/_es_update.py
+++ b/src/ert/analysis/_es_update.py
@@ -3,11 +3,13 @@ from __future__ import annotations
 import functools
 import logging
 import time
+import warnings
 from collections.abc import Callable, Iterable, Sequence
 from typing import (
     TYPE_CHECKING,
     Generic,
     Self,
+    TextIO,
     TypeVar,
 )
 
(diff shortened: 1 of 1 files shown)


SAKavli:
Did you manage to recreate this @berland ? If so how? I've tried with the folder from Eivinds repo, but maybe I've set something up wrong.

Issue #9059, opened 2024-10-28
Bug/error-prone GUI-dark_storage "*H" requests control flow


yngve-sk:
The way `history` data for gen data / summary is fetched from the GUI is currently error-prone. Control flow goes like this:

User has plotter open and selects a key -> `plot_window.py:PlotWIndow.updatePlot()` is invoked, leading to this part:
```
            if str(key).endswith("H") or "H:" in str(key):
                plot_context.history_data = DataFrame()
            else:
                try:
                    plot_context.history_data = self._api.history_data(
                        key,
                        [e.id for e in plot_context.ensembles()],
                    )
```

Meaning, it will always try to get the history data for the selected key. This leads to one invocation of 
`df = self.data_for_key(ensemble_id, history_key)`
per selected ensemble, leading to this request
```
client.get(
                f"/ensembles/{ensemble.id}/records/{PlotApi.escape(key)}",
                headers={"accept": "application/x-parquet"},
                timeout=self._timeout,
            )
```

finally leading to `dark_storage/common.py:data_for_key`, where it will try to load the response "*H" and just eat the error and return an empty dataframe. 

https://github.com/equinor/ert/blob/9817d757bb1fca85a57bc7e41e82ee77480060c4/src/ert/dark_storage/common.py#L97C5-L97C17

Solution: Make it possible to check if there is history data for a response key in a more straight-forward way, would suggest to just maintain a map/list of response keys with history, that is updated every time the list of responses / active ensembles changes.

PR #10941, opened 2025-05-23
Remove records concept for parameters & responses

yngve-sk:
**Issue**
Resolves [#10684](https://github.com/equinor/ert/issues/10684)
Resolves [#9059](https://github.com/equinor/ert/issues/9059)

**Approach**
Make separate endpoints for params and responses

PS: Will look into adding webviz ert plots into ert plotter to not maintain backward compat. The removals of some endpoints in this PR can be considered blocked by that.


src/ert/dark_storage/endpoints/parameters.py
```diff
@@ -0,0 +1,125 @@
...
+
+def data_for_parameter(ensemble: Ensemble, key: str) -> pd.DataFrame:
+    group, _ = _extract_parameter_group_and_key(key)
+    try:
+        df = ensemble.load_scalars(group)
+    except KeyError:
+        return pd.DataFrame()
+
+    if df.is_empty():
+        return pd.DataFrame()
```
oyvindeide: Can it be empty? It probably could in the past, but not sure about now

src/ert/dark_storage/endpoints/parameters.py
```diff
@@ -0,0 +1,125 @@
...
+    param_group, param_key = key.split(":")
+    return param_group, param_key
+
+
+def data_for_parameter(ensemble: Ensemble, key: str) -> pd.DataFrame:
+    group, _ = _extract_parameter_group_and_key(key)
+    try:
+        df = ensemble.load_scalars(group)
+    except KeyError:
+        return pd.DataFrame()
```
oyvindeide, line 109: Does this make sense? We should probably let it raise and catch it in the end point to give an appropriate response.

src/ert/dark_storage/endpoints/observations.py
```diff
@@ -30,3 +37,86 @@ def get_observations(
...
+                    "errors": _obs_df["errors"].to_list(),
+                    "x_axis": _obs_df["x_axis"].to_list(),
+                }
+            )
+
+    return observations
+
+
+def get_all_observations(experiment: Experiment) -> list[dict[str, Any]]:
+    return _get_observations(experiment)
```
oyvindeide: Suggest just inlining this

src/ert/dark_storage/endpoints/parameters.py
```diff
@@ -0,0 +1,125 @@
...
+    data_2d = da[:, :, z]
+
+    buffer = io.BytesIO()
+    np.save(buffer, data_2d)
+
+    return Response(content=buffer.getvalue(), media_type="application/octet-stream")
+
+
+def _extract_parameter_group_and_key(key: str) -> tuple[str, str] | tuple[None, None]:
+    key = key.removeprefix("LOG10_")
```
oyvindeide, line 95: This would be for a separate issue?

src/ert/dark_storage/endpoints/responses.py
```diff
@@ -0,0 +1,171 @@
...
+    assert response_key is not None
+    assert response_type is not None
+
+    if response_type == "summary":
+        summary_data = ensemble.load_responses(
+            response_key,
+            tuple(ensemble.get_realization_list_with_responses()),
+        )
+        if summary_data.is_empty():
+            return pd.DataFrame()
```
oyvindeide: Should be error?

src/ert/dark_storage/endpoints/responses.py
```diff
@@ -0,0 +1,171 @@
...
+        df = df.set_index(["Date", "Realization"])
+        # This performs the same aggragation by mean of duplicate values
+        # as in ert/analysis/_es_update.py
+        df = df.groupby(["Date", "Realization"]).mean()
+        data = df.unstack(level="Date")
+        data.columns = data.columns.droplevel(0)
+        try:
+            return data.astype(float)
+        except ValueError:
+            return data
```
oyvindeide: Something that can still happen?

src/ert/dark_storage/endpoints/responses.py
```diff
@@ -0,0 +1,171 @@
...
+
+    if response_type == "gen_data":
+        try:
+            # Call below will ValueError if key ends with H,
+            # requested via PlotAPI.history_data
```
oyvindeide: Can it? Not for gen_data?

src/ert/dark_storage/endpoints/responses.py
```diff
@@ -0,0 +1,171 @@
...
+            pivoted = vals.drop("response_key", "report_step").pivot(
+                on="index", values="values"
+            )
+            data = pivoted.to_pandas().set_index("realization")
+            data.columns = data.columns.astype(int)
+            data.columns.name = "axis"
+            try:
+                return data.astype(float)
+            except ValueError:
+                return data
```
oyvindeide: Can this happen?

src/ert/dark_storage/endpoints/parameters.py
yngve-sk: I think we should defer to future issue/PR. It is "blocked" by these tests (will fail when removing the try-catch clause):
`test_plot_api_handles_non_existant_gen_kw`
`test_plot_api_handles_empty_gen_kw`
Not sure if the behavior these tests enforce even makes sense

src/ert/dark_storage/endpoints/parameters.py
yngve-sk: Yes, removing the LOG10 thing and using the metadata transform instead, will likely require a tweak in webviz ert + ert plotter/plotapi), I think this is the issue?: https://github.com/equinor/ert/issues/10664

src/ert/dark_storage/endpoints/responses.py
yngve-sk: Seems it can be removed without causing error, was previously "locked" in because of the plotter trying to get history data all the time, but now it checks beforehand, so removing it.

src/ert/dark_storage/endpoints/responses.py
yngve-sk: Removed try-catch. there is one left that CAN be removed but currently it causes `test_data_for_response_returns_empty_gen_data_config` to fail, so will defer that to future issue.

src/ert/dark_storage/endpoints/responses.py
yngve-sk: Nope, removed^

src/ert/dark_storage/endpoints/parameters.py
oyvindeide: No, asking for a non-existing key should give an error, not an empty response, but fine to leave it for another issue. Can you create it when merging?

src/ert/dark_storage/endpoints/responses.py
yngve-sk: Don't think so, trycatch is removed

src/ert/dark_storage/endpoints/responses.py
```diff
@@ -0,0 +1,157 @@
...
+) -> pd.DataFrame:
+    response_key, response_type = _extract_response_type_and_key(
+        key, ensemble.experiment.response_key_to_response_type
+    )
+
+    if response_key is None:
+        return pd.DataFrame()
+
+    assert response_key is not None
+    assert response_type is not None
```
oyvindeide, line 117: Can they actually be None? Or is the typing wrong?

oyvindeide (approved):
LGTM! One minor comment

src/ert/dark_storage/endpoints/responses.py
yngve-sk: Suggest deferring to separate issue to remove the `None` typing.
Webviz-ert fails when it is removed, giving this:
```
  File "webviz-ert/webviz_ert/data_loader/__init__.py", line 307, in compute_misfit
    response_key, report_step = response_name.split("@", 1)
    ^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: not enough values to unpack (expected 2, got 1)

```
, and these tests fail: 
`test_data_for_response_doesnt_mistake_history_for_response`
`test_plot_api_handles_urlescape`

Screenshot 2025-06-18 at 14 17 15

PR #10959, opened 2025-05-27
Have cli warn for DESIGN_MATRIX with different num_realizations

### LINKED ISSUE #10799: Should warn when DESIGN_MATRIX changes NUM_REALIZATIONS when running CLI

Currently there is no warning when running the cli with DESIGN_MATRIX in the case were NUM_REALIZATIONS are changed due to the design matrix entries.

Expected a warning printed to the console, probably logged as well.

### Steps to reproduce

```Python

```

- [x] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [x] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

jonathan-eq:
**Issue**
Resolves #10799 
**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/cli/main.py
+++ b/src/ert/cli/main.py
@@ -25,7 +25,7 @@ from ert.mode_definitions import (
 from ert.namespace import Namespace
 from ert.plugins import ErtPluginManager
 from ert.run_models.event import StatusEvents
-from ert.run_models.model_factory import create_model
+from ert.run_models.model_factory import _realizations, create_model
 from ert.storage import open_storage
 from ert.storage.local_storage import local_storage_set_ert_config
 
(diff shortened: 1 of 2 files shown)


src/ert/cli/main.py
```diff
@@ -104,7 +104,43 @@ def run_cli(args: Namespace, plugin_manager: ErtPluginManager | None = None) ->
...
+        )
+        config_num_realizations = ert_config.runpath_config.num_realizations
 
+        if (
+            hasattr(args, "realizations")
+            and args.realizations is None
+            and len(dm_num_realizations) != ert_config.runpath_config.num_realizations
+        ):
+            if (
+                len(ert_config.analysis_config.design_matrix.active_realizations)
```
xjules, line 122: ~~This seems to be incorrect. We have discussed that num_realizations should always take the precedents, or?~~ My bad this is correct.

larsevj: But what will you do if num realizations are greater than the entries? I think we agreed on the minimum.

src/ert/cli/main.py
```diff
@@ -104,7 +104,43 @@ def run_cli(args: Namespace, plugin_manager: ErtPluginManager | None = None) ->
...
+                )
+            else:
+                print(
+                    f"NUM_REALIZATIONS ({config_num_realizations}) is less than the "
+                    "number of realizations in DESIGN_MATRIX "
+                    f"({len(dm_num_realizations)}). Using the realizations from "
+                    f"NUM_REALIZATIONS ({config_num_realizations})"
+                )
+                model.active_realizations = _realizations(
+                    args, config_num_realizations
```
larsevj, line 142: Do not believe this will work if you have nonconsecutive realization entries in the design matrix. 
And also this assumes that we have already changed the ensemble size due to the design matrix, so we should instead change it at the same place for both cli and gui if we want the same behaviour for both? 
Although might be tricky since the warning is different in the two cases.

src/ert/cli/main.py
xjules: Yes, it should be minimum. Afterwards we need to check if there are any overlapping realizations if not we do ValidationError.

jonathan-eq:
Closing this as it won't be merged

PR #11000, opened 2025-05-28
save/load params through numpy when updating

yngve-sk:


xjules:
I would do the change (to employ numpy endpoints) only for the tests then. Would skip the copy_unupdated_parameters as is for the moment since going through numpy for "just" copying files might not be very efficient.

PR #11013, opened 2025-05-30
Make implicit masking of some parameter configs more explicit

yngve-sk:
Towards having numpy endpoint not be only silently "update only" by masking out excluded params

### DIFF
--- a/src/ert/config/gen_kw_config.py
+++ b/src/ert/config/gen_kw_config.py
@@ -383,6 +383,11 @@ class GenKwConfig(ParameterConfig):
     ) -> npt.NDArray[np.float64]:
         return ensemble.load_parameters(self.name, realizations)["values"].values.T
 
+    def load_parameters_for_update(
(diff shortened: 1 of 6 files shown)


src/ert/config/gen_kw_config.py
```diff
@@ -383,6 +383,11 @@ def load_parameters(
     ) -> npt.NDArray[np.float64]:
         return ensemble.load_parameters(self.name, realizations)["values"].values.T
 
+    def load_parameters_for_update(
```
xjules, line 386: Not sure how this makes sense at the moment? 
1) There is the `update` flag on each field, surf and genkw parameter and it is not used here.
2) I'd prefer that there would be a local_ensemble.load_parameters_for_update(...) at least it would give more sense to me.

yngve-sk:
Change here will be done / superceded by https://github.com/equinor/ert/pull/11071

Issue #10520, opened 2025-04-01
A single GenKwConfig: make union of all genkw configs


xjules:
Before we implement Config for ScalarParameters as a first step, we should try to combine all GenKw Configs into one, wherein all genkw instances will be unified into a single of GenKW:
```python
class DataSource(StrEnum):
    DESIGN_MATRIX = "design_matrix"
    SAMPLED = "sampled"

class GenKwConfig(ParameterConfig):
    type: Literal["gen_kw"] = "gen_kw"
    distribution: Annotated[
        UnifSettings
        | LogNormalSettings
        | LogUnifSettings
        | DUnifSettings
        | RawSettings
        | ConstSettings
        | NormalSettings
        | TruncNormalSettings
        | ErrfSettings
        | DerrfSettings
        | TriangularSettings,
        Field(discriminator="name"),
    ]
    forward_init: bool = False
    update: bool = True
    group: str = "DEFAULT"
    input_source: DataSource = DataSource.SAMPLED

ensemble.load_parameters("COEEFS") -> dataframe will all params
ensemble.load_parameters("a") -> dataframe will just "a" 
```

The goal is to migrate away from groups and instead we should give more options to individual parameters. In this way we can choose which parameters to sample and which load from the file - one by one.
  
As an inspiration, something similar has been done for GenData:
https://github.com/equinor/ert/pull/8554

xjules:
This needs to include validation that each genkw prior's file is unique.

PR #11016, opened 2025-05-30
Consolidate parameter sampling logic in runmodels

yngve-sk:


src/ert/run_models/base_run_model.py
```diff
@@ -847,6 +853,64 @@ def _evaluate_and_postprocess(
         return num_successful_realizations
 
 
+class HasDesignParameters:
```
xjules, line 856: This class makes the entire design matrix very explicit. This PR
https://github.com/equinor/ert/pull/11071

will make design matrix parameters only a attribute of parameter; ie. it does not need to be explicit anymore.

xjules:
It makes sense, but it might not be necessary in the end due to this
blocked by https://github.com/equinor/ert/issues/10520 
and this
https://github.com/equinor/ert/pull/10793

PR #11037, opened 2025-06-04
Don't update parameters with zero variance during adaptive localization

### LINKED ISSUE #9714: Handling zero standard deviation in adaptive localization for field parameters in ERT
During updates in ERT with adaptive localization for both facies and petrophysical field parameters (such as permeability and porosity), a recurring issue with zero standard deviation values for certain grid cells has been observed. These values usually arise from fixed geological zones with no variance or due to the presence of inactive grid cells at the edge of the defined areas. As a consequence, there may be misleading cross-correlations or divide by zero errors during the update step after the initial forecast. The current extrapolation methods for undefined grid cell values can result in multiple grid cells having the same fixed value across all realizations, thereby leading to a zero std.dev.

The problem is not easily solvable by users as it requires implementing additional algorithms or injecting noise into the values, which is impractical. Users are also unable to discern whether error messages arising from this issue can be safely ignored or entail critical malfunctions.

#### Proposed Solutions
1. **Early Error Reporting and Input Validation:** ERT should conduct early checks for zero std.dev values:
    - If all grid cell values have a std.dev of zero, the field parameters in question should not be included in updates, and users should be notified to remove these parameters since they cannot be updated.
    - If only some of the grid cells exhibit a std.dev of zero, filtering should be applied to update only those cells with a positive std.dev, while maintaining the initial value of cells with a zero std.dev across all realizations.

2. **Adaptive Localization Processing Optimization:**
    - The process should be optimized to skip the processing of grid cell values that cannot be updated due to a zero std.dev. This would not only prevent unnecessary computations but also reduce support queries and confusion arising from erroneous cross-correlations and divide by zero errors during adaptive localization.

3. **Documentation and User Guidance:**
    - Provide clear documentation and feedback mechanisms informing users about the detection of suspicious values (e.g., a zero std.dev) and advising on correct procedures or automated handling by ERT.

#### Screenshot of error messages:
Image

dafeda:
**Issue**
Resolves #9714

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/analysis/_es_update.py
+++ b/src/ert/analysis/_es_update.py
@@ -305,6 +305,31 @@ def analysis_ES(
         param_ensemble_array = source_ensemble.load_parameters_numpy(
             param_group, iens_active_index
         )
+
+        # Calculate variance for each parameter
+        param_variance = np.var(param_ensemble_array, axis=1)
+        # Create mask for non-zero variance parameters
+        non_zero_variance_mask = ~np.isclose(param_variance, 0.0)
+
+        log_msg = (
+            f"Updating {np.sum(non_zero_variance_mask)} parameters "
+            f"{'with' if module.localization else 'without'} "
+            f"adaptive localization."
+        )
+        logger.info(log_msg)
+        progress_callback(AnalysisStatusEvent(msg=log_msg))
+
+        log_msg = f"There are {num_obs} responses and {ensemble_size} realizations."
+        logger.info(log_msg)
+        progress_callback(AnalysisStatusEvent(msg=log_msg))
+
+        log_msg = (
+            f"There are {(~non_zero_variance_mask).sum()} parameters with 0 variance "
+            f"that will not be updated."
+        )
+        logger.info(log_msg)
+        progress_callback(AnalysisStatusEvent(msg=log_msg))
+
         if module.localization:
             with threadpool_limits(limits=OPTIMAL_NUM_THREADS):
                 config_node = source_ensemble.experiment.parameter_configuration[
@@ -325,7 +350,10 @@ def analysis_ES(
                 start = time.time()
                 cross_correlations: list[npt.NDArray[np.float64]] = []
                 for param_batch_idx in batches:
-                    X_local = param_ensemble_array[param_batch_idx, :]
+                    update_idx = param_batch_idx[
+                        non_zero_variance_mask[param_batch_idx]
+                    ]
+                    X_local = param_ensemble_array[update_idx, :]
                     if isinstance(config_node, GenKwConfig):
                         correlation_batch_callback = functools.partial(
                             correlation_callback,
@@ -333,13 +361,13 @@ def analysis_ES(
                         )
                     else:
                         correlation_batch_callback = None
-                    param_ensemble_array[param_batch_idx, :] = (
+                    param_ensemble_array[update_idx, :] = (
                         smoother_adaptive_es.assimilate(
                             X=X_local,
                             Y=S,
                             D=D,
                             # The user is responsible for scaling observation covariance
-                            # (esmda usage):
+                            # (ESMDA usage)
                             alpha=1.0,
                             correlation_threshold=module.correlation_threshold,
                             cov_YY=cov_YY,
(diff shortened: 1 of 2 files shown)


src/ert/analysis/_es_update.py
```diff
@@ -305,6 +305,31 @@ def correlation_callback(
...
+        log_msg = f"There are {num_obs} responses and {ensemble_size} realizations."
+        logger.info(log_msg)
+        progress_callback(AnalysisStatusEvent(msg=log_msg))
+
+        log_msg = (
+            f"There are {(~non_zero_variance_mask).sum()} parameters with 0 variance "
+            f"that will not be updated."
+        )
+        logger.info(log_msg)
+        progress_callback(AnalysisStatusEvent(msg=log_msg))
```
xjules, line 331: Why 3 `AnalysisStatusEvent` instead of one?

src/ert/analysis/_es_update.py
```diff
@@ -305,6 +305,31 @@ def correlation_callback(
         param_ensemble_array = source_ensemble.load_parameters_numpy(
             param_group, iens_active_index
         )
+
+        # Calculate variance for each parameter
+        param_variance = np.var(param_ensemble_array, axis=1)
+        # Create mask for non-zero variance parameters
+        non_zero_variance_mask = ~np.isclose(param_variance, 0.0)
```
xjules, line 312: Do you want to set the tolerance to some non-default values?

dafeda: Not sure. What do you suggest?

src/ert/analysis/_es_update.py
dafeda: Not sure I understand your question. Is it possible to send multiple messages in one `AnalysisStatusEvent`?

xjules: No. one log_msg combining all three statements.

src/ert/analysis/_es_update.py
```diff
@@ -368,9 +396,9 @@ def correlation_callback(
 
         else:
             # In-place multiplication is not yet supported, therefore avoiding @=
-            param_ensemble_array = param_ensemble_array @ T.astype(  # noqa: PLR6104
-                param_ensemble_array.dtype
-            )
+            param_ensemble_array[non_zero_variance_mask] = param_ensemble_array[  # noqa: PLR6104
```
xjules, line 399: so this is the place :grin: 
nice!

tests/ert/ui_tests/cli/test_field_parameter.py
```diff
@@ -476,32 +477,33 @@ def test_field_param_update_using_heat_equation_zero_var_params_and_adaptive_loc
...
+                    "There are 50 parameters with 0 variance that will not be updated."
+                    in caplog.text
+                )
 
         param_config = config.ensemble_config.parameter_configs["COND"]
         prior_result = new_prior.load_parameters("COND")["values"]
         posterior_result = new_posterior.load_parameters("COND")["values"]
+
+        # Make sure parameters with zero variance are not updated.
+        assert (prior_result.values[:, :, :5, 0] == 1).all()
```
xjules: not sure that I understand this test or maybe I do but future generations might not :smile: 
can we just do?:
```python
(prior_result.values[:, :, :5, 0] ==posterior_result.values[:, :, :5, 0]).all()
```

src/ert/analysis/_es_update.py
xjules: Not sure either. The only thing I can (fore)see that when numpy changes defaults we will have hard time figure out when it fails. Maybe having them explicit as they are in the current version `rtol=1e-05, atol=1e-08`

dafeda: Can't see numpy doing that to be honest, as it could potentially break lots of things.
What I can see happening however, is that we'll need to increase the tolerance so that it works on real data.
I suggest we release it in bleeding and ask users to do some testing.

tests/ert/ui_tests/cli/test_field_parameter.py
dafeda: Nice, that looks better. I've pushed a fix.

src/ert/analysis/_es_update.py
dafeda: Current code is easy to read I think so will leave as is for now, but thanks.

Issue #11027, opened 2025-06-02
Everest run includes INFO logging from opentelemetry


berland:
When running `everest run` on a example config, some noise is emitted in the terminal:

```
Optimization run failed, with error: Optimization failed: not enough successful realizations to proceed.
To re-run optimization case use command:
`  everest run --new-run everest_config.yml`

2025-06-02 15:20:03,319 azure.core.pipeline.policies.http_logging_policy INFO: Request URL: 'https://norwaywest-0.in.applicationinsights.azure.com//v2.1/track'
Request method: 'POST'
Request headers:
    'Content-Type': 'application/json'
    'Content-Length': '1459'
    'Accept': 'application/json'
    'x-ms-client-request-id': '4afc339a-3fb4-11f0-941c-80e82cfd812f'
    'User-Agent': 'azsdk-python-azuremonitorclient/unknown Python/3.11.11 (Linux-4.18.0-553.46.1.el8_10.x86_64-x86_64-with-glibc2.28)'
A body is sent with the request
2025-06-02 15:20:03,397 azure.core.pipeline.policies.http_logging_policy INFO: Response status: 200
Response headers:
    'Transfer-Encoding': 'chunked'
    'Content-Type': 'application/json; charset=utf-8'
    'Server': 'Microsoft-HTTPAPI/2.0'
    'Strict-Transport-Security': 'REDACTED'
    'X-Content-Type-Options': 'REDACTED'
    'Date': 'Mon, 02 Jun 2025 13:20:03 GMT'
2025-06-02 15:20:03,398 azure.monitor.opentelemetry.exporter.export._base INFO: Transmission succeeded: Item received: 2. Items accepted: 2
```

Such log messages should not be logged to the terminal, and it must be ensure that they are not sent to Azure either. Align with similar behaviour for `ert`.

PR #11049, opened 2025-06-04
Prepare erts logger configuration to work with everest

berland:
The file path must be set in code instead of in the config file such that the config file can be used for multiple applications

**Issue**
Resolves first step of #11027

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/__main__.py
+++ b/src/ert/__main__.py
@@ -625,9 +625,11 @@ def main() -> None:
 
     with open(LOGGING_CONFIG, encoding="utf-8") as conf_file:
         config_dict = yaml.safe_load(conf_file)
-        for _, v in config_dict["handlers"].items():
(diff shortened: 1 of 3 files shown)


StephanDeHoop (approved):
Code makes sense! Just not sure if I follow what you mean with: "such that the config file can be used for multiple applications" (and how this helps) (sorry if it is a silly question :)!)

berland:
In 7abd15993764c1c2bbfb2d91dfe287baa5431b4a you see how Everest can use the same logger configuration when this PR is in place.

berland:
This PR is mostly about removing the default filename "ert-log.txt" from the configuration file. Everest could have overwritten it, but it is more explicit not to have any default when it is supposed to vary.

PR #11053, opened 2025-06-04
Have ert use minimum of NUM_REALIZATIONS and DESIGN_MATRIX realizations

### LINKED ISSUE #10921: NUM_REALIZATIONS should take precedence over DESIGN_MATRIX
We should have the same behavior for both cli and gui and that is that ensemble_size is determined by NUM_REALIZATIONS and not overwritten by design matrix. 
Still we should do a warning that NUM_REALIZATIONS is different the number of realizations in the DESIGN_MATRIX when in use.
Additionally, make sure that test_run will work and in case DESIGN_MATRIX is missing real=0, then this should provide a meaningful error message.

jonathan-eq:
**Issue**
Resolves #10921 

**Approach**
This commit makes ert use the minimum of NUM_REALIZATIONS from config and design matrix. Now we can see in the logs which num_realization is being used, and it is also printed in the terminal when running CLI.

This commit also improves the error message for when a single test run fails due to the first active realization being inactive.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/cli/main.py
+++ b/src/ert/cli/main.py
@@ -98,6 +98,17 @@ def run_cli(args: Namespace, plugin_manager: ErtPluginManager | None = None) -> 
     except ValueError as e:
         raise ErtCliError(f"{args.mode} was not valid, failed with: {e}") from e
 
+    if (
+        ert_config.analysis_config.design_matrix is not None
+        and ert_config.analysis_config.design_matrix.active_realizations is not None
+    ):
+        _log_if_num_realization_was_updated(
+            ert_config.runpath_config.num_realizations,
+            len(ert_config.analysis_config.design_matrix.active_realizations),
+            model.active_realizations.count(True),
+            hasattr(args, "realizations") and args.realizations is not None,
+        )
+
     if args.port_range is None and model.queue_system == QueueSystem.LOCAL:
         # This is within the range for ephemeral ports as defined by
         # most unix flavors https://en.wikipedia.org/wiki/Ephemeral_port
@@ -154,3 +165,30 @@ def run_cli(args: Namespace, plugin_manager: ErtPluginManager | None = None) -> 
         # If monitor has not reported, give some info if the job failed
         msg = end_event.msg if args.disable_monitoring else ""
         raise ErtCliError(msg)
+
+
+def _log_if_num_realization_was_updated(
+    config_num_realizations: int,
+    dm_num_realizations: int,
+    active_realizations_count: int,
+    has_realizations_specified: bool = False,
+) -> None:
+    if dm_num_realizations == config_num_realizations:
+        return
+    if has_realizations_specified:
+        print(
+            "Using realizations intersected between realizations_specified "
+            f"and DESIGN_MATRIX ({active_realizations_count})"
+        )
+    else:
+        print(
+            f"NUM_REALIZATIONS ({config_num_realizations}) is "
+            + ("greater " if dm_num_realizations < config_num_realizations else "less ")
+            + f"than the number of realizations in DESIGN_MATRIX "
+            f"({dm_num_realizations}). Using the realizations from "
+            + (
+                f"DESIGN_MATRIX ({dm_num_realizations})"
+                if dm_num_realizations < config_num_realizations
+                else f"NUM_REALIZATIONS ({config_num_realizations})"
+            )
+        )
--- a/src/ert/run_models/model_factory.py
+++ b/src/ert/run_models/model_factory.py
@@ -39,13 +39,14 @@ if TYPE_CHECKING:
     from ert.namespace import Namespace
     from ert.run_models.event import StatusEvents
 
+logger = logging.getLogger(__name__)
+
 
 def create_model(
     config: ErtConfig,
     args: Namespace,
     status_queue: SimpleQueue[StatusEvents],
 ) -> BaseRunModel:
-    logger = logging.getLogger(__name__)
     logger.info(
         "Initiating experiment",
         extra={
@@ -57,7 +58,6 @@ def create_model(
 
     if args.mode == TEST_RUN_MODE:
         return _setup_single_test_run(config, args, status_queue)
-    validate_minimum_realizations(config, args)
     if args.mode == ENSEMBLE_EXPERIMENT_MODE:
         return _setup_ensemble_experiment(config, args, status_queue)
     if args.mode == EVALUATE_ENSEMBLE_MODE:
@@ -85,7 +85,11 @@ def _setup_single_test_run(
     experiment_name = (
         "single-test-run" if args.experiment_name is None else args.experiment_name
     )
-
+    active_realizations = _get_and_validate_active_realizations_list(args, config)
+    if not active_realizations[0]:
+        raise ConfigValidationError(
+            "Cannot run single test run when the first realization is inactive."
+        )
     return SingleTestRun(
         random_seed=config.random_seed,
         runpath_file=config.runpath_file,
@@ -112,9 +116,10 @@ def _setup_single_test_run(
     )
 
 
-def validate_minimum_realizations(config: ErtConfig, args: Namespace) -> None:
+def validate_minimum_realizations(
+    config: ErtConfig, active_realizations: list[bool]
+) -> None:
     min_realizations_count = config.analysis_config.minimum_required_realizations
-    active_realizations = _get_active_realizations_list(args, config)
     active_realizations_count = int(np.sum(active_realizations))
     if active_realizations_count < min_realizations_count:
         config.analysis_config.minimum_required_realizations = active_realizations_count
@@ -130,7 +135,8 @@ def _setup_ensemble_experiment(
     args: Namespace,
     status_queue: SimpleQueue[StatusEvents],
 ) -> EnsembleExperiment:
-    active_realizations = _get_active_realizations_list(args, config)
+    active_realizations = _get_and_validate_active_realizations_list(args, config)
+    validate_minimum_realizations(config, active_realizations)
     experiment_name = args.experiment_name
     assert experiment_name is not None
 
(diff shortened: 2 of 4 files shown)


src/ert/run_models/model_factory.py
```diff
@@ -195,14 +200,77 @@ def _get_active_realizations_list(args: Namespace, config: ErtConfig) -> list[bo
...
+            ) & np.array(dm_active_realizations[:ensemble_size])
+            if np.any(intersected_realizations):
+                return intersected_realizations
+            else:
+                raise ConfigValidationError(
+                    "The specified realizations do not intersect "
+                    "with the active realizations in the design matrix "
+                    "and num_realizations."
+                )
+        return dm_active_realizations[:ensemble_size]
```
jonathan-eq: I don't like this. I think if you specify NUM_REALIZATIONS 20, you should get 20 active and runnable realizations. If the design matrix only has realizations 25-50, I think you should get realization 25-45 instead of an error. 

I think we need better definitions for what ensemble_size is (is it all the True masks in active_realizations, or the invalid and valid realizations), and what `active_realizations` means. Is `active_realizations` the list of all the realizations (runnable and invalid), or just the True values in the mask=

src/ert/cli/main.py
```diff
@@ -97,7 +97,39 @@ def run_cli(args: Namespace, plugin_manager: ErtPluginManager | None = None) ->
...
+    if (
+        ert_config.analysis_config.design_matrix is not None
+        and ert_config.analysis_config.design_matrix.active_realizations is not None
+    ):
+        dm_num_realizations = (
+            ert_config.analysis_config.design_matrix.active_realizations
+        )
+        config_num_realizations = ert_config.runpath_config.num_realizations
+
+        if len(dm_num_realizations) != ert_config.runpath_config.num_realizations:
```
xjules: this is wrong I think.
`config.num_realizations` is ensemble_size, which is a number representing max_id of all the realizations that are to be run; ie. up to config.num_realizations-1. This implies that if should be more like this:
`if max(dm_num_realizations) == config.num_realizations-1`

src/ert/run_models/model_factory.py
jonathan-eq: Upon discussion, it seems like the ert docs are wrong for this. NUM_REALIZATIONS should be the end_id of the realization - 1. NOT the number of realizations to run.

src/ert/cli/main.py
larsevj: `dm_num_realizations` is a mask, so if design matrix only contains real 200 it will have length 201.
`active_realizations` is poorly named both in design_matrix.py and ensemble_experiment.py and the other run model files i believe.

src/ert/run_models/model_factory.py
```diff
@@ -195,14 +200,77 @@ def _get_active_realizations_list(args: Namespace, config: ErtConfig) -> list[bo
...
+
+    return _realizations(args, ensemble_size).tolist()
+
+
+def _get_and_validate_active_realizations_list(
```
larsevj, line 200: Why is this repeated twice?

src/ert/run_models/model_factory.py
```diff
@@ -389,7 +457,7 @@ def _setup_multiple_data_assimilation(
 
 
 def _realizations(args: Namespace, ensemble_size: int) -> npt.NDArray[np.bool_]:
-    if args.realizations is None:
+    if getattr(args, "realizations", None) is None:
```
larsevj: Why is this changed at all?
And why is it different from the logic above:
`if hasattr(args, "realizations") and args.realizations is not None:` ?

src/ert/run_models/model_factory.py
jonathan-eq: It was due to it being called twice in certain scenarios, and we would see duplicate logs. Will find another way to do it.

src/ert/run_models/model_factory.py
jonathan-eq: I am not sure why this worked beforehand, but if we don't specify `--realizations`, it would blow up here, with and without using design matrix. 
I updated it to be more concise, but it turned out less readable. Will undo.

src/ert/cli/main.py
xjules: Anyway, this needs to be fixed by retrieving the max_id of the design matrix realizations.

jonathan-eq: Yes, thats the length of dm_num_realizations

xjules: Oh, this took me a while :) Then I agree, the names should be changed definitely. 
`dm_num_realizations` should be changed at least in this PR -> dm_realization_mask or the like

src/ert/cli/main.py
```diff
@@ -97,7 +97,39 @@ def run_cli(args: Namespace, plugin_manager: ErtPluginManager | None = None) ->
...
+            if args.realizations and args.realizations is not None:
```
xjules: Don't know what this means? Is this first redudant?

src/ert/cli/main.py
```diff
@@ -97,7 +97,39 @@ def run_cli(args: Namespace, plugin_manager: ErtPluginManager | None = None) ->
...
+                print(
+                    "Using realizations intersected between realizations_specified "
+                    f"and DESIGN_MATRIX ({model.active_realizations.count(True)})"
+                )
+            elif (
+                len(ert_config.analysis_config.design_matrix.active_realizations)
```
xjules: is this `dm_num_realizations` already?

src/ert/cli/main.py
```diff
@@ -97,7 +97,39 @@ def run_cli(args: Namespace, plugin_manager: ErtPluginManager | None = None) ->
...
+                < ert_config.runpath_config.num_realizations
+            ):
+                print(
```
xjules: Can this be combined into a single print statement?
```python
print("Ensemble size specified by NUM_REAZATIONS {...} is different than REALIZATIONS {...} defined in the DESIGN_MATRIX. Using ...."
```

src/ert/cli/main.py
jonathan-eq: Yes, I don't know why I didnt use the variable here. Will change 🔍

src/ert/cli/main.py
jonathan-eq: Yes, but it drops readability a lot. I changed it in the newest commit. Is it worth it?

src/ert/run_models/model_factory.py
```diff
@@ -85,7 +85,11 @@ def _setup_single_test_run(
     experiment_name = (
         "single-test-run" if args.experiment_name is None else args.experiment_name
     )
-
+    active_realizations = _get_and_validate_active_realizations_list(args, config)
```
xjules, line 88: nice

src/ert/cli/main.py
xjules: Have you fixed it?
```python
if args.realizations and args.realizations is not None:
```

jonathan-eq: Oh, I thought you meant line 109

jonathan-eq: I think this is a bug we had from before. args does not always have the realizations property, so we have to check with hasattr(). I guess I forgot it here. Will change 👌

tests/ert/ui_tests/cli/analysis/test_design_matrix.py
```diff
@@ -491,3 +494,116 @@ def test_design_matrix_on_esmda_fail_without_updateable_parameters(
...
+
+
+@pytest.mark.parametrize("realizations_in_design_matrix", [5, 15])
+def test_run_poly_example_with_different_realization_count_chooses_smaller_and_warns(
+    realizations_in_design_matrix,
+    copy_poly_case_with_design_matrix,
+    capsys,
+):
+    num_realizations_in_user_config = 10
+    if realizations_in_design_matrix < num_realizations_in_user_config:
```
xjules: Can you do these expected messages as part of parametrization or is it too complicated?

tests/ert/ui_tests/cli/analysis/test_design_matrix.py
```diff
@@ -491,3 +494,116 @@ def test_design_matrix_on_esmda_fail_without_updateable_parameters(
...
+        "--experiment-name",
+        "test-experiment",
+    )
+    config_path = ErtConfig.from_file("poly.ert").config_path
+
+    realizations_run = os.listdir(Path(config_path) / "poly_out")
+    assert len(realizations_run) == min(
+        realizations_in_design_matrix, num_realizations_in_user_config
+    )
+    assert expected_message in capsys.readouterr().out
```
xjules, line 556: Just wondering if we need to run the entire run_cli just for the sake of creating run_models?

src/ert/cli/main.py
```diff
@@ -97,7 +97,37 @@ def run_cli(args: Namespace, plugin_manager: ErtPluginManager | None = None) ->
...
+                        else "less "
+                    )
+                    + f"than the number of realizations in DESIGN_MATRIX "
+                    f"({dm_num_realizations}). Using the realizations from "
+                    + (
+                        f"DESIGN_MATRIX ({dm_num_realizations})"
+                        if dm_num_realizations < config_num_realizations
+                        else f"NUM_REALIZATIONS ({config_num_realizations})"
+                    )
+                )
```
jonathan-eq: We actually only want to test this without having to run the entire cli. Lets put this into a helper function `log_if_num_realization_was_updated`, and only test that. Is there a better name we should use? @xjules

src/ert/cli/main.py
```diff
@@ -97,7 +97,37 @@ def run_cli(args: Namespace, plugin_manager: ErtPluginManager | None = None) ->
         )
     except ValueError as e:
         raise ErtCliError(f"{args.mode} was not valid, failed with: {e}") from e
-
+    if (
```
xjules, line 101: as agreed, let's put this code into a separate function and then we will have an easier way to test it.

tests/ert/ui_tests/cli/analysis/test_design_matrix.py
jonathan-eq: I think we need it as an integration test. This is to prove that Ert actually choses the smaller NUM_REALIZATIONS, and not only says that it will (that is guaranteed to be a future bug).

src/ert/cli/main.py
jonathan-eq: I changed my mind^

tests/ert/ui_tests/cli/analysis/test_design_matrix.py
jonathan-eq: Yes

xjules (changes requested):
Nice @jonathan-eq ! Could you squash the commits?

PR #11059, opened 2025-06-05
Update to ropt 0.21

verveerpj:
The main change is that the optimizer is now initialized with a dictionary instead of an EnOptConfig object, which could not be serialized.

Other changes:
- Import the ropt exit code with a clear alias
- Adjustments for the changed ropt configuration:
    - Adjust everest2ropt
    - Pass initial values separately to the optimizer
    - Move creation of transforms out of everest2ropt
- Adjust some renamed method calls
- Update relevant test snapshots
- Update tests:
    - Updates for ropt configuration by dict
    - Update test for discrete optimizer for changed option
    - Adjust tests for transforms

      --exec 'just rapid-tests'`)

## When applicable


pyproject.toml
```diff
@@ -143,8 +143,8 @@ everest = [
     "fastapi",
     "decorator",
     "colorama",
-    "ropt[pandas]>=0.20.1,<0.21",
-    "ropt-dakota>=0.20,<0.21",
+    "ropt[pandas]>=0.21,<0.22",
```
berland: Why do we have `<0.22` in here, I assume we have no knowledge of whether that will be incompatible?

verveerpj: We will got to 0.22.0 when the update breaks compatibility, and we do not want older version of everest pull in a newer incompatible version of ropt. Updates that are compatible will be 0.21.x. Kind of semantic versioning where you would do that with the major version, but this is a 0.x.x instable, so using minor version for that for now.

Happy to change it if preferred.

berland: Having it in place increases the probability of getting a failing build, even if the future ropt 0.22 happens to be compatible with both todays Everest and the future Everest (that is, the tests would pass fine, but the build breaks because pip says versions are incompatible - that is a kind of Komodo bleeding noise that we dont want)

I think a better plan is to have a future ropt 0.22 that can support both todays Everest and the future Everest. When/if we get to a point that such compatibility costs too much, then we add the upper bound.

verveerpj: It seems that upper  bounds are seen as undesirable in the python world anyway. I will remove them.

verveerpj: I removed them, but I see a potential issue:

Next time that `ropt` is updated on PyPI, all ERT PR's will try to use it, since there is no upper bound. If it is not backwards compatible, all those PR's will break unless Everest is updated to use it. This will block all PR's until that is done, and even then all PR's must be rebased.

Putting the upper bound will prevent this. I don't think it is an issue, `ropt` itself does not have dependencies with upper bounds, and it is not used somewhere else in Komodo, hence the upper bound should not break anything.

verveerpj: > I removed them, but I see a potential issue:
> 
> Next time that `ropt` is updated on PyPI, all ERT PR's will try to use it, since there is no upper bound. If it is not backwards compatible, all those PR's will break unless Everest is updated to use it. This will block all PR's until that is done, and even then all PR's must be rebased.
> 
> Putting the upper bound will prevent this. I don't think it is an issue, `ropt` itself does not have dependencies with upper bounds, and it is not used somewhere else in Komodo, hence the upper bound should not break anything.

Ah, I forgot, we use `uv` with lock files now, so this should be a non-issue. No upper bounds needed then.

tests/everest/test_ropt_initialization.py
```diff
@@ -297,11 +204,89 @@ def safe_default(obj):
...
+    run_model = EverestRunModel.create(ever_config)
+    monkeypatch.setattr(run_model, "enopt_config", _patched_everest2ropt(ever_config))
+    evaluator_server_config = EvaluatorServerConfig()
     with pytest.raises(ValueError, match=r"Validation error\(s\) in ropt"):
-        everest2ropt_patch.everest2ropt(ever_config)
+        run_model.run_experiment(evaluator_server_config)
+
+
+def test_transforms_controls_scaling():
+    config = EverestConfig.load_file(os.path.join(_CONFIG_DIR, _CONFIG_FILE))
```
yngve-sk: Do these have to rely on the `test_data/mocked_test_case/mocked_test_case.yml`? If possible, please use EverestConfig.with_defaults and set only what is necessary. Same question goes for the other introduced tests here (+probably some/most of the existing but that is outside the scope of this PR). This mainly tests only the mapping from some stuff in the everest config to a given ROPT config.

verveerpj:
> PR itself looks mostly OK, only small comment/question wrt newly introduced tests. Also nice summary of what has been done in the commit body 👌

Done

yngve-sk (approved):
LGTM!

PR #11068, opened 2025-06-06
Allow negative and zero objective weights

### LINKED ISSUE #11067: Allow zero and negative objective weights in Everest
Currently objective weights must be larger than zero. However, there are use cases in multi-objective optimization where the weights might be zero or negative:
- Finding extreme cases in a trade-off: set one objective weight to one, other to zero. Although this may be achieved by using separate configurations, this is conveniently more done by setting the weight to zero, especially if the cost of calculating the extra objective is negligible.
- Composing an multi-objective from a weighted differences of objectives.

It should be possible to have zero or negative weights.

verveerpj:
**Issue**
Resolves #11067

**Approach**
Change pydantic class, add and update tests.

      --exec 'just rapid-tests'`)

## When applicable


src/everest/config/everest_config.py
```diff
@@ -602,6 +603,14 @@ def validate_objective_function_weights_for_all_or_none(cls, functions):
                 "Weight should be given either for all of the objectives"
                 " or for none of them"
             )
+        if weights and sum(weights) < float_info.epsilon:
```
StephanDeHoop, line 606: Just out of curiosity: why should this necessarily be true? Because of the normalization of the weights somewhere else in the code (which would then change the sign of the weights)? Because I can imagine cases where you want to penalize one objective by giving it a negative weight (which might be larger than the weight for the actual objective(s)), or should that objective be negative anyway (since it is a penalty) instead of using a negative weight? If yes, what is the use-case for negative weights?

StephanDeHoop (approved):
Looks good, just one question

src/everest/config/everest_config.py
verveerpj: The use case for negative weights is indeed for something like a penalty. However, when the sum becomes negative, the normalization may go wrong in the optimizer. It is possible that it works, but then your intention seems basically that you are turning the maximization into a minimization. That may work (might need some fixes in the optimizer code), but it is confusing, so for now better disallow it.

StephanDeHoop: Makes sense, thanks :)!

PR #11071, opened 2025-06-06
Rebasing union_genkw

xjules:
**Issue**
Draft PR to rebase union_genkw to main

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/gen_kw_config.py
+++ b/src/ert/config/gen_kw_config.py
@@ -3,8 +3,10 @@ from __future__ import annotations
 import math
 import os
 import warnings
+from collections import defaultdict
 from collections.abc import Callable
 from dataclasses import dataclass
+from enum import StrEnum
 from hashlib import sha256
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Self, cast, overload
@@ -49,19 +51,31 @@ def _get_abs_path(file: str | None) -> str | None:
     return file
 
 
+class DataSource(StrEnum):
+    DESIGN_MATRIX = "design_matrix"
+    SAMPLED = "sampled"
+
+
 @dataclass
 class TransformFunctionDefinition:
     name: str
     param_name: str
+    group_name: str
     values: list[Any]
+    update: bool = True
+    input_source: DataSource = DataSource.SAMPLED
 
 
 @dataclass
 class GenKwConfig(ParameterConfig):
     transform_function_definitions: list[TransformFunctionDefinition]
+    forward_init: bool = False
+    update: bool = True
+    name: str = "GEN_KW"
 
     def __post_init__(self) -> None:
         self.transform_functions: list[TransformFunction] = []
+        self.groups: dict[str, list[TransformFunctionDefinition]] = defaultdict(list)
         for e in self.transform_function_definitions:
             if isinstance(e, dict):
                 self.transform_functions.append(
@@ -73,6 +87,8 @@ class GenKwConfig(ParameterConfig):
                 self.transform_functions.append(
                     self._parse_transform_function_definition(e)
                 )
+            self.groups[e.group_name].append(e)
+        self.update = any(param.update for param in self.transform_function_definitions)
         self._validate()
 
     def __contains__(self, item: str) -> bool:
@@ -103,6 +119,17 @@ class GenKwConfig(ParameterConfig):
 
     @classmethod
     def templates_from_config(
+        cls, gen_kw_list: list[list[str | dict[str, str]]]
+    ) -> list[tuple[str, str]]:
+        return [
+            template
+            for gen_kw in gen_kw_list
+            if (template := GenKwConfig._templates_from_config_single(gen_kw))
+            is not None
+        ]
+
+    @classmethod
+    def _templates_from_config_single(
         cls, gen_kw: list[str | dict[str, str]]
     ) -> tuple[str, str] | None:
         gen_kw_key = cast(str, gen_kw[0])
@@ -134,7 +161,19 @@ class GenKwConfig(ParameterConfig):
         return None
 
     @classmethod
-    def from_config_list(cls, gen_kw: list[str | dict[str, str]]) -> Self:
+    def from_config_list(cls, gen_kw_list: list[list[str | dict[str, str]]]) -> Self:
+        return cls(
+            transform_function_definitions=[
+                item
+                for gen_kw in gen_kw_list
+                for item in GenKwConfig._from_config_list_single_genkw(gen_kw)
+            ]
+        )
+
+    @classmethod
+    def _from_config_list_single_genkw(
+        cls, gen_kw: list[str | dict[str, str]]
+    ) -> list[TransformFunctionDefinition]:
         gen_kw_key = cast(str, gen_kw[0])
 
         options = cast(dict[str, str], gen_kw[-1])
@@ -178,6 +217,8 @@ class GenKwConfig(ParameterConfig):
                         TransformFunctionDefinition(
                             name=items[0],
                             param_name=items[1],
+                            group_name=gen_kw_key,
+                            update=update_parameter,
                             values=items[2:],
                         )
                     )
@@ -199,12 +240,7 @@ class GenKwConfig(ParameterConfig):
                 "to exclude this from updates, set UPDATE:FALSE.\n",
                 gen_kw_key,
             )
-        return cls(
-            name=gen_kw_key,
-            forward_init=False,
-            transform_function_definitions=transform_function_definitions,
-            update=update_parameter,
-        )
+        return transform_function_definitions
 
     def _validate(self) -> None:
         errors = []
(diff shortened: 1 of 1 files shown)


src/ert/config/ensemble_config.py
```diff
@@ -110,11 +109,20 @@ def _check_for_duplicate_gen_kw_param_names(gen_kw_list: list[GenKwConfig]) -> N
...
+        Get the GenKwConfig Returns None if the GenKwConfig does not exist.
+        """
+        gen_kw = [
+            config
+            for config in self.parameter_configs.values()
+            if isinstance(config, GenKwConfig)
         ]
+        if gen_kw:
+            return gen_kw[0]
+        return None
```
yngve-sk: Style suggestion:
```
return next(
    (config for config in self.parameter_configs.values() if isinstance(config, GenKwConfig)),
    None
)
```

src/ert/config/gen_kw_config.py
```diff
@@ -49,32 +57,56 @@ def _get_abs_path(file: str | None) -> str | None:
...
 
 
 @dataclass
 class GenKwConfig(ParameterConfig):
-    transform_function_definitions: list[TransformFunctionDefinition]
+    transform_function_definitions: list[TransformFunctionDefinition] = field(
+        default_factory=list
+    )
+    forward_init: bool = False
+    update: bool = True
```
yngve-sk, line 82: Maybe defer to other PR, but this is obsolete and can be removed when we have one `update` per transform fn def?

src/ert/config/gen_kw_config.py
```diff
@@ -49,32 +57,56 @@ def _get_abs_path(file: str | None) -> str | None:
...
+    name: str = SCALAR_NAME
 
     def __post_init__(self) -> None:
         self.transform_functions: list[TransformFunction] = []
+        self.groups: dict[str, list[TransformFunction]] = defaultdict(list)
```
yngve-sk: Could we make this logic be a `@cached_property` instead of storing as attribute? Adding as attribute makes it a bit less clean when we want to serialize this

src/ert/dark_storage/common.py
```diff
@@ -241,15 +242,16 @@ def data_for_key(
...
         return data_for_response(ensemble, key)
 
-    parameter_config = ensemble.experiment.parameter_configuration
-    parameter_group, parameter_key = _extract_parameter_group_and_key(key)
-    if (
-        parameter_group in parameter_config
-        and parameter_key in parameter_config[parameter_group].parameter_keys
-    ):
-        return data_for_parameter(ensemble, key)
+    # parameter_config = ensemble.experiment.parameter_configuration
```
yngve-sk: Remove?

src/ert/dark_storage/common.py
```diff
@@ -126,6 +126,7 @@ def _extract_parameter_group_and_key(key: str) -> tuple[str, str] | tuple[None,
 
 def data_for_parameter(ensemble: Ensemble, key: str) -> pd.DataFrame:
     group, _ = _extract_parameter_group_and_key(key)
+
```
yngve-sk: Added newline here seems a bit accidental

src/ert/storage/local_ensemble.py
```diff
@@ -557,6 +557,18 @@ def _load_parameters_lazy(
         df = pl.scan_parquet(group_path)
         return df
 
+    @property
+    def _scalar_config(self) -> GenKwConfig | None:
```
yngve-sk, line 557: I think this should ideally not be in `LocalEnsemble`, the logic branching relating to it should be moved to parameter configs, so that storage can keep being agnostic of parameter types. As far as I can see I think you need to check that groups exist (why can't the scalar group be in the `experiment.parameter_configurations` the same as other configs?). And the `update` check, should ideally be treated the same for all configs too, to avoid this kind of branching.

src/ert/storage/migration/to12.py
```diff
@@ -0,0 +1,66 @@
...
+
+        for ens in ensembles:
+            with open(ens / "index.json", encoding="utf-8") as f:
+                ens_file = json.load(f)
+                if ens_file["experiment_id"] != experiment_id:
+                    continue
+
+            datasets = {}
+            parameters = {}
+            one_gen_kw = {}
```
yngve-sk: The parameter occur once per experiment, so loop over ensembles (for just editing configs at least) will result in some potential rewrites

src/ert/storage/migration/to12.py
```diff
@@ -0,0 +1,66 @@
...
+                    group = param_config["name"]
+                    for tfd in param_config.get("transform_function_definitions", []):
+                        new_tfd = tfd.copy()
+                        new_tfd["group_name"] = group
+                        new_tfd["input_source"] = "sampled"
+                        new_tfd["update"] = param_config.get("update", False)
+                        one_gen_kw["transform_function_definitions"].append(new_tfd)
+                    tfd = {}
+                    if (ens / f"{_escape_filename(group)}.nc").exists():
+                        df = pl.read_parquet(ens / f"{_escape_filename(group)}.parquet")
```
yngve-sk, line 50: This part converts xarray to polars, looks OK to me, but I feel the split between them could be more clear, I don't think they need to be done in one function, would lean a bit towards doing it in two functions. One for migrating the configs, and another for migrating the datasets. I think you can migrate the parameter configs first, without depending on checking if there are datasets?

src/ert/storage/migration/to12.py
```diff
@@ -0,0 +1,66 @@
...
+        ensembles = path.glob("ensembles/*")
+
+        experiment_id = None
+        with open(experiment / "index.json", encoding="utf-8") as f:
+            exp_index = json.load(f)
+            experiment_id = exp_index["id"]
+
+        with open(experiment / "parameter.json", encoding="utf-8") as fin:
+            parameters_json = json.load(fin)
+
```
yngve-sk, line 24: Would migrate parameter configs here, and store whatever info you need to do the dataset migration per-ensemble

src/ert/config/design_matrix.py
```diff
@@ -128,56 +130,39 @@ def merge_with_other(self, dm_other: DesignMatrix) -> None:
...
-            elif set(design_keys) & set(existing_keys):
-                raise ConfigValidationError(
-                    "Overlapping parameter names found in design matrix!\n"
-                    f"{DESIGN_MATRIX_GROUP}:{design_keys}\n{parameter_group.name}:{existing_keys}"
-                    "\nThey need to match exactly or not at all."
-                )
-            else:
-                new_param_config += [parameter_group]
-        return new_param_config, design_parameter_group
+            for existing_parameter in config.transform_function_definitions:
```
yngve-sk, line 141: Style: Would see if this function can be simplified a bit doing `next` instead of `for`-`continue` "pattern" that's here. Suggestion ish
```
    def merge_with_existing_parameters(
        self, existing_parameters: list[ParameterConfig]
    ) -> list[ParameterConfig]:
        config = next(
            (config_ for config_ in existing_parameters if isinstance(config_, GenKwConfig)), None
        )
        if not config:
            return [GenKwConfig(
                transform_function_definitions=self.parameter_configuration.transform_function_definitions)]

        existing_defs = {
            p.name:p
            for p in config.transform_function_definitions
            if p.input_source != DataSource.DESIGN_MATRIX
        }

        new_defs = {
            p.name:p for p in
            self.parameter_configuration.transform_function_definitions
        }

        for name in existing_defs.keys() & new_defs.keys():
            new_defs[name].group_name = existing_defs[name].group_name

        merged_defs = list(new_defs.values()) + [
            p for name, p in existing_defs.items() if name not in new_defs
        ]

        return [GenKwConfig(transform_function_definitions=merged_defs)]
```

xjules: It changes the function behavior though, since the proposed version returns also not GenKwConfigs. Yours would only modify GenKw.

src/ert/config/gen_kw_config.py
xjules: This actually needs to be there as otherwise:
```
src/ert/config/gen_kw_config.py:76: in <module>
    @dataclass
     ^^^^^^^^^
/usr/lib64/python3.11/dataclasses.py:1232: in dataclass
    return wrap(cls)
           ^^^^^^^^^
/usr/lib64/python3.11/dataclasses.py:1222: in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash,
/usr/lib64/python3.11/dataclasses.py:1027: in _process_class
    _init_fn(all_init_fields,
/usr/lib64/python3.11/dataclasses.py:545: in _init_fn
    raise TypeError(f'non-default argument {f.name!r} '
E   TypeError: non-default argument 'update' follows default argument
```

src/ert/analysis/_enif_update.py
```diff
@@ -232,20 +232,21 @@ def analysis_EnIF(
...
             f"{(time.time() - start) / 60} minutes"
         )
-        _copy_unupdated_parameters(
-            list(source_ensemble.experiment.parameter_configuration.keys()),
-            parameters,
-            iens_active_index,
-            source_ensemble,
-            target_ensemble,
-        )
+    _copy_unupdated_parameters(
```
yngve-sk, line 243: Nice bugfix for (my) copypasting _es_update when implementing enif, should be just a separate bugfix commit for this for ease of backporting etc.

src/ert/config/parameter_config.py
```diff
@@ -109,6 +109,7 @@ def copy_parameters(
         source_ensemble: Ensemble,
         target_ensemble: Ensemble,
         realizations: npt.NDArray[np.int_],
+        update_mask: bool = False,
```
yngve-sk, line 112: Reading this makes me wonder what `update_mask=True` means (without going in and just reading in the code what it does). Does it mean include updated or exclude updated? Would opt to rename the argument to be less potentially ambiguous.

src/ert/storage/local_ensemble.py
```diff
@@ -569,11 +577,26 @@ def load_parameters(
         otherwise it will return the raw values.
 
         """
+        group_keys: list[str] | None = None
+        config: ParameterConfig | None = None
         if group not in self.experiment.parameter_configuration:
-            raise KeyError(f"{group} is not registered to the experiment.")
-        config = self.experiment.parameter_configuration[group]
+            if self._scalar_config:
```
yngve-sk, line 583: Would suggest making this logic general as to not do this special casing, would push some of that responsibility onto param configs/experiment

src/ert/storage/local_ensemble.py
```diff
@@ -604,8 +627,23 @@ def load_parameters(
...
+        self, group: str, realizations: npt.NDArray[np.int_], update_mask: bool = False
     ) -> npt.NDArray[np.float64]:
+        if (
+            group not in self.experiment.parameter_configuration
+            and self._scalar_config
+            and group in self._scalar_config.groups
+        ):
+            df = self.load_parameters(group, realizations).drop("realization")
+            if update_mask:
+                update_keys = [
```
yngve-sk, line 616: Passing `update_mask` (<-- should be renamed to not be ambiguous) further to the parameter config would let us stay "agnostic" here in the ensemble storage

src/ert/config/parameter_config.py
xjules: ~~Yes, update_mask means to include only the updated parameters.~~
You are right, the semantics here is actually to copy only the non_updated_parameters, will revisit the naming here.

xjules:
Replaced by https://github.com/equinor/ert/pull/11716
Will close this once the above gets merged.

PR #11091, opened 2025-06-11
Fix Event reporter not exiting after failed retries

jonathan-eq:
**Approach**
This commit fixes the issue where the fm_dispatch event reporter would be stuck forever trying to resend events/CONNECT/DISCONNECT if it didn't get ACK messages from the dealer.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/_ert/forward_model_runner/client.py
+++ b/src/_ert/forward_model_runner/client.py
@@ -55,7 +55,13 @@ class Client:
         self._receiver_task: asyncio.Task[None] | None = None
 
     async def __aenter__(self) -> Self:
-        await self.connect()
+        try:
+            await self.connect()
+        except ClientConnectionError:
+            logger.error(
+                "No ack for dealer connection. Connection was not established!"
+            )
+            raise
         return self
 
     async def __aexit__(
@@ -162,5 +168,5 @@ class Client:
                 backoff = min(backoff * 2, 10)  # Exponential backoff
             retries -= 1
         raise ClientConnectionError(
-            f"{self.dealer_id} Failed to send {message!r} to {self.url} after retries!"
+            f"{self.dealer_id} Failed to send {message!r} to {self.url} after retrying!"
         )
--- a/src/_ert/forward_model_runner/reporting/event.py
+++ b/src/_ert/forward_model_runner/reporting/event.py
@@ -75,7 +75,9 @@ class Event(Reporter):
         self._ens_id = None
         self._real_id = None
         self._event_queue: queue.Queue[events.Event | EventSentinel] = queue.Queue()
-        self._event_publisher_thread = ErtThread(target=self._event_publisher)
+        self._event_publisher_thread = ErtThread(
+            target=self._event_publisher, should_raise=False
+        )
         self._done = threading.Event()
         self._ack_timeout = ack_timeout
         self._max_retries = max_retries
@@ -124,6 +126,7 @@ class Event(Reporter):
                         return
                     except ClientConnectionError as exc:
                         logger.error(f"Failed to send event: {exc}")
+                        raise exc
 
         try:
             asyncio.run(publisher())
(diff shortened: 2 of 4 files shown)


src/_ert/forward_model_runner/client.py
```diff
@@ -55,7 +55,13 @@ def __init__(
         self._receiver_task: asyncio.Task[None] | None = None
 
     async def __aenter__(self) -> Self:
-        await self.connect()
+        try:
+            await self.connect()
+        except ClientConnectionError:
```
xjules, line 60: Is this one necessary? Since here ti will just re-raise automatically

jonathan-eq: It is for the logging. If it fails here, it should have the same format as the if it fails during disconnect.

src/_ert/forward_model_runner/client.py
```diff
@@ -162,5 +168,5 @@ async def send(self, message: str | bytes, retries: int | None = None) -> None:
                 backoff = min(backoff * 2, 10)  # Exponential backoff
             retries -= 1
         raise ClientConnectionError(
-            f"{self.dealer_id} Failed to send {message!r} to {self.url} after retries!"
+            f"{self.dealer_id} Failed to send {message!r} to {self.url} after retrying!"
```
xjules, line 171: Why?

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -75,7 +75,9 @@ def __init__(
         self._ens_id = None
         self._real_id = None
         self._event_queue: queue.Queue[events.Event | EventSentinel] = queue.Queue()
-        self._event_publisher_thread = ErtThread(target=self._event_publisher)
+        self._event_publisher_thread = ErtThread(
+            target=self._event_publisher, should_raise=False
```
xjules, line 79: this needs some explanation

src/_ert/forward_model_runner/client.py
jonathan-eq: While looking at a bunch of these errors, it got very confusing. It looks like a f-string where we forgot to enter the number of retries. `Failed to send SOME_MESSAGE after 3 retries`

tests/ert/unit_tests/forward_model_runner/test_event_reporter.py
```diff
@@ -226,3 +226,90 @@ def test_report_with_reconnected_reporter_but_finished_jobs(unused_tcp_port):
             reporter._event_publisher_thread.join()
         assert reporter._done.is_set()
     assert len(mock_server.messages) == 3, "expected 3 Job running messages"
+
+
+def test_event_reporter_does_not_hang_after_failed_retries_for_events(
```
xjules: `test_event_reporter_does_not_hang_after_failed_to_send_events` ?

src/_ert/forward_model_runner/reporting/event.py
jonathan-eq: The default ErtThread behavior is to propagate the exception to the main thread, which we do not want in this case. We just want it to stop. I do not think we need ErtThread is this case (or in any case anymore). It should be considered refactored out in the future.

tests/ert/unit_tests/forward_model_runner/test_event_reporter.py
xjules: Also, can we use parameterize instead; ie. to account for different scenarios.

jonathan-eq: What about `test_event_reporter_does_not_hang_after_failing_to_send_events`

tests/ert/unit_tests/forward_model_runner/test_event_reporter.py
```diff
@@ -226,3 +226,36 @@ def test_report_with_reconnected_reporter_but_finished_jobs(unused_tcp_port):
...
+
+
+@pytest.mark.parametrize(
+    "mocked_server_signal, expected_message",
+    [
+        pytest.param(4, "No ack for dealer disconnection", id="failed_disconnect"),
+        pytest.param(5, "No ack for dealer connection", id="failed_connect"),
+        pytest.param(1, "Failed to send event", id="failed_to_send_event"),
+    ],
+)
```
xjules, line 238: nice!

src/_ert/forward_model_runner/reporting/event.py
xjules: Could you include some of the description into the commit message?

src/_ert/forward_model_runner/client.py
xjules: ok

PR #11108, opened 2025-06-12
Use polars to read the design matrix

larsevj:
      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/design_matrix.py
+++ b/src/ert/config/design_matrix.py
@@ -1,15 +1,14 @@
 from __future__ import annotations
 
+from collections import Counter
 from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, cast
 
 import numpy as np
-import pandas as pd
-from pandas.api.types import is_integer_dtype
+import polars as pl
 
 from ert.config.gen_kw_config import GenKwConfig, TransformFunctionDefinition
-from ert.shared.status.utils import convert_to_numeric
 
 from .parsing import ConfigValidationError, ErrorInfo
 
@@ -34,7 +33,9 @@ class DesignMatrix:
             ) = self.read_and_validate_design_matrix()
         except (ValueError, AttributeError) as exc:
             raise ConfigValidationError.with_context(
-                f"Error reading design matrix {self.xls_filename}: {exc}",
+                f"Error reading design matrix {self.xls_filename}"
+                f" ({self.design_sheet} {self.default_sheet or ''}):"
+                f" {exc}",
                 str(self.xls_filename),
             ) from exc
 
@@ -75,29 +76,29 @@ class DesignMatrix:
             errors.append(
                 ErrorInfo(
                     f"Design Matrices '{self.xls_filename.name} ({self.design_sheet} "
-                    f"{self.default_sheet})' and '{dm_other.xls_filename.name} "
-                    f"({dm_other.design_sheet} {dm_other.default_sheet})' do not "
+                    f"{self.default_sheet or ''})' and '{dm_other.xls_filename.name} "
+                    f"({dm_other.design_sheet} {dm_other.default_sheet or ''})' do not "
                     "have the same active realizations!"
                 )
             )
 
-        common_keys = set(self.design_matrix_df.columns) & set(
-            dm_other.design_matrix_df.columns
-        )
+        common_keys = set(
+            self.design_matrix_df.select(pl.exclude("realization")).columns
+        ) & set(dm_other.design_matrix_df.columns)
         non_identical_cols = set()
         if common_keys:
             for key in common_keys:
-                if not self.design_matrix_df[key].equals(
-                    dm_other.design_matrix_df[key]
+                if not self.design_matrix_df.select(key).equals(
+                    dm_other.design_matrix_df.select(key)
                 ):
                     non_identical_cols.add(key)
             if non_identical_cols:
                 errors.append(
                     ErrorInfo(
                         f"Design Matrices '{self.xls_filename.name} "
-                        f"({self.design_sheet} {self.default_sheet})' and "
+                        f"({self.design_sheet} {self.default_sheet or ''})' and "
                         f"'{dm_other.xls_filename.name} ({dm_other.design_sheet} "
-                        f"{dm_other.default_sheet})' "
+                        f"{dm_other.default_sheet or ''})' "
                         "contains non identical columns with the same name: "
                         f"{non_identical_cols}!"
                     )
@@ -107,19 +108,22 @@ class DesignMatrix:
             raise ConfigValidationError.from_collected(errors)
 
         try:
-            self.design_matrix_df = pd.concat(
+            self.design_matrix_df = pl.concat(
                 [
                     self.design_matrix_df,
-                    dm_other.design_matrix_df.drop(list(common_keys), axis=1),
+                    dm_other.design_matrix_df.select(
+                        pl.exclude([*list(common_keys), "realization"])
+                    ),
                 ],
-                axis=1,
+                how="horizontal",
             )
         except ValueError as exc:
             raise ConfigValidationError(
                 f"Error when merging design matrices "
-                f"'{self.xls_filename.name} ({self.design_sheet} {self.default_sheet})'"
+                f"'{self.xls_filename.name} ({self.design_sheet}"
+                f" {self.default_sheet or ''})'"
                 f" and '{dm_other.xls_filename.name} ({dm_other.design_sheet} "
-                f"{dm_other.default_sheet})': {exc}!"
+                f"{dm_other.default_sheet or ''})': {exc}!"
             ) from exc
 
         for tfd in dm_other.parameter_configuration.transform_function_definitions:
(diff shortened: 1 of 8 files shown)


src/ert/config/design_matrix.py
```diff
@@ -181,105 +180,144 @@ def merge_with_existing_parameters(
 
     def read_and_validate_design_matrix(
         self,
-    ) -> tuple[list[bool], pd.DataFrame, GenKwConfig]:
+    ) -> tuple[list[bool], pl.DataFrame, GenKwConfig]:
         # Read the parameter names (first row) as strings to prevent pandas from
         # modifying them. This ensures that duplicate or empty column names are
         # preserved exactly as they appear in the Excel sheet. By doing this, we
         # can properly validate variable names, including detecting duplicates or
         # missing names.
```
jonathan-eq: Is this comment still relevant?

src/ert/config/design_matrix.py
```diff
@@ -181,105 +180,144 @@ def merge_with_existing_parameters(
...
-            io=self.xls_filename,
+        except pl.exceptions.NoDataError as err:
+            raise ValueError("Design sheet headers are empty.") from err
+        design_matrix_df = pl.read_excel(
+            self.xls_filename,
             sheet_name=self.design_sheet,
-            header=None,
-            skiprows=1,
+            has_header=False,
+            drop_empty_cols=False,
```
jonathan-eq: Do we not want to drop empty columns?

src/ert/config/design_matrix.py
```diff
@@ -181,105 +180,144 @@ def merge_with_existing_parameters(
...
+        ).with_columns(pl.col(pl.Float32, pl.Float64).fill_nan(None))
+        string_cols = [
+            col
+            for col, dtype in design_matrix_df.schema.items()
+            if dtype in {pl.String, pl.Categorical}
+        ]
+        design_matrix_df = design_matrix_df.with_columns(
+            [
+                pl.when(
+                    pl.col(col).str.to_lowercase().is_in(["nan", "null", "none", ""])
```
jonathan-eq, line 229: Can we include whitespace? Something like regex `\s+`

src/ert/config/design_matrix.py
```diff
@@ -181,105 +180,144 @@ def merge_with_existing_parameters(
...
-            design_matrix_df = design_matrix_df.set_index(
-                "REAL", drop=True, verify_integrity=True
-            )
+        design_matrix_df = design_matrix_df.select(
+            design_matrix_df.columns[i] for i in columns_to_keep
+        )
+        param_names = tuple(param_names[i] for i in columns_to_keep)
 
-        if error_list := DesignMatrix._validate_design_matrix(design_matrix_df):
+        if error_list := DesignMatrix._validate_design_matrix(
```
jonathan-eq: Can we rename this to `errors`?

src/ert/config/design_matrix.py
```diff
@@ -181,105 +180,144 @@ def merge_with_existing_parameters(
...
 
+        design_matrix_df.columns = list(param_names)
+        if "REAL" in design_matrix_df.schema:
+            design_matrix_df = design_matrix_df.rename({"REAL": "realization"})
+            real_dt = design_matrix_df.schema.get("realization")
+            assert real_dt is not None
+            if (
+                not real_dt.is_integer()
+                or design_matrix_df.get_column("realization").lt(0).any()
+                or design_matrix_df.get_column("realization").is_duplicated().any()
```
jonathan-eq: I wonder if it is possible to combine these two before doing `.any()`

src/ert/config/design_matrix.py
```diff
@@ -181,105 +180,144 @@ def merge_with_existing_parameters(
...
         """
-        if design_matrix.empty:
+        if design_matrix.is_empty():
             return []
         errors = []
-        column_na_mask = design_matrix.columns.isna()
-        if not design_matrix.columns[~column_na_mask].is_unique:
-            errors.append("Duplicate parameter names found in design sheet")
+        param_name_count = Counter(p for p in param_names if p is not None)
+        duplicate_param_names = [(n, c) for n, c in param_name_count.items() if c > 1]
```
jonathan-eq, line 316: Very readable! Good job 🙆‍♂️

src/ert/config/design_matrix.py
larsevj: Yes, but will update to say polars instead of pandas.

src/ert/config/design_matrix.py
larsevj: A bit tricky since we "have to" read the header separately from the body of the excel sheet, so if we were to drop columns here we would then have to match them up with the header/parameter_names correctly. Therefore we explicitly drop the empty columns below manually with the `columns_to_keep` mask.

src/ert/config/design_matrix.py
larsevj: Will try it out with some tests.

src/ert/config/design_matrix.py
larsevj: Will try it out.

src/ert/config/design_matrix.py
```diff
@@ -181,105 +180,144 @@ def merge_with_existing_parameters(
...
+            drop_empty_cols=False,
+            drop_empty_rows=True,
+            raise_if_empty=False,
+            infer_schema_length=None,
+            read_options={"skip_rows": 1},
+        ).with_columns(pl.col(pl.Float32, pl.Float64).fill_nan(None))
+        string_cols = [
+            col
+            for col, dtype in design_matrix_df.schema.items()
+            if dtype in {pl.String, pl.Categorical}
```
xjules: How do we get the `pl.Categorical` type in here?

src/ert/config/design_matrix.py
```diff
@@ -181,105 +180,144 @@ def merge_with_existing_parameters(
...
             sheet_name=self.design_sheet,
-            header=None,
-            skiprows=1,
+            has_header=False,
+            drop_empty_cols=False,
+            drop_empty_rows=True,
+            raise_if_empty=False,
+            infer_schema_length=None,
+            read_options={"skip_rows": 1},
+        ).with_columns(pl.col(pl.Float32, pl.Float64).fill_nan(None))
```
xjules: There will be definitely some `pl.Int` columns too

src/ert/config/design_matrix.py
```diff
@@ -107,12 +106,12 @@ def merge_with_other(self, dm_other: DesignMatrix) -> None:
...
         try:
-            self.design_matrix_df = pd.concat(
+            self.design_matrix_df = pl.concat(
                 [
                     self.design_matrix_df,
-                    dm_other.design_matrix_df.drop(list(common_keys), axis=1),
+                    dm_other.design_matrix_df.select(pl.exclude(list(common_keys))),
                 ],
-                axis=1,
+                how="horizontal",
```
xjules, line 118: not sure about this one. In another PR I do some concat and use "diagonal" to get the full join. Horizontal is only in case that you add new columns - which is the case here I guess. On the other side, there could be new rows defined in dm_other, but maybe this is checked somewhere else before we do the merge.

src/ert/config/design_matrix.py
```diff
@@ -87,8 +86,8 @@ def merge_with_other(self, dm_other: DesignMatrix) -> None:
         non_identical_cols = set()
         if common_keys:
             for key in common_keys:
-                if not self.design_matrix_df[key].equals(
-                    dm_other.design_matrix_df[key]
+                if not self.design_matrix_df.select(key).equals(
+                    dm_other.design_matrix_df.select(key)
```
larsevj, line 92: check if this gives double errors on different active realizations

src/ert/config/design_matrix.py
larsevj: Do not think that an int can be nan, so in that case the column will be Float is my guess ?

src/ert/config/design_matrix.py
larsevj: Should be solved by now stripping the whitespace when reading the design matrix above, but will add test case for this.

src/ert/config/design_matrix.py
larsevj: This should be checked above when comparing active realizations. I would rather not have polars fill in null values, and then for ert to fail later on.

src/ert/config/design_matrix.py
larsevj: I guess we dont, so ill remove it.

src/ert/config/design_matrix.py
```diff
@@ -75,29 +76,29 @@ def merge_with_other(self, dm_other: DesignMatrix) -> None:
             errors.append(
                 ErrorInfo(
                     f"Design Matrices '{self.xls_filename.name} ({self.design_sheet} "
-                    f"{self.default_sheet})' and '{dm_other.xls_filename.name} "
-                    f"({dm_other.design_sheet} {dm_other.default_sheet})' do not "
+                    f"{self.default_sheet or ''})' and '{dm_other.xls_filename.name} "
```
xjules, line 79: Just wondering if `merge_with_other` should just exploit `update`
https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.update.html#polars.DataFrame.update

larsevj: I am not sure what this comment references, the concatenation?
Currently all columns (except the `realization` column) must be uniquely named, unless they are exactly alike (to follow design2param behavior i think).

src/ert/config/design_matrix.py
```diff
@@ -181,105 +185,154 @@ def merge_with_existing_parameters(
...
+                )
+                .then(None)
+                .otherwise(pl.col(col))
+                .alias(col)
+                for col in string_cols
+            ]
         )
-        design_matrix_df.columns = param_names.to_list()
-        design_matrix_df = design_matrix_df.dropna(axis=1, how="all")
+        columns_to_keep = [
```
xjules, line 238: Shall we check:
```python
if len(param_names) != len(design_matrix_df.columns):
    raise ValueError(...)
```

ie. blank columns?

src/ert/config/design_matrix.py
```diff
@@ -181,105 +185,154 @@ def merge_with_existing_parameters(
...
 
-        if error_list := DesignMatrix._validate_design_matrix(design_matrix_df):
-            error_msg = "\n".join(error_list)
+        if errors := DesignMatrix._validate_design_matrix(
+            design_matrix_df, param_names
+        ):
+            error_msg = "\n".join(errors)
             raise ValueError(f"Design matrix is not valid, error(s):\n{error_msg}")
 
+        design_matrix_df.columns = list(param_names)
```
xjules, line 255: `df.columns` is read-only. -> 
```python
design_matrix_df = design_matrix_df.rename(
    dict(zip(design_matrix_df.columns, param_names))
)
```

src/ert/config/design_matrix.py
larsevj: We currently drop blank columns here via by keeping only if `s.null_count() != design_matrix_df.height` or param_names[i] .
Partially empty columns are validated later.

src/ert/config/design_matrix.py
larsevj: ? https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.columns.html

src/ert/config/design_matrix.py
larsevj: Todo, write a comment!

src/ert/config/design_matrix.py
xjules: mr gpt was using apparently some old version.

tests/ert/unit_tests/sensitivity_analysis/test_design_matrix.py
```diff
@@ -351,58 +344,64 @@ def test_reading_design_matrix_validate_cells(tmp_path, values, error_msg):
...
 def test_default_values_used(tmp_path):
     design_path = tmp_path / "design_matrix.xlsx"
-    design_matrix_df = pd.DataFrame(
+    design_matrix_df = pl.DataFrame(
         {
             "REAL": [0, 1, 2, 3],
             "a": [1, 2, 3, 4],
             "b": [0, 2, 0, 1],
-            "c": ["low", "high", "medium", "low"],
+            "c": ["low", "high", "     medium", "low"],
```
xjules: there is an extra white space `"   medium"`

larsevj: That is part of the test. To check if the whitespace is stripped. See the assert further down.

xjules: Well in that case should this be made more explicit? Encoding it in the test name + small comment.
`test_correct_values_used`?

larsevj: Added a separate test for it now

xjules (approved):
Nice one! :rocket:

PR #11116, opened 2025-06-13
Fix handling of time-outs for SLURM

### LINKED ISSUE #11115: Optimizations hang on a SLURM cluster if a job is terminated due to a time-out

An optimization run with long-running jobs hit the specified maximum run time. This was reported as a failure in the console, but the optimization then just hangs. The log file repeatedly reports that the slurm driver could not parse the output of scontrol/sattr. This appears to be not handled properly leading to an infinite loop.

The time-out should be reported and handled as a normal failure.

### Steps to reproduce

```Python
- Specify `max_runtime` in the Everest config
- Add a job that runs longer than `max_runtime` (i.e. sleep)
```

- [x] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [ ] RHEL8
- [x] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [x] Slurm queue
- [ ] OpenPBS queue

verveerpj:
**Issue**
Resolves #11115

**Approach**
The SLURM drive did not handler time-out, these were detected as a failure, but not properly handled, leading to an infinite loop. This PR adds the TIMEOUT as a job status and handles it, reporting the job as failed with an appropriate exit code.

In addition, a test is added that works both with the SLURM test mock, as well as with a real SLURM cluster.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/scheduler/slurm_driver.py
+++ b/src/ert/scheduler/slurm_driver.py
@@ -5,6 +5,7 @@ import datetime
 import itertools
 import logging
 import shlex
+import signal
 import stat
 import time
 from collections.abc import Iterator
@@ -17,6 +18,7 @@ from .driver import SIGNAL_OFFSET, Driver, FailedSubmit, create_submit_script
 from .event import Event, FinishedEvent, StartedEvent
 
 SLURM_FAILED_EXIT_CODE_FETCH = SIGNAL_OFFSET + 66
+SLURM_TERMINATED_EXIT_CODE = SIGNAL_OFFSET + signal.SIGTERM
 
 logger = logging.getLogger(__name__)
 
@@ -26,6 +28,7 @@ class JobStatus(Enum):
     COMPLETED = auto()
     RUNNING = auto()
     FAILED = auto()
+    TERMINATED = auto()
     CANCELLED = auto()
     COMPLETING = auto()
     CONFIGURING = auto()
@@ -38,7 +41,12 @@ class JobData:
     status: JobStatus | None = None
 
 
-END_STATES = {JobStatus.FAILED, JobStatus.COMPLETED, JobStatus.CANCELLED}
+END_STATES = {
+    JobStatus.FAILED,
+    JobStatus.TERMINATED,
+    JobStatus.COMPLETED,
+    JobStatus.CANCELLED,
+}
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


tests/ert/unit_tests/scheduler/test_slurm_driver.py
```diff
@@ -458,3 +460,37 @@ async def test_slurm_uses_sacct(
...
     # Make sure sacct was tried:
     assert "scontrol failed, trying sacct" in caplog.text
+
+
+async def test_slurm_timeout(tmp_path, caplog, pytestconfig):
+    os.chdir(tmp_path)
+    caplog.set_level(logging.INFO)
+
+    if pytestconfig.getoption("slurm"):
+        cmds = ["sleep 300"]
```
berland: (I suppose plural form of `cmd` is misleading here, as a list here is only one command)

verveerpj: Indeed, will fix.

src/ert/scheduler/slurm_driver.py
```diff
@@ -26,6 +28,7 @@ class JobStatus(Enum):
     COMPLETED = auto()
     RUNNING = auto()
     FAILED = auto()
+    TIMEOUT = auto()
```
berland: Should this be called "TIMEOUT" as it seems the same thing will happen if someone issues a `kill -9 ` on the correct pid on the correct compute node. Call it "TERMINATED" instead? (as the signal does not say why it was terminated).

verveerpj: I have not tried that, but that makes sense. Will do.

berland: I wonder if these JobStatus enums are the strings that `scontrol` can return, as it is used in the `_parse_squeue_output`. Can you check the state reported by `scontrol` when something is terminated due to a timeout?

verveerpj: That will be `TIMEOUT`

berland (approved):
I think this looks correct, but I assume you have done the relevant testing on a real slurm cluster. 

There could be a residual issue with what happens if the slurm job is terminated by SIGKILL and not SIGTERM. 

If you do a kill -SIGTERM manually on the compute node, then I assume this PR will make it look like it was terminated (you have tested this?). Then the question is what happens if you do kill -SIGKILL <pid>.

PR #11123, opened 2025-06-16
Improve naming and asserts of autoscale test

dafeda:
      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
+++ b/tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
@@ -142,18 +142,32 @@ def test_that_correlated_and_independent_observations_are_grouped_separately(
 
 
 @pytest.mark.parametrize("nr_observations", [0, 1, 2])
-def test_misfit_preprocessor_single(nr_observations):
-    """We create nr_observations responses using the linear function y = ax.
-    We don`t have enough information to cluster properly, so we don`t try
-    to scale"""
-    rng = np.random.default_rng(1234)
+def test_edge_cases_with_few_observations_return_default_values(nr_observations):
+    """Test that edge cases with 0-2 observations return default scaling values.
+    We do not know why this is the case.
+    """
     nr_realizations = 1000
     Y = np.ones((nr_observations, nr_realizations))
+
+    rng = np.random.default_rng(1234)
     parameters_a = rng.normal(10, 1, nr_realizations)
+
     for i in range(nr_observations):
         Y[i] = (i + 1) * parameters_a
-    result, *_ = main(Y, Y.mean(axis=1))
+
+    scale_factors, clusters, nr_components = main(Y, Y.mean(axis=1))
+
     np.testing.assert_equal(
-        result,
+        scale_factors,
+        np.array(nr_observations * [1.0]),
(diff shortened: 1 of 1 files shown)


tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
```diff
@@ -142,18 +142,29 @@ def test_that_correlated_and_independent_observations_are_grouped_separately(
 
 
 @pytest.mark.parametrize("nr_observations", [0, 1, 2])
-def test_misfit_preprocessor_single(nr_observations):
-    """We create nr_observations responses using the linear function y = ax.
-    We don`t have enough information to cluster properly, so we don`t try
-    to scale"""
-    rng = np.random.default_rng(1234)
+def test_that_perfectly_correlated_responses_are_not_scaled(nr_observations):
```
xjules: `test_that_responses_with_correlation_one_are_not_scaled`?

dafeda: Thanks, but I prefer the original :)

larsevj: I think the idea behind this test is just to test the following code path in misfit preprocessor:
```
    if len(obs_errors) <= 2:
        logger.info("Observations not correlated or only correlated each other")
        return scale_factors, np.ones(len(obs_errors), dtype=int), nr_components
```
because this test will fail if you increase the number of observations beyond 2, as then the correlated group will be scaled as expected.

larsevj: Now why we do not want to scale in the case that you only input 2 observations, I am not entirely sure.

dafeda: Nice catch Lars!
Should we just rename the test and leave as is for now?

larsevj: I would say rename test and keep the doc-string ( or improve it a bit), but the additional asserts cannot hurt to have as well.

dafeda: I have changed the name of the test to `test_edge_cases_with_few_observations_return_default_values`.

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
```diff
@@ -142,18 +142,32 @@ def test_that_correlated_and_independent_observations_are_grouped_separately(
...
 
 @pytest.mark.parametrize("nr_observations", [0, 1, 2])
-def test_misfit_preprocessor_single(nr_observations):
-    """We create nr_observations responses using the linear function y = ax.
-    We don`t have enough information to cluster properly, so we don`t try
-    to scale"""
-    rng = np.random.default_rng(1234)
+def test_edge_cases_with_few_observations_return_default_values(nr_observations):
+    """Test that edge cases with 0-2 observations return default scaling values.
+    We do not know why this is the case.
```
larsevj, line 147: I guess it trivially makes sense for 0 and 1, but not 2.

PR #11132, opened 2025-06-17
Fix model_factory realization_list validation returning np array and GUI not showing correct NUM_REALIZATIONS when using designmatrix

### LINKED ISSUE #11136: GUI not showing correct NUM_REALIZATIONS

After we added some validation in (5d94cb9b78fc9f1f505d34e1e4493435ab34c67e), the GUI is no longer showing the correct number of realizations, and the field for specifiying active realizations is no longer working. 

Image
(This is with 201 realizations in design matrix and NUM_REALIZATIONS set to 100.)
In this case, `number of realizations` should show 100, and `Active realizations` should NOT show realization 200. 
_No response_

### Steps to reproduce

```Python

```

- [ ] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

jonathan-eq:
**Issue**
Resolves #11136 
The first commit fixes the issue where the `_get_and_validate_active_realization_list` might return a numpy array instead of a `list[bool]` for one of the code paths.

The second commit fixes the issue where the GUI would always show active realizations from design_matrix, even if it was greater than NUM_REALIZATIONS. Ert chooses the minimum of the two, and would run the correct realizations.

**Approach**
🚗 

DM Realizations: 201
NUM_REALIZATIONS 100
In this case, `number of realizations` should show 100, and `Active realizations` should NOT show realization 200. 
**Before**
Image

**After**
image

Image

Image

(This is with 201 realizations in design matrix and NUM_REALIZATIONS set to 100.)

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/ert_config.py
+++ b/src/ert/config/ert_config.py
@@ -7,6 +7,7 @@ import re
 from collections import defaultdict
 from collections.abc import Mapping, Sequence
 from datetime import datetime
+from functools import cached_property
 from os import path
 from pathlib import Path
 from typing import Any, ClassVar, Self, no_type_check, overload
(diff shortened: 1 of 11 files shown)


src/ert/config/ert_config.py
```diff
@@ -1161,6 +1165,49 @@ def _log_config_dict(cls, content_dict: dict[str, Any]) -> None:
         else:
             logger.info(f"Content of the config_dict: {config_dict_content}")
 
+    def get_ensemble_size(self) -> int:
```
xjules: wondering if `ensemble_size` and `active_realizations` could be a `@cached_property` instead?

xjules: What do you think @oyvindeide ?

jonathan-eq: I am not sure, it is not heavy to calculate, and is not called often either. This is just to fix the bug, so it can (should) be refactored later on.

src/ert/gui/simulation/ensemble_experiment_panel.py
```diff
@@ -38,6 +40,8 @@ def __init__(
         self,
         analysis_config: AnalysisConfig,
         ensemble_size: int,
+        active_realizations: list[bool],
+        config_num_realization: int,
```
xjules, line 44: Now this gets `ensemble_size`, `active_realizations` and `config_num_realization` - I think  a comment would be nice to explain the source of each and why is this required.
PS: why is `activate_realizations` not enough here?

jonathan-eq: It is used in the error message when you hover the warming triangle. I will add a comment 👌

jonathan-eq:
@xjules Very good! https://github.com/equinor/ert/issues/11318 🚢

PR #11134, opened 2025-06-17
Test misfit corr groups

dafeda:
      --exec 'just rapid-tests'`)

## When applicable


tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
```diff
@@ -59,10 +59,12 @@ def test_that_get_nr_primary_components_is_according_to_theory(p, rho, seed):
...
     Y = (np.linalg.cholesky(Sigma) @ X).T
 
-    # Adding a bit to the thresholds because of numerical accuracy.
-    assert get_nr_primary_components(Y, threshold_1 + 0.01) == 1
-    assert get_nr_primary_components(Y, threshold_2 + 0.01) == 2
-    assert get_nr_primary_components(Y, threshold_3 + 0.01) == 3
+    # To get 1 component, the threshold must be <= the variance
+    # of the 1st component.
+    # Same for the other components.
+    assert get_nr_primary_components(Y, threshold_1 - 0.01) == 1
```
larsevj, line 65: Would it not be better to +1 to the right side of the asserts? As we now expects one more component?

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
```diff
@@ -71,24 +73,19 @@ def test_that_correlated_and_independent_observations_are_grouped_separately(
...
-    of clusters.
-    It returns the number of components that explain **less** than the
-    variance specified as the threshold.
-    With a threshold of 0.95, the expression used is as follows:
-
-    max(len([1 for i in variance_ratio[:-1] if i < 0.95]), 1),
+    last two are independent of both the main group and each other.
+
+    This will result in a request for 3 clusters, correctly separating the
+    data into its three natural groups:
```
larsevj, line 84: Now this test finally makes some sense.

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
```diff
@@ -99,61 +96,163 @@ def test_that_correlated_and_independent_observations_are_grouped_separately(
...
+    # basis for the two correlation groups.
+    params_a = rng.standard_normal(nr_realizations)
+    params_b = rng.standard_normal(nr_realizations)
+
+    # Create the final response matrix Y
+    Y = np.zeros((nr_observations, nr_realizations))
+
+    # Create Group A: `nr_obs_group_a` perfectly correlated
+    # responses based on `params_a`
+    for i in range(nr_obs_group_a):
```
larsevj, line 215: What if these are "perfectly negatively correlated" instead? I suspect one will not necessarily get 2 groups then, but what is reasonable in that case?

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
dafeda: I think that could work, but we would have to add a bit to the thresholds. Something like this I think:

```
assert get_nr_primary_components(Y, threshold_1  + 0.01) == 2
```

Would you prefer that?

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
dafeda: I remember you saying something about this, but don't recall the details.
Why do you suspect different results with negative correlations?

larsevj: If you have observation a,b,c with correlations:

$$
\begin{bmatrix}
1 & 0.9 & -0.9 \\
0.9 & 1 & 0.1 \\
-0.9 & 0.1 & 1
\end{bmatrix}
$$

you would with our method get distances `ab: 1.01`, `ac: 2.8` and `bc: 2.2`.
Whereas if you where to consider positive and negative correlations as equal (take absolute value) you would get:
`ab: 0.812`, `ac: 0.812` and `bc: 1.27`.
Now whether we want to consider positive and negative correlations as equal is up for question i guess?

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
larsevj: I take it back, think it looks good as is.

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
dafeda: Ah I see. Nice example.

Here's a trivial example that shows that perfectly negatively correlated responses are treated as being far apart, while perfectly positively correlated responses are treated as being close.

```python
corr_matrix = np.array([
    [ 1.0, -1.0],
    [-1.0,  1.0]
])

Z = linkage(corr_matrix, "average", "euclidean")
print("Distance perfect negative correlation: ", Z[0, 2])

corr_matrix = np.array([
    [ 1.0, 1.0],
    [1.0,  1.0]
])

Z = linkage(corr_matrix, "average", "euclidean")
print("Distance perfect positive correlation: ", Z[0, 2])
```

```console
Distance perfect negative correlation:  2.8284271247461903
Distance perfect positive correlation:  0.0
```

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
```diff
@@ -99,61 +96,279 @@ def test_that_correlated_and_independent_observations_are_grouped_separately(
...
     np.testing.assert_equal(obs_errors, obs_error_copy)
 
 
 @pytest.mark.parametrize("nr_observations", [0, 1, 2])
-def test_misfit_preprocessor_single(nr_observations):
-    """We create nr_observations responses using the linear function y = ax.
-    We don`t have enough information to cluster properly, so we don`t try
-    to scale"""
-    rng = np.random.default_rng(1234)
+def test_that_perfectly_correlated_responses_are_not_scaled(nr_observations):
```
larsevj: Is this commit intended to be part of the pr, does there not exist a separate pr for this test?

dafeda: I based this branch on the branch where that test was made.
I have updated the other PR so that it's ready for merging.

dafeda: Should be fixed now.

src/ert/analysis/misfit_preprocessor.py
```diff
@@ -45,7 +54,10 @@ def get_nr_primary_components(
     # sum to get the cumulative proportion of variance explained by each successive
     # component.
     variance_ratio = np.cumsum(singulars**2) / np.sum(singulars**2)
-    return max(len([1 for i in variance_ratio[:-1] if i < threshold]), 1)
+
+    num_components = np.searchsorted(variance_ratio, threshold, side="left") + 1
```
xjules, line 58: nice one!

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
```diff
@@ -99,44 +96,49 @@ def test_that_correlated_and_independent_observations_are_grouped_separately(
...
-    np.testing.assert_equal(
-        scale_factors,
-        np.array(
-            nr_correlated_obs
-            * [np.sqrt(nr_correlated_obs / nr_components_correlated_group)]
-            + nr_uncorrelated_obs
-            * [np.sqrt(nr_uncorrelated_obs / nr_components_uncorrelated_group)]
-        ),
-    )
+    # Check that the three labels are all different
```
xjules, line 117: Just out of curiosity; isn't it expected that the clusters will always have zero overlap?

dafeda: From my understanding hierarchical clustering produces non-overlapping clusters.
Do you agree @larsevj ?

xjules: So why are we asserting that clusters are different? These can be removed then, or?

dafeda: Well, it's a property of the method. If someone tries to replace it with a method that does not have this property, the test will fail.

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
```diff
@@ -171,3 +173,205 @@ def test_edge_cases_with_few_observations_return_default_values(nr_observations)
...
+
+    # Create Group A: (a+i)*X_1 pattern - all positively correlated
+    a = 1  # base scaling factor for group A
+    for i in range(nr_obs_group_a):
+        Y[i] = (a + i) * X_1
+
+    # Create Group B: (b+j)*(-1)^j*X_2 pattern - checkerboard correlation
+    b = 1  # base scaling factor for group B
+    for j in range(nr_obs_group_b):
+        sign = (-1) ** j  # alternates: +1, -1, +1, -1, ...
```
xjules, line 306: maybe more efficient: `sign = -1 if j % 2 else 1`

dafeda: The modulo version is approx 1 second faster with 10 million groups. The number of groups in this test is maximum 6.

```python
import time

n = 10_000_000

# Test exponentiation
start = time.perf_counter()
for j in range(n):
   sign = (-1) ** j
time_exp = time.perf_counter() - start

# Test modulo
start = time.perf_counter()
for j in range(n):
   sign = -1 if j % 2 else 1
time_mod = time.perf_counter() - start

print(f"Exponentiation: {time_exp:.4f}s")
print(f"Modulo:         {time_mod:.4f}s")
print(f"Winner: {'Exponentiation' if time_exp < time_mod else 'Modulo'}")
```

```console
Exponentiation: 1.5748s
Modulo:         0.6303s
Winner: Modulo
```

xjules:
These tests timeout:
```
FAILED tests/ert/unit_tests/analysis/test_es_update.py::test_update_report[0-misfit_preprocess0]@snake_oil_case_storage - Failed: Timeout (>360.0s) from pytest-timeout.
FAILED tests/ert/unit_tests/analysis/test_es_update.py::test_update_report[0-misfit_preprocess2]@snake_oil_case_storage - Failed: Timeout (>360.0s) from pytest-timeout.
FAILED tests/ert/unit_tests/analysis/test_es_update.py::test_update_report[0-misfit_preprocess3]@snake_oil_case_storage - AssertionError: value does not match the expected value in snapshot 
```

dafeda:
> These tests timeout:
> 
> ```
> FAILED tests/ert/unit_tests/analysis/test_es_update.py::test_update_report[0-misfit_preprocess0]@snake_oil_case_storage - Failed: Timeout (>360.0s) from pytest-timeout.
> FAILED tests/ert/unit_tests/analysis/test_es_update.py::test_update_report[0-misfit_preprocess2]@snake_oil_case_storage - Failed: Timeout (>360.0s) from pytest-timeout.
> FAILED tests/ert/unit_tests/analysis/test_es_update.py::test_update_report[0-misfit_preprocess3]@snake_oil_case_storage - AssertionError: value does not match the expected value in snapshot 
> ```

I have updated the snapshots.

xjules (approved):
Not sure if @larsevj has any more comments, but it's :rocket: for my part

Issue #10212, opened 2025-03-05
Replace %d with <IENS> in docs


berland:
Possibly all usages of "`%d`" in docs and code should be replaced by the more precise template `<IENS>` (when `%d` is used for the realization index). 

```
.../ert/docs] rg "%d"
ert/reference/queue/lsf_queue.ert
1:JOBNAME queue_test_%d

ert/reference/configuration/data_types.rst
406:``INIT_FILE:/path/to/priors/files%d`` which tells where the prior files are:
410:		GEN_KW  MY-FAULTS   MULTFLT.tmpl   MULTFLT.INC   MULTFLT.txt    INIT_FILES:priors/multflt/faults%d

ert/reference/queue/local_queue.ert
1:JOBNAME queue_test_%d

ert/reference/configuration/observations.rst
428:the ``GEN_DATA`` result files must have an embedded ``%d`` to indicate the
435:   GEN_DATA RFT_BH67 INPUT_FORMAT:ASCII RESULT_FILE:rft_BH67_%d    REPORT_STEPS:20

ert/reference/configuration/keywords.rst
490:Deprecated syntax still allow use of two `%d` specifers. Use of more than two `%d` specifiers,
504:        -- Using RUNPATH with two %d specifers.
505:        RUNPATH /mnt/my_scratch_disk/realization-%d/iteration-%d
578:        ECLBASE BASE_ECL_NAME%d
585:``<IENS>`` instead of ``%d`` is a better option:
739:        FIELD  ID  PARAMETER  <OUTPUT_FILE>  INIT_FILES:/path/%d  FORWARD_INIT:True  INIT_TRANSFORM:FUNC  OUTPUT_TRANSFORM:FUNC  MIN:X  MAX:Y
756:  Filename to load initial field from. Must contain ``%d`` if ``FORWARD_INIT`` is set to ``False``.
763:    Indicates that the specified files are generated by a forward model and do not require an embedded ``%d``.
766:    Means that the files must be pre-generated before running ERT and require an embedded ``%d`` to differentiate between different realizations.
794:the integer format specified ``%d`` - which ERT will replace with the
799:   FIELD ... INIT_FILES:/path/poro_%d.grdecl
832:filename given should contain a ``%d`` which will be replaced with realization
946:same naming scheme for all realizations (ex: ``gd_%d`` which may resolve to
947:``gd_0``, ``gd_1`` where `%d` is the report step).
[...]

PR #11166, opened 2025-06-19
25.06.make experiments serializable

### LINKED ISSUE #10976: Make response configs serializable
So that we can serialize/deserialize instances of `BaseRunModels` implementations

### LINKED ISSUE #10977: Make ParameterConfigs serializable
So that we can serialize/deserialize instances of `BaseRunModels` implementations

yngve-sk:
**Issue**
Resolves #10977
Resolves #10976

### DIFF
--- a/tests/ert/unit_tests/run_models/test_experiment_serialization.py
+++ b/tests/ert/unit_tests/run_models/test_experiment_serialization.py
@@ -1,0 +1,499 @@
+import shutil
+import string
+import tempfile
+from collections import defaultdict
+from datetime import datetime
+from pathlib import Path
+from typing import Any
+
+from hypothesis import given
+from hypothesis import strategies as st
+
+from ert.config import (
+    Field,
+    ForwardModelStep,
+    GenDataConfig,
+    HookRuntime,
+    ModelConfig,
+    QueueSystem,
+    SummaryConfig,
+    SurfaceConfig,
+    Workflow,
+)
+from ert.config.gen_kw_config import GenKwConfig, TransformFunctionDefinition
+from ert.config.parsing import SchemaItemType
+from ert.config.queue_config import (
+    LocalQueueOptions,
+    LsfQueueOptions,
+    QueueConfig,
+    SlurmQueueOptions,
+    TorqueQueueOptions,
+)
+from ert.field_utils import FieldFileFormat
+from ert.run_models import EnsembleExperiment
+from ert.substitutions import Substitutions
+
+
+def realistic_text(min_size=1, max_size=8):
+    safe_chars = string.ascii_letters + string.digits + "_-"
+    return st.text(alphabet=safe_chars, min_size=min_size, max_size=max_size)
+
+
+optional_nonempty_str = st.one_of(st.none(), realistic_text())
+
+
+# === Strategy for base fields in QueueOptions ===
+@st.composite
+def base_queue_fields(draw):
+    return {
+        "max_running": draw(st.integers(min_value=0, max_value=100)),
+        "submit_sleep": draw(st.floats(min_value=0.0, max_value=10.0)),
+        "num_cpu": draw(st.integers(min_value=1, max_value=64)),
+        "realization_memory": draw(st.integers(min_value=0, max_value=1_000_000)),
+        "job_script": shutil.which("fm_dispatch.py") or "fm_dispatch.py",
+        "project_code": draw(optional_nonempty_str),
+        "activate_script": draw(optional_nonempty_str),
+    }
+
+
+# === Strategies for each queue type ===
+
+
+@st.composite
+def lsf_queue_options_strategy(draw):
+    base = draw(base_queue_fields())
+    return LsfQueueOptions(
+        name=QueueSystem.LSF,
+        bhist_cmd=draw(optional_nonempty_str),
+        bjobs_cmd=draw(optional_nonempty_str),
+        bkill_cmd=draw(optional_nonempty_str),
+        bsub_cmd=draw(optional_nonempty_str),
+        exclude_host=draw(optional_nonempty_str),
+        lsf_queue=draw(optional_nonempty_str),
+        lsf_resource=draw(optional_nonempty_str),
+        **base,
+    )
+
+
+@st.composite
+def torque_queue_options_strategy(draw):
+    base = draw(base_queue_fields())
+    return TorqueQueueOptions(
+        name=QueueSystem.TORQUE,
+        qsub_cmd=draw(optional_nonempty_str),
+        qstat_cmd=draw(optional_nonempty_str),
+        qdel_cmd=draw(optional_nonempty_str),
+        queue=draw(optional_nonempty_str),
+        cluster_label=draw(optional_nonempty_str),
+        job_prefix=draw(optional_nonempty_str),
+        keep_qsub_output=draw(st.booleans()),
+        **base,
+    )
+
+
+@st.composite
+def slurm_queue_options_strategy(draw):
+    base = draw(base_queue_fields())
+    return SlurmQueueOptions(
+        name=QueueSystem.SLURM,
+        sbatch_cmd=draw(optional_nonempty_str),
+        squeue_cmd=draw(optional_nonempty_str),
+        scancel_cmd=draw(optional_nonempty_str),
+        queue=draw(optional_nonempty_str),
+        cluster_label=draw(optional_nonempty_str),
+        job_prefix=draw(optional_nonempty_str),
+        keep_sbatch_output=draw(st.booleans()),
+        **base,
+    )
+
+
+@st.composite
+def local_queue_options_strategy(draw):
+    base = draw(base_queue_fields())
+    return LocalQueueOptions(name=QueueSystem.LOCAL, **base)
+
+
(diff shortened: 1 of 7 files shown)


src/ert/config/gen_kw_config.py
```diff
@@ -49,30 +51,32 @@ def _get_abs_path(file: str | None) -> str | None:
...
-class TransformFunctionDefinition:
+class TransformFunctionDefinition(BaseModel):
     name: str
     param_name: str
     values: list[Any]
 
 
-@dataclass
 class GenKwConfig(ParameterConfig):
+    type: Literal["gen_kw"] = "gen_kw"
```
xjules: Once we have pydantic distributions, maybe we can remove `transform_function_definitions` attribute completely. This requires some discussion though.

src/ert/config/distribution.py
```diff
@@ -2,29 +2,31 @@
...
-from pydantic import field_validator, model_validator
-from pydantic.dataclasses import dataclass
+from pydantic import BaseModel, field_validator, model_validator
 from scipy.stats import norm
 
 from .parsing import ConfigValidationError, ErrorInfo
 
 
-class TransSettingsValidation:
+class TransSettingsValidation(BaseModel):
```
xjules: Nice, I wanted to do it, but completely forgot :D

src/ert/config/field.py
```diff
@@ -95,6 +95,14 @@ class Field(ParameterConfig):
     grid_file: str
     mask_file: Path | None = None
 
+    @field_serializer("output_file")
```
xjules: what about `serialize_grid_file`?

src/ert/config/gen_kw_config.py
```diff
@@ -91,34 +90,54 @@ class TransformFunction:
...
+    def model_post_init(self, ctx: Any) -> None:
+        transform_functions: list[TransformFunction] = []
+
+        errors = []
         for e in self.transform_function_definitions:
-            if isinstance(e, dict):
-                self.transform_functions.append(
-                    self._parse_transform_function_definition(
-                        TransformFunctionDefinition(**e)
+            try:
```
xjules: if this try: except: is here I guess the equivalent try: except: can be removed from `_parse_transform_function_definition`

src/ert/config/parameter_config.py
```diff
@@ -137,11 +136,11 @@ def load_parameter_graph(self) -> nx.Graph[int]:
         """
 
     def to_dict(self) -> dict[str, Any]:
```
xjules: if this function `to_dict()` still needed?

src/ert/storage/local_experiment.py
```diff
@@ -314,11 +315,19 @@ def get_surface(self, name: str) -> IrapSurface:
 
     @cached_property
     def parameter_configuration(self) -> dict[str, ParameterConfig]:
-        params = {}
-        for data in self.parameter_info.values():
-            param_type = data.pop("_ert_kind")
-            params[data["name"]] = _KNOWN_PARAMETER_TYPES[param_type](**data)
-        return params
+        adapter = TypeAdapter(
```
xjules: much nicer!

tests/ert/unit_tests/config/test_gen_kw_config.py
```diff
@@ -320,7 +353,6 @@ def test_gen_kw_distribution_errors(tmpdir, distribution, mean, std, error):
             "Incorrect number of values: \\['0', '1', '2'\\], "
             "provided for variable MYNAME with distribution UNIFORM.",
         ),
-        ("MYNAME", "Too few instructions provided in"),
```
xjules: hm, is this due to pydantic? I thought that this was triggered prior to pydantic creation; ie. when we try to parse TFD.

src/ert/config/gen_kw_config.py
```diff
@@ -91,34 +90,54 @@ class TransformFunction:
     @property
     def parameter_list(self) -> dict[str, float]:
         """Return the parameters of the distribution as a dictionary."""
-        return {k: v for k, v in asdict(self.distribution).items() if k != "name"}
+        return self.distribution.model_dump(exclude="name")
```
xjules: This one might be:
`return self.distribution.model_dump(exclude={"name"})`

src/ert/config/field.py
yngve-sk: Its type is already a str, so it does not need conversion

src/ert/config/gen_kw_config.py
yngve-sk: Hmm I think we need both? One definition may have multiple errors pertaining to the distribution

src/ert/config/parameter_config.py
yngve-sk: Sadly yes, in migrations, suggest removing those in another PR though

tests/ert/unit_tests/config/test_gen_kw_config.py
yngve-sk: Yep, the typing was non optional but the `dataclasses.dataclass` did not enforce it. So having a custom validation error for this is now superceded by pydantic validations

berland:
Do we want 7000 lines added to the repository for this? Can we do unit testing instead of snapshots? (these snapshots look like the will give us some maintenance burden in the future)

yngve-sk:
> Do we want 7000 lines added to the repository for this? Can we do unit testing instead of snapshots? (these snapshots look like the will give us some maintenance burden in the future)

As per our discussion, snapshots is still better than unit tests trying to exhaustively assert correspondences between expected keys/values in the object. I will write some more background into into the PR description and commit body.

tests/ert/unit_tests/run_models/test_experiment_serialization.py
```diff
@@ -0,0 +1,826 @@
...
+    EnsembleInformationFilter,
+    EnsembleSmoother,
+    MultipleDataAssimilation,
+    create_model,
+)
+from ert.storage import open_storage
+
+
+@pytest.fixture(autouse=True)
+def isolate_test_cwd(tmp_path, monkeypatch):
```
berland: can you reuse the existing t`ests/conftest.py::change_to_tmpdir()` fixture?

tests/ert/unit_tests/run_models/test_experiment_serialization.py
```diff
@@ -0,0 +1,826 @@
...
+    config = ErtConfig.from_file(config_file)
+    config.random_seed = 1  # Ensure deterministic
+
+    _create_and_verify_runmodel_snapshot(
+        config,
+        snapshot,
+        cli_args=Namespace(
+            mode=ES_MDA_MODE,
+            realizations="1,2",
+            target_ensemble="iter-%d",
```
berland: can we use `<ITER>` here? ref https://github.com/equinor/ert/issues/10212

tests/ert/unit_tests/run_models/test_experiment_serialization.py
```diff
@@ -0,0 +1,826 @@
...
+
+    snapshot.assert_match(
+        serialized,
+        f"{Path(config.user_config_file).stem}.json",
+    )
+
+
+cases_to_test = [
+    # "heat_equation/config.ert",
+    # "poly_example/poly.ert",
```
berland: is poly in progress? Should probably be included?

tests/ert/unit_tests/run_models/test_experiment_serialization.py
```diff
@@ -0,0 +1,826 @@
...
+    "snake_oil/snake_oil.ert",
+    "snake_oil_field/snake_oil_field.ert",
+    # "snake_oil_field/snake_oil_surface.ert", # Missing irap file?..
+]
+
+# Note2reviewer:
```
berland: Thanks for the comment. Read and burn line 670. 😄 🔥

tests/ert/unit_tests/run_models/test_experiment_serialization.py
```diff
@@ -0,0 +1,826 @@
...
+        serialized,
+        f"{Path(config.user_config_file).stem}.json",
+    )
+
+
+cases_to_test = [
+    # "heat_equation/config.ert",
+    # "poly_example/poly.ert",
+    "snake_oil/snake_oil.ert",
+    "snake_oil_field/snake_oil_field.ert",
```
berland: if snake_oil is almost a subset of snake_oil_field, then the latter might suffice?

tests/ert/unit_tests/run_models/test_experiment_serialization.py
yngve-sk: I can use it for the snapshot tests, but we need some way to explicitly generate a new tmpdir per invocation from  hypothesis, for the hypothesis tests. I added some comments for this.

tests/ert/unit_tests/run_models/test_experiment_serialization.py
yngve-sk: I think this is fixed now by including only relevant examples

berland:
The json files disappeared during squashing, but I think it looks fine now!

berland (approved):

PR #11167, opened 2025-06-19
Check files specified in config with schema type EXISTING_PATH for read access

### LINKED ISSUE #10716: Make sure that template files have the correct access permissions
When reading templates either from RUN_TEMPLATE or GEN_KW template options we should make sure that the files have the correct read access. 

Also from DATA_FILE. 

Check on startup that the files can be read (as happens with other files, eg OBS_CONFIG)

Now it just crashes on start simulation

flikka:
**Issue**
Resolves #10716 

**Approach**
Should make sense that the arguments that we already define as "EXISTING_PATH" and "EXISTING_PATH_INLINE" should BOTH exist (checked today) and readable (not checked, altthough _INLINE will be validated due to a failing read_file). Now they fail "the same way".

TODO: Check for unintended other consequences, test, and make some new tests, for DATA_FILE, RUN_TEMPLATE and GEN_KW

bilde

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/config/parsing/test_lark_parser.py
+++ b/tests/ert/unit_tests/config/parsing/test_lark_parser.py
@@ -196,6 +196,48 @@ def test_that_invalid_boolean_values_are_handled_gracefully():
         )
 
 
+@pytest.mark.parametrize(
+    "config_file_contents",
+    [
+        dedent(
+            """\
+             NUM_REALIZATIONS 1
+             RUN_TEMPLATE file.txt out.txt
+             """
+        ),
+        dedent(
+            """\
+             NUM_REALIZATIONS 1
+             DATA_FILE file.txt
+             """
+        ),
+        dedent(
+            """\
+             NUM_REALIZATIONS 1
+             GEN_KW file.txt file.txt
+             """
+        ),
+    ],
+)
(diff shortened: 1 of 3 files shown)


jonathan-eq (commented):
Beautiful tests!
I think it is worth looking into mr.copilots suggetions, we might have a set format that is used

tests/ert/unit_tests/config/parsing/test_lark_parser.py
```diff
@@ -196,6 +196,48 @@ def test_that_invalid_boolean_values_are_handled_gracefully():
...
+        dedent(
+            """\
+             NUM_REALIZATIONS 1
+             GEN_KW file.txt file.txt
+             """
+        ),
+    ],
+)
+@pytest.mark.usefixtures("use_tmpdir")
+def test_that_file_without_read_fails_gracefully(config_file_contents):
```
jonathan-eq: Great test! Can we rename it `test_that_file_without_read_access_fails_gracefully`? I think that would be the cherry on top 🍒

jonathan-eq (changes requested):
Very good! A tiny change, and we will merge it when the tests pass 🤩

PR #11182, opened 2025-06-23
Move env_save to cover the everest tests

### LINKED ISSUE #11173: Everest tests should check for clean environment
Like the ERT tests do, currently Everest tests seem to not do this

frode-aarstad:
**Issue**
Resolves #11173 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,5 @@
+import os
+
 import pytest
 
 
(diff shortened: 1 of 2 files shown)


src/everest/bin/config_branch_script.py
```diff
@@ -127,6 +127,8 @@ def config_branch_entry(args: list[str] | None = None) -> None:
         yaml.dump(yml_config, f)
     print(f"New config file {options.output_config} created.")
 
+    cleanup_logging()
```
berland, line 130: Are each of these cleanup_logging() calls a result of the env_save fixture, as in the tests would not pass if not?

berland: A follow-up question would be if the tests pass after the first commit in this PR? If not, they must be squashed.

frode-aarstad: Yes and we should probably squash the two commits

berland (approved):
👍🏻

PR #11186, opened 2025-06-23
Simplify ExtParam saving

yngve-sk:
Replace with using only `.formatted_control_names` from `ControlConfig` instead of having multiple independent computations of it


src/ert/config/ext_param_config.py
```diff
@@ -35,55 +35,18 @@ class ExtParamConfig(ParameterConfig):
...
     def parameter_keys(self) -> list[str]:
-        if isinstance(self.input_keys, dict):
-            flattened = []
-            for key, subkeys in self.input_keys.items():
-                for subkey in subkeys:
-                    flattened.append(key + subkey)
-            return flattened
-        else:
-            return self.input_keys
+        return self.input_keys
```
verveerpj, line 38: Do we need this property if we just forwarding `input_keys`, or can we rename the latter?

src/ert/run_models/everest_run_model.py
```diff
@@ -25,7 +25,6 @@
 from typing_extensions import TypedDict
 
 from ert.config import (
-    ExtParamConfig,
     ParameterConfig,
     ResponseConfig,
```
verveerpj: Removing the last comma will make this format nicer.

src/everest/config/control_config.py
```diff
@@ -224,7 +223,10 @@ def formatted_control_names(self) -> list[str]:
             if isinstance(variable, ControlVariableGuessListConfig):
                 for index in range(1, len(variable.initial_guess) + 1):
                     formatted_names.append(f"{self.name}.{variable.name}.{index}")
-            elif variable.index is not None:
+            elif (
+                isinstance(variable, ControlVariableConfig)
```
verveerpj: Is the `isinstance `really necessary?

tests/everest/test_controls.py
```diff
@@ -258,7 +259,7 @@ def test_control_bad_variables(variables, control_data_no_variables: dict):
         ControlConfig.model_validate(data)
 
 
-def test_controls_variables_array_vs_index():
+def test_controls_ordering_is_consistent_for_ropt_and_extparam():
```
verveerpj, line 262: This is good, but the indexing is now in order, so that will not detect if we get a regression where some sorting is introduced. This name index is misleading in the sense that you would naturally assume they must be ordered. So maybe some test with indices that are not ordered that checks this is useful?

verveerpj (approved):
This cleans up the code nicely, only small comments.

src/ert/config/ext_param_config.py
yngve-sk: Hmm it is an abstract property of `ParameterConfig` and is mainly used to describe which "keys" it can plot that goes through dark storage, so for now it has to stay

src/everest/config/control_config.py
yngve-sk: mypy induced

verveerpj: That is one of the issues with mypy imho. I always use an assert, but I am not sure what is the better option to resolve things like this. Any opinions?

tests/everest/test_controls.py
yngve-sk: Added a test with "incorrect" index sorting that asserts the `index` is disregarded. Will create issue for it being confusing

src/everest/config/control_config.py
yngve-sk: Edit: Seems now it can be removed (locally for me at least), maybe I had some temp different state when I added it in, hmm. Will remove if it passes on here too

tests/everest/test_controls.py
yngve-sk: https://github.com/equinor/ert/issues/11219

PR #11203, opened 2025-06-24
Improved typing and removed some tmp vars

JHolba:
      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/analysis/event.py
+++ b/src/ert/analysis/event.py
@@ -65,7 +65,7 @@ class AnalysisDataEvent(AnalysisEvent):
 class AnalysisErrorEvent(AnalysisEvent):
     event_type: Literal["AnalysisErrorEvent"] = "AnalysisErrorEvent"
     error_msg: str
-    data: DataSection | None = None
+    data: DataSection
 
 
 class AnalysisCompleteEvent(AnalysisEvent):


src/ert/run_models/base_run_model.py
```diff
@@ -179,8 +179,13 @@ class BaseRunModel(BaseModelWithContextSupport, ABC):
...
     _total_iterations: int = PrivateAttr(default=1)
 
-    def __init__(self, _total_iterations: int | None = None, **data: Any) -> None:
-        status_queue = data.pop("status_queue", None)
+    def __init__(
+        self,
+        *,
+        status_queue: queue.SimpleQueue[StatusEvents],
+        _total_iterations: int | None = None,
+        **data: Any,
```
jonathan-eq: What is the Any type here? Should it be a dict/typeddict?

src/ert/run_models/initial_ensemble_run_model.py
```diff
@@ -23,8 +23,9 @@ class InitialEnsembleRunModel(BaseRunModel, ABC):
     ert_templates: list[tuple[str, str]]
     _observations: dict[str, pl.DataFrame] | None = PrivateAttr()
 
-    def __init__(self, **data: Any) -> None:
-        observations = data.pop("observations", None)
+    def __init__(
+        self, *, observations: dict[str, pl.DataFrame] | None, **data: Any
```
jonathan-eq, line 27: Same here, is there a way to make the data type more specific than Any?

jonathan-eq (changes requested):
Very good :)
Some questions, and maybe squash the commits

src/ert/run_models/base_run_model.py
JHolba: data is all of the class variables. It would be possible to name them all there, but I don't think it should be part of this PR.

src/ert/run_models/evaluate_ensemble.py
```diff
@@ -44,8 +44,7 @@ def run_experiment(
         self.log_at_startup()
         self._is_rerunning_failed_realizations = rerun_failed_realizations
         ensemble = self._storage.get_ensemble(UUID(self.ensemble_id))
-        experiment = ensemble.experiment
```
berland, line 47: Can you describe the change more precisely in the commit message? "Small refactor" is not helpful when later looking at the git blame..

JHolba: Would this be better?

> Improve typing
> Remove two temporary vars that are not needed

berland: That is better yes!

JHolba:
> Can you make a better PR title and tick off the box for it in the top message?

done

src/ert/run_models/initial_ensemble_run_model.py
berland: (unanswered)

JHolba: same answer as https://github.com/equinor/ert/pull/11203#discussion_r2169061539 :)
it is possible, but it would have to be edited every time the dataclass it inherits adds new variables. I'm not sure if we want to do that and I don't think it needs to be in this PR either way.

jonathan-eq (approved):
🚢

PR #11209, opened 2025-06-24
Proposing new autoscaling algorithm based on kmeans

dafeda:
**NB! This is just a quick sketch of how we might do variance inflation using simple KMeans clustering instead of the complicated and not well understood method currently implemented. I have not analysed it carefully at all, and I'm sure there are better ways of doing this.**

## TODO

- Use silhouette analysis as recommended by docs:
https://scikit-learn.org/stable/auto_examples/cluster/plot_kmeans_silhouette_analysis.html#sphx-glr-auto-examples-cluster-plot-kmeans-silhouette-analysis-py

- Consider doing PCA:
https://scikit-learn.org/stable/modules/clustering.html#k-means
"Running a dimensionality reduction algorithm such as [Principal component analysis (PCA)](https://scikit-learn.org/stable/modules/decomposition.html#pca) prior to k-means clustering can alleviate this problem and speed up the computations."

## Drogon

The variance inflation algorithm based on KMmeans produces 3 clusters on Drogon compared to more than 70 produced by the current method.
3 clusters might be more natural since we have 3 observation types (I think). Seismic, pressure and tracer (?).

      --exec 'just rapid-tests'`)

## When applicable


misfit-preprocessor.py
```diff
@@ -0,0 +1,403 @@
...
+    .list.first()
+    > 0
+)
+
+# %% [markdown]
+# # Add column with normalized misfit and sort by it
+
+# %%
+# First calculate the misfit column
+df_filtered = df_filtered.with_columns(
```
xjules, line 76: Maybe, once you have the misfits pl code, can you inject it into local_ensemble.py as `def compute_misfits`?
Thus we can remove src/ert/libres_facade.py

misfit-preprocessor.py
```diff
@@ -0,0 +1,403 @@
...
+    max_k = min(10, len(data) // 5)  # Don't try more clusters than we have data points
+    K_range = range(2, max_k + 1)  # Try 2, 3, 4, ... clusters
+
+    inertias = []  # This will store how "spread out" each clustering solution is
+    for k in K_range:
+        kmeans = KMeans(n_clusters=k, random_state=42, n_init=10)
+        kmeans.fit(data.reshape(-1, 1))
+        inertias.append(
+            kmeans.inertia_
+        )  # inertia = sum of distances from points to cluster centers
```
xjules, line 310: what kind of distance measure does it use?

misfit-preprocessor.py
```diff
@@ -0,0 +1,403 @@
...
+
+    # Find the "elbow" - where the improvement starts slowing down
+    if len(inertias) >= 3:
+        diffs = np.diff(inertias)  # How much inertia decreases each step
+        diff2 = np.diff(diffs)  # How much the decrease rate changes
+        elbow_idx = np.argmax(diff2)  # Find biggest change in decrease rate
+        optimal_k = K_range[elbow_idx + 1]
+    else:
+        optimal_k = 3  # If we don't have enough data points, just use 3 clusters
```
xjules, line 319: I like this one :D

misfit-preprocessor.py
```diff
@@ -0,0 +1,403 @@
...
+        inertias.append(
+            kmeans.inertia_
+        )  # inertia = sum of distances from points to cluster centers
+
+    # Find the "elbow" - where the improvement starts slowing down
+    if len(inertias) >= 3:
+        diffs = np.diff(inertias)  # How much inertia decreases each step
+        diff2 = np.diff(diffs)  # How much the decrease rate changes
+        elbow_idx = np.argmax(diff2)  # Find biggest change in decrease rate
+        optimal_k = K_range[elbow_idx + 1]
```
xjules, line 317: A very naive question. Don't we want to select `k` such that sum of inertias is minimum out of all?

misfit-preprocessor.py
dafeda: From the [documentation](https://scikit-learn.org/stable/modules/clustering.html#k-means):

image

Does this answer your question?

The docs also recommend using `silhouette analysis` analysis to select the number of cluster, so we might want to use something like that instead:

https://scikit-learn.org/stable/auto_examples/cluster/plot_kmeans_silhouette_analysis.html#sphx-glr-auto-examples-cluster-plot-kmeans-silhouette-analysis-py

misfit-preprocessor.py
dafeda: The reason we don't want that I think, is because inertia always decreases as the number of clusters increases.
So we would get 0 intertia if every point is its own cluster.

PR #11211, opened 2025-06-24
Improve exception handling in plot window

### LINKED ISSUE #10889: Storage server times out on OSErrors
If the storage directory is inaccessible and throws OSErrors (at least PermissionErrors have been observed), then it will print the exception to the terminal and probably exit(?). Ert will hang until it reaches a timeout.

## To reproduce

* `rm -rf logs poly_out storage`
* Start `ert gui poly.ert`
* Immediately do `chmod 000 storage` in a different terminal (_immediately_ means "before dark_storage has finished starting)
* Click "Create plot"

Observe an exception printed in your terminal and  Ert showing a spinning mouse cursor until timeout.

## Expected behaviour

Storage server sends a message back to to Ert that it got an OSError. Ert then displays this using the Suggestor widget.

frode-aarstad:
There are several things that can go wrong when calling the api from the plot window. This change will handle all BaseExceptions and print better feedback to the user.

**Issue**
Resolves #10889 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/gui/tools/plot/plot_window.py
+++ b/src/ert/gui/tools/plot/plot_window.py
@@ -5,7 +5,6 @@ from typing import TYPE_CHECKING, cast
 
 import numpy as np
 import pandas as pd
-from httpx import RequestError
 from pandas import DataFrame
 from PyQt6.QtCore import Qt
 from PyQt6.QtCore import pyqtSlot as Slot
@@ -23,6 +22,7 @@ from PyQt6.QtWidgets import (
 )
 
 from ert.gui.ertwidgets import CopyButton, showWaitCursorWhileWaiting
+from ert.services._base_service import ServerBootFail
 from ert.utils import log_duration
 
 from .customize import PlotCustomizer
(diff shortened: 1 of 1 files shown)


src/ert/gui/tools/plot/plot_window.py
```diff
@@ -92,6 +93,7 @@ def create_error_dialog(title: str, content: str) -> QDialog:
     layout.addWidget(text)
 
     qd.setLayout(layout)
+    qd.resize(450, 150)
```
StephanDeHoop, line 96: I never worked with a `QDialog`; is this ever a problem that the UI scales in a way that some functionality is invisible/unusable?

frode-aarstad: Not sure what you mean. This change was to make the error dialog more horizontal to look better with the new error messages I added.

StephanDeHoop: Ah ok, clear, thanks!

PR #11221, opened 2025-06-25
Add deprecation warnings for non-iso data formats

### LINKED ISSUE #11141: Deprecate non-ISO8601 dates in everest
Everest allows non ISO-8601 date formats. That is a bad thing, and should be removed eventually.

In case user config depends on these date formats, we need to trigger a DeprecationWarning and also log it.

This date string allows e.g. "2000-1-1" when fed through `datetime.datetime.strptime`:
https://github.com/equinor/ert/blob/460da29930b91bbb5493092cc6efa51045c240be/src/everest/strings.py#L5

Illustrated in tests:
https://github.com/equinor/ert/blob/460da29930b91bbb5493092cc6efa51045c240be/tests/everest/test_everlint.py#L313-L316

verveerpj:
**Issue**
Resolves #11141 

**Approach**
None-iso dates of this form: "2000-1-1" will be accepted, but will emit a configuration warning.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/config/well_config.py
+++ b/src/everest/config/well_config.py
@@ -1,3 +1,4 @@
+import logging
 from datetime import datetime
 
 from pydantic import (
@@ -8,7 +9,8 @@ from pydantic import (
     field_validator,
 )
 
-from everest.strings import DATE_FORMAT
+from ert.config import ConfigWarning
+from everest.strings import EVEREST
 
 
 class WellConfig(BaseModel):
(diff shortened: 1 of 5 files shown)


berland:
Will this be logged? I don't think the `deprecated_warn` will yield logs from checking the code.

berland:
Great. Can you confirm that the warning shows up on stdout/stderr while running as a user?

If the GUI is active, will it be shown in the Suggestor-GUI element (I do not know if that works for Everest, if not is an issue to be written) or in stdout/stderr or both?

verveerpj:
> Great. Can you confirm that the warning shows up on stdout/stderr while running as a user?
> 
> If the GUI is active, will it be shown in the Suggestor-GUI element (I do not know if that works for Everest, if not is an issue to be written) or in stdout/stderr or both?

It does indeed show up in the console. I do not know about the GUI, it does not work on my setup (not sure if it should, never seriously tried before)

berland:
When run with `--gui` it shows up on the console as without GUI. But that is ok for now.

berland (approved):
👍🏻

Issue #11217, opened 2025-06-24
Tests tainting repository


berland:
At least one test is tainting the repository, causing failures (flaky!) like:

https://github.com/equinor/ert/actions/runs/15849527065/job/44679308843?pr=11186

Such failures block the release pipeline.

Tip for reproduction: `touch storage; chmod 000 storage` and then see which test that fails.

berland:
Reopened as the flakyness is still present.

berland:
Not observed the last weeks. Closing.

PR #11223, opened 2025-06-26
Make GHA workflows fail when polluting project dir

yngve-sk:
**Issue**
Resolves [#11217](https://github.com/equinor/ert/issues/11217)

### DIFF
--- a/.github/workflows/test_ert.yml
+++ b/.github/workflows/test_ert.yml
@@ -52,7 +52,11 @@ jobs:
     - name: GUI Test
       if: inputs.test-type == 'gui-tests'
       run: |
+        touch storage
+        chmod 000 storage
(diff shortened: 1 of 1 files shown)


.github/workflows/test_ert.yml
```diff
@@ -52,7 +52,10 @@ jobs:
     - name: GUI Test
       if: inputs.test-type == 'gui-tests'
       run: |
+        touch storage
+        chmod 000 storage
         uv run just ert-gui-tests
+        rm storage
```
berland, line 59: I think this will fail, you need to do `chmod 700 storage` before you can remove it.

yngve-sk: oops, fixed

PR #11238, opened 2025-06-27
Turn distributions into pydantic class

xjules:
Refactor distributions to become a pydantic class.
This provides a nicer validation and the actual transformation.

image

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/distribution.py
+++ b/src/ert/config/distribution.py
@@ -1,0 +1,353 @@
+from __future__ import annotations
+
+import math
+import warnings
+from dataclasses import fields, is_dataclass
+from typing import Any, Literal, Self
+
+import numpy as np
+from pydantic import field_validator, model_validator
+from pydantic.dataclasses import dataclass
+from scipy.stats import norm
+
+from .parsing import ConfigValidationError, ErrorInfo
+
+
+class TransSettingsValidation:
+    @classmethod
+    def create(cls, *args: Any, **kwargs: Any) -> Self:
+        return cls(*args, **kwargs)
+
+    @classmethod
+    def get_param_names(cls) -> list[str]:
+        assert is_dataclass(cls), "This method should only be called on dataclasses"
+        return [f.name for f in fields(cls) if f.init and f.name != "name"]
+
+
+@dataclass
+class UnifSettings(TransSettingsValidation):
+    name: Literal["uniform"] = "uniform"
+    min: float = 0.0
+    max: float = 1.0
+
+    @model_validator(mode="after")
+    def valid_unif_params(self) -> Self:
+        if not (self.min < self.max):
+            raise ConfigValidationError(
+                f"Minimum {self.min} must be strictly less than the maximum {self.max}"
+                " for uniform distribution"
+            )
+        return self
+
+    def transform(self, x: float) -> float:
+        y = float(norm.cdf(x))
+        return y * (self.max - self.min) + self.min
+
+
+@dataclass
+class LogUnifSettings(TransSettingsValidation):
+    name: Literal["logunif"] = "logunif"
+    min: float = 0.0
+    max: float = 1.0
+
+    @model_validator(mode="after")
+    def valid_logunif_params(self) -> Self:
+        if not (self.min < self.max):
+            raise ConfigValidationError(
+                f"Minimum {self.min} must be strictly less than the maximum {self.max}"
+                " for log uniform distribution"
+            )
+        return self
+
+    def transform(self, x: float) -> float:
+        log_min, log_max = math.log(self.min), math.log(self.max)
+        tmp = norm.cdf(x)
+        # Shift according to max / min
+        return math.exp(log_min + tmp * (log_max - log_min))
+
+
+@dataclass
+class DUnifSettings(TransSettingsValidation):
+    name: Literal["dunif"] = "dunif"
+    steps: int = 1000
+    min: float = 0.0
+    max: float = 1.0
+
+    @model_validator(mode="after")
+    def valid_dunif_params(self) -> Self:
+        errors = []
+        if not (self.min < self.max):
+            errors.append(
+                ErrorInfo(
+                    message=f"Minimum {self.min} must be strictly less"
+                    f" than the maximum {self.max} for duniform distribution"
+                )
+            )
+        if not self.steps > 1:
+            errors.append(
+                ErrorInfo(
+                    message=f"Number of steps {self.steps} must be larger"
+                    " than 1 for duniform distribution"
+                )
+            )
+        if errors:
+            raise ConfigValidationError.from_collected(errors)
+        return self
+
+    def transform(self, x: float) -> float:
+        y = norm.cdf(x)
+        return (math.floor(y * self.steps) / (self.steps - 1)) * (
+            self.max - self.min
+        ) + self.min
+
+
+@dataclass
+class NormalSettings(TransSettingsValidation):
+    name: Literal["normal"] = "normal"
+    mean: float = 0.0
+    std: float = 1.0
+
+    @field_validator("std")
+    @classmethod
+    def std_must_be_positive(cls, value: float) -> float:
+        if value < 0:
+            raise ConfigValidationError(f"Negative STD {value} for normal distribution")
+        return value
+
+    def transform(self, x: float) -> float:
+        return x * self.std + self.mean
+
+
+@dataclass
+class LogNormalSettings(TransSettingsValidation):
+    name: Literal["lognormal"] = "lognormal"
+    mean: float = 0.0
+    std: float = 1.0
+
+    @field_validator("std")
+    @classmethod
+    def std_must_be_positive(cls, value: float) -> float:
+        if value <= 0:
+            raise ConfigValidationError(
+                f"Negative STD {value} for lognormal distribution"
+            )
+        return value
+
+    def transform(self, x: float) -> float:
+        return math.exp(x * self.std + self.mean)
+
+
+@dataclass
+class TruncNormalSettings(TransSettingsValidation):
+    name: Literal["truncated_normal"] = "truncated_normal"
+    mean: float = 0.0
+    std: float = 1.0
+    min: float = 0.0
+    max: float = 1.0
+
+    @model_validator(mode="after")
+    def valid_trunc_normal_params(self) -> Self:
+        errors = []
+        if not (self.min < self.max):
+            errors.append(
+                ErrorInfo(
+                    message=f"Minimum {self.min} must be strictly less than"
+                    f" the maximum {self.max} for truncated_normal distribution"
+                )
+            )
+        if self.std <= 0:
+            errors.append(
+                ErrorInfo(
+                    message=f"Negative STD {self.std} for truncated normal distribution"
+                )
+            )
+        if errors:
+            raise ConfigValidationError.from_collected(errors)
+        return self
+
+    def transform(self, x: float) -> float:
+        y = x * self.std + self.mean
+        return max(min(y, self.max), self.min)  # clamp
+
+
+@dataclass
+class RawSettings(TransSettingsValidation):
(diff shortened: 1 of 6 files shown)


src/ert/config/gen_kw_config.py
```diff
@@ -56,6 +70,30 @@ class TransformFunctionDefinition:
     values: list[Any]
 
 
+@dataclass
```
yngve-sk, line 73: Can we use pydantic dataclasses for interop with the pydantic BaseModels / ErtConfig etc? (Goes for all instances of dataclasses in this PR)

src/ert/config/distribution.py
```diff
@@ -0,0 +1,273 @@
...
+            return []
+
+
+@dataclass
+class TransUnifSettings(TransSettingsValidation):
+    name: Literal["uniform"] = "uniform"
+    min: float = 0.0
+    max: float = 1.0
+
+    def trans(self, x: float) -> float:
```
larsevj: Any reason why the full word transformation is not used throughout? I do not think that this shortening is very describing of what is going on?
Also would suggest to omit trans or transformation from the class names, because even though the transformation is a property of the class, the class itself represent a distribution.

src/ert/config/distribution.py
```diff
@@ -0,0 +1,273 @@
...
+        # mean is the expectation of log( y )
+        return math.exp(x * self.std + self.mean)
+
+
+@dataclass
+class TransTruncNormalSettings(TransSettingsValidation):
+    name: Literal["truncated_normal"] = "truncated_normal"
+    mean: float = 0.0
+    std: float = 1.0
+    min: float = 0.0
```
larsevj, line 145: No validation of min vs max?

src/ert/config/distribution.py
```diff
@@ -0,0 +1,273 @@
...
+@dataclass
+class TransLogNormalSettings(TransSettingsValidation):
+    name: Literal["lognormal"] = "lognormal"
+    mean: float = 0.0
+    std: float = 1.0
+
+    @field_validator("std")
+    @classmethod
+    def std_must_be_positive(cls, value: float) -> float:
+        if value < 0:
```
larsevj: Do not think that std can be 0

src/ert/config/distribution.py
```diff
@@ -0,0 +1,273 @@
...
+            return self.max - math.sqrt((1 - y) * inv_norm_right)
+
+
+@dataclass
+class TransErrfSettings(TransSettingsValidation):
+    name: Literal["errf"] = "errf"
+    min: float = 0.0
+    max: float = 1.0
+    skewness: float = 0.0
+    width: float = 1.0
```
larsevj, line 236: width must be >0

src/ert/config/distribution.py
```diff
@@ -0,0 +1,273 @@
...
+        log_min, log_max = math.log(self.min), math.log(self.max)
+        tmp = norm.cdf(x)
+        log_y = log_min + tmp * (log_max - log_min)  # Shift according to max / min
+        return math.exp(log_y)
+
+
+@dataclass
+class TransDUnifSettings(TransSettingsValidation):
+    name: Literal["dunif"] = "dunif"
+    steps: int = 1000
```
larsevj, line 72: I believe steps must be int, > 1

src/ert/config/distribution.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def trans(self, x: float) -> float:
+        y = float(norm.cdf(x))
+        return y * (self.max - self.min) + self.min
+
+
+@dataclass
+class TransLogUnifSettings(TransSettingsValidation):
+    name: Literal["logunif"] = "logunif"
+    min: float = 0.0
```
larsevj, line 50: Both must  be positive, max > min

src/ert/config/gen_kw_config.py
xjules: ~~Yes, I was actually considering it. I can change it...~~ It seems that this would require more refactor; especially wrt. to ConfigValidationError. Will do it in another PR.

src/ert/config/distribution.py
xjules: We haven't done that before; so was thinking to add extra validation in another PR.

src/ert/config/distribution.py
xjules: 1) I agree to change `trans`, but would then go for `trasform` rather then `transformation`.
2) I'll leave `TransSettingsValidation`, but can remove `Trans` from all the sub-classes.

src/ert/config/distribution.py
```diff
@@ -0,0 +1,273 @@
...
+    max: float = 1.0
+
+    def trans(self, x: float) -> float:
+        log_min, log_max = math.log(self.min), math.log(self.max)
+        tmp = norm.cdf(x)
```
yngve-sk, line 64: any special reason for not inlining into expresion below?

src/ert/config/distribution.py
```diff
@@ -0,0 +1,273 @@
...
+
+@dataclass
+class TransTriangularSettings(TransSettingsValidation):
+    name: Literal["triangular"] = "triangular"
+    min: float = 0.0
+    mode: float = 0.5
+    max: float = 1.0
+
+    @model_validator(mode="after")
+    def valid_traingular_params(self) -> Self:
```
yngve-sk: typo 🚆

src/ert/config/distribution.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+@dataclass
+class TransUnifSettings(TransSettingsValidation):
+    name: Literal["uniform"] = "uniform"
+    min: float = 0.0
+    max: float = 1.0
+
+    def trans(self, x: float) -> float:
+        y = float(norm.cdf(x))
```
yngve-sk, line 43: Nitpick: float cast wasn't there before, some typing issue?

src/ert/config/gen_kw_config.py
```diff
@@ -56,6 +70,30 @@ class TransformFunctionDefinition:
...
+        | TransNormalSettings
+        | TransTruncNormalSettings
+        | TransErrfSettings
+        | TransDerrfSettings
+        | TransTriangularSettings,
+        Field(discriminator="name"),
+    ]
+
+    @property
+    def parameter_list(self) -> dict[str, float]:
```
yngve-sk, line 92: More of a comment than a question, doesn't need to be addressed: Usages of this (legacy storage `dump`/dark storage `experiments` endpoint fn seem to be all set to be removed) but this is just to keep those things alive for now I guess

tests/ert/unit_tests/config/test_gen_kw_config.py
```diff
@@ -77,7 +77,7 @@ def test_gen_kw_config_get_priors():
             TransformFunctionDefinition("KEY6", "DUNIF", ["3", "0", "1"]),
             TransformFunctionDefinition("KEY7", "ERRF", ["0", "1", "2", "3"]),
             TransformFunctionDefinition("KEY8", "DERRF", ["1", "1", "2", "3", "4"]),
-            TransformFunctionDefinition("KEY9", "LOGUNIF", ["0", "1"]),
+            TransformFunctionDefinition("KEY9", "LOGUNIF", ["1", "2"]),
```
yngve-sk, line 80: Was this change necessary for it to work?

tests/ert/unit_tests/config/test_gen_kw_config.py
```diff
@@ -136,7 +136,7 @@ def test_gen_kw_config_get_priors():
     assert {
         "key": "KEY9",
         "function": "LOGUNIF",
-        "parameters": {"MIN": 0, "MAX": 1},
+        "parameters": {"MIN": 1, "MAX": 2},
```
yngve-sk, line 139: connected w above comment

tests/ert/unit_tests/config/test_gen_kw_config.py
```diff
@@ -417,7 +417,8 @@ def test_gen_kw_trans_func(tmpdir, params, xinput, expected):
             transform_function_definitions=[tfd],
         )
         tf = gkw.transform_functions[0]
-        assert abs(tf.calculate(xinput, float_args) - expected) < 10**-15
+        # assert abs(tf.calculate(xinput, float_args) - expected) < 10**-15
```
yngve-sk: commented code

src/ert/config/gen_kw_config.py
```diff
@@ -541,167 +515,12 @@ def _parse_transform_function_definition(
                     f"{t.name} with distribution {t.param_name}.",
                     self.name,
                 ) from e
+        try:
+            dist = get_distribution(t.param_name, t.values)
+        except ValueError as e:
+            raise ConfigValidationError.with_context(
```
yngve-sk: Behavior same as before for GUI suggester if this is hit? Does/should it show the context etc here?

src/ert/config/distribution.py
xjules: ~~hm, yes can be removed!~~ Yes, mypy complains otherwise.

src/ert/config/distribution.py
xjules: no, purely copy&paste from the old code. Can be inlined.

src/ert/config/distribution.py
xjules: Fixed

tests/ert/unit_tests/config/test_gen_kw_config.py
xjules: Yes, now the validation kicked in and it failed.

tests/ert/unit_tests/config/test_gen_kw_config.py
xjules: yes, the same as above

src/ert/config/gen_kw_config.py
xjules: image

src/ert/config/gen_kw_config.py
xjules: `parameter_list` is actually not a best name for the function as it only extract pydantic attributes of the class. Maybe this should be renamed?

src/ert/config/gen_kw_config.py
yngve-sk: Is there a test that hits this error? Just the error itself, not necessarily GUI

xjules: yes, in here there are tests for various errors etc.: tests/ert/unit_tests/config/test_gen_kw_config.py

src/ert/config/gen_kw_config.py
yngve-sk: Can be future work

src/ert/config/distribution.py
```diff
@@ -0,0 +1,356 @@
...
+    @classmethod
+    def std_must_be_positive(cls, value: float) -> float:
+        if value <= 0:
+            raise ConfigValidationError(
+                f"Negative STD {value} for lognormal distribution"
+            )
+        return value
+
+    def transform(self, x: float) -> float:
+        # mean is the expectation of log( y )
```
yngve-sk: where is `y`?

src/ert/config/distribution.py
```diff
@@ -0,0 +1,356 @@
...
+
+
+class TransSettingsValidation:
+    @classmethod
+    def create(cls, *args: Any, **kwargs: Any) -> Self:
+        return cls(*args, **kwargs)
+
+    @classmethod
+    def get_param_names(cls) -> list[str]:
+        if is_dataclass(cls):
```
yngve-sk: are they not always dataclasses? do we need this conditional?

src/ert/config/distribution.py
xjules: removing the comment.

PR #11254, opened 2025-06-30
Split out memory tests

yngve-sk:
Background: Sometimes we need to set some env vars to configure OS memory collection (which affects the memory we measure from a python context).

Can be run as just ert-memory-tests

### DIFF
--- a/justfile
+++ b/justfile
@@ -21,7 +21,7 @@ ert-cli-tests:
     pytest {{pytest_args}} tests/ert/ui_tests/cli
 
 ert-memory-tests:
-    _RJEM_MALLOC_CONF="dirty_decay_ms:100,muzzy_decay_ms:100" pytest {{pytest_args}} tests/ert -m "memory_test"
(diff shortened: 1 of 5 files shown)


justfile
```diff
@@ -12,16 +12,19 @@ pytest_args := env("ERT_PYTEST_ARGS", "--quiet")
...
+    pytest -n auto --dist loadgroup tests/ert/unit_tests tests/everest --hypothesis-profile=fast -m "not (integration_test or flaky or memory_test)"
 
 ert-gui-tests:
     pytest {{pytest_args}} --mpl tests/ert/ui_tests/gui
 
 ert-cli-tests:
     pytest {{pytest_args}} tests/ert/ui_tests/cli
 
+ert-memory-tests:
+    _RJEM_MALLOC_CONF="dirty_decay_ms:100,muzzy_decay_ms:100" pytest {{pytest_args}} tests/ert -m "memory_test"
```
berland: we can probably afford at least "`-n 2`" here.

berland (approved):

PR #11256, opened 2025-06-30
Split memory tests on komodo

yngve-sk:


justfile
```diff
@@ -21,10 +21,10 @@ ert-cli-tests:
     pytest {{pytest_args}} tests/ert/ui_tests/cli
 
 ert-memory-tests:
-    _RJEM_MALLOC_CONF="dirty_decay_ms:100,muzzy_decay_ms:100" pytest -n 2 {{pytest_args}} tests/ert -m "memory_test"
+    _RJEM_MALLOC_CONF="dirty_decay_ms:100,muzzy_decay_ms:100" pytest -n 2 {{pytest_args}} tests/ert -m "memory_test or limit_memory"
```
berland: Should there be a memray option here?

yngve-sk: Edit: It should be there but can't be mixed with the `memory_test` marked ones, so it has a separate invocation.
```
test_field_param_memory 
```

PR #11260, opened 2025-06-30
Test if server is running after the last time interval

berland:
The exponential backoff code had forgot to test if the server was actually running after the last sleep. That means the timeout was effectively 60 seconds instead of the claimed 120 seconds.

Also refactored the sleep times to be more eager at the start. During normal cpu loads, the startup time seems to be around 8-9 seconds on Linux and on Mac hardware. The previous backoff scheme then dictated it would always wait for 14 seconds before it senses that t the server is actually running.

The currently computed sleep times are as follows:

[0.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.08, 1.175, 1.351, 1.703, 2.407, 3.815, 6.630, 12.26, 23.52, 46.043]

which sums up to 120 seconds.

**Issue**
60 seconds extra wait time should reduce some flakyness (we waited the same amount of time before, but for no good)

**Approach**
🧠 

      --exec 'just rapid-tests'`)

## When applicable


berland:
TODO: Test interactively with LSF involved.

PR #11270, opened 2025-07-01
Have evaluator send terminate message instead of driver kill

jonathan-eq:
**Issue**
Improves #9832

**Approach**
This commit makes it so that the evaluator sends a `TERMINATE_MSG`
to the connected dispatchers, letting them know it is time to shut down.
The corresponding task in the dispatcher calls SIGTERM on the pgroup,
which would look the same as if we used the queue specific commands
from the driver (bkill, qdel). The evaluator waits five seconds
after sending the message, then calls `scheduler.killAllJobs()`
that forcefully kills all jobs that did not exit gracefully within
the set sleep period. This should improve the issue with too many open
files, as we do not call the shell commands for stopping jobs as often
(only if graceful shutdown did not succeeed).

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/_ert/forward_model_runner/client.py
+++ b/src/_ert/forward_model_runner/client.py
@@ -20,6 +20,7 @@ DISCONNECT_MSG = b"DISCONNECT"
 ACK_MSG = b"ACK"
 HEARTBEAT_MSG = b"BEAT"
 HEARTBEAT_TIMEOUT = 5.0
+TERMINATE_MSG = b"TERMINATE"
 
 
 class Client:
@@ -44,6 +45,7 @@ class Client:
         # wherein the linger is set to 0 to discard all messages in the queue
         self.socket.setsockopt(zmq.LINGER, 0)
         self.dealer_id = dealer_name or f"dispatch-{uuid.uuid4().hex[:8]}"
+        self.received_terminate_message: asyncio.Event = asyncio.Event()
         self.socket.setsockopt_string(zmq.IDENTITY, self.dealer_id)
 
         if token is not None:
@@ -118,6 +120,8 @@ class Client:
                             f"{self.dealer_id} heartbeat failed - reconnecting."
                         )
                     last_heartbeat_time = asyncio.get_running_loop().time()
+                elif raw_msg == TERMINATE_MSG:
+                    self.received_terminate_message.set()
                 else:
                     await self.process_message(raw_msg.decode("utf-8"))
             except zmq.ZMQError as exc:
@@ -157,7 +161,6 @@ class Client:
                     f"reconnecting: {exc}"
                 )
             except asyncio.CancelledError:
-                self.term()
                 raise
 
             if retries > 0:
--- a/src/_ert/forward_model_runner/reporting/event.py
+++ b/src/_ert/forward_model_runner/reporting/event.py
@@ -2,8 +2,11 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import os
 import queue
+import signal
 import threading
+import uuid
 from pathlib import Path
 from typing import Final
 
@@ -16,7 +19,10 @@ from _ert.events import (
     ForwardModelStepSuccess,
     dispatcher_event_to_json,
 )
-from _ert.forward_model_runner.client import Client, ClientConnectionError
+from _ert.forward_model_runner.client import (
+    Client,
+    ClientConnectionError,
+)
 from _ert.forward_model_runner.reporting.base import Reporter
 from _ert.forward_model_runner.reporting.message import (
     _STEP_EXIT_FAILED_STRING,
@@ -97,43 +103,64 @@ class Event(Reporter):
         if self._event_publisher_thread.is_alive():
             self._event_publisher_thread.join()
 
+    async def handle_publish(self, client: Client):
+        event = None
+        start_time = None
+        while True:
+            try:
+                if self._done.is_set() and start_time is None:
+                    start_time = asyncio.get_event_loop().time()
+                if event is None:
+                    event = self._event_queue.get(timeout=0.1)
+                    if event is self._sentinel:
+                        break
+                if (
+                    start_time
+                    and (asyncio.get_event_loop().time() - start_time)
+                    > self._finished_event_timeout
+                ):
+                    break
+                await client.send(dispatcher_event_to_json(event), self._max_retries)
+                event = None
+            except asyncio.CancelledError:
+                return
+            except ClientConnectionError as exc:
+                logger.error(f"Failed to send event: {exc}")
+                return
+            except queue.Empty:
+                await asyncio.sleep(0)
+
+    async def listen_for_terminate_message(self, client: Client):
+        try:
+            await client.received_terminate_message.wait()
+
+            logger.info("Received a TERMINATE message. Terminating the forward model")
+            pgid = os.getpgid(os.getpid())
+            os.killpg(pgid, signal.SIGTERM)
+        except asyncio.CancelledError:
+            return
+
     def _event_publisher(self):
         async def publisher():
             async with Client(
                 url=self._evaluator_url,
                 token=self._token,
                 ack_timeout=self._ack_timeout,
+                dealer_name=f"dispatch-real-{self._real_id}-{uuid.uuid4().hex[:6]}",
             ) as client:
-                event = None
-                start_time = None
-                while True:
-                    try:
-                        if self._done.is_set() and start_time is None:
-                            start_time = asyncio.get_event_loop().time()
-                        if event is None:
-                            event = self._event_queue.get()
-                            if event is self._sentinel:
-                                break
-                        if (
-                            start_time
-                            and (asyncio.get_event_loop().time() - start_time)
-                            > self._finished_event_timeout
-                        ):
-                            break
-                        await client.send(
-                            dispatcher_event_to_json(event), self._max_retries
-                        )
-                        event = None
-                    except asyncio.CancelledError:
-                        return
-                    except ClientConnectionError as exc:
-                        logger.error(f"Failed to send event: {exc}")
-                        raise exc
+                publisher_task = asyncio.create_task(
+                    self.handle_publish(client), name="publisher_task"
+                )
+                listener_task = asyncio.create_task(
+                    self.listen_for_terminate_message(client),
+                    name="terminate_message_listener_task",
+                )
+                await publisher_task
+                if not listener_task.done():
+                    listener_task.cancel()
+                    await listener_task
 
-        try:
-            asyncio.run(publisher())
-        except ClientConnectionError as exc:
-            raise ClientConnectionError("Couldn't connect to evaluator") from exc
+        asyncio.run(publisher())
 
     def report(self, msg):
         self._statemachine.transition(msg)
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -19,6 +19,7 @@ from _ert.events import (
     EnsembleSucceeded,
     FMEvent,
     ForwardModelStepChecksum,
+    ForwardModelStepFailure,
     RealizationEvent,
     SnapshotInputEvent,
     dispatcher_event_from_json,
@@ -27,7 +28,9 @@ from _ert.forward_model_runner.client import (
     ACK_MSG,
     CONNECT_MSG,
     DISCONNECT_MSG,
+    TERMINATE_MSG,
 )
+from _ert.forward_model_runner.fm_dispatch import FORWARD_MODEL_TERMINATED_MSG
 from ert.ensemble_evaluator import identifiers as ids
 
 from ..config import QueueSystem
@@ -148,6 +151,7 @@ class EnsembleEvaluator:
                                 "Experiment cancelled by user during evaluation"
                             )
                         )
+                        self.stop()
 
             except TimeoutError:
                 if closetracker_received:  # THIS SHOULD NOT BE NEEDED ANYMORE
@@ -169,6 +173,26 @@ class EnsembleEvaluator:
                 logger.debug("Run model cancelled - during evaluation - cancel sent")
                 self._end_event.clear()
             await asyncio.sleep(0.1)
+
+    async def _send_terminate_message_to_dispatchers(self) -> None:
+        event = TERMINATE_MSG
+
+        await asyncio.gather(
+            *(
+                self._router_socket.send_multipart([identity, b"", event])
+                for identity in self._dispatchers_connected
+            )
+        )
+        if self._ensemble._scheduler is not None:
+            for identity in self._dispatchers_connected:
+                real_id = int(identity.decode("utf-8").split("-")[2])
+                self._ensemble._scheduler.mark_job_as_being_killed_by_evaluator(real_id)
+
+    async def _terminate_all_dispatchers(self) -> None:
+        if (scheduler := self.ensemble._scheduler) is not None:
+            await scheduler._running.wait()
+        await self._send_terminate_message_to_dispatchers()
+        await self._ensemble.cancel()
 
     async def _append_message(self, snapshot_update_event: EnsembleSnapshot) -> None:
         event = EESnapshotUpdate(
@@ -298,6 +322,15 @@ class EnsembleEvaluator:
                     f"Ignoring since I am {self.ensemble.id_}"
                 )
                 return
+            if (
+                type(event) is ForwardModelStepFailure
+                and event.error_msg == FORWARD_MODEL_TERMINATED_MSG
+                and self._ensemble._scheduler is not None
+            ):
+                self._ensemble._scheduler.confirm_job_killed_by_evaluator(
+                    int(event.real)
+                )
+
             if type(event) is ForwardModelStepChecksum:
                 await self.forward_checksum(event)
             else:
@@ -395,7 +428,8 @@ class EnsembleEvaluator:
             logger.debug("Cancelling current ensemble")
             self._ee_tasks.append(
                 asyncio.create_task(
-                    self._ensemble.cancel(), name="ensemble_cancellation_task"
+                    self._terminate_all_dispatchers(),
+                    name="dispatcher_termination_task",
                 )
             )
         else:
@@ -450,7 +484,7 @@ class EnsembleEvaluator:
                 elif task.get_name() in {
                     "ensemble_task",
                     "listener_task",
-                    "ensemble_cancellation_task",
+                    "dispatcher_termination_task",
                     "publisher_task",
                     "monitor_end_event_task",
                 }:
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -167,6 +167,7 @@ class EnsembleEvaluator:
                 logger.debug("Run model cancelled - during evaluation")
                 await self._signal_cancel()
                 logger.debug("Run model cancelled - during evaluation - cancel sent")
+                self._end_event.clear()
             await asyncio.sleep(0.1)
 
     async def _append_message(self, snapshot_update_event: EnsembleSnapshot) -> None:
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -67,6 +67,7 @@ class Job:
     """
 
     DEFAULT_FILE_VERIFICATION_TIMEOUT = 120
+    WAIT_PERIOD_FOR_TERM_MESSAGE_TO_CANCEL = 5
 
     def __init__(self, scheduler: Scheduler, real: Realization) -> None:
         self.real = real
@@ -81,6 +82,8 @@ class Job:
         self._end_time: float | None = None
         self.remaining_file_verification_time = self.DEFAULT_FILE_VERIFICATION_TIMEOUT
         self.remember_remaining_time()
+        self._started_killing_by_evaluator: bool = False
+        self._was_killed_by_evaluator = asyncio.Event()
 
     def remember_remaining_time(self) -> None:
         self.previous_file_verification_time = self.remaining_file_verification_time
@@ -154,7 +157,21 @@ class Job:
 
         except asyncio.CancelledError:
             await self._send(JobState.ABORTING)
-            await self.driver.kill(self.iens)
+            killed_by_evaluator = False
+            if self._started_killing_by_evaluator:
+                with suppress(asyncio.TimeoutError):
+                    killed_by_evaluator = await asyncio.wait_for(
+                        self._was_killed_by_evaluator.wait(),
+                        timeout=self.WAIT_PERIOD_FOR_TERM_MESSAGE_TO_CANCEL,
+                    )
+            if not killed_by_evaluator:
+                logger.warning(
+                    f"Realization {self.iens} was not killed gracefully by "
+                    "TERM message. Killing it with the scheduler"
+                )
+                await self.driver.kill(self.iens)
+            else:
+                logger.info(f"Realization {self.iens} was killed by the evaluator")
             with suppress(asyncio.CancelledError):
                 self.returncode.cancel()
             await self._send(JobState.ABORTED)
--- a/src/ert/scheduler/scheduler.py
+++ b/src/ert/scheduler/scheduler.py
@@ -130,6 +130,16 @@ class Scheduler:
         for task in pending:
             logger.debug(f"Task {task.get_name()} was not killed properly!")
 
+    def mark_job_as_being_killed_by_evaluator(self, real_id: int) -> None:
+        """Mark the job as being killed by the evaluator"""
+        self._jobs[real_id]._started_killing_by_evaluator = True
+
+    def confirm_job_killed_by_evaluator(self, real_id: int) -> None:
+        """Set the job as killed by the evaluator, if the flag for \
+            evaluator started killing job is already set."""
+        if self._jobs[real_id]._started_killing_by_evaluator:
+            self._jobs[int(real_id)]._was_killed_by_evaluator.set()
+
     async def _update_avg_job_runtime(self) -> None:
         while True:
             iens = await self.completed_jobs.get()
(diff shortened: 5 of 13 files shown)


jonathan-eq:
We should also test that we dont kill bkill for fm_dispatchers that already quit gracefully

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -97,41 +104,67 @@ def stop(self, exited_event: Exited | None = None):
...
+                    break
+                await client.send(event_to_json(event), self._max_retries)
+                event = None
+            except asyncio.CancelledError:
+                return
+            except ClientConnectionError as exc:
+                logger.error(f"Failed to send event: {exc}")
+                raise exc
+            except queue.Empty:
+                await asyncio.sleep(0)
```
jonathan-eq, line 131: This is needed to manually yield. We could use an asyncio.Queue, but that would require a lot more refactoring and making the fm_dispatcher async.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -57,6 +58,8 @@ class HeartbeatEvent(Enum):
 
 
 class EnsembleEvaluator:
+    WAIT_PERIOD_FOR_GRACEFUL_SHUTDOWN = 5.0
```
jonathan-eq: I am not sure what we should set this to. I think 5s should be a good starting point. If it works, it will use no additional files (as we already have the zmq connection in place), whereas running scheduler.killAllJobs uses two files per realization.

src/_ert/forward_model_runner/client.py
```diff
@@ -32,6 +33,7 @@ def __init__(
         token: str | None = None,
         dealer_name: str | None = None,
         ack_timeout: float | None = None,
+        store_terminate_messages: bool = False,
```
berland: why is this an option? Is there a downside to always storing them? (and can there be more than one such message for each forward_model_runner?

jonathan-eq: The client is re-used in monitor, and we will never get the terminated events there. Maybe it will be fine to have it always set to True.

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -97,41 +104,66 @@ def stop(self, exited_event: Exited | None = None):
...
+
+    async def handle_listen(self, client: Client):
+        while True:
+            try:
+                event = await client.received_messages.get()
+                if event == TERMINATE_MSG:
+                    pgid = os.getpgid(os.getpid())
+                    os.killpg(pgid, signal.SIGTERM)
+                    return
```
berland: (this line (`return`) is probably never reached as it should be killed before it happens ?)

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -97,41 +104,66 @@ def stop(self, exited_event: Exited | None = None):
...
+                logger.error(f"Failed to send event: {exc}")
+                raise exc
+            except queue.Empty:
+                await asyncio.sleep(0)
+
+    async def handle_listen(self, client: Client):
+        while True:
+            try:
+                event = await client.received_messages.get()
+                if event == TERMINATE_MSG:
```
berland: This event should probably be logged.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -105,6 +108,20 @@ async def _publisher(self) -> None:
...
+            *(
+                self._router_socket.send_multipart([identity, b"", event])
+                for identity in self._dispatchers_connected
+            )
+        )
+
+    async def _terminate_all_dispatchers(self) -> None:
+        await self._send_terminate_messages_to_dispatchers()
+        await asyncio.sleep(self.WAIT_PERIOD_FOR_GRACEFUL_SHUTDOWN)
+        await self._ensemble.cancel()
```
berland, line 195: Is it possible to log here which realizations that need an extra kill?

src/_ert/forward_model_runner/reporting/event.py
jonathan-eq: We can probably change this to an asyncio.Event() instead of a queue, and have the task just await the event instead of the queue. We would not need the loop as we only expect one message at most.

src/_ert/forward_model_runner/client.py
```diff
@@ -157,7 +161,6 @@ async def send(self, message: str | bytes, retries: int | None = None) -> None:
                     f"reconnecting: {exc}"
                 )
             except asyncio.CancelledError:
-                self.term()
```
jonathan-eq, line 160: This caused issues as we would terminate the zmq client both here, and then again when we were sending the final DISCONNECT message.

src/ert/ensemble_evaluator/evaluator.py
berland: If all dispatchers happily terminate before the grace period, we should not have to wait for the grace period.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -105,6 +106,25 @@ async def _publisher(self) -> None:
...
+        )
+
+    async def _terminate_all_dispatchers(self) -> None:
+        if self.ensemble._scheduler is not None:
+            await self.ensemble._scheduler._running.wait()
+            self.ensemble._scheduler._cancelled = True
+        await self._send_terminate_messages_to_dispatchers()
+        if self.ensemble._scheduler is not None:
+            for scheduler_job in self._ensemble._scheduler._jobs.values():
+                scheduler_job.returncode.cancel()
```
jonathan-eq: The best here would be to only cancel the ones that ACK the message. That requires the refactoring outlined in https://github.com/equinor/ert/issues/10354. Ensemble Evaluator has NO clue what dispatcher is connected to which realization.

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: Yes, it is fire and forget in an asyncio.Task

jonathan-eq:
This will log 
```
2025-07-15 09:10:20,873 - ert.scheduler.lsf_driver - ert_gui_simulation_thread - INFO - Realization 3 (LSF-id: 926334) failed
```
when it has been killed by evaluator. I am not sure if it does the same when killed by bkill, but it should probably output `Realization was X was killed` instead.

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -97,43 +104,64 @@ def stop(self, exited_event: Exited | None = None):
...
+        except asyncio.CancelledError:
+            return
+
     def _event_publisher(self):
         async def publisher():
             async with Client(
                 url=self._evaluator_url,
                 token=self._token,
                 ack_timeout=self._ack_timeout,
+                dealer_name=f"dispatch-iens-{self._real_id}-{uuid.uuid4().hex[:6]}",
```
jonathan-eq: We add the realization being run into the dealer name, so the evaluator knows which "realization" it is communicating with.

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -97,43 +104,64 @@ def stop(self, exited_event: Exited | None = None):
...
-                    except asyncio.CancelledError:
-                        return
-                    except ClientConnectionError as exc:
-                        logger.error(f"Failed to send event: {exc}")
-                        raise exc
+                publisher_task = asyncio.create_task(
+                    self.handle_publish(client), name="publisher_task"
+                )
+                listener_task = asyncio.create_task(
+                    self.handle_listen(client), name="listener_task"
```
jonathan-eq: This is a new task that is listening for messages from the evaluator. Before this, it was only one-way communication.

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: I added a mechanism for this, and we will log what realizations were killed by the evaluator (via TERMINATE_MSG) and which ones were killed by the scheduler (via driver.kill).

tests/ert/unit_tests/scheduler/test_scheduler.py
```diff
@@ -308,7 +308,7 @@ async def kill():
         await asyncio.sleep(1)
 
     realization.max_runtime = 1
-
+    monkeypatch.setattr(Job, "WAIT_PERIOD_FOR_TERM_MESSAGE_TO_CANCEL", 0)
```
jonathan-eq: This is now necessary, otherwise it will wait here for 5 seconds. Maybe there is a better way to do this?

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -298,6 +328,14 @@ async def handle_dispatch(self, dealer: bytes, frame: bytes) -> None:
...
                 )
                 return
+            if (
+                type(event) is ForwardModelStepFailure
+                and event.error_msg == FORWARD_MODEL_TERMINATED_MSG
+                and self._ensemble._scheduler is not None
+            ):
+                self._ensemble._scheduler._jobs[
+                    int(event.real)
+                ]._was_killed_by_evaluator.set()
```
jonathan-eq: This does not have to be due to the TERMINATE_MSG. This can also be by MAX_RUNTIME, and that will cause issues.

jonathan-eq:
> This has a problem with killing by MAX_RUNTIME.

Fixed the problem, but killing with MAX_RUNTIME will not use the TERMINATE message.

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -97,43 +104,64 @@ def stop(self, exited_event: Exited | None = None):
...
+            except asyncio.CancelledError:
+                return
+            except ClientConnectionError as exc:
+                logger.error(f"Failed to send event: {exc}")
+                raise exc
+            except queue.Empty:
+                await asyncio.sleep(0)
+
+    async def handle_listen(self, client: Client):
+        try:
```
yngve-sk, line 134: This is now only used to wait for a termination signal that may never come? Naming suggests it might listen for more incoming events?

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -97,43 +104,64 @@ def stop(self, exited_event: Exited | None = None):
...
+                listener_task = asyncio.create_task(
+                    self.handle_listen(client), name="listener_task"
+                )
+                await publisher_task
+                if not listener_task.done():
+                    listener_task.cancel()
+                    await listener_task
 
-        try:
+        with contextlib.suppress(ClientConnectionError):
```
yngve-sk: Why suppress here?

src/_ert/forward_model_runner/reporting/event.py
jonathan-eq: Yes, I will rename it `listen_for_terminate_message`.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -46,6 +49,11 @@
 EVENT_HANDLER = Callable[[list[SnapshotInputEvent]], Awaitable[None]]
 
 
+def _get_iens_from_dealer_name(dealer_name: str) -> int:
```
yngve-sk: style pref: Would prefer to call a realization nr real/realization instead of iens (widely used in ERT from old times but a bit ambiguous imo). Also 1 use so would consider inlining with comment / moving the fn to where it is used

src/_ert/forward_model_runner/reporting/event.py
jonathan-eq: We have already logged the exception, but we never retrieve it from the thread running publisher. We might as well not re-raise it, so I can safely remove the suppress. Will fix 🔧

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -298,6 +334,17 @@ async def handle_dispatch(self, dealer: bytes, frame: bytes) -> None:
                     f"Ignoring since I am {self.ensemble.id_}"
                 )
                 return
+            if (
+                type(event) is ForwardModelStepFailure
+                and event.error_msg == FORWARD_MODEL_TERMINATED_MSG
+                and self._ensemble._scheduler is not None
+                and self._ensemble._scheduler._jobs[
```
yngve-sk: I think you mentioned this already but it does look a bit like some of this logic should be in `LegacyEnsemble` or `Scheduler` from reading this

jonathan-eq: Yes, all responsibility for the fm_dispatch communication should be moved to scheduler, but there have been some discussion over how this would affect the experiment server milestone. Maybe the scheduler should also be on the experiment server 🤔

src/ert/scheduler/job.py
```diff
@@ -154,7 +157,21 @@ async def _submit_and_run_once(self, sem: asyncio.BoundedSemaphore) -> None:
 
         except asyncio.CancelledError:
             await self._send(JobState.ABORTING)
-            await self.driver.kill(self.iens)
+            killed_by_evaluator = False
+            if self._started_killing_by_evaluator:
+                with suppress(asyncio.TimeoutError):
+                    killed_by_evaluator = await asyncio.wait_for(
```
yngve-sk, line 163: Hm. Do we really need `self._started_killing_by_evaluator` as well as the event stored as attrs? Would it be possible to just keep an event to `await` that will return whether it was interrupted?

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: I also think the scheduler should live in evaluator, not in the ensemble

src/ert/scheduler/job.py
jonathan-eq: I tried that, but there is a code path for jobs being killed (MAX_RUNTIME) that does not through the evaluator. This will cause that to unnecessarily hang for 5s.
I think there was another reason too, where jobs could be cancelled without the user calling terminate

jonathan-eq:
This is the end of terminate experiment on main, but we are not getting the same on this PR. Will look into it. 
image

jonathan-eq:
## Results using `watch -n 0.1 "lsof -c ert | wc -l"`:

| Realizations | 150 | 200 |
| ------------- | ------------- |------------- |
| Main (open files) | 895 | 1019 |
| PR (open files) | 603 | 657 |

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -97,43 +103,64 @@ def stop(self, exited_event: Exited | None = None):
...
+                await publisher_task
+                if not listener_task.done():
+                    listener_task.cancel()
+                    await listener_task
 
-        try:
-            asyncio.run(publisher())
-        except ClientConnectionError as exc:
-            raise ClientConnectionError("Couldn't connect to evaluator") from exc
+        asyncio.run(publisher())
```
xjules, line 163: How does this handle `ClientConnectionError`?

src/ert/ensemble_evaluator/evaluator.py
xjules: I agree this should be `scheduler.kill_job(real: int)` or something similar.

src/_ert/forward_model_runner/reporting/event.py
jonathan-eq: It will never get it, it is already handled upstream. If `self.handle_publish()` gets `ClientConnectionError` it logs the error, then quits.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -395,7 +436,8 @@ async def _signal_cancel(self) -> None:
             logger.debug("Cancelling current ensemble")
             self._ee_tasks.append(
                 asyncio.create_task(
-                    self._ensemble.cancel(), name="ensemble_cancellation_task"
+                    self._terminate_all_dispatchers(),
```
xjules, line 431: So I guess `ensemble.cancel()` has not been removed?

jonathan-eq: No, we are still using it inside of `self._terminate_all_dispatchers`

src/ert/scheduler/scheduler.py
```diff
@@ -119,9 +120,14 @@ async def cancel_all_jobs(self) -> None:
         await self._cancel_job_tasks()
 
     async def _cancel_job_tasks(self) -> None:
-        for task in self._job_tasks.values():
+        for iens, task in self._job_tasks.items():
             if not task.done():
                 task.cancel()
+            else:
+                logger.info(
+                    f"Realization {iens} stopped itself after getting the "
```
xjules: Is this message correct? This looks like the task is not done yet.

src/ert/scheduler/scheduler.py
```diff
@@ -96,6 +96,7 @@ def __init__(
         self.warnings_extracted: bool = False
 
         self._cancelled = False
+        self._cancelled_by_evaluator = False
```
xjules: Why do we need two variables here `_cancelled` and `_cancelled_by_evaluator`?

tests/ert/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -281,135 +321,293 @@ def test_overspent_cpu_is_logged(
...
+
+@pytest.mark.timeout(15)
+async def test_signal_cancel_sends_terminate_message_to_dispatchers(evaluator_to_use2):
+    (evaluator, _) = evaluator_to_use2
+    token = evaluator._config.token
+    url = evaluator._config.get_uri()
+    evaluator.ensemble._cancellable = True
+    async with (
+        Client(
+            url, token=token, dealer_name=f"dispatch-real-0-{uuid.uuid4().hex[:6]}"
```
xjules, line 457: is it important to provide dealer_name?

tests/ert/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -171,86 +182,115 @@ def event_handler(snapshot):
...
+    evaluator._batching_interval = 0.5  # batching can be faster for tests
+    run_task = asyncio.create_task(evaluator.run_and_get_successful_realizations())
+    await evaluator._server_started
+    yield evaluator
+    evaluator.stop()
+    await run_task
+
+
+@pytest.fixture(name="evaluator_to_use2")
+async def evaluator_to_use_fixture2(make_ee_config):
```
xjules: This is not a great name for the fixture; ie. `evaluator_to_use2`. Maybe something `..._with_event_queue` or something?

src/ert/scheduler/scheduler.py
jonathan-eq: Yes, it is already done, but now I am wondering if this will include the job tasks that were already done before we started cancelling (the ones that finished successfully), or if those are already removed upon their completion. Will check :)

tests/ert/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
jonathan-eq: They are doing exactly the same, but the new one is also yielding the event_queue. I will remove the old one, and fixup the tests using it.

tests/ert/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
jonathan-eq: Yes, the evaluator expects it, otherwise it will fail hard.

src/ert/scheduler/scheduler.py
jonathan-eq: Hmmm, maybe not. I will double check. There was a reason for it, but maybe it got changed 🤔

tests/ert/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
jonathan-eq: The old was not in use anymore :)
There is another evaluator_to_use in conftest. I will see if I can merge all three fixtures.

jonathan-eq: The fixture used in test_ensemble_legacy causes hanging if merged with the other one. That is probably the reason they are two different ones. I will try merging them in a different PR.

src/ert/ensemble_evaluator/evaluator.py
xjules: maybe `scheduler.confirm_job_killed(real: int)` ?

jonathan-eq: Hmm, but that doesnt tell us if it was killed by scheduler or evaluator

src/ert/scheduler/job.py
```diff
@@ -154,7 +157,22 @@ async def _submit_and_run_once(self, sem: asyncio.BoundedSemaphore) -> None:
...
+                        self._was_killed_by_evaluator.wait(),
+                        timeout=self.WAIT_PERIOD_FOR_TERM_MESSAGE_TO_CANCEL,
+                    )
+
+            if not killed_by_evaluator:
+                logger.warning(
+                    f"Realization {self.iens} was not killed by the evaluator. "
+                    "Killing it with the scheduler"
```
xjules: `Realization {self.iens} was not killed gracefully by TERM message, executing driver kill.` ?

src/ert/scheduler/scheduler.py
```diff
@@ -114,14 +114,14 @@ async def kill_all_jobs(self) -> None:
...
         await self._running.wait()
-        self._cancelled = True
         logger.info("Cancelling all jobs")
         await self._cancel_job_tasks()
 
     async def _cancel_job_tasks(self) -> None:
         for task in self._job_tasks.values():
             if not task.done():
                 task.cancel()
+
```
xjules: extra empty line

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -480,6 +524,7 @@ async def run_and_get_successful_realizations(self) -> list[int]:
         try:
             await self._monitor_and_handle_tasks()
         finally:
+            print("IN FINALLY ")
```
xjules: leftover

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -87,6 +90,7 @@ def __init__(
         self._complete_batch: asyncio.Event = asyncio.Event()
         self._server_started: asyncio.Future[None] = asyncio.Future()
         self._dispatchers_connected: set[bytes] = set()
+        self._dispatcher_identity_to_real: dict[bytes, int] = {}
```
xjules: I don't think that this extra variable is needed, since we loop over all the jobs regardless.

jonathan-eq: It only loops over the jobs that are still connected to EE. This is to avoid setting the flag for jobs that have finished

src/ert/scheduler/job.py
jonathan-eq: The driver already logs when it kills jobs, but I can update the message for killed by evaluator. I am not sure if we need to mention the details of how it was killed, maybe it is already fine, or should we specify that it was with TERM message?

src/ert/ensemble_evaluator/evaluator.py
xjules: `self._dispatchers_connected:` this contains the set of the connected dispatchers only.

jonathan-eq: Yes, I will change it to reuse that instead. Very good idea!

src/ert/scheduler/job.py
xjules: I meant only to suggest a better rewording.

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -97,43 +103,64 @@ def stop(self, exited_event: Exited | None = None):
...
+        except asyncio.CancelledError:
+            return
+
     def _event_publisher(self):
         async def publisher():
             async with Client(
                 url=self._evaluator_url,
                 token=self._token,
                 ack_timeout=self._ack_timeout,
+                dealer_name=f"dispatch-real-{self._real_id}-{uuid.uuid4().hex[:6]}",
```
xjules, line 149: I guess we don't need this when this's been merged: https://github.com/equinor/ert/pull/11402 right?

jonathan-eq: We will need both. #11402 fixes a different issue

src/ert/scheduler/job.py
jonathan-eq: I updated it some. Maybe it is a bit better

src/ert/scheduler/scheduler.py
jonathan-eq: After some discussion, we could either have _cancelled_by_evaluator to kill all realizations, but if we keep it the way with `killing_with_evaluator` and `killed_by_evaluator` flags on each job, we can more easily kill specific realizations in the future. I think we decided to remove the `cancelled_by_evaluator` scheduler flag, and keep the job specific flags instead 👌

src/ert/scheduler/scheduler.py
jonathan-eq: It is logged another place, so we can remove it here :)

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: We should move `.evaluate(...)` to evaluator instead of LegacyEnsemble in the future, but it will be as is for now :)

tests/ert/unit_tests/scheduler/test_scheduler.py
xjules: TBH I don't understand why  `monkeypatch.setattr(Job, "WAIT_PERIOD_FOR_TERM_MESSAGE_TO_CANCEL", 0)` is needed in all the tests? I thought that this is only relevant when `killed_by_evaluator` is set?

jonathan-eq: You are right, this is from a previous iteration of the PR. Good catch!

src/_ert/forward_model_runner/reporting/event.py
xjules: Yes, I see. Esspecially when we want to implement the single realization kill :+1:

xjules (approved):
Great job @jonathan-eq ! Just remember to squash the last two commits. :rocket:

PR #11286, opened 2025-07-02
Remove refcase from SummaryConfig

yngve-sk:
Seems unused. `*H` vectors still plot for snake oil in plotter as usual. `Refcase` is only used for observations parsing, and is therefore completely irrelevant after that.

### DIFF
--- a/docs/ert/reference/configuration/keywords.rst
+++ b/docs/ert/reference/configuration/keywords.rst
@@ -1449,13 +1449,8 @@ field property, such as FOPT, WGRNAME need not be set to FIELD.
         -- Using the SUMMARY keyword to add diagnostic variables
         SUMMARY WOPR:MY_WELL
         SUMMARY RPR:8
-        SUMMARY F*          -- Use of wildcards requires that you have entered a REFCASE.
-
-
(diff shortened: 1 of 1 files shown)


xjules (approved):
:scissors:

src/ert/storage/local_experiment.py
```diff
@@ -353,6 +353,8 @@ def response_configuration(self) -> dict[str, ResponseConfig]:
         responses = {}
         for data in self.response_info.values():
             ert_kind = data.pop("_ert_kind")
+            data.pop("refcase", None)
+
```
berland, line 357: Is it possible to leave a code comment on why this is popped? Is it for legacy data on disk - meaning it can be removed after a long grace period?

JHolba: It seems to be the identifier for which kind of response is stored. We use it to decide which constructor to use. Maybe ert_kind isn't the best name for it :)

yngve-sk: Created issue for removal https://github.com/equinor/ert/issues/11289 , will be removed in next storage migration

Issue #11204, opened 2025-06-24
Dark storage raises when trying to unpack parameter groups and keys


andreas-el:
In komodo version `2025.06.02`;

```
 File "/prog/res.../ert/dark_storage/common.py", line 267, in data_for_key
    parameter_group, parameter_key = _extract_parameter_group_and_key(key)
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/prog/res.../ert/dark_storage/common.py", line 126, in _extract_parameter_group_and_key
    param_group, param_key = key.split(":")
    ^^^^^^^^^^^^^^^^^^^^^^
ValueError: too many values to unpack (expected 2)
 from url: http://127.0.0.1:51856/ensembles/9a369b20-f81b-44e6-9cc0-5a5c2eca5a17/records/WOPR%253A%253A%252B%253A%252B%253A%252B%253A%252B.
```

Trace:
https://fmu-dashboard-fdeqekb8chekcqhu.neu.grafana.azure.com/d/ceauatgtsa9s0a/traceview?orgId=1&from=now-30d&to=now&var-operation_id=46cacec974c8df97af191eff5732ad92

yngve-sk:
I'm assuming this bug is due to the dark storage refactor + a user having a parameter key with `:` in it. Quickfix would be to split by first hit, a bit better fix would be to remove the `:` prefixing alltogether (in spite of this being the old behavior), but I think that will come out more nicely once @xjules is done with the "removing groups" part. Removing the `:` split from webviz-ert would also have to be done here.

PR #11298, opened 2025-07-03
Handle multiple colons in dark storage keys

yngve-sk:
Resolves [#11204](https://github.com/equinor/ert/issues/11204)

### DIFF
--- a/src/ert/gui/tools/plot/plot_api.py
+++ b/src/ert/gui/tools/plot/plot_api.py
@@ -306,7 +306,7 @@ class PlotApi:
 
     def _history_key(self, key: str) -> str:
         if ":" in key:
-            head, tail = key.split(":", 2)
+            head, tail = key.split(":", 1)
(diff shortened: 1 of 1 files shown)


tests/ert/unit_tests/gui/tools/plot/test_plot_api.py
```diff
@@ -329,6 +330,41 @@ def test_plot_api_handles_non_existant_gen_kw(api_and_storage):
...
+    api, storage = api_and_storage
+    experiment = storage.create_experiment(
+        parameters=[
+            GenKwConfig(
+                name="group",
+                forward_init=False,
+                update=False,
+                transform_function_definitions=[
+                    TransformFunctionDefinition(
+                        name="subgroup:1:2:2", param_name="NORMAL", values=[0, 0.1]
```
xjules: maybe using `RAW` distribution would be enough and then the expected value will be the same as stored.

Issue #11289, opened 2025-07-02
Remove `refcase` conditional in `LocalExperiment`


yngve-sk:
It is currently there to ignore the `refcase` when loading old summary configs with it still included. This issue shall be closed when adding the next storage migration, wherein `refcase` shall be removed from every summary config in storage.

PR #11306, opened 2025-07-04
Use pydantic for responses and parameters

yngve-sk:
Resolves https://github.com/equinor/ert/issues/10976
Resolves https://github.com/equinor/ert/issues/10977
Resolves https://github.com/equinor/ert/issues/11289

### DIFF
--- a/src/ert/analysis/_es_update.py
+++ b/src/ert/analysis/_es_update.py
@@ -378,10 +378,7 @@ def analysis_ES(
 
                 if cross_correlations:
                     assert isinstance(config_node, GenKwConfig)
-                    parameter_names = [
-                        t["name"]  # type: ignore
-                        for t in config_node.transform_function_definitions
-                    ]
+                    parameter_names = config_node.parameter_keys
                     cross_correlations_ = np.vstack(cross_correlations)
                     if cross_correlations_.size != 0:
                         source_ensemble.save_cross_correlations(
--- a/src/ert/config/field.py
+++ b/src/ert/config/field.py
@@ -5,12 +5,12 @@ import logging
 import os
 from functools import cached_property
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Self, cast, overload
+from typing import TYPE_CHECKING, Any, Literal, Self, cast, overload
 
 import networkx as nx
 import numpy as np
 import xarray as xr
-from pydantic.dataclasses import dataclass
+from pydantic import field_serializer
 
 from ert.field_utils import FieldFileFormat, Shape, read_field, read_mask, save_field
 from ert.substitutions import substitute_runpath_name
@@ -80,8 +80,8 @@ def adjust_graph_for_masking(
     return G
 
 
-@dataclass
 class Field(ParameterConfig):
+    type: Literal["field"] = "field"
     nx: int
     ny: int
     nz: int
@@ -94,6 +94,14 @@ class Field(ParameterConfig):
     output_file: Path
     grid_file: str
     mask_file: Path | None = None
+
+    @field_serializer("output_file")
+    def serialize_output_file(self, path: Path) -> str:
+        return str(path)
+
+    @field_serializer("mask_file")
+    def serialize_mask_file(self, path: Path | None) -> str | None:
+        return str(path) if path is not None else None
 
     @property
     def parameter_keys(self) -> list[str]:
--- a/src/ert/config/gen_kw_config.py
+++ b/src/ert/config/gen_kw_config.py
@@ -3,17 +3,17 @@ from __future__ import annotations
 import math
 import os
 from collections.abc import Callable
-from dataclasses import asdict, dataclass
+from dataclasses import dataclass
 from hashlib import sha256
 from pathlib import Path
-from typing import TYPE_CHECKING, Annotated, Any, Self, cast, overload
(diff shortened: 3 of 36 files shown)


src/ert/analysis/_es_update.py
```diff
@@ -379,8 +379,7 @@ def correlation_callback(
                 if cross_correlations:
                     assert isinstance(config_node, GenKwConfig)
                     parameter_names = [
```
xjules: `config_node.parameter_keys`

src/ert/config/gen_kw_config.py
```diff
@@ -91,34 +90,54 @@ class TransformFunction:
...
 
 
-@dataclass
 class GenKwConfig(ParameterConfig):
+    type: Literal["gen_kw"] = "gen_kw"
     transform_function_definitions: list[TransformFunctionDefinition]
 
-    def __post_init__(self) -> None:
-        self.transform_functions: list[TransformFunction] = []
+    _transform_functions: list[TransformFunction] = PrivateAttr()
```
xjules, line 100: Just wondering if this can be made into @cached_property instead? Maybe it is ok though.
```python
@cached_property
def transform_functions(self):
     # parsing the tfd

```

src/ert/config/gen_kw_config.py
```diff
@@ -514,18 +531,12 @@ def _parse_transform_function_definition(
...
-                    ctx["error"].errors += f" parameter {t.name}"
-                else:
-                    for err in custom_error:
-                        err.set_context(self.name)
-                        err.message += f" parameter {t.name}"
+            error_to_raise = ConfigValidationError.from_pydantic(
+                error=e, context=self.name
+            )
+            for error_info in error_to_raise.errors:
+                error_info.message += f" parameter {t.name}"
```
xjules, line 541: nice!

src/ert/config/gen_kw_config.py
yngve-sk: It could be and I intended to originally, but we need to trigger the parsing early, to get configvalidation errors to GUI etc, so then I think it is more clean to just do this than to have a cached_property and trigger it, to trigger the validations

src/ert/storage/local_experiment.py
```diff
@@ -128,7 +152,7 @@ def create(
         parameter_data = {}
         for parameter in parameters or []:
             parameter.save_experiment_data(path)
-            parameter_data.update({parameter.name: parameter.to_dict()})
+            parameter_data.update({parameter.name: parameter.model_dump(mode="json")})
```
xjules, line 155: since `to_dict()` was replaced by pydantic.model_dump why are we still keeping `to_dict()` functions :top: ?

src/ert/config/field.py
```diff
@@ -95,6 +95,14 @@ class Field(ParameterConfig):
     grid_file: str
```
xjules, line 95: Not sure about grid file. Does it always need to be specified? It is either the global one or as a named attribute for FIELD - this is actually WIP though.

src/ert/storage/local_experiment.py
yngve-sk: Hmm, can actually be removed by just moving the logic to migrations (only place it is needed anyway)

src/ert/config/gen_kw_config.py
xjules: but you can trigger it in `model_post_init` just by calling 
`_ = self.transform_functions`

src/ert/config/field.py
yngve-sk: It is (currently) non-optional: Ref
```
...
            if grid_file_path is None:
                raise ConfigValidationError.with_context(
                    "In order to use the FIELD keyword, a GRID must be supplied.",
                    field_list,
                )
   ...
         return FieldConfig.from_config_list(grid_file_path, dims, field_list)
```

src/ert/config/gen_kw_config.py
yngve-sk: I described this in my answer above, still think it is a bit more implicit to do it that way w/ a cached_property

xjules: Agreed.

src/ert/config/field.py
xjules: I see then it's fine 👍

src/ert/config/gen_kw_config.py
```diff
@@ -91,34 +90,54 @@ class TransformFunction:
...
-@dataclass
 class GenKwConfig(ParameterConfig):
+    type: Literal["gen_kw"] = "gen_kw"
     transform_function_definitions: list[TransformFunctionDefinition]
 
-    def __post_init__(self) -> None:
-        self.transform_functions: list[TransformFunction] = []
+    _transform_functions: list[TransformFunction] = PrivateAttr()
+
+    def model_post_init(self, ctx: Any) -> None:
```
xjules: This will not work according mr. gpt!

xjules: Suggestion is to use this:
```python
def build_transform_functions(self) -> Self:
...
```

xjules: Clarified via testing and the exceptions are propagated correctly.

xjules (approved):
Nice work @yngve-sk ! Please squash the commits / rebase with some nice commit messages describing the overall goal potentially.

PR #11316, opened 2025-07-04
Leave note in everest test-data

### LINKED ISSUE #11301: Everest is not able to deal with all initial guesses in math_func/config_minimal.yml
Experimenting with math_func/config_minimal.yml, I see that Everest has severe difficulties converging unless the initial guess is right.

I would expect that a simple 3D distance function is possible to optimize within the [-1,1] cube, and that any initial guess inside this box will be sufficient for Everest to locate the correct point (0.5, 0.5, 0.5). 

If this problem is to difficult, an ever simpler problem should be presented as "minimal". 

However, setting the initial guess negative, typically the `y` point will jump to 1, and then that control point cannot converge further. 

Also setting the initial point to (1,1,1) will make it fail converging. 

Example diff:
```
$ git diff
diff --git a/test-data/everest/math_func/config_minimal.yml b/test-data/everest/math_func/config_minimal.yml
index d170c401d3..e7cc3692af 100644
--- a/test-data/everest/math_func/config_minimal.yml
+++ b/test-data/everest/math_func/config_minimal.yml
@@ -4,7 +4,7 @@ controls:
     min: -1.0
     max: 1.0
     scaled_range: [-1, 1]
-    initial_guess: 0.1
+    initial_guess: -0.1
     perturbation_magnitude: 0.001
     variables:
       - name: x
@@ -12,6 +12,11 @@ controls:
       - name: z
```
then

`$ everest run --new-run config_minimal.yml `

```
gives:
  point.x: 0.99722
  point.y: 1
  point.z: 0.99397

  distance: -0.74124
```

berland:
**Issue**
Resolves #11301 

**Approach**
📜 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/test-data/everest/math_func/config_minimal.yml
+++ b/test-data/everest/math_func/config_minimal.yml
@@ -19,7 +19,7 @@ objective_functions:
 optimization:
   algorithm: optpp_q_newton
   convergence_tolerance: 0.001
-  max_batch_num: 4
+  max_batch_num: 4  # Remove this to let Everest converge
   constraint_tolerance: 0.1  # To test that it is ignored.
 
 


berland:
One sample timing test indidates that this adds 20 seconds to the test suite (`time just everest-tests`). Should check if this parameter should be reinjected dynamically if the tests do not require convergence.

PR #11323, opened 2025-07-07
Remove unnecessary mocked everest test cases&files

yngve-sk:


tests/everest/test_controls.py
```diff
@@ -13,49 +12,24 @@
...
-    ):
-        EverestConfig.model_validate(config.to_dict())
 
-    config.controls[1].name = exp_grp_name + "_new"
-    EverestConfig.model_validate(config.to_dict())
+    with pytest.raises(ValidationError, match=r"Subfield\(s\) `name` must be unique"):
+        EverestConfig.model_validate(min_config)
+
+    min_config["controls"][1]["name"] = existing_name + "_new"
+    EverestConfig.model_validate(min_config)
```
verveerpj: I know this was already in the test before, but do we really need that?

tests/everest/test_controls.py
```diff
@@ -183,41 +157,38 @@ def test_control_variable_initial_guess_above_max():
...
-        EverestConfig.model_validate(config.to_dict())
+        ControlConfig.model_validate(min_config["controls"][0])
 
+
+def test_control_variable_weird_names(min_config):
+    """We would potentially like to support variable names with
+    underscores, but currently Seba is using this as a separator between
+    the group name and the variable name in such a way that having an
+    underscore in a variable name will not behave nicely..
+    """
```
verveerpj: This is not true anymore. If `Everest` supports underscores in the name, `ropt` should have not issue with it.

tests/everest/test_controls.py
```diff
@@ -183,41 +157,38 @@ def test_control_variable_initial_guess_above_max():
...
     weirdo_name = "something/with-symbols_=/()*&%$#!"
-    new_config = EverestConfig.load_file(mocked_config)
+    new_config = EverestConfig.model_validate(min_config)
```
verveerpj, line 169: We support this? That is a questionable decision...

tests/everest/test_input_constraints_config.py
```diff
@@ -3,35 +3,6 @@
 
 from ert.config.parsing.config_errors import ConfigWarning
 from everest.config import EverestConfig
-from tests.everest.utils import relpath
-
-
-def test_input_constraint_initialization():
```
verveerpj, line 9: Is this test covered elsewhere?

verveerpj (approved):
Only some small comments

tests/everest/test_input_constraints_config.py
yngve-sk: Hmm, did I maybe miss something here? As far as I could tell, I think this test is a bit tautological, in that it tests that things that exist in the config file also end up in the config. Now the parsing from yml to pydantic is pretty direct, so this would be tested by any test that includes input constraints with upper/lower bounds, and also with some weights.

tests/everest/test_controls.py
yngve-sk: Not really, will remove. Tested by all tests involving a valid controls config without duplicate names

tests/everest/test_controls.py
yngve-sk: Removing the comment as it is deprecated

tests/everest/test_controls.py
yngve-sk: I guess we do, but as we don't have any validation that disallows this I will leave it, could have an issue for making out lives easier by making it a bit more strict though?

verveerpj: Yes, it might be a good idea to discuss this at least.

tests/everest/test_input_constraints_config.py
verveerpj: I agree, we don't really need this test.

tests/everest/test_controls.py
yngve-sk: Set up an issue: https://github.com/equinor/ert/issues/11326

Issue #11219, opened 2025-06-25
ControlVariableConfig.index is confusing/not clear


yngve-sk:
From reading docs and reading code it is not so easy to see exactly what this "index" is other than a label. I expected it to mean something for sorting / correspondence with giving a `ControlVariableGuessList`. For example:

```
ControlConfig(
        name="well_priorities",
        type="well_control",
        variables=[
            {"name": "WELL-1", "initial_guess": 0.54, "index": 2},
            {"name": "WELL-1", "initial_guess": 0.58, "index": 1},
            {"name": "WELL-1", "initial_guess": 0.5, "index": 3},
            {"name": "WELL-2", "initial_guess": 0.58, "index": 2},
            {"name": "WELL-2", "initial_guess": 0.5, "index": 1},
            {"name": "WELL-2", "initial_guess": 0.56, "index": 3},
            {"name": "WELL-3", "initial_guess": 0.52, "index": 2},
            {"name": "WELL-3", "initial_guess": 0.56, "index": 1},
            {"name": "WELL-3", "initial_guess": 0.58, "index": 3},
        ],
```

Would make it seem that the `"index"` would be "equivalent" to this:
```
 ControlConfig(
        name="well_priorities",
        type="well_control",
        variables=[
            {"name": "WELL-1", "initial_guess": [0.58, 0.54, 0.5]},
            {"name": "WELL-2", "initial_guess": [0.5, 0.58, 0.56]},
            {"name": "WELL-3", "initial_guess": [0.56, 0.52, 0.58]},
        ],
```

Other than that, having an explicit `index` makes it short up as a suffix for the formatted control name (used in ERT storage and probably for plotting etc), this should be revisited/potentially simplified.

PR #11324, opened 2025-07-07
Fix FlattenedControls

### LINKED ISSUE #11321: Everest seems to ignore `enabled: False` in the config

When using this in the config file:

```yaml
controls:
- name: my_controls
  type: generic_control
  min: 0
  max: 259
  control_type: integer
  variables:
    - name: indexes
      initial_guess: [1,2,3]
    - name: number
      min: 0
      max: 3 
      initial_guess: [2]
      enabled: False
```
I could still see the control values of `my_controls.number` changing across the batches and perturbations while it should remain the same. 

_No response_

### Steps to reproduce

```Python

```

- [ ] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

verveerpj:
**Issue**
Resolves #11321

**Approach**
The original code was buggy, since it did not handle all fields in the case the variable field has class ControlVariableGuessListConfig . The result was that some fields would fall back to the values in the control
section, or to defaults. The code has been rewritten such that the parsing of all fields is shared by the ControlVariableGuessListConfig and ControlVariableConfig cases. A test to check this has been added.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/config/utils.py
+++ b/src/everest/config/utils.py
@@ -1,124 +1,92 @@
-from collections.abc import Generator
+from copy import deepcopy
 from typing import Any, Literal
 
 from .control_config import ControlConfig
-from .control_variable_config import (
-    ControlVariableConfig,
-    ControlVariableGuessListConfig,
-)
+from .control_variable_config import ControlVariableGuessListConfig
 from .sampler_config import SamplerConfig
 
 
 class FlattenedControls:
     def __init__(self, controls: list[ControlConfig]) -> None:
-        self._controls = []
-        self._samplers: list[SamplerConfig] = []
-
-        for control in controls:
-            control_sampler_idx = -1
-            variables = []
-            for variable in control.variables:
-                match variable:
-                    case ControlVariableConfig():
-                        var_dict, control_sampler_idx = self._add_variable(
-                            control, variable, control_sampler_idx
-                        )
-                        variables.append(var_dict)
-                    case ControlVariableGuessListConfig():
-                        var_dicts, control_sampler_idx = self._add_variable_guess_list(
-                            control, variable, control_sampler_idx
-                        )
-                        variables.extend(var_dicts)
-            self._inject_defaults(control, variables)
-            self._controls.extend(variables)
-
-        self.names = [control["name"] for control in self._controls]
+        control_dicts = _get_control_dicts(controls)
+        self.names = [control["name"] for control in control_dicts]
         self.types: list[Literal["real", "integer"]] = [
-            control["control_type"] for control in self._controls
+            control["control_type"] for control in control_dicts
         ]
-        self.initial_guesses = [control["initial_guess"] for control in self._controls]
-        self.lower_bounds = [control["min"] for control in self._controls]
-        self.upper_bounds = [control["max"] for control in self._controls]
-        self.scaled_ranges = [control["scaled_range"] for control in self._controls]
-        self.enabled = [control["enabled"] for control in self._controls]
+        self.initial_guesses = [control["initial_guess"] for control in control_dicts]
+        self.lower_bounds = [control["min"] for control in control_dicts]
+        self.upper_bounds = [control["max"] for control in control_dicts]
+        self.scaled_ranges = [control["scaled_range"] for control in control_dicts]
+        self.enabled = [control["enabled"] for control in control_dicts]
         self.perturbation_magnitudes = [
-            control["perturbation_magnitude"] for control in self._controls
+            control["perturbation_magnitude"] for control in control_dicts
         ]
         self.perturbation_types = [
-            control["perturbation_type"] for control in self._controls
+            control["perturbation_type"] for control in control_dicts
         ]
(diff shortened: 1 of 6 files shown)


src/everest/config/utils.py
```diff
@@ -1,124 +1,92 @@
...
+            "scaled_range",
+        ]:
+            if var_dict.get(key) is None:
+                var_dict[key] = getattr(control, key)
+
+    control_dicts: list[dict[str, Any]] = []
+    for control in controls:
+        for variable in control.variables:
+            if isinstance(variable, ControlVariableGuessListConfig):
+                for index, guess in enumerate(variable.initial_guess, start=1):
```
StephanDeHoop, line 50: Why do we start from `index=1`?

src/everest/config/utils.py
```diff
@@ -1,124 +1,92 @@
...
+                    var_dict = deepcopy(variable.model_dump())
+                    var_dict["name"] = (control.name, variable.name, index)
+                    var_dict["initial_guess"] = guess
+                    _inject_defaults(control, var_dict)
+                    control_dicts.append(var_dict)
+            else:
```
StephanDeHoop, line 56: We don't have the explicit check for `ControlVariableConfig` anymore (not that is was here, but in the other logic we had only two cases). Are there more variable types than `ControlVariableConfig` and `ControlVariableGuessListConfig`? I guess this just follows from how the user specifies the controls in the everest-config.

src/everest/config/utils.py
```diff
@@ -1,124 +1,92 @@
...
+                var_dict = deepcopy(variable.model_dump())
+                var_dict["name"] = (
+                    (control.name, variable.name)
+                    if variable.index is None
```
StephanDeHoop, line 60: Reason I am asking is, `ControlVariableConfig` has property `index`, but might this throw an exception (e.g., `AttributeError`) if we have a variable type that doesn't have this property?

tests/everest/test_controls.py
```diff
@@ -401,3 +401,73 @@ def test_controls_ordering_disregards_index():
...
+    assert initial1 == initial2
+    assert ropt_config1["names"]["variable"] == ropt_config2["names"]["variable"]
+    for key in [
+        "lower_bounds",
+        "upper_bounds",
+        "perturbation_magnitudes",
+        "perturbation_types",
+        "mask",
+    ]:
+        assert ropt_config1["variables"][key] == ropt_config2["variables"][key]
```
StephanDeHoop, line 473: Was the bug in the way ropt handled the control/variables specification from everest (i.e., did the bug fix also entail a fix in ropt?)? Or was just this mapping incorrect? I follow the logic of the new way, it's a bit hard to completely decipher why it was failing in the first place (maybe add it the commit body or PR description) (since the bug seems to be related to controls/variables changing even though they should be constant).

src/everest/config/utils.py
verveerpj: That was decided when the feature was implemented first. Most users will start counting from one.

src/everest/config/utils.py
verveerpj: No, currently there are only these two types. Should we add another one, it just a question of making this a `match` statement.

src/everest/config/utils.py
verveerpj: It would, if we add another type, we will have to amend this code.

tests/everest/test_controls.py
verveerpj: It was failing because the `ControlVariableGuessListConfig` case did not handle any of the other fields at all. Which meant it would take the values from the parent control field, or default values. In reported case, the `enabled` field was not handled. This PR makes sure that the code for handling all fields is now shared by `ControlVariableGuessListConfig` and `ControlVariableConfig`. I will amend the commit/PR comments.

src/everest/config/utils.py
StephanDeHoop: Ah ok sounds good, somehow I thought that this means that, for example, `config_advanced.yml` still starts counting from `index: 0` and will get skipped now, but that's wrong because those are specified as `ControlVariableConfig` and not `ControlVariableGuessListConfig` (I just realize that now :"))

verveerpj: The entire indexing situation is a bit unfortunate, since if you specify the manually you can specify them in any order/and or offset. That is not indexing, that is just naming with numbers...

tests/everest/test_controls.py
StephanDeHoop: Ah that makes sense, thanks :)!

src/everest/config/utils.py
StephanDeHoop: True haha :")

verveerpj: It came up before: #11219

PR #11333, opened 2025-07-09
Start storage server on KeyError from storage_server.json

### LINKED ISSUE #11325: Investigate cert KeyError
Observed crash in logs for Ert 14.4.0:

```
          File "/prog/res/komodo/2025.07.00-py311-rhel8-numpy1/root/lib64/python3.11/site-packages/ert/__main__.py", line 644, in main
            args.func(args, context.plugin_manager)
          File "/prog/res/komodo/2025.07.00-py311-rhel8-numpy1/root/lib64/python3.11/site-packages/ert/__main__.py", line 191, in run_gui_wrapper
            run_gui(args, ert_plugin_manager)
          File "/usr/lib64/python3.11/contextlib.py", line 81, in inner
            return func(*args, **kwds)
                   ^^^^^^^^^^^^^^^^^^^
          File "/prog/res/komodo/2025.07.00-py311-rhel8-numpy1/root/lib64/python3.11/site-packages/ert/gui/main.py", line 116, in run_gui
            with StorageService.init_service(project=os.path.abspath(ens_path)):
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
          File "/prog/res/komodo/2025.07.00-py311-rhel8-numpy1/root/lib64/python3.11/site-packages/ert/services/storage_service.py", line 62, in init_service
            _ = service.fetch_url()
                ^^^^^^^^^^^^^^^^^^^
          File "/prog/res/komodo/2025.07.00-py311-rhel8-numpy1/root/lib64/python3.11/site-packages/ert/services/storage_service.py", line 78, in fetch_url
            verify=con_info["cert"],
                   ~~~~~~~~^^^^^^^^
        KeyError: 'cert'
        ",
        "type": "KeyError"
```

berland:
If KeyErrors occurs while trying to connect to the server, then storage_server.json is malformed and should be regarded the same as a TimeoutError, i.e. a new server will be started and storage_server.json is overwritten

**Issue**
Resolves #11325 

**Approach**
catch

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/services/storage_service.py
+++ b/src/ert/services/storage_service.py
@@ -60,7 +60,7 @@ class StorageService(BaseService):
             service = cls.connect(timeout=0, project=kwargs.get("project", os.getcwd()))
             # Check the server is up and running
             _ = service.fetch_url()
-        except TimeoutError:
+        except (TimeoutError, KeyError):
             return cls.start_server(*args, **kwargs)
         return _Context(service)
 
--- a/tests/ert/unit_tests/services/test_storage_service.py
+++ b/tests/ert/unit_tests/services/test_storage_service.py
@@ -2,6 +2,7 @@ import glob
 import json
 import os
 import socket
+from unittest.mock import patch
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


src/ert/services/storage_service.py
```diff
@@ -60,7 +60,7 @@ def init_service(cls, *args: Any, **kwargs: Any) -> _Context[StorageService]:
             service = cls.connect(timeout=0, project=kwargs.get("project", os.getcwd()))
             # Check the server is up and running
             _ = service.fetch_url()
-        except TimeoutError:
+        except (TimeoutError, KeyError):
```
frode-aarstad, line 63: Perhaps we should log this exception before continuing

tests/ert/unit_tests/services/test_storage_service.py
```diff
@@ -51,6 +52,35 @@ def test_that_service_can_be_started_with_existing_conn_info_json(tmp_path):
...
+                socket.gethostname(),
+                socket.getfqdn(),
+            )
+        ],
+        "authtoken": "dummytoken",
+    }
+    with open(tmp_path / "storage_server.json", mode="w", encoding="utf-8") as f:
+        json.dump(connection_info, f)
+    StorageService.init_service(project=str(tmp_path))
+    start_server_mock.assert_called_once()
```
yngve-sk, line 81: Is this part just for performance optimization/avoiding teardown? Test would still be valid without it I think? in that it boots up without error'ing

src/ert/services/storage_service.py
berland: We should, but it was slightly awkward in a class method. Is it risk-free to convert to an instance method?

tests/ert/unit_tests/services/test_storage_service.py
berland: Yes, this is for performance, makes the test fly ini 0.2 sec instead of many. Manually confirmed that the test will fail if the fix is not in place.

src/ert/services/storage_service.py
berland: I will leave that as a residual issue, as we are still not handling all scenarios with a pre-existing storage_server.json.

yngve-sk, line 63: I think this will work. Sidenote: This whole BaseService setup is scheduled for refactoring/simplification, so having this in because it works for a while is also OK for me. Did not see any obvious "nice" place to explicitly delete the file if it exists. It is not obvious where the file is created either.

yngve-sk (approved):

PR #11339, opened 2025-07-09
Revert to previous behavior of FlattenedControls

verveerpj:
**Issue**
The previous fix of the FlattenedControls slightly changed the way perturbations were sampled, and hence changed the gradients. This commit reverts to the old behavior where a single default sampler is used, if no samplers are defined in the configuration.

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/config/utils.py
+++ b/src/everest/config/utils.py
@@ -70,21 +70,37 @@ def _get_samplers(
 ) -> tuple[list[SamplerConfig | None], list[int]]:
     samplers: list[SamplerConfig | None] = []
     sampler_indices: list[int] = []
+
+    default_sampler_index: int | None = None
+
     for control in controls:
-        control_sampler_idx: int | None = None
+        control_sampler_index: int | None = None
+
         for variable in control.variables:
-            if variable.sampler is None:
-                if control_sampler_idx is None:
-                    samplers.append(control.sampler)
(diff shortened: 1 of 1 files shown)


src/everest/config/utils.py
```diff
@@ -70,21 +70,37 @@ def _get_samplers(
...
+        control_sampler_index: int | None = None
+
         for variable in control.variables:
-            if variable.sampler is None:
-                if control_sampler_idx is None:
+            if variable.sampler is not None:
+                # Use the sampler of the variable:
+                samplers.append(variable.sampler)
+                variable_sampler_index = len(samplers) - 1
+            elif control.sampler is not None:
```
StephanDeHoop, line 84: Somehow in the old code I assumed that there will always be a either a control or variable sampler. This is more clear indeed! The change in results was because neither the control and variable has samplers and a different default sampler was used? Or was it because of something else?

StephanDeHoop (approved):
Looks good, just one question :)!

src/everest/config/utils.py
verveerpj: In the change a new default sampler was defined for each control, instead of a single default sampler. That caused the issue because you added a new control. I reverted to a single default, so I assume that solves the issue.

PR #11352, opened 2025-07-10
Add manual update EnIF to ert

yngve-sk:
Resolves https://github.com/equinor/ert/issues/11423

Screenshot 2025-09-17 at 09 09 45

### DIFF
--- a/tests/ert/ui_tests/gui/test_full_manual_update_workflow.py
+++ b/tests/ert/ui_tests/gui/test_full_manual_update_workflow.py
@@ -3,13 +3,13 @@ import shutil
 
 import numpy as np
 import polars as pl
+import pytest
 from PyQt6.QtCore import Qt
 from PyQt6.QtWidgets import QComboBox, QToolButton, QTreeView, QWidget
 
 from ert.data import MeasuredData
 from ert.gui.simulation.evaluate_ensemble_panel import EvaluateEnsemblePanel
 from ert.gui.simulation.experiment_panel import ExperimentPanel
-from ert.gui.simulation.manual_update_panel import ManualUpdatePanel
 from ert.gui.simulation.run_dialog import RunDialog
 from ert.gui.tools.manage_experiments import ManageExperimentsPanel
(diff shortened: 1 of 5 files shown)


src/ert/gui/simulation/manual_enif_update_panel.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+@dataclass
+class Arguments:
+    mode: str
+    realizations: str
+    ensemble_id: str
+    target_ensemble: str
+
+
```
larsevj: I am not a huge fan of adding more copy paste code for the panels, that need to be maintained, so would prefer this to be an option in the existing manual update panel.

That might be a lot more work though...

tests/ert/ui_tests/gui/test_full_manual_update_workflow.py
```diff
@@ -28,9 +30,8 @@ def test_manual_analysis_workflow(ensemble_experiment_has_run, qtbot):
 
     # Select correct experiment in the simulation panel
     experiment_panel = get_child(gui, ExperimentPanel)
-    simulation_settings = get_child(experiment_panel, ManualUpdatePanel)
     simulation_mode_combo = get_child(experiment_panel, QComboBox)
-    simulation_mode_combo.setCurrentText(ManualUpdate.name())
+    simulation_mode_combo.setCurrentText(mode)
```
larsevj: Is it possible to to check that these to parametrized tests produces different output? or output similar to their enif/es equivalents?

yngve-sk: Would be possible, but the test is generalized to just checking that the posterior has less variance, don't think it would be a good idea to do in a GUI test. EnIF is covered by snapshot tests (not specifically manual update runmodel, but it runs through the same codepath as the "enif" snapshots).

src/ert/gui/simulation/manual_enif_update_panel.py
yngve-sk: Added EnIF dropdown to manual update instead

PR #11366, opened 2025-07-11
Make Everviz work with old storages

### LINKED ISSUE #11353: Everviz not working for older cases on bleeding/stable

Cases run with an ert before this PR: https://github.com/equinor/ert/pull/11052 
can not be opened with a newert ERT, due to a missing status.json file, in `visualization_script.py`.

_No response_

### Steps to reproduce

Run math_func/advanced with older Everest, then try open it with newer Everest and watch it fail

### LINKED ISSUE #11356: Enable visualization of in-progress everest results
Definition of done:
* status_file dependency must be removed, it should not be concerned with its contents
* UX is OK for these cases:
* * Open everviz with non-existing storage should print an error like `no storage found for case, no data to show until one batch is complete`
* * Open everviz with empty (but existing) storage should be successful, but show no data
* * Open everviz on case when it is running and at least one batch has been written
* * Reload of chart (or reload page) in web browser will render data that is currently in storage

Caching must be removed / revised from `export_everest_opt_results_to_csv`, as it would need to be updated for every batch. 

Keep in mind, might encounter race conditions

yngve-sk:
**Issue**
Resolves #11356 
Resolves #11353 

**Approach**
Use storage instead. Initiate storage migration only if storage is old. Attempt to open readonly storage to not interfere with writing in progress (if running simulation)

### DIFF
--- a/src/ert/storage/migration/to12.py
+++ b/src/ert/storage/migration/to12.py
@@ -1,7 +1,26 @@
 import json
 from pathlib import Path
+from typing import Any
 
 info = "Migrate response and parameter configs to include type"
+
+
+def migrate_everest_param(config: dict[str, Any]) -> dict[str, Any]:
+    formatted_control_names = []
+
+    name = config["name"]
+    input_keys = config["input_keys"]
+
+    if isinstance(input_keys, list):
+        return config
+
+    # It is a dict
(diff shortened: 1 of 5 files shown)


verveerpj (approved):
Seems you broke a test... Otherwise looking good.

Issue #11386, opened 2025-07-21
test_cli.py run_cli is slower than actual CLI


yngve-sk:
Running the forward model takes longer when running via the `run_cli` function. Running 
```
ert ensemble_information_filter --disable-monitoring --target-ensemble "iter-%d" --realizations 1,2,4,8,16 test-data/ert/poly_example/poly.ert
```

takes a few seconds, whereas `test_enif` itself takes 30s-1min to run. Other tests using the `run_cli` also run a lot slower than they do when running from the CLI.

berland:
I speculate that the `justfile` commands are used in CI, there overbooking CPU easily makes the whole thing slower.

PR #11388, opened 2025-07-21
Don't limit max running for cli/gui tests

yngve-sk:
**Issue**
Resolves https://github.com/equinor/ert/issues/11386


jonathan-eq:
How much faster is this?

PR #11395, opened 2025-07-28
Activate default value for Active realizations

### LINKED ISSUE #11156: When running "Evaluate Ensemble" based on 100% finished ensemble, validation errors occur after run

If you (for some reason, a bit corner casy) have an ensemble with "all finished" realizations, and try to run the "evaluate ensemble" on it, you have to specify (no default) in the "Active Realizations" (rename suggested in https://github.com/equinor/ert/issues/11155) 

When the experiment/simulation is finished, this fields gets updated again (with blank values) and there is a validation error that shows "through" to the simulation panel/window. 

Image

Either set "all" (eg. 0.-99) as default or don't show the validation error unless you actually interact with the "start simulation" panel again. Should perhaps fix the latter in any case, but a suggestion of setting the prefilled values to "all" is in this case discussed and kind of agreed with @dafeda and @oyvindeide 

### Steps to reproduce

```Python

```

- [ ] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

SAKavli:
When running evaluate ensemble, this value is reset back to nothing, causing a red warning containing "Missing Argument!" to pop up in the experiment result panel. A quick fix for this is giving a default value to reset back to, as discussed in the related issue.

**Issue**
Resolves #11156

**Approach**

The current behavior of EvaluateEnsemblePanel is to populate "active realizations" with the intersection of realizations with parameters and those which failed.

When all realizations succeeds, meaning active realizations is set to nothing, having a fallback value being all realizations prevents getting a "Missing Argument!" popup from EvaluateEnsemblePanel.

Before:

https://github.com/user-attachments/assets/0bd85920-d75f-40c8-946f-2fbb9932f12a

After:

https://github.com/user-attachments/assets/8d462011-262a-4594-b6d1-bf0574aad384

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/gui/simulation/evaluate_ensemble_panel.py
+++ b/src/ert/gui/simulation/evaluate_ensemble_panel.py
@@ -103,6 +103,8 @@ class EvaluateEnsemblePanel(ExperimentConfigPanel):
                 mask = np.logical_and(
                     parameters, np.logical_or(missing_responses, failures)
                 )
+                if not any(mask):
+                    mask = parameters
                 self._active_realizations_field.model.setValueFromMask(mask)  # type: ignore
         except OSError as err:
             logger.error(str(err))
--- a/tests/ert/ui_tests/gui/test_evaluate_ensemble_panel.py
+++ b/tests/ert/ui_tests/gui/test_evaluate_ensemble_panel.py
@@ -1,0 +1,41 @@
+from PyQt6.QtCore import Qt, QTimer
+from PyQt6.QtWidgets import QComboBox, QWidget
+
+from ert.gui.simulation import ExperimentPanel
+from ert.gui.simulation.evaluate_ensemble_panel import EvaluateEnsemblePanel
+from ert.gui.simulation.run_dialog import RunDialog
+from tests.ert.ui_tests.gui.conftest import (
+    DEFAULT_NUM_REALIZATIONS,
+    get_child,
+    wait_for_child,
+)
+from tests.ert.ui_tests.gui.test_restart_ensemble_experiment import (
+    handle_run_path_dialog,
+)
+
+
+def test_evaluate_ensemble_active_realizations_resets_to_all_realizations_with_parameters_when_all_realizations_are_successful(  # noqa
+    qtbot, opened_main_window_poly
+):
+    gui = opened_main_window_poly
+
+    num_reals = DEFAULT_NUM_REALIZATIONS
+    expected_active_reals = f"0-{num_reals - 1}" if num_reals > 1 else "0"
+
+    experiment_panel = gui.findChild(ExperimentPanel)
+    simulation_mode_combo = experiment_panel.findChild(QComboBox)
+    simulation_mode_combo.setCurrentText("Ensemble experiment")
+
+    run_experiment = experiment_panel.findChild(QWidget, name="run_experiment")
+    QTimer.singleShot(1000, lambda: handle_run_path_dialog(gui, qtbot))
+    qtbot.mouseClick(run_experiment, Qt.MouseButton.LeftButton)
+
+    run_dialog = wait_for_child(gui, qtbot, RunDialog)
+    qtbot.waitUntil(lambda: run_dialog.is_simulation_done() is True, timeout=60000)
+    qtbot.waitUntil(lambda: run_dialog._tab_widget.currentWidget() is not None)
+
+    evaluate_ensemble_active_realizations = get_child(
+        experiment_panel, EvaluateEnsemblePanel
+    )._active_realizations_field.model.getValue()
+
+    assert evaluate_ensemble_active_realizations == expected_active_reals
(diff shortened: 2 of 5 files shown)


jonathan-eq:
I like the cursor circling in the video 🥇

jonathan-eq:
What happens if you run ensemble_experiment with a couple realizations failing, and select evaluate ensemble? Will it auto-select the failing ones for all run_models, or just evaluate experiment?

src/ert/gui/simulation/evaluate_ensemble_panel.py
```diff
@@ -103,6 +103,8 @@ def _realizations_from_fs(self) -> None:
                 mask = np.logical_and(
                     parameters, np.logical_or(missing_responses, failures)
                 )
+                if not any(mask):
+                    mask = np.logical_and(parameters, ~missing_responses)
```
jonathan-eq, line 107: Beautiful!

jonathan-eq: Does this make mask(active_realizations?) the realizations where we have parameters AND responses?

jonathan-eq: ^Will that be all realizations that previously succeeded?

jonathan-eq: I wonder if it would be the same if we set it to `mask = parameters` instead. Wouldnt that also select all realizations in ensemble?

SAKavli: Yes, I think you're right. There are two other panels doing roughly the same job, but a bit differently from each other. Given a second thought, it makes much more sense like you suggest.

https://github.com/equinor/ert/blob/402ed51e1f55bb4979f413470119b05ec946517a/src/ert/gui/simulation/multiple_data_assimilation_panel.py#L313

https://github.com/equinor/ert/blob/402ed51e1f55bb4979f413470119b05ec946517a/src/ert/gui/simulation/manual_update_panel.py#L127

jonathan-eq:
Very good! Should be easy to change if someone disagrees after the summer, as we now know what pieces to touch up. Should we squash some commits, or is it ready?

src/ert/gui/simulation/evaluate_ensemble_panel.py
SAKavli: Yes. Might not be the intended behavior we want, like you mentioned.

SAKavli:
> Very good! Should be easy to change if someone disagrees after the summer, as we now know what pieces to touch up. Should we squash some commits, or is it ready?

I'll squash the two last ones. The two first are kind of out of scope, picked them up as I believe that was where I would make changes. Do you think it's fine to keep in this PR?

It seemed as if `_custom` was only used within the class, couldn't find any references and no tests broke, but other than that idk where it comes from. I think it was from 2020 or something.

tests/ert/ui_tests/gui/test_restart_no_responses_and_parameters.py
```diff
@@ -87,12 +87,25 @@ def open_gui(tmp_path, monkeypatch, run_experiment, tmp_path_factory):
         yield gui
 
 
-def test_sensitivity_restart(open_gui, run_experiment):
+@pytest.fixture
+def open_gui_succeeding(tmp_path, monkeypatch, run_experiment, tmp_path_factory):
```
jonathan-eq: Do we need tmp_path_factory?

SAKavli: Perhaps not, just copy paste of the previous test. Can try to remove it and see whether the red light start blinking 🚦

tests/ert/ui_tests/gui/test_evaluate_ensemble_panel.py
```diff
@@ -0,0 +1,34 @@
...
+    run_dialog = wait_for_child(gui, qtbot, RunDialog)
+    qtbot.waitUntil(lambda: run_dialog.is_simulation_done() is True, timeout=60000)
+    qtbot.waitUntil(lambda: run_dialog._tab_widget.currentWidget() is not None)
+
+    evaluate_ensemble_active_realizations = (
+        get_child(experiment_panel, EvaluateEnsemblePanel)
+        .findChild(StringBox)
+        .model.getValue()
+    )
+    assert evaluate_ensemble_active_realizations == "0-9"
```
SAKavli: Tested on main, main gives `None` for this value, which is the default value if `show_default` in `ActiveRealizationsModel` is `False`

andreas-el (approved):
Nice work! 💯

SAKavli:
Squashed last 4 commits into 1

PR #11397, opened 2025-07-29
Improve typing for src/_ert

jonathan-eq:
**Approach**
Improve typing for src/_ert.
This bring down the mypy errors from 130 to 4.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/_ert/forward_model_runner/fm_dispatch.py
+++ b/src/_ert/forward_model_runner/fm_dispatch.py
@@ -8,8 +8,10 @@ import sys
 import time
 from collections.abc import Generator, Iterable
 from datetime import datetime
+from typing import TYPE_CHECKING, Any
 
 from _ert.forward_model_runner import reporting
+from _ert.forward_model_runner.reporting.base import Reporter
 from _ert.forward_model_runner.reporting.message import (
     Exited,
     Finish,
@@ -17,6 +19,17 @@ from _ert.forward_model_runner.reporting.message import (
     ProcessTreeStatus,
 )
 from _ert.forward_model_runner.runner import ForwardModelRunner
+
+if TYPE_CHECKING:
+    from ert.config.forward_model_step import ForwardModelJSON
+
+    class ForwardModelDescriptionJSON(ForwardModelJSON):
+        ens_id: str | None
+        real_id: int
+        dispatch_url: str | None
+        ee_token: str | None
+        experiment_id: str | None
+
 
 # This is incorrecty named, but is kept to avoid a breaking change.
 # "job" was previously used for what is now called a "forward_model_step".
@@ -32,11 +45,11 @@ logger = logging.getLogger(__name__)
 
 
 def _setup_reporters(
-    is_interactive_run,
-    ens_id,
-    dispatch_url,
-    ee_token=None,
-    experiment_id=None,
+    is_interactive_run: bool,
+    ens_id: str | None,
+    dispatch_url: str | None,
+    ee_token: str | None = None,
+    experiment_id: str | None = None,
 ) -> list[reporting.Reporter]:
     reporters: list[reporting.Reporter] = []
     if is_interactive_run:
@@ -52,7 +65,7 @@ def _setup_reporters(
     return reporters
 
 
-def _setup_logging(directory: str = "logs"):
+def _setup_logging(directory: str = "logs") -> None:
     fm_runner_logger = logging.getLogger("_ert.forward_model_runner")
     memory_csv_logger = logging.getLogger("_ert.forward_model_memory_profiler")
 
@@ -82,11 +95,11 @@ def _setup_logging(directory: str = "logs"):
     fm_runner_logger.setLevel(logging.DEBUG)
 
 
-def _wait_for_retry():
+def _wait_for_retry() -> None:
     time.sleep(FILE_RETRY_TIME)
 
 
-def _read_fm_description_file(retry=True):
+def _read_fm_description_file(retry: bool = True) -> "ForwardModelDescriptionJSON":
     try:
         with open(FORWARD_MODEL_DESCRIPTION_FILE, encoding="utf-8") as json_file:
             return json.load(json_file)
@@ -130,7 +143,9 @@ def _report_all_messages(
             _stop_reporters_and_sigkill(reporters)
 
 
-def _stop_reporters_and_sigkill(reporters, exited_event: Exited | None = None):
+def _stop_reporters_and_sigkill(
+    reporters: Iterable[Reporter], exited_event: Exited | None = None
+) -> None:
     _stop_reporters(reporters, exited_event)
     pgid = os.getpgid(os.getpid())
     os.killpg(pgid, signal.SIGKILL)
@@ -144,7 +159,16 @@ def _stop_reporters(
             reporter.stop(exited_event=exited_event)
 
 
-def fm_dispatch(args):
+class Namespace(argparse.Namespace):
+    """
+    fm_dispatch argument parser namespace
+    """
+
+    run_path: str | None = None
+    steps: list[str]
+
+
+def fm_dispatch(args: list[str]) -> None:
     parser = argparse.ArgumentParser(
         description=(
             "Run all the forward model steps specified in jobs.json, "
@@ -161,7 +185,7 @@ def fm_dispatch(args):
         ),
     )
 
-    parsed_args = parser.parse_args(args[1:])
+    parsed_args = parser.parse_args(args[1:], namespace=Namespace())
 
     # If run_path is defined, enter into that directory
     if parsed_args.run_path is not None:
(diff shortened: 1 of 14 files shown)


src/_ert/forward_model_runner/forward_model_step.py
```diff
@@ -388,11 +387,11 @@ def _check_target_file_is_written(
...
         return f"Could not find target_file:{target_file}"
 
-    def _assert_arg_list(self):
-        errors = []
-        if "arg_types" in self.step_data:
-            arg_types = self.step_data["arg_types"]
-            arg_list = self.step_data.get("argList")
+    def _assert_arg_list(self) -> list[str]:
+        errors: list[str] = []
+        if "arg_types" in self.step_data:  # This seems to be NEVER true(?)
```
jonathan-eq, line 392: This might be a bug from last fall, that was never picked up because of the missing typing.

src/_ert/forward_model_runner/fm_dispatch.py
```diff
@@ -18,6 +20,17 @@
...
 
+if TYPE_CHECKING:
+    from ert.config.forward_model_step import ForwardModelJSON
+
+    class ForwardModelDescriptionJSON(ForwardModelJSON):
+        ens_id: str | None
+        real_id: int
+        dispatch_url: str | None
+        ee_token: str | None
+        experiment_id: str | None
```
jonathan-eq, line 31: This is added by scheduler right before we start the fm_dispatchers

src/_ert/forward_model_runner/fm_dispatch.py
```diff
@@ -18,6 +20,17 @@
 )
 from _ert.forward_model_runner.runner import ForwardModelRunner
 
+if TYPE_CHECKING:
+    from ert.config.forward_model_step import ForwardModelJSON
```
jonathan-eq, line 24: After some digging, this seems to be the content of jobs.json.

src/_ert/forward_model_runner/forward_model_step.py
```diff
@@ -388,11 +387,11 @@ def _check_target_file_is_written(
...
+            arg_types = self.step_data.get("arg_types")
```
jonathan-eq: Some places we use dict["some_key"], other places we do dict.get("some_key"). This should be more consistent, and we should add some better error handling in case we are missing keys in jobs.json that are required. That will be a different PR.

src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -135,37 +143,49 @@ async def publisher():
...
             )
-            self._dump_event(event)
             if not msg.success():
                 logger.error(f"Step {step_name} FAILED to start")
-                event = ForwardModelStepFailure(**step_msg, error_msg=msg.error_message)
+                error_message = (
+                    msg.error_message
+                    if msg.error_message is not None
+                    else "Unknown error"
+                )
```
jonathan-eq, line 214: Not sure if this is ever the case, but we should set it just in case.

src/_ert/forward_model_runner/reporting/file.py
```diff
@@ -53,72 +66,64 @@ def report(self, msg: Message):
...
+                fm_step_status["start_time"] = datetime_serialize(msg.timestamp)
             else:
                 logger.error(f"Forward model step {msg.step.name()} FAILED to start")
-                error_msg = msg.error_message
-                fm_step_status.update(
-                    status=_STEP_STATUS_FAILURE,
-                    error=error_msg,
-                    end_time=data_util.datetime_serialize(msg.timestamp),
-                )
+                fm_step_status["status"] = "Failure"
```
jonathan-eq: This is no longer using the _STEP_STATUS_RUNNING variable, but is instead documented in the string typing of the TypedDict using Literal["Failure"|"Running"|...]

src/_ert/forward_model_runner/reporting/file.py
```diff
@@ -53,72 +66,64 @@ def report(self, msg: Message):
...
-            fm_step_status.update(
-                max_memory_usage=msg.memory_status.max_rss,
-                current_memory_usage=msg.memory_status.rss,
-                cpu_seconds=msg.memory_status.cpu_seconds,
-                status=_STEP_STATUS_RUNNING,
-            )
+            fm_step_status["max_memory_usage"] = msg.memory_status.max_rss
+            fm_step_status["current_memory_usage"] = msg.memory_status.rss
+            fm_step_status["cpu_seconds"] = msg.memory_status.cpu_seconds
+            fm_step_status["status"] = "Running"
```
jonathan-eq: mypy did NOT like dict.update(...)

src/_ert/forward_model_runner/reporting/statemachine.py
```diff
@@ -25,23 +26,23 @@ def __init__(self) -> None:
         steps = (Start, Running, Exited)
         checksum = (Checksum,)
         finished = (Finish,)
-        self._handler: dict[Message, Callable[[Message], None]] = {}
+        self._handler: dict[Any, Callable[[Any], None]] = {}
```
jonathan-eq, line 29: It was very difficult to try getting the typing here to work, as we are using tuples of subclasses of Message, so it will be typing.Any for now.

src/_ert/forward_model_runner/forward_model_step.py
flikka: A bug, in that it should be removed?

src/_ert/forward_model_runner/forward_model_step.py
```diff
@@ -388,11 +387,11 @@ def _check_target_file_is_written(
...
+            arg_list = self.step_data["argList"]
```
xjules: You have changed "arg_types" via get and "argList" to step_data["argList"] ?

src/_ert/forward_model_runner/forward_model_step.py
flikka: I suppose we should use .get if we want to provide a default, or have None as default? Not sure I understand why you swapped these though, we check here that arg_types is in step_data, no? But the other one (argList) would now raise error (guess that is ok?). But makes more sense to use brackets for the arg_types too, no?

src/_ert/forward_model_runner/forward_model_step.py
jonathan-eq: I think we want it. It might already be validated on the client side of ert, but before we did it in fm_dispatch too. I am not sure if it is no longer needed on the dispatcher side too, or if it the typing elsewhere that is incorrect. I will double check. @oyvindeide

src/ert/config/forward_model_step.py
```diff
@@ -51,7 +51,7 @@ class ForwardModelStepJSON(TypedDict):
     """
 
     name: str
-    executable: list[str]
```
xjules, line 54: fascinating!

src/_ert/forward_model_runner/util/data.py
```diff
@@ -1,14 +1,30 @@
...
 
-from _ert.forward_model_runner.reporting.message import _STEP_STATUS_WAITING
+if TYPE_CHECKING:
+    from _ert.forward_model_runner.forward_model_step import ForwardModelStep
 
 
-def create_step_dict(step):
+class StepDict(TypedDict, total=False):
+    name: str
+    status: Literal["Waiting", "Running", "Failure", "Success"]
```
xjules: I don't like strings here tbh. What about:
```python
class StepStatus(StrEnum):
    WAITING = "Waiting"
    RUNNING = "Running"
    FAILURE = "Failure"
    SUCCESS = "Success"

class StepDict(TypedDict, total=False):
    name: str
    status: StepStatus
...
```

src/_ert/forward_model_runner/fm_dispatch.py
```diff
@@ -144,7 +159,16 @@ def _stop_reporters(
             reporter.stop(exited_event=exited_event)
 
 
-def fm_dispatch(args):
+class Namespace(argparse.Namespace):
```
xjules, line 162: `FMArgsNamespace` ?

xjules: Also it seems that the defaults are not pased automatically. Suggestion from chatgpt:
```python
class Namespace(argparse.Namespace):
def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.run_path: Optional[str] = None
        self.steps: list[str] = []
```

src/_ert/forward_model_runner/util/data.py
flikka: Good suggestion!

src/_ert/forward_model_runner/fm_dispatch.py
jonathan-eq: Hmmm, not sure. This is a direct copy of what we already have in Ert, so I think we should keep it that way.

PR #11398, opened 2025-07-29
Refactor RunModel to use asyncio.Event for cancellation

### LINKED ISSUE #11340: Refactor run_model.py to not use SimpleQueue to cancel evaluator
Currently, when we want to cancel an experiment, the run model puts a "END"-message into a simple queue that the evaluator polls. This should be refactored to be either an asyncio.Event or a threading.Event. We could also move ownership of the queue from run_model to evaluator, but this would require some refactoring of evaluator so that it does not send an initial snapshot on initialization if experiment is cancelled pre-evaluation. See https://github.com/equinor/ert/pull/10726 and https://github.com/equinor/ert/pull/10726#discussion_r2192301462

jonathan-eq:
This commit changes RunModel and EnsembleEvaluator to use asyncio.Event to cancel instead of queue.SimpleQueue.

**Issue**
Resolves #11340

**Approach**
🏎️ 

      --exec 'just rapid-tests'`)

## When applicable


tests/ert/unit_tests/test_tracking.py
```diff
@@ -281,7 +281,7 @@ def test_setting_env_context_during_run(
 
     thread = ErtThread(
         name="ert_cli_simulation_thread",
-        target=model.start_simulations_thread,
+        target=model.api.start_simulations_thread,
```
jonathan-eq, line 284: I am not really sure why we created RunModelAPI if we only used it one place(?)

jonathan-eq:
This will cause problems for everest, as the optimizer cannot take in an async function. This will cause us calling async_method -> sync_method -> async_method. It might be possible to create a workaround by using asyncio.run_coroutine_threadsafe(), but it does not seem to work. asyncio.run() would work, but there is already an event_loop running so it does not work out of the box.

@xjules, do you remember how we solved this in Ert, we had that problem there too. Did we just force everything to be async? That won't work here, as the optimizer is from ropt.

PR #11410, opened 2025-08-05
Move evaluate from ensemble to EnsembleEvaluator

jonathan-eq:
**Approach**
It makes more sense that evaluate be in EnsembleEVALUATOR instead of ensemble. This also moves scheduler and driver from ensemble to evaluator, so we don't have to do as nested method calls anymore (still nested, but slightly better).
 
      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -32,6 +32,8 @@ from _ert.forward_model_runner.client import (
 )
 from _ert.forward_model_runner.fm_dispatch import FORWARD_MODEL_TERMINATED_MSG
 from ert.ensemble_evaluator import identifiers as ids
+from ert.scheduler import create_driver
+from ert.scheduler.scheduler import Scheduler
 
 from ..config import QueueSystem
 from ._ensemble import FMStepSnapshot
@@ -70,6 +72,8 @@ class EnsembleEvaluator:
         event_handler: Callable[[EEEvent], None] | None = None,
     ) -> None:
         self._config: EvaluatorServerConfig = config
+        if self._config is None:
+            raise ValueError("no config for evaluator")
         self._ensemble: Ensemble = ensemble
 
         self._events: asyncio.Queue[SnapshotInputEvent] = asyncio.Queue()
@@ -104,6 +108,18 @@ class EnsembleEvaluator:
 
         self._publisher_receiving_timeout: float = 60.0
         self._evaluation_result: asyncio.Future[bool] = asyncio.Future()
+        self._scheduler = Scheduler(
+            create_driver(self.ensemble._queue_config.queue_options),
+            self.ensemble.active_reals,
+            self._manifest_queue,
+            self._events,
+            max_submit=self.ensemble._queue_config.max_submit,
+            max_running=self.ensemble._queue_config.max_running,
+            submit_sleep=self.ensemble._queue_config.submit_sleep,
+            ens_id=self.ensemble.id_,
+            ee_uri=self._config.get_uri(),
+            ee_token=self._config.token,
+        )
 
     async def _publisher(self) -> None:
         heartbeat_interval = 0.1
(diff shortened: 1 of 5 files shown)


src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -104,6 +109,19 @@ def __init__(
 
         self._publisher_receiving_timeout: float = 60.0
         self._evaluation_result: asyncio.Future[bool] = asyncio.Future()
+        driver = create_driver(self.ensemble._queue_config.queue_options)
```
flikka: Perhaps just inline this below?

jonathan-eq: Not sure, it was not inlined before

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -540,6 +552,65 @@ def _get_ens_id(source: str) -> str:
...
+        except asyncio.CancelledError:
+            print("Cancelling evaluator task!")
+            return
+        # Dispatch final result from evaluator - SUCCEEDED or CANCELLED
+        if scheduler_finished_successfully:
+            await self.send_event(EnsembleSucceeded(ensemble=self.ensemble.id_))
+        else:
+            await self.send_event(EnsembleCancelled(ensemble=self.ensemble.id_))
+
+    async def send_event(self, event: EnsembleEvent) -> None:
```
xjules: this is not necessary. `_events` queue is owned by the evaluator so you can put there events directly.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -183,16 +201,15 @@ async def _send_terminate_message_to_dispatchers(self) -> None:
...
+            real_id = int(identity.decode("utf-8").split("-")[2])
+            self._scheduler.mark_job_as_being_killed_by_evaluator(real_id)
 
     async def _terminate_all_dispatchers(self) -> None:
-        if (scheduler := self.ensemble._scheduler) is not None:
-            await scheduler._running.wait()
+        await self._scheduler._running.wait()
         await self._send_terminate_message_to_dispatchers()
-        await self._ensemble.cancel()
+        await self.cancel()
```
xjules: This helper function is not needed anymore I think.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -540,6 +552,65 @@ def _get_ens_id(source: str) -> str:
...
+        # Dispatch final result from evaluator - SUCCEEDED or CANCELLED
+        if scheduler_finished_successfully:
+            await self.send_event(EnsembleSucceeded(ensemble=self.ensemble.id_))
+        else:
+            await self.send_event(EnsembleCancelled(ensemble=self.ensemble.id_))
+
+    async def send_event(self, event: EnsembleEvent) -> None:
+        await self._events.put(event)
+
+    async def cancel(self) -> None:
```
xjules: we can can directly: `await self._scheduler.kill_all_jobs()`; ie. this function can be removed.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -104,6 +109,19 @@ def __init__(
 
         self._publisher_receiving_timeout: float = 60.0
         self._evaluation_result: asyncio.Future[bool] = asyncio.Future()
+        driver = create_driver(self.ensemble._queue_config.queue_options)
+        self._scheduler = Scheduler(
```
xjules, line 111: Why is the scheduler created in the `__init__` constructor and not only when calling evaluate?

jonathan-eq: Why would we only have it when calling evaluate? Then we would still need the 
`if self._scheduler is not None: ...` everywhere

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: Lets do it, I don't see any reason why it shouldnt be

src/ert/ensemble_evaluator/evaluator.py
jonathan-eq: If not, we have to do `if self._scheduler is not None` checks four different places in evaluator. I think it is better the way it is now.

xjules (approved):
Nice PR @jonathan-eq ! :rocket: 
Squash the commits also.

PR #11440, opened 2025-08-12
Move responsibility to save parameters into Ensemble

xjules:
In order to make scalars more actionable, we need to move saving of parameters one level higher to get the proper abstraction.

**Approach**
ParameterConfig gets create_storage_datasets iterator, which prepares
the datasets (xr.Dataset or pl.Dataframe), which then gets saved into
the storage by the ensemble. This allows us to handle hybrid ways of
storing data; ie. either per realization or per ensemble.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/parameter_config.py
+++ b/src/ert/config/parameter_config.py
@@ -1,12 +1,14 @@
 from __future__ import annotations
 
 from abc import abstractmethod
+from collections.abc import Iterator
 from hashlib import sha256
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal
 
 import networkx as nx
 import numpy as np
+import polars as pl
 import xarray as xr
 from pydantic import BaseModel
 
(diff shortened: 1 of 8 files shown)


src/ert/config/parameter_config.py
```diff
@@ -85,15 +87,14 @@ def write_to_runpath(
...
-    ) -> None:
+    def gen_datasets_storage(
+        self, data: npt.NDArray[np.float64], iens_active_index: npt.NDArray[np.int_]
+    ) -> Iterator[tuple[pl.DataFrame | xr.Dataset, int | None]]:
         """
-        Save the parameter in internal storage for the given ensemble
+        Create an xarray Dataset or polars DataFrame from the data array
+        and the active index.
         """
+        raise NotImplementedError("create_dataset must be implemented in subclasses")
```
yngve-sk: I think this can just be removed since there is `@abstractmethod` annotation which will fail if a subclass fails to implement it

src/ert/config/parameter_config.py
```diff
@@ -85,15 +87,14 @@ def write_to_runpath(
         """
 
     @abstractmethod
-    def save_parameters(
-        self,
-        ensemble: Ensemble,
-        realization: int,
-        data: npt.NDArray[np.float64],
-    ) -> None:
+    def gen_datasets_storage(
```
yngve-sk: Too similar to `gen_data` 🧐 Would revise

xjules: `iter_storage_datasets` ?

yngve-sk: It should say that it creates datasets intended for storage, not sure if `iter_storage_datsets` imply that. `create_storage_datasets` + rename the argument `data` to `from_data` and invoke it with kwarg contention, it will read a bit more clear

yngve-sk (approved):
LGTM!

PR #11478, opened 2025-08-15
Fix some incorrect uses of import convention (pep328 and pep8)

eivindjahren:
As discussed ad nauseam, naming and imports should conform to common python conventions

https://peps.python.org/pep-0008/#public-and-internal-interfaces
https://peps.python.org/pep-0328/

      --exec 'just rapid-tests'`)

## When applicable


JHolba (approved):
Nice changes. Makes sense to import at package level instead of module level as it is more explicit what we consider the API.

SAKavli (approved):
I was not aware that this import convention is what is agreed upon in ERT, looks nice!
The change in `specific_warning_handler` makes a lot of sense, thanks for tidying up!

PR #11483, opened 2025-08-15
Print where everest runs

yngve-sk:
**Issue**
Resolves https://github.com/equinor/ert/issues/11453


src/everest/bin/everest_script.py
```diff
@@ -43,16 +44,47 @@
...
-            "You are running your optimization locally.\n"
-            "Pressing Ctrl+C will stop the optimization and exit."
-        )
+    client_machine_hostname = socket.gethostname()
+    server_queue_system = options.config.server.queue_system.name
+    simulator_queue_system = options.config.simulator.queue_system.name
+
+    server_info_str = "The optimization will be run by an experiment server on " + (
+        f"this machine ({client_machine_hostname}). "
+        f"Pressing Ctrl+C will stop the optimization and exit."
```
xjules, line 60: ~~So Ctrl+C will not work to stop the simulator when the everserver runs on LSF?~~ I can see down bellow that it will not stop it :)

src/everest/config/everest_config.py
```diff
@@ -834,21 +834,24 @@ def load_file(cls, config_file: str) -> Self:
...
-            "install_jobs": ert_config.installed_forward_model_steps,
-        }
-        activate_script = ErtPluginManager().activate_script()
-        if has_site_config:
-            context["queue_system"] = QueueConfig.from_dict(site_config).queue_options
-        if activate_script:
-            context["activate_script"] = activate_script
-        with init_context(context):
-            return cls(**config_dict)
+        with ErtPluginContext():
```
xjules, line 837: nice!

xjules (approved):
I think it is gooood! Just squash the commits once green.

PR #11484, opened 2025-08-15
Use property to simplify remember logic in remaining verification time

eivindjahren:
Also adds documentation of parameters and return value to log_warnings_from_file

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -80,13 +80,25 @@ class Job:
         self._requested_max_submit: int | None = None
         self._start_time: float | None = None
         self._end_time: float | None = None
-        self.remaining_file_verification_time = self.DEFAULT_FILE_VERIFICATION_TIMEOUT
-        self.remember_remaining_time()
+        self._remaining_file_verification_time = self.DEFAULT_FILE_VERIFICATION_TIMEOUT
+        self._previous_file_verification_time = self._remaining_file_verification_time
         self._started_killing_by_evaluator: bool = False
         self._was_killed_by_evaluator = asyncio.Event()
 
-    def remember_remaining_time(self) -> None:
-        self.previous_file_verification_time = self.remaining_file_verification_time
+    @property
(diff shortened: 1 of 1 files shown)


src/ert/scheduler/job.py
```diff
@@ -80,13 +80,25 @@ def __init__(self, scheduler: Scheduler, real: Realization) -> None:
...
+        return self._remaining_file_verification_time
+
+    @remaining_file_verification_time.setter
+    def remaining_file_verification_time(self, value: int) -> None:
+        self._previous_file_verification_time = self._remaining_file_verification_time
+        self._remaining_file_verification_time = value
+
+    def elapsed_time(self) -> int:
+        return (
+            self._previous_file_verification_time
```
SAKavli, line 99: I got curious now, should this have been 
```py
self.previous_file_verification_time
```
now that we have a property for this or is it fine?

src/ert/scheduler/job.py
```diff
@@ -183,11 +195,9 @@ async def _submit_and_run_once(self, sem: asyncio.BoundedSemaphore) -> None:
     async def log_time_spent_above_threshold_waiting_for_files(
         self, method_name: str
     ) -> None:
-        if self.previous_file_verification_time <= 0:
+        if self._previous_file_verification_time <= 0:
```
SAKavli, line 198: I guess the same applies here as well. Did we make the property just for the `setter` aspect of it, or should we also stick to the getter?

src/ert/scheduler/job.py
eivindjahren: Could be, but I think in this case the three functions that we added all go together in understanding the underlying values.

src/ert/scheduler/job.py
eivindjahren: I think we did it correctly in that elapsed_time and the file_verification_time getter/setters are the ones who need to deal with the state of the fields.

PR #11495, opened 2025-08-18
Ignore site config for everest tests

yngve-sk:
Fixes failing tests on komodo. Simulator queue config is picked up from site config, which is again copied into server config. If the config is serialized, it will now contain site config settings in server and simulator config, hence they must both be overwritten.

### DIFF
--- a/tests/everest/test_detached.py
+++ b/tests/everest/test_detached.py
@@ -235,7 +235,7 @@ def test_generate_queue_options_no_config():
         ),
     ],
 )
-def test_generate_queue_options_use_simulator_values(
+def test_that_server_queue_system_defaults_to_simulator_queue_options(
     monkeypatch, queue_class, expected_queue_kwargs
 ):
     monkeypatch.setattr(


tests/everest/test_everserver.py
```diff
@@ -173,12 +173,14 @@ async def test_status_exception(_, change_to_tmpdir, min_config):
...
 @pytest.mark.timeout(240)
 @patch("sys.argv", ["name", "--output-dir", "everest_output"])
+@pytest.mark.usefixtures("no_site_config")
 async def test_status_max_batch_num(copy_math_func_test_data_to_tmp):
     config = EverestConfig.load_file("config_minimal.yml")
     config_dict = {
         **config.model_dump(exclude_none=True),
         "optimization": {"algorithm": "optpp_q_newton", "max_batch_num": 1},
         "simulator": {"queue_system": {"name": "local"}},
+        "server": {"queue_system": {"name": "local"}},
```
berland: When we explicitly skip the site config, should not both simulator and server queue system default to "local"?

berland: I guess you can remove the line with "simulator" set to "local" as well.

tests/everest/test_detached.py
```diff
@@ -162,7 +162,7 @@ def test_detached_mode_config_base(min_config, monkeypatch, tmp_path):
     everest_config = EverestConfig.load_file("config.yml")
 
     # Expect it to take the default of the ERT QueueConfig
-    assert everest_config.simulator.queue_system == LocalQueueOptions(max_running=0)
+    assert everest_config.simulator.queue_system == LocalQueueOptions(max_running=8)
```
berland, line 165: Where does this 8 stem from?

yngve-sk: Default set within SimulatorConfig currently, ref
```

    @field_validator("queue_system", mode="before")
    @classmethod
    def default_local_queue(cls, v: Any, info: ValidationInfo) -> Any:
        if v is None:
            options = None
            if info.context:
                options = info.context.get("queue_system")
        return v
```

berland: This number 8 is probaly related to #11505

tests/everest/conftest.py
```diff
@@ -70,17 +71,17 @@ def make_config(forward_model_sleep_time: int = 1) -> Generator[str, None, None]
...
             job_path.write_text(
                 dedent(f"""#!/usr/bin/env python3
-                    import sys
-                    import os
-                    import time
-                    from pathlib import Path
+import sys
+import os
+import time
+from pathlib import Path
```
berland: Looks like a revert of: https://github.com/equinor/ert/commit/e9a7421a236687799b43efd7573ebfe3c4ed9755

If this is to be included, should be in a separate commit.

tests/everest/test_detached.py
```diff
@@ -227,37 +222,30 @@ def test_generate_queue_options_no_config():
...
-            },
-            TorqueQueueOptions(max_running=1, keep_qsub_output=True),
+            TorqueQueueOptions,
+            {"name": "torque", "keep_qsub_output": True, "max_running": 1},
         ),
     ],
 )
 def test_generate_queue_options_use_simulator_values(
-    queue_options, expected_result, monkeypatch
+    monkeypatch, queue_class, queue_kwargs
```
berland: why not use the string "expected" in the args? It tends to make things more readable.

tests/everest/test_detached.py
```diff
@@ -227,37 +222,30 @@ def test_generate_queue_options_no_config():
...
-                "name": "torque",
-                "keep_qsub_output": "1",
-            },
-            TorqueQueueOptions(max_running=1, keep_qsub_output=True),
+            TorqueQueueOptions,
+            {"name": "torque", "keep_qsub_output": True, "max_running": 1},
         ),
     ],
 )
 def test_generate_queue_options_use_simulator_values(
```
berland: I think "server" is missing from this test name. It seems we are testing the feature of copying simulator settings to server settingsl

PR #11512, opened 2025-08-19
Fix deprecation tests

eivindjahren:
      --exec 'just rapid-tests'`)

## When applicable


tests/ert/unit_tests/config/parsing/test_config_schema_deprecations.py
```diff
@@ -12,28 +12,30 @@
...
+def test_is_angle_bracketed_detects_words_surrounded_by_brackets():
     assert DeprecationInfo.is_angle_bracketed("<KEY>")
     assert not DeprecationInfo.is_angle_bracketed("KEY")
     assert not DeprecationInfo.is_angle_bracketed("K<E>Y")
     assert not DeprecationInfo.is_angle_bracketed("")
 
 
 @pytest.mark.parametrize("kw", JUST_REMOVE_KEYWORDS)
-def test_that_suggester_gives_simple_migrations(kw):
+def test_that_a_warning_is_shown_for_keys_that_can_just_be_removed(kw):
```
SAKavli, line 23: I feel like using the word "just" here can cause some ambiguity. Would the test intention be clearer and more concise without it?

tests/ert/unit_tests/config/parsing/test_config_schema_deprecations.py
```diff
@@ -42,57 +44,59 @@ def test_that_suggester_gives_gen_kw_migrations(kw):
...
         ConfigWarning, match=f"The {kw} keyword has been removed. For most cases "
     ):
         ErtConfig.from_file_contents(f"NUM_REALIZATIONS 1\n{kw}\n")
 
 
-def test_that_suggester_gives_refcase_list_migration():
+def test_that_a_deprecation_message_is_shown_when_the_refcase_list_keyword_is_used():
     with pytest.warns(
         ConfigWarning,
         match="The corresponding plotting functionality was removed in 2015",
```
SAKavli, line 67: I am curious, does it make sense to parameterize these tests? Perhaps if the exact match on deprecation message is not so important?

tests/ert/unit_tests/test_run_path_creation.py
```diff
@@ -562,62 +562,135 @@ def test_num_cpu_subst(append, numcpu, make_run_path):
...
+            tmpdir + "/simulations/realization-3/ITER-0",
+        ]
+
+
+@pytest.mark.filterwarnings(
+    "ignore:.*RUNPATH keyword contains deprecated "
+    "value placeholders.*:ert.config.ConfigWarning"
+)
+@pytest.mark.parametrize("iens_placeholder", [("%d"), ("<IENS>")])
+def test_that_runpaths_with_iens_only_will_substituted_with_just_iens_index(
```
SAKavli: Is there missing a "be" here? "only will be substituted with ..."
Also, does "only" and "just" have duplicate meaning in this sentence?
Suggestion:
`test_that_runpaths_with_iens_will_only_be_substituted_with_iens_index`

tests/ert/unit_tests/test_run_path_creation.py
SAKavli, line 1: I completely agree with your changes to this file, this is much more readable than the old parameterization.

tests/ert/unit_tests/test_run_path_creation.py
```diff
@@ -562,62 +562,135 @@ def test_num_cpu_subst(append, numcpu, make_run_path):
...
-        NUM_REALIZATIONS 1
-        RUNPATH {run_path}
-        """
-    )
-    if expected_raise:
-        with pytest.raises(ConfigValidationError, match=f"{msg}.*{run_path}`."):
-            _ = ErtConfig.from_file_contents(config_text)
-    else:
-        _ = ErtConfig.from_file_contents(config_text)
+def test_that_mixing_printf_format_placeholders_and_bracketed_is_invalid(runpath):
```
SAKavli, line 636: Could you explain what we are trying to test here? Or why this should be invalid? I didn't quite get it from the title. Could just be skill issue from my side.

tests/ert/unit_tests/config/parsing/test_config_schema_deprecations.py
eivindjahren: The idea behind the "just" was that there was no reason to not remove, and no other consideration is required. Other suitable words are "simply" and "obviously". Any preference on word?

tests/ert/unit_tests/config/parsing/test_config_schema_deprecations.py
eivindjahren: It sort of makes sense, but I think it doesn't add much value in this case and introduces some confusion in the error message from pytest on failed tests. You would now get the following message from pytest:
```
FAILED test_config_schema_deprecations.py::test_that_a_deprecation_message_is_shown_when_the_refcase_list_keyword_is_used
```

while if you used parametrize you get

```
test_config_schema_deprecations.py::test_that_a_deprecation_message_is_shown_when_a_deprecated_keyword_is_used[REFCASE_LIST, The corresponding plotting functionality was removed in 2015]
```

tests/ert/unit_tests/test_run_path_creation.py
eivindjahren: Yes, you are correct, I did another take on the test and came up with `test_that_runpaths_with_just_iens_will_be_substituted_with_just_iens_index`

tests/ert/unit_tests/test_run_path_creation.py
eivindjahren: So I am completely missing the rationale for the choice here but the idea is that you cannot mix and match %d and <ITER> styles. Ie. this is invalid `"simulations/realization-%d/iter-<ITER>"`. If you have a suggestion for a more informative name, then go ahead :)

Issue #11477, opened 2025-08-14
Save serialized runmodels in storage


yngve-sk:
For all ERT runmodels and the everest runmodel, storage should contain the serialized runmodel in the `index.json` of an experiment. Furthermore, the metadata.json file should be deprecated and replaced by reading from the serialized runmodel.

yngve-sk:
Blocked by https://github.com/equinor/ert/issues/10972

PR #11513, opened 2025-08-19
Save serialized experiment config in index.json

yngve-sk:
Goes in with https://github.com/equinor/semeio/pull/866
**Issue**
Resolves https://github.com/equinor/ert/issues/11477

Made experiment config be cached, it is a performance hog, ref run [link] https://github.com/equinor/ert/issues/12804(https://codspeed.io/equinor/ert/branches/yngve-sk%3A25.08.save-runmodel-configs-in-storage?uri=tests%2Fert%2Fperformance_tests%2Ftest_obs_and_responses_performance.py%3A%3Atest_time_performance_of_joining_observations_and_responses%5Bsetup_benchmark0%5D&runnerMode=Simulation&utm_source=github&utm_medium=check&utm_content=benchmark&page=1)

Followup issues:
* Streamline / clarify / restructure cached getting of responses, parameters, observations from storage. Now it seems a bit ad-hoc, and is a potential performance hog especially if something is called many times with high frequency. https://github.com/equinor/ert/issues/12802
* Make LocalExperiment.experiment_config typed - split out experiment configs into own file so they can be imported from ERT storage and run models without introducing circular imports  https://github.com/equinor/ert/issues/12803

### DIFF
--- a/src/ert/gui/tools/manage_experiments/storage_widget.py
+++ b/src/ert/gui/tools/manage_experiments/storage_widget.py
@@ -22,8 +22,7 @@ from PyQt6.QtWidgets import (
     QWidget,
 )
 
-from ert.config import ErrorInfo, ErtConfig, RFTConfig
-from ert.config._create_observation_dataframes import create_observation_dataframes
+from ert.config import ErrorInfo, ErtConfig
 from ert.gui.ertnotifier import ErtNotifier
 from ert.gui.ertwidgets import CreateExperimentDialog, Suggestor
 from ert.storage import Ensemble, Experiment
@@ -170,26 +169,29 @@ class StorageWidget(QWidget):
         if create_experiment_dialog.exec():
             try:
                 with self._notifier.write_storage() as storage:
-                    response_configs = (
+                    parameter_configuration = (
+                        self._ert_config.parameter_configurations_with_design_matrix
+                    )
+                    response_configuration = (
                         self._ert_config.ensemble_config.response_configuration
                     )
-
-                    rft_config = next(
-                        (r for r in response_configs if r.type == "rft"),
-                        None,
-                    )
-                    if rft_config is not None:
-                        rft_config = cast(RFTConfig, rft_config)
-
                     ensemble = storage.create_experiment(
-                        parameters=self._ert_config.parameter_configurations_with_design_matrix,
-                        responses=response_configs,
-                        observations=create_observation_dataframes(
-                            observations=self._ert_config.observation_declarations,
-                            rft_config=rft_config,
-                        ),
+                        experiment_config={
+                            "parameter_configuration": [
+                                c.model_dump(mode="json")
+                                for c in parameter_configuration
+                            ],
+                            "response_configuration": [
+                                c.model_dump(mode="json")
+                                for c in response_configuration
+                            ],
+                            "observations": [
+                                d.model_dump(mode="json")
+                                for d in self._ert_config.observation_declarations
+                            ],
+                            "ert_templates": self._ert_config.ert_templates,
+                        },
                         name=create_experiment_dialog.experiment_name,
-                        templates=self._ert_config.ert_templates,
                     ).create_ensemble(
                         name=create_experiment_dialog.ensemble_name,
                         ensemble_size=self._ensemble_size,
--- a/src/ert/storage/local_experiment.py
+++ b/src/ert/storage/local_experiment.py
@@ -24,7 +24,10 @@ from ert.config import (
     SurfaceConfig,
 )
 from ert.config import Field as FieldConfig
-from ert.config.parsing.context_values import ContextBoolEncoder
+from ert.config._create_observation_dataframes import (
+    create_observation_dataframes,
+)
+from ert.config._observations import Observation
 
 from .mode import BaseMode, Mode, require_write
 
@@ -54,6 +57,7 @@ class _Index(BaseModel):
     # from a different experiment. For example, a manual update
     # is a separate experiment from the one that created the prior.
     ensembles: list[UUID]
+    experiment: dict[str, Any] = {}
     status: ExperimentStatus | None = Field(default=None)
 
 
@@ -80,9 +84,6 @@ class LocalExperiment(BaseMode):
     arguments. Provides methods to create and access associated ensembles.
     """
 
-    _parameter_file = Path("parameter.json")
-    _responses_file = Path("responses.json")
-    _metadata_file = Path("metadata.json")
     _templates_file = Path("templates.json")
     _index_file = Path("index.json")
 
@@ -118,63 +119,21 @@ class LocalExperiment(BaseMode):
         storage: LocalStorage,
         uuid: UUID,
         path: Path,
-        *,
-        parameters: list[ParameterConfig] | None = None,
-        responses: list[ResponseConfig] | None = None,
-        observations: dict[str, pl.DataFrame] | None = None,
-        simulation_arguments: dict[Any, Any] | None = None,
+        experiment_config: dict[str, Any],
         name: str | None = None,
-        templates: list[tuple[str, str]] | None = None,
     ) -> LocalExperiment:
-        """
-        Create a new LocalExperiment and store its configuration data.
-
-        Parameters
-        ----------
-        storage : LocalStorage
-            Storage instance for experiment creation.
-        uuid : UUID
-            Unique identifier for the new experiment.
-        path : Path
-            File system path for storing experiment data.
-        parameters : list of ParameterConfig, optional
-            List of parameter configurations.
-        responses : list of ResponseConfig, optional
-            List of response configurations.
-        observations : dict of str to encoded observation datasets, optional
-            Observations dictionary.
-        simulation_arguments : SimulationArguments, optional
-            Simulation arguments for the experiment.
-        name : str, optional
-            Experiment name. Defaults to current date if None.
-        templates : list of tuple[str, str], optional
-            Run templates for the experiment. Defaults to None.
-
-        Returns
-        -------
-        local_experiment : LocalExperiment
-            Instance of the newly created experiment.
-        """
         if name is None:
             name = datetime.today().isoformat()
 
         storage._write_transaction(
             path / cls._index_file,
-            _Index(id=uuid, name=name, ensembles=[])
+            _Index(id=uuid, name=name, ensembles=[], experiment=experiment_config)
             .model_dump_json(indent=2, exclude_none=True)
             .encode("utf-8"),
         )
 
-        parameter_data = {}
-        for parameter in parameters or []:
-            parameter.save_experiment_data(path)
-            parameter_data.update({parameter.name: parameter.model_dump(mode="json")})
-        storage._write_transaction(
-            path / cls._parameter_file,
-            json.dumps(parameter_data, indent=2).encode("utf-8"),
-        )
-
-        if templates:
+        templates = experiment_config.get("ert_templates")
+        if templates is not None:
             templates_path = path / "templates"
             templates_path.mkdir(parents=True, exist_ok=True)
             templates_abs: list[tuple[str, str]] = []
@@ -191,27 +150,29 @@ class LocalExperiment(BaseMode):
                 json.dumps(templates_abs).encode("utf-8"),
             )
 
-        response_data = {}
-        for response in responses or []:
-            response_data.update({response.type: response.model_dump(mode="json")})
-        storage._write_transaction(
-            path / cls._responses_file,
-            json.dumps(response_data, default=str, indent=2).encode("utf-8"),
-        )
-
-        if observations:
+        observation_declarations = experiment_config.get("observations")
+        if observation_declarations:
             output_path = path / "observations"
-            output_path.mkdir()
-            for response_type, dataset in observations.items():
-                storage._to_parquet_transaction(
-                    output_path / f"{response_type}", dataset
-                )
+            output_path.mkdir(parents=True, exist_ok=True)
 
-        simulation_data = simulation_arguments or {}
-        storage._write_transaction(
-            path / cls._metadata_file,
-            json.dumps(simulation_data, cls=ContextBoolEncoder).encode("utf-8"),
-        )
+            responses_list = experiment_config.get("response_configuration", [])
+            rft_config_json = next(
+                (r for r in responses_list if r.get("type") == "rft"), None
+            )
+            rft_config = (
+                _responses_adapter.validate_python(rft_config_json)
+                if rft_config_json is not None
+                else None
+            )
+
+            obs_adapter = TypeAdapter(Observation)  # type: ignore
+            obs_objs: list[Observation] = []
+            for od in observation_declarations:
+                obs_objs.append(obs_adapter.validate_python(od))
+
+            datasets = create_observation_dataframes(obs_objs, rft_config)
+            for response_type, df in datasets.items():
+                storage._to_parquet_transaction(output_path / response_type, df)
 
         return cls(storage, path, Mode.WRITE)
 
@@ -286,15 +247,9 @@ class LocalExperiment(BaseMode):
         raise KeyError(f"Ensemble with name '{name}' not found")
 
     @property
-    def metadata(self) -> dict[str, Any]:
-        path = self.mount_point / self._metadata_file
-        if not path.exists():
-            raise ValueError(f"{self._metadata_file!s} does not exist")
-        return json.loads(path.read_text(encoding="utf-8"))
-
-    @property
     def relative_weights(self) -> str:
-        return self.metadata.get("weights", "")
+        assert self.experiment_config is not None
+        return self.experiment_config.get("weights", "")
 
     @property
     def name(self) -> str:
@@ -324,9 +279,8 @@ class LocalExperiment(BaseMode):
 
     @property
     def parameter_info(self) -> dict[str, Any]:
-        return json.loads(
-            (self.mount_point / self._parameter_file).read_text(encoding="utf-8")
-        )
+        parameters_list = self.experiment_config.get("parameter_configuration", [])
+        return {parameter["name"]: parameter for parameter in parameters_list}
 
     @property
     def templates_configuration(self) -> list[tuple[str, str]]:
@@ -348,9 +302,8 @@ class LocalExperiment(BaseMode):
 
     @property
     def response_info(self) -> dict[str, Any]:
-        return json.loads(
-            (self.mount_point / self._responses_file).read_text(encoding="utf-8")
-        )
+        responses_list = self.experiment_config.get("response_configuration", [])
+        return {response["type"]: response for response in responses_list}
 
     def get_surface(self, name: str) -> IrapSurface:
         """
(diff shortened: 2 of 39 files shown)


yngve-sk:
Blocked by https://github.com/equinor/ert/issues/11673

src/ert/config/workflow_job.py
```diff
@@ -102,6 +104,7 @@ class _WorkflowJob(BaseModelWithContextSupport, ABC):
     max_args: int | None = None
     arg_types: list[SchemaItemType] = field(default_factory=list)
     stop_on_fail: bool = False
+    source_file: str | None = None
```
berland: Should we mention the special meaning of `None` here too?

berland: (that it it is available through a plugin)

yngve-sk: Fixed

src/ert/storage/migration/to15.py
```diff
@@ -0,0 +1,27 @@
+import json
+from pathlib import Path
+
+info = "Migrate metadata.json contents to be in experiment index.json .experiment field"
+
+
+def migrate_metadata_json_into_experiment_index(path: Path) -> None:
+    for experiment_path in path.glob("experiments/*"):
+        experiment_json = {}
+        with open(experiment_path / "metadata.json", encoding="utf-8") as fin:
```
berland: This should trigger a ruff lint issue for not using pathlib. Maybe you need to rebase on main.

src/ert/run_models/ert_runmodel_configs.py
```diff
@@ -0,0 +1,239 @@
...
+            runtime_plugins.inject_installed_workflow_jobs(user_installed_workflow_jobs)
+
+        return values
+
+    @field_serializer("user_installed_workflow_jobs")
+    def serialize_workflow_jobs(
+        self, user_installed_workflow_jobs: dict[str, WorkflowJob]
+    ) -> dict[str, str]:
+        return {
+            k: v.source_file
```
berland: Can there be issues with relative/absolute path here? (I have not thought it through)

berland: Ert should be opinionated whether it wants to preserve the path as-is, or if it need to make it e.g. absolute for this to be robust.

tests/ert/unit_tests/run_models/test_model_factory.py
```diff
@@ -334,9 +334,9 @@ def test_evaluate_ensemble_paths(
...
                 realizations=None,
             ),
-            MagicMock(),
+            queue.SimpleQueue(),
         )
 
-    base_path = tmp_path / "simulations"
-    expected_path = [str(base_path / expected) for expected in expected_path]
-    assert set(model.paths) == set(expected_path)
+        base_path = tmp_path / "simulations"
```
berland: was indenting this needed in order to have the patch on get_ensemble active?

src/ert/run_models/ert_runmodel_configs.py
```diff
@@ -0,0 +1,239 @@
...
+    log_path: Path
+    random_seed: int
+    start_iteration: int = 0
+    minimum_required_realizations: int = 0
+    supports_rerunning_failed_realizations: ClassVar[bool] = False
+    user_installed_workflow_jobs: dict[str, WorkflowJob]
+    user_installed_fm_steps: dict[str, ForwardModelStep]
+
+    # Note: Must be model validator to be invoked before other "before"
+    # field validators downstream
```
oyvindeide: Can we reconsider the design to avoid this? You have to do this because of:
```python
        user_installed_fm_steps = {
            k: v
            for k, v in config.installed_forward_model_steps.items()
            if k not in runtime_plugins.installed_forward_model_steps and v is not None
        }
```
?
Feels like it should be possible to do it in a more intuitive way

src/ert/storage/migration/to15.py
yngve-sk: Hmm, rebased, but this already uses pathlib? path is a `pathlib.Path` so the glob results also are?

tests/ert/unit_tests/run_models/test_model_factory.py
yngve-sk: Seems not, I dedented it, mock should only be needed for `create_model` invocation

src/ert/run_models/ert_runmodel_configs.py
yngve-sk: I think absolute path is closer to what we want? I also think someone suggested just storing the entire workflow/forward model file in storage and point to it there. Absolute paths will have some implicaitons wrt trying to move projects/renaming folders too..

src/ert/run_models/ert_runmodel_configs.py
oyvindeide: And I guess this happens because we resolve the executable in the config level? Instead of doing it later?

yngve-sk: I think it would require solving https://github.com/equinor/ert/issues/11983 , main issue is they are merged into one thing in ErtConfig, then we have to separate it again.

yngve-sk: TLDR summary of discussion on this (will likely try some of these in separate PRs)

This can be made more clean by doing one or more of:
* Split context validation out for runmodels
* Separate context of site installed vs user installed workflow jobs / forward model steps
* Forward model steps can be serialized as-is -> should not need to validate them against installed / not installed for runmodels
* Workflow jobs still need to be referenced somehow, to be reconstructed outside of ErtConfig-context, still a bit up for discussion how that can be done.

src/ert/storage/local_experiment.py
```diff
@@ -160,25 +150,17 @@ def create(
...
-        )
-
-        if templates:
+        templates = experiment_config.get("ert_templates")
+        if templates is not None:
             templates_path = path / "templates"
             templates_path.mkdir(parents=True, exist_ok=True)
             templates_abs: list[tuple[str, str]] = []
-            for idx, (src, dst) in enumerate(templates):
+            for idx, (src, dst) in enumerate(experiment_config["ert_templates"]):
```
andreas-el: Can't you use the `templates` variable you already extracted "ert_templates" into?

src/ert/storage/local_storage.py
```diff
@@ -21,17 +21,18 @@
 from pydantic import BaseModel, Field
 
 import ert.storage
-from ert.config import ErtConfig, ParameterConfig, ResponseConfig
+from ert.config import ErtConfig
 from ert.shared import __version__
 
 from .local_ensemble import LocalEnsemble
 from .local_experiment import LocalExperiment
+from .migration import to20
```
andreas-el: Wasn't `to20` removed? And could you explain shortly why?

src/ert/storage/local_storage.py
berland, line 333: This docstring is now outdated (it tells marginally more than the code, so it can probably be removed)

tests/ert/unit_tests/test_run_path_creation.py
```diff
@@ -1009,8 +1029,11 @@ def test_that_contents_of_gridfile_is_logged(storage, caplog):
     )
 
     experiment_id = storage.create_experiment(
-        parameters=config.ensemble_config.parameter_configuration,
-        responses=config.ensemble_config.response_configuration,
+        experiment_config={
+            "parameter_configuration": config.ensemble_config.parameter_configuration,
+            "response_configuration": config.ensemble_config.response_configuration,
+            "ert_templates": config.ert_templates,
```
andreas-el: Needed?

src/ert/gui/tools/manage_experiments/storage_widget.py
```diff
@@ -170,26 +169,29 @@ def _addItem(self) -> None:
...
-                        rft_config = cast(RFTConfig, rft_config)
-
                     ensemble = storage.create_experiment(
-                        parameters=self._ert_config.parameter_configurations_with_design_matrix,
-                        responses=response_configs,
-                        observations=create_observation_dataframes(
-                            observations=self._ert_config.observation_declarations,
-                            rft_config=rft_config,
-                        ),
+                        experiment_config={
```
berland, line 179: Should this `experiment_config` have been a Pydantic object? (could be a issue to solve later)

tests/ert/ui_tests/cli/test_field_parameter.py
```diff
@@ -432,10 +432,20 @@ def test_field_param_update_using_heat_equation_zero_var_params_and_adaptive_loc
         init_temp_scale = prior.load_parameters("INIT_TEMP_SCALE")
         corr_length = prior.load_parameters("CORR_LENGTH")
 
+        obs_decls = config.observation_declarations
```
andreas-el: Perhaps inline this?

src/ert/storage/local_storage.py
yngve-sk: Oops! fixed

src/ert/storage/local_storage.py
yngve-sk: Indeed, quite a lot of the boilerplate/CRUD methods in storage have these types of comments, suggest removing separate from this PR though?

tests/ert/unit_tests/test_run_path_creation.py
yngve-sk: Seems not, removing

tests/ert/ui_tests/gui/test_manage_experiments_tool.py
```diff
@@ -53,11 +55,19 @@ def test_design_matrix_in_manage_experiments_panel(
     assert config.analysis_config.design_matrix is not None
 
     with notifier.write_storage() as storage:
+        design_parameters = (
+            config.analysis_config.design_matrix.parameter_configurations
+        )
```
andreas-el: Perhaps inline?

src/ert/gui/tools/manage_experiments/storage_widget.py
yngve-sk: It could be indeed a union between all experiment configs, suggest doing a property parsing the JSON to an instance of an experiment config. Suggest deferring to later, doing this would require us to split out all runmodel configs into an own separate file so that it can be imported from ERT storage AND runmodels without causing a circular import (doing it now would give circular imports). But it is very easy to do and would improve readability, could be picked up after this and resolved fast I think.

tests/ert/unit_tests/plugins/test_export_misfit.py
```diff
@@ -10,10 +8,10 @@
...
 
 
-@pytest.mark.skipif(
-    sys.platform.startswith("darwin"),
-    reason="https://github.com/equinor/ert/issues/7533",
-)
+# @pytest.mark.skipif(
+#    sys.platform.startswith("darwin"),
+#    reason="https://github.com/equinor/ert/issues/7533",
+# )
```
andreas-el: Should remove this is if these tests do not hang/fail on Mac anymore.

yngve-sk: Indeed, will do in separate PR and check. AFAIK we only run mac tests on main, so this is untested so far, quite separate issue though.

src/ert/storage/local_storage.py
berland: As this PR renders it incorrect, at least this one should be removed now.

tests/ert/unit_tests/storage/test_local_ensemble.py
```diff
@@ -178,13 +188,15 @@ def test_that_load_scalar_keys_raises_index_error_for_missing_realizations(tmp_p
...
-                    group="group1",
-                    distribution={"name": "uniform", "min": 0, "max": 1},
-                ),
-            ]
+            experiment_config={
+                "parameter_configuration": [
+                    GenKwConfig(
+                        name="param1",
+                        group="group1",
+                        distribution={"name": "uniform", "min": 0, "max": 1},
```
andreas-el, line 196: Looks like you're missing;
```
.model_dump(mode="json"),
```

andreas-el: Could this single parameter be set globally in the test, and rather referenced? Seems to be a lot of duplicates of this code.

yngve-sk: It could be but I think it still makes sense to have the declaration inlined instead of adding one level of indirection, even though it is a default value. Prefer to defer this discussion to another issue, if it is to be fixed up.

tests/ert/unit_tests/test_load_forward_model.py
```diff
@@ -131,7 +136,12 @@ def test_load_forward_model_summary(
         """
     )
     experiment_id = storage.create_experiment(
-        responses=ert_config.ensemble_config.response_configuration
+        experiment_config={
+            "response_configuration": [
+                rc.model_dump(mode="json")
+                for rc in ert_config.ensemble_config.response_configuration
+            ]
```
andreas-el, line 143: I can't help to think that this could have been a helper function instead, e.g. `get_response_configuration_json` which returns this dict entry?

tests/ert/unit_tests/test_load_forward_model.py
```diff
@@ -290,7 +305,11 @@ def test_loading_from_any_available_iter(storage, run_args, itr):
     )
     prior_ensemble = storage.create_ensemble(
         storage.create_experiment(
-            responses=ert_config.ensemble_config.response_configuration
+            experiment_config={
+                "response_configuration": (
+                    ert_config.ensemble_config.response_configuration
+                ),
+            }
```
andreas-el, line 313: This is a bit confusing; why does this differ from the other experiment_config ~50 lines up?
i.e.
```
         experiment_config={
                "response_configuration": [
                    rc.model_dump(mode="json")
                    for rc in ert_config.ensemble_config.response_configuration
                ],
            }
```

tests/ert/unit_tests/gui/tools/plot/test_plot_api.py
```diff
@@ -510,26 +563,54 @@ def test_that_response_keys_do_not_match_keys_that_are_substrings(
...
+                    "type": "summary_observation",
+                    "name": "FOPR",
+                    "key": "FOPR",
+                    "date": date.isoformat(),
+                    "value": 0.01,
+                    "error": 0.2,
+                    "east": None,
+                    "north": None,
+                    "radius": None,
+                },
```
andreas-el: Potential for doing a `zip` like you've done elsewhere.
Seemingly only name (+key), value & error changes here.

tests/ert/unit_tests/test_load_forward_model.py
yngve-sk: Nice catch! This should have a `model_dump(mode="json")` over each response config yes

tests/ert/unit_tests/test_load_forward_model.py
yngve-sk: It could have been but prefer to defer this simplification for future work, that helper would not be used outside of tests (in the current state of the code).

tests/ert/unit_tests/analysis/test_es_update.py
```diff
@@ -29,24 +30,40 @@ def uniform_parameter():
...
+        },
+        {
+            "type": "general_observation",
+            "name": "OBSERVATION",
+            "data": "RESPONSE",
+            "restart": 0,
+            "index": 2,
+            "value": 1.0,
+            "error": 10.0,
+        },
```
andreas-el: Could be something like;
```
errors = [0.1, 1.0, 10.0]

observations = [
    {
        "type": "general_observation",
        "name": "OBSERVATION",
        "data": "RESPONSE",
        "restart": 0,
        "index": i,
        "value": 1.0,
        "error": e,
    }
    for i, e in enumerate(errors)
]
```

andreas-el: If you decide to go this, there are other places this is applicable too.

tests/ert/unit_tests/analysis/test_es_update.py
```diff
@@ -558,11 +555,14 @@ def test_smoother_snapshot_alpha(
 
     # alpha is a parameter used for outlier detection
 
-    resp = GenDataConfig(keys=["RESPONSE"])
+    resp = GenDataConfig(keys=["RESPONSE"]).model_dump(mode="json")
```
andreas-el: Could change the variable name to reflect content better.
E.g. response_json or something similar.

tests/ert/unit_tests/storage/migration/test_to24.py
```diff
@@ -0,0 +1,225 @@
...
+    )
+
+    (exp_path / "index.json").write_text(
+        json.dumps({"id": "test-id", "name": "test-experiment", "ensembles": []}),
+        encoding="utf-8",
+    )
+
+    responses = {
+        "summary": {"type": "summary", "name": "summary", "keys": ["FOPR"]},
+        "gen_data": {"type": "gen_data", "name": "gen_data", "keys": ["SOME"]},
```
andreas-el: If possible, have different names from types. This is a common switchup to do, and this test won't catch that.

yngve-sk:
> You've touched how RFTs are handled (at least some). Do we have tests that will catch errors that could have been introduced by your changes concerning RFTs?

There are / should be no changes to how RFTs are handled in this PR. 
* In ERTConfig there is still a step that invokes the `create_observation_dataframes` which again triggers a creation of an `RFTConfig` that is passed onto runmodel. It also edits the existing RFT config

After this, the observation declarations and response configs should be exactly the same as before.
When it hits the runmodel / storage, that same RFT config is fetched out from the response config and passed as an argument to `_create_observation_dataframes`

tests/ert/unit_tests/storage/migration/test_to24.py
yngve-sk: Added some overlapping response keys

tests/ert/unit_tests/test_summary_response.py
```diff
@@ -41,7 +41,12 @@ def test_load_summary_response_restart_not_zero(
         ert_config = ErtConfig.from_file("config.ert")
 
         experiment_id = storage.create_experiment(
-            responses=ert_config.ensemble_config.response_configuration
+            experiment_config={
+                "response_configuration": [
```
berland, line 45: This also passes with `"response_configuration": ert_config.ensemble_config.response_configuration`

yngve-sk: Yes, though we should pass JSON as it respects the separation of storage vs runmodel (same python context now but should be possible to have them separate). Should/could be revisited when adding types to the experiment config though, as the pydantic typing can be directly translated into API docs

src/ert/storage/local_experiment.py
```diff
@@ -82,7 +86,6 @@ class LocalExperiment(BaseMode):
 
     _parameter_file = Path("parameter.json")
     _responses_file = Path("responses.json")
```
berland: Should this `_responses_file` also be removed? It is only used in `test_storage_migration.py` where it can probably be used as a hardcoded string.

tests/ert/unit_tests/test_load_forward_model.py
```diff
@@ -23,7 +23,12 @@ def func(config_text):
         ert_config = ErtConfig.from_file_contents(config_text)
         prior_ensemble = storage.create_ensemble(
             storage.create_experiment(
-                responses=ert_config.ensemble_config.response_configuration
+                experiment_config={
+                    "response_configuration": [
+                        rc.model_dump(mode="json")
+                        for rc in ert_config.ensemble_config.response_configuration
```
berland, line 29: This works with the simpler
```
            storage.create_experiment(
                experiment_config={
                    "response_configuration": ert_config.ensemble_config.response_configuration
                }
```

berland: Is the model_dump() added to ensure everything is dicts and not special objects?

yngve-sk: Same reply as in https://github.com/equinor/ert/pull/11513#discussion_r2762875738

src/ert/storage/local_experiment.py
yngve-sk: Removed it

PR #11537, opened 2025-08-21
Add setting summary log statement to run_model

### LINKED ISSUE #11518: Ensure active realizations is logged in the first chunk in run_model
To make queries regarding parameters, realizations and active realizations simultaneously, we need all of this information to be in the same log chunk.

SAKavli:
Logs are chunked if their message becomes too long. To be able to query logs for conditions based on multiple of these values, we need to ensure they are in the same log chunk.

This will make a short summary of properties (to be amended?) we wish to query for in the first chunk of the log.

**Issue**
Resolves #11518 

Log examples:
image
image

Note: `evaluate_ensemble` have `num_parameters: NA` because `EvaluateEnsemble` doesn't have the `parameter_config` property as it's not inheriting `InitialEnsembleRunModel`

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/run_models/run_model.py
+++ b/src/ert/run_models/run_model.py
@@ -230,16 +230,25 @@ class RunModel(BaseModel, ABC):
             for key, value in self.__dict__.items()
             if key not in keys_to_drop
         }
-        num_params_log = "with"
-        if hasattr(self, "parameter_configuration"):
-            num_params = sum(
+        settings_summary = {
+            "run_model": self.name(),
+            "num_realizations": self.runpath_config.num_realizations,
+            "num_active_realizations": self.active_realizations.count(True),
+            "num_parameters": sum(
                 len(param_config.parameter_keys)
                 for param_config in self.parameter_configuration
             )
-            num_params_log = f"with {num_params} parameters and"
+            if hasattr(self, "parameter_configuration")
+            else "NA",
+            "localization": getattr(
+                settings_dict.get("analysis_settings", {}), "localization", "NA"
+            ),
+        }
 
         logger.info(
-            f"Running '{self.name()}' {num_params_log} settings {settings_dict}"
+            f"Running '{self.name()}'\n\n"
+            f"Settings summary: {settings_summary}\n\n"
+            f"Settings: {settings_dict}"
         )
 
     @classmethod
--- a/tests/ert/unit_tests/run_models/test_base_run_model.py
+++ b/tests/ert/unit_tests/run_models/test_base_run_model.py
@@ -624,7 +624,7 @@ def test_run_model_logs_number_of_parameters(use_tmpdir):
     rm = create_run_model(parameter_configuration=[parameters])
 
     def mock_logging(_, log_str):
-        regex = r"Running '\w+' with (\d+) parameters"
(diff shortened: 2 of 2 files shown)


src/ert/run_models/run_model.py
```diff
@@ -230,16 +230,22 @@ def log_at_startup(self) -> None:
...
         }
-        num_params_log = "with"
-        if hasattr(self, "parameter_configuration"):
-            num_params = sum(
+        summary_dict = {
+            "run_model": self.name(),
+            "num_realizations": self.runpath_config.num_realizations,
+            "num_active_realizations": self.active_realizations.count(True),
+            "num_parameters": sum(
                 len(param_config.parameter_keys)
```
lucaeg, line 238: Why would you typically want to query these in the logs?

src/ert/run_models/run_model.py
```diff
@@ -230,16 +230,22 @@ def log_at_startup(self) -> None:
...
-            num_params_log = f"with {num_params} parameters and"
+            if hasattr(self, "parameter_configuration")
+            else "NA",
+        }
 
         logger.info(
-            f"Running '{self.name()}' {num_params_log} settings {settings_dict}"
+            f"Running '{self.name()}'\n\n"
+            f"Settings summary: {summary_dict}\n\n"
+            f"Settings: {settings_dict}"
```
lucaeg, line 251: Conceptually, what is the difference between the settings dict and the summary dict?

lucaeg (commented):
Looks good to me! I added two general questions :)

src/ert/run_models/run_model.py
SAKavli: @Blunde1  was interested in how our users use certain run models, especially ensemble smoother (ES) and ES multiple data assimilation (ES-MDA) with and without adaptive localization.
There are some use cases where some correlations between number of realizations and number of parameters can be considered as user errors, giving bad simulation results.
I am looking into creating a dashboard to display these stats, then I query the logs on multiple dimensions in the same query, for example:

```kusto
Apptraces
| where Message contains "run_model: ensemble_smoother"
| extend NumParams = toint(extract(@"num_parameters: (\d+)", 1, Message))
| where NumParams <= 50
| extend UseLocalization = tobool(extract(@"localization: (\w+)", 1, Message))
| where UseLocalization == True
```

The problem lies in that logs larger than a threshold (was it 5000?) are divided into chunks. If this information is not contained in the same log, we can't query the runs as the log statements does not contain every property. Trying to stitch multiple logs together before querying could maybe be possible, but a lot of work!

src/ert/run_models/run_model.py
SAKavli: Settings dict = All information about the run
Summary dict = The information from settings which we wish to query which must be short and contained in the first log chunk

src/ert/run_models/run_model.py
lucaeg: Thanks for a good explanation!

src/ert/run_models/run_model.py
lucaeg: OK, thanks!

Issue #10972, opened 2025-05-27
Make observations serializable for runmodels


yngve-sk:
Observations should be part of a standalone experiment config (serialized runmodel), a way to serialize observations should be added to support this. Assuming it is small, we could simply do JSON, or stream it as parquet. Maybe it needs some discussion.

PR #11538, opened 2025-08-21
Make observations serializable

yngve-sk:
**Issue**
Resolves https://github.com/equinor/ert/issues/10972

**Approach**
Encode observations to dicts.
Added schema and type to support multiple
types for future use, especially wrt big
observation files, it makes more sense to
encode them to something more compressed that
can be inlined.

### DIFF
--- a/src/ert/storage/local_experiment.py
+++ b/src/ert/storage/local_experiment.py
@@ -6,7 +6,7 @@ from collections.abc import Generator
 from datetime import datetime
 from functools import cached_property
 from pathlib import Path
-from typing import TYPE_CHECKING, Annotated, Any
+from typing import TYPE_CHECKING, Annotated, Any, Literal
 from uuid import UUID
 
 import polars as pl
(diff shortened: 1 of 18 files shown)


src/ert/storage/local_experiment.py
```diff
@@ -60,6 +62,40 @@ class _Index(BaseModel):
 )
 
 
+# https://github.com/pola-rs/polars/issues/13152#issuecomment-1864600078
+# PS: Serializing/deserializing schema is scheduled to be added to polars core,
+# ref https://github.com/pola-rs/polars/issues/20426
+# then this workaround can be omitted.
+def str_to_dtype(dtype_str: str) -> pl.DataType:
+    # the other imports here
```
berland: what other imports?

yngve-sk: Oops! Was a dev artifact, removed it.

berland:
What you wrote in the "Approach" in the PR could also be added to the git commit body message so it will persist.

berland (approved):
👍🏻

PR #11593, opened 2025-08-26
Reimplement auto-scaling of objectives/constraints

### LINKED ISSUE #11571: Reimplement auto-scaling for objectives and output constraints
1. Autoscaling of multiple objectives should be replaced by auto-scaling of the total objective, since that is what is important for the optimization algorithms.
2. Autoscaling of constraints should be done on a per-constraint basis.
3. It should not be possible to select auto-scaling individually for objectives and constraints, if auto-scaling is attempted it should be done for all objectives and constraints. The autoscale flags should be removed and replaced by a single autoscaling flag in the optimization section.
4. Auto-scaling should be on by default.
5. Manual scaling should still be possible on a individual basis for both objectives or constraints, but should be disabled if auto-scaling is on.

verveerpj:
**Issue**
Resolves #11571

**Approach**
First commit:

The `config_advanced.yml` example was found to be using too many perturbations and batches. These were reduced to safe run-time, without changing the test results.

Second commit:
- Individual `auto_scale` options are removed from the `objective_functions` and `output_constraints` sections.
- An `auto_scale` option is added to the `optimization` sections. It is `False` by default.
- The `objective_functions` and `output_constraints` sections still have `scale` options, however these are mutually exclusive with the `auto_scale` option in the `optimization` section.
- The transform code that handles auto-scaling has been modified to reflect that auto-scaling is now on for all objectives and constraints, or completely off.
- Auto-scaling of the objectives is now done by scaling each by the  weighted objective function.
- The `ropt` library now reports weighted objective function after scaling it back to the user domain. For this reason the results and snapshots for the `config_multiobj.yml` case have changed.
- The `ropt` library does not require the `weighted_objective_from_optimizer` method anymore in the transform code, it has been removed.
- The changes in `ropt` are not backwards compatible and therefore it has been bumped to version 0.24.0. The `pyproject.toml` and `uv.lock` files have been updated accordingly.

**Note**
This will require an update of `ropt` to 0.24.0 in Komodo

      --exec 'just rapid-tests'`)

## When applicable


src/everest/config/output_constraint_config.py
```diff
@@ -68,3 +60,16 @@ def validate_target_or_bounds(cls, values: dict[str, Any]) -> dict[str, Any]:
...
+    @model_validator(mode="before")
+    @classmethod
+    def deprecate_normalization(cls, values: dict[str, Any]) -> dict[str, Any]:
+        errors = []
+        if "auto_scale" in values:
+            errors.append(
+                "auto_scale is deprecated and has been replaced with "
+                "auto_scale in the optimization section"
+            )
+        if errors:
```
yngve-sk: errors has only one item, could just direct raise?

src/everest/config/optimization_config.py
```diff
@@ -195,6 +195,13 @@ class OptimizationConfig(BaseModel, extra="forbid"):
 parallel, if supported by the optimization algorithm.
 
 The default is to use parallel evaluation if supported.
+""",
+    )
+    auto_scale: bool = Field(
```
yngve-sk, line 200: I guess here you will report back with some user feedback on whether they are fine with consolidating the scaling of responses/constraints (I don't know if they ever wanted only one of them on)

src/everest/config/everest_config.py
```diff
@@ -284,6 +284,24 @@ class EverestConfig(BaseModelWithContextSupport):
     config_path: Path = Field()
     model_config = ConfigDict(extra="forbid", frozen=True)
 
+    @model_validator(mode="after")
+    def validate_scales(self) -> Self:
+        if not self.optimization or self.optimization.auto_scale:
+            if any(item.scale is not None for item in self.objective_functions):
+                raise ValueError(
```
yngve-sk: This is not needed if you removed the auto_scale option from objectives & constraints anyway?

src/everest/config/optimization_config.py
verveerpj: Yes

src/ert/run_models/everest_run_model.py
```diff
@@ -438,6 +439,10 @@ def _create_optimizer(self) -> tuple[BasicOptimizer, list[float]]:
                 variables=self._transforms["control_scaler"],
                 objectives=self._transforms["objective_scaler"],
                 nonlinear_constraints=self._transforms["constraint_scaler"],
+                objective_weights=np.fromiter(
```
yngve-sk, line 442: Is this just a cosmetic change? ROPT used to get this info through the objectives as objs before?

src/everest/config/everest_config.py
verveerpj: This is to ensure that the `auto_scale` in the optimization section is mutually exclusive with the `scale` options in the objective/constraint sections.

src/everest/optimizer/opt_model_transforms.py
```diff
@@ -379,25 +383,40 @@ def calculate_auto_scales(
...
-        weights = np.tile(self._weights[realizations, np.newaxis], constraints.shape[1])
-        weights[np.isnan(constraints)] = 0.0
-        weights /= np.sum(weights, axis=0)
-        auto_scales = np.abs(np.sum(constraints * weights, axis=0))
-        auto_scales = np.where(auto_scales > 0.0, auto_scales, 1.0)
-        self._scales[self._auto_scales] *= auto_scales[self._auto_scales]
-        self._needs_auto_scale_calculation = False
+        if self._auto_scale:
+            if np.all(np.isnan(constraints)):  # All forward models failed, don't handle
+                return
```
yngve-sk: If this happens, is the consequence user-facing in any way?

verveerpj: Yes, that is handled by stopping the iteration and returning an appropriate exit code. But this will be called before that is done, so here we just skip, and let it be handled later. Should I maybe comment?

src/everest/optimizer/opt_model_transforms.py
```diff
@@ -201,7 +201,11 @@ class ObjectiveScaler(ObjectiveTransform):
     """
 
     def __init__(
-        self, scales: list[float], auto_scales: list[bool], weights: list[float]
+        self,
+        auto_scale: bool,
+        scales: list[float],
```
yngve-sk, line 206: Are `scales` used / "active" even if `auto_scale` is `False` / unspecified?

verveerpj: No, they are not.

test-data/everest/math_func/config_advanced.yml
```diff
@@ -71,8 +70,9 @@ install_templates: []
 optimization:
   algorithm: optpp_q_newton
   convergence_tolerance: 0.005
-  perturbation_num: 7
+  perturbation_num: 5
```
yngve-sk, line 73: Why this change? Appears disjoint from reimplementing autoscaling

test-data/everest/math_func/config_multiobj.yml
```diff
@@ -15,18 +15,16 @@ controls:
...
     weight: 0.25
-    scale: 1.0
 
 
 optimization:
   algorithm: optpp_q_newton
   convergence_tolerance: 0.005
   perturbation_num: 5
-  max_batch_num: 3
+  max_batch_num: 4
```
yngve-sk: Similar comment as the one above

test-data/everest/math_func/config_advanced.yml
verveerpj: Yes, just to make it a bit faster. 5 is sufficient.

test-data/everest/math_func/config_multiobj.yml
verveerpj: Because autoscale is now on by default, results are slightly different, an extra batch was needed to converge this test.

src/everest/optimizer/opt_model_transforms.py
```diff
@@ -274,18 +261,36 @@ def calculate_auto_scales(
...
         """
-        weights = np.tile(self._weights[realizations, np.newaxis], objectives.shape[1])
-        weights[np.isnan(objectives)] = 0.0
-        weights /= np.sum(weights, axis=0)
-        auto_scales = np.abs(np.sum(objectives * weights, axis=0))
-        auto_scales = np.where(auto_scales > 0.0, auto_scales, 1.0)
-        self._scales[self._auto_scales] *= auto_scales[self._auto_scales]
-        self._needs_auto_scale_calculation = False
+        if self._auto_scale:
+            if np.all(np.isnan(objectives)):  # All forward models failed, don't handle
```
yngve-sk: Could you add a small sentence about why we don't handle this case? I find myself asking why we end up in this part of the code in the first place, if the entire forward model is all nan. Does this full forward model failure get handled elsewhere? Goes into ROPT and gets handled there?

src/everest/optimizer/opt_model_transforms.py
```diff
@@ -274,18 +261,36 @@ def calculate_auto_scales(
...
+                "Auto-scaling of the objective failed "
+                "to estimate a positive scale factor"
+            )
+            realization_weights = np.tile(
+                self._realization_weights[realizations, np.newaxis], objectives.shape[1]
+            )
+            realization_weights[np.isnan(objectives)] = 0.0
+            rw_sum = np.sum(realization_weights, axis=0)
+            if np.any(rw_sum < np.finfo(np.float64).eps):
+                raise RuntimeError(msg)
```
yngve-sk: What is the reason for doing the same check on the weights first, before computing scale and doing on scale? Looks like performance optimization (no pun intended) but not sure if it is needed enough to not just do it once on the `auto_scale`?

src/everest/optimizer/opt_model_transforms.py
```diff
@@ -274,18 +261,36 @@ def calculate_auto_scales(
...
-        weights[np.isnan(objectives)] = 0.0
-        weights /= np.sum(weights, axis=0)
-        auto_scales = np.abs(np.sum(objectives * weights, axis=0))
-        auto_scales = np.where(auto_scales > 0.0, auto_scales, 1.0)
-        self._scales[self._auto_scales] *= auto_scales[self._auto_scales]
-        self._needs_auto_scale_calculation = False
+        if self._auto_scale:
+            if np.all(np.isnan(objectives)):  # All forward models failed, don't handle
+                return
+            msg = (
```
yngve-sk: Too generic varname, pertains to the runtime error

src/everest/optimizer/opt_model_transforms.py
```diff
@@ -379,25 +383,40 @@ def calculate_auto_scales(
...
+            w_sum = np.sum(weights, axis=0)
+            msg = (
+                "Auto-scaling of the constraints failed "
+                "to estimate a positive scale factor"
+            )
+            if w_sum < np.finfo(np.float64).eps:
+                raise RuntimeError(msg)
+            weights /= w_sum
+            auto_scales = np.abs(np.sum(constraints * weights, axis=0))
+            if np.any(auto_scales < np.finfo(np.float64).eps):
```
yngve-sk: This logic looks the same as the one for objectives, could we pull it into an own fn to dedupe? Comments on the objectives part is same as on this

src/everest/optimizer/opt_model_transforms.py
verveerpj: The issue is that the `_forward_model_evaluator` does not really handle the case that all forward models failed, it just returns the objectives all set to `nan` and then `ropt` handles that by stopping the optimization. That is fine, but `_forward_model_evaluator` still calls the code that does this scaling calculation, and if that raises an error, `ropt` would never get the change to stop the iteration properly. I will try to catch this problem earlier in `_forward_model_evaluator`, so that this will not be needed anymore.

src/everest/optimizer/opt_model_transforms.py
verveerpj: Otherwise you get a division by zero in the next line, I'd rather have the error. Could be handled by modifying the calculation in the next line to be zero, but I find this more explicit.

src/everest/optimizer/opt_model_transforms.py
verveerpj: Renamed to error message

src/everest/optimizer/opt_model_transforms.py
verveerpj: Not enterily the same, but I wil try.

src/everest/optimizer/opt_model_transforms.py
```diff
@@ -443,3 +441,20 @@ def get_optimization_domain_transforms(
...
+    functions: NDArray[np.float64],
+    error_msg: str,
+) -> NDArray[np.float64]:
+    realization_weights = np.tile(
+        realization_weights[realizations, np.newaxis], functions.shape[1]
+    )
+    realization_weights[np.isnan(functions)] = 0.0
+    rw_sum = np.sum(realization_weights, axis=0)
+    if np.any(rw_sum < np.finfo(np.float64).eps):
+        raise RuntimeError(error_msg)
```
yngve-sk, line 458: is there a test that hits this? And why do we need to raise that error here when it is also potentially raised after? Will the condition here sometimes occur when the latter condition (at call site) does not?

tests/everest/snapshots/test_api_snapshots/test_api_snapshots/config_advanced.yml/snapshot.json
```diff
@@ -2,16 +2,14 @@
...
-    4,
-    5
+    4
   ],
   "batches": [
     0,
     1,
     2,
     3,
-    4,
```
yngve-sk, line 13: These edits to test num perturbations/num batches are causing big diffs here, do they need to be in the same commit as the change to autoscaling? Would prefer separate commit here

src/everest/optimizer/opt_model_transforms.py
verveerpj: This is for the unlikely case that the sum of the realizations weights is zero. This is an unlikely case, that can happen if `ropt` modifies the realization weights (which can happen). Is a safe-guard for an unlikely case, so I think a test goes a bit too far. If this test is not raised here, the next line will give you potentially a divide by zero, so caching it already is a bit better.

tests/everest/snapshots/test_api_snapshots/test_api_snapshots/config_advanced.yml/snapshot.json
verveerpj: They do need to be in the same commit, because otherwise some tests will fail. They could be done in a separate commit _before_ the autoscale changes, but the logic why doing that is then lost.

yngve-sk: Hmm. It seems a bit weird however that this change makes the config advanced example to not work past batch 4, or now need more perturbations than before... So some justification/explanation should be there. Could you please explain why they would fail with this change and not before?

src/everest/optimizer/opt_model_transforms.py
yngve-sk: I think if it is a case that is considered realistic enough to put into code, we should have a simple test for it, unless it involves setting up a lot of stuff. 

> the next line will give you potentially a divide by zero

This is a good reason for doing it actually, but was not obvious to me, could you write this into a comment in the code?

verveerpj: I made a new test file, specifically for the domain transforms. Moved some tests there, and added some tests, including for these errors.

tests/everest/snapshots/test_api_snapshots/test_api_snapshots/config_advanced.yml/snapshot.json
verveerpj: As it turns out, the change in the advanced example does indeed not affect the scaling tests. It does however save some time, so I have split it off in a separate commit, and added that before the other changes.

berland:
Merging of incompatible ropt-changes must be synchronized with merging of a PR to bleeding in komodo-releases, to avoid expected failures like https://github.com/equinor/komodo-releases/actions/runs/17400212168/job/49391430538

PR #11622, opened 2025-08-27
Add logging and minor reporter.stop bug fix

### LINKED ISSUE #11578: Log missing communication to the compute node
We have encountered several issues when there was no communication to the ert evaluator and the compute node. Especially when running everest, since everest uses experiment server, which already can run on a compute node. 

**What should be done:**
We should log when there was no communication between the evaluator and the job runner and the job finished successfully. This means that the scheduler reported realization SUCCESS, but there were no messages in the evaluator (grep `Failed to send` in the forward-model-log file).

**We should log:**
- compute node: present in the STATUS file
- port: present in the config
- in case of everest also the server node

example of the log:
```
raise ClientConnectionError(
_ert.forward_model_runner.client.ClientConnectionError: dispatch-41605287 Failed to send b'CONNECT' to ipc:///tmp/socket-8f0e3a41 after retrying!
```

jonathan-eq:
**Issue**
Resolves #11578 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -4,6 +4,7 @@ import asyncio
 import logging
 import threading
 import traceback
+from collections import defaultdict
 from collections.abc import Awaitable, Callable, Iterable, Sequence
 from typing import Any, cast, get_args
 
@@ -32,8 +33,10 @@ from _ert.forward_model_runner.client import (
 )
 from _ert.forward_model_runner.fm_dispatch import FORWARD_MODEL_TERMINATED_MSG
 from ert.ensemble_evaluator import identifiers as ids
+from ert.ensemble_evaluator import state
 from ert.scheduler import create_driver
 from ert.scheduler.scheduler import Scheduler
+from ert.shared.net_utils import get_machine_name
 
 from ..config import QueueSystem
 from ._ensemble import FMStepSnapshot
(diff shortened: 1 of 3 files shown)


andreas-el:
Backport to 14.6 perhaps?

xjules (approved):
Nice job! Could you squash the commits? :rocket:

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -515,6 +518,46 @@ def log_exception(task_exception: BaseException, task_name: str) -> None:
...
+        for real_id, real_data in (
+            self.ensemble.snapshot.data().get("reals", {}).items()
+        ):
+            if all(
+                fm_step_data["status"] == state.FORWARD_MODEL_STATE_INIT
+                for fm_step_data in real_data.get("fm_steps", {}).values()
+            ):
+                host_name = real_data.get("exec_hosts")
+                if host_name == "-":
+                    host_name = "localhost"
```
jonathan-eq, line 540: We hard code this in scheduler/lsf-driver so that it shows up in the gui. It should probably be hardcoded there instead, but that will be for a different PR.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -515,6 +518,46 @@ def log_exception(task_exception: BaseException, task_name: str) -> None:
...
+        hosts_with_missing_connection_to_realization_id: dict[str, list[str]] = (
+            defaultdict(list)
+        )
+        for real_id, real_data in (
+            self.ensemble.snapshot.data().get("reals", {}).items()
+        ):
+            if all(
+                fm_step_data["status"] == state.FORWARD_MODEL_STATE_INIT
+                for fm_step_data in real_data.get("fm_steps", {}).values()
+            ):
```
jonathan-eq, line 537: This should be improved in the future, so that it does not need all fm_steps to be pending, but rather that atleast one is stuck on pending. This does not work for the time being, as we do not change fm_step_status for the remaining fm_steps to `cancelled` after the previous one fails (they are stuck on `pending`).

PR #11633, opened 2025-08-28
Restructure some tests in test_runpath_creation.py

eivindjahren:
      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/config/test_ert_config.py
+++ b/tests/ert/unit_tests/config/test_ert_config.py
@@ -2455,3 +2455,24 @@ def test_validation_error_on_invalid_design_matrix_parameter_name(
                 DESIGN_MATRIX {tmp_path}/my_design_matrix.xlsx
                 """
         )
+
+
+def test_that_the_runpath_keyword_sets_the_runpath_substitution():
+    assert (
+        ErtConfig.from_dict({"RUNPATH": "my/runpath"}).substitutions["<RUNPATH>"]
+        == "my/runpath"
+    )
+
+
+@pytest.mark.parametrize("eclbase_substitution", ["<ECLBASE>", "<ECL_BASE>"])
+def test_that_the_eclbase_keyword_sets_the_eclbase_substitution(eclbase_substitution):
+    datafile = "input.data"
+    assert (
+        ErtConfig.from_dict(
+            {
+                "ECLBASE": datafile,
+                "JOBNAME": "job<IENS>",
+            }
+        ).substitutions[eclbase_substitution]
+        == datafile
+    )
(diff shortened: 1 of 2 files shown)


tests/ert/unit_tests/test_run_path_creation.py
```diff
@@ -46,42 +41,38 @@ def test_create_run_args(prior_ensemble, config_dict):
         f"{Path().absolute()}/simulations/realization-{i}/iter-0"
         for i in range(ensemble_size)
     ]
```
SAKavli, line 43: Should there be an assert here somewhere testing the substitution?

SAKavli: Or was the idea to merge this test with the next one perhaps?

tests/ert/unit_tests/config/test_ert_config.py
```diff
@@ -2455,3 +2455,24 @@ def test_validation_error_on_invalid_design_matrix_parameter_name(
                 DESIGN_MATRIX {tmp_path}/my_design_matrix.xlsx
                 """
         )
+
+
+def test_that_the_runpath_keyword_sets_the_runpath_substitution():
```
SAKavli, line 2460: I take it as these are new tests and not restructured? Reviewing could probably go faster if this is separated commits for context.

eivindjahren: The test is split out from `test_create_run_args` that is why it is in the same commit.

tests/ert/unit_tests/test_run_path_creation.py
eivindjahren: No the testing of the substitutions were put into separate tests

SAKavli: So we are just testing that `create_run_arguments` doesn't crash given certain input? Just had to make sure there was no mistake here. Sounds good.

tests/ert/unit_tests/config/test_ert_config.py
SAKavli: Now I see, thanks!

SAKavli (approved):
🧙‍♂️

tests/ert/unit_tests/test_run_path_creation.py
eivindjahren: This is the assert:
```
  assert [real.runpath for real in run_args] == [
        f"{Path().absolute()}/simulations/realization-{i}/iter-0"
        for i in range(ensemble_size)
    ]
```
The other asserts were moved elsewhere :)

PR #11641, opened 2025-08-28
Make unmatched quotes raise config validation error

### LINKED ISSUE #10585: Quoting error not catched by parser
If the following quoting error is done in `poly.ert`:

`FORWARD_MODEL poly_eval(<FOO>=""bar")
`

(assuming the user intented to write `"bar"` as the argument to `FOO`) then there will be no error or warning message from Ert, while the argument provided to `poly_eval` will be the empty string. 

Empty string is correct enough, but then the `bar"` should be considered noise in the input that at least should give a warning.

This issue is a follow-up to https://github.com/equinor/ert/issues/10568

eivindjahren:
**Issue**
Resolves #10585

**Approach**
Dissallows quotes inside of string

      --exec 'just rapid-tests'`)

## When applicable


tests/ert/unit_tests/config/parsing/test_lark_parser.py
```diff
@@ -345,3 +345,40 @@ def test_that_giving_too_few_arguments_to_enkf_alpha_raises_config_validation_er
...
+def test_that_quotes_can_be_escaped():
+    contents = parse_contents(
+        """
+        NUM_REALIZATIONS 1
+        FORWARD_MODEL poly_eval(<FOO>="\\"bar")
+        """,
+        file_name="config.ert",
+        schema=init_user_config_schema(),
+    )
+    assert contents["FORWARD_MODEL"] == [["poly_eval", [("<FOO>", '\\"bar')]]]
```
berland, line 371: Can this be called "escaping"? The backslash character(s) is carried on to the forward model step as is, and not treating the quote character.

eivindjahren: It is escape in the sense that it doesn't end the string in our config. I think keeping the backslash is the correct behavior in order to preserve character locations and all information.

berland: 👍🏻  Lets keep it that way until a user enters this narrow corner and proposes a valid use-case..

PR #11652, opened 2025-08-29
Have fm_steps in snapshot be set to cancelled if any previous fail

### LINKED ISSUE #11651: FM steps in GUI stuck on `Pending` after cancellation or failure

If a forward model is cancelled or fails, the forward model is marked as having failed, but the remaining forward model steps are still displayed as still `Pending`. This can be very misleading as the user might think something is still running or waiting to run, even though it absolutely never will.

Image

The remaining forward model steps are labeled as `Cancelled`

### Steps to reproduce

Add `raise RuntimeError("bogus")` to poly_eval.py

- [ ] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

jonathan-eq:
**Issue**
Resolves #11651

**Approach**
This commit makes it so that `Pending` forward model steps have their state set to `Cancelled` if their realization fails and they have not started yet. EnsembleSnapshot makes sure that it does not overwrite non-pending states when merging in the  RealizationFailed event. This fixes the issue where the fm_steps were stuck in Pending in the GUI if the realization fails or is cancelled.

Before:
image

After:
image

      --exec 'just rapid-tests'`)

## When applicable


src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -297,6 +299,26 @@ def update_from_event(
...
+                                "status", state.FORWARD_MODEL_STATE_INIT
+                            )
+                            == state.FORWARD_MODEL_STATE_INIT
+                        ) and (
+                            source_snapshot._fm_step_snapshots[fm_idx].get("status")
+                            == state.FORWARD_MODEL_STATE_INIT
+                        ):
+                            self._fm_step_snapshots[fm_idx]["status"] = (
+                                state.FORWARD_MODEL_STATE_CANCELLED
+                            )
```
jonathan-eq: This will only overwrite in cases where the forward model step is still in Pending state after the realization has failed.

tests/ert/unit_tests/test_tracking.py
```diff
@@ -124,6 +124,7 @@ def check_expression(original, path_expression, expected: list[str], msg_start):
             "",
             (
                 "    import os\n"
+                "    import sys\n"
```
jonathan-eq, line 127: This was an error in the test. We tried sys.exit(1) without importing sys, causing it to fail for reasons other than expected.

jonathan-eq:
It works on tgx as well 👌

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -297,6 +299,26 @@ def update_from_event(
...
+                        fm_idx = (event.real, fm_step_id)
+                        if fm_idx not in self._fm_step_snapshots:
+                            self._fm_step_snapshots[fm_idx] = FMStepSnapshot()
+                        if (
+                            self._fm_step_snapshots[fm_idx].get(
+                                "status", state.FORWARD_MODEL_STATE_INIT
+                            )
+                            == state.FORWARD_MODEL_STATE_INIT
+                        ) and (
+                            source_snapshot._fm_step_snapshots[fm_idx].get("status")
```
xjules: not sure that I get this `source_snapshot._fm_step_snapshots[fm_idx].get("status")` if statement.

jonathan-eq: We have a list of events that we merge into a SnapshotUpdateEvent based on the source_snapshot, then we merge that snapshotupdateevent into the source_snapshot. The statement is so that we don't overwrite FORWARD_MODEL_STATE_FAILED/SUCCEEDED with CANCELLED

jonathan-eq: Wait, you are right, we already did the check, I dont think we need it twice.

jonathan-eq: There was another edgecase where we could get FMStepFailed event in the same list as RealizationFailed due to the batching. This would cause us to overwrite Failed with Cancelled.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -297,6 +299,23 @@ def update_from_event(
                 end_time = convert_iso8601_to_datetime(timestamp)
             if type(event) is RealizationFailed:
                 message = event.message
+                # Mark remaining forward model steps that will never run as cancelled
+                for fm_step_id, fm_step in source_snapshot.get_fm_steps_for_real(
+                    event.real
+                ).items():
+                    if fm_step["status"] == state.FORWARD_MODEL_STATE_INIT:
```
xjules, line 306: Now it makes :)

tests/ert/unit_tests/test_tracking.py
xjules: good catch!

PR #11662, opened 2025-08-29
Fix data_for_response matches substrings

### LINKED ISSUE #11564: dark_storage data_for_response may match substrings of the summary key when not all summary keys are present
```python
@pytest.mark.parametrize("well_name", ("46/3-7s", "FOPR"))
def test_that_responses_do_not_match_substring(api_and_storage, well_name):
    api, storage = api_and_storage
    key = f"WBHP:{well_name}"
    date = datetime(year=2024, month=10, day=4)
        parameters=[],
        responses=[
            SummaryConfig(
                name="summary",
                input_files=["CASE.UNSMRY", "CASE.SMSPEC"],
                keys=[key, history_key(key), "FOPR"],
            )
        ],
        observations={
            "summary": pl.DataFrame(
                {
                    "observation_key": [key, history_key(key), "FOPR"],
                    "response_key": [key, history_key(key), "FOPR"],
                    "time": [date, date, date],
                    "observations": [0.2, 0.1, 0.01],
                    "std": [0.1, 0.1, 0.2],
                }
            )
        },
    )
    ensemble = experiment.create_ensemble(ensemble_size=1, name="ensemble")
    assert api.data_for_response(str(ensemble.id), key).empty
    ensemble.save_response(
        "summary",
        pl.DataFrame(
            {
                "response_key": ["FOPR"],
                "time": [pl.Series([date]).dt.cast_time_unit("ms")],
                "values": [pl.Series([1.0], dtype=pl.Float32)],
            }
        ).explode("values", "time"),
        0,
    )
    assert api.data_for_response(str(ensemble.id), "FOPR").to_csv() == dedent(
        """\
        Realization,2024-10-04
        0,1.0
        """
    )
    assert api.data_for_response(str(ensemble.id), key).empty # fails when the well name is FOPR
```

HakonSohoel:
**Issue**
Resolves #11564

**Approach**
Introduced the test from the issue and verified that is was failing. 
Solved the bug to make all tests pass. 
Verified that *H responses are still working in the plotter by running Snake Oil.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/dark_storage/endpoints/responses.py
+++ b/src/ert/dark_storage/endpoints/responses.py
@@ -86,14 +86,8 @@ response_to_pandas_x_axis_fns: dict[str, Callable[[tuple[Any, ...]], Any]] = {
 def _extract_response_type_and_key(
     key: str, response_key_to_response_type: dict[str, str]
 ) -> tuple[str, str] | tuple[None, None]:
-    # Check for exact match first. For example if key is "FOPRH"
-    # it may stop at "FOPR", which would be incorrect
-    response_key = next((k for k in response_key_to_response_type if k == key), None)
-    if response_key is None:
-        response_key = next(
-            (k for k in response_key_to_response_type if k in key and key != f"{k}H"),
-            None,
-        )
+    # Only allow exact matches.
+    response_key = key if key in response_key_to_response_type else None
 
     if response_key is None:
         return None, None
--- a/tests/ert/unit_tests/gui/tools/plot/test_plot_api.py
+++ b/tests/ert/unit_tests/gui/tools/plot/test_plot_api.py
@@ -22,6 +22,7 @@ from ert.dark_storage.app import app
 from ert.gui.tools.plot.plot_api import PlotApi, PlotApiKeyDefinition
 from ert.services import StorageService
 from ert.storage import open_storage
+from ert.summary_key_type import history_key
 from tests.ert.unit_tests.gui.tools.plot.conftest import MockResponse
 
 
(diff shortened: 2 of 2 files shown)


tests/ert/performance_tests/test_dark_storage_performance.py
```diff
@@ -416,3 +418,53 @@ def test_plotter_on_all_snake_oil_responses_memory(api_and_snake_oil_storage):
...
+        0,
+    )
+    assert api.data_for_response(str(ensemble.id), "FOPR").to_csv() == dedent(
+        """\
+        Realization,2024-10-04
+        0,1.0
+        """
+    )
+    assert api.data_for_response(
+        str(ensemble.id), key
```
SAKavli: Why do we expect this to fail for `FOPR`? I notice `FOPR` is being used as `response_key`.

tests/ert/performance_tests/test_dark_storage_performance.py
```diff
@@ -416,3 +418,53 @@ def test_plotter_on_all_snake_oil_responses_memory(api_and_snake_oil_storage):
     # thresholds are set to about 1.5x local memory used
     assert total_memory_mb < 5000
     assert peak_memory_mb < 1500
+
+
+@pytest.mark.parametrize("well_name", ("46/3-7s", "FOPR"))
+def test_that_responses_do_not_match_substring(api_and_storage, well_name):
```
SAKavli: Could we find a title which capture even more of the specific test case? Or perhaps use the term `substring` inside the test, is that referring to `well_name`?

src/ert/dark_storage/endpoints/responses.py
```diff
@@ -86,14 +86,8 @@ async def get_response(
...
-    response_key = next((k for k in response_key_to_response_type if k == key), None)
-    if response_key is None:
-        response_key = next(
-            (k for k in response_key_to_response_type if k in key and key != f"{k}H"),
-            None,
-        )
+    # Only allow exact matches.
+    response_key = key if key in response_key_to_response_type else None
 
     if response_key is None:
```
SAKavli, line 92: Could cut down 1 line of code by:
```py
if (response_key := key if key in response_key_to_response_type else None) is None:
```
But I don't know think this looks any better than how it is per now.

tests/ert/performance_tests/test_dark_storage_performance.py
HakonSohoel: Removed the misleading comment. Test would fail for FOPR before the bug fix.

tests/ert/performance_tests/test_dark_storage_performance.py
HakonSohoel: New name proposal: `test_that_response_keys_do_not_match_keys_that_are_substrings`

Specifically we test that "WBHP:FOPR" will not match "FOPR" but that is just one example.

src/ert/dark_storage/endpoints/responses.py
HakonSohoel: No strong opinion here, for me it's slightly more readable on separate lines.

eivindjahren: I suggest
```python
    if (response_key := response_key_to_response_type.get(key)) is not None:
            response_type = response_key_to_response_type.get(response_key)
            assert response_type is not None
            return response_key, response_type
    return None, None
```
However, in any case I would prefer:
```python
response_key = response_key_to_response_type.get(key)
```
over the ternary.

tests/ert/performance_tests/test_dark_storage_performance.py
SAKavli: Nice 🎉

src/ert/dark_storage/endpoints/responses.py
SAKavli:
```bash
>>> a = {1: 2}
>>> a.get(1)
2
```
We are interested in keys not values in this case, will your suggestion work?

HakonSohoel:
The comment and logic was introduced in #10941. It seems to me that the comment has probably been added together with the line:
`response_key = next((k for k in response_key_to_response_type if k == key), None)`
while the next part might have been like this:
```
    if response_key is None:
        response_key = next(
            (k for k in response_key_to_response_type if k in key),
            None,
        )
```
Later, as part of the same commit, the `and key != f"{k}H"` part has probably been added without updating the comment.

yngve-sk:
I remember tweaking the logic here as there was some bug at some point, and the comment is (sloppy) worded towards that. At some point it checked if the response key was a substring of the requested key, with things like FOPR (response key) being a substring of the requested key (FOPRH), when really it should pick the exact match first. I think it is best to leave this logic behind as the source behind it https://github.com/equinor/ert/issues/9059 etc is gone now. Ideally we will just now attach metadata to the response metadatas and use that, and have no use the `*H` notation for history throughout the code.

PR #11666, opened 2025-08-30
Resolve type errors regarding ConfigDict by using dict[Any, Any]

eivindjahren:
Towards: #11668

In order to fully do so, it was necessary to replace MaybeWithContext with Any, which conceptually is better.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/parsing/config_dict.py
+++ b/src/ert/config/parsing/config_dict.py
@@ -1,4 +1,3 @@
-from .context_values import ContextString
-from .types import MaybeWithContext
+from typing import Any
 
-ConfigDict = dict[ContextString | str, MaybeWithContext]
+ConfigDict = dict[Any, Any]
--- a/src/ert/config/analysis_config.py
+++ b/src/ert/config/analysis_config.py
@@ -6,7 +6,7 @@ from dataclasses import field
 from math import ceil
 from os.path import realpath
 from pathlib import Path
-from typing import Any, Final, no_type_check
+from typing import Any, Final
 
 from pydantic import Field, PositiveFloat, ValidationError
 from pydantic.dataclasses import dataclass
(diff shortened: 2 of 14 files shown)


src/ert/config/parsing/config_dict.py
```diff
@@ -1,4 +1,3 @@
-from .context_values import ContextString
-from .types import MaybeWithContext
+from typing import Any
 
-ConfigDict = dict[ContextString | str, MaybeWithContext]
+ConfigDict = dict[Any, Any]
```
SAKavli, line 3: How would you argue that `dict[Any, Any]` is conceptually better than `MaybeWithContext`? I don't disagree, just curious

eivindjahren: `MaybeWithContext` was just an indicator that this could be a contextvariable, but if doesn't that is ok. Using an `Any` for that is much easier to deal with.

SAKavli: That sounds reasonable, thanks!

SAKavli (approved):
I think the code becomes easier to understand with this change, very nice @eivindjahren ! I've been confused by the `MaybeWithContext` in the past myself.

As a rookie, I think it would go faster to review with more details regarding the motivation of what we are trying to achieve - then I don't need to deduce it from the changes. 
For example an issue or writing a more detailed commit body. We discussed it slightly on the last squad retro when you were absent. What do you think?

eivindjahren:
The test failure I had is totally new according to codecov. It seems likely to be related to the change 50521854bb697cace4dc4e0fcf5f260139808058 . I think we will have to keep an eye on this. I will rerun a couple of times more to see how flaky it is.

PR #11678, opened 2025-09-01
Implement auto-scaling for input constraints

### LINKED ISSUE #11342: Automatic scaling of objectives, controls and constraints in Everest
It was always recommended to scale objectives, controls and output constraints to similar scales. It was expected to be done by the Everest user. Then auto scaling of the controls by default was introduced but not by default for the objectives and output constraints. Maybe we should consider all 3 of them _on_ by default?

### LINKED ISSUE #11572: Implement autoscaling for input constraints
- Input constraints can be scaled by their right-hand side and/or their coefficients.
- This should be done if autoscaling is enabled 
- Manual scaling should also be possible, only if autoscaling is off

verveerpj:
**Issue**
Resolves #11572
Closes #11342

**Approach**
- Add a `scale` option to the input constraint configuration. These are mutually exclusive with the `auto_scale` option in the `optimization` section.
- Update the control variable transforms to scale the input constraints in one of the two following ways:
  - If `auto_scale` is on in the `optimization` section, scale each   equation by the maximum of its coefficients or right-hand-side.
  - Use the scales, if defined, and `auto_scale` is off.
- Tests are added for auto-scaling and manual scaling.

--

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/config/input_constraint_config.py
+++ b/src/everest/config/input_constraint_config.py
@@ -1,5 +1,5 @@
 import numpy as np
-from pydantic import BaseModel, Field, field_validator
+from pydantic import BaseModel, Field, PositiveFloat, field_validator
 
 
 class InputConstraintConfig(BaseModel, extra="forbid"):
@@ -34,7 +34,14 @@ class InputConstraintConfig(BaseModel, extra="forbid"):
     upper_bound: float = Field(
         default=np.inf,
         description="""Only control values that satisfy the following
-         equation will be allowed: sum of (<control> * weight) <= upper_bound`
+        equation will be allowed: sum of (<control> * weight) <= upper_bound`
+        """,
+    )
+    scale: PositiveFloat | None = Field(
+        default=None,
+        description="""Scaling of input constraints (scale).
+        scale is a normalization factor which can be used to scale the input constraint.
+        The bounds or target, and the weights will be scaled with this number.
         """,
     )
 
--- a/src/everest/optimizer/opt_model_transforms.py
+++ b/src/everest/optimizer/opt_model_transforms.py
@@ -13,6 +13,7 @@ from ropt.transforms.base import (
 
 from everest.config import (
     ControlConfig,
+    InputConstraintConfig,
     ModelConfig,
     ObjectiveFunctionConfig,
     OutputConstraintConfig,
@@ -27,10 +28,20 @@ class EverestOptModelTransforms(TypedDict):
 
 
 class ControlScaler(VariableTransform):
-    """Transformation object to define a linear scaling of controls.
+    """Transformation object to define scaling related to the scales.
 
-    This object defines a linear scaling from lower and upper bounds [lb, ub] in
-    the user domain to a target range in the optimizer domain.
+    For scaling of the controls itself, this object defines a linear scaling
+    from lower and upper bounds [lb, ub] in the user domain to a target range in
+    the optimizer domain.
+
+    Constraints on linear combinations on the controls are defined by the
+    `input_constraints` section of the configuration. If a linear transformation
+    of the controls is defined, the linear
+     constraints are also transformed accordingly. In
+    addition, each of the linear constraints can be scaled by an overall factor,
+    either automatically if the `auto_scale` option in the `optimization`
+    section is set, or manually if the `scale` option in the `input_constraints`
+    section is set.
     """
 
     def __init__(
@@ -39,6 +50,8 @@ class ControlScaler(VariableTransform):
         upper_bounds: Sequence[float],
         scaled_ranges: Sequence[tuple[float, float]],
         control_types: list[Literal["real", "integer"]],
+        auto_scale_input_constraints: bool,
+        input_constraint_scales: list[float] | None,
     ) -> None:
         """Transformation object to define a linear scaling.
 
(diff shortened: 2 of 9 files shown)


src/everest/optimizer/opt_model_transforms.py
```diff
@@ -109,35 +126,66 @@ def linear_constraints_to_optimizer(
...
-        Here, the linear transformation of the variables to the optimizer domain
-        is given by the scaling factors $\mathbf{s}_i$ and the offsets $\mathbf{o}_i$:
+            $$
+            \hat{\mathbf{x}}_i = \frac{\mathbf{x}_i - \mathbf{o}_i}{\mathbf{s}_i}
+            $$
 
-        $$
-        \hat{\mathbf{x}}_i = \frac{\mathbf{x}_i - \mathbf{o}_i}{\mathbf{s}_i}
-        $$
+            In the optimizer domeain, the coefficients and right-hand-side
```
berland: domeain -> domain

src/everest/optimizer/opt_model_transforms.py
```diff
@@ -109,35 +126,66 @@ def linear_constraints_to_optimizer(
...
+                    np.where(np.isfinite(lower_bounds), np.abs(lower_bounds), 0.0),
+                    np.where(np.isfinite(upper_bounds), np.abs(upper_bounds), 0.0),
+                    np.max(np.abs(coefficients), axis=1),
+                ],
+                axis=0,
+            )
+        else:
+            scales = np.asarray(self._input_constraint_scales, np.float64)
+        if scales is not None:
+            coefficients = coefficients / scales[:, np.newaxis]  # noqa: PLR6104
```
berland: augmented assigment does not work with numpy arrays?

verveerpj: It does work, but the arguments of this method may be immutable arrays (`ropt` uses immutable numpy arrays and calls this method). `ruff` does not know that, hence the `noqa`, I will put a little comment to clarify.

verveerpj: Thinking about that, it may have not been a good decision that `ropt` is sending immutable arrays, it might be better to make them mutable again before passing them to this method. I may fix that if there is a new release of `ropt` in the future (not making a specific release for that...)

src/everest/config/input_constraint_config.py
```diff
@@ -34,7 +34,14 @@ class InputConstraintConfig(BaseModel, extra="forbid"):
     upper_bound: float = Field(
         default=np.inf,
         description="""Only control values that satisfy the following
-         equation will be allowed: sum of (<control> * weight) <= upper_bound`
+        equation will be allowed: sum of (<control> * weight) <= upper_bound`
+        """,
+    )
+    scale: float | None = Field(
```
berland: This should be a positive int, as negative values will crash it, and a zero value will yield divide by zero.

src/everest/optimizer/opt_model_transforms.py
```diff
@@ -109,35 +126,68 @@ def linear_constraints_to_optimizer(
...
+            transforming variables to the optimizer domain, the coefficients
+            ($\mathbf{A}$) and right-hand-side values ($\mathbf{b}$) must be
+            converted to remain valid.
+
+            Here, the linear transformation of the variables to the optimizer
+            domain is given by the scaling factors $\mathbf{s}_i$ and the
+            offsets $\mathbf{o}_i$:
+
+            $$
+            \hat{\mathbf{x}}_i = \frac{\mathbf{x}_i - \mathbf{o}_i}{\mathbf{s}_i}
```
berland: I don't think this is rendered anywhere, as our sphinx-build setup does not seem to include the API (?) (fixing that is of course a separate issue)

berland: (but if we intend not to render it anywhere, LaTeX markup is perhaps not appropriate, though beautiful)

verveerpj: Yes, that is a good point, I have removed the formatting, and since the equations are quite simple that looks better.

src/everest/optimizer/opt_model_transforms.py
```diff
@@ -109,35 +126,61 @@ def linear_constraints_to_optimizer(
...
             coefficients = coefficients * self._scaling_factors  # noqa: PLR6104
+
+        # Correct for constraint scaling:
+        scales: NDArray[np.float64] | None
+        if self._auto_scale_input_constraints:
+            scales = np.max(
+                [
+                    np.where(np.isfinite(lower_bounds), np.abs(lower_bounds), 0.0),
+                    np.where(np.isfinite(upper_bounds), np.abs(upper_bounds), 0.0),
+                    np.max(np.abs(coefficients), axis=1),
```
berland, line 170: why do we have axis=1 here, aren't all these 1-dim np.array objects?

verveerpj: `coefficients` is a matrix. Each row is the coefficients for one linear equation. By using `axis=1` we get for each equation the maximum of its coefficients.

src/everest/optimizer/opt_model_transforms.py
```diff
@@ -157,6 +200,7 @@ def bound_constraint_diffs_from_optimizer(
             The re-scaled bounds.
         """
         if self._scaling_factors is not None:
```
berland, line 199: Can these really be `None`? They might be able to be reduced to an empty list, but None I don't see that the code allows.

verveerpj: You are correct, that seems to be a left-over from the time scales could be optional. Will fix.

berland: (it probably applies to self._offsets too)

Issue #11673, opened 2025-09-01
Decentralize ErtConfig <-> plugin manager interactions


yngve-sk:
Workflows, forward model steps, queue options, and site config are set via the ERT plugin manager. Currently this goes through the ErtPluginContext(), applied at the top level for ERT config.

 This should be decoupled from the ErtConfig object, by using the same pattern as we do in everest. I.e., we pass down a context that may or may not contain a plugin manager, then each individual sub-config object can get the things relevant to it. 

Example: the queue config will read the queue options from the plugin manager, workflows can use the plugin manager to look up installed workflows, and so on.

Reasons for doing this:
* Opens up for getting the plugin context without going through ErtConfig
* Lets us still create models with / without a specific plugin context, which is more convenient wrt testing
* Runmodel serialization will be aware of context and not have to go through ErtConfig / create its own plugin manager every time it needs to read something.

Definition of done:
ERT handles its plugin context the same way as Everest.

PR #11698, opened 2025-09-02
Consolidate ert & everest interactions with plugins

yngve-sk:
**Issue**
Resolves https://github.com/equinor/ert/issues/11673

**Approach**
* Resolve circular dependency by moving workflows to ert.config instead of ert.plugins, as ert.config can no longer import ert.plugins, but ert.plugins needs to import config to parse/read the site config.
* Pass in a typed object with plugin context to pydantic validators
* ERT config.with_plugins now takes in the context from the ErtPluginContext(), removing any direct interactions with the ErtPluginManager. The custom error handling, initialization and heavy use of staticmethods in ErtConfig makes it unsuitable for (without a big refactor) taking in the context via pydantic in sub-validators.
* RunModels receive the necessary context in through pydantic validation context

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/ert_config.py
+++ b/src/ert/config/ert_config.py
@@ -491,8 +491,10 @@ def workflows_from_dict(
     dict[str, Workflow],
     defaultdict[HookRuntime, list[Workflow]],
 ]:
-    workflow_jobs = dict(copy.copy(installed_workflows)) if installed_workflows else {}
-    workflow_jobs = workflow_jobs_from_dict(content_dict, workflow_jobs)
+    workflow_jobs = workflow_jobs_from_dict(
+        content_dict,
+        dict(copy.copy(installed_workflows)) if installed_workflows else {},
+    )
     workflows, hooked_workflows = create_and_hook_workflows(
         content_dict, workflow_jobs, substitutions
     )


src/ert/config/__init__.py
```diff
@@ -56,11 +71,15 @@
     "EnsembleConfig",
     "ErrorInfo",
     "ErtConfig",
+    "ErtPlugin",
+    "ErtScript",
+    "ErtScript",
```
oyvindeide: Duplicate

src/ert/config/__init__.py
```diff
@@ -80,6 +101,20 @@
...
+    "PostUpdateFixtures",
+    "PostUpdateFixtures",
+    "PreExperimentFixtures",
+    "PreExperimentFixtures",
+    "PreFirstUpdateFixtures",
+    "PreFirstUpdateFixtures",
+    "PreSimulationFixtures",
+    "PreSimulationFixtures",
+    "PreUpdateFixtures",
+    "PreUpdateFixtures",
```
oyvindeide, line 120: Duplicates

tests/ert/ui_tests/cli/run_cli.py
```diff
@@ -3,9 +3,9 @@
 
 from ert.__main__ import ert_parser
 from ert.cli.main import run_cli as cli_runner
-from ert.plugins import ErtPluginManager
+from ert.plugins.plugin_manager import ErtPluginContext, ErtRuntimePlugins
 
-PM = ErtPluginManager()
+RUNTIME_PLUGINS = ErtPluginContext().__enter__()  # noqa: PLC2801
```
oyvindeide: This does not seem like a good idea, possible to refactor the tests?

tests/ert/unit_tests/plugins/dummy_plugins.py
```diff
@@ -18,27 +18,27 @@ def forward_model_configuration():
 
 @plugin(name="dummy")
 def ecl100_config_path():
-    return "/dummy/path/ecl100_config.yml"
+    return "dummy/path/ecl100_config.yml"
```
oyvindeide, line 21: Why is this required?

tests/ert/unit_tests/plugins/test_plugin_context.py
```diff
@@ -15,9 +17,47 @@
 ]
 
 
+class ErtPluginContextLegacy(ErtPluginContext):
```
oyvindeide: Can we refactor to avoid this?

src/everest/config/everest_config.py
```diff
@@ -27,9 +27,9 @@
 from pydantic_core.core_schema import ValidationInfo
 from ruamel.yaml import YAML, YAMLError
 
-from ert.base_model_context import BaseModelWithContextSupport, init_context
```
oyvindeide, line 30: `init_context` is no longer used after this

src/everest/config/simulator_config.py
```diff
@@ -113,7 +113,7 @@ def default_local_queue(cls, v: Any, info: ValidationInfo) -> Any:
         if v is None:
             options = None
             if info.context:
-                options = info.context.get("queue_system")
+                options = info.context.queue_options
             return options or LocalQueueOptions(max_running=8)
```
oyvindeide, line 117: `options` will now always be something, so this line does not work anymore?

src/ert/cli/main.py
```diff
@@ -35,14 +35,20 @@ class ErtCliError(Exception):
...
-def run_cli(args: Namespace, plugin_manager: ErtPluginManager | None = None) -> None:
+def run_cli(args: Namespace, runtime_plugins: ErtRuntimePlugins | None = None) -> None:
     ert_dir = os.path.abspath(os.path.dirname(args.config))
     os.chdir(ert_dir)
     # Changing current working directory means we need to update
     # the config file to be the base name of the original config
     args.config = os.path.basename(args.config)
 
-    ert_config = ErtConfig.with_plugins().from_file(args.config)
+    if runtime_plugins is not None:
```
oyvindeide, line 45: Is this purely for testing?

tests/ert/unit_tests/plugins/dummy_plugins.py
yngve-sk: The tests will now actually create an ertconfig from the site config lines, so they require files to exist. Leading slash would point to root dir on the computer running the tests

src/everest/config/simulator_config.py
yngve-sk: Hm, I think it can still be None? queue options are set here:
```
queue_options=site_config.queue_config.queue_options
            if has_site_config
            else None,
```
and typing there does include optional

tests/ert/unit_tests/plugins/test_plugin_context.py
yngve-sk: Probably.. But the tests in this file do stuff to edit the plugin manager behavior and test against it, and it would be ideal to not have to expose the plugin manager itself through the API for only the sake of these tests. With removing site config it might be possible to just rewrite / remove these tests as the site config part won't be needed anymore.

tests/ert/ui_tests/cli/run_cli.py
yngve-sk: Made it slightly better, will see very soon if consolidating the `init_context` / `with ErtPluginContext(...)` / `ErtRuntimePlugins` concepts will open up for doing this in a cleaner way

src/ert/cli/main.py
yngve-sk: I'm not entirely sure, but I see now this is not the same behavior as before, the previous behavior was weird though: passing in a plugin manager, then doing one default via .with_plugins(), which in practice will always be the same, unless it is done via testing. We have the pattern of passing the plugin manager (now replaced with plugin context) from the top and down, which I think is OK, so I think the clause could be removed, and we could just do `ErtConfig.with_plugins(runtime_plugins).from_file(args.config)`

src/ert/config/ert_config.py
```diff
@@ -1054,6 +1014,10 @@ def from_dict(cls, config_dict: ConfigDict) -> Self:
         substituter = Substitutions(substitutions)
         for key, val in config_dict.get("SETENV", []):
             env_vars[key] = substituter.substitute(val)
+
+        # Insert env vars from plugins/site config
+        for key, val in cls.ENV_VARS.items():
+            env_vars[key] = val  # Q: Substitute these? From site
```
oyvindeide: No, I think warn if duplicate, then take user value.

src/ert/config/queue_config.py
```diff
@@ -314,19 +343,52 @@ class QueueConfig(BaseModel):
     max_runtime: int | None = None
 
     @classmethod
-    def from_dict(cls, config_dict: ConfigDict) -> QueueConfig:
-        selected_queue_system = QueueSystem(
-            config_dict.get("QUEUE_SYSTEM", QueueSystem.LOCAL)
+    def from_dict(
+        cls,
+        config_dict: ConfigDict,
+        site_queue_options: QueueOptions | None = None,
```
oyvindeide, line 374: This is easier than changing to: `config_content_to_dict`?

src/ert/plugins/plugin_manager.py
```diff
@@ -182,6 +186,24 @@ def get_forward_model_configuration(self) -> dict[str, dict[str, Any]]:
...
+
+            if len(plugin_response) > 1:
+                plugin_names = [
+                    plugin.plugin_metadata.plugin_name for plugin in plugin_response
+                ]
+                raise ValueError(
+                    f"Only one site configuration is allowed, got {plugin_names}"
+                )
+            return ErtRuntimePlugins.model_validate(plugin_response[0].data)
+        except AttributeError:
```
oyvindeide, line 203: What is this from?

tests/everest/conftest.py
```diff
@@ -287,27 +285,12 @@ def no_plugins():
...
         patch("everest.simulator.everest_to_ert.ErtPluginContext", patched_context),
-        patch("ert.config.ert_config.ErtPluginManager", patched),
-        patch("everest.config.everest_config.ErtPluginManager", patched_everest),
+        patch("everest.config.everest_config.ErtPluginContext", patched_everest),
     ):
         yield
-
-
-@pytest.fixture(autouse=True)
-def no_site_config():
```
oyvindeide, line 304: How will testing in an environment with a site config work now?

yngve-sk: Can change the fixture a bit, but basically we just need to mock-override the `RuntimePlugins` class to return `None` of a default empty set of plugins, for the actual site config environment to never reach the test. Did this in some Everest tests already, ref 
```
with mock.patch(
        "ert.plugins.plugin_manager.ErtRuntimePlugins",
        return_value=ErtRuntimePlugins(
            environment_variables={"HOW_MANY_CPU": "<NUM_CPU>"}
        ),
    ):
```
could just return an empty instead

src/ert/config/queue_config.py
yngve-sk: I'm assuming you mean `ErtConfig.from_dict` where this is being called? I think it is easier to do it here, there are a lot of edge cases packed in here and the logic currently (still in this PR) reads complex. So I think we do need one place where we have the
1. list of lines in ertconfig that sets a queue option (some prefixed by `QUEUE_OPTION <SYSTEM>`, some are just not prefixed)
2. the options set in the site config

the order in which we want things is something like user setting > site config setting > default setting, and `QUEUE_OPTION <O>` settings should override global settings `<O>`, and there are some loggings given when we override things, I also added a logging for when a user config overrides site config as seen in the revised `test_that_overwriting_QUEUE_OPTIONS_warns`.

oyvindeide: I was thinking that this function would be:
```python
def config_content_to_dict(config: ConfigDict) -> unpack[QueueConfig]:
    return extract_queue_options_from_config_dict
```

src/ert/config/ert_config.py
```diff
@@ -668,10 +663,13 @@ class ErtConfig(BaseModel):
     DEFAULT_ENSPATH: ClassVar[str] = "storage"
     DEFAULT_RUNPATH_FILE: ClassVar[str] = ".ert_runpath_list"
     PREINSTALLED_FORWARD_MODEL_STEPS: ClassVar[Mapping[str, ForwardModelStep]] = {}
-    PREINSTALLED_WORKFLOWS: ClassVar[dict[str, ErtScriptWorkflow]] = {}
+    PREINSTALLED_WORKFLOWS: ClassVar[dict[str, WorkflowJob]] = {}
     ENV_PR_FM_STEP: ClassVar[dict[str, dict[str, Any]]] = {}
     ACTIVATE_SCRIPT: ClassVar[str | None] = None
```
oyvindeide: Will this be configured twice?

src/ert/plugins/plugin_manager.py
yngve-sk: I'm not sure, was copypaste from other plugin implementation, `_site_config_lines`, thinking it might be OK to remove all these...

src/ert/config/ert_config.py
yngve-sk: Yes, it will be read from plugins in ertconfig, but also in runmodels when deserializing (in the case we deserialize without going through ertconfig)

oyvindeide: But `activate_script` is a part of `QueueOptions`, so could remove that hook completely?

oyvindeide: https://github.com/equinor/ert/blob/be7bb50b8d49e7b0e762401a508da66adffc0eb8/src/ert/config/queue_config.py#L76

yngve-sk: Edit: It seems only we use the hook, so will remove

src/ert/config/workflow_job.py
```diff
@@ -143,10 +152,36 @@ class ErtScriptWorkflow(_WorkflowJob):
...
 
-    ert_script: type[ErtScript] = None  # type: ignore
+    ert_script: builtins.type[ErtScript] = None  # type: ignore
     description: str = ""
     examples: str | None = None
     category: str = "other"
+    type: Literal["ert_script"] = "ert_script"
+
+    @field_serializer("ert_script")
+    def serialize_ert_script(self, _: str | builtins.type[ErtScript]) -> str:
```
oyvindeide, line 159: Is the `.builtin` needed because it is pydantic now?

yngve-sk: ruff sets it to `builtins.type`

src/ert/config/workflow_job.py
```diff
@@ -133,6 +141,7 @@ def location(self) -> str | None: ...
 
 class ExecutableWorkflow(_WorkflowJob):
     executable: str | None = None
+    type: Literal["executable"] = "executable"
 
     def location(self) -> str | None:
```
oyvindeide, line 144: This is now obsolete and `type` could be used instead?

src/ert/config/ert_config.py
```diff
@@ -780,40 +776,19 @@ def __eq__(self, other: object) -> bool:
...
+    def with_plugins(runtime_plugins: ErtRuntimePlugins) -> type[ErtConfig]:
         class ErtConfigWithPlugins(ErtConfig):
             PREINSTALLED_FORWARD_MODEL_STEPS: ClassVar[
                 Mapping[str, ForwardModelStep]
-            ] = preinstalled_fm_steps
-            PREINSTALLED_WORKFLOWS = (
-                pm.get_ertscript_workflows().get_workflows()
-                | pm.get_legacy_ertscript_workflows().get_workflows()
+            ] = runtime_plugins.installed_forward_model_steps
+            PREINSTALLED_WORKFLOWS = dict(runtime_plugins.installed_workflow_jobs)
```
eivindjahren, line 805: There seems to be some changes in this commit that doesn't belong to it. specifically `with_plugins` has a lot of changes not related to resolving circular dependencies. Also, some classmethods are removed that is not related to resolving circular dependencies.

docs/ert/getting_started/howto/plugin_system.rst
```diff
@@ -233,7 +233,7 @@ Full example:
            print("Hello World")
 
    @ert.plugin(name="my_plugin")
-    def ertscript_workflow(config: ert.WorkflowConfigs):
+    def ertscript_workflow(config: ert.config.WorkflowConfigs):
```
eivindjahren: This we really do not want. A plugin should never have to import anything from `ert.config`. This change to the docs is not necessary as WorkflowConfigs is already exported through the top-level `__init__.py` file.

eivindjahren (changes requested):
There is a change to the plugin system docs that we really need to avoid. Other than that, more granular commits would make this a lot easier to review. Mostly the first commit seems to contain changes that are at odds with the commit message title and body.

docs/ert/getting_started/howto/plugin_system.rst
yngve-sk: fixed

src/ert/config/ert_config.py
yngve-sk: Split out move of workflows to config module into own commit (requires the dynamic import of `ErtPluginManager` from `ert_config.py`)

src/ert/config/ert_config.py
```diff
@@ -808,10 +810,9 @@ class ErtConfigWithPlugins(ErtConfig):
             PREINSTALLED_FORWARD_MODEL_STEPS: ClassVar[
                 Mapping[str, ForwardModelStep]
             ] = preinstalled_fm_steps
-            PREINSTALLED_WORKFLOWS = (
+            PREINSTALLED_WORKFLOWS = dict(
```
eivindjahren: There are still some changes here that are unrelated to moving workflows into the config module

src/ert/config/workflow_config.py
```diff
@@ -139,7 +138,7 @@ def add_workflow(
         self._workflows.append(workflow)
         self.parsers[workflow.name] = parser
 
-    def get_workflows(self) -> dict[str, ErtScriptWorkflow]:
+    def get_workflows(self) -> Mapping[str, ErtScriptWorkflow]:
```
eivindjahren: This change is also unrelated to moving workflow_config into the config module

src/ert/plugins/hook_implementations/workflows/__init__.py
```diff
@@ -12,7 +12,7 @@
 from .misfit_preprocessor import MisfitPreprocessor
 
 if TYPE_CHECKING:
-    from ert.plugins.workflow_config import WorkflowConfigs
+    from ert.config.workflow_config import WorkflowConfigs
```
eivindjahren: This import should be `from ert.config import WorkflowConfigs` to not touch internal structure of the `ert.config` module.

src/ert/plugins/plugin_manager.py
```diff
@@ -388,11 +388,19 @@ def get_documentation_for_workflows(self) -> dict[str, JobDoc]:
         return job_docs
 
     def get_ertscript_workflows(self) -> WorkflowConfigs:
+        # from ert.config import (
```
eivindjahren: We should avoid code scars.

eivindjahren (dismissed):
Found some small issues that should be resolved. I will look closer at the actual changes to see if there are bigger suggestions.

src/ert/config/ert_config.py
```diff
@@ -780,41 +776,19 @@ def __eq__(self, other: object) -> bool:
...
+    def with_plugins(runtime_plugins: ErtRuntimePlugins) -> type[ErtConfig]:
         class ErtConfigWithPlugins(ErtConfig):
             PREINSTALLED_FORWARD_MODEL_STEPS: ClassVar[
                 Mapping[str, ForwardModelStep]
-            ] = preinstalled_fm_steps
-            PREINSTALLED_WORKFLOWS = dict(
-                pm.get_ertscript_workflows().get_workflows()
-            ) | dict(pm.get_legacy_ertscript_workflows().get_workflows())
-            ENV_PR_FM_STEP: ClassVar[dict[str, dict[str, Any]]] = env_pr_fm_step_
-            ACTIVATE_SCRIPT = pm.activate_script()
```
eivindjahren, line 816: The removal of the `activate_script` hook seems a rather independent part of what is being done here. It would simplify understanding what has changed by having a separate commit for that change alone.

src/ert/config/workflow_job.py
```diff
@@ -143,11 +150,36 @@ class ErtScriptWorkflow(_WorkflowJob):
     Single workflow configuration object
     """
 
-    ert_script: type[ErtScript] = None  # type: ignore
+    ert_script: builtins.type[ErtScript] = None  # type: ignore
```
eivindjahren, line 153: Making `ert_script` serializable seems to not be related to the change here, and could probably easily be in its own commit.

src/ert/plugins/hook_specifications/site_config.py
```diff
@@ -1,8 +1,9 @@
+from ert.plugins import ErtRuntimePlugins
 from ert.plugins.plugin_manager import hook_specification
 
 
 @hook_specification
-def site_config_lines() -> list[str]:  # type: ignore
+def site_configurations() -> ErtRuntimePlugins:  # type: ignore
     """
-    :return: List of lines to append to site config file
+    :return: dict of ert runtime plugins
```
eivindjahren: "dict of" seems to be at odds with the type annotation here. Also not very descriptive so the docstring should probably be changed all together.

tests/ert/ui_tests/cli/test_cli.py
```diff
@@ -684,6 +684,7 @@ def test_that_prior_is_not_overwritten_in_ensemble_experiment(
             "COEFFS", np.arange(num_realizations)
         )
         np.testing.assert_array_equal(parameter_values, prior_values)
+
```
eivindjahren: This extra line should probably not be here.

tests/ert/unit_tests/cli/test_model_hook_order.py
```diff
@@ -18,8 +18,8 @@
     PreSimulationFixtures,
     PreUpdateFixtures,
     QueueConfig,
-    fixtures_per_hook,
 )
+from ert.config.workflow_fixtures import fixtures_per_hook
```
eivindjahren: This change to the import statements should not be here.

tests/ert/unit_tests/config/test_ert_config.py
```diff
@@ -324,7 +315,7 @@ def run(self, *args):
     Path("plugin.py").write_text(
         dedent(
             """
-            from ert.plugins import ErtPlugin
+            from ert.config import ErtPlugin
```
eivindjahren, line 318: This change should be part of the previous commit. Please make sure the tests pass on all commits.

src/ert/config/ert_config.py
```diff
@@ -1396,7 +1307,7 @@ def uppercase_subkeys_and_stringify_subvalues(
     return fixed_dict
 
 
-def _forward_model_step_from_config_contents(
+def forward_model_step_from_config_contents(
```
eivindjahren, line 1354: Please put the rename of this function into its own commit.

tests/ert/unit_tests/config/test_forward_model.py
```diff
@@ -50,7 +50,7 @@ def test_load_forward_model():
 
     assert fm_step.min_arg is None
 
-    fm_step = _forward_model_step_from_config_contents(contents, "CONFIG", name="Step")
+    fm_step = forward_model_step_from_config_contents(contents, "CONFIG", name="Step")
```
eivindjahren, line 54: Commit message should probably be `Rename function forward_model_step_from_config_contents` or `Remove leading underscore from forward_model_step_from_config_contents name`.

src/ert/plugins/hook_specifications/site_config.py
yngve-sk: Changed to
```
    """
    Configure global ERT settings for a specific computing environment (a "site").
    """
```
return type is `ErtRuntimePlugins`, which I think describes the rtype close enough? Not sure if this will/should be a part of our plugins docs either, at least it is not intended for users to override site-wide configurations as we maintain one per cluster.

tests/ert/unit_tests/config/test_queue_config.py
```diff
@@ -176,36 +176,108 @@ def test_invalid_realization_memory(invalid_memory_spec: str, error_message: str
...
     assert (
-        f"Overwriting QUEUE_OPTION {queue_system} MAX_RUNNING: \n Old value:"
-        " 10 \n New value: 10"
-    ) not in caplog.text
+        f"Overwriting QUEUE_OPTION {queue_system.upper()} JOB_SCRIPT: \n "
+        f"Old value: usr_dispatch2.py \n New value: usr_dispatch3.py\n" in captured_log
+    )
+
+
+def test_that_user_given_queue_settings_overwrites_site_config_(
```
eivindjahren: Test name has an extra underscore after it.

tests/ert/unit_tests/plugins/test_plugin_manager.py
```diff
@@ -34,21 +29,15 @@ def test_with_plugins():
         "test": "test",
         "test2": "test",
     }
-    assert pm.get_flow_config_path() == "/dummy/path/flow_config.yml"
-    assert pm.get_ecl100_config_path() == "/dummy/path/ecl100_config.yml"
-    assert pm.get_ecl300_config_path() == "/dummy/path/ecl300_config.yml"
+    assert pm.get_flow_config_path() == "dummy/path/flow_config.yml"
+    assert pm.get_ecl100_config_path() == "dummy/path/ecl100_config.yml"
```
eivindjahren, line 33: The change from absolute path to relative path seems rather arbitrary. Is it related to the change being made?

yngve-sk: It now has to point to some mocked up files as it goes through `ErtPluginContext.__enter__`, which triggers some validations. ref `mock_dummy_plugins` in `test_plugin_context.py`, previous paths were pointing to the root directory, which was also dubious.

src/ert/plugins/hook_specifications/site_config.py
eivindjahren: This seems fine :)

eivindjahren:
A couple of extra things we should probably test for:
* What if multiple plugins provide the `site_configurations` hook.
* There is a test for warning for overriding queue options, but we should have the same for environment variables, workflow jobs and forward model steps. It is also important that we assert which takes precedence.

yngve-sk:
> A couple of extra things we should probably test for:
> 
> * What if multiple plugins provide the `site_configurations` hook.

Added test for this

> * There is a test for warning for overriding queue options, but we should have the same for environment variables, workflow jobs and forward model steps. It is also important that we assert which takes precedence.

Added for overrides of envvars. 

For workflows and forward model steps: They overriding is handled to some extent within the plugin manager, so all we should (imo) explicitly test is that user settings override "installed" (from plugins, not only limited to site config) ones.

**Workflows:** ref `test_that_adding_a_workflow_twice_warns`,  `test_that_ert_warns_on_duplicate_workflow_jobs`. Legacy workflows are not overriding, there new additions are skipped, this is asserted in `test_workflow_config_duplicate_log_message`. Suggest to not touch/revise this in this PR.

**Forward model steps:** Already tested to some extent, `test_that_installing_two_forward_model_steps_with_the_same_name_warn`, `test_that_installing_two_forward_model_steps_with_the_same_name_warn_with_dir`. Wrote in a clause to give a warning for overrides of site config, as well as a test for precedence. Should be rewritten more but this should be deferred to a future PR.

tests/ert/unit_tests/config/test_queue_config.py
yngve-sk: fixed

src/ert/plugins/plugin_manager.py
```diff
@@ -408,73 +369,106 @@ def add_span_processor_to_trace_provider(self) -> None:
...
-                )
+        site_configurations = self.plugin_manager.get_site_configurations()
+
+        ecl100_config_path = self.plugin_manager.get_ecl100_config_path()
+        ecl300_config_path = self.plugin_manager.get_ecl300_config_path()
+        flow_config_path = self.plugin_manager.get_flow_config_path()
+
+        config_env_vars = {}
+        if ecl100_config_path is not None:
+            config_env_vars["ECL100_SITE_CONFIG"] = ecl100_config_path
```
oyvindeide, line 408: Fine to keep this pattern, but this is quite old, so could create an issue to delete this and just read it through hooks.

berland:
> > No commit message mention anything about what is the breaking change. Can that be made explicit?
> 
> fixed, does this look OK?
> 
> ```
> Rewrite site config to not go through ErtConfig
> 
> Take in site-wide configurations of ERT through
> typed python object instead of ERT config lines.
> 
> This is a non major version bump,
> backward-compatible breaking change, and will
> not work with older specifications of
> site-configurations that are intended to append
> lines to the ERT config. Newer configs must
> instead provide their plugins through the
> site_configurations hook.
> ```
The text says "non major", but you maybe mean "major"? But how we version Ert does not need to go into the commit message, it is more like the versioning is the result of the commits. So the important bit is to say that this commit is breaking, and what it is breaking.

yngve-sk:
> > > No commit message mention anything about what is the breaking change. Can that be made explicit?
> > 
> > 
> > fixed, does this look OK?
> > ```
> > Rewrite site config to not go through ErtConfig
> > 
> > Take in site-wide configurations of ERT through
> > typed python object instead of ERT config lines.
> > 
> > This is a non major version bump,
> > backward-compatible breaking change, and will
> > not work with older specifications of
> > site-configurations that are intended to append
> > lines to the ERT config. Newer configs must
> > instead provide their plugins through the
> > site_configurations hook.
> > ```
> 
> The text says "non major", but you maybe mean "major"? But how we version Ert does not need to go into the commit message, it is more like the versioning is the result of the commits. So the important bit is to say that this commit is breaking, and what it is breaking.

oops, the "non" was accidental indeed. Reworded to be along the lines of  "this is a breaking change":

```
This is a breaking change that requires ERT
plugins to be provided through the
site_configurations hook.
```

src/ert/plugins/plugin_manager.py
yngve-sk: Made issue for removing https://github.com/equinor/ert/issues/11879

oyvindeide (approved):
LGTM!

PR #11708, opened 2025-09-02
Plot parameters with log distribution in linear scale instead of loga…

### LINKED ISSUE #11579: Plot parameters with log distribution in linear scale instead of logarithmic
Currently, the default plot for log-type distributions generates a histogram on a logarithmic scale (top).
However, according to the ERT documentation, the representation is should be in the original (linear) scale, obtained after applying the exponential transformation to the data (bottom).

Image

Image

jonathan-eq:
…rithmic

**Issue**
Resolves #11579

**Approach**

      --exec 'just rapid-tests'`)

## When applicable


src/ert/gui/tools/plot/plot_api.py
```diff
@@ -127,9 +127,7 @@ def parameters_api_key_defs(self) -> list[PlotApiKeyDefinition]:
                             observations=False,
                             dimensionality=metadata["dimensionality"],
                             metadata=metadata["userdata"],
-                            log_scale=(metadata["transformation"] or "None")
```
jonathan-eq, line 130: Maybe we want to keep log_scale around in case we want a toggle in the GUI

xjules: > Maybe we want to keep log_scale around in case we want a toggle in the GUI

Based on the discussion with @oyvindeide we have decided to remove it completely (ie. log_scale). If having such a GUI checkbox, then this can be solved on the fly in the plot / histogram view.

PR #11714, opened 2025-09-02
WIP: Bind all interfaces

flikka:
**Issue**
Resolves #11705 and partly #11407 

**Approach**
_Short description of the approach_

(Screenshot of new behavior in GUI if applicable)

- [ ] PR title captures the intent of the changes, and is fitting for release notes.
- [ ] Added appropriate release note label
- [ ] Commit history is consistent and clean, in line with the [contribution guidelines](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md).
- [ ] Make sure unit tests pass locally after every commit (`git rebase -i main
      --exec 'just rapid-tests'`)

## When applicable
- [ ] **When there are user facing changes**: Updated documentation
- [ ] **New behavior or changes to existing untested code**: Ensured that unit tests are added (See [Ground Rules](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md#ground-rules)).
- [ ] **Large PR**: Prepare changes in small commits for more convenient review
- [ ] **Bug fix**: Add regression test for the bug
- [ ] **Bug fix**: Add backport label to latest release (format: 'backport release-branch-name')

PR #11715, opened 2025-09-02
Use StorageService.session to set config (url, auth, etc) in everest

### LINKED ISSUE #11407: Everest not hanging on cluster

At our site, we have two clusters both running SLURM. One runs Everest fine, the other will hang waiting for the Everest server. The reason is that the this clusters does not allow external connections. The server seems to depend on those, hence clients cannot connect.

I expect this to be an issue that can happen on any cluster, depending on its configuration. It is likely not relevant that this cluster happens to use SLURM.

The Everest server should function, even if only internal connections between nodes are allowed. The servers should listen to the correct interface and communicate the proper information to clients to connect to it.

### Steps to reproduce

```Python
Reproduceable on a cluster where external connections are blocked. I will expand in a comment.
```

- [x] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [ ] RHEL8
- [x] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [x] Slurm queue
- [ ] OpenPBS queue

flikka:
**Issue**
Fixes #11407 
.. by using ERTs StorageService.session to ensure the correct, functioning url is selected. 

**Approach**
In order to use the correct url (full domain name vs. hostname), use the existing StorageService.session to set up a client. 
Next step is to actually use the session's get/post - separate issue here: https://github.com/equinor/ert/issues/11745

Should be merged together with #11714 to get the full enjoyment needed for the TNO non-fqdn setup to work.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/bin/everest_script.py
+++ b/src/everest/bin/everest_script.py
@@ -9,14 +9,15 @@ import signal
 import socket
 import threading
 from functools import partial
+from pathlib import Path
 
 from _ert.threading import ErtThread
 from ert.config import QueueSystem
+from ert.services import StorageService
 from everest.config import EverestConfig, ServerConfig
 from everest.detached import (
     ExperimentState,
     everserver_status,
-    server_is_running,
     start_experiment,
     start_server,
     wait_for_server,
@@ -162,7 +163,15 @@ async def run_everest(options: argparse.Namespace) -> None:
         )
 
     server_state = everserver_status(everserver_status_path)
-    if server_is_running(*ServerConfig.get_server_context(options.config.output_dir)):
+    try:
+        StorageService.session(
+            Path(ServerConfig.get_session_dir(options.config.output_dir)), timeout=1
+        )
+        server_running = True
+    except TimeoutError:
+        server_running = False
+
+    if server_running:
         config_file = options.config.config_file
         print(
             "An optimization is currently running.\n"
@@ -211,12 +220,17 @@ async def run_everest(options: argparse.Namespace) -> None:
 
         print("Waiting for server ...")
         logger.debug("Waiting for response from everserver")
-        wait_for_server(options.config.output_dir, timeout=600)
+        client = StorageService.session(
+            Path(ServerConfig.get_session_dir(options.config.output_dir))
+        )
+        wait_for_server(client, timeout=600)
         print("Everest server found!")
         logger.info("Got response from everserver. Starting experiment")
 
         start_experiment(
-            server_context=ServerConfig.get_server_context(options.config.output_dir),
+            server_context=ServerConfig.get_server_context_from_conn_info(
+                client.conn_info
+            ),
             config=options.config,
         )
 
@@ -229,7 +243,7 @@ async def run_everest(options: argparse.Namespace) -> None:
                 if options.disable_monitoring
                 else run_detached_monitor,
                 name="Everest CLI monitor thread",
-                args=[ServerConfig.get_server_context(options.config.output_dir)],
+                args=[ServerConfig.get_server_context_from_conn_info(client.conn_info)],
                 daemon=True,
             )
             monitor_thread.start()
@@ -237,14 +251,14 @@ async def run_everest(options: argparse.Namespace) -> None:
             monitor_thread.join()
         elif options.disable_monitoring:
             run_empty_detached_monitor(
-                server_context=ServerConfig.get_server_context(
-                    options.config.output_dir
+                server_context=ServerConfig.get_server_context_from_conn_info(
+                    client.conn_info
                 )
             )
         else:
             run_detached_monitor(
-                server_context=ServerConfig.get_server_context(
-                    options.config.output_dir
+                server_context=ServerConfig.get_server_context_from_conn_info(
+                    client.conn_info
                 )
             )
 
--- a/src/everest/bin/utils.py
+++ b/src/everest/bin/utils.py
@@ -24,6 +24,7 @@ from ert.ensemble_evaluator import (
     SnapshotUpdateEvent,
 )
 from ert.logging import LOGGING_CONFIG
+from ert.services import StorageService
 from everest.config import EverestConfig
 from everest.config.server_config import ServerConfig
 from everest.detached import (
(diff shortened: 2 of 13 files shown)


src/everest/bin/utils.py
```diff
@@ -94,7 +95,10 @@ def setup_logging(options: argparse.Namespace) -> Generator[None, None, None]:
 def handle_keyboard_interrupt(signum: int, _: Any, options: argparse.Namespace) -> None:
     print("\n" + "=" * 80)
     if options.config.server_queue_system == QueueSystem.LOCAL:
-        server_context = ServerConfig.get_server_context(options.config.output_dir)
+        client = StorageService.session(
+            Path(ServerConfig.get_session_dir(options.config.output_dir))
+        )
+        server_context = ServerConfig.get_server_context_from_client(client)
         if server_is_running(*server_context):
```
jonathan-eq: Can we change this to the same format you did elsewhere in the PR (Try...Except with the timeout)?

src/everest/config/server_config.py
```diff
@@ -56,18 +57,35 @@ def get_server_url(output_dir: str) -> str:
...
-        - url of the server
-        - path to the .cert file
-        - password for the certificate file
+    def get_server_context_from_client(
+        client: Client,
+    ) -> tuple[str, str, tuple[str, str]]:
+        """Get server connection context information from a storage session.
+
+        Returns a tuple containing the server URL, certificate file path,
+        and authentication credentials. NOTE: This function is to bridge the gap between
```
xjules: I like the comment. Could you maybe add a sentence what do you mean by the gap term?

tests/everest/entry_points/test_everest_entry.py
```diff
@@ -113,13 +130,24 @@ def test_everest_entry(
     return_value={"status": ExperimentState.completed, "message": None},
 )
 @patch("everest.bin.everest_script.start_experiment")
+@patch("everest.config.ServerConfig.get_server_context_from_client")
+@patch(
+    "ert.services.StorageService.session",
+    side_effect=[
+        TimeoutError("Boom!"),
+        TimeoutError("Boom2!"),
+        TimeoutError("Boom3!"),
```
xjules: Do we need all the `Booms`?

src/everest/config/server_config.py
```diff
@@ -56,18 +57,38 @@ def get_server_url(output_dir: str) -> str:
         return f"https://{server_info['host']}:{server_info['port']}/experiment_server"
 
     @staticmethod
-    def get_server_context(output_dir: str) -> tuple[str, str, tuple[str, str]]:
-        """Returns a tuple with
-        - url of the server
-        - path to the .cert file
-        - password for the certificate file
+    def get_server_context_from_client(
+        client: Client,
```
xjules: just nit-picking here, but why we need to sent an entire client here? We do need only conn_info or ?

flikka: Agree!

tests/everest/entry_points/test_everest_entry.py
flikka: You don't like Boom? No, we dont need them :-)

src/everest/bin/everest_script.py
```diff
@@ -162,7 +163,15 @@ async def run_everest(options: argparse.Namespace) -> None:
         )
 
     server_state = everserver_status(everserver_status_path)
-    if server_is_running(*ServerConfig.get_server_context(options.config.output_dir)):
+    try:
+        StorageService.session(
+            Path(ServerConfig.get_session_dir(options.config.output_dir)), timeout=1
```
xjules, line 168: So what does happen in the code (of StorageService.session) is among other things the function connect:
```python
_base_service.py
    @classmethod
    def connect(
        cls,
        *,
        project: os.PathLike[str],
        timeout: int | None = None,
    ) -> Self:
        if cls._instance is not None:
            cls._instance.wait_until_ready()
            assert isinstance(cls._instance, cls)
            return cls._instance

        path = Path(project)
        name = f"{cls.service_name}_server.json"
        # Note: If the caller actually pass None, we override that here...
        if timeout is None:
            timeout = 240
        t = -1
        while t < timeout:
            if (path / name).exists():
                with (path / name).open() as f:
                    return cls((), conn_info=json.load(f), project=str(path))

            sleep(1)
            t += 1

        raise TimeoutError("Server not started")
```

Just by a naive thought process having timeout=1 triggers that basically immediately - especially given the server run on LSF. Shall we maybe reduce the sleep there?

flikka: Not sure I understand, the timeout in this call is set to 1, which is intended. You suggest to "reduce the sleep here" - not following.

This is supposed to fail fast if it is not running. The replaced "server_is_running" did a "get" against the endpoint, with a one second timeout. Now we check here the presense of the server file, with similar timeout. So the intention is to be quick.

xjules: > Not sure I understand, the timeout in this call is set to 1, which is intended. You suggest to "reduce the sleep here" - not following.

Right. Maybe I understand the code above wrongly. I thought that we are to connect to an existing instance of the storage_service (`if cls._instance is not None == False`) and thus we we only check for path (on lsf node) to contain the `xxx_server.json`. Here I thought that having `timeout==1` can fail as this is our (tgx or local) machine that calls connect and thus it might take some time to get the ` xxx_server.json` file visible and since we do `sleep(1)` while checking for the file presence this might gets "wrongly" triggered.

flikka: Well, we do check for path, indeed. But the purpose is to assess if there is a server running or not. If there is no server running, waiting more, will not help. And LSF or not, not really relevant I believe. This is only relevant if we eg. are waiting for a startup. This is just to assess, when we start the client side, is the server running (eg. if we started it, and then restart the client for example)

xjules: > And LSF or not, not really relevant I believe. 

I agree with all you said, but this :)
Anyway, I tend to believe it's correct.

Issue #11695, opened 2025-09-02
Move config dict validation from parsing to ErtConfig.from_dict


eivindjahren:
In order to have alternate input formats (needed for https://github.com/equinor/scout/issues/1292), schema validation should ideally be done in `ErtConfig.from_dict` to avoid having to duplicate the validation for each input format.

As of https://github.com/equinor/ert/commit/36e01f24429db2b58f069310a462c0a95c218b84 the config file is first turned into a list instructions (technically `Tree[Instruction]`) via `ert.config.parser.lark_parser._parse_contents`. Then that is turned into `ConfigDict` AND validated by `ert.config.parser.lark_parser._tree_to_dict`. The idea would be to separate `_transform_tree` into the parts necessary to correctly construct the dictionary and the validation (I believe this is simply `SchemaItem.apply_constraints`).

The same will need to happen to the OBS_CONFIG file, that it is read and parser before `ErtConfig.from_dict` as a list of dictionaries (as returned by `ert.config.parsing.observation_parser._parse_content_list`), and that it is validated in `ErtConfig.from_dict` (calls `ert.config.parsing.observation_parser._validate_conf_content`).

eivindjahren:
We have completed the parts of this that is relevant for observations, so putting this on hold for now.

PR #11722, opened 2025-09-03
Make ContextString=FileContextToken

eivindjahren:
Towards #11695

Because the output of the observations parser consists of lists, dicts and FileContextToken, it greatly simplifies things to simply have ContextString=FileContextToken. In order to do that, we also need to replace keyword_token/keyword_context concept with just the token of an instruction (which already was the token of the keyword).

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/config/test_workflow.py
+++ b/tests/ert/unit_tests/config/test_workflow.py
@@ -126,6 +126,7 @@ def test_that_redefine_in_workflow_overwrites_in_subsequent_lines():
     )
 
     commands = [(name, args[0]) for (name, args) in wf.cmd_list]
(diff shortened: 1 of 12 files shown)


tests/ert/unit_tests/config/test_workflow.py
```diff
@@ -126,6 +126,7 @@ def test_that_redefine_in_workflow_overwrites_in_subsequent_lines():
     )
 
     commands = [(name, args[0]) for (name, args) in wf.cmd_list]
+    print(commands)
```
andreas-el, line 129: Probably leave out

PR #11725, opened 2025-09-03
Remove perturbation_magnitude from eightcells example

berland:
Documentation says that this should normally not be touched, and removing it does not make a significant impact to optimization results from eightcells

**Issue**
Resolves irrelevant config

**Approach**
✂️ 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/test-data/everest/eightcells/everest/model/config.yml
+++ b/test-data/everest/eightcells/everest/model/config.yml
@@ -9,7 +9,6 @@ controls:
   -
     name: well_rate
     type: generic_control
-    perturbation_magnitude: 150
     variables:
       -
         name: OP1


test-data/everest/eightcells/everest/model/config.yml
```diff
@@ -9,7 +9,6 @@ controls:
   -
     name: well_rate
     type: generic_control
-    perturbation_magnitude: 150
```
verveerpj, line 12: This will effectively change the perturbation magnitude to 49, changing the result. That is fine, hopefully it shows a bit of an improvement still.

But what is the intention? Making the config as simple as possible?

test-data/everest/eightcells/everest/model/config.yml
```diff
@@ -9,7 +9,6 @@ controls:
...
     variables:
```
verveerpj, line 12: I note that the variables section has identical min/max values, these could be lifted up to the controls section.

test-data/everest/eightcells/everest/model/config.yml
berland: The documentation says:

> NOTE: In most cases this should not be configured, and the default value should be used

which implies that I should have to argue to keep it in (if I need to argue to keep it in, we should change the docs..)

verveerpj: Was not aware of that note, but in that case: LGTM!

test-data/everest/eightcells/everest/model/config.yml
berland: will keep them explicit for now as it is not obvious to me yet that they should be equal

PR #11713, opened 2025-09-02
Create experiment sorting for plotting window

SAKavli:
The simulation arguments are dumped to 'metadata.json'. However, this feature is currently only used by MultipleDataAssimilation.

For other run_models, they simply pass down "None" for this value. Therefore it makes more sense to make it an optional parameter.

I've also introduced the 'time_generated' to the dataclass such that experiments/ensembles can be sorted by this value.

**Issue**
Resolves #11680 

**Approach**
_Short description of the approach_

(Screenshot of new behavior in GUI if applicable)

- [ ] PR title captures the intent of the changes, and is fitting for release notes.
- [ ] Added appropriate release note label
- [ ] Commit history is consistent and clean, in line with the [contribution guidelines](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md).
- [ ] Make sure unit tests pass locally after every commit (`git rebase -i main
      --exec 'just rapid-tests'`)

## When applicable
- [ ] **When there are user facing changes**: Updated documentation
- [ ] **New behavior or changes to existing untested code**: Ensured that unit tests are added (See [Ground Rules](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md#ground-rules)).
- [ ] **Large PR**: Prepare changes in small commits for more convenient review
- [ ] **Bug fix**: Add regression test for the bug
- [ ] **Bug fix**: Add backport label to latest release (format: 'backport release-branch-name')

PR #11732, opened 2025-09-04
Sort ensembles chronologically in plotter

SAKavli:
**Issue**
Resolves #11713 

**Approach**
This sorts ensembles chronologically.

Before change:
image

After change:
image

- [X] Make sure unit tests pass locally after every commit (`git rebase -i main
      --exec 'just rapid-tests'`)

## When applicable


src/ert/gui/tools/plot/plottery/plots/histogram.py
```diff
@@ -146,7 +136,9 @@ def plotHistogram(
             PlotTools.showGrid(axes[ensemble.name], plot_context)
 
     min_count = 0
-    max_count = max(subplot.get_ylim()[1] for subplot in axes.values())
+    max_count = (
+        max(subplot.get_ylim()[1] for subplot in axes.values()) if axes.values() else 0
```
SAKavli, line 140: Not entirely sure if 0 is a good default value here? It's used for:
```py
if custom_limits.count_maximum is not None:
    max_count = custom_limits.count_maximum
...
subplot.set_ylim(min_count, max_count)
```

eivindjahren: You could test `None`, but I imagine it would have the same result.

src/ert/gui/tools/plot/plottery/plots/histogram.py
```diff
@@ -33,14 +33,13 @@ def plot(
         observation_data: pd.DataFrame,
         std_dev_images: dict[str, npt.NDArray[np.float32]],
     ) -> None:
-        plotHistogram(figure, plot_context, ensemble_to_data_map, observation_data)
```
eivindjahren, line 36: Probably want the commit title `Remove unused parameter observations_data` here, as the verb `delete` is a bit unconventional for this.

tests/ert/unit_tests/gui/plottery/test_ensemble_plot.py
```diff
@@ -48,3 +49,41 @@ def test_ensemble_plot_handles_rate(plot_context: PlotContext):
...
             assert mock_plotLines.call_args[0][4] is None
+
+
+def test_plot_ensemble_selection_widget_sorts_input_ensembles_reverse_chronologically(
+    qtbot,
+):
+    first_ensemble = "ens1"
+    second_ensemble = "ens2"
+    third_ensemble = "ens3"
+
```
eivindjahren: I think we can make this a bit simpler by using a list
```python
ensembles = [
  EnsembleObject(name=f"ens_{i}", id="id", hidden=False, experiment_name="exp", started_at=t)
  for i, t in enumerate(["2000-00-00T00:00:00", "2001-11-11T00:00:00", "2002-12-22T00:00:00"])
]
list_widget = EnsembleSelectListWidget(ensembles)
assert [list_widget.item(i).text() for i in range(len(ensembles))] == ["ens3", "ens2", "ens1"]

File: src/ert/gui/tools/plot/plottery/plots/histogram.py (hunk 1 above)
Comment (SAKavli) [reply src/ert/gui/tools/plot/plottery/plots/histogram.py:140]: I've tried without success to reproduce an example of where the max_count is set to 0. Either the Histogram window is empty, or there is always something to plot. With a single test realization run I believe I've discovered a bug, but apart from that the window is empty and the max count is set to funny number `0.05500000000000001
`.
https://github.com/equinor/ert/issues/11775

Comment (SAKavli) [reply src/ert/gui/tools/plot/plottery/plots/histogram.py:140]: So haven't been able to try with `None`, but I can't imagine this being a problem either.

File: tests/ert/unit_tests/gui/plottery/test_ensemble_plot.py (hunk 3 above)
Comment (SAKavli) [reply tests/ert/unit_tests/gui/plottery/test_ensemble_plot.py]: That is better, yes! Made something along those lines

File: tests/ert/unit_tests/dark_storage/test_http_endpoints.py [hunk 4]
```diff
@@ -141,7 +141,17 @@ def test_get_summary_response(
     userdata = resp_ensemble["userdata"]
 
     assert resp_ensemble["size"] == 5
-    assert userdata == {"name": "default_0", "experiment_name": "ensemble-experiment"}
```
SAKavli, line 144: Perhaps this is a bit too lazy of a fix, but this seems to work. The timestamp on ensembles is generated at run time of the test. Could probably have mocked something somewhere, but I did not spend time on this. I will look into it if it is worth the time!

PR #11747, opened 2025-09-05
Split and extend observation parsing tests

eivindjahren:
In order to work on the observation file format we will need to separate tests of the file format
from the validation tests (independent of format). This is a step in that direction.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/config/test_observations.py
+++ b/tests/ert/unit_tests/config/test_observations.py
@@ -1,7 +1,10 @@
 from contextlib import ExitStack as does_not_raise
 from datetime import datetime
 
+import hypothesis.strategies as st
 import pytest
+from hypothesis import given
+from pytest import MonkeyPatch, TempPathFactory
 from resdata.summary import Summary
 
 from ert.config import (
@@ -10,6 +13,8 @@ from ert.config import (
     ErtConfig,
 )
 from ert.config.parsing import parse_observations
+
+from .summary_generator import summaries
 
 
 def make_observations(obs_config_contents, parse=True):
@@ -632,22 +637,22 @@ def test_that_out_of_bounds_segments_are_truncated(tmpdir, start, stop, message)
             )
 
 
-@pytest.mark.parametrize("with_ext", [True, False])
-@pytest.mark.parametrize(
-    "keys",
-    [
-        [("FOPR", "SM3/DAY", None), ("FOPRH", "SM3/DAY", None)],
-        [("WWIR", "SM3/DAY", "WNAME"), ("WWIRH", "SM3/DAY", "WNAME")],
-    ],
-)
 @pytest.mark.filterwarnings("ignore:Config contains a SUMMARY key")
-def test_that_history_observations_are_loaded(tmpdir, keys, with_ext):
-    with tmpdir.as_cwd():
-        key, _, wname = keys[0]
-        local_name = key if wname is None else (key + ":" + wname)
-        run_sim(datetime(2014, 9, 10), keys)
+@given(
+    std=st.floats(min_value=0.1, max_value=1.0e3),
+    with_ext=st.booleans(),
+    summary=summaries(summary_keys=st.just(["FOPR", "FOPRH"])),
+)
+def test_that_history_observations_values_are_fetched_from_refcase(
+    tmp_path_factory: TempPathFactory, summary, with_ext, std
+):
+    with MonkeyPatch.context() as patch:
+        patch.chdir(tmp_path_factory.mktemp("history_observation_values_are_fetched"))
+        smspec, unsmry = summary
+        smspec.to_file("ECLIPSE_CASE.SMSPEC")
+        unsmry.to_file("ECLIPSE_CASE.UNSMRY")
 
-        ert_config = ErtConfig.from_dict(
+        observations = ErtConfig.from_dict(
             {
                 "ECLBASE": "ECLIPSE_CASE",
                 "REFCASE": f"ECLIPSE_CASE{'.DATA' if with_ext else ''}",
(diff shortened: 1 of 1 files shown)


tests/ert/unit_tests/config/test_observations.py
```diff
@@ -1086,117 +996,290 @@ def test_that_common_observation_error_validation_is_handled(
...
+            SUMMARY_OBSERVATION FOPR
+            {
+                HOURS = -1;
+                KEY = FOPR;
+            };
+        """)
+
+
+def test_that_date_must_be_a_date_in_summary_observation():
+    with pytest.raises(ConfigValidationError, match="Please use ISO date format"):
```
SAKavli, line 1206: Does the please wording sound optional? Is it better with something angry? 😡
Wait a minute, I guess this message is defined elsewhere in Ert. But maybe still valid?

SAKavli (approved):
These changes makes the tests much more describing, næis 👏😎

tests/ert/unit_tests/config/test_observations.py
eivindjahren: Yea, I think we can reconsider this message. But its a bit convoluted now. Probably better to refactor first.

tests/ert/unit_tests/config/test_observations.py
```diff
@@ -152,10 +151,50 @@ def test_that_using_summary_observations_without_eclbase_shows_user_error():
...
+        ConfigValidationError, match=r"must be given a positive value|strictly > 0"
+    ):
+        make_observations(
+            [
+                (
+                    "SUMMARY_OBSERVATION",
+                    "FOPR",
+                    {
+                        "KEY": "FOPR",
+                        "VALUE": "0",  # ERROR becomes zero when mode is REL
```
SAKavli, line 190: I am a little confused by this comment, as `value` is zero, does that have anything to do with the value of `ERROR`? If `ERROR` becomes zero because of `REL` mode, should the comment be put on the `ERROR `value?

tests/ert/unit_tests/config/test_observations.py
```diff
@@ -197,17 +196,54 @@ def test_that_computed_error_must_be_greater_than_zero_in_summary_observations()
...
-            list(std),
-            list(range(len(std))),
-            list(range(len(std))),
+@pytest.mark.usefixtures("use_tmpdir")
+@pytest.mark.parametrize("std", [-1.0, 0, 0.0])
+def test_that_error_must_be_greater_than_zero_in_general_observations(std):
+    with pytest.raises(
+        ConfigValidationError, match=r"must be given a positive value|strictly > 0"
+    ):
+        make_observations(
```
SAKavli, line 206: All of the objects passed into `make_observations` are rather long, is it possible to shorten them down with some technique? Or are all of them unique? I suppose it wouldn't make any sense creating an object for this as that is what we will do in the milestone.

Might not be anything to do here, just mentioning it in case you get a bright idea!

tests/ert/unit_tests/config/test_observations.py
```diff
@@ -636,22 +641,22 @@ def test_that_out_of_bounds_segments_are_truncated(tmpdir, start, stop, message)
...
-)
 @pytest.mark.filterwarnings("ignore:Config contains a SUMMARY key")
-def test_that_history_observations_are_loaded(tmpdir, keys, with_ext):
-    with tmpdir.as_cwd():
-        key, _, wname = keys[0]
-        local_name = key if wname is None else (key + ":" + wname)
-        run_sim(datetime(2014, 9, 10), keys)
+@given(
+    std=st.floats(min_value=0.1, max_value=1.0e3),
+    with_ext=st.booleans(),
```
SAKavli, line 643: What does `std` and `with_ext` mean in this context? Is it more readable without abbreviations?

tests/ert/unit_tests/config/test_observations.py
```diff
@@ -636,22 +641,22 @@ def test_that_out_of_bounds_segments_are_truncated(tmpdir, start, stop, message)
...
+    summary=summaries(summary_keys=st.just(["FOPR", "FOPRH"])),
+)
+def test_that_history_observations_values_are_fetched_from_refcase(
+    tmp_path_factory: TempPathFactory, summary, with_ext, std
+):
+    with MonkeyPatch.context() as patch:
+        patch.chdir(tmp_path_factory.mktemp("history_observation_values_are_fetched"))
+        smspec, unsmry = summary
```
SAKavli, line 651: I don't understand these abbreviations either, but not sure if we want to touch them as it makes sense for domain people.

tests/ert/unit_tests/config/test_observations.py
```diff
@@ -660,22 +665,26 @@ def test_that_history_observations_are_loaded(tmpdir, keys, with_ext):
...
-        )
+        ).observations["summary"]
 
-        observations = ert_config.enkf_obs
-        assert [o.observation_key for o in observations] == [local_name]
-        assert observations[local_name].observations[datetime(2014, 9, 11)].value == 1.0
-        assert observations[local_name].observations[datetime(2014, 9, 11)].std == 100.0
+        steps = len(unsmry.steps)
+        assert list(observations["response_key"]) == ["FOPR"] * steps
+        assert list(observations["observations"]) == pytest.approx(
```
SAKavli, line 677: Could you briefly explain why you think property based testing makes sense here?

SAKavli (changes requested):

tests/ert/unit_tests/config/test_observations.py
eivindjahren: Maybe, but relative mode means that the value is taken into account: `computed_error = ERROR * VALUE`.

tests/ert/unit_tests/config/test_observations.py
eivindjahren: I will think about it, but they are rather unique, so I think I would like to keep it independent.

tests/ert/unit_tests/config/test_observations.py
eivindjahren: std means standard deviation which is the same as error. We might want to think about having better naming in the future, but in the model update setting the number is thought of as standard deviation, but in the observation setting it is observational error (measurement error).

tests/ert/unit_tests/config/test_observations.py
eivindjahren: These abbreviations you can find in the opm-flow manual, and we have some references/explanation here: https://github.com/equinor/ert/blob/main/src/ert/config/_read_summary.py

tests/ert/unit_tests/config/test_observations.py
eivindjahren: I am not sure what value we get out of having a property based test here, but since the test was so close to being one, I thought it would be better to move away from using `run_sim()` and instead use `summaries()`. The test then becomes more general, and not a whole lot more complex. Mainly the added complexity is the list comprehension over the steps of the unsmry file.

SAKavli (approved):

PR #11753, opened 2025-09-08
Add before validator for better validation error message. Remove default value for `type`.

### LINKED ISSUE #11476: EVEREST linting fails with zero validation for forward model job options

EVEREST linting fails with no validation error when there is a forward model for flow without the specification of the optional keyword `type` like this:

```yaml
forward_model:
  - distance3 --point-file point.json --target 0.5 0.5 0.5 --out distance
  - job: flow point.json --enable-tuning=true
    results:
      file_name: point.json
      # type: summary
```

Running `everest lint` yields the following output:

```bash
(2025.07.04-py311-rhel8-numpy1) [user]$ everest lint config_minimal.yml 
usage: everest lint <config_file>
everest: error: Loading config file <config_minimal.yml> failed with:
Found 0 validation error:



``` 

Uncommenting `# type: summary` fixes the issue and passes the linting check.

Related #11293 (?)

Although `type` seems to be optional, a reasonable error message would be preferential to this. 

### Steps to reproduce

Run `everest lint...` after adding the job below to the forward model `minimal_config.yml` case:

```yaml
  - job: flow point.json --enable-tuning=true
    results:
      file_name: point.json
      # type: summary
```

- [x] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [x] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

StephanDeHoop:
**Issue**
Resolves #11476 

**Approach**
I never got the `Found 0 validation error:` message when commenting `type`, I got this instead:
```
Found 1 validation error:

line: 45, forward_model -> 1 -> results
    * Unable to extract tag using discriminator 'type' (type=union_tag_not_found)
```

(hence I removed the `bug` tag, if you feel that the tag is warranted please feel free to add it)

In my opinion not a very easy to understand error message, I took the liberty to add this before validator to improve the wording (if it is overkill please let me know I can remove it). Also, the `type` keyword is used a `Discriminator` meaning it is not optional at all (while our keyword-reference in the docs says it is optional). I therefore removed the default value since it is a required keyword (i.e., hence `Discriminator`).

If anyone has an idea why I never got that `Found 0 validation error:` I'm curious to know.

      --exec 'just rapid-tests'`)

## When applicable


src/everest/config/forward_model_config.py
```diff
@@ -38,3 +39,15 @@ class ForwardModelStepConfig(BaseModelWithContextSupport):
         )
         | None
     ) = Field(default=None)
+
+    @model_validator(mode="before")
+    @classmethod
+    def check_discriminator(cls, values: dict[str, Any]) -> dict[str, Any]:
```
verveerpj, line 45: Can you add a test for this case, including checking the error msg?

src/everest/config/forward_model_config.py
```diff
@@ -38,3 +39,15 @@ class ForwardModelStepConfig(BaseModelWithContextSupport):
...
+        results = values.get("results")
+        if results is not None and "type" not in results:
+            raise ValueError(
+                "Missing required field 'type' in 'results'. This field is used to "
```
verveerpj: "used to" -> "needed to"

verveerpj (changes requested):
LGTM, but needs a test.

tests/everest/test_forward_model_config.py
```diff
@@ -0,0 +1,24 @@
...
+
+from everest.config.forward_model_config import ForwardModelStepConfig
+
+
+def test_forward_model_step_config_valid():
+    config = ForwardModelStepConfig(
+        job="example_job",
+        results={"type": "summary", "file_name": "output.txt"},
+    )
+    assert config.results.type == "summary"
```
verveerpj: Maybe this is overkill?

StephanDeHoop: Yea I thought so while writing it. If we only have the missing type test maybe I should amend it to the general config test file (it is massive though)? Or is it OK to have one test in a separate file?

verveerpj: I would put it in the general test file.

verveerpj:
One small last comment. The commit message is too long. According to `CONTRIBUTING.md` the subject line should be max 50, the body at 72.

StephanDeHoop:
> One small last comment. The commit message is too long. According to `CONTRIBUTING.md` the subject line should be max 50, the body at 72.

Ah sorry I wasn't aware of that, will make sure to adhere to these standards in the future. My apologies!

PR #11755, opened 2025-09-08
Have scheduler kill realizations in batches

### LINKED ISSUE #9832: Assess "Too many open files"

jonathan-eq:
**Issue**
Resolves #9832 

**Approach**
* Fix Forward model was terminated message overwriting real error (MAX_RUNTIME) when merging event into snapshot 
* This commit changes the driver.kill(...) to take in a list of
realizations to kill instead of killing them one by one. This should
help us with the TooManyOpenFiles issue by ensuring we call bkill/qdel
in batches instead of invoking the command for every realization.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -70,6 +70,7 @@ class Job:
     DEFAULT_FILE_VERIFICATION_TIMEOUT = 120
     WAIT_PERIOD_FOR_TERM_MESSAGE_TO_CANCEL = 5
     WAIT_PERIOD_FOR_SUBMIT_TO_FINISH = 5
+    WAIT_PERIOD_FOR_SCHEDULER_TO_KILL_IN_BATCH = 45
 
     def __init__(self, scheduler: Scheduler, real: Realization) -> None:
         self.real = real
@@ -86,6 +87,7 @@ class Job:
         self._previous_file_verification_time = self._remaining_file_verification_time
         self._started_killing_by_evaluator: bool = False
         self._was_killed_by_evaluator = asyncio.Event()
+        self._was_killed_by_scheduler = asyncio.Event()
 
     @property
     def remaining_file_verification_time(self) -> int:
@@ -199,7 +201,17 @@ class Job:
                         f"Realization {self.iens} was not killed gracefully by "
                         "TERM message. Killing it with the driver"
                     )
-                await self.driver.kill(self.iens)
+                await self._scheduler.schedule_kill(self.iens)
+                try:
+                    await asyncio.wait_for(
+                        self._was_killed_by_scheduler.wait(),
+                        timeout=self.WAIT_PERIOD_FOR_SCHEDULER_TO_KILL_IN_BATCH,
+                    )
+                except TimeoutError:
+                    logger.warning(
+                        f"Realization {self.iens} did not get confirmation from "
+                        "scheduler that the batch killing with driver ran successfully."
+                    )
             else:
                 logger.info(f"Realization {self.iens} was killed by the evaluator")
             with suppress(asyncio.CancelledError):
--- a/src/ert/scheduler/local_driver.py
+++ b/src/ert/scheduler/local_driver.py
@@ -5,7 +5,7 @@ import contextlib
 import logging
 import signal
 from asyncio.subprocess import Process
-from collections.abc import MutableMapping
+from collections.abc import Iterable, MutableMapping
 from contextlib import suppress
 from pathlib import Path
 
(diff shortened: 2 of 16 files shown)


src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -154,6 +154,10 @@ def merge_snapshot(self, ensemble: EnsembleSnapshot) -> EnsembleSnapshot:
         for real_id, other_real_data in ensemble._realization_snapshots.items():
             self._realization_snapshots[real_id].update(other_real_data)
         for fm_step_id, other_fm_data in ensemble._fm_step_snapshots.items():
+            if (
+                previous_error := self._fm_step_snapshots[fm_step_id].get("error")
+            ) and other_fm_data.get("error") == FORWARD_MODEL_TERMINATED_MSG:
+                other_fm_data["error"] = previous_error
```
jonathan-eq, line 160: This was a strange timing related bug that was created when we added the FORWARD_MODEL_TERMINATED_MSG in 37bc811564d8e040e7aaf3ac4ded94b82194dfb4. If a RealizationFailed event with a message comes in batch A, and as a consequence (example due to MAX_RUNTIME) a ForwardModelStepFailure event with message FORWARD_MODEL_TERMINATED_MSG comes in batch B; the second "irrelevant" error message will overwrite the initial one.

jonathan-eq: In the same commit, we took into account this timing condition if it was in the same batch, but not for different batches.

xjules: Why are we overwriting the error message instead of appending?

jonathan-eq: The FORWARD_MODEL_TERMINATED_MSG is the default, and it is uninteresting if we have an actual error message.

src/ert/scheduler/local_driver.py
```diff
@@ -39,21 +39,24 @@ async def submit(
...
-            await self._dispatch_finished_event(iens, signal.SIGTERM + SIGNAL_OFFSET)
-
-        except KeyError:
-            logger.info(f"Realization {iens} is already killed")
-            return
-        except Exception as err:
-            logger.error(f"Killing realization {iens} failed with error {err}")
-            raise err
+    async def kill(self, realizations: Iterable[int]) -> None:
+        for realization in realizations:
```
xjules, line 43: I think to think about this for a while :)

jonathan-eq: It is the same as before, just a loop.

src/ert/scheduler/job.py
```diff
@@ -199,7 +200,8 @@ async def _submit_and_run_once(self, sem: asyncio.BoundedSemaphore) -> None:
                         f"Realization {self.iens} was not killed gracefully by "
                         "TERM message. Killing it with the driver"
                     )
-                await self.driver.kill(self.iens)
+                await self._scheduler.schedule_kill(self.iens)
+                await self._was_killed_by_scheduler.wait()
```
jonathan-eq: Add timeout so it won't hang.

jonathan-eq: Maybe a test too^

jonathan-eq:
This seems to have worked 
```
    async def _batch_killing(self) -> None:
        kill_tasks = []
        while True:
            await self._wait_for_next_batch()
            if len(self._realization_ids_to_kill) > 0:
                realizations_to_kill = self._realization_ids_to_kill
                self._realization_ids_to_kill = []
                logger.info(
                    "Scheduler killing a batch of "
                    f"{len(realizations_to_kill)} realizations"
                )
                start_time = time.time()
                kill_tasks.append( asyncio.create_task(self.driver.kill(realizations_to_kill)))
                logger.info(f"Scheduler killed batch in {time.time() - start_time}s")
                for realization in realizations_to_kill:
                    self._jobs[realization]._was_killed_by_scheduler.set()
            if (
                self._stop_kill_task.is_set()
                and len(self._realization_ids_to_kill) == 0
            ):
                await asyncio.gather(*kill_tasks)
                return
```

src/ert/scheduler/job.py
```diff
@@ -70,6 +70,7 @@ class Job:
     DEFAULT_FILE_VERIFICATION_TIMEOUT = 120
     WAIT_PERIOD_FOR_TERM_MESSAGE_TO_CANCEL = 5
     WAIT_PERIOD_FOR_SUBMIT_TO_FINISH = 5
+    WAIT_PERIOD_FOR_SCHEDULER_TO_KILL_IN_BATCH = 45
```
jonathan-eq, line 73: This often takes a lot of time if there are a lot of job to kill, especially if most of them are not running yet. 300 jobs (mostly pending) takes 13-15s.

tests/ert/unit_tests/scheduler/test_scheduler.py
```diff
@@ -742,3 +746,145 @@ async def test_log_warnings_from_forward_model_is_run_once_per_ensemble(
...
+
+    monkeypatch.setattr(
+        scheduler.Scheduler, "_wait_for_next_batch", mock_wait_for_next_batch
+    )
+    mock_driver = MagicMock()
+    mock_driver.kill = AsyncMock()
+    sch = scheduler.Scheduler(mock_driver, realizations=realizations)
+
+    first_batch = True
+    for odd_or_even_reminder in [0, 1]:
```
flikka: I find this a bit hard to read, this odd_even_reminder loop. Is it possible to explain a bit with comments, ore perhaps rename things/rewrite a bit?  Is the point to have two batches, some realizations in one and others in the other? Perhaps just rename odd_or_even_reminder to "batch number" (and in the expected batch, just say if i%2: batch 0, else batch 1? And some comments?

src/ert/scheduler/scheduler.py
```diff
@@ -372,3 +380,56 @@ def _update_jobs_json(
...
+        while True:
+            await self._wait_for_next_batch()
+            if len(self._realization_ids_to_kill) > 0:
+                realizations_to_kill = self._realization_ids_to_kill
+                self._realization_ids_to_kill = []
+                kill_tasks.append(asyncio.create_task(self._kill(realizations_to_kill)))
+
+            if (
+                self._stop_kill_task.is_set()
+                and len(self._realization_ids_to_kill) == 0
```
flikka: I guess you couldn't just put this up into the while condition? Don't like the while True in general..

jonathan-eq: Yes, I think we can. This should maybe also be done in Evaluator, all of the tasks there use `while True:..`

PR #11756, opened 2025-09-08
Change everest runpath naming scheme

### LINKED ISSUE #10018: Improve identification of everest output simulation folders
Since there are on-going discussions related to renaming the `geo_realization_<id>` folders in the output simulation directory created by EVEREST, we have the opportunity to introduce another improvement. It is common feedback from new users (also from our recent EVEREST course at Equinor) that it is not always straightforward to identify which simulation folders correspond to "current control evaluation" simulations and which simulation folders correspond to "perturbation" simulations. I propose to rename the simulation folders to indicate that more clearly. See below an example of my proposed solution.

For a case with `perturbation_num: 2`, `realizations: [0, 1, 2]` and `speculative: true`, we could have: 

```
- batch_0
    - realization_0
        - evaluation
        - perturbation_0
        - perturbation_1 
    - realization_1
        - evaluation
        - perturbation_0
        - perturbation_1 
    - realization_2
        - evaluation
        - perturbation_0
        - perturbation_1 
```

In current situation, the same setup would result in the following (less user-friendly / informative) folder naming structure: 

```
- batch_0
    - geo_realization_0
        - simulation_0
        - simulation_3
        - simulation_4 
    - geo_realization_1
        - simulation_1
        - simulation_5
        - simulation_6 
    - geo_realization_2
        - simulation_2
        - simulation_7
        - simulation_8 
```

@verveerpj @StephanDeHoop @slawekszklarz

verveerpj:
**Issue**
Resolves #10018

**Approach**
- rename simulations that are simple evaluations to `evaluation_<idx>`
- rename simulations that are perturbations to `perturbation_<idx>`
- rename the `geo_realization` component of the path to `realization`

where `idx` re-starts from zero each batch and each realization, for both evaluations and perturbations.

This is implemented by an additional replacement key `<SIM_ID>` similar to the existing `<GEO_ID>` key.

      --exec 'just rapid-tests'`)

## When applicable


berland:
There are some bitmap pictures in the documentation, e.g. on https://everest.readthedocs.io/en/latest/development.html that at some point should maybe be updated when doing this change (they might not even be totally in sync with what is in Everest today).

verveerpj:
> There are some bitmap pictures in the documentation, e.g. on https://everest.readthedocs.io/en/latest/development.html that at some point should maybe be updated when doing this change (they might not even be totally in sync with what is in Everest today).

That is true, unfortunately they are png's, we are going to need the source files.

berland:
This should perhaps be replaced by SIM_ID?
https://github.com/equinor/ert/blob/cc143be29bf5b26e5479cf611ce95aff7b04e8a7/docs/everest/config.rst?plain=1#L36

berland:
Can you read through development.rst also and check if any updates apply there?

verveerpj:
> Can you read through development.rst also and check if any updates apply there?

The text is fine, but the figures need updating. I will try to track down the sources.

verveerpj:
> > Can you read through development.rst also and check if any updates apply there?
> 
> The text is fine, but the figures need updating. I will try to track down the sources.

I have now update the figure and the text.

docs/everest/images/Everest_vs_Ert_01.png
berland, line 1: This image still says `model_realization`, it looks like this is changed to `realization`?

berland (approved):
I am ok with this, but let us try to hear from some stakeholders: @oyvindeide, Tudor Popa

tup1985:
> geo_realization -> realization | model_realization, we should pick one but currently have a mix that seems a bit ad-hoc/random, we should agree on one thing to go for and go for only that

I have the same comment, as I browsed through the changes. The rest of the nesting structure sounds ok to me, `experiment/batch/realization/evaluation, perturbation`
I am not concerned how naming is being handled internally, but for the user the naming should be consistent.

PR #11781, opened 2025-09-10
Make everserver tolerate intermittent httpx glitches

### LINKED ISSUE #11776: Everserver crashes on httpx.RemoteProtocolError

berland:
There are rare occurences where a httpx.RemoteProtocolError is thrown on the client side when it is talking to a experiment server even on the same host.

With this patch, this exception is caught and ignored when it happens during normal status polling, since status polling is retried twice a second we can just wait and see if the next connection attempt is successful.

For failsafe reasons, the client is also recreated, as the reason for the glitch can be connected to this object as well. The client object keeps a connection pool, and the server may have decided to close some of the long-lived keep-alive connections to save resources. There is no builtin retry mechanism in httpx for this situation.

This unfortunately means that a the context manager goes away, as it will not support recreating itself on failure. In this commit, that affects the mocking of the service.

**Issue**
Resolves #11776 

**Approach**
Ignore and retry

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/detached/everserver.py
+++ b/src/everest/detached/everserver.py
@@ -11,6 +11,7 @@ from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import Any
 
+import httpx
 import yaml
 from cryptography import x509
 from cryptography.hazmat.backends import default_backend
@@ -20,6 +21,7 @@ from cryptography.x509.oid import NameOID
 from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
 from pydantic import BaseModel
 
+from ert.dark_storage.client import Client as DarkStorageClient
 from ert.ensemble_evaluator import (
     EnsembleSnapshot,
     FullSnapshotEvent,
@@ -166,10 +168,14 @@ def main() -> None:
         else None
     )
 
+    def _create_client(server_path: str) -> DarkStorageClient:
+        return StorageService.session(project=server_path)
+
     with (
         tracer.start_as_current_span("everest.everserver", context=ctx),
         NamedTemporaryFile() as log_file,
     ):
+        client = None
         try:
             _configure_loggers(
                 detached_dir=Path(ServerConfig.get_detached_node_dir(output_dir)),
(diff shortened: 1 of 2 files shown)


src/everest/detached/everserver.py
```diff
@@ -189,19 +195,30 @@ def main() -> None:
                 timeout=240, project=server_path, logging_config=log_file.name
             ) as server:
                 server.fetch_conn_info()
-                with StorageService.session(project=server_path) as client:
-                    update_everserver_status(status_path, ExperimentState.running)
-                    done = False
-                    while not done:
+                client = _create_client(server_path)
+                update_everserver_status(status_path, ExperimentState.running)
```
yngve-sk, line 199: completely unrelated to PR but I wonder if this makes sense

src/everest/detached/everserver.py
```diff
@@ -189,19 +195,30 @@ def main() -> None:
...
                             "/experiment_server/status", auth=server.fetch_auth()
                         )
-                        status = ExperimentStatus(**response.json())
-                        done = status.status not in {
-                            ExperimentState.pending,
-                            ExperimentState.running,
-                        }
-                        time.sleep(0.5)
+                    except httpx.RemoteProtocolError:
+                        logger.warning(
```
yngve-sk, line 207: Sure `httpx.RemoteProtocolError` is the only error we need to handle for this?

src/everest/detached/everserver.py
```diff
@@ -166,10 +168,14 @@ def main() -> None:
         else None
     )
 
+    def _create_client(server_path: str) -> DarkStorageClient:
+        return StorageService.session(project=server_path)
```
yngve-sk, line 172: I'm assuming the `httpx.Client` has same behavior when used as context manager and just as a regular obj?

src/everest/detached/everserver.py
berland: I assumed it was kind of bootstrapping. 🤷🏻

src/everest/detached/everserver.py
berland: Yes, this pattern is listed in the docs: https://www.python-httpx.org/advanced/clients/#usage

src/everest/detached/everserver.py
yngve-sk: Seems it just defaults to think it is running until it hears otherwise 🙈 Ideally the experiment status should come only from querying the experiment server

src/everest/detached/everserver.py
berland: There are other related, but none that we have observed IRL.

PR #11797, opened 2025-09-12
Simplify error handling in _create_observations.py

eivindjahren:
Simplify validation in _create_observations.py by avoiding to throw ValueError followed by catch and reraise ObservationConfigError.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/_create_observations.py
+++ b/src/ert/config/_create_observations.py
@@ -43,10 +43,6 @@ class _SummaryObservation:
     std: float
     std_scaling: float = 1.0
 
-    def __post_init__(self) -> None:
-        if self.std <= 0:
-            raise ValueError("Observation uncertainty must be strictly > 0")
-
 
 @dataclass(eq=False)
 class _GenObservation:
@@ -283,12 +279,11 @@ def _handle_history_observation(
         )
     data: dict[datetime, _SummaryObservation] = {}
     for date, error, value in zip(refcase.dates, std_dev, values, strict=False):
-        try:
-            data[date] = _SummaryObservation(
-                summary_key, summary_key, value, float(error)
-            )
-        except ValueError as err:
-            raise ObservationConfigError.with_context(str(err), summary_key) from None
+        if float(error) <= 0:
+            raise ObservationConfigError.with_context(
+                "Observation uncertainty must be strictly > 0", summary_key
+            ) from None
+        data[date] = _SummaryObservation(summary_key, summary_key, value, float(error))
 
     return {
         summary_key: _SummaryObs(
@@ -428,6 +423,10 @@ def _handle_summary_observation(
             obs_key,
         )
     try:
+        if float(std_dev) <= 0:
+            raise ObservationConfigError.with_context(
+                "Observation uncertainty must be strictly > 0", summary_key
+            ) from None
         return {
             obs_key: _SummaryObs(
                 summary_key,


src/ert/config/_create_observations.py
SAKavli, line 1: I assume commit message was intended to say `_create_ ...`

src/ert/config/_create_observations.py
```diff
@@ -393,51 +391,51 @@ def _handle_summary_observation(
...
+        restart = _get_restart(summary_dict, obs_key, time_map, has_refcase)
+        date = time_map[restart]
 
     if restart == 0:
         raise ObservationConfigError.with_context(
             "It is unfortunately not possible to use summary "
             "observations from the start of the simulation. "
             f"Problem with observation {obs_key}"
-            f"{' at ' + str(_get_time(summary_dict, time_map[0])) if summary_dict.restart is None else ''}",  # noqa: E501
+            f"{' at ' + str(_get_time(summary_dict, time_map[0], obs_key)) if summary_dict.restart is None else ''}",  # noqa: E501
```
SAKavli, line 422: What happens if `_get_time()` raises ValueError inside here?

tests/ert/unit_tests/config/test_observations.py
```diff
@@ -676,6 +676,53 @@ def test_that_non_existent_time_map_file_is_invalid():
...
+                        "VALUE": "1.0",
+                        "ERROR": "0.1",
+                    },
+                )
+            ],
+            parse=False,
+        )
+
+
+def test_that_general_observation_must_contain_either_value_or_obs_file():
```
SAKavli, line 707: The title of this test first gave me the impression we are testing for both values present like in the previous test. Can we emphasize we are testing that the absence of both is raising the error?

src/ert/config/_create_observations.py
```diff
@@ -538,27 +539,25 @@ def _handle_general_observation(
...
+                            general_observation.error,
+                        )
+                        if general_observation.value is not None
+                        and general_observation.error is not None
+                        else None
                     ),
-                },
-            )
-        }
-    except ValueError as err:
```
SAKavli, line 564: The reason for removing this catch statement is because it should be caught in the outer scope where the method is used, yes?

src/ert/config/_create_observations.py
```diff
@@ -475,9 +473,12 @@ def _create_gen_obs(
...
             obs_file if obs_file is not None else "",
         )
-    return GenObservation(
-        values.tolist(), stds.tolist(), indices.tolist(), std_scaling.tolist()
-    )
+    try:
+        return GenObservation(
+            values.tolist(), stds.tolist(), indices.tolist(), std_scaling.tolist()
+        )
+    except ValueError as err:
```
SAKavli: From the scope of this PR, this method is called from inside `_handle_general_observation`. In that method, we removed errorhandling as it was caught from outer scope. Won't the same apply here, or is this called from elsewhere without handling?

src/ert/config/_create_observations.py
eivindjahren: You get an "error raised while handling exception". I have tried to figure out whether that is possible, but I want to tackle that separately.

src/ert/config/_create_observations.py
eivindjahren: Seems like GenObservation is only called here. I will look into simplifying that also.

eivindjahren: I took care of that also.

src/ert/config/_create_observations.py
eivindjahren: The catching of the ValueError is now moved in. So we are more specific about which ValueError we catch. However, I also moved those ValueErrors so that we can just throw an ObservationConfigError initially and not have to rethrow anything.

src/ert/config/_create_observations.py
```diff
@@ -391,54 +394,51 @@ def _handle_summary_observation(
...
+            "GENERAL_OBSERVATION must contain either VALUE and ERROR or OBS_FILE",
+            context=context,
         )
 
     if scalar_value is not None and obs_file is not None:
-        raise ValueError(
-            "Exactly one the scalar_value and obs_file arguments must be present"
+        raise ObservationConfigError.with_context(
+            "GENERAL_OBSERVATION cannot contain both VALUE/ERROR and OBS_FILE",
+            context=obs_file,
```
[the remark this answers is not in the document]
eivindjahren: No the context is obs_file on purpose here.

src/ert/config/_create_observations.py
```diff
@@ -265,8 +263,16 @@ def _handle_history_observation(
             segment_instance,
         )
     data: dict[datetime, SummaryObservation] = {}
+    errors: list[ObservationConfigError] = []
```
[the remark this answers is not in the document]
eivindjahren: This is not correct.

PR #11800, opened 2025-09-12
Log usage of observation keywords

### LINKED ISSUE #11815: Log usage of all observation types and observation keywords
If the users observations config file contains
```
        GENERAL_OBSERVATION MY_OBS {
            DATA       = MY_RESPONSE;
            INDEX_LIST = 0,2,4,6,8;
            RESTART    = 0;
            OBS_FILE   = obs.txt;
        };
```
we want to log that `GENERAL_OBSERVATION`, `DATA`, `INDEX_LIST`, `RESTART` and `OBS_FILE` is used. We do not want to log any of the values.

Documentation of which types and keywords exists can be found here: https://github.com/equinor/ert/blob/main/docs/ert/reference/configuration/observations.rst

SAKavli:
Resolves #11815 
Resolves https://github.com/equinor/scout/issues/1349

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/ert_config.py
+++ b/src/ert/config/ert_config.py
@@ -5,7 +5,7 @@ import logging
 import os
 import pprint
 import re
-from collections import defaultdict
+from collections import Counter, defaultdict
 from collections.abc import Mapping
 from datetime import datetime
 from functools import cached_property
@@ -665,23 +665,17 @@ def create_list_of_forward_model_steps_to_run(
 def log_observation_keys(
     observations: list[SimpleHistoryStatement | ObservationStatement],
 ) -> None:
-    observation_types_count: dict[str, int] = defaultdict(int)
-    observation_keywords_count: dict[str, int] = defaultdict(int)
-
-    for observation in observations:
-        observation_type = getattr(observation[0], "value", observation[0])
-        observation_types_count[observation_type] += 1
-
-        if observation_body := (observation[2] if len(observation) == 3 else None):
-            for key in observation_body:
-                if isinstance(key, tuple):
-                    observation_keywords_count["SEGMENT"] += 1
-                else:
-                    observation_keywords_count[key] += 1
+    observation_type_counts = Counter(str(o[0]) for o in observations)
+    observation_keyword_counts = Counter(
+        "SEGMENT" if isinstance(key, tuple) else str(key)
+        for o in observations
+        if len(o) == 3
+        for key in o[2]
+    )
 
     logger.info(
-        f"Count of observation types:\n\t{dict(observation_types_count)}\n"
-        f"Count of observation keywords:\n\t{dict(observation_keywords_count)}"
+        f"Count of observation types:\n\t{dict(observation_type_counts)}\n"
+        f"Count of observation keywords:\n\t{dict(observation_keyword_counts)}"
     )
 
 
--- a/tests/ert/unit_tests/config/test_observations.py
+++ b/tests/ert/unit_tests/config/test_observations.py
@@ -52,8 +52,8 @@ def run_simulator(summary_values=SUMMARY_VALUES):
     """
     summary = Summary.writer("MY_REFCASE", datetime(2000, 1, 1), 10, 10, 10)
 
-    summary.add_variable("FOPR", unit="SM3/DAY")
-    summary.add_variable("FOPRH", unit="SM3/DAY")
+    for key in summary_values:
+        summary.add_variable(key, unit="SM3/DAY")
 
     mini_step_count = 10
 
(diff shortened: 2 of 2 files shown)


tests/ert/unit_tests/config/test_ert_config.py
```diff
@@ -2479,3 +2479,19 @@ def test_that_the_eclbase_keyword_sets_the_eclbase_substitution(eclbase_substitu
         ).substitutions[eclbase_substitution]
         == datafile
     )
+
+
+@pytest.mark.usefixtures("copy_snake_oil_case")
+def test_ert_config_logs_observation_types_and_keywords(caplog):
+    with caplog.at_level(logging.INFO):
+        _ = ErtConfig.from_file("snake_oil.ert")
+    assert "Count of observation types" in caplog.text
```
SAKavli: Are these string comparisons too slow from a testing perspective? As `caplog.text` is quite large.

eivindjahren: On mac, `copy_snake_oil_case` is quite slow, but the string comparison is not that bad. You could probably speed things up by creating your test data on the fly: 
```python
obs_config_contents = """
            SUMMARY_OBSERVATION FOPR
            {
                VALUE = 1;
                KEY = FOPR;
            };
"""
_ = ErtConfig.from_dict(
        {
            "NUM_REALIZATIONS": 1,
            "ECLBASE": "BASEBASEBASE",
            "SUMMARY": "*",
            "GEN_DATA": [["GEN", {"RESULT_FILE": "gen%d.txt", "REPORT_STEPS": "1"}]],
            "TIME_MAP": ("time_map.txt", "2020-01-01\n2020-01-02\n"),
            "OBS_CONFIG": (
                "obs_config_file",
                parse_observations(obs_config_contents, "obs_config_file"),
            ),
        }
    )
```
Then you can create multiple tests with different `obs_config_content` to write out what you expect to be present in the log message.

eivindjahren: If you put this test in `test_observations.py` then there is a function `make_observations` you can use to do this.

SAKavli: Neat, I will try that!

SAKavli: I force pushed since I forget to use --fixup flag in commit, sorry about that! Have fixed it now.

src/ert/config/ert_config.py
```diff
@@ -660,6 +662,29 @@ def create_list_of_forward_model_steps_to_run(
     return fm_steps
 
 
+def log_observation_keys(
+    observations: list[SimpleHistoryStatement | ObservationStatement],
+) -> None:
+    observation_types_count: dict[str, int] = defaultdict(int)
```
eivindjahren: You probably want to use a Counter here: https://docs.python.org/3/library/collections.html#counter-objects

src/ert/config/ert_config.py
```diff
@@ -660,6 +662,29 @@ def create_list_of_forward_model_steps_to_run(
...
+    observation_keywords_count: dict[str, int] = defaultdict(int)
+
+    for observation in observations:
```
eivindjahren: I think we can use:
 ```python
observation_type_counts = Counter(
    str(o[0]) for o in observations
)
observation_keyword_counts = Counter(
  "SEGMENT" if isinstance(key, tuple) else str(key)
  for o in observations
  for key in o[2] if len(o) == 3
)
```

Comment (SAKavli) [reply src/ert/config/ert_config.py]: This is beautiful, thanks @eivindjahren !

Comment (eivindjahren) [reply src/ert/config/ert_config.py]: Well, maybe but turns out it didn't work :D . I'll help fix it tomorrow :)

Comment (eivindjahren) [reply src/ert/config/ert_config.py]: I wonder whether python lets us put the if statement between the for statements because conceptually that is what we need to do. Another option is to normalize the output from `parse_observations` so that we always have a triple.

Comment (eivindjahren) [reply src/ert/config/ert_config.py]: Just checked, and it works correctly with 
```python
observation_keyword_counts = Counter(
 "SEGMENT" if isinstance(key, tuple) else str(key)
 for o in observations
 if len(o) == 3
 for key in o[2]
)
```

SAKavli: I will try this! Will the generator skip the iterations if `len(o) != 3`?

tests/ert/unit_tests/config/test_observations.py
```diff
@@ -1703,3 +1704,27 @@ def test_that_setting_an_unknown_key_in_a_segment_is_not_valid():
                 };
             };
         """)
+
+
+def test_ert_config_logs_observation_types_and_keywords(caplog):
+    obs_config_contents = """
+    GENERAL_OBSERVATION OBS1 {
```
eivindjahren, line 1712: Maybe put in `HISTORY_OBSERVATION FOPR;` and a use of `SEGMENT` into this test, then you have to use `make_refcase_observations` and add the `use_tmpdir` fixture.

src/ert/config/ert_config.py
eivindjahren: > I will try this! Will the generator skip the iterations if `len(o) != 3`?

Yes :)

eivindjahren:
> Do you want me to backport this @eivindjahren ?

Yes

PR #11812, opened 2025-09-15
Split width validation into field validator for errf distribution

dafeda:
      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/distribution.py
+++ b/src/ert/config/distribution.py
@@ -228,25 +228,22 @@ class ErrfSettings(TransSettingsValidation):
     skewness: float = 0.0
     width: float = 1.0
 
+    @field_validator("width")
+    @classmethod
+    def width_must_be_positive(cls, value: float) -> float:
+        if value <= 0:
+            raise ConfigValidationError(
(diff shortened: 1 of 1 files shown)


src/ert/config/distribution.py
```diff
@@ -228,25 +228,22 @@ class ErrfSettings(TransSettingsValidation):
...
-    @model_validator(mode="after")
-    def valid_errf_params(self) -> Self:
-        errors = []
-        if not self.width > 0:
-            errors.append(
-                ErrorInfo(
-                    message=f"The width {self.width} must be greater than"
-                    " 0 for errf distribution"
-                )
+    @field_validator("width")
```
jonathan-eq, line 231: Previously, we got all the errors for both width validation and min/max. Now that these will be different validation methods, will we get all errors or just the first?

jonathan-eq (commented):
Good change 👍

src/ert/config/distribution.py
dafeda: Did not think of this. I think field_validators run first so we'll get a single error.

PR #11830, opened 2025-09-15
Change storage precision and data size for parameters to 32bit

xjules:
Update step requires 32bit so there is no need to store 64bit dataframes
**Approach**
 - 64bit -> 32bit
 - update snapshot
   - For float32 snapshots rounding up to 7 digits is more realistic

- [ ] change the snapshots

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/ui_tests/cli/test_cli.py
+++ b/tests/ert/ui_tests/cli/test_cli.py
@@ -541,7 +541,7 @@ def test_that_es_mda_on_poly_case_matches_snapshot(snapshot):
             ensemble_data.columns.name = None
             ensemble_data.index.name = "Realization"
             ensemble_data = ensemble_data.sort_index(axis=1)
-            data.append(ensemble_data)
+            data.append(ensemble_data.astype("float32"))
     result = pd.concat(
         data,
         keys=[f"iter-{iter_}" for iter_ in range(len(data))],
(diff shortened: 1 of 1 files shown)


xjules:
> Seem sensible. But, do we have a general "philosophy" on float precision? For example the parameters stuff here:
> 
> https://github.com/equinor/ert/blob/26203ad8e1377cf10e5420ce3cc4fd3511726273/src/ert/config/gen_kw_config.py#L320
> 
> should that also be 32 bit?
> And tests are failing :-/

When it is stored as 32bit the schema should say 32 bit. The tests are failing due to the snapshots.
I'll labeled this as blocked due to merging the genkw PR first.

src/ert/analysis/misfit_preprocessor.py
```diff
@@ -61,7 +61,7 @@ def get_nr_primary_components(
 
 
 def cluster_responses(
-    responses: npt.NDArray[np.float64],
+    responses: npt.NDArray[np.float32],
```
JHolba, line 64: should this be np.floating instead? it seems responses can be np.float64 coming from main?

xjules: yes, probably it should be `np.floating` I have changed it in some places :smile:

tests/ert/ui_tests/cli/test_cli.py
```diff
@@ -553,7 +553,7 @@ def test_that_es_mda_on_poly_case_matches_snapshot(snapshot):
         numpy_suffix = "_numpy1"
 
     snapshot.assert_match(
-        result.to_csv(float_format="%.12g"),
+        result.to_csv(float_format="%.7g"),
```
flikka, line 556: Why 7?

xjules: I need to revisit the precision.

PR #11834, opened 2025-09-16
Use ensemble.id as dict key as name is not unique

### LINKED ISSUE #11832: Inconsistent values when plotting

When running several different experiments with different random seed, I see a difference in the response values, but the parameters appear to be the same for all experiments, related to: [11775](https://github.com/equinor/ert/issues/11775)

Responses:

Image

Parameters:

Image

That both responses and parameters are different between experiments

### Steps to reproduce

```Python
Open ert
Run ensemble experiment
Close ert to refresh the random seed
Open ert
Run ensemble experiment
Open plotter and inspect the responses and parameters
```

- [ ] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

dafeda:
**Issue**
Resolves #11832

**Approach**
`ensemble.name` is not unique across experiments so we should use `ensemble.id` as key in dict.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/gui/tools/plot/plottery/plots/histogram.py
+++ b/src/ert/gui/tools/plot/plottery/plots/histogram.py
@@ -66,32 +66,33 @@ def plotHistogram(
     categories: set[str] = set()
     max_element_count = 0
     categorical = False
+
     for ensemble, datas in ensemble_to_data_map.items():
         if datas.empty:
-            data[ensemble.name] = pd.Series(dtype="float64")
+            data[ensemble.id] = pd.Series(dtype="float64")
             continue
 
-        data[ensemble.name] = datas[0]
+        data[ensemble.id] = datas[0]
 
-        if data[ensemble.name].dtype == "object":
+        if data[ensemble.id].dtype == "object":
             try:
-                data[ensemble.name] = convert_to_numeric(data[ensemble.name])
+                data[ensemble.id] = convert_to_numeric(data[ensemble.id])
             except AttributeError:
-                data[ensemble.name] = data[ensemble.name].convert_objects(
+                data[ensemble.id] = data[ensemble.id].convert_objects(
                     convert_numeric=True
                 )
 
-        if data[ensemble.name].dtype == "object":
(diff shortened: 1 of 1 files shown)


src/ert/gui/tools/plot/plottery/plots/histogram.py
```diff
@@ -66,32 +66,33 @@ def plotHistogram(
     categories: set[str] = set()
     max_element_count = 0
     categorical = False
+
     for ensemble, datas in ensemble_to_data_map.items():
         if datas.empty:
-            data[ensemble.name] = pd.Series(dtype="float64")
+            data[ensemble.id] = pd.Series(dtype="float64")
```
xjules, line 72: would be nice to do type annotations on `data` (line 63)?

dafeda: Yeah maybe, but I'm just fixing a bug here.

dafeda:
> test_histogram_plot_for_constant_distribution

Don't see a straight forward way to adapt that test to this specific case. Do you?

src/ert/gui/tools/plot/plottery/plots/histogram.py
xjules: yes, but this would help regardless to identify what data actually is.

xjules:
> > test_histogram_plot_for_constant_distribution
> 
> Don't see a straight forward way to adapt that test to this specific case. Do you?

There are a couple of thoughts but let just say we don't do it now.

src/ert/gui/tools/plot/plottery/plots/histogram.py
xjules: `data: dict[UUID, pd.Series] = {}`
or something along those lines

xjules (approved):

Issue #11775, opened 2025-09-10
Histogram is empty for single realization test run


SAKavli:
### What happened? (You can include a screenshot if it helps explain)

Some iffy behavior here:

https://github.com/user-attachments/assets/46f537a9-810b-4b0e-8ed8-2eefc3a22763

### Steps to reproduce

Run poly example (single test run) and inspect parameters in create plot

### Environment where bug has been observed

- [ ] Python 3.11
- [ ] Python 3.12
- [x] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

oyvindeide:
I edited your issue a bit, as there was multiple in one. The random seed not refreshing in the same session is by design, but see that it can be confusing, but there is a plotting bug which I split into: https://github.com/equinor/ert/issues/11832

PR #11836, opened 2025-09-16
Hist single test run bug

dafeda:
**Issue**
Resolves [#11775](https://github.com/equinor/ert/issues/11775)

# Snapshot failure

The code on `main` produces this histogram:

image

The code on this `branch` produces this:

image

This is the data being plotted:

```python
np.array([3.033689, 3.478627, 4.760788, 4.424522, 4.571844])
```

Default `matplotlib` gives this:

```python
import matplotlib.pyplot as plt
import numpy as np

data = np.array([3.033689, 3.478627, 4.760788, 4.424522, 4.571844])

_ = plt.hist(data)
```

image

To me, the new histogram is a better representation of the data as it shows its bimodal nature. There's a clear separation between lower values (3.03, 3.48) and higher values (4.42, 4.57, 4.76).

      --exec 'just rapid-tests'`)

## When applicable


src/ert/gui/tools/plot/plottery/plots/histogram.py
```diff
@@ -184,7 +184,9 @@ def _plotHistogram(
 ) -> Rectangle:
     bins: Sequence[float] | int
     if minimum is not None and maximum is not None:
-        bins = np.linspace(minimum, maximum, bin_count)  # type: ignore
+        # Ensure we have at least 2 bin edges to create 1 bin
```
xjules, line 187: In case of test run, one bin will be empty. Which is alright I guess.

Issue #9844, opened 2025-01-23
guidance/validation on perturbation magnitude?


yngve-sk:
Should not be too big or too small, exact details should be figured out, but this could/should be validated in the config

PR #11843, opened 2025-09-17
Update the everest keyword reference docs

### LINKED ISSUE #11734: Autogenerated documentation for Everest preserves irrelevant linebreaks
Looking at the screenshot:

Image

there are linebreaks in the wrong places, and these exact linebreaks can be seen to stem from the source code:

https://github.com/equinor/ert/blob/abaa2de84d316dedba8440236641a73b4ad9973c/src/everest/config/input_constraint_config.py#L17-L19

Expected behaviour is that these linebreaks from the source code are not preserved, and the html renderer should itself decide where to put linebreaks.

verveerpj:
**Issue**
Resolves #11734

**Approach**
- Fix `json_schema_for_humans` configuration
- Update individual key descriptions.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/docs/everest/conf.py
+++ b/docs/everest/conf.py
@@ -49,6 +49,9 @@ config = GenerationConfiguration(
     with_footer=False,
     collapse_long_descriptions=False,
     collapse_long_examples=False,
+    description_is_markdown=True,
+    markdown_options={"breaks": {"on_newline": False}},
+    deprecated_from_description=True,
 )
 with open("config_schema.json", "w", encoding="utf-8") as fout:
     json.dump(EverestConfig.model_json_schema(), fout)
--- a/src/everest/config/control_config.py
+++ b/src/everest/config/control_config.py
@@ -1,5 +1,6 @@
 import logging
 from itertools import chain
+from textwrap import dedent
 from typing import (
     Annotated,
     Any,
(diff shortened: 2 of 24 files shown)


src/everest/config/control_config.py
```diff
@@ -49,124 +50,171 @@ class ControlConfig(BaseModel):
...
+
+            The `perturbation_type` keyword defines whether the perturbation
+            magnitude (`perturbation_magnitude`) should be treated as an
+            absolute value or as relative to the dynamic range of the controls.
+
+            Note: currently the dynamic range is computed with respect to all
+            controls, so defining a relative perturbation type for control types
+            with different dynamic ranges might have unintended effects.
+            """
+        ),
```
andreas-el, line 168: Just curious, why isn't this setting a bool instead?

src/everest/config/control_config.py
```diff
@@ -49,124 +50,171 @@ class ControlConfig(BaseModel):
...
-Initial guess for the control group all control variables with initial_guess not
-defined will be assigned this value. Individual initial_guess values in the control
-variables will overwrite this value.
-""",
+        description=dedent(
+            """
+            Initial guess for the control group.
+
+            The initial guess value that is assigned to all control variables in
+            the control group unless overridden at the variable level.
```
andreas-el: Other descriptions separate this slightly;
I think I prefer having this as a separate line
```
This can be overridden at the variable level.
```
```
This group default may be overridden by individual variables.
This group default may be overridden at the variable level.
```
You can consider wording.

src/everest/config/control_config.py
```diff
@@ -49,124 +50,171 @@ class ControlConfig(BaseModel):
...
+            """
+            The perturbation type for the control group.
+
+            The `perturbation_type` keyword defines whether the perturbation
+            magnitude (`perturbation_magnitude`) should be treated as an
+            absolute value or as relative to the dynamic range of the controls.
+
+            Note: currently the dynamic range is computed with respect to all
+            controls, so defining a relative perturbation type for control types
+            with different dynamic ranges might have unintended effects.
```
andreas-el: Is this something Everest could warn when simulating? Perhaps we already do?

src/everest/config/control_config.py
```diff
@@ -49,124 +50,171 @@ class ControlConfig(BaseModel):
...
-whilst for more discrete controls larger values should give a better
-result. However, this is a balance as too large or too small
-of values also cause issues.
-
-NOTE: In most cases this should not be configured, and the default value should be used.
-        """,
+        description=dedent(
+            """
+            The magnitude of the perturbation of all control variables in the
+            control group unless overridden at the variable level.
```
andreas-el: Suggest splitting this. Notes like this usually are on separate lines, and sometimes with different style.

src/everest/config/forward_model_config.py
```diff
@@ -10,14 +11,25 @@
 
 
 class ForwardModelResult(BaseModelWithContextSupport):
-    file_name: str = Field(description="Output file produced by the forward model")
+    file_name: str = Field(
+        description=dedent(
+            """
+        The output file produced by the forward model.
+        """
```
andreas-el: Resolve indent.
Perhaps an example filename?
Not clear if everest adds suffix etc.

src/everest/config/control_config.py
verveerpj: Not sure, presumably to keep the option open to extend it with different types.

src/everest/config/simulator_config.py
```diff
@@ -41,47 +42,63 @@ def check_removed_config(queue_system: Any) -> None:
...
-     the forward models. This can for example be used in conjunction with the Eclipse
-     parallel keyword for multiple CPU simulation runs. This keyword has no effect
-     when running with the local queue.
-
-    This number is specified in Ert as NUM_CPU.""",
+        description=dedent(
+            """
+            Defines the number of CPUs when running the forward models. This can
+            for example be used in conjunction with the Eclipse parallel keyword
+            for multiple CPU simulation runs. This keyword has no effect when
```
andreas-el: Split line?

src/everest/config/simulator_config.py
```diff
@@ -41,47 +42,63 @@ def check_removed_config(queue_system: Any) -> None:
...
         default=None,
-        description="""Maximum allowed running time of a forward model. When
-        set, a job is only allowed to run for max_runtime seconds.
-        A value of 0 means unlimited runtime.
-        """,
+        description=dedent(
+            """
+            Maximum allowed running time of a forward model job.
+
+            When set, a job is only allowed to run for `max_runtime` seconds.
```
andreas-el: Perhaps only;
```
Maximum allowed running time for each individual forward model job specified in seconds.

A value of 0 means unlimited runtime.
```

src/everest/config/simulator_config.py
```diff
@@ -91,20 +108,28 @@ class SimulatorConfig(BaseModelWithContextSupport, extra="forbid"):
...
-        Defines how many times should the queue system retry a forward model.
-
-    A forward model may fail for reasons that are not due to the forward model
-    itself, like a node in the cluster crashing, network issues, etc.  Therefore, it
-    might make sense to resubmit a forward model in case it fails.
-    resumbit_limit defines the number of times we will resubmit a failing forward model.
-    If not specified, a default value of 1 will be used.""",
+        description=dedent(
+            """
+        Defines how many times the queue system may retry a forward model.
```
andreas-el: Specifies how many times a forward model may be submitted to the queue system.

src/everest/config/simulator_config.py
```diff
@@ -91,20 +108,28 @@ class SimulatorConfig(BaseModelWithContextSupport, extra="forbid"):
...
+
+        A forward model may fail for reasons that are not due to the forward
+        model itself, like a node in the cluster crashing, network issues, etc.
+        Therefore, it might make sense to resubmit a forward model in case it
+        fails. `resumbit_limit` defines the number of times we will resubmit a
+        failing forward model. If not specified, a default value of 1 will be
+        used.
```
andreas-el: A forward model may fail for external reasons, examples include node in the cluster crashing, network issues, etc.

src/everest/config/server_config.py
```diff
@@ -32,7 +33,11 @@ class ServerConfig(BaseModel):
         | None
     ) = Field(
         default=None,
-        description="Defines which queue system the everest server is submitted to",
+        description=dedent(
+            """
+            Defines which queue system the everest server is submitted to.
```
andreas-el, line 38: Should we include examples?

src/everest/config/control_config.py
verveerpj: I checked what is happening in the code, and this comment does not make sense in this context. It may refer to the way the default value for `perturbation_magnitude` is calculated. And I noticed now that the approach for calculating that default is pretty bad. To my knowledge, current practice (e.g. in the tutorials) is to always manually set this. We should probably make `perturbation_magnitude` mandatory.

src/everest/config/sampler_config.py
```diff
@@ -10,32 +11,42 @@
...
+            have a method named `A`, pick a specific backend `B` by putting `B/A` in
+            the `method` field.
+            """
+        ),
+    )
+    method: str = Field(
+        default="norm",
+        description=dedent(
+            """
+            The sampling method or distribution used by the sampler backend.
```
andreas-el, line 28: What are the alternatives?

src/everest/config/control_config.py
verveerpj: There is an issue about the `perturbation_magnitude`, (#9844) I will amend.

src/everest/config/forward_model_config.py
verveerpj: This a base class, hence any restrictions on the filename will depend on the derived class. As far as I can tell, no specific limitations on e.g. suffix for now.

src/everest/config/sampler_config.py
verveerpj: Depends on the sampler backend. Currently we only use the SciPy based backend included with `ropt`. I have copied the description of the available methods from the `ropt` manual.

src/everest/config/control_variable_config.py
```diff
@@ -18,89 +20,157 @@ class _ControlVariable(BaseModel):
...
-defined by [min, max] to the range defined by scaled_range (default [0, 1]).
-""",
+        description=dedent(
+            """
+            Enable/disable the variable.
+
+            If set to `False` the variable will not be changed during
+            optimization. Overrides the value of `enabled` in the control group.
+            """
+        ),
```
andreas-el, line 57: Sounds more like `constant` or `fixed_value` ?
I assume default is `True` ?
Will the value be overlooked (not used), or just not updated?
Split line.

verveerpj: The default is `None`, meaning it does not override the control group value.

berland:
Check off the tickboxes in the first PR post.

docs/everest/conf.py
```diff
@@ -49,6 +49,9 @@
     with_footer=False,
     collapse_long_descriptions=False,
     collapse_long_examples=False,
+    description_is_markdown=True,
+    markdown_options={"breaks": {"on_newline": False}},
```
berland, line 53: The changes in this file is probably what solves the original issue. Should perhaps be in its own commit with an instructive git commit message.

src/everest/config/environment_config.py
```diff
@@ -1,41 +1,64 @@
...
+            - `debug`: Detailed information, typically of interest only when
+               diagnosing problems.
+            - `info`: Confirmation that things are working as expected.
+            - `warning`: An indication that something unexpected happened, or
+               indicative of some problem in the near future (e.g. `disk space
+               low`). The software is still working as expected.
+            - `error`: Due to a more serious problem, the software has not been
+               able to perform some function.
+            - `critical`: A serious error, indicating that the program itself
+               may be unable to continue running.
```
andreas-el: This is definitely `will not be able to`

src/everest/config/input_constraint_config.py
```diff
@@ -7,42 +9,75 @@ class InputConstraintConfig(BaseModel, extra="forbid"):
...
+        description=dedent(
+            """
+            Scaling of input constraints.
+
+            `scale` is a normalization factor which can be used to scale the
+            input constraint. The bounds or target, and the weights will be
+            scaled with this number.
+
+            This option may not be set if `auto_scale` is set in the
+            `optimization` section.
```
andreas-el: This option will be disabled, or overridden perhaps?

src/everest/config/objective_function_config.py
```diff
@@ -12,43 +13,65 @@
...
+            differently will change their relative weights.
 
-    The value of each objective function is divided by the related scaling value.
-    When optimizing with respect to multiple objective functions, it is important
-    that the scaling is set so that all the scaled objectives have the same order
-    of magnitude. Ultimately, the scaled objectives are used in computing
-    the weighted sum that Everest tries to optimize.
-    """,
+            This option may not be set if `auto_scale` is set in the
+            `optimization` section.
```
andreas-el: This gives a feeling that the config will fail validation if this is set. Perhaps overridden or disabled?

src/everest/config/output_constraint_config.py
```diff
@@ -1,53 +1,74 @@
...
+            scaled with this number. For example, if the upper bound is 0.5 and
+            the scaling is 10, then the function evaluation value will be
+            divided by 10 and bounded from above by 0.05.
 
-Both the upper_bound and the function evaluation value will be scaled with this number.
-That means that if, e.g., the upper_bound is 0.5 and the scaling is 10, then the
-function evaluation value will be divided by 10 and bounded from above by 0.05.
-""",
+            This option may not be set if `auto_scale` is set in the
+            `optimization` section.
```
andreas-el: Possible reword.

docs/everest/conf.py
verveerpj: Good point, just before merging I will make sure there are sensible commits.

andreas-el (approved):
You did well here! Nicely done! 🧹 ✨

PR #11845, opened 2025-09-17
Make sure everserver logs reach attached azure logger

### LINKED ISSUE #11626: Everserver (as _storage_main.py) logs to not reach Azure log instance
Everserver has its custom log configuration where logs end up on disk. If it used the Ert/Everests plugin manager, it can also log to other locations (as in an Azure log instance)

StephanDeHoop:
**Issue**
Resolves #11626

**Approach**
~~(TODO: need to rebase this once the PR moving to `ErtPluginManager` is in)~~

Current logs from the `everserver` will not reach the Azure logs, this is most likely due to:
logs from children not being propagated to the root + `everserver` logs being filtered out due to the hook implementation in `fmu_site_configuration`. This PR attempts to fix that (i.e., changing the name of the logger to start with `everest`, alternatively we could also add `everserver` to the allowed modules in the hook implementation). 

A lot of logs might now end up in the Azure logs, not sure if we want all of them + a bit scared if I somehow introduce duplicate logs (due to the propagate True, I don't understand all this logging business at a low enough level I am afraid) or bloat due to third party loggers that are not reaching the root logger (due to `"disable_existing_loggers": False`). 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/detached/everserver.py
+++ b/src/everest/detached/everserver.py
@@ -61,11 +61,12 @@ def _configure_loggers(
             "class": "logging.FileHandler",
             "formatter": "default",
             "level": log_level,
-            "filename": path,
+            "filename": str(path),
         }
 
     logging_config = {
         "version": 1,
+        "disable_existing_loggers": False,
         "handlers": {
             "endpoint_log": make_handler_config(
                 detached_dir / "endpoint.log", logging_level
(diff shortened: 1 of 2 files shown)


berland:
Can you tick of the relevant boxes in the PR?

frode-aarstad:
> > Is it possible to make a test that tests that goes through the logger and checks that the correct loggers are active?
> 
> Do you mean an integration test all the way to Azure or something simpler?

Something simpler. Perhaps checking the logging object and make sure the expected handlers are there

PR #11852, opened 2025-09-18
Add logging for obs types and keywords

SAKavli:
This is a manual backport of https://github.com/equinor/ert/commit/565430f946c514fa240e6d1714df9bea8fd078c5, sanity checks are welcome.


src/ert/config/parsing/observations_parser.py
```diff
@@ -170,6 +190,8 @@ def _parse_content_list(
                 end_column=e.column + 1,
             )
         ) from e
+    else:
+        return parsed_obs
```
SAKavli, line 194: Ruff did not let me return inside the try statement https://docs.astral.sh/ruff/rules/try-consider-else/

eivindjahren: That seems fine

Issue #11875, opened 2025-09-19
Make new experiment for manual update/ESMDA restart


yngve-sk:
To support multiple experiments using the same ensemble as prior

PR #11883, opened 2025-09-22
Use write transaction for writing experiment index.json

yngve-sk:
**Issue**
Resolves https://github.com/equinor/ert/issues/11875


frode-aarstad (approved):
Looks good. Nitpick: lowercase first letter in last commit

src/ert/storage/local_experiment.py
```diff
@@ -142,8 +142,9 @@ def create(
         if name is None:
             name = datetime.today().isoformat()
 
-        (path / "index.json").write_text(
-            _Index(id=uuid, name=name).model_dump_json(indent=2)
+        storage._write_transaction(
+            path / "index.json",
```
andreas-el, line 150: Could the file be a constant?

src/ert/storage/migration/to13.py
```diff
@@ -7,7 +7,49 @@
...
+            ensemble_to_experiment[index_["id"]] = index_["experiment_id"]
+
+    experiment_to_ensemble = {
+        exp_id: [
+            ens_id
+            for ens_id, ens_exp_id in ensemble_to_experiment.items()
+            if ens_exp_id == exp_id
+        ]
+        for exp_id in set(ensemble_to_experiment.values())
+    }
```
andreas-el: Can this be refactored slightly? We don't need `ensemble_to_experiment` if we populate `experiment_to_ensemble` during the loop?
```
    experiment_to_ensemble = {}
    for ensemble in path.glob("ensembles/*"):
        with open(ensemble / "index.json", encoding="utf-8") as fin:
            index_ = json.load(fin)
            ensemble_id = index_["id"]
            experiment_id = index_["experiment_id”]
           if experiment_id not in experiment_to_ensemble:
               experiment_to_ensemble[experiment_id] = []
           experiment_to_ensemble[experiment_id].append(ensemble_id)
```

src/ert/storage/migration/to13.py
```diff
@@ -7,7 +7,49 @@
...
+        with open(experiment_path / "index.json", encoding="utf-8") as fin:
+            old_json = json.load(fin)
+            exp_id = old_json["id"]
+
+        if exp_id not in experiment_to_ensemble:
+            with open(experiment_path / "index.json", "w", encoding="utf-8") as fout:
+                json.dump(
+                    old_json | {"ensembles": []},
+                    fout,
+                    indent=4,
```
andreas-el: Why do we alter between the indentation set in the json-files?

src/ert/storage/local_experiment.py
```diff
@@ -229,14 +235,24 @@ def create_ensemble(
...
             ensemble_size=ensemble_size,
             iteration=iteration,
             name=name,
             prior_ensemble=prior_ensemble,
         )
 
+        self.add_ensemble(ensemble.id)
+        return ensemble
+
+    def add_ensemble(self, ensemble_id: UUID) -> None:
```
andreas-el: Slightly misleading function name? It adds and most importantly writes?

src/ert/storage/local_experiment.py
yngve-sk: Hmm it could be, prefer to defer discussion for that to another PR/issue. Not directly related to this change, beyond wrapping it in _write_transaction which is in a separate commit.

src/ert/storage/local_experiment.py
yngve-sk: Hmm, nice catch. I think I must have had another idea for this fn, see it is only used one place, so will just inline the logic to `LocalExperiment.create_ensemble` to avoid this confusion.

src/ert/storage/migration/to13.py
yngve-sk: Oopsie, set it to 2

src/ert/storage/migration/to13.py
yngve-sk: Done

andreas-el (approved):
Well done! 🎖️

PR #11891, opened 2025-09-23
Ignore site config queue options if user queue system differs

yngve-sk:
Resolves bleeding fails

### DIFF
--- a/tests/ert/unit_tests/config/test_queue_config.py
+++ b/tests/ert/unit_tests/config/test_queue_config.py
@@ -18,6 +18,7 @@ from ert.config.queue_config import (
     SlurmQueueOptions,
     TorqueQueueOptions,
 )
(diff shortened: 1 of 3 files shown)


tests/ert/unit_tests/config/test_queue_config.py
```diff
@@ -548,3 +549,11 @@ def test_that_job_script_from_queue_options_takes_precedence_over_global(
         "QUEUE_OPTION LSF JOB_SCRIPT fm_dispatch_lsf.py\n"
     )
     assert config.queue_config.queue_options.job_script == "fm_dispatch_lsf.py"
+
+
+def test_that_site_queue_options_are_ignored_with_differing_usr_queue_system_arg():
```
berland: usr -> user (also in commit message)

PR #11912, opened 2025-09-25
Deprecate the --new-run argument for everest

### LINKED ISSUE #11910: Deprecate --new-run option
We now always use rolling directories when performing a everest run so this options has no effect

frode-aarstad:
**Issue**
Resolves #11910 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/bin/utils.py
+++ b/src/everest/bin/utils.py
@@ -425,17 +425,10 @@ def report_on_previous_run(
     server_state = everserver_status(everserver_status_path)
     if server_state["status"] == ExperimentState.failed:
         error_msg = server_state["message"]
-        print(
-            f"Optimization run failed, with error: {error_msg}\n"
-            "To re-run optimization case use command:\n"
-            f"`  everest run --new-run {config_file}`\n"
-        )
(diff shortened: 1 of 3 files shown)


src/everest/bin/utils.py
```diff
@@ -425,18 +425,9 @@ def report_on_previous_run(
...
-        )
+        print(f"Optimization run failed, with error: {error_msg}\n")
     else:
-        print(
-            f"Optimization completed.\n"
-            "\nTo re-run the optimization use command:\n"
-            f"  `everest run --new-run {config_file}`\n"
-            f"Results are stored in {optimization_output_dir}"
-        )
+        print("Optimization completed.\n\nTo re-run the optimization use command:\n")
```
verveerpj: I think you can remove the `\nTo re-run the optimization use command:\n` part? I don't think we need to suggest that users are able to run it again...

Do we still need the `f"Results are stored in {optimization_output_dir}"` message?

frode-aarstad: Forgot the re-run part. Will add the message where results are stored back

PR #11915, opened 2025-09-25
Log correct number of failed realizations when restarting simulations

### LINKED ISSUE #11907: Rerunning failed realizations cause calculation for logged failures to be wrong
Image

Adding this to the poly example:

```
diff --git a/test-data/ert/poly_example/poly_eval.py b/test-data/ert/poly_example/poly_eval.py
index 33c67d8a11..f09ff3de17 100755
--- a/test-data/ert/poly_example/poly_eval.py
+++ b/test-data/ert/poly_example/poly_eval.py
@@ -12,7 +12,12 @@ def _evaluate(coeffs, x):


 if __name__ == "__main__":
+    from random import random
+
+    if random() > 0.5:
+        raise ValueError("Random error occurred")
+
```

Then I ran ensemble_experiment, and re-ran twice.
During the first run 49 realizations failed, during the second run, likely 31 of those 49 succeeded (but only 18 failed) and so on.

andreas-el:
Resolves #11907

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/run_models/run_model.py
+++ b/src/ert/run_models/run_model.py
@@ -800,7 +800,10 @@ class RunModel(BaseModelWithContextSupport, ABC):
         num_successful_realizations = len(successful_realizations)
         self.validate_successful_realizations_count()
         logger.info(f"Experiment ran on QUEUESYSTEM: {self.queue_config.queue_system}")
-        logger.info(f"Experiment ran with number of realizations: {self.ensemble_size}")
+        logger.info(
+            f"Experiment ran with number of realizations: {len(starting_realizations)} "
+            f"out of a total {self.ensemble_size}"
+        )
         logger.info(
             f"Experiment run ended with number of realizations succeeding: "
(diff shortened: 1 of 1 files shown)


src/ert/run_models/run_model.py
```diff
@@ -807,7 +807,7 @@ def _evaluate_and_postprocess(
         )
         logger.info(
             f"Experiment run ended with number of realizations failing: "
-            f"{self.ensemble_size - num_successful_realizations}"
+            f"{len(starting_realizations) - num_successful_realizations}"
```
berland, line 813: The similar log message on line 803 is perhaps equally wrong? Maybe the log (on line 803) should state both the size of the ensemble, and the size of the starting realizations?

andreas-el: Have updated a new proposal

berland (approved):

PR #11933, opened 2025-09-26
Allow inline data in install_data

### LINKED ISSUE #11895: Update the install_data keyword to accept inline data in the Everest configuration file
This will add a `data` sub-field to the `install_data` keywords, which is mutually exclusive with the `source` and `link` fields. If given the content will be dumped to the named `target` file. The output format may be json or yaml and will be derived from the file name.

The initial purpose of this is to replace the `wells` section in the Everest configuration file. This would be achieved by moving the content of `wells` to an entry in `install_data` and specifying `wells.json` as the output.

Secondary purpose is to add a simple way to generate small data files that are shared by all the forward models.

verveerpj:
**Issue**
Resolves #11895

**Approach**
Extends the `install_data` section of the Everest configuration with a `data` fields that is mutually exclusive with the `source` field. If present, the content of that field is used instead of a source file to generate the target file. Supported outputs are json and yaml. This feature is intended to replace the `wells` field in the configuration, which is to be deprecated. Since it is a generic mechanism, it will also be documented as a new feature.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/config/install_data_config.py
+++ b/src/everest/config/install_data_config.py
@@ -1,25 +1,30 @@
+from __future__ import annotations
+
+import json
 from pathlib import Path
 from textwrap import dedent
+from typing import Any
 
-from pydantic import BaseModel, Field, field_validator, model_validator
+from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
 
 
-class InstallDataConfig(BaseModel, extra="forbid"):
-    source: str = Field(
+class InstallDataConfig(BaseModel):
+    source: str | None = Field(
+        default=None,
         description=dedent(
             """
             Path to file or directory that needs to be copied or linked into the
             directory where the forward model executes.
             """
-        )
-    )  # existing path
+        ),
+    )
     target: str = Field(
         description=dedent(
             """
             Relative destination path to copy or link the given source to.
             """
         )
-    )  # path
+    )
     link: bool = Field(
         default=False,
         description=dedent(
@@ -28,6 +33,25 @@ class InstallDataConfig(BaseModel, extra="forbid"):
             if not set the source will be copied at the given target.
             """
         ),
+    )
+    data: dict[str, Any] | None = Field(
+        default=None,
+        description=dedent(
+            """
+            Data to be exported to the target file.
+
+            If provided, the data is exported to a JSON file. If `target` has no
+            `.json` extension, it is added.
+
+            The `data` and `source` fields are mutually exclusive.
+
+            If `data` is provided, `link` will be ignored.
+            """
+        ),
+    )
+
+    model_config = ConfigDict(
+        extra="forbid",
     )
 
     @field_validator("link", mode="before")
@@ -40,7 +64,29 @@ class InstallDataConfig(BaseModel, extra="forbid"):
         return link
 
     @model_validator(mode="after")
-    def validate_target(self) -> "InstallDataConfig":
-        if self.target in {".", "./"}:
+    def validate_target(self) -> InstallDataConfig:
+        if self.data is not None:
+            if self.source is not None:
+                raise ValueError("The data and source options are mutually exclusive.")
+            if self.target.strip() in {"", ".", "./"}:
+                raise ValueError("A target name must be provided with data.")
+
+            ext = Path(self.target).suffix.lower()
+            if ext not in {".json", ""}:
+                raise ValueError(f"Invalid target extension {ext} (.json expected).")
+
+            return self
+
+        if self.source is None:
+            raise ValueError("Either source or data must be provided.")
+
+        if self.target.strip() in {".", "./"}:
             self.target = Path(self.source).name
         return self
+
+    def inline_data_as_str(self) -> tuple[str, str]:
+        assert self.data is not None
+        return (
+            str(Path(self.target).with_suffix(".json")),
+            json.dumps(self.data, indent=2),
+        )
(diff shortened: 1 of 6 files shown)


berland:
Can you prepare everest-models to not depend on wells.json before we merge this?

src/everest/config/install_data_config.py
```diff
@@ -29,6 +35,31 @@ class InstallDataConfig(BaseModel, extra="forbid"):
...
+            """
+            Data to be exported to the target file.
+
+            If provided, the data is exported to `target`, according to its file
+            extension:
+
+            - `.json`: Export to json.
+            - `.yaml` or `.yml`: Export to yaml.
+
+            If `target` has no extension, json format is used, and a `.json`
```
berland: Do we want this automatic addition of json extension? That means it will be impossible to have full control of the filename.

src/everest/config/install_data_config.py
```diff
@@ -40,7 +71,36 @@ def validate_link_type(cls, link: bool | None) -> bool | None:
...
     @model_validator(mode="after")
-    def validate_target(self) -> "InstallDataConfig":
-        if self.target in {".", "./"}:
+    def validate_target(self) -> InstallDataConfig:
+        if self.data is not None:
+            if self.source is not None:
+                raise ValueError("The data and source options are mutually exclusive.")
+            if self.target.strip() in {"", ".", "./"}:
+                raise ValueError("A target name must be provided with data.")
+
```
berland, line 73: Is it an idea to test that the file location is writable?

src/everest/config/validation_utils.py
```diff
@@ -50,21 +50,35 @@ def __init__(
...
-            if "<GEO_ID>" not in data.source:
-                self._set_symlink(data.source, data.target, None)
-
         os.chdir(self._temp_dir.name)
+        for item in self._install_data:
+            try:
+                Path(item.target).resolve().relative_to(Path.cwd())
+            except ValueError as err:
+                raise ValueError(
+                    f"Target location outside of runpath {item.target} not allowed!"
```
berland: Can this be more precise, it reads like runpath is the same as item.target?

berland: Maybe just move `{item.target}` to between 'location' and 'outside'?

berland:
We should have codecov working on this PR to check that everything has been tested.

berland:
I found the codecov report at https://app.codecov.io/github/equinor/ert/commit/39e90e7e16deda107e15e356510e618c210eb819 and there are a few uncovered lines, can you add tests for those?

tests/everest/test_config_validation.py
```diff
@@ -506,6 +509,18 @@ def test_that_install_data_allows_runpath_root_as_target(
...
+            {"link": True, "target": "bar.json"},
+            "Either source or data must be provided",
+        ),
+        (
+            {"source": "baz/", "data": {"foo": 1}, "link": True, "target": "bar.json"},
+            "The data and source options are mutually exclusive",
+        ),
+        (
+            {"data": {"foo": 1}, "link": True, "target": "bar.txt"},
+            "Invalid target extension",
```
berland: (hmm, codecov says this is never hit, something does not work somewhere..)

tests/everest/test_config_validation.py
```diff
@@ -561,6 +576,38 @@ def test_that_install_data_source_exists(change_to_tmpdir):
     )
 
 
+def test_that_install_data_with_inline_data_generates_a_file(
+    copy_math_func_test_data_to_tmp,
+):
+    config = EverestConfig.load_file("config_minimal.yml")
+    config_dict = {
+        **config.model_dump(exclude_none=True),
+        "optimization": {
```
berland, line 589: (why do we need to set/overwrite `optimization` in this test?)

berland: ah, maybe you are limiting the time taken to run the test.

src/everest/config/install_data_config.py
verveerpj: In this way the user can enter the equivalent of the current `wells` section. They just copy the content of wells section to the `install_data` field, and set the `target` field to `wells`. No need to know the implementation detail that this needs to be a `json` file.

This feature is also meant for quick inclusion of data that is representable as a dict and serializable to json, for consumption by the forward models. JSON seems a reasonable default, hence adding and implicit json extension makes sense. We don't need full control over the filename, if other file types are needed, the user should install a manually written file with `install_data `or `install_templates`.

However, I see the point that it may be confusing that it could be something else than JSON, i.e. YAML., We could either reason that they are equivalent so yaml files are fine, or we could avoid potential confusion by only allowing JSON. What do you think?

src/everest/config/install_data_config.py
verveerpj: The file location is the run path, so it does not exist yet. The target file will be written by the same copy forward model as used by `install_data`, it presumable uses the necessary safeguards.

tests/everest/test_config_validation.py
verveerpj: `pytest` runs that test without issue, so maybe codecov has issues with parametrized tests?

src/everest/config/install_data_config.py
verveerpj: On second thought, I think allowing other formats like  yaml is premature, removing yaml support.

berland: fair enough. Keep the json extension.

tests/everest/test_config_validation.py
verveerpj: Seems to be ok now.

PR #11937, opened 2025-09-29
Increase memory limit

yngve-sk:
https://github.com/equinor/komodo-releases/actions/runs/18082964074/job/51450362321


berland:
Kudos for the commit message!

PR #11945, opened 2025-09-29
Add deprecation message for FORWARD_MODEL DESIGN2PARAMS usage

### LINKED ISSUE #11935: A suggester should pop up that `DESIGN_MATRIX` can be used when `DESIGN2PARAMS` is used in config
We would like users to switch to DESIGN_MATRIX and thus this would be of a great help.

flikka:
**Issue**
Resolves #11935 

**Approach**
Adding FORWARD_MODEL DESIGN2PARAMS to deprecation list

image

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/parsing/config_schema_deprecations.py
+++ b/src/ert/config/parsing/config_schema_deprecations.py
@@ -1,7 +1,28 @@
 from functools import partial
-from typing import cast
+from typing import Any, cast
 
 from .deprecation_info import DeprecationInfo
+
+
+def make_forward_model_design2params_deprecation_message(line: list[Any]) -> str:
+    design_matrix_file = "<name_of_design_matrix_file.xlsx>"
+    design_sheet = "<name_of_design_sheet>"
+    defaults_sheet = "<name_of_defaults_sheet>"
+    arguments_dict = dict(line[1])
+
+    design_matrix_file = arguments_dict.get("<xls_filename>", design_matrix_file)
+    design_sheet = arguments_dict.get("<designsheet>", design_sheet)
+    defaults_sheet = arguments_dict.get("<defaultssheet>", defaults_sheet)
+
+    msg = (
+        f"FORWARD_MODEL DESIGN2PARAMS will be replaced with DESIGN_MATRIX. "
+        "Please change configuration line with FORWARD_MODEL DESIGN2PARAMS,\n"
+        "To this format:\n "
+        f"'DESIGN_MATRIX {design_matrix_file} DESIGN_SHEET:{design_sheet} "
+        f"DEFAULT_SHEET:{defaults_sheet}'"
+    )
+    return msg
+
 
 REPLACE_WITH_GEN_KW = [
     "RELPERM",
@@ -203,5 +224,10 @@ deprecated_keywords_list = [
         "configuration path if provided."
         "It has been replaced with the path to the directory containing the "
         "configuration file for the experiment.",
+    ),
+    DeprecationInfo(
+        keyword="FORWARD_MODEL",
+        message=make_forward_model_design2params_deprecation_message,
+        check=lambda line: line[0] == "DESIGN2PARAMS",
     ),
 ]
--- a/tests/ert/unit_tests/config/parsing/test_config_schema_deprecations.py
+++ b/tests/ert/unit_tests/config/parsing/test_config_schema_deprecations.py
@@ -1,8 +1,10 @@
 import warnings
+from unittest.mock import patch
 
 import pytest
 
 from ert.config import ConfigWarning, ErtConfig
+from ert.config.forward_model_step import ForwardModelStep
 from ert.config.parsing.config_schema_deprecations import (
     JUST_REMOVE_KEYWORDS,
     REPLACE_WITH_GEN_KW,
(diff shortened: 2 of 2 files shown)


src/ert/config/parsing/config_schema_deprecations.py
```diff
@@ -204,4 +204,10 @@
         "It has been replaced with the path to the directory containing the "
         "configuration file for the experiment.",
     ),
+    DeprecationInfo(
```
xjules, line 228: Not sure if we want to remove completely `DESIGN2PARAMS`. Maybe @oyvindeide can comment on this?
When it comes to the message itself I thought more in lines:
```
"FORWARD_MODEL DESIGN2PARAMS is still supported, but we recommend using DESIGN_MATRIX instead. 
Please consider updating your configuration; see the ERT manual for details."
```

berland: If there are scenarios where DESIGN2PARAMS is really needed in a user config, can the user in some way mute the warning?

flikka: > Not sure if we want to remove completely `DESIGN2PARAMS`. Maybe @oyvindeide can comment on this? When it comes to the message itself I thought more in lines:
> 
> ```
> "FORWARD_MODEL DESIGN2PARAMS is still supported, but we recommend using DESIGN_MATRIX instead. 
> Please consider updating your configuration; see the ERT manual for details."
> ```

Ah, I assumed we made DESIGN_MATRIX to replace DESIGN2PARAMS. Could we check with the user-representatives etc? As @berland  points out, if these both are equally supported I guess we need a different way of informing the user. And to be honest, if it is a fully worthy feature, I think it is more appropriate to put information in the manual or similar? Speaking of which, seems like we don't write anything about DESIGN2PARAMS in the manual? https://ert.readthedocs.io/en/latest/search.html?q=DESIGN2PARAMS&check_keywords=yes&area=default

flikka: @oyvindeide confirms that we want to remove the DESIGN2PARAMS forward model, starting with a deprecation.

berland: A rogue user could have some rogue code generating the xls file as part of the forward model, and then run DESIGN2PARAMS on it, but it would surprise me if we actually have anyone doing that.

xjules: @flikka additionally would it be possible to print a line that uses `DESIGN_MATRIX` representing equivalent to `DESIGN2PARAMS` usage (ie. with DESIGN_SHEET, DEFAULT_SHEET and the filename)?

xjules: > A rogue user could have some rogue code generating the xls file as part of the forward model, and then run DESIGN2PARAMS on it, but it would surprise me if we actually have anyone doing that.

I think there are "super" users actually doing what you said, but I don't see a way to distinguish between those cases.

flikka: > @flikka additionally would it be possible to print a line that uses `DESIGN_MATRIX` representing equivalent to `DESIGN2PARAMS` usage (ie. with DESIGN_SHEET, DEFAULT_SHEET and the filename)?

@xjules  You mean "To use the design matrix fuctionality, add a line 'DESIGN_MATRIX <name_of_design_matrix>.xlsx DESIGN_SHEET:DesignSheet DEFAULT_SHEET:DefaultSheet'" 

Or did I misunderstand? Isnt this a little bit to much info, and now we have two places to maintain the "manual"? But perhaps I misunderstood!

xjules: > You mean "To use the design matrix fuctionality, add a line 'DESIGN_MATRIX <name_of_design_matrix>.xlsx DESIGN_SHEET:DesignSheet DEFAULT_SHEET:DefaultSheet'"

yes, exactly!

flikka:
So maybe like this?
image

flikka:
> > So maybe like this?
> 
> Looks very good! I have two small suggestions:
> 
>     * Adding this sentence somewhere:
>       Please see https://ert.readthedocs.io/en/latest/reference/configuration/keywords.html#design-matrix for more details on DESIGN_MATRIX
> 
>     * "have been replaced" -> "will be replaced"

Adding a link (with link) in the suggestor requires some changes to the Suggestor and Depcrecations code, is it important enough? I have changed it to "will be replaced", and removed the "example", I think it is good enough to point it out, and refer to the manual. Gets to crowded and messy if we write "the manual" in these boxes, and one more place to maintain it. This is my suggestion:

image

xjules:
> > > So maybe like this?
> > 
> > 
> > Looks very good! I have two small suggestions:
> > ```
> > * Adding this sentence somewhere:
> >   Please see https://ert.readthedocs.io/en/latest/reference/configuration/keywords.html#design-matrix for more details on DESIGN_MATRIX
> > 
> > * "have been replaced" -> "will be replaced"
> > ```
> 
> Adding a link (with link) in the suggestor requires some changes to the Suggestor and Depcrecations code, is it important enough? I have changed it to "will be replaced", and removed the "example", I think it is good enough to point it out, and refer to the manual. Gets to crowded and messy if we write "the manual" in these boxes, and one more place to maintain it. This is my suggestion:
> 
> image

 - There is a type "will been" ...
 - Adding link to docs was suggested by @eivindjahren . We do use it https://github.com/equinor/ert/blob/c8742981dc59773893a361dd0a8d82a89c37a82f/src/ert/config/parsing/config_schema_deprecations.py#L79

src/ert/config/parsing/config_schema_deprecations.py
```diff
@@ -204,4 +204,11 @@
...
+        keyword="FORWARD_MODEL",
+        message="FORWARD_MODEL DESIGN2PARAMS will been replaced with DESIGN_MATRIX. "
```
xjules: As discussed let's switch to the first version.

flikka: Since the config is already parsed, if we want something more than just a generic message, I think it is feasible to make a special thing that picks out arguments, and puts it into the "new" wine. see updated PR

src/ert/config/parsing/config_schema_deprecations.py
```diff
@@ -1,8 +1,29 @@
...
+    design_sheet = "<name_of_design_sheet>"
+    defaults_sheet = "<name_of_defaults_sheet>"
+    arguments_dict = dict(line[1])
+
+    design_matrix_file = arguments_dict.get("<xls_filename>", design_matrix_file)
+    design_sheet = arguments_dict.get("<designsheet>", design_sheet)
+    defaults_sheet = arguments_dict.get("<defaultssheet>", defaults_sheet)
+
+    msg = (
+        f"FORWARD_MODEL DESIGN2PARAMS will been replaced with DESIGN_MATRIX. "
```
xjules: -> "will be"

tests/ert/unit_tests/config/parsing/test_config_schema_deprecations.py
```diff
@@ -201,3 +203,57 @@ def test_that_a_deprecation_message_is_shown_for_use_of_the_job_prefix_queue_opt
...
+    )
+
+    with (
+        patch(
+            "ert.config.ert_config.installed_forward_model_steps_from_dict",
+            return_value={"DESIGN2PARAMS": mock_design2params_step},
+        ),
+        pytest.warns(
+            ConfigWarning,
+            match="FORWARD_MODEL DESIGN2PARAMS will been replaced with DESIGN_MATRIX. "
```
xjules: -> "will be"

Issue #11355, opened 2025-07-11
Remove redundant everest status file


yngve-sk:
Everest status file given by path:
```
    @staticmethod
    def get_everserver_status_path(output_dir: str) -> str:
        """Returns path to the everest server status file"""
        return os.path.join(ServerConfig.get_session_dir(output_dir), SERVER_STATUS)
```

is currently read from multiple places. Reading this should not be necessary, as the experiment server holds the status, so it should be queried for the status instead. 

We assume that it is completely unnecessary to have this information stored in a readily available file. The only thing that should be readable by clients of everest (everviz, cli monitor, etc) should be what is in **storage** . 

(If it turns out we do need a status file, it should be written to storage once, on the experiment server side, most likely from the context of `.run_experiment`)

If doing this all at once turns out too big, it is possible to just spawn a sub issue here and remove usages one by one.

frode-aarstad:
I think we would need something in storage indicating the status of the experiment. Perhaps in experiment/metadata.json. So we can query the status when there are no everserver running

PR #11963, opened 2025-09-30
Stop reading status from file in everest_script.py and monitor_script.py

frode-aarstad:
**Issue**
Resolves partly #11355

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/bin/utils.py
+++ b/src/everest/bin/utils.py
@@ -23,6 +23,7 @@ from ert.ensemble_evaluator import (
     FullSnapshotEvent,
     SnapshotUpdateEvent,
 )
+from ert.ensemble_evaluator.event import EndEvent
 from ert.logging import LOGGING_CONFIG
 from ert.plugins.plugin_manager import ErtPluginManager
 from ert.services import StorageService
@@ -214,6 +215,8 @@ class _DetachedMonitor:
                     self._clear_lines = 0
             if SIM_PROGRESS_ID in status:
                 match status[SIM_PROGRESS_ID]:
+                    case EndEvent(msg=msg):
+                        return
                     case FullSnapshotEvent(snapshot=snapshot, iteration=batch):
                         if snapshot is not None:
                             self._snapshots[batch] = snapshot
@@ -404,7 +407,9 @@ class _DetachedMonitor:
             print(colorama.Cursor.UP(), end=colorama.ansi.clear_line())
 
 
-def run_detached_monitor(server_context: tuple[str, str, tuple[str, str]]) -> None:
+def run_detached_monitor(
+    server_context: tuple[str, str, tuple[str, str]],
+) -> None:
     monitor = _DetachedMonitor()
     start_monitor(server_context, callback=monitor.update)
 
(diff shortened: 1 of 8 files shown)


berland:
> Yes. Added a link to the "master" issue

Add it to the top message in this PR, that is where I always look for it. Also check the tick-boxes.

src/everest/bin/utils.py
```diff
@@ -216,6 +217,14 @@ def update(self, status: dict[str, Any]) -> None:
...
             if SIM_PROGRESS_ID in status:
                 match status[SIM_PROGRESS_ID]:
+                    case EndEvent(msg=msg) as end_event:
+                        if end_event.failed:
+                            raise SystemError(
+                                f"{Fore.RED}FAILED{Fore.RESET}"
+                                f" Everest run failed with: {msg}"
+                            )
+                        else:
+                            raise SystemExit(f"Everest run finished with: {msg}")
```
berland: I don't think we should use SystemExit for this. When it is called with something that evaluates to not-null, it implies that something bad has happened, which is not the case here.

Additionally, there should probably be an Everest specific exception (or reuse a similar Ert exception) when the run fails.

berland: I think this should be rewritten to not use exceptions (there is also a risk of logging noise if the exceptions bubble through spans)

frode-aarstad: Will update this when #11976 has been merged

frode-aarstad:
> Btw, should we also remove this code:
> 
> https://github.com/equinor/ert/blob/31d1e547f0de33e590fcd79037f4c00deac4cbb5/src/everest/bin/monitor_script.py#L84-L88
> ?
> 
> Since the `_DetachedMonitor.update()` is taking care of it with your changes (minus the `SystemExit`), or am I misunderstanding how this works?

Yes. This can also be removed. Thanks for the reminder

src/everest/detached/client.py
```diff
@@ -261,8 +258,9 @@ def start_monitor(
                     logger.error("Error when processing event %s", exc_info=e)
 
                 time.sleep(polling_interval)
+
     except Exception:
-        logger.debug(traceback.format_exc())
+        logger.exception(traceback.format_exc())
```
berland, line 263: looks unrelated to the git commit message (?) Upgrading to exception might be sensible, but might also trigger some extra actions in the logging system. Would just like to read some motivation for the change.

src/everest/bin/utils.py
```diff
@@ -214,6 +215,17 @@ def update(self, status: dict[str, Any]) -> None:
                     self._clear_lines = 0
             if SIM_PROGRESS_ID in status:
                 match status[SIM_PROGRESS_ID]:
+                    case EndEvent(msg=msg) as end_event:
+                        if end_event.failed:
+                            print(
+                                f"Everest run {Fore.RED}failed{Fore.RESET} with: {msg}"
```
berland: Is this replacing `print(end_event.msg)` below in client.py? Or maybe something else in `utils.py` where there are colours in use.

src/everest/detached/client.py
frode-aarstad: It is unrelated. Exceptions should be logged as exceptions

src/everest/bin/utils.py
frode-aarstad: The colors are added in this commit they were not there before

PR #11973, opened 2025-10-01
Fix OpenPBSDriver.kill() getting malformed input

jonathan-eq:
**Approach**
This commit fixes the issue where we get `qdel: illegally formed job identifier: 'job_1 job_2'` when killing jobs. This is fixed by not joining the list of job ids into a single argument

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/scheduler/test_generic_driver.py
+++ b/tests/ert/unit_tests/scheduler/test_generic_driver.py
@@ -159,28 +159,45 @@ async def test_repeated_submit_same_iens(driver: Driver, tmp_path, monkeypatch):
 @pytest.mark.flaky(reruns=5)
 async def test_kill_actually_kills(driver: Driver, tmp_path, pytestconfig, monkeypatch):
     monkeypatch.chdir(tmp_path)
-    finished = False
+    realizations = {0, 1}
+    iens_status = {
+        realization: {"finished": False, "started": False}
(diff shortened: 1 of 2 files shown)


tests/ert/unit_tests/scheduler/test_generic_driver.py
```diff
@@ -159,28 +159,45 @@ async def test_repeated_submit_same_iens(driver: Driver, tmp_path, monkeypatch):
...
+        realization: {"finished": False, "started": False}
+        for realization in realizations
+    }
+
     driver._poll_period = 0.01
 
-    async def kill_job_once_started(iens):
-        nonlocal driver
-        await driver.kill(iens)
+    async def kill_jobs_once_all_started(realizations):
```
berland, line 170: This callback is called for each of the realizations when they are started AFAIK, so you are not guaranteed that both are started when you get here. Check the definition of `poll() ` in `tests/ert/utils.py`

jonathan-eq: Yes, I keep tabs on what jobs have started, and if it is the last one calling this, it will also calls driver.kill().

berland: ah, I misread your code 🤦🏻

PR #11976, opened 2025-10-01
Write experiment status to experiment metadata

frode-aarstad:
**Issue**
Resolves #my_issue

**Approach**

      --exec 'just rapid-tests'`)

## When applicable


frode-aarstad:
> Hm, ideally the experiment status will come through emitted events, not sure if it is a good idea to explicitly store it in storage, but if we need it, it should be there only temporarily until we have a way of reacting to events from where we read it. In ERT we don't store it, we just handle it reactively through runmodel events.

Then we need to store the events because its needed when you start the monitor and there is no everest running

yngve-sk:
> > Hm, ideally the experiment status will come through emitted events, not sure if it is a good idea to explicitly store it in storage, but if we need it, it should be there only temporarily until we have a way of reacting to events from where we read it. In ERT we don't store it, we just handle it reactively through runmodel events.
> 
> Then we need to store the events because its needed when you start the monitor and there is no everest running

Hm, the websocket endpoint with events will replay all events, but not sure, looking at it if there is a direct indicator of the status. It comes from everserver, and is assigned in a context outside the runmodel itself, it could be exposed through an endpoint or via the runmodel, not sure exactly what would be best here. Agree that storage is the closest short-term solution, any opinions @oyvindeide ?

src/ert/dark_storage/endpoints/experiment_server.py
```diff
@@ -191,9 +191,7 @@ async def start_experiment(
         runner = ExperimentRunner(config)
         try:
             background_tasks.add_task(runner.run)
-            shared_data.status = ExperimentStatus(
-                status=ExperimentState.running, message="Experiment started"
-            )
+
```
berland: maybe remove these whitespace changes?

src/ert/dark_storage/endpoints/experiment_server.py
```diff
@@ -191,9 +191,7 @@ async def start_experiment(
         runner = ExperimentRunner(config)
         try:
             background_tasks.add_task(runner.run)
-            shared_data.status = ExperimentStatus(
```
berland, line 197: was this superfluouos even before this PR?

src/ert/run_models/everest_run_model.py
```diff
@@ -54,6 +54,8 @@
     OutputConstraintConfig,
 )
 from everest.config.forward_model_config import SummaryResults
+from everest.detached.client import ExperimentState
+from everest.detached.everserver import ExperimentStatus
```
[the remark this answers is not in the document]
berland: Good comment, but should be in a separate PR.

src/ert/storage/local_experiment.py
```diff
@@ -269,6 +269,17 @@ def metadata(self) -> dict[str, Any]:
         with open(path, encoding="utf-8") as f:
             return json.load(f)
 
+    @require_write
+    def update_metadata(self, data: dict[str, Any]) -> None:
+        """
+        Update the metadata file for the experiment.
+        """
+        path = self.mount_point / self._metadata_file
+        assert path.exists()
```
berland: If this happens, do we really want an AssertionError or would we rather like an OSError which will be thrown in  `_write_transaction` below?

src/ert/dark_storage/endpoints/experiment_server.py
frode-aarstad: Yes. It was set to the same thing in the run() below

src/ert/storage/local_experiment.py
```diff
@@ -37,6 +37,7 @@
 class _Index(BaseModel):
     id: UUID
     name: str
+    status: dict[str, Any] | None = None
```
yngve-sk: status: ExperimentStatus | None = None

src/ert/storage/local_experiment.py
```diff
@@ -281,6 +283,20 @@ def name(self) -> str:
     def id(self) -> UUID:
         return self._index.id
 
+    @property
+    def status(self) -> dict[str, Any] | None:
```
yngve-sk: should use ExperimentStatus type instead of dict[str, Any]

src/ert/storage/local_experiment.py
```diff
@@ -281,6 +283,20 @@ def name(self) -> str:
...
+        return self._index.status
+
+    @status.setter
+    @require_write
+    def status(self, status: dict[str, Any] | None) -> None:
```
yngve-sk: should use ExperimentStatus type instead of dict[str, Any]

src/ert/run_models/run_model.py
```diff
@@ -860,3 +860,9 @@ def _merge_parameters_from_design_matrix(
                 raise ErtRunError(str(exc)) from exc
 
         return parameters_config, design_matrix
+
+    def set_experiment_status(
```
yngve-sk: Would just directly use the status setter you made on experiment, and do that from the other runmodels too

frode-aarstad: Then I need to rewrite the runmodels to have access to the experiment which they do not have currently

yngve-sk: I checked some and I agree, it seems the refactor of runmodels made less duplication of code but also made it harder to do this, can keep as is for now. 

We should consider refactoring it to make the experiment / initial sampling ensemble initialization happen within the scope of the runmodel implementation (manual update, ensemble experiment, esmda etc) itself

frode-aarstad: Yes, creating the experiment before starting the run_model seems like a good idea.

src/ert/storage/local_experiment.py
frode-aarstad: This code has now been removed

src/ert/run_models/everest_run_model.py
```diff
@@ -490,22 +509,46 @@ def run_experiment(
...
+                        message="Optimization aborted.", status=ExperimentState.stopped
+                    )
                 case RoptExitCode.TOO_FEW_REALIZATIONS:
                     self._exit_code = (
                         EverestExitCode.TOO_FEW_REALIZATIONS
                         if self.get_number_of_successful_realizations() > 0
                         else EverestExitCode.ALL_REALIZATIONS_FAILED
                     )
+                    experiment_status = ExperimentStatus(
+                        message="Too few realizations."
```
andreas-el: This is short for `Too few realizations succeeded` ?

src/ert/run_models/everest_run_model.py
```diff
@@ -490,22 +509,46 @@ def run_experiment(
         ):
             self._ever_storage.export_everest_opt_results_to_csv()
 
+        experiment_status = None
         if self._exit_code is None:
             match optimizer_exit_code:
                 case RoptExitCode.MAX_FUNCTIONS_REACHED:
                     self._exit_code = EverestExitCode.MAX_FUNCTIONS_REACHED
+                    experiment_status = ExperimentStatus(
+                        message="Maximum number of function evaluations reached.",
```
andreas-el: This is very nitpick. The messages passed to ExperimentStatus should either end with period, or not.

src/ert/storage/local_experiment.py
```diff
@@ -34,9 +35,24 @@
...
+    running = auto()
+    completed = auto()
+    stopped = auto()
+    failed = auto()
+    never_run = auto()
+
+
+class ExperimentStatus(BaseModel):
+    message: str = ""
+    status: ExperimentState = ExperimentState.pending
```
andreas-el: Shouldn't the state be `never_run` ? I see that it was pending prior to you refactor.

src/ert/run_models/everest_run_model.py
frode-aarstad: Aparently short for "Too few realizations are evaluated successfully"

src/ert/storage/local_experiment.py
frode-aarstad: It is still pending. never_run can probably be removed in a seperate PR

src/ert/run_models/run_model.py
yngve-sk: Runmodels now have access to the experiment, so this function should not be needed.

src/everest/bin/utils.py
```diff
@@ -477,3 +461,19 @@ def show_scaled_controls_warning() -> None:
                 logging.getLogger(EVEREST).error(str(e))
         case "n":
             raise SystemExit(0)
+
+
+def get_experiment_status(config: EverestConfig) -> ExperimentStatus:
```
yngve-sk: Could we take in just the storage dir here? The entire config to only take one dir from it seems excessive. Could pass in `config.storage_dir` from call site, ref
```
    @property
    def storage_dir(self) -> str:
        return self._get_output_subdirectory(STORAGE_DIR)
```

yngve-sk: Is this covered by some test(s)?

frode-aarstad: Added a test and changed parameter to string

src/everest/bin/utils.py
```diff
@@ -477,3 +461,19 @@ def show_scaled_controls_warning() -> None:
...
             raise SystemExit(0)
+
+
+def get_experiment_status(output_dir: str) -> ExperimentStatus:
+    """
+    Reads the experiment status from storage. If no experiments are found,
+    returns an ExperimentStatus with status 'never_run'. We assume that there is
+    only one experiment for each everest run in storage.
+    """
+    storage_dir = os.path.join(output_dir, STORAGE_DIR)
```
yngve-sk: There is a `storage_dir` property on `EverestConfig` that does the same join as above, not sure why it should be re-specified here?

yngve-sk:
Not sure if it needs/should be migrated _right now_. `everest results` works OK now when running on a case made before this PR (ERT main). When Everest storage goes through ERT storage (pending: https://github.com/equinor/ert/issues/11129 ) we will however need a migration for this.

src/ert/storage/migration/to15.py
```diff
@@ -0,0 +1,19 @@
+import json
+from pathlib import Path
+
+info = "Add status field to experiment index.json files"
+
+
+def migrate_index_json(path: Path) -> None:
```
yngve-sk, line 14: Could you write a small test for this? ref `test_to13.py` `setup_experiments_and_ensembles`

src/ert/storage/migration/to15.py
```diff
@@ -0,0 +1,23 @@
...
+
+info = "Add status field to experiment index.json files"
+
+
+def update_json(index_json: dict[str, Any]) -> dict[str, Any]:
+    index_json["status"] = {"message": "", "status": "completed"}
+    return index_json
+
+
+def migrate_index_json(path: Path) -> None:
```
yngve-sk, line 14:
```suggestion
def migrate_experiment_status_to_index_json(path: Path) -> None:
```

src/ert/storage/migration/to15.py
```diff
@@ -0,0 +1,23 @@
+import json
+from pathlib import Path
+from typing import Any
+
+info = "Add status field to experiment index.json files"
+
+
+def update_json(index_json: dict[str, Any]) -> dict[str, Any]:
```
yngve-sk:
```suggestion
def add_experiment_status_to_index_json(index_json: dict[str, Any]) -> dict[str, Any]:
```

frode-aarstad: ok as is

src/ert/run_models/everest_run_model.py
```diff
@@ -617,22 +628,46 @@ def run_experiment(
...
+                        status=ExperimentState.failed,
+                    )
                 case _:
                     self._exit_code = EverestExitCode.COMPLETED
+                    experiment_status = ExperimentStatus(
+                        message="Experiment completed", status=ExperimentState.completed
+                    )
+
+        if experiment_status is not None:
+            self._experiment.status = experiment_status
```
berland, line 670: Does this also maybe warrant a `logger.info` with the same information?

src/ert/storage/migration/to15.py
```diff
@@ -0,0 +1,23 @@
...
+    index_json["status"] = {"message": "", "status": "completed"}
+    return index_json
+
+
+def migrate_index_json(path: Path) -> None:
+    for experiment in path.glob("experiments/*"):
+        index_path = experiment / "index.json"
+        with open(index_path, encoding="utf-8") as f:
+            index_data = json.load(f)
+        with open(index_path, "w", encoding="utf-8") as f:
```
berland: This should use pathlib (will not `ruff` complain about this if you have rebased on `main` ??)

src/ert/run_models/everest_run_model.py
berland: Nope, this is already logged in `everest_script.py`

berland:
When running `ert` on  the `poly.ert` case, the status in the index.json ends up as "pending" when the experiement is completed.

src/ert/storage/migration/to15.py
frode-aarstad: `pre-commit run --all-files ` does not fail

src/everest/bin/utils.py
```diff
@@ -477,3 +461,17 @@ def show_scaled_controls_warning() -> None:
...
+    Reads the experiment status from storage. We assume that there is
+    only one experiment for each everest run in storage.
+    """
+    with open_storage(storage_dir, "r") as storage:
+        if len(list(storage.experiments)) == 0:
+            return None
+        experiment_status = next(iter(storage.experiments)).status
+        if experiment_status is not None:
+            return experiment_status
+        return None
```
andreas-el: Could this be;
```
    with open_storage(storage_dir, "r") as storage:
        experiments = list(storage.experiments)
        return None if not experiments else experiments[0].status
```
the next-iter solution works fine, but felt slightly off wrt the docstring.

frode-aarstad: Seems like a good improvement. Will change

PR #11984, opened 2025-10-02
Create forward model steps without going through ert config

yngve-sk:
Everest models PR (tests pass locally with that + this PR ERT installed): https://github.com/equinor/everest-models/pull/182

First commit inlines the initialization to `EverestRunModel.create`. Subsequent commits move the mapping from an Everest subconfig (installdata, install job etc) to erT forward model steps to the respective everest subconfig. This makes it more explicit and testable how the everest config elements map to ERT forward model steps.

Requires update to everest-models, all runmodel.create() must be done from within a plugin context. In ERT it is currently done within plugin context without site queue options.

**Issue**
Resolves https://github.com/equinor/ert/issues/11953

### DIFF
--- a/src/everest/config/install_data_config.py
+++ b/src/everest/config/install_data_config.py
@@ -130,6 +130,14 @@ class InstallDataConfig(BaseModel):
     ) -> ForwardModelStep:
         target = self.target
 
+        def _missing_fm_msg(fm_name: str) -> str:
+            return (
+                f"Using install_data in Everest requires the "
+                f"ERT forward model: {fm_name} to be installed. "
+                f"It was not found in the installed forward model steps: "
+                + (", ".join(installed_fm_steps.keys()))
+            )
+
         if self.source is not None:
             source = self.source.replace("<CONFIG_PATH>", config_directory)
             if not os.path.isabs(source):
@@ -145,13 +153,19 @@ class InstallDataConfig(BaseModel):
                 fm_name = "copy_file"
 
             fm_step_instance = copy.deepcopy(installed_fm_steps.get(fm_name))
-            assert fm_step_instance is not None
+            if fm_step_instance is None:
+                error_message = _missing_fm_msg(fm_name)
+                raise KeyError(error_message)
+
             fm_step_instance.arglist = [source, target]
             return fm_step_instance
         else:
             data_storage = (Path(output_directory) / ".internal_data").resolve()
             data_file = data_storage / Path(self.target).with_suffix(".json")
             fm_step_instance = copy.deepcopy(installed_fm_steps.get("copy_file"))
-            assert fm_step_instance is not None
+            if fm_step_instance is None:
+                error_message = _missing_fm_msg("copy_file")
+                raise KeyError(error_message)
+
             fm_step_instance.arglist = [str(data_file), Path(data_file).name]
             return fm_step_instance
--- a/tests/everest/test_res_initialization.py
+++ b/tests/everest/test_res_initialization.py
@@ -25,7 +25,7 @@ from ert.config.queue_config import (
 )
 from ert.plugins import ErtPluginContext, ErtRuntimePlugins
 from ert.run_models.everest_run_model import EverestRunModel, _get_internal_files
-from everest.config import EverestConfig
+from everest.config import EverestConfig, InstallDataConfig
 from everest.simulator.everest_to_ert import (
     everest_to_ert_config_dict,
     get_substitutions,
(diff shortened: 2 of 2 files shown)


src/ert/run_models/everest_run_model.py
```diff
@@ -64,9 +66,9 @@
     get_optimization_domain_transforms,
 )
 from everest.simulator.everest_to_ert import (
+    _get_well_file,
```
verveerpj: We have to have a look at the `get_internal_files` function in `everest_to_ert`. I think that should also move to the run model, then this becomes a local function.

src/everest/config/install_data_config.py
```diff
@@ -44,3 +74,32 @@ def validate_target(self) -> "InstallDataConfig":
...
+            source = os.path.join(config_directory, source)
+
+        is_dir = _is_dir_all_model(source, model_realizations)
+
+        fm_name: str | None = None
+        if self.link:
+            fm_name = "symlink"
+        elif is_dir:
+            fm_name = "copy_directory"
+        else:
```
verveerpj: This is going to need to change since PR #11998 has been merged.

verveerpj (commented):
It looks good, but it will need to be updated. Also: the create method is becoming really long, maybe we should split it a bit?

src/everest/config/install_data_config.py
yngve-sk: Updated accordingly

src/ert/run_models/everest_run_model.py
yngve-sk: Moved to the EverestRunModel

src/ert/run_models/everest_run_model.py
```diff
@@ -136,6 +138,42 @@ class _EvaluationInfo:
...
+
+def _get_install_data_files(ever_config: EverestConfig) -> Iterator[tuple[Path, str]]:
+    data_storage = (Path(ever_config.output_dir) / ".internal_data").resolve()
+    for item in ever_config.install_data or []:
+        if item.data is not None:
+            target, data = item.inline_data_as_str()
+            yield (data_storage / Path(target).name, data)
+
+
+def get_internal_files(ever_config: EverestConfig) -> dict[Path, str]:
```
verveerpj: Is this a local function? (needs _ ?)

verveerpj (approved):
LGTM! One tiny comment.

src/ert/run_models/everest_run_model.py
```diff
@@ -302,8 +309,123 @@ def create(
...
+        copy_wellfile.arglist = [str(well_path), str(well_path.name)]
+        forward_model_steps.append(copy_wellfile)
+
+        # map templating to template_render job
+        for tmpl_request in everest_config.install_templates:
+            fm_step_instance = copy.deepcopy(installed_fm_steps.get("template_render"))
+            assert fm_step_instance is not None
+            res_input = [control.name for control in everest_config.controls]
+            res_input = [fn + ".json" for fn in res_input]
+            res_input.append(str(_get_well_file(everest_config)[0]))
```
andreas-el: Can this be moved outside the loop?

yngve-sk: fixed

src/everest/config/install_template_config.py
```diff
@@ -29,3 +34,34 @@ class InstallTemplateConfig(BaseModel, extra="forbid"):
...
     )
+
+    def to_ert_forward_model_step(
+        self,
+        control_names: list[str],
+        installed_fm_steps: dict[str, ForwardModelStep],
+        well_path: str,
+    ) -> ForwardModelStep:
+        fm_step_instance = copy.deepcopy(installed_fm_steps.get("template_render"))
+        assert fm_step_instance is not None
```
andreas-el: Rather raise a different exception instead?
Is it possible to continue without this specific failing forward model?

src/everest/config/install_data_config.py
```diff
@@ -90,3 +120,41 @@ def inline_data_as_str(self) -> tuple[str, str]:
...
+                fm_name = "symlink"
+            elif is_dir:
+                fm_name = "copy_directory"
+            else:
+                fm_name = "copy_file"
+
+            assert isinstance(fm_name, str)
+
+            fm_step_instance = copy.deepcopy(installed_fm_steps.get(fm_name))
+            assert fm_step_instance is not None
```
andreas-el: I'd rather have some exception here if possible to log this properly.

src/everest/config/install_data_config.py
```diff
@@ -90,3 +120,41 @@ def inline_data_as_str(self) -> tuple[str, str]:
...
+
+            fm_name: str | None = None
+            if self.link:
+                fm_name = "symlink"
+            elif is_dir:
+                fm_name = "copy_directory"
+            else:
+                fm_name = "copy_file"
+
+            assert isinstance(fm_name, str)
```
andreas-el: Won't this a string regardless? Why offer the alternatives?
We could use fm_name = "copy_file" as the default?

src/everest/config/install_template_config.py
yngve-sk: Should be a hardfail, as the templating depends on this forward model. Now raising error. Note that this should never happen though, as template_render is built into ERT, severe bug if it is somehow gone.

src/everest/config/install_data_config.py
yngve-sk: Fixed

src/everest/config/install_data_config.py
yngve-sk: Fixed

PR #12000, opened 2025-10-06
Remove the default for perturbation_magnitude

### LINKED ISSUE #9844: guidance/validation on perturbation magnitude?

verveerpj:
**Issue**
Resolves #9844

**Approach**
The default for the perturbation_magnitude would be wrong in case the relative perturbation type is selected. In addition, choosing a good default is non-trivial and current practice is to always set it manually. This PR makes this mandatory.

This will also require a fix in everest-models: https://github.com/equinor/everest-models/pull/184

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/config/control_config.py
+++ b/src/everest/config/control_config.py
@@ -167,8 +167,7 @@ class ControlConfig(BaseModel):
             """
         ),
     )
-    perturbation_magnitude: float | None = Field(
-        default=None,
+    perturbation_magnitude: float = Field(
         description=dedent(
             """
             The magnitude of the perturbation of all control variables in the
@@ -177,6 +176,14 @@ class ControlConfig(BaseModel):
             This controls the magnitude of perturbations (e.g. the standard
             deviation in case of a normal distribution) of controls, used to
             approximate the gradient.
+
+            The interpretation of this field depends on the value of the
+            `perturbation_type` field:
+
+            - `absolute`: The given value is used as-is.
+            - `relative`: The perturbation magnitude is calculated by
+              multiplying the given by value by the difference between the `max`
+              and `min` fields.
 
             This default value can be overridden at the variable level.
             """
--- a/src/everest/config/control_variable_config.py
+++ b/src/everest/config/control_variable_config.py
@@ -98,6 +98,20 @@ class _ControlVariable(BaseModel):
             """
         ),
     )
+    perturbation_type: Literal["absolute", "relative"] = Field(
+        default="absolute",
+        description=dedent(
+            """
+            The perturbation type for the control group.
+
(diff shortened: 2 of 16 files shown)


berland:
Should it be deprecated for some time before we remove it?

andreas-el:
Should we backport a warning or error if the default pertubation_magnitude is None, and `relative` is used?

Also, could an option be to not require the perturbation_magnitude set, but warn (as a deprecation) that users need to explicitly set this?
Then error if `relative` is used and perturbation_magnitude is None?

verveerpj:
> Should we backport a warning or error if the default pertubation_magnitude is None, and `relative` is used?

I don't think it is very critical, most likely because `perturbation_magnitude` is always set by the users. And we have not really seen it go wrong when they don't.

> 
> Also, could an option be to not require the perturbation_magnitude set, but warn (as a deprecation) that users need to explicitly set this? Then error if `relative` is used and perturbation_magnitude is None?

I would rather not make it more complicated, simply requiring it now will probably not even be noticed...

andreas-el:
There are zero uses of `perturbation_type='relative'` in the logs, so perhaps the effect will be negligible.

Also, for `perturbation_magnitude=None` there are almost exclusively just developers in Scout doing testing with that setting. There is one where I can't see who ran Everest, but this is before when we redacted user information from the logs.
[Azure logs](https://portal.azure.com#@3aa4a235-b6e2-48d5-9195-7fcf05b459b0/blade/Microsoft_OperationsManagementSuite_Workspace/Logs.ReactView/resourceId/%2Fsubscriptions%2Fe823d00d-e01a-4af3-b71b-61df62a8e3ea%2Fresourcegroups%2Ffmu-logging%2Fproviders%2Fmicrosoft.operationalinsights%2Fworkspaces%2Ffmu-logs/source/LogsBlade.AnalyticsShareLinkToQuery/q/H4sIAAAAAAAAA0WNy2oCURBE9%252FmKzl2NEGT8AAOusggJgo9NCEM7U7k2Yt%252Bhu0cN%252BPE6CSSrgqpzqEXfr41b%252BMOVznsYaC1HvEBhHOjomTiXalbXdTf5Q97gzhnUFg0WdUo9LAbbcUjR5shZJYYO8%252FeiSHcNl4B2tHEYzSmKh4nmamllFAX%252BkYb7lj7%252FP37YR1Gq0lfjbRmiaSU9UdrzaTfmt%252BYTfHow8RBWh47t6nWx%252FV0PaXIDKV07Et0AAAA%253D/limit/1000)

Agreed, i think this is small enough to not bother with.

andreas-el (approved):
Nicely done! 🥇

PR #12009, opened 2025-10-07
Fix displayed paths in the everest gui

### LINKED ISSUE #11355: Remove redundant everest status file

frode-aarstad:
**Issue**
Resolves #11355

Screenshot 2025-10-07 at 12 04 15

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/dark_storage/endpoints/experiment_server.py
+++ b/src/ert/dark_storage/endpoints/experiment_server.py
@@ -18,6 +18,7 @@ from fastapi import (
     WebSocketException,
 )
 from fastapi.encoders import jsonable_encoder
+from fastapi.responses import JSONResponse
 from fastapi.security import HTTPBasic, HTTPBasicCredentials
 from starlette import status
 from starlette.requests import Request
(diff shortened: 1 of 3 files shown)


andreas-el (approved):
Well done! 🎉

PR #12012, opened 2025-10-07
Use assert_never for matches that should be exhaustive

eivindjahren:
This uses assert_never to type-check that matches that should be exhaustive stay that way. Will also produce a runtime error when the match is not exhaustive.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/scheduler/__init__.py
+++ b/src/ert/scheduler/__init__.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from os import getuid
 from pwd import getpwuid
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, assert_never
 
 from ert.config import QueueSystem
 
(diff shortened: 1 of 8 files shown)


src/ert/scheduler/__init__.py
```diff
@@ -33,9 +33,8 @@ def create_driver(queue_options: QueueOptions) -> Driver:
                     **queue_options.driver_options,
                 )
             )
-    raise NotImplementedError(
-        "Only LOCAL, SLURM, TORQUE and LSF drivers are implemented"
```
SAKavli, line 37: I suppose a config validation error is raised if something else is raised such that we dont need the message?

src/ert/scheduler/job.py
```diff
@@ -434,6 +434,8 @@ async def _send(self, state: JobState) -> None:
                 event = RealizationSuccess(real=str(self.iens))
                 self._end_time = time.time()
                 await self._scheduler.completed_jobs.put(self.iens)
+            case default:
+                assert_never(default)
```
SAKavli, line 438: Is this one superfluous as we have typing of input?
```py
state: JobState
```

SAKavli: Well, if this is to catch bugs it might be fine

src/ert/scheduler/__init__.py
eivindjahren: What is raised is an assertion error. This should never happen as long as typing is followed.

PR #12029, opened 2025-10-08
Skip recreating ensemble directory in storage

### LINKED ISSUE #12025: local_ensemble.set_failure will make directories uncritically
The `mkdir` in:

https://github.com/equinor/ert/blob/3b514301fd98e92efd832f96edfe02b9faa323e9/src/ert/storage/local_ensemble.py#L335

will uncritically make parent directories in storage.

An observed user pattern is that users realize that an ensemble will fail, and then they delete storage (and maybe also tries to stop Ert by asking it to kill its realizations). The mentioned `mkdir` will be run by forthcoming failing realizations, and recreates the ensemble directory in a new storage. 

This storage is however garbled (missing index.json and more), and will on next opening of Ert trigger an exception (probably un-noticed by the user) with the message "`Failed to load an ensemble from path`". 

This is at least explaining a subset of the causes for #11855

berland:
When writing errors.json (or responses) we should not recreate the ensemble directory in storage if it does not exist. Then the user has probably deleted it manually while Ert is running, and whatever happens is undefined behaviour.

**Issue**
Resolves #12025

**Approach**
Restrict the `mkdir`.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/storage/local_ensemble.py
+++ b/src/ert/storage/local_ensemble.py
@@ -946,7 +946,7 @@ class LocalEnsemble(BaseMode):
             )
 
         output_path = self._realization_dir(realization)
-        Path.mkdir(output_path, parents=True, exist_ok=True)
+        Path(output_path).mkdir(exist_ok=True)
 
         self._storage._to_parquet_transaction(
             output_path / f"{response_type}.parquet", data
--- a/tests/ert/unit_tests/storage/test_local_storage.py
+++ b/tests/ert/unit_tests/storage/test_local_storage.py
@@ -1167,6 +1167,57 @@ def test_set_failure_will_not_recreate_ensemble_directory(storage):
     assert not dummy_ensemble._path.exists()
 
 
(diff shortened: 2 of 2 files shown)


src/ert/storage/local_ensemble.py
```diff
@@ -332,8 +331,14 @@ def set_failure(
         """
 
         filename: Path = self._realization_dir(realization) / self._error_log_name
-        os.makedirs(os.path.dirname(filename), exist_ok=True)
         error = _Failure(type=failure_type, message=message or "", time=datetime.now())
+        if not filename.parent.parent.exists():
+            logger.warning(
+                f"Storage {filename.parent.parent} does not exist, "
```
jonathan-eq, line 337: Why filename.parent.parent and not filename.parent?

jonathan-eq: Right, it has to create `REALIZATION-<IENS>` but not `ensembles`...

PR #12033, opened 2025-10-09
Move design matrix merge out of runmodels

yngve-sk:
Required for https://github.com/equinor/ert/pull/11513 , now we store only the parsed design matrix dataframe in the runmodel, so that there is no need to do the merging automatically within the runmodel.

### DIFF
--- a/src/ert/run_models/initial_ensemble_run_model.py
+++ b/src/ert/run_models/initial_ensemble_run_model.py
@@ -7,7 +7,6 @@ from polars.datatypes import DataTypeClass
 from pydantic import BaseModel, Field, field_validator
 
 from ert.config import (
-    DesignMatrix,
     EverestConstraintsConfig,
     EverestObjectivesConfig,
     ExtParamConfig,
@@ -59,7 +58,7 @@ class DictEncodedDataFrame(BaseModel):
 
 class InitialEnsembleRunModel(RunModel, ABC):
     experiment_name: str
-    design_matrix: DesignMatrix | None
+    design_matrix: DictEncodedDataFrame | None
     parameter_configuration: list[
         Annotated[
             (GenKwConfig | SurfaceConfig | FieldConfig | ExtParamConfig),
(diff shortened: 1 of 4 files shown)


src/ert/run_models/initial_ensemble_run_model.py
```diff
@@ -37,7 +36,7 @@ def str_to_dtype(dtype_str: str) -> pl.DataType:
     return dtype
 
 
-class DictEncodedObservations(BaseModel):
+class DictEncodedDataFrame(BaseModel):
```
xjules, line 39: This looks very complex. Chatgpt suggests:
```python
class ModelWithPolars(BaseModel):
    df: pl.DataFrame
@field_validator("df", mode="before")
@classmethod
def _serialize_df(self, df: pl.DataFrame):
        return {
            "columns": df.columns,
            "dtypes": [str(dt) for dt in df.dtypes],  # optional but handy
            "rows": df.rows(),  # -> list[list], JSON-friendly
        }
@field_serializer("df", when_used="json")
def _coerce_df(cls, v):
        # Accept a real DataFrame
        if isinstance(v, pl.DataFrame):
            return v
        # Accept our JSON shape: {"columns": [...], "dtypes": [...], "rows": [...]}
        if isinstance(v, dict) and {"columns", "rows"} <= set(v):
            # Optional: reconstruct dtypes if provided
            if "dtypes" in v:
                dtype_map = {c: getattr(pl, dt) if isinstance(dt, str) else dt
                             for c, dt in zip(v["columns"], v["dtypes"], strict=True)}
                return pl.DataFrame(v["rows"], schema=dtype_map)
            return pl.DataFrame(v["rows"], schema=v["columns"])
        # Accept a list[dict] (records)
        if isinstance(v, list):
            return pl.DataFrame(v)
        raise TypeError("Unsupported df input for ModelWithPolars.df")
```

yngve-sk: (I'm not sure if the above is less complex? 😅) But we also use this to encode polars dataframes, specifically for observations. The above function does the same thing, only that the existing uses the .to_dicts() format with one object per row, and a schema. It is debatable how we serialize the dataframes though. Serializable JSON schema in polars is coming up (https://github.com/pola-rs/polars/issues/20426), then the serialization / deserialization code will be easier.

src/ert/sample_prior.py
```diff
@@ -33,6 +34,8 @@ def sample_prior(
     parameter_configs = ensemble.experiment.parameter_configuration
     if parameters is None:
         parameters = list(parameter_configs.keys())
+
+    design_matrix_df = design_matrix_dicts.to_polars() if design_matrix_dicts else None
```
xjules: nice!

Issue #12016, opened 2025-10-07
File name adjustment for NTFS compatibility


GlauciaSchnoeller:
Hello,

Recently, we had a change in our infrastructure with the migration of the filesystem (NTFS) on Windows. This caused errors when using ERT to store files generated in the shared area.
We encountered issues with some files named in the ISO 8601 format. These files include the ":" separator for hours, which is not allowed by NTFS on Windows. This format is also used in other files.
We replaced ":" with "-" in these files, which resolved the issue.

We would like to know if it is possible to change the format from, for example:
ert-log-poly-ert-2025-09-30T14:39.txt
to:
ert-log-poly-ert-2025-09-30T14-39.txt

This substitution could be applied only on Windows filesystems to keep the original format elsewhere.
We understand that the current format is a convention of the tool, so we want to confirm if this change would be appropriate before opening a PR with the adjustment.

Thank you for your assistance!

berland:
Thanks for notifying. According to https://en.wikipedia.org/wiki/ISO_8601#Times we should then just drop the colon.

I have notices that the colon causes some issues with bash completion on Linux as well, so I am probably in favour of this.

PR #12040, opened 2025-10-09
Make ISO timestamps safe for filenames

GlauciaSchnoeller:
**Issue**
Resolves [#12016](https://github.com/equinor/ert/issues/12016)

**Approach**  
Create `file_safe_timestamp` function to make ISO timestamps safe for filenames by replacing ":".

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/_ert/forward_model_runner/fm_dispatch.py
+++ b/src/_ert/forward_model_runner/fm_dispatch.py
@@ -19,6 +19,7 @@ from _ert.forward_model_runner.reporting.message import (
     ProcessTreeStatus,
 )
 from _ert.forward_model_runner.runner import ForwardModelRunner
+from _ert.utils import file_safe_timestamp
 
 if TYPE_CHECKING:
     from ert.config.forward_model_step import ForwardModelJSON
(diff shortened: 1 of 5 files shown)


src/ert/utils/__init__.py
```diff
@@ -30,3 +30,8 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
         return wrapper
 
     return decorator
+
+
+def file_safe_timestamp(timestamp: str) -> str:
+    """Convert an ISO timestamp string to a file-safe version"""
```
berland: It could be added something in the docstring that this implies that we are using the "basic" format  ISO8601 in the time part, while keeping the "extended" format for the date part. Mixing the two formats is not ISO8601 compliant, but datetime.datetime.fromisoformat() is still able to parse it.

GlauciaSchnoeller: Thanks a lot for the observation! 
It really makes sense. I have already updated the docstring as you suggested.

GlauciaSchnoeller:
Thanks! During test execution, I observed that importing the ert.utils module inside _ert caused new environment variables to be created. This modifies the global environment and can interfere with module initialization, leading to test failures.

For this reason, I simplified the implementation. The file_safe_timestamp function was changed locally in the external modules (_ert and everest), replacing the import with a direct timestamp.replace(":", "") call. This removes the dependency between ert and _ert and ensures more stable imports. Although not the most ideal solution for code reuse, it prevents environment conflicts during module initialization.

berland (approved):
Good job!

PR #12044, opened 2025-10-10
Change ExtParamConfig to use polars instead of xarray

### LINKED ISSUE #12005: Refactor ExtParamConfig to use pl.DataFrame as storage format
The current implementation of ExtParamConfig in ERT stores parameter data as an xarray.Dataset, which is inconsistent with how scalar parameters (GenKwConfig) are handled using polars.DataFrame.
This should allow for better ensemble-level data operations, serialization, and integration with the new storage and migration logic that assumes pl.DataFrame for scalar-like parameters.

Suggested approach:

```python
class ExtParamConfig(GenKwConfig):
    type: Literal["everest_parameters"] = "everest_parameters"
    input_keys: list[str] = field(default_factory=list)
```

jonathan-eq:
**Issue**
Resolves #12005

**Approach**
This should have no effect on the result whatsoever, as both the input
and output will look exactly the same. The difference is that values will be put in the ensemble SCALARS.parquet instead of netcdf files in each realization directory. This data is only used between the ropt optimizers and Everest Run Model, and it won't be used afterwards, so it shouldnt really require a storage migration but we add one just in case.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/ext_param_config.py
+++ b/src/ert/config/ext_param_config.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import json
 from collections.abc import Iterator, Mapping, MutableMapping
-from dataclasses import field
 from pathlib import Path
 from typing import TYPE_CHECKING, Literal
 
@@ -26,15 +25,9 @@ if TYPE_CHECKING:
 
 
 class ExtParamConfig(ParameterConfig):
-    """Create an ExtParamConfig for @key with the given @input_keys
-
-    @input_keys can be either a list of keys as strings or a dict with
-    keys as strings and a list of suffixes for each key.
-    If a list of strings is given, the order is preserved.
-    """
+    """Create an ExtParamConfig for @key with the given @group"""
 
     type: Literal["everest_parameters"] = "everest_parameters"
-    input_keys: list[str] = field(default_factory=list)
     forward_init: bool = False
     output_file: str = ""
     forward_init_file: str = ""
--- a/tests/ert/unit_tests/storage/test_local_storage.py
+++ b/tests/ert/unit_tests/storage/test_local_storage.py
@@ -958,12 +958,8 @@ def test_that_all_parameters_and_gen_data_consolidation_works(
         experiment = storage.create_experiment(
             responses=[GenDataConfig(keys=["R1", "R2"])],
             parameters=[
-                ExtParamConfig(
-                    name="point.P1", group="point", input_keys=["point.P1", "point.P2"]
-                ),
-                ExtParamConfig(
-                    name="point.P2", group="point", input_keys=["point.P1", "point.P2"]
-                ),
+                ExtParamConfig(name="point.P1", group="point"),
+                ExtParamConfig(name="point.P2", group="point"),
             ],
         )
 
--- a/src/everest/config/control_config.py
+++ b/src/everest/config/control_config.py
@@ -318,7 +318,6 @@ class ControlConfig(BaseModel):
             ExtParamConfig(
                 group=self.name,
                 name=control_name,
-                input_keys=self.formatted_control_names,
                 output_file=self.name + ".json",
             )
             for control_name in self.formatted_control_names


src/ert/storage/local_ensemble.py
```diff
@@ -863,14 +880,25 @@ def save_parameters(
...
-                df = df.drop(
-                    [c for c in dataset.columns if c != "realization"], strict=False
-                )
-                df_full = (
-                    df.join(dataset, on="realization", how="left")
-                    .unique(subset=["realization"], keep="first")
-                    .sort("realization")
-                )
+                if might_have_new_realizations:
+                    df = df.drop(  # Why we do this??
```
jonathan-eq: We ONLY delete the columns that are in the new dataset

tests/ert/unit_tests/storage/test_local_storage.py
```diff
@@ -900,7 +900,9 @@ def test_that_all_parameters_and_gen_data_consolidation_works(
 
         experiment = storage.create_experiment(
             responses=[GenDataConfig(keys=["R1", "R2"])],
-            parameters=[ExtParamConfig(name="point", input_keys=["P1", "P2"])],
+            parameters=[
+                ExtParamConfig(name="point", input_keys=["point.P1", "point.P2"])
```
jonathan-eq: This had to be changed for us to use `save_parameters_numpy`, and was part of an earlier storage migration.

src/ert/run_models/everest_run_model.py
```diff
@@ -706,25 +706,21 @@ def _run_forward_model(
...
-                    np.array([sim_id]),
-                )
-                offset += n_param_keys
+            # Save controls to ensemble
+            ensemble.save_parameters_numpy(
+                sim_to_control_vector[:, offset : (offset + n_param_keys)],
+                ext_param_config.name,
+                np.array(sim_vec),
+            )
+            offset += n_param_keys
```
jonathan-eq: We do the same as for genkw here. Instead of looping over reach realization, we are looping over each parameter. This fixes the bug I had with joining polars.DataFrames.

src/ert/config/ext_param_config.py
```diff
@@ -80,20 +82,13 @@ def create_storage_datasets(
...
-                int(realization),
-                xr.Dataset(
-                    {
-                        "values": ("names", from_data[:, i]),
-                        "names": [
-                            x.split(f"{self.name}.")[1].replace(".", "\0")
-                            for x in self.parameter_keys
-                        ],
-                    }
-                ),
```
jonathan-eq, line 94: I do not see why this would work. Does `from_data[:, i]` mean that we only yield column 0 for realization 0, only column 1 for realization 1, only column 2 for realization 2?

src/ert/storage/local_ensemble.py
```diff
@@ -606,11 +611,15 @@ def load_parameters(
...
-            p
-            for p in self.experiment.parameter_configuration.values()
-            if group in {p.name, p.group_name}
-        ]
+        cfgs: list[ParameterConfig] = []
+
+        for p in self.experiment.parameter_configuration.values():
+            if group in {p.name, p.group_name}:
+                cfgs.append(p)
+            if isinstance(p, ExtParamConfig) and any(
```
yngve-sk: This breaks the abstraction we currently have, LocalEnsemble should ideally know only of `ParameterConfig`, would resolve this by changing ExtParamConfig to be same as current GenKwConfig, with same cardinality settings, and a single instance of `ExtParamConfig` per individual control, with the `group` as metadata. If we do this as an intermediate step we will need 2 storage migrations: One for migrating the xr to polars, and one for migrating the config from current format to new one, might as well just do it in one go imo,

src/ert/config/ext_param_config.py
yngve-sk: one column in `from_data`, `from_data[:, i]` is all parameters (within the extparam group) for one realization. so each yield gives the realization nr and the corresponding dataset/polars series

jonathan-eq:
New:
`control,batch,value
point.x.0,initial,0.25
point.x.1,initial,0.25
point.x.2,initial,0.25
point.x.0,initial,0.25
point.x.1,initial,0.25
point.x.2,initial,0.25
point.x.0,best,0.1445293090389739
point.x.1,best,0.017081595559766807
point.x.2,best,0.365267270783631
point.x.0,best,0.1445293090389739
point.x.1,best,0.017081595559766807
point.x.2,best,0.365267270783631
`

New:
`
batch_id,realization,simulation_id,perturbation,batch_distance,batch_point.x.0,batch_point.x.1,batch_point.x.2,batch_total_objective_value,batch_x-0_coord,distance,grad(distance wrt point.x.0),grad(distance wrt point.x.1),grad(distance wrt point.x.2),grad(distance.total wrt point.x.0),grad(distance.total wrt point.x.1),grad(distance.total wrt point.x.2),point.x.0,point.x.1,point.x.2,x-0_coord
0,0,,0,-1.6875,0.9999708162864437,-0.00003602003640486248,-0.00011296466080669269,-1.6875,0.25,-6.197460174560547,-0.4957872898436779,-0.49438596681209324,0.49561241321653354,-0.4957872898436779,-0.49438596681209324,0.49561241321653354,0.2538282445129496,0.24751682992526736,0.23974638454801972,0.2538279891014099
0,0,,1,-1.6875,0.9999708162864437,-0.00003602003640486248,-0.00011296466080669269,-1.6875,0.25,-6.19789981842041,-0.4957872898436779,-0.49438596681209324,0.49561241321653354,-0.4957872898436779,-0.49438596681209324,0.49561241321653354,0.24996884704125027,0.25263343360307505,0.24743589567425595,0.2499690055847168
0,0,,2,-1.6875,0.9999708162864437,-0.00003602003640486248,-0.00011296466080669269,-1.6875,0.25,-6.169020175933838,-0.4957872898436779,-0.49438596681209324,0.49561241321653354,-0.4957872898436779,-0.49438596681209324,0.49561241321653354,0.24853067346762697,0.24651802978838783,0.25233575728243945,0.24853099882602692
0,0,,3,-1.6875,0.9999708162864437,-0.00003602003640486248,-0.00011296466080669269,-1.6875,0.25,-6.191420078277588,-0.4957872898436779,-0.49438596681209324,0.49561241321653354,-0.4957872898436779,-0.49438596681209324,0.49561241321653354,0.24427056529371927,0.257227541473465,0.2528332936646901,0.24427099525928497
0,0,,4,-1.6875,0.9999708162864437,-0.00003602003640486248,-0.00011296466080669269,-1.6875,0.25,-6.183549880981445,-0.4957872898436779,-0.49438596681209324,0.49561241321653354,-0.4957872898436779,-0.49438596681209324,0.49561241321653354,0.25064428907515635,0.24940559927105557,0.2583987698188737,0.2506439983844757
`

New SCALARS.parquet
image

src/ert/config/ext_param_config.py
jonathan-eq: Wait, so is every column a new realization, and not every row?

yngve-sk: Hm, probably saying rows and columns is confusing, but `from_data[:, i]` is all params for that one realization. Whether the outer dimension is cols and inner is rows... If each realization had a column, one column corresponds to the vector `from_data[:i,]`, for example if `from_data` is 
```
[[1,2,3], [4,5,6], [7,8,9]], then the realization with index0 (`[:, 0]`) would have param values [1, 4, 7]
```

ref 
```
>>> import numpy as np
>>> a=np.array([[1,2,3],[4,5,6],[7,8,9]])
>>> a.shape
(3, 3)
>>> a[:, 0]
array([1, 4, 7])
```

jonathan-eq: Ahhh, I see. Thanks!

src/ert/config/ext_param_config.py
```diff
@@ -32,20 +33,13 @@ class ExtParamConfig(ParameterConfig):
...
-    @property
-    def parameter_keys(self) -> list[str]:
-        return self.input_keys
-
-    @property
-    def metadata(self) -> list[ParameterMetadata]:
-        return []
-
     type: Literal["everest_parameters"] = "everest_parameters"
     input_keys: list[str] = field(default_factory=list)
```
jonathan-eq: I tried getting rid of this key, but tests started failing. I am not sure what this does, but there was a storage migration for it three months ago.

jonathan-eq: `input_keys`^

src/ert/config/ext_param_config.py
xjules, line 28: Another suggestion is to Rename `ExtParamConfig` to `EverestControl` to avoid multiple storage migrations.

jonathan-eq: Is this something we want? We can also remove the `input_keys` attribute, but that will also require a storage migration. 
I think both of those ideas are outside of the scope of this PR, I just wanted to `Change ExtParamConfig to use polars instead of xarray`

xjules: Ok, will create an issue for it. :heavy_check_mark:

src/ert/config/ext_param_config.py
xjules: But it is in use here: https://github.com/equinor/ert/blob/aab1ef06155385097ab3e72dd34b9fe38cc43932/src/everest/config/control_config.py#L317

PR #12050, opened 2025-10-10
Change ForwardModelStep to BaseModel

yngve-sk:
Split out from https://github.com/equinor/ert/pull/11513

### DIFF
--- a/src/ert/config/forward_model_step.py
+++ b/src/ert/config/forward_model_step.py
@@ -1,10 +1,9 @@
 from __future__ import annotations
 
 import logging
-from abc import abstractmethod
-from dataclasses import dataclass, field
-from typing import ClassVar, Literal, NotRequired
+from typing import ClassVar, Literal, NotRequired, Self
 
+from pydantic import BaseModel, Field, field_validator, model_validator
 from typing_extensions import TypedDict, Unpack
 
 from .parsing import ConfigValidationError, ConfigWarning, SchemaItemType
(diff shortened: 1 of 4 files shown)


src/ert/config/forward_model_step.py
```diff
@@ -184,28 +181,32 @@ def check_required_keywords(self) -> None:
                 self.name,
             )
 
-    def __post_init__(self) -> None:
-        # We unescape backslash here to keep backwards compatability ie. If
-        # the arglist contains a '\n' we interpret it as a newline.
-        self.arglist = [
```
yngve-sk: I think we could remove this part in own commit / PR

src/ert/config/forward_model_step.py
```diff
@@ -184,28 +181,32 @@ def check_required_keywords(self) -> None:
...
-            s.encode("utf-8", "backslashreplace").decode("unicode_escape")
-            for s in self.arglist
-        ]
-
+    @model_validator(mode="after")
+    def set_default_env(self) -> Self:
         self.environment.update(ForwardModelStep.default_env)
+        return self
 
+    @model_validator(mode="after")
```
andreas-el, line 189: If the functions are not explicitly used, we could combine them, e.g.

```
    def set_file_attributes(self) -> Self:
        for attr in ['stdout_file', 'stderr_file', 'stdin_file']:
            value = getattr(self, attr, None)
            if value is None:
                setattr(self, attr, f"{self.name}.{attr.split('_')[0]}")
            elif value == "null":
                setattr(self, attr, None)
        return self
```

yngve-sk: Fixup'd, added only for stdout/stderr, set stdin to field validator

PR #12052, opened 2025-10-10
Add a parameter viewer to experiment panels

### LINKED ISSUE #11936: Implement a window showing all experiment parameters as a table (tree?) view
This concerns only the parameters represented in the storage. Additionally, this should visualize and provide a way to adjust: 
 - `input_source` for the parameter. If the parameter is provided as both (gen_kw & desing) then this can be toggled
 - `update` flag. If `input_source` is defined as sampled then this can be toggled.

@flikka : I think the "choose the source for this parameter" should be a different issue. Mixing it with this "view task" seems wrong. Is this task about showing storage data for parameters (as it appeared), or is to "set up parameters" (as one may suspect).

flikka:
**Issue**
Resolves #11936 

**Approach**
Add a button in relevant experiment panels that opens a window showing a tree-view of parameters, grouped by type

image

      --exec 'just rapid-tests'`)

## When applicable


src/ert/gui/simulation/ensemble_experiment_panel.py
```diff
@@ -39,6 +40,7 @@ class EnsembleExperimentPanel(ExperimentConfigPanel):
     def __init__(
         self,
         analysis_config: AnalysisConfig,
+        ensemble_config: EnsembleConfig,
```
xjules: I think that we want to remove ensemble_config (in the near future). Would it be possible to just sent the required  "info " into the panels. Ie. just experiment and dm parameters.
This would require to change the merge function to be a class method (I guess) that would accept parameter_configurations instances that are to be merged.

flikka: Sure, anything is possible! Let's talk about it, not quite clear what you mean is coming in the future...

xjules: Copy from the discussion:
Yes, merge_with_existing_parameters requires to send in the full design_matrix config, while what we actually need is only two sets of parameters then the panels would only require
```python
parameter_configurations: list[ParameterConfig]
design_parameter_configs: list[ParameterConfig] | None
```

yngve-sk: `EnsembleConfig` only lives in `ErtConfig`, its information is expressed only through parameter / response configs downstream. In the runmodels we only deal with those. Ideally we would not want the GUI to form a direct dependency on EnsembleConfig, as it set up a hurdle towards keeping frontend (GUI)/backend (ErtConfig in this case) separation.

src/ert/gui/ertwidgets/parameterviewer.py
```diff
@@ -0,0 +1,140 @@
...
+                self.type_nodes[parameter.type], [parameter.name]
+            )
+            parameter_node.setData(
+                0, Qt.ItemDataRole.UserRole, parameter.update
+            )  # Store update value for filtering
+            self.parameter_nodes.append(parameter_node)
+            QTreeWidgetItem(parameter_node, [f"Update: {parameter.update}"])
+            QTreeWidgetItem(parameter_node, [f"Forward Init: {parameter.forward_init}"])
+            QTreeWidgetItem(parameter_node, [f"Group: {parameter.group_name.upper()}"])
+            if hasattr(parameter, "input_source"):
```
xjules, line 83: this is very implicit. Maybe directly `if parameter.type=="gen_kw"`?

src/ert/gui/ertwidgets/parameterviewer.py
```diff
@@ -0,0 +1,140 @@
...
+
+from ert.config import ParameterConfig
+
+
+class ParametersViewer(QDialog):
+    def __init__(
+        self, merged_parameters: list[ParameterConfig], parent: QWidget | None = None
+    ) -> None:
+        super().__init__(parent)
+        self.merged_parameters = merged_parameters
```
xjules: I think we can call it safely just `self._parameters` as it displays what we provide.

tests/ert/ui_tests/gui/test_parameterviewer.py
```diff
@@ -0,0 +1,193 @@
...
+    viewer.filter_parameters()
+
+    # Check visible parameters have expected update values
+    for i in range(tree.topLevelItemCount()):
+        type_node = tree.topLevelItem(i)
+        for j in range(type_node.childCount()):
+            param_node = type_node.child(j)
+            update_val = param_node.data(0, Qt.ItemDataRole.UserRole)
+
+            if not param_node.isHidden():
```
xjules, line 192: Can param_node be hidden?

xjules (changes requested):
Besides avoiding ensemble_config, I just had some smaller comments.
Nice work so far!

src/ert/gui/ertwidgets/parameterviewer.py
flikka: I would say it is very _explicit_? show a parameters input_source if it has this attribute?

tests/ert/ui_tests/gui/test_parameterviewer.py
flikka: Indeed, that is what the test is about. Filtering "hides" parameter nodes from the view, we test that only the ones with Update flag XX is shown for the three different scenarios

src/ert/gui/ertwidgets/parameterviewer.py
xjules: codeing-wise it is implicit, but that with `gen_kw` we "claim" it is only gen_kw having the `input_source`

tests/ert/ui_tests/gui/test_parameterviewer.py
xjules: Oh, I've misunderstood the test then. I got it now.

xjules (approved):
No more comments Dr. Flikka!
well played :)
Don't forget to squash the commits.

PR #12055, opened 2025-10-13
Move storage initialization to `.run_experiment` level

yngve-sk:
Makes it possible to access experiment and prior from `.run_experiment` context. Unblocks storing experiment state in the same way for all runmodels, ref https://github.com/equinor/ert/pull/11976

### DIFF
--- a/src/ert/run_models/multiple_data_assimilation.py
+++ b/src/ert/run_models/multiple_data_assimilation.py
@@ -2,12 +2,17 @@ from __future__ import annotations
 
 import functools
 import logging
-from typing import Any, ClassVar
+from typing import Any, ClassVar, cast
 from uuid import UUID
 
 from pydantic import PrivateAttr
 
-from ert.config import PostExperimentFixtures, PreExperimentFixtures
+from ert.config import (
+    ParameterConfig,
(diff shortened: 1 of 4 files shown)


src/ert/run_models/multiple_data_assimilation.py
```diff
@@ -98,12 +103,28 @@ def run_experiment(
...
-                self.target_ensemble % 0,
+            experiment_storage = self._storage.create_experiment(
+                parameters=cast(list[ParameterConfig], self.parameter_configuration),
+                observations={k: v.to_polars() for k, v in self.observations.items()}
+                if self.observations is not None
+                else None,
+                responses=cast(list[ResponseConfig], self.response_configuration),
+                name=self.experiment_name,
+                templates=self.ert_templates,
+                simulation_arguments=sim_args,
```
frode-aarstad, line 114: Is simulation_arguments not needed in the classes above?

yngve-sk: Only used for initializing the experiment from `_sample_and_evaluate_ensemble` on main as far as I can tell

Issue #12024, opened 2025-10-08
Bug related to the mask for field parameters(current bleeding version)


oddvarlia:
### What happened?

Running Drogon case with individual ERTBOX parameters per 3D field parameter group with different definition of the number of layers of the ERTBOX grid per geological zone fails due to no appropriate handling of the active/inactive mask. The approach to use individual ertbox definition per field parameter will probably require that the mask is saved together with the field parameter since it vary in size from field to field (in realization from geological zone to zone). The current implementation has one common mask file for all fields, not one per field. But note that the mask is actually not used when using individual ertbox definition per field (or even for a common ertbox). The mask in only relevant when using the 'old' approach  when the grid is fixed and only one geological zone is present in the geomodel. 

The missing handling of the mask file results in inconsistency in some dimensionality check since it is impossible to share the same mask file for all fields anymore when using individual size (ertbox definition) per 3D field parameter group. The error occurs in ERT storage. 

### What did you expect to happen?

No error when using individual ertbox definition (ertbox grid) per field.

### Steps to reproduce

Use a case .e.g Drogon with individual ERTBOX per zone with different NZ for each geological zone. All realizations will fail when saving to ERT storage.

<img width="681" height="785" alt="Image" src="https://github.com/user-attachments/assets/15f895ee-00c7-4bde-aed6-588e7316222c" />

This is how my FIELD keywords look like (I have defined a symbol for each input ertbox grid).
NZ for Valysar, Therys and Volon are all different with size NZ = 66 for Therys as the largest one. The other two have 15 and 20 layers.
<img width="1496" height="246" alt="Image" src="https://github.com/user-attachments/assets/256945bc-8ef8-4c02-be76-9acce5699eeb" />

oddvarlia:
I have some suggestions here related to solving this issue:

[...]

PR #12066, opened 2025-10-14
Remove mask from field

dafeda:
Resolves: #12024 

We only support updating fields on an ertbox grid which never has inactive cells.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/field.py
+++ b/src/ert/config/field.py
@@ -65,24 +65,6 @@ def create_flattened_cube_graph(px: int, py: int, pz: int) -> nx.Graph[int]:
     return G
 
 
-def adjust_graph_for_masking(G: nx.Graph[int]) -> nx.Graph[int]:
-    """
-    Adjust the graph G according to the masking indices.
-    Removes nodes specified by the mask and relabels the remaining nodes
-    to have consecutive labels from 0 to G.number_of_nodes - 1.
-    Parameters:
-    - G: The graph to adjust
-    - mask: Boolean mask flattened array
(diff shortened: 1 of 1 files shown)


src/ert/config/field.py
```diff
@@ -67,9 +65,7 @@ def create_flattened_cube_graph(px: int, py: int, pz: int) -> nx.Graph[int]:
     return G
 
 
-def adjust_graph_for_masking(
-    G: nx.Graph[int], mask: npt.NDArray[np.bool_]
-) -> nx.Graph[int]:
+def adjust_graph_for_masking(G: nx.Graph[int]) -> nx.Graph[int]:
```
xjules: I guess this function can be removed completely?

src/ert/config/field.py
```diff
@@ -324,13 +305,18 @@ def create_storage_datasets(
...
-                mask=self.mask,
+                data=np.zeros(
+                    self.ertbox_params.nx
+                    * self.ertbox_params.ny
+                    * self.ertbox_params.nz
+                ),
                 fill_value=np.nan,
                 dtype=from_data.dtype,
             )
             ma[~ma.mask] = from_data[:, i]
```
xjules: why we need to do this. is it not enough to?
```python
data = from_data[:, i].reshape(self.ertbox_params.nx, self.ertbox_params.ny, self.ertbox_params.nz)
ds = xr.Dataset({"values": (["x", "y", "z"], data})
```

Issue #12067, opened 2025-10-14
Implement toggle button for using log scale for parameter visualization


xjules:
Since we don't do the LOG10_ prefix anymore, the usage of log scale when visualizing parameters can be done on a custom basis by having a toggle (or a checkbox), which would allow to change axis from linear to log.

PR #12069, opened 2025-10-14
Log scale toggle

xjules:
**Issue**
Resolves https://github.com/equinor/ert/issues/12067

- [x] remove log scale toggle from other plots; ie. it will be available only for histograms
- [x] something has changed in the plot_window and thus new axes are created for each plot. This needs to be fixed.

**Approach**
chebox for log scale
- **linear scale**
image

- **log scale**
image

      --exec 'just rapid-tests'`)

## When applicable


src/ert/gui/tools/plot/plot_widget.py
```diff
@@ -133,7 +141,21 @@ def __init__(
...
 
+        self._log_checkbox = QCheckBox("Log scale", self)
+        self._log_checkbox.setObjectName("log_scale_checkbox")
+        self._log_checkbox.setCheckable(True)
+        # only for supported plots see _log_axis_for_plotter
+        self._log_checkbox.setVisible(False)
+        self._log_checkbox.setToolTip("Toggle data domain to log scale and back")
+        self._log_checkbox.toggled.connect(self._on_log_toggled)
+
+        btn_row = QHBoxLayout()
```
jonathan-eq: This might get confused with the button row at the bottom. If it only contains the log_checkbox, maybe we can call it `log_checkbox_row` or something like that

tests/ert/ui_tests/gui/test_main_window.py
```diff
@@ -395,6 +406,17 @@ def click_tab_index(pos: int) -> None:
...
+        # wait until the checkbox exists
+        qtbot.waitUntil(lambda: get_log_checkbox() is not None, timeout=2000)
+        cb = get_log_checkbox()
+
+        # wait until it becomes visible
+        qtbot.waitUntil(
+            lambda: cb.isVisibleTo(plot_window._central_tab.currentWidget()) is True,
+            timeout=2000,
+        )
+        assert cb.isVisibleTo(plot_window._central_tab.currentWidget())
```
jonathan-eq: I wonder if this extra check is something we need. If the previous qtbot.waitUntil does not raise timeouterror, we already know it is visible, so it might be redundant to check for it here as well.

xjules: maybe not. I think I saw this pattern somewhere that we tend to use.
Will remove it.

tests/ert/ui_tests/gui/test_main_window.py
```diff
@@ -395,6 +406,16 @@ def click_tab_index(pos: int) -> None:
...
+        # wait until the checkbox exists
+        qtbot.waitUntil(lambda: get_log_checkbox() is not None, timeout=2000)
+        cb = get_log_checkbox()
+
+        # wait until it becomes visible
+        qtbot.waitUntil(
+            lambda: cb.isVisibleTo(plot_window._central_tab.currentWidget()) is True,
+            timeout=2000,
+        )
+
```
jonathan-eq, line 418: Did you figure out the reason why it wasn't visible to qtbot?

src/ert/gui/tools/plot/plot_widget.py
```diff
@@ -165,64 +165,15 @@ def __init__(
...
-
     @Slot(bool)
     def _on_log_toggled(self, checked: bool) -> None:
-        ok = self._apply_log_state(checked)
-        if not ok:
-            self._log_checkbox.blockSignals(True)
-            self._log_checkbox.setChecked(False)
-            self._log_checkbox.blockSignals(False)
-        self._sync_log_checkbox()
+        self.updatePlotRequested.emit()
```
jonathan-eq: I don't think we need this method or this signal. Can we alias the `self._log_checkbox.toggled`-signal instead of having `updatePlotRequested`?

jonathan-eq (commented):
Look better now :)

tests/ert/ui_tests/gui/test_main_window.py
xjules: Not sure, it might be related to the qbot. However, `isVisibleTo(...)` works.

src/ert/gui/tools/plot/plot_widget.py
xjules: I think you will need to elaborate on this :) In essence we want to call `updatePlot(...)` of the parent window and this might be the smallest way to do it.

tests/ert/ui_tests/gui/test_main_window.py
```diff
@@ -371,15 +373,24 @@ def click_tab_index(pos: int) -> None:
             tab_center = tab_bar.tabRect(pos).center()
             qtbot.mouseClick(tab_bar, Qt.MouseButton.LeftButton, pos=tab_center)
 
+        def get_log_checkbox():
+            w = plot_window._central_tab.currentWidget()
+            return w.findChild(QCheckBox, "log_scale_checkbox")
```
jonathan-eq, line 378: I think we can get rid of this and instead just call 
```
return w.findChild(plot_window, "log_scale_checkbox")
```

PR #12072, opened 2025-10-15
Log specified and defaulted parameters for parameter configurations

SAKavli:
Issue
Resolves https://github.com/equinor/ert/issues/11196

Instead of logging this in each parameter class, we can instead log this once in the parent class ParameterConfig.

ExtParam did not need any tinkering as the constructor is just directly called is some places. Not through from_config_list.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/parameter_config.py
+++ b/src/ert/config/parameter_config.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 from abc import abstractmethod
 from collections.abc import Callable, Iterator
 from enum import StrEnum, auto
@@ -11,12 +12,14 @@ import networkx as nx
 import numpy as np
 import polars as pl
 import xarray as xr
-from pydantic import BaseModel
+from pydantic import BaseModel, model_validator
 
 if TYPE_CHECKING:
     import numpy.typing as npt
 
     from ert.storage import Ensemble
+
+logger = logging.getLogger(__name__)
 
 
 class InvalidParameterFile(Exception):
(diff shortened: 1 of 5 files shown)


src/ert/config/parameter_config.py
```diff
@@ -52,6 +55,18 @@ class ParameterConfig(BaseModel):
     forward_init: bool
     update: bool
 
+    @model_validator(mode="after")
+    def log_parameters_on_instantiation(self) -> ParameterConfig:
+        specified_parameters = self.model_fields_set
+        defaulted_parameters = set(self.model_fields.keys()) - specified_parameters
+        msg = (
+            f"{ParameterConfig.__name__}: {type(self).__name__} instance created\n"
```
xjules: Not sure but maybe `self.type` should be enough here?

tests/ert/unit_tests/config/test_parameter_config.py
```diff
@@ -0,0 +1,169 @@
...
+    )
+
+
+PARAMETER_LOG_PATTERN = (
+    r"Parameters with input values:\n\{([^}]+)\}\n"
+    r"Parameters with defaulted values:\n\{([^}]+)\}"
+)
+
+
+def test_parameters_are_logged_for_gen_kw_instances(caplog):
```
xjules: Could we add parameterize here for checking `update=True/False` is logged correctly or just maybe add two parameters in the test?

xjules (approved):
Nice work @SAKavli ! 🚀

PR #12076, opened 2025-10-16
Move EnsembleConfig.refcase to ErtConfig.refcase

yngve-sk:
EnsembleConfig does nothing but just hold a reference to `refcase`

### DIFF
--- a/tests/ert/unit_tests/config/test_ensemble_config.py
+++ b/tests/ert/unit_tests/config/test_ensemble_config.py
@@ -1,14 +1,10 @@
-from datetime import datetime
 from pathlib import Path
 from textwrap import dedent
 
 import pytest
-import xtgeo
-from resdata.summary import Summary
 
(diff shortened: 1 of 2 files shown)


src/ert/config/ert_config.py
```diff
@@ -713,14 +715,19 @@ class ErtConfig(BaseModel):
...
     def observations(self) -> dict[str, pl.DataFrame]:
         if self._observations is None:
             computed = create_observation_dataframes(
                 self.observation_declarations,
-                self.ensemble_config,
+                self.refcase,
+                cast(
+                    GenDataConfig | None,
+                    self.ensemble_config.response_configs.get("gen_data", None),
+                ),
```
andreas-el, line 730: Could extract this to a separate private function as this is used only within this file.

andreas-el (approved):
Fine work! You can decide yourself if you want to extract the function or not.
Either is fine. 🍖

Issue #11990, opened 2025-10-02
Wrong number of Parameters when using Design Matrix


flikka:
### What happened? (You can include a screenshot if it helps explain)

I have a design matrix with params a, b, c

In the GUI (e.g. ensemble experiment), it says 0

<img width="1404" height="825" alt="Image" src="https://github.com/user-attachments/assets/1d924631-5744-4cc0-a35c-015ef942cbb1" />

### What did you expect to happen?

That it says 3 (if I have three in the design matrix, and in general, the total number of parameters (from design matrix + distribution))

### Steps to reproduce

```Python
Make a case with design matrix
```

### Environment where bug has been observed

- [x] Python 3.11
- [ ] Python 3.12
- [x] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

oyvindeide:
Think this is working, was testing with v16.0, closing

PR #12079, opened 2025-10-16
Wrong number of Parameters when using Design Matrix

dafeda:
**Issue**
Resolves [#11990](https://github.com/equinor/ert/issues/11990)

Testing out whether a Claude Sonnet 4.5 agent can solve the issue #11990 that is labeled as "good first issue".

## Prompt used in VSCode Copilot Github Agent:

This is a description of an issue labeled as "good first issue".
Make an attempt solving it, and if you are unable to, try to improve the issue description so that it is easier to pick up.
There's a screen-shot attached to the issue, but I am unable to paste screenshots here, so you're going to have to do without.

I have a design matrix with params a, b, c

In the GUI (e.g. ensemble experiment), it says 0

What did you expect to happen?
That it says 3 (if I have three in the design matrix, and in general, the total number of parameters (from design matrix + distribution))

Steps to reproduce
Make a case with design matrix

## Screenshot

Here I open up the poly-case that uses the `DESIGN_MATRIX` keyword pointing to an excel file with 6 parameters, `SENSNAME`, `SENSCASE`, `RMS_SEED`, `a`, `b`, and `c`.

image

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/gui/ertwidgets/models/test_ertsummary.py
+++ b/tests/ert/unit_tests/gui/ertwidgets/models/test_ertsummary.py
@@ -135,10 +135,9 @@ def test_that_design_matrix_parameters_are_included_in_the_parameter_count(mock_
 
 
 @pytest.mark.usefixtures("use_tmpdir")
-def test_get_parameters_with_design_matrix_from_config(
+def test_design_matrix_parameters_counted_when_loaded_from_real_config(
     copy_poly_case_with_design_matrix,
 ):
-    """Test that design matrix parameters are counted when loaded from a real config"""
     # Create a design matrix with 3 parameters (a, b, c)
     design_dict = {"REAL": [0, 1, 2], "a": [1, 2, 3], "b": [4, 5, 6], "c": [7, 8, 9]}
     copy_poly_case_with_design_matrix(design_dict, [])


tests/ert/unit_tests/gui/ertwidgets/models/test_ertsummary.py
```diff
@@ -96,6 +100,59 @@ def test_getParameters(mock_ert):
     assert parameter_count == 223
 
 
+def test_getParameters_with_design_matrix(mock_ert):
```
eivindjahren: Should probably rename `getParameters` to snake_case rather than have mixed case here. The convention is to use snake_case for all function names and variable names.

tests/ert/unit_tests/gui/ertwidgets/models/test_ertsummary.py
```diff
@@ -96,6 +100,59 @@ def test_getParameters(mock_ert):
...
+    """Test that design matrix parameters are included in the parameter count"""
```
eivindjahren: This docstring should be the test name.

tests/ert/unit_tests/gui/ertwidgets/models/test_ertsummary.py
dafeda: The function its testing is called `getParameters` though.

eivindjahren: Yes, but might as well rename it while we are here.

dafeda: I actually think this is out of scope of the issue being fixed, but have pushed your suggestion anyway.

tests/ert/unit_tests/gui/ertwidgets/models/test_ertsummary.py
dafeda: Makes sense, done.

eivindjahren: No need for the docstring any more ;)

src/ert/gui/summarypanel.py
```diff
@@ -97,7 +97,7 @@ def updateSummary(self) -> None:
 
         self.addColumn(text.getText())
 
-        parameter_list, parameter_count = summary.getParameters()
+        parameter_list, parameter_count = summary.get_parameters()
```
eivindjahren, line 100: Forgot to rename the implementation :)

tests/ert/unit_tests/gui/ertwidgets/models/test_ertsummary.py
eivindjahren: Sure, but as a good default practice when implementing a new feature would be to make sure the existing code follows the coding conventions of the project. We default on the pep8 naming convention of `CamelCase` for classes and `snake_case` for function names. There are a myriad of other things we want from the code as well such as typing without blanket ignores and everything being Any, being well-tested, test naming convention, etc. that we have put into the pile "fix as you go" rather than trying to fix everything all at once.

We could try to automatically rename all functions to snake_case but its not high priority and kind of error-prone, so instead it is fixed whenever the code is otherwise updated. Its probably less error prone today than it was 5 years ago so if you feel like trying to update all names to follow pep8, its probably doable.

dafeda: Alright, that makes sense, thanks.

src/ert/gui/summarypanel.py
dafeda: Done 👍

tests/ert/unit_tests/gui/ertwidgets/models/test_ertsummary.py
```diff
@@ -91,11 +95,63 @@ def test_getForwardModels(mock_ert):
...
+    parameter_list, parameter_count = ErtSummary(mock_ert).get_parameters()
+
+    # Check that design matrix parameters are counted
+    assert "DESIGN_MATRIX (3)" in parameter_list
+    # Original count (223) + 3 design matrix parameters
+    assert parameter_count == 226
+
+
+@pytest.mark.usefixtures("use_tmpdir")
+def test_design_matrix_parameters_counted_when_loaded_from_real_config(
```
eivindjahren, line 138: A small thing, but this test name reads a bit better with a `that` inserted. Something like:

```diff
- def test_design_matrix_parameters_counted_when_loaded_from_real_config(
+ def test_that_design_matrix_parameters_are_counted_when_loaded_from_a_real_config(
```

PR #12081, opened 2025-10-16
Rewrite some queueconfig parsing

yngve-sk:
Towards circumventing ERT config dict from EverestRunModel


frode-aarstad (approved):
👍🏻

src/ert/config/queue_config.py
```diff
@@ -378,31 +399,18 @@ def from_dict(
         usr_queue_system = config_dict.get("QUEUE_SYSTEM")
         site_queue_system = site_queue_options.name if site_queue_options else None
 
-        config_dict["QUEUE_SYSTEM"] = selected_queue_system = QueueSystem(
+        selected_queue_system = QueueSystem(
```
andreas-el: Should we perhaps re-add the write-back so we have a complete config_dict in case someone logs or uses repr or similar?

PR #12084, opened 2025-10-16
Update param view docs

### LINKED ISSUE #11936: Implement a window showing all experiment parameters as a table (tree?) view

flikka:
**Issue**
Resolves #11936 

**Approach**
Forgot to add some updated screenshot and info when introducing the parameter viewer

      --exec 'just rapid-tests'`)

## When applicable


docs/ert/getting_started/configuration/poly_new/guide.rst
```diff
@@ -249,7 +249,8 @@ Creating parameters
 To sample different parameters across realizations in ERT,
 you'll need to define the prior distribution for each parameter.
 Furthermore, you'll detail how ERT can identify and inject these parameters into
-each simulation run via a templating mechanism.
+each simulation run via a templating mechanism. If you have parameters defined, you can inspect them
```
xjules, line 252: I think  you need to have parameters in order ert to run. 🤔

src/ert/gui/simulation/ensemble_experiment_panel.py
```diff
@@ -118,7 +118,9 @@ def __init__(
                 merged_parameters
             )
 
-        layout.addRow("Parameters:", get_parameters_button(merged_parameters, self))
+        if merged_parameters:
```
xjules, line 121: can this even happen?

docs/ert/getting_started/configuration/poly_new/guide.rst
xjules: It can run without params. So it is fine.

PR #12088, opened 2025-10-17
Simplify QueueConfig.from_dict logic

yngve-sk:
Defaults will anyway be applied when making the pydantic object

### DIFF
--- a/src/ert/config/queue_config.py
+++ b/src/ert/config/queue_config.py
@@ -407,44 +407,29 @@ class QueueConfig(BaseModelWithContextSupport):
             selected_queue_system, config_dict
         )
 
-        usr_job_script = usr_queue_options_dict.get("job_script")
-        site_job_script = site_queue_options_dict.get("job_script")
-
-        config_dict["JOB_SCRIPT"] = (
-            usr_job_script
-            or site_job_script
-            or shutil.which("fm_dispatch.py")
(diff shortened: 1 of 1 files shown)


src/ert/config/queue_config.py
```diff
@@ -407,44 +407,29 @@ def from_dict(
...
         )
 
-        usr_job_script = usr_queue_options_dict.get("job_script")
-        site_job_script = site_queue_options_dict.get("job_script")
-
-        config_dict["JOB_SCRIPT"] = (
-            usr_job_script
-            or site_job_script
-            or shutil.which("fm_dispatch.py")
-            or "fm_dispatch.py"
```
andreas-el, line 421: Will we not be missing these if usr_job_script and site_job_script are both None?

yngve-sk: Hmm this is maybe a bit subtle but it can never be explicit `None` in those dicts, it will be unset, so defaults apply when constructing it

job script from user config dict (usr_job_script) will either: Not exist in the usr dict, or not be None. Added some tests to clarify this. 

site_job_script comes from a `.model_dump(exclude_unset=True)` of the site config options, so if it is set to the default it'll just not be there. When we do `QueueOptions.create_queue_options` defaults are applied.

PR #12114, opened 2025-10-22
Have fm_dispatch remove event reporter after it shuts down

jonathan-eq:
**Issue**
Resolves **pew** **pew** 🔫  

**Approach**
If the event reporter's publisher zmq client gets a ClientConnectionError, it exits and no more messages are attempted delivered. The event_reporter.report(...) method is however still being called, and this causes a new log message `Event ... scheduled for delivery` every time. There is some
logic in fm_dispatch already to remove faulty reporters, and this commit makes the event reporter hit that code path.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/_ert/forward_model_runner/reporting/event.py
+++ b/src/_ert/forward_model_runner/reporting/event.py
@@ -137,6 +137,7 @@ class Event(Reporter):
                 return
             except ClientConnectionError as exc:
                 logger.error(f"Failed to send event: {exc}")
+                self._reporter_exception = exc
                 return
             except queue.Empty:
                 await asyncio.sleep(0)


src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -137,6 +137,7 @@ async def handle_publish(self, client: Client) -> None:
                 return
             except ClientConnectionError as exc:
                 logger.error(f"Failed to send event: {exc}")
+                self._reporter_exception = exc
```
xjules, line 140: But this can be a temporal drop out, so why to bring down reporter instead of reconnecting?

xjules: clarified via discusssion

jonathan-eq: Yes, we want to take down the reporter so it won't cause the fm_dispatcher to hang if mothership Ert goes down unexpectedly.

Issue #12065, opened 2025-10-14
Possible to open two ert instances towards the same storage


oyvindeide:
### What happened? (You can include a screenshot if it helps explain)

It is now possible to open two ert towards the same storage version, we seem to be ignoring the lock file.

### What did you expect to happen?

ert should give an error when opening two ert on the same path.

### Steps to reproduce

```console
Open ert with for example poly.ert
Open a new ert with the same poly.ert
```

### Environment where bug has been observed

- [ ] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

PR #12118, opened 2025-10-23
Fix plotter offering to select uninitialized ensembles

### LINKED ISSUE #12075: Plotting offers to select uninitialized ensembles leading to an error

When creating an empty experiment and unitilized ensemble, it shows it plotting window, which then shows an error:
```
 [ERROR] ert.dark_storage.endpoints.parameters: 'No SCALAR dataset in storage for ensemble ensemble'
```

Image

The ensemble shouldn't be listed in plotting window when being not initialized.

### Steps to reproduce

`ert gui poly.ert`
Manage experiments panel -> create new experiment -> create plot -> select ensemble

```Python

```

- [x] Python 3.11
- [x] Python 3.12
- [x] macOS
- [x] RHEL8
- [x] RHEL9
- [x] Local queue
- [x] LSF queue
- [x] Slurm queue
- [x] OpenPBS queue

jonathan-eq:
**Issue**
Resolves #12075

**Approach**
This commit adds an extra field to the ensemble object returned from the dark storage api - `realization_storage_states`. The value of this new field is a map between RealizationStorageStates and the number of realizations in the ensemble with this RealizationStorageState. We then use this to hide uninitialized ensembles from the ensemble selector in the plotter.
Example: '{"realization_storage_states": {
    "UNDEFINED": 6,
    "PARAMETERS_LOADED": 94
}}'

The dark storage endpoint now returns this object: 
```
{
  "size": 100,
  "active_realizations": [],
  "id": "d056f89d-99fd-421c-b34e-8f117e69d9e6",
  "experiment_id": "43ca3156-4dbc-43f7-ab08-8332e2dd55cf",
  "userdata": {
    "name": "ensemble",
    "experiment_name": "new_experiment",
    "started_at": "2025-10-23T09:55:17.473356"
  },
  "realization_storage_states": {
    "UNDEFINED": 6,
    "PARAMETERS_LOADED": 94
  }
}
```

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/dark_storage/endpoints/ensembles.py
+++ b/src/ert/dark_storage/endpoints/ensembles.py
@@ -1,4 +1,5 @@
 import logging
+from collections import Counter
 from uuid import UUID
 
 from fastapi import APIRouter, Body, Depends, HTTPException
@@ -38,4 +39,7 @@ def get_ensemble(
             "started_at": ensemble.started_at,
         },
         size=ensemble.ensemble_size,
+        realization_storage_states=Counter(
+            state for states in ensemble.get_ensemble_state() for state in states
+        ),
     )
--- a/src/ert/dark_storage/json_schema/ensemble.py
+++ b/src/ert/dark_storage/json_schema/ensemble.py
@@ -4,6 +4,8 @@ from uuid import UUID
 
 from pydantic import BaseModel
 
+from ert.storage.realization_storage_state import RealizationStorageState
+
 
 class _Ensemble(BaseModel):
     size: int
@@ -19,3 +21,4 @@ class EnsembleOut(_Ensemble):
     id: UUID
     experiment_id: UUID | None = None
     userdata: Mapping[str, Any]
+    realization_storage_states: Mapping[RealizationStorageState, int] | None = None
--- a/src/ert/gui/tools/plot/plot_api.py
+++ b/src/ert/gui/tools/plot/plot_api.py
@@ -20,6 +20,7 @@ from resfo_utilities import history_key
 
 from ert.config import ParameterMetadata, ResponseMetadata
 from ert.services import StorageService
+from ert.storage.realization_storage_state import RealizationStorageState
 
 logger = logging.getLogger(__name__)
 
@@ -92,18 +93,27 @@ class PlotApi:
                             f"/ensembles/{ensemble_id}", timeout=self._timeout
                         )
                         self._check_response(response)
-                        response_json = response.json()
+                        response_json: dict[str, Any] = response.json()
                         ensemble_name: str = response_json["userdata"]["name"]
                         experiment_name: str = response_json["userdata"][
                             "experiment_name"
                         ]
                         ensemble_started_at = response_json["userdata"]["started_at"]
+                        ensemble_undefined = False
+                        if realization_storage_states := response_json.get(
+                            "realization_storage_states"
+                        ):
+                            ensemble_undefined = (
+                                RealizationStorageState.PARAMETERS_LOADED
+                                not in set(realization_storage_states)
+                            )
                         self._all_ensembles.append(
                             EnsembleObject(
                                 name=ensemble_name,
                                 id=ensemble_id,
                                 experiment_name=experiment_name,
-                                hidden=ensemble_name.startswith("."),
+                                hidden=ensemble_name.startswith(".")
+                                or ensemble_undefined,
                                 started_at=ensemble_started_at,
                             )
                         )
--- a/tests/ert/unit_tests/gui/tools/plot/conftest.py
+++ b/tests/ert/unit_tests/gui/tools/plot/conftest.py
@@ -87,29 +87,47 @@ def mocked_requests_get(*args, **kwargs):
 
     ensemble = {
         "/ensembles/ens_id_1": {
-            "name": "ensemble_1",
-            "experiment_name": "experiment",
-            "started_at": "2012-12-10T00:00:00",
+            "userdata": {
+                "name": "ensemble_1",
+                "experiment_name": "experiment",
+                "started_at": "2012-12-10T00:00:00",
+            }
         },
         "/ensembles/ens_id_2": {
-            "name": ".ensemble_2",
-            "experiment_name": "experiment",
-            "started_at": "2012-12-10T01:00:00",
+            "userdata": {
+                "name": ".ensemble_2",
+                "experiment_name": "experiment",
+                "started_at": "2012-12-10T01:00:00",
+            }
         },
         "/ensembles/ens_id_3": {
-            "name": "default_0",
-            "experiment_name": "experiment",
-            "started_at": "2012-12-10T02:00:00",
+            "userdata": {
+                "name": "default_0",
+                "experiment_name": "experiment",
+                "started_at": "2012-12-10T02:00:00",
+            }
         },
         "/ensembles/ens_id_4": {
-            "name": "default_1",
-            "experiment_name": "experiment",
-            "started_at": "2012-12-10T03:00:00",
+            "userdata": {
+                "name": "default_1",
+                "experiment_name": "experiment",
+                "started_at": "2012-12-10T03:00:00",
+            }
         },
         "/ensembles/ens_id_5": {
-            "name": "default_manyobs",
-            "experiment_name": "experiment",
-            "started_at": "2012-12-10T04:00:00",
+            "userdata": {
+                "name": "default_manyobs",
+                "experiment_name": "experiment",
+                "started_at": "2012-12-10T04:00:00",
+            }
+        },
+        "/ensembles/ens_id_uninitialized": {
+            "userdata": {
+                "name": "uninitialized_ensemble",
+                "experiment_name": "experiment",
+                "started_at": "2012-12-10T04:00:00",
+            },
+            "realization_storage_states": {"UNDEFINED": 1},
         },
     }
 
@@ -316,6 +334,7 @@ def mocked_requests_get(*args, **kwargs):
                 "ens_id_3",
                 "ens_id_4",
                 "ens_id_5",
+                "ens_id_uninitialized",
             ],
             "parameters": {
                 "default": [
@@ -379,7 +398,7 @@ def mocked_requests_get(*args, **kwargs):
     ]
 
     if args[0] in ensemble:
-        return MockResponse({"userdata": ensemble[args[0]]}, 200)
+        return MockResponse(ensemble[args[0]], 200)
     elif args[0] in observations:
         return MockResponse(
             observations[args[0]],
--- a/tests/ert/unit_tests/gui/tools/plot/test_plot_api.py
+++ b/tests/ert/unit_tests/gui/tools/plot/test_plot_api.py
@@ -114,10 +114,11 @@ def test_case_structure(api):
         "default_0",
         "default_1",
         "default_manyobs",
+        "uninitialized_ensemble",
     ]
 
     assert ensembles == expected
-    assert hidden_case == [".ensemble_2"]
+    assert hidden_case == [".ensemble_2", "uninitialized_ensemble"]
 
 
 def test_can_load_data_and_observations(api):


src/ert/gui/tools/plot/plot_api.py
```diff
@@ -79,18 +80,26 @@ def get_all_ensembles(self) -> list[EnsembleObject]:
...
                             "experiment_name"
                         ]
                         ensemble_started_at = response_json["userdata"]["started_at"]
+                        ensemble_undefined = False
+                        if realization_storage_states := response_json.get(
+                            "realization_storage_states"
+                        ):
+                            ensemble_undefined = list(realization_storage_states) == [
+                                RealizationStorageState.UNDEFINED
+                            ]  # All realizations are UNDEFINED
```
jonathan-eq: This is done to avoid KeyError in the case of this bug https://github.com/equinor/ert/issues/12065

src/ert/storage/realization_storage_state.py
```diff
@@ -1,9 +1,9 @@
...
-    PARAMETERS_LOADED = 2
-    RESPONSES_LOADED = 4
-    FAILURE_IN_CURRENT = 8
-    FAILURE_IN_PARENT = 16
+class RealizationStorageState(StrEnum):
+    UNDEFINED = "UNDEFINED"
+    PARAMETERS_LOADED = "PARAMETERS_LOADED"
+    RESPONSES_LOADED = "RESPONSES_LOADED"
+    FAILURE_IN_CURRENT = "FAILURE_IN_CURRENT"
+    FAILURE_IN_PARENT = "FAILURE_IN_PARENT"
```
jonathan-eq: Does this require a storage migration somewhere? 
Having this as a normal enum made the api response very non readable 
```
{
  "size": 100,
  "active_realizations": [],
  "id": "d056f89d-99fd-421c-b34e-8f117e69d9e6",
  "experiment_id": "43ca3156-4dbc-43f7-ab08-8332e2dd55cf",
  "userdata": {
    "name": "ensemble",
    "experiment_name": "new_experiment",
    "started_at": "2025-10-23T09:55:17.473356"
  },
  "realization_storage_states": {
    "1": 6,
    "8": 94
  }
}
```

tests/ert/unit_tests/gui/tools/plot/conftest.py
```diff
@@ -87,29 +87,47 @@ def mocked_requests_get(*args, **kwargs):
...
+                "started_at": "2012-12-10T04:00:00",
+            }
+        },
+        "/ensembles/ens_id_uninitialized": {
+            "userdata": {
+                "name": "uninitialized_ensemble",
+                "experiment_name": "experiment",
+                "started_at": "2012-12-10T04:00:00",
+            },
+            "realization_storage_states": {"UNDEFINED": 1},
```
jonathan-eq, line 130: I had to change all of other ones to add `realization_storage_states` outside of `userdata`.

tests/ert/unit_tests/gui/tools/plot/test_plot_api.py
```diff
@@ -114,10 +114,11 @@ def test_case_structure(api):
         "default_0",
         "default_1",
         "default_manyobs",
+        "uninitialized_ensemble",
     ]
 
     assert ensembles == expected
-    assert hidden_case == [".ensemble_2"]
+    assert hidden_case == [".ensemble_2", "uninitialized_ensemble"]
```
jonathan-eq, line 121: ✅

src/ert/gui/tools/plot/plot_api.py
```diff
@@ -79,18 +80,26 @@ def get_all_ensembles(self) -> list[EnsembleObject]:
...
                         )
                         self._check_response(response)
-                        response_json = response.json()
+                        response_json: dict[str, Any] = response.json()
                         ensemble_name: str = response_json["userdata"]["name"]
                         experiment_name: str = response_json["userdata"][
                             "experiment_name"
                         ]
                         ensemble_started_at = response_json["userdata"]["started_at"]
+                        ensemble_undefined = False
```
xjules, line 102: here, ensemble_undefined is a bool while down it is a list.

src/ert/storage/realization_storage_state.py
```diff
@@ -1,9 +1,9 @@
...
+from enum import StrEnum
 
 
-class RealizationStorageState(Enum):
-    UNDEFINED = 1
-    PARAMETERS_LOADED = 2
-    RESPONSES_LOADED = 4
-    FAILURE_IN_CURRENT = 8
-    FAILURE_IN_PARENT = 16
+class RealizationStorageState(StrEnum):
```
xjules, line 4: `RealizationStorageState.UNDEFINED.name` should do the trick.

src/ert/gui/tools/plot/plot_api.py
jonathan-eq: No, it is the result of comparing two lists

src/ert/storage/realization_storage_state.py
jonathan-eq: Yes, lets do that

src/ert/storage/realization_storage_state.py
```diff
@@ -7,3 +8,12 @@ class RealizationStorageState(Enum):
     RESPONSES_LOADED = 4
     FAILURE_IN_CURRENT = 8
     FAILURE_IN_PARENT = 16
+
+
+RealizationStorageStateNames = Literal[
```
yngve-sk: Why not just make this a StrEnum instead of introducing the literal type below? I think old storage migrations had some dependency on this being int, but looks like it should be OK to remove now?

src/ert/dark_storage/json_schema/ensemble.py
```diff
@@ -19,3 +21,4 @@ class EnsembleOut(_Ensemble):
     id: UUID
     experiment_id: UUID | None = None
     userdata: Mapping[str, Any]
+    realization_storage_states: Mapping[RealizationStorageStateNames, int] | None = None
```
yngve-sk: Would opt to use the enum from storage directly (iff it can be converted to StrEnum), should still work with FastAPI the same as the literal type

src/ert/storage/realization_storage_state.py
jonathan-eq: We were not sure if we do some bitwise operations on this somewhere downstream, so we decided to leave it be

src/ert/dark_storage/json_schema/ensemble.py
xjules: Do we need `RealizationStorageStateNames` at all?

jonathan-eq: Yes, StrEnum would be better, but we would risk changing behavior for something elsewhere

src/ert/dark_storage/endpoints/ensembles.py
```diff
@@ -38,4 +39,7 @@ def get_ensemble(
             "started_at": ensemble.started_at,
         },
         size=ensemble.ensemble_size,
+        realization_storage_states=Counter(
```
yngve-sk, line 42: (nitpick/optional) naming preference, "realization states" instead of "realization storage states" but this depends on the already established RealizationStorageState name

src/ert/dark_storage/json_schema/ensemble.py
jonathan-eq: Typing, otherwise I would keep it as string

xjules: We have converted it back to normal Enum due to the underlying int specs. But why not just use `Mapping[str, int]`?

src/ert/dark_storage/endpoints/ensembles.py
jonathan-eq: We already use the concept of "RealizationState" in evaluator/scheduler/snapshot context, and we already have the RealizationStorageState available in LocalEnsemble. Maybe a more suitable name in the response would be better

src/ert/gui/tools/plot/plot_api.py
```diff
@@ -79,18 +80,26 @@ def get_all_ensembles(self) -> list[EnsembleObject]:
...
+                        if realization_storage_states := response_json.get(
+                            "realization_storage_states"
+                        ):
+                            ensemble_undefined = list(realization_storage_states) == [
```
yngve-sk: Suggest slight adjustment of logic. Now it will exclude only if all are undefined. Would include iff: there is at least one realization with parameters (or responses, but the responses will only be there if there are also parameters).

i.e., could check if there is at least one occurrence of `PARAMETERS_LOADED` in the list of states, and make it hidden if this check fails.

src/ert/dark_storage/json_schema/ensemble.py
jonathan-eq: I want the typing to be as specific as possible^

src/ert/gui/tools/plot/plot_api.py
jonathan-eq: We already do this. If there is at least one occurence of `PARAMETERS_LOADED` and the rest `UNDEFINED`, then `list(realization_storage_states)` would return `['UNDEFINED','PARAMETERS_LOADED']`, and the check would fail.

Should we change it to `
'PARAMETERS_LOADED' in list(realization_storage_states)`?

yngve-sk: Hmm to me that looks more readable, to express it as a criterion of inclusion instead of exclusion. Also suggest `set(realization_storage_states)` instead of list (functionally same, only for readability)

src/ert/dark_storage/endpoints/ensembles.py
yngve-sk: Hmm I see, totally fine to defer this to future PR(s) then.

src/ert/dark_storage/json_schema/ensemble.py
xjules: So what about this:
```python
def valid_storage_state(self) -> Self:
...
```

src/ert/gui/tools/plot/plot_api.py
```diff
@@ -79,18 +80,27 @@ def get_all_ensembles(self) -> list[EnsembleObject]:
...
+                                RealizationStorageState.PARAMETERS_LOADED
+                                not in set(realization_storage_states)
+                            )
                         self._all_ensembles.append(
                             EnsembleObject(
                                 name=ensemble_name,
                                 id=ensemble_id,
                                 experiment_name=experiment_name,
-                                hidden=ensemble_name.startswith("."),
+                                hidden=ensemble_name.startswith(".")
```
xjules, line 115: alright, we do support hidden ensembles then? How this come to be in the storage?

jonathan-eq: No idea, but it was already supported

src/ert/dark_storage/endpoints/ensembles.py
```diff
@@ -38,4 +39,7 @@ def get_ensemble(
...
+            state for states in ensemble.get_ensemble_state() for state in states
```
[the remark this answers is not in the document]
jonathan-eq: Not sure about this one

src/ert/gui/tools/plot/plot_api.py
```diff
@@ -92,18 +93,27 @@ def get_all_ensembles(self) -> list[EnsembleObject]:
...
                             "experiment_name"
                         ]
                         ensemble_started_at = response_json["userdata"]["started_at"]
+                        ensemble_undefined = False
+                        if realization_storage_states := response_json.get(
+                            "realization_storage_states"
+                        ):
+                            ensemble_undefined = (
+                                RealizationStorageState.PARAMETERS_LOADED
+                                not in set(realization_storage_states)
```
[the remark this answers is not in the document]
jonathan-eq: Don't think so, SIM118 Use `key not in dict` instead of `key not in dict.keys()`

src/ert/dark_storage/endpoints/ensembles.py
xjules, line 42: how do we make sure that the storage has been invalidated?

src/ert/gui/tools/plot/plot_api.py
xjules: We can remove the hidden "." feature and only use ensemble_undefined 👍

src/ert/dark_storage/endpoints/ensembles.py
jonathan-eq: What do you mean by invalidated?

xjules: That the state has been updated.

xjules (approved):
All resolved! Nice @jonathan-eq ! 🚀

PR #12119, opened 2025-10-23
Fix selecting a key in PlotWindow fetches same data multiple times

### LINKED ISSUE #12104: Selecting a key in the PlotWindow can lead to multiple data-loads via updatePlot
If you have an experiment / storage with different types of parameters, when selecting different types of data (e.g. changing from GEN_KW key to FIELD) we trigger "updatePlot" several times:
- Response -> GEN_KW, two times (one keyPressed, which triggers currentTabChanged - both of which individually triggers updatePlot).
- GEN_KW -> FIELD, three times, but not quite sure why. 
- FIELD -> GEN_KW, two times (same as the first)

Each results in calling down to the Rest API, which is an unnceccesary round-trip

jonathan-eq:
**Issue**
Resolves #12104

**Approach**
This commit fixes the issue by fixing the order of disconnecting/connecting the currentTabChanged signal in the `keySelected(...) method of PlotWindow. Before, we reconnected the signal before setting the widget manually, and this caused the signal to emit, and updating the plot twice.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/gui/tools/plot/plot_window.py
+++ b/src/ert/gui/tools/plot/plot_window.py
@@ -387,8 +387,6 @@ class PlotWindow(QMainWindow):
             self._central_tab.setTabEnabled(
                 self._central_tab.indexOf(plot_widget), plot_widget in available_widgets
             )
-        self._central_tab.currentChanged.connect(self.currentTabChanged)
-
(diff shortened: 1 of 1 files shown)


flikka (approved):
Small and neat!

PR #12140, opened 2025-10-27
Store all info about Everest objectives/constraints in ERT storage

yngve-sk:
**Issue**
prerequisite of removing EverestStorage


yngve-sk:
Doing one PR per obj/constraint instead
objectives: https://github.com/equinor/ert/pull/12248
constraints: TODO

PR #12144, opened 2025-10-27
Update Everest cli docs

### LINKED ISSUE #12127: Everest documentation: Move cli doc text to the source code where appropriate
The cli is implemented via `argparse` which supports documentation that is shown with the `--help` option. The manual includes this via an `argparse` directive, but also adds some extra text. Ideally this all resides with the `argparse` implementation in the source code.

Blocked by #12125

verveerpj:
**Issue**
Resolves #12127

**Approach**
- Clean up the argparse doc strings.
- Move and rewrite content from cli.rst to the argparse doc strings.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/bin/config_branch_script.py
+++ b/src/everest/bin/config_branch_script.py
@@ -28,23 +28,20 @@ def _build_args_parser() -> argparse.ArgumentParser:
     arg_parser = argparse.ArgumentParser(
         description=dedent(
             """
-            Create a new config file with updated controls.
+            Create a configuration file with updated controls based on a
+            previous run.
 
-            The *everest branch* command creates a new configuration file based
-            on the configuration file from a previous optimization run. The new
-            file is identical to the old configuration, but has updated control
-            values that are retrieved from the batch specified by the mandatory
-            `--batch` argument.
+            The *everest branch* command reads an existing configuration and
+            creates a new configuration file, with initial control values taken
+            from a batch of a previous optimization run.
 
-            **Warning:**
-            Do not remove the optimization output folder before running the
-            *branch* command or it will fail.
+            **Warning:** Do not remove the optimization output folder before
+            running the *branch* command or it will fail.
 
-            **Note:**
-            The *branch* command does not provide true restart functionality.
-            The new config file merely defines new initial controls defined by
-            the given batch. An optimization run based on the new configuration
-            will generate slightly different results.
+            **Note:** The *branch* command does not provide true restart
+            functionality. The new config file merely sets new initial controls
+            defined by the given batch. Subsequent optimization runs will likely
+            produce slightly different results.
             """
         ),
         formatter_class=ArgParseFormatter,
--- a/src/everest/bin/everest_script.py
+++ b/src/everest/bin/everest_script.py
@@ -106,7 +106,7 @@ def _build_args_parser() -> argparse.ArgumentParser:
     arg_parser = argparse.ArgumentParser(
         description=dedent(
             """
-            Start an "optimization run.
+            Start an optimization run.
 
             Closing the console or interrupting the `everest run` process does
             not terminate the optimization. To continue monitoring the running
--- a/src/everest/bin/main.py
+++ b/src/everest/bin/main.py
@@ -79,7 +79,7 @@ class EverestMain:
         )
 
     def export(self, args: list[str]) -> None:
-        """Deprecated."""
+        """Deprecated. Results are stored by default in the output directory."""
         everexport_entry(args)
 
     def lint(self, args: list[str]) -> None:


src/everest/bin/everest_script.py
```diff
@@ -102,16 +104,23 @@ def _build_args_parser() -> argparse.ArgumentParser:
     """Build arg parser"""
 
     arg_parser = argparse.ArgumentParser(
-        description=(
-            "Everest console runner, start an "
-            "optimization case based on a given config file"
+        description=dedent(
+            """
+            Start an "optimization run.
```
StephanDeHoop:
```suggestion
            Start an optimization run.
```

src/everest/bin/config_branch_script.py
```diff
@@ -23,23 +26,43 @@ def _yaml_config(
 
 def _build_args_parser() -> argparse.ArgumentParser:
     arg_parser = argparse.ArgumentParser(
-        description="Create new config file with updated controls "
-        "from specified simulation batch number\n"
-        "**Warning**: Previous simulation output folder will be overwritten"
-        " if it was placed outside the everest output folder",
+        description=dedent(
+            """
+            Create a new config file with updated controls.
```
StephanDeHoop:
```suggestion
            Create a new configuration file with updated controls based on a previous run.
```

src/everest/bin/config_branch_script.py
```diff
@@ -23,23 +26,43 @@ def _yaml_config(
...
+
+            The *everest branch* command creates a new configuration file based
```
StephanDeHoop: Maybe change paragraph to:
The 'everest branch' command reads an existing configuration and writes a new one with initial control values taken from the specified batch (i.e., required --batch argument) of a prior optimization.

src/everest/bin/config_branch_script.py
```diff
@@ -23,23 +26,43 @@ def _yaml_config(
...
+            file is identical to the old configuration, but has updated control
+            values that are retrieved from the batch specified by the mandatory
+            `--batch` argument.
+
+            **Warning:**
+            Do not remove the optimization output folder before running the
+            *branch* command or it will fail.
+
+            **Note:**
+            The *branch* command does not provide true restart functionality.
```
StephanDeHoop: Maybe change paragraph to:
This is not a full restart. The new config only sets initial controls from the given batch; subsequent optimization runs may diverge and produce slightly different results.

src/everest/bin/monitor_script.py
```diff
@@ -42,25 +44,35 @@ def _build_args_parser() -> argparse.ArgumentParser:
     """Build arg parser"""
 
     arg_parser = argparse.ArgumentParser(
-        description=(
-            "Everest console monitor a running optimization case based on a config file"
+        description=dedent(
+            """
+            Monitor a running optimization process.
+
+            Closing the console or interrupting the `everest run` process does
```
StephanDeHoop, line 51: I guess we cannot really get around the code-duplication here and in the description for `everest run`?

src/everest/bin/everconfigdump_script.py
```diff
@@ -18,7 +18,7 @@ def _build_args_parser() -> argparse.ArgumentParser:
         usage="""everest show <config_file>""",
     )
     arg_parser.add_argument(
-        "config_file", help="The path to the everest configuration file"
+        "config_file", help="The path to the everest configuration file."
```
StephanDeHoop, line 21: Maybe a nitpick, but as it appears in 8 different places, maybe we can change it to this?
```suggestion
        "config_file", help="Path to the Everest configuration file (YAML)."
```

src/everest/bin/main.py
```diff
@@ -51,52 +51,51 @@ def __init__(self, args: list[str]) -> None:
     @classmethod
     def methods_help(cls) -> str:
         """Return documentation of the public methods in this class"""
-        pubmets = [m for m in dir(cls) if not m.startswith("_")]
-        pubmets.remove("methods_help")  # Current method should not show up in desc
+        pubmets = ["run", "monitor", "kill", "lint", "render", "branch", "results"]
```
StephanDeHoop, line 54: I guess this is OK since we don't really add new methods/commands, but if we do we would have to add them here too (maybe easy to forget, maybe not?).

src/everest/bin/main.py
```diff
@@ -51,52 +51,51 @@ def __init__(self, args: list[str]) -> None:
...
         maxlen = max(len(m) for m in pubmets)
         docstrs = [getattr(cls, m).__doc__ for m in pubmets]
         doclist = [
-            m.ljust(maxlen + 1) + d for m, d in zip(pubmets, docstrs, strict=False)
+            m.ljust(maxlen + 2) + d for m, d in zip(pubmets, docstrs, strict=False)
```
StephanDeHoop, line 58: +2 because of --? Or something else?

src/everest/bin/main.py
```diff
@@ -51,52 +51,51 @@ def __init__(self, args: list[str]) -> None:
...
-        """Start the graphical user interface (Removed)"""
+        """Removed."""
         print(
             "The gui command has been removed. "
             "Please use the run command with the --gui option instead."
         )
 
     def export(self, args: list[str]) -> None:
-        """Export data from a completed optimization case"""
+        """Deprecated."""
```
StephanDeHoop: Maybe good to say what is going to replace it in the future (if replaced at all, and what work-around is suggested in the future if not replaced, or is this not used at all by anyone?)?

src/everest/bin/visualization_script.py
```diff
@@ -13,16 +14,32 @@
...
+
 
 def _build_args_parser() -> argparse.ArgumentParser:
     arg_parser = argparse.ArgumentParser(
-        description="Start possible plugin containing everest visualization",
+        description=dedent(
+            """
+            Start an everest visualization plugin.
+
+            If no visualization plugin is installed the message: ``No
```
StephanDeHoop, line 26: Just wondering, I guess we mainly use the `` ... `` such that in the actual docs it will be rendered better, but if someone does --help in the terminal, is there a better to suit both, or is this acceptable?

src/everest/bin/utils.py
```diff
@@ -475,3 +475,13 @@ def get_experiment_status(storage_dir: str) -> ExperimentStatus | None:
     with open_storage(storage_dir, "r") as storage:
         experiments = list(storage.experiments)
         return None if not experiments else experiments[0].status
+
+
+class ArgParseFormatter(argparse.HelpFormatter):
```
StephanDeHoop, line 480: Don't know much about `argparse.HelpFormatter`, but after Googling, is it a good idea to also immediately add: `argparse.ArgumentDefaultsHelpFormatter` for displaying defaults (or because we mainly have required keywords without defaults this is not really necessary)?

docs/everest/backends/index.rst
```diff
@@ -0,0 +1,15 @@
...
+Backends
+********
+
+.. toctree::
+    :hidden:
+
+    optimization
+    sampling
+
+The optimization functionality of Everest is determined by optimization backends that provide the low-level optimization algorithms. For details about the optimization backends that are included with Everest, see section :ref:`cha_optimization_backends`.
```
StephanDeHoop, line 11:
```suggestion
Everest's optimization capabilities are powered by modular optimization backends, which implement the underlying algorithms. For details on the backends included with Everest, see :ref:`cha_optimization_backends`.
```

docs/everest/backends/index.rst
```diff
@@ -0,0 +1,15 @@
...
+
+Similarly, the sampling methods that are used to generate perturbations for gradient calculation are provided by sampling backends, which are described in section: :ref:`cha_sampling_backends`.
```
StephanDeHoop, line 13:
```suggestion
Similarly, sampling backends are used to generate perturbations for gradient estimation during optimization. These are described in :ref:`cha_sampling_backends`.
```

docs/everest/backends/index.rst
```diff
@@ -0,0 +1,15 @@
...
+
+Additional backends may be provided with your local Everest installation, consult your systems manager for options.
```
StephanDeHoop, line 15:
```suggestion
Your local Everest installation may include additional backends. Please consult your system administrator for available options.
```

docs/everest/workflow_jobs.rst
StephanDeHoop, line 1: Isn't this file completely moved to forward_models --> custom.rst What is the purpose of keeping this?

docs/everest/forward_models/custom.rst
```diff
@@ -0,0 +1,39 @@
+
+.. _cha_creating_custom_jobs:
+
+
+Custom workflow jobs
+====================
+
+To use a custom workflow job in everest, the job needs to be added to the `install_workflow_jobs` section of the config file.
```
StephanDeHoop, line 8:
```suggestion
Everest supports custom workflow jobs that can be triggered before or after each batch of simulations. These jobs allow users to integrate external tools or scripts into the simulation workflow.

To define a custom job, add it to the `install_workflow_jobs` section of your configuration file:
```
and remove line 10?

docs/everest/forward_models/custom.rst
```diff
@@ -0,0 +1,39 @@
...
+      pre_simulation:
+        - job_name <job arguments>
+        - another_job <job arguments>
+      post_simulation:
+        - job_name <job arguments>
+        - another_job <job arguments>
+
+Currently `pre_simulation` and `post_simulation` triggers are defined, which run the specified jobs just before, and directly after, running each batch of simulations.
+
+For each batch evaluation, a runpath file is written containing a list of the simulation folders. The location of that file can be passed to a job using the pre-defined `runpath_file` variable that will be replaced with the full path to the runpath file. For example, this will pass the location of the runpath file as the first argument to a workflow job:
```
StephanDeHoop, line 33:
```suggestion
For each batch evaluation, Everest writes a runpath file containing the list of simulation folders. You can pass the location of this file to a job using the predefined `runpath_file` variable. For example:
```

docs/everest/forward_models/index.rst
```diff
@@ -0,0 +1,44 @@
...
+Forward model jobs
+******************
+
+.. toctree::
+    :hidden:
+
+    builtin
+    custom
+
+Some of the forward models are connected and often the output of one is the
```
StephanDeHoop, line 15: I know you didn't write this, but maybe change this whole paragraph to:
Some forward models in Everest are interconnected, with the output of one often serving as the input to another. However, they share a consistent data format, which allows certain jobs to be run independently when their dependencies are not required.

For example, if the `well_constraints` job is not needed, it can be omitted. Similarly, if `well_order` is not a concern, the input to `well_constraints` can be manually configured and reused in the Everest configuration file (avoiding the need to run `drill_planner` each time).

Note that some jobs are mandatory. For instance, `add_templates` must be executed before running the `schmerge` job, as it provides essential input data.

src/everest/bin/config_branch_script.py
verveerpj: Modified, somewhat along those lines.

src/everest/bin/config_branch_script.py
verveerpj: Do not fully agree, "full" is not the correct, it is not a restart at all, in the sense that it does continue in precisely the same way. Rewrite using some of your formulation as a suggestion.

src/everest/bin/monitor_script.py
verveerpj: Not really that would be a bit forced.

src/everest/bin/everconfigdump_script.py
verveerpj: I don't think we need to say here that is a yaml file. Users need to have some pre-knowledge of the format of the file anyway before they can start using the cli.

src/everest/bin/main.py
verveerpj: yes, but I wanted to put them in a specific order. Adding a new command means adding documentation, so I don't think it is bad.

src/everest/bin/main.py
verveerpj: Just some formatting, I found it sligthly more visually pleasing to have a bit of extra space between command and description.

src/everest/bin/main.py
verveerpj: It would not matter much since this string is not used in the cli or manual. However, I agree that is good to put something for documenation purposes, so I added something similar to what `everexport_script` does:

"Deprecated. Results are stored by default in the output directory."

src/everest/bin/main.py
StephanDeHoop: Makes sense, thanks!

src/everest/bin/visualization_script.py
verveerpj: I do not understand. Both the cli and the manual seem to render this fine.

src/everest/bin/utils.py
verveerpj: As far as I can tell we only use defaults to make optional flags, and these are shown as optional in the cli help. There does not seem to be a need to format defaults.

docs/everest/forward_models/index.rst
verveerpj: This will be rewritten in the context of another issue.

docs/everest/forward_models/custom.rst
verveerpj: Not using it, to be done in anohter issue.

docs/everest/backends/index.rst
verveerpj: Will not do this now, for an upcoming issue.

docs/everest/workflow_jobs.rst
verveerpj: no, these are workflow jobs, not the same as forward models. If there is any overlap, that will be resolved in a future issue.

src/everest/bin/visualization_script.py
StephanDeHoop: I did something wrong myself, please ignore this comment :)!

verveerpj:
@StephanDeHoop  I just noticed that somehow this PR got mixed up with another one, causing you to review things that do not belong there, and that will probably change. I hope I fixed this PR now, using your suggestions that were appropriate for the PR. 

Sorry for the confusion, you ended up doing some unnecessary reviewing work ...

docs/everest/backends/index.rst
verveerpj: For another issue.

docs/everest/backends/index.rst
verveerpj: Using this in the next issue.

PR #12146, opened 2025-10-27
Catch PermissionError in _load_index() and _load_ensembles()

### LINKED ISSUE #12089: Catch PermissionError when loading ensembles
In

https://github.com/equinor/ert/blob/99a6646102bd582a922cc4b61c2e93e76ec85aa3/src/ert/storage/local_storage.py#L218-L225

we should catch a PermissionError explicitly, as this is something that can happen and would propagate. Users have been observed to copy each others storages, and with varying correctness of permissions. 

When any of the files needed to load an ensemble (try with `chmod 000 ensembles/index.json`) and make sure an error is shown to the user, and then log the incident as a `logger.warning` or `logger.error` (It is not an exception when it is a permission error)

frode-aarstad:
**Issue**
Resolves #12089

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/storage/test_local_storage.py
+++ b/tests/ert/unit_tests/storage/test_local_storage.py
@@ -1283,37 +1283,34 @@ def test_that_permission_error_is_logged_in_load_ensembles(snake_oil_storage, ca
     ensemble = snake_oil_storage.get_experiment_by_name(
         "ensemble-experiment"
     ).get_ensemble_by_name("default_0")
-    path = Path(ensemble._path)
-    mode = path.stat().st_mode
-    os.chmod(path, 0o000)  # no permissions
+    Path(ensemble._path).chmod(0o000)  # no permissions
     snake_oil_storage._ensembles.clear()
     try:
         with caplog.at_level(logging.ERROR), pytest.raises(PermissionError):
             snake_oil_storage._load_ensembles()
             assert (
-                f"Permission error when loading ensemble from path: {path}."
+                f"Permission error when loading ensemble from path: {ensemble._path}."
                 in caplog.records[0].message
             )
             assert len(snake_oil_storage._ensembles) == 0
     finally:
-        os.chmod(path, mode)
+        Path(ensemble._path).chmod(0o755)  # restore permissions
 
 
 def test_that_permission_error_is_logged_in_load_index(snake_oil_storage, caplog):
     old_index = snake_oil_storage._index
-    path = Path(snake_oil_storage.path / "index.json")
-    mode = path.stat().st_mode
-    os.chmod(path, 0o000)  # no permissions
+    index_path = Path(snake_oil_storage.path / "index.json")
+    index_path.chmod(0o000)  # no permissions
     try:
         with caplog.at_level(logging.ERROR), pytest.raises(PermissionError):
             index = snake_oil_storage._load_index()
             assert index == old_index
             assert (
-                f"Permission error when loading index from path: {path}."
+                f"Permission error when loading index from path: {index_path}."
                 in caplog.records[0].message
             )
     finally:
-        os.chmod(path, mode)
+        index_path.chmod(0o744)  # restore permissions
 
 
 @dataclass


src/ert/dark_storage/common.py
```diff
@@ -15,7 +20,11 @@ def get_storage() -> Storage:
     if _storage is None:
         try:
             return (_storage := open_storage(os.environ["ERT_STORAGE_ENS_PATH"]))
+        except ErtStoragePermissionError as err:
+            logger.error(f"Permission error accessing storage: {err!s}")
+            raise InternalServerError("Permission error accessing storage") from None
```
andreas-el, line 25: Out of curiosity, why not propagate the error content as before?
I.e. `{err!s})` as part of the mesage?

src/ert/__main__.py
```diff
@@ -666,6 +666,9 @@ def main() -> None:
         with ErtPluginContext(logger=logging.getLogger()) as runtime_plugins:
             logger.info(f"Running ert with {args} in {os.getcwd()}")
             args.func(args, runtime_plugins)
+    except ErtStoragePermissionError as err:
+        logger.error(str(err))
+        sys.exit(str(err))
```
andreas-el, line 671: This will happen very early on, so it does not make sense to update the span?
Other blocks include;
```
        span.set_status(Status(StatusCode.ERROR))
        span.record_exception(err)
```

src/ert/dark_storage/common.py
```diff
@@ -15,7 +20,11 @@ def get_storage() -> Storage:
     if _storage is None:
         try:
             return (_storage := open_storage(os.environ["ERT_STORAGE_ENS_PATH"]))
+        except ErtStoragePermissionError as err:
+            logger.error(f"Permission error accessing storage: {err!s}")
```
andreas-el, line 24: Is there a reason for using logger.error for this exception, and logger.exception for the one below?

src/ert/storage/__init__.py
```diff
@@ -32,11 +32,20 @@ class ErtStorageException(Exception):
...
 def open_storage(
     path: str | os.PathLike[str], mode: ModeLiteral | Mode = "r"
 ) -> Storage:
     try:
         return LocalStorage(Path(path), Mode(mode))
+    except PermissionError as err:
+        raise ErtStoragePermissionError(
+            "Permission error when accessing storage at: "
+            f"{path} with mode: '{mode}'. Error: {err}"
+        ) from err
```
andreas-el, line 48: Codecov complains this is not covered. Would it be easy to have a test for this?

src/ert/dark_storage/common.py
andreas-el, line 24: Would it be possible to test this `ErtStoragePermissionError` since we add this in the PR?

src/ert/dark_storage/common.py
frode-aarstad: In "public" APIs its not really recommended add details of the internal architecture in return messages

src/ert/__main__.py
frode-aarstad: A requirement of this issue was not to propagate this type of PermissionError to azure

src/ert/dark_storage/common.py
frode-aarstad: Requested by @berland in the issue

tests/ert/unit_tests/storage/test_local_storage.py
```diff
@@ -340,6 +341,26 @@ def test_open_empty_write(tmp_path):
     assert (tmp_path / "empty").is_dir()
 
 
+def test_open_with_no_permissions(tmp_path):
+    with open_storage(tmp_path, mode="w") as storage:
+        experiment = storage.create_experiment()
+        ensemble = storage.create_ensemble(experiment, name="foo", ensemble_size=1)
+    path = Path(ensemble._path)
+    mode = path.stat().st_mode
+    os.chmod(path, 0o000)  # no permissions
```
berland, line 350: `path` is a `pathlib` object, then you  can do `path.chmod(0o000)` directly

tests/ert/unit_tests/storage/test_local_storage.py
```diff
@@ -340,6 +341,26 @@ def test_open_empty_write(tmp_path):
     assert (tmp_path / "empty").is_dir()
 
 
+def test_open_with_no_permissions(tmp_path):
+    with open_storage(tmp_path, mode="w") as storage:
+        experiment = storage.create_experiment()
+        ensemble = storage.create_ensemble(experiment, name="foo", ensemble_size=1)
+    path = Path(ensemble._path)
+    mode = path.stat().st_mode
```
berland, line 349: maybe call this `original_permissions` or `original_mode` or something as it is only used at the end (it could probably also be dropped, and then just do `chmod(0o700)` at the end, as all we need is deletability.

PR #12154, opened 2025-10-27
Avoid plotter if plot api versions differ

### LINKED ISSUE #12037: KeyError: 'started_at' in plot_api.py due to API "mismatch"

We observed an exception `'started_at' - in get_all_ensembles ensemble_started_at = response_json["userdata"]["started_at"] ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^ KeyError: 'started_at' `

Triggered by plotting functionality in ERT, location of the thing: https://github.com/equinor/ert/blob/66bb8133fb57817b95f5e27e4da7ad32db920301/src/ert/gui/tools/plot/plot_api.py#L87

No exception

### Steps to reproduce

```Python
Run ERT version 2025.09 - run an experiment
Start (while the other is running) ERT 2025.10

Open Plotting (in 2025.10). Observe failure.

The problem is that the newest ERT exposes and uses a new property from the "API". When the new ert starts up, it will _NOT_ start a new api / service, since one is already running (checks the file storage/storage_server.json). But this OLD server does not have this property.

So the question is: Should we be "backwards compatible" with old version of the API. I suggest NO: Lets rather check, when starting ert, that the "ERT" version of the API (add an endpoint) is the same as the mothership ERT.
```

- [x] Python 3.11
- [x] Python 3.12
- [x] macOS
- [x] RHEL8
- [x] RHEL9
- [x] Local queue
- [x] LSF queue
- [x] Slurm queue
- [x] OpenPBS queue

andreas-el:
**Issue**
Resolves #12037 

⚠️ Note that I conducted recording altering the comparison function from `!=` to `==`, hence identical versions reported incompatible.

**Updated 2025-10-29:**
Screenshot 2025-10-29 at 10 30 34

https://github.com/user-attachments/assets/00da80dd-3d41-4d91-a48b-704bc736a25e

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/dark_storage/endpoints/version.py
+++ b/src/ert/dark_storage/endpoints/version.py
@@ -1,0 +1,16 @@
+import logging
+
+from fastapi import APIRouter, Body, Depends
+
+from ert.dark_storage.common import get_storage, get_storage_api_version
+
+router = APIRouter(tags=["version"])
+logger = logging.getLogger(__name__)
+
+DEFAULT_STORAGE = Depends(get_storage)
+DEFAULT_BODY = Body(...)
+
+
+@router.get("/version")
+def get_version() -> str:
+    return get_storage_api_version()
--- a/src/ert/gui/main_window.py
+++ b/src/ert/gui/main_window.py
@@ -81,8 +81,7 @@ class SidebarToolButton(QToolButton):
         if event:
             if event.button() == Qt.MouseButton.RightButton:
                 self.right_clicked.emit()
-            else:
-                super().mousePressEvent(event)
+            super().mousePressEvent(event)
 
 
 class ErtMainWindow(QMainWindow):
--- a/src/ert/gui/tools/plot/plot_api.py
+++ b/src/ert/gui/tools/plot/plot_api.py
@@ -53,6 +53,19 @@ class PlotApi:
         self._all_ensembles: list[EnsembleObject] | None = None
         self._timeout = 120
 
+    @property
+    def api_version(self) -> str:
+        with StorageService.session(project=self.ens_path) as client:
+            try:
+                response = client.get("/version", timeout=self._timeout)
+                self._check_response(response)
+                api_version = str(response.json())
+            except Exception as exc:
+                logger.exception(exc)
+                raise exc
+            else:
+                return api_version
+
     @staticmethod
     def escape(s: str) -> str:
         return quote(quote(s, safe=""))
--- a/src/ert/dark_storage/common.py
+++ b/src/ert/dark_storage/common.py
@@ -1,5 +1,7 @@
 import logging
 import os
+import re
+from importlib import metadata
 
 from ert.dark_storage.exceptions import InternalServerError
 from ert.storage import (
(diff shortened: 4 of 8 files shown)


src/ert/dark_storage/endpoints/version.py
```diff
@@ -0,0 +1,16 @@
...
+
+router = APIRouter(tags=["version"])
+logger = logging.getLogger(__name__)
+
+DEFAULT_STORAGE = Depends(get_storage)
+DEFAULT_BODY = Body(...)
+
+
+@router.get("/version")
+def get_running_storage_version() -> str:
```
andreas-el: Perhaps find a better name

src/ert/gui/main_window.py
```diff
@@ -165,14 +166,51 @@ def __init__(
         self.__add_tools_menu()
         self.__add_help_menu()
 
+        error_topic = "Plot API version mismatch detected"
+        self.msg_box = QMessageBox()
+        self.msg_box.setIcon(QMessageBox.Icon.Critical)
+        self.msg_box.setWindowTitle(error_topic)
+        self.msg_box.setStandardButtons(QMessageBox.StandardButton.Ok)
+        self.msg_box.setModal(False)
```
andreas-el: EventHandling and Modal boxes interfere with one another.
Using non-modal should not be a problem as the plotting is still denied regardless of modality.

src/ert/dark_storage/endpoints/version.py
frode-aarstad: why not just get_version?

src/ert/gui/main_window.py
frode-aarstad: Is this event handling a problem when running a simulation? And does this work when running a simulation?

src/ert/gui/tools/plot/plot_api.py
```diff
@@ -52,6 +52,15 @@ def __init__(self, ens_path: Path) -> None:
         self.ens_path = ens_path
         self._all_ensembles: list[EnsembleObject] | None = None
         self._timeout = 120
+        self._api_version = ""
+
+    @property
+    def api_version(self) -> str:
+        with StorageService.session(project=self.ens_path) as client:
+            response = client.get("/version", timeout=self._timeout)
+            self._check_response(response)
```
frode-aarstad: what if check_response raises an exception?

src/ert/gui/main_window.py
andreas-el: It won't interfere with the simulation. I tested both with version match and mismatch.

The event handling issue affects the mouse release signal when modal boxes are used. All buttons become unresponsive until this signal is dispatched.

src/ert/gui/tools/plot/plot_api.py
andreas-el: Good question, most other similar access patterns does not have any exception handling. I've added a try-except block to at least log.

src/ert/gui/main_window.py
andreas-el: This is totally rewritten in the new proposal.

PR #12160, opened 2025-10-28
Revise plugin manager usage

yngve-sk:
(Requires updating Everest models: https://github.com/equinor/everest-models/pull/189 )
*** Intentionally leaving EverestRunModel.create out of this, but it should need one kind of plugin context.

ERT plugins pertain to the packages installed in the python, these are not subject to change during runtime. Passing in custom plugins is only for testing.

### DIFF
--- a/src/ert/plugins/plugin_manager.py
+++ b/src/ert/plugins/plugin_manager.py
@@ -380,92 +380,82 @@ class ErtRuntimePlugins(BaseModel):
     help_links: dict[str, str] = Field(default_factory=dict)
 
 
-class ErtPluginContext:
-    def __init__(
-        self,
-        plugins: list[object] | None = None,
-        logger: logging.Logger | None = None,
-    ) -> None:
-        self.plugin_manager = ErtPluginManager(plugins=plugins)
-        self._logger = logger
+def get_site_plugins(
+    plugin_manager: ErtPluginManager | None = None,
+) -> ErtRuntimePlugins:
+    if plugin_manager is None:
+        plugin_manager = ErtPluginManager()
 
-    @staticmethod
-    def get_site_plugins(
-        plugin_manager: ErtPluginManager | None = None,
-    ) -> ErtRuntimePlugins:
-        if plugin_manager is None:
-            plugin_manager = ErtPluginManager()
+    site_configurations = plugin_manager.get_site_configurations()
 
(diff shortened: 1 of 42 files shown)


src/ert/plugins/plugin_manager.py
```diff
@@ -477,9 +477,11 @@ def get_site_plugins(
         return runtime_plugins
 
     @staticmethod
-    def setup_logging(root_logger: logging.Logger | None) -> None:
+    def setup_logging(root_logger: logging.Logger | None = None) -> None:
         pm = ErtPluginManager.get_default()
-        pm.add_logging_handle_to_root(logger=root_logger)
+        if root_logger is not None:
+            pm.add_logging_handle_to_root(logger=root_logger)
+
```
oyvindeide: Not related to the rename in the commit?

src/ert/plugins/plugin_manager.py
```diff
@@ -52,6 +52,14 @@ class JobDoc(TypedDict):
 
 
 class ErtPluginManager(pluggy.PluginManager):
+    _cached_default: ErtPluginManager | None = None
+
+    @classmethod
+    def get_default(cls) -> ErtPluginManager:
+        if cls._cached_default is None:
+            cls._cached_default = cls()
+        return cls._cached_default
```
oyvindeide: If we want it to be a singleton, it should probably be an actual singleton?

src/ert/plugins/plugin_manager.py
oyvindeide: This should probably be renamed as it is no longer a context

src/ert/plugins/plugin_manager.py
yngve-sk: Fixed

src/ert/plugins/plugin_manager.py
yngve-sk: Tried but leads to circular imports, defer to later / separate issue.

src/ert/plugins/plugin_manager.py
yngve-sk: Renamed, open to other suggestions, but I think something with "site" makes sense, and it is not purely plugins since we also use it to pass logging handles to external plugins.

yngve-sk: Edit: Removed the namespace class

PR #12164, opened 2025-10-28
Split runmodels into config/runtime

yngve-sk:
This is mainly just moving stuff
(Split out from https://github.com/equinor/ert/pull/11513)
Towards serializing/deserializing runmodels without any state.

(cherry picked from commit eb6e27a99035fc437f1f11a7db4def42bcd5ef4c)

### DIFF
--- a/src/ert/storage/local_experiment.py
+++ b/src/ert/storage/local_experiment.py
@@ -23,9 +23,7 @@ from ert.config import (
     ResponseConfig,
     SurfaceConfig,
 )
-from ert.config import (
-    Field as FieldConfig,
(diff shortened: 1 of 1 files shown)


src/ert/storage/local_experiment.py
```diff
@@ -23,7 +23,9 @@
     ResponseConfig,
     SurfaceConfig,
 )
-from ert.config import Field as FieldConfig
+from ert.config import (
+    Field as FieldConfig,
```
berland: rm comma to allow reformatting back to original.

PR #12165, opened 2025-10-29
Add tests for creation of the Everest runmodel

berland:
No changes to code, but this is needed for upcoming refactoring. Partial overlap with what is tested in test_res_initialization.py

**Issue**
Lack of testing of `EverestRunModel.create()`.

**Approach**
🧑‍🏭 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/everest/test_everest_run_model.py
+++ b/tests/everest/test_everest_run_model.py
@@ -1,0 +1,125 @@
+from collections.abc import Callable
+from unittest.mock import Mock
+
+import pytest
+
+from ert.config.queue_config import (
+    LocalQueueOptions,
+    LsfQueueOptions,
+    QueueOptions,
+    SlurmQueueOptions,
+    TorqueQueueOptions,
+)
+from ert.plugins import ErtPluginContext
+from ert.run_models.everest_run_model import EverestRunModel
+from everest.config import EverestConfig
+
+
+@pytest.fixture
+def create_runmodel(min_config: dict, monkeypatch: pytest.MonkeyPatch) -> Callable:
+    monkeypatch.setattr("ert.run_models.run_model.open_storage", Mock())
+
+    def _create_runmodel(
+        queue_system: dict[str, str | int | bool | float],
+    ) -> EverestRunModel:
+        with ErtPluginContext() as runtime_plugins:
+            return EverestRunModel.create(
+                EverestConfig(
+                    **(min_config | {"simulator": {"queue_system": queue_system}})
+                ),
+                runtime_plugins=runtime_plugins,
+            )
+
+    return _create_runmodel
+
+
+@pytest.mark.parametrize("queue_system", ["lsf", "local", "torque", "slurm"])
+def test_that_queue_system_name_passes_through_create(
+    create_runmodel: Callable, queue_system: str
+) -> None:
+    runmodel = create_runmodel({"name": queue_system})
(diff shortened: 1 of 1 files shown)


tests/everest/test_everest_runmodel.py
```diff
@@ -0,0 +1,150 @@
...
+        simulator = {}
+    if "name" not in queue_system:
+        queue_system["name"] = "local"
+    with ErtPluginContext() as runtime_plugins:
+        return EverestRunModel.create(
+            EverestConfig(
+                **(
+                    minimal_config
+                    | {"simulator": simulator}
+                    | {"simulator": {"queue_system": queue_system}}
```
yngve-sk: won't this override the above statement?
ref
```
>>> {"a": 2} | {"a": { "b": 3}} | {"a": {"c": 3}}
{'a': {'c': 3}}
```

tests/everest/test_everest_runmodel.py
```diff
@@ -0,0 +1,150 @@
...
+    LsfQueueOptions,
+    QueueOptions,
+    SlurmQueueOptions,
+    TorqueQueueOptions,
+)
+from ert.plugins import ErtPluginContext
+from ert.run_models.everest_run_model import EverestRunModel
+from everest.config import EverestConfig
+
+minimal_config = {
```
yngve-sk: Do we need this instead of the `min_config` fixture?

berland: We don't!

tests/everest/test_everest_runmodel.py
berland: Fixed, this was a remnant from earlier versions of this file.

tests/everest/test_everest_run_model.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def _create_runmodel(
+    min_config: dict,
+    queue_system: dict[str, str | int | bool | float] | None = None,
+) -> EverestRunModel:
+    with ErtPluginContext() as runtime_plugins:
+        return EverestRunModel.create(
+            EverestConfig(
+                **(min_config | {"simulator": {"queue_system": queue_system}})
```
yngve-sk: Works now, but I tried rebasing on it without giving a queue_system, then I need this:
```
**(min_config | ({"simulator": {"queue_system": queue_system}}) if queue_system is not None else {})
```
So either make it non-optional or support `None` here

berland: removed None.

tests/everest/test_everest_run_model.py
```diff
@@ -0,0 +1,124 @@
...
+        ],
+    ],
+)
+def test_queue_options_properties_pass_through_create(
+    min_config: dict,
+    config: dict[str, str | int | float | bool],
+    config_class: QueueOptions,
+    monkeypatch: pytest.MonkeyPatch,
+):
+    monkeypatch.setattr("ert.run_models.run_model.open_storage", Mock())
```
yngve-sk: This now has to go with every use of _create_runmodel if we create more of them, or if we do create only one + use tmpdir, a bit implicit. could perhaps move the mock to _create_runmodel?

berland: Will try to make _create_runmodel into a fixture so it can do the patching.

PR #12169, opened 2025-10-29
Remove unneeded "ensemble size" number argument to panels

### LINKED ISSUE #12059: Replace ensemble_size with len(active_realizations mask) in GUI panels
Continuation of #10923
Ensemble size is just the length of the active_realizations mask, and it is not neccessary to pass both to both the different gui panels, and for the validators

flikka:
**Issue**
Resolves #12059 

**Approach**
Removing the "ensemble_size" from evaluate_ensemble_panel and manual_update_panel.
Some other changes were needed to get the size from elsewhere.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/gui/simulation/evaluate_ensemble_panel.py
+++ b/src/ert/gui/simulation/evaluate_ensemble_panel.py
@@ -32,9 +32,7 @@ class Arguments:
 
 
 class EvaluateEnsemblePanel(ExperimentConfigPanel):
-    def __init__(
-        self, ensemble_size: int, run_path: str, notifier: ErtNotifier
-    ) -> None:
+    def __init__(self, run_path: str, notifier: ErtNotifier) -> None:
         self.notifier = notifier
         super().__init__(EvaluateEnsemble)
         self.setObjectName("Evaluate_parameters_panel")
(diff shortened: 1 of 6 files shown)


src/ert/gui/simulation/evaluate_ensemble_panel.py
```diff
@@ -49,16 +47,22 @@ def __init__(
...
+        else:
+            ensemble_size = "No ensemble selected"
+
+        self._number_of_realizations_label = QLabel(f"<b>{ensemble_size}</b>")
+        layout.addRow(
+            QLabel("Number of realizations:"), self._number_of_realizations_label
+        )
 
         self._active_realizations_field = StringBox(
             ActiveRealizationsModel(ensemble_size, show_default=False),  # type: ignore
```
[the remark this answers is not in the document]
flikka: I'll fix it differently

src/ert/validation/ensemble_realizations_argument.py
```diff
@@ -17,11 +17,13 @@ class EnsembleRealizationsArgument(RangeStringArgument):
...
         ensemble: Callable[[], Union["Ensemble", None]],
-        max_value: int | None,
         required_realization_storage_states: Iterable["RealizationStorageState"],
         **kwargs: bool,
     ) -> None:
-        super().__init__(max_value, **kwargs)
+        ens = ensemble()
+        size = ens.ensemble_size if ens is not None else None
+        super().__init__(size, **kwargs)
+
```
jonathan-eq, line 26: Does it work to have this logic only in the constructor, or do we need it in the validate method too?

flikka: Yes, I believe so. This is to handle the case where we dont have any exisiting experiments, and the validator is initialized without ensemble. If the validate is called without ensemble set (it should not happen, because the text field is not editable then) - but if it happens, yes, then we alread have an assert not none for that (for typing probably).

PR #12171, opened 2025-10-29
Initialize substitutions without going through ERT config dict

yngve-sk:


src/ert/run_models/everest_run_model.py
```diff
@@ -372,12 +371,16 @@ def create(
...
-        substitutions = get_substitutions(
-            config_dict,
-            runpath_config,
-            runpath_file,
-            queue_config.queue_options.num_cpu,
-        )
+        substitutions = {
+            "<RUNPATH_FILE>": str(runpath_file),
+            "<RUNPATH>": runpath_config.runpath_format_string,
+            "<ECL_BASE>": runpath_config.eclbase_format_string,
```
berland, line 377: I perturbed these lines slightly (mixing the dict values, and/or removing some) and no errors will be caught by `just rapid-tests`.  Some errors seemed to got caught doing `just everest-tests` (which takes a lot of time to run). Could you have a look if some unit-tests should be added?

tests/everest/test_everest_run_model.py
```diff
@@ -123,3 +140,36 @@ def test_queue_options_properties_pass_through_create(
 ) -> None:
     runmodel = create_runmodel(config)
     assert runmodel.queue_config.queue_options == config_class(**config)
+
+
+def test_substitutions_from_everest_config(
+    monkeypatch: pytest.MonkeyPatch, tmp_path: Path, create_runmodel: Callable
+):
```
berland: (if we ever activate `ruff --select ANN` you need `-> None`  here )

yngve-sk: fixd

tests/everest/test_everest_run_model.py
```diff
@@ -123,3 +140,36 @@ def test_queue_options_properties_pass_through_create(
...
     assert runmodel.queue_config.queue_options == config_class(**config)
+
+
+def test_substitutions_from_everest_config(
+    monkeypatch: pytest.MonkeyPatch, tmp_path: Path, create_runmodel: Callable
+) -> None:
+    monkeypatch.chdir(tmp_path)
+    config_path = Path("./hello/world/strong_optimizer.yml")
+    config_dir = config_path.parent
+    os.makedirs(config_path.parent)
```
berland: Here you can do `config.path.parent.mkdir()` for consistency in pathlib usage (and then the `import os` can be dropped).

tests/everest/test_everest_run_model.py
```diff
@@ -125,5 +137,50 @@ def test_queue_options_properties_pass_through_create(
...
+        "<ECL_BASE>": "ECLBASE<IENS>",
+        "<ECLBASE>": "ECLBASE<IENS>",
+        "<NUM_CPU>": "1337",
+        "<CONFIG_PATH>": str(config_dir),
+        "<CONFIG_FILE>": "strong_optimizer",
+    }
+
+
+@pytest.mark.parametrize("random_seed", [None, 1234])
+def test_random_seed(create_runmodel: Callable, random_seed: int | None) -> None:
```
berland: Consider making the function name more precise. I chose `passes through` as a kind of description of what we want the create function to do for the parameters (unless it is doing something more fancy)

yngve-sk: It was moved, not introduced, so I did not rename it. Will add a separate rename commit

tests/everest/test_everest_run_model.py
```diff
@@ -175,7 +175,9 @@ def test_substitutions_from_everest_config(
 
 
 @pytest.mark.parametrize("random_seed", [None, 1234])
-def test_random_seed(create_runmodel: Callable, random_seed: int | None) -> None:
+def test_that_random_seed_pass_through_create(
```
berland: nitpick: pass -> passes

PR #12174, opened 2025-10-30
Avoid queue_options from plugins when testing

berland:
Otherwise the plugins can fail the tests

**Issue**
Resolves Komodo bleeding test failure

**Approach**
Modify the plugin context directly in the test. This object is pr now not a singleton object but that might change later?

The `no_plugin` fixture did not help, as it does not patch the correct places for this test. Considered the proposed change as fine as a fix to the `no_plugins` fixture.

      --exec 'just rapid-tests'`)

## When applicable


jonathan-eq (approved):
Tested this manually, and it works :)
I am just a bit curious as to why we even use plugins in the tests to begin with 🤔

PR #12202, opened 2025-10-31
Rename ertbox -> grid_geometry

oyvindeide:
**Issue**
Resolves #my_issue

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/field.py
+++ b/src/ert/config/field.py
@@ -14,10 +14,10 @@ import xtgeo
 from pydantic import field_serializer
 
 from ert.field_utils import (
-    ErtboxParameters,
     FieldFileFormat,
+    GridGeometry,
     Shape,
-    calculate_ertbox_parameters,
+    calculate_grid_geometry,
     get_shape,
     read_field,
     save_field,
@@ -68,7 +68,7 @@ def create_flattened_cube_graph(px: int, py: int, pz: int) -> nx.Graph[int]:
 class Field(ParameterConfig):
     type: Literal["field"] = "field"
     dimensionality: Literal[3] = 3
-    ertbox_params: ErtboxParameters
+    grid_geometry: GridGeometry
     file_format: FieldFileFormat
     output_transformation: str | None = None
     input_transformation: str | None = None
@@ -165,7 +165,7 @@ class Field(ParameterConfig):
         try:
             if grid_extension == ".egrid":
                 grid = xtgeo.grid_from_file(grid_file_path)
-                ertbox_params = calculate_ertbox_parameters(grid)
+                grid_geometry = calculate_grid_geometry(grid)
             else:
                 dims = get_shape(grid_file_path)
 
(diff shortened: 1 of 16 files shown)


src/ert/storage/migration/to16.py
```diff
@@ -0,0 +1,30 @@
...
+from typing import Any
+
+info = "Rename ertbox to grid dimensions"
+
+
+def migrate_field_param(parameters_json: dict[str, Any]) -> dict[str, Any]:
+    new_configs = {}
+    for param_config in parameters_json.values():
+        if param_config["type"] == "field":
+            param_config["grid_dimensions"] = param_config.pop("ertbox_params")
```
xjules: I guess this should be `param_config.pop("ertbox_params", None)`?

oddvarlia:
A comment: I am not sure renaming ertbox to grid_dimensions is a good approach since this contains more than grid dimensions. The `class ErtboxParameters` contains not only grid dimensions, but equally important a definition of a local coordinate system used for calculating the transformation of observations positions into the local grid coordinate system.

oddvarlia (changes requested):
I am not sure we should rename this to grid_dimensions since the `class ErtboxParameters` not only contains grid dimensions (nx,ny,nz) but equally important a definition of the local coordinate system for the field parameters. grid_dimensions is not a very descriptive name of the `class ErtboxParameters`

src/ert/config/field.py
```diff
@@ -71,7 +71,7 @@ def create_flattened_cube_graph(px: int, py: int, pz: int) -> nx.Graph[int]:
 class Field(ParameterConfig):
     type: Literal["field"] = "field"
     dimensionality: Literal[3] = 3
-    ertbox_params: ErtboxParameters
+    grid_geometry: GridGeometry
```
xjules, line 71: I like `grid_geometry`!

tests/ert/unit_tests/config/test_field.py
```diff
@@ -380,9 +380,9 @@ def test_calculate_ertbox_parameters_synthetic_grid(origin, increment, rotation,
 
 
 @pytest.fixture
-def field_with_ertbox_params():
+def field_with_grid_params():
     # Return a Field object for test purpose
-    # Define parameters for the grid (ertbox parameters)
+    # Define parameters for the grid (grid parameters)
```
xjules, line 388: `# Define parameters for the grid (its geometry)`

xjules (approved):
Good renaming!
Optional: write a simple unit test `test_to22.py`

src/ert/config/field.py
oddvarlia: GridGeometry is better. but what about all documentation. ERTBOX has been used for a long time in courses and documentation including APS and its documentation. I don't think I will change it in APS repo, at least not now, but APS and ERT does not share code. But `fmutools` and APS share code and ERTBOX is consistenty used there also. The repo` subscript` also uses the ERTBOX as name for the field parameter grids.

oddvarlia: But maybe we can still use ERTBOX as terminology, but use GridGeometry as the name of the ERT python class. My argument for this is that GridGeometry is very general, but ERTBOX is for field parameter grid to be used as field parameters in ERT.

PR #12206, opened 2025-11-03
Move EverestConfig.with_defaults to conftest

### LINKED ISSUE #12186: Remove EverestConfig.with_defaults, only used in tests
This function is only in place to facilitate tests. That means it should go into `tests/everest/conftest.py` instead. 

Additionally, it is very close to the already existing `min_config` fixture, which can probably be used instead.

https://github.com/equinor/ert/blob/7f31d53ea047d8ae23b86f9a8409d364707865af/src/everest/config/everest_config.py#L1018-L1041

frode-aarstad:
**Issue**
Resolves #12186 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/everest/conftest.py
+++ b/tests/everest/conftest.py
@@ -27,25 +27,12 @@ from everest.config import (
 )
 from everest.config.control_config import ControlConfig
 from everest.detached import everserver
-from tests.everest.utils import get_optimal_result, relpath
+from tests.everest.utils import MIN_CONFIG, get_optimal_result, relpath
 
-MIN_CONFIG = dedent(
-    """
-    model: {"realizations": [0]}
-    controls:
-      -
-        name: my_control
-        type: well_control
-        min: 0
-        max: 0.1
-        perturbation_magnitude: 0.01
-        variables:
-          - { name: test, initial_guess: 0.1 }
-    objective_functions:
-      - {name: my_objective}
-    config_path: .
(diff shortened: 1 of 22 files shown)


tests/everest/conftest.py
```diff
@@ -321,3 +321,11 @@ def no_plugins():
         patch("ert.plugins.ErtPluginManager", patched_plugin_manager),
     ):
         yield
+
+
+def everest_config_with_defaults(**kwargs) -> EverestConfig:
```
verveerpj: Maybe this should be a fixture?

verveerpj: Or otherwise, it should maybe to got `test_util.py`

tests/everest/conftest.py
```diff
@@ -255,25 +273,7 @@ def run_config(test_data_case: str):
 
 @pytest.fixture
 def min_config():
```
berland: Suggest to move this fixture to the top, right after the definiton of MIN_CONFIG.

tests/everest/conftest.py
berland: Move to top, close to the other MIN_CONFIG stuff.

frode-aarstad: Its used slightly different than a fixture. Will move it to utils for now

PR #12213, opened 2025-11-03
Defer loading of ertscript workflows

yngve-sk:
Towards https://github.com/equinor/ert/pull/11513
Background: Workflow jobs come either site-installed plugins, or user-installed workflow jobs. Site-installed workflow jobs do not need serialization, as they can be resolved from site plugins by name. User installed workflow jobs do need to store a reference to the workflow script iself.

### DIFF
--- a/src/ert/config/workflow_job.py
+++ b/src/ert/config/workflow_job.py
@@ -6,10 +6,23 @@ import os
 import textwrap
 from abc import ABC, abstractmethod
 from dataclasses import field
-from typing import TYPE_CHECKING, Self, TypeAlias, cast
+from typing import (
+    TYPE_CHECKING,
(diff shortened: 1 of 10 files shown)


src/ert/config/workflow_job.py
```diff
@@ -42,7 +55,11 @@ def workflow_job_parser(file: str) -> ConfigDict:
     return parse(file, schema=schema)
 
 
-def workflow_job_from_file(config_file: str, name: str | None = None) -> WorkflowJob:
+def workflow_job_from_file(
+    config_file: str,
+    origin: Literal["user", "site"] | None = "site",
```
berland: Allowing `None` is probably not needed?

PR #12215, opened 2025-11-03
Add workflow for testing towards installed dummy site config

yngve-sk:
For early feedback on interactions with site configurations, only checking for queue options.

### DIFF
--- a/.github/workflows/test_ert_with_site_config.yml
+++ b/.github/workflows/test_ert_with_site_config.yml
@@ -1,0 +1,93 @@
+name: Run ERT tests against dummy site config
+
+on: [ push, pull_request, workflow_dispatch ]
+
+jobs:
+  test-against-lsf:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: ['ubuntu-latest']
+        test-type: [ 'gui-tests', 'rapid-tests' ]
+        python-version: ["3.12"]
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v5
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v6
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - uses: ./.github/actions/install_dependencies_qt
+        with:
+          os: ${{ matrix.os }}
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip setuptools wheel pytest uv
+
+      - name: Use uv and make venv
+        run: |
+          uv venv .venv
+
+      - name: Install ERT + dummy plugin package that sets site config
+        run: |
+          set -euo pipefail
+
+          # Create package structure
+          mkdir -p dummy_site_config/dummy_plugin
+
+          # --- pyproject.toml ---
+          cat > dummy_site_config/pyproject.toml <<'END_OF_FILE'
+          [build-system]
+          requires = ["setuptools", "wheel"]
+          build-backend = "setuptools.build_meta"
+
+          [project]
+          name = "dummy-site-configuration"
+          version = "0.1.0"
+          description = "Dummy ERT plugin for testing"
+          requires-python = ">=3.8"
+
+          [project.entry-points.ert]
+          dummy_site_config = "dummy_plugin"
+          END_OF_FILE
+
+          # --- __init__.py ---
+          cat > dummy_site_config/dummy_plugin/__init__.py <<'END_OF_FILE'
+          from ert import plugin
+          from ert.plugins import ErtRuntimePlugins
+          from ert.config.queue_config import LsfQueueOptions
+
+          @plugin(name="fmu-site-configuration")
+          def site_configurations():
+              return ErtRuntimePlugins(
+                  queue_options=LsfQueueOptions(
+                      lsf_resource="select[cs && x86_64Linux]",
+                      lsf_queue="mr",
+                      max_running=10000,
+                      submit_sleep=1,
+                      realization_memory=10737418240
+                  ),
+              )
+          END_OF_FILE
+
+          # Install the dummy package
+          source .venv/bin/activate
+          uv sync --extra dev
+          uv pip install ./dummy_site_config
+
+      - name: Rapid tests
+        if: matrix.test-type == 'rapid-tests'
+        run: |
+          source .venv/bin/activate
+          uv run just rapid-tests
+
+      - name: GUI screenshot tests
+        if: matrix.test-type == 'gui-tests'
+        run: |
+          source .venv/bin/activate
+          uv run pytest tests/ert/ui_tests/gui/test_docs_screenshots.py


.github/workflows/test_ert_with_site_config.yml
```diff
@@ -0,0 +1,96 @@
...
+
+          @hook_implementation
+          def plugin(name="dummy-site-configuration"):
+              return ErtRuntimePlugins(
+                  queue_options=LsfQueueOptions(
+                      lsf_resource="select[cs && x86_64Linux]",
+                      lsf_queue="mr",
+                      max_running=10000,
+                      submit_sleep=1,
+                  ),
```
berland, line 74: add realization_memory to catch my blunder as well.

berland:
Is this possible to achieve also by mocking a site-config in selected tests? (running this entire workflow will certainly lead to an extra usage of cpu-minutes).

Also, if it is a mocked site-config, it could be fast enough for unit-tests.

yngve-sk:
It is possible to mock site config for tests, the fixture `use_site_configurations_with_lsf_queue_options` applies LSF queue options to runtime plugins (and thus also site config). Problem is it wouldn't catch the problem of us running tests (that work without site config installed) on komodo and have it run as expected, unless we beforehand know the test should be forced to run with LSF queue options (then we can't really set it to local). Would somehow be possible to set it up though I guess, but not sure it would give us the feedback beforehand. 

> Is this possible to achieve also by mocking a site-config in selected tests? (running this entire workflow will certainly lead to an extra usage of cpu-minutes).
> 
> Also, if it is a mocked site-config, it could be fast enough for unit-tests.

tests/conftest.py
```diff
@@ -130,11 +127,11 @@ def setup_svg_search_path():
...
-    ):
-        yield
+    return None
+    # def ErtRuntimePluginsWithNoQueueOptions(**kwargs):
+    #    return ErtRuntimePlugins(**(kwargs | {"queue_options": None}))
+    # with patch(
+    #    "ert.plugins.plugin_manager.ErtRuntimePlugins",
+    #    ErtRuntimePluginsWithNoQueueOptions,
+    # ):
+    #    yield
```
andreas-el: Leftovers?

tests/ert/ui_tests/gui/test_docs_screenshots.py
```diff
@@ -291,7 +291,7 @@ def ignore_pngs(src, files):
     return next(gui_generator)
 
 
-@pytest.mark.usefixtures("use_site_configurations_with_no_queue_options")
+# @pytest.mark.usefixtures("use_site_configurations_with_no_queue_options")
```
andreas-el: Remove or re-add

.github/workflows/test_ert_with_site_config.yml
```diff
@@ -0,0 +1,106 @@
+name: Run ERT tests against dummy site config
+
+on:
+  push:
+  pull_request:
```
andreas-el: Why not add dispatch capability for this also?

.github/workflows/test_ert_with_site_config.yml
```diff
@@ -0,0 +1,106 @@
...
+
+jobs:
+  test-against-lsf:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: ['ubuntu-latest']
+        test-type: [ 'gui-tests', 'rapid-tests' ]
+        python-version: ["3.11"]
```
andreas-el: Perhaps lean forward and use `3.12`

.github/workflows/test_ert_with_site_config.yml
```diff
@@ -0,0 +1,106 @@
...
+
+      - name: Rapid tests
+        if: matrix.test-type == 'rapid-tests'
+        run: |
+          source .venv/bin/activate
+          touch storage
+          chmod 000 storage
+          OMP_NUM_THREADS=1 pytest -n auto --benchmark-disable --dist loadgroup tests/ert/unit_tests tests/everest --hypothesis-profile=fast -m "not (integration_test or flaky or memory_test or limit_memory)" --timeout=10 --session-timeout=120
+          chmod 700 storage
+          rm storage
```
andreas-el: This is a file, not folder? What is the intent for using this file with no permissions set?

tests/ert/ui_tests/gui/test_docs_screenshots.py
yngve-sk: Just to demonstrate/test that it fails when the fixture is gone (which is what we expect), with fixture back they will all pass, will remove

tests/conftest.py
yngve-sk: Will put back, was tmp to ensure it failed without the fixture

.github/workflows/test_ert_with_site_config.yml
yngve-sk: copypaste remnant, ref https://github.com/equinor/ert/pull/11223 , will remove though, file pollution should be unrelated to site config.

.github/workflows/test_ert_with_site_config.yml
```diff
@@ -0,0 +1,96 @@
...
+          # Install the dummy package
+          source .venv/bin/activate
+          uv sync --extra dev
+          uv pip install ./dummy_site_config
+
+      - name: Rapid tests
+        if: matrix.test-type == 'rapid-tests'
+        run: |
+          source .venv/bin/activate
+          OMP_NUM_THREADS=1 pytest -n auto --benchmark-disable --dist loadgroup tests/ert/unit_tests tests/everest --hypothesis-profile=fast -m "not (integration_test or flaky or memory_test or limit_memory)" --timeout=10 --session-timeout=720
```
yngve-sk: Seems to exceed timeout... https://github.com/equinor/ert/actions/runs/19266220017/job/55082581828

Also does not include git-lfs storage migration tests (I think this is OK)

.github/workflows/test_ert_with_site_config.yml
```diff
@@ -0,0 +1,93 @@
...
+          # Install the dummy package
+          source .venv/bin/activate
+          uv sync --extra dev
+          uv pip install ./dummy_site_config
+
+      - name: Rapid tests
+        if: matrix.test-type == 'rapid-tests'
+        run: |
+          source .venv/bin/activate
+          OMP_NUM_THREADS=1 pytest -n auto --benchmark-disable --dist loadgroup tests/ert/unit_tests tests/everest --hypothesis-profile=fast -m "not (integration_test or flaky or memory_test or limit_memory)" --timeout=10 --session-timeout=120
```
berland: call `just rapid-tests` directly?

.github/workflows/test_ert_with_site_config.yml
```diff
@@ -0,0 +1,93 @@
...
+        with:
+          os: ${{ matrix.os }}
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip setuptools wheel pytest uv
+
+      - name: Use uv and make venv
+        run: |
+          python -m uv venv .venv
```
berland: `uv venv .venv` will do the same?

.github/workflows/test_ert_with_site_config.yml
```diff
@@ -0,0 +1,93 @@
...
+
+      - name: Install ERT + dummy plugin package that sets site config
+        run: |
+          set -euo pipefail
+
+          # Create package structure
+          mkdir -p dummy_site_config/dummy_plugin
+
+          # --- pyproject.toml ---
+          cat > dummy_site_config/pyproject.toml <<'PYPROJECT'
```
berland: Replace `PYPROJECT` by `END_OF_FILE`. Makes it clearer.

.github/workflows/test_ert_with_site_config.yml
```diff
@@ -0,0 +1,93 @@
...
+          version = "0.1.0"
+          description = "Dummy ERT plugin for testing"
+          requires-python = ">=3.8"
+
+          [project.entry-points.ert]
+          dummy_site_config = "dummy_plugin"
+          PYPROJECT
+
+          # --- __init__.py ---
+          cat > dummy_site_config/dummy_plugin/__init__.py <<'PYPLUGIN'
```
berland: ditto

tests/conftest.py
```diff
@@ -130,11 +127,10 @@ def setup_svg_search_path():
...
-        "ert.plugins.plugin_manager.ErtRuntimePlugins",
-        ErtRuntimePluginsWithNoQueueOptions,
-    ):
-        yield
+    # def ErtRuntimePluginsWithNoQueueOptions(**kwargs):
+    #    return ErtRuntimePlugins(**(kwargs | {"queue_options": None}))
+    # with patch(
+    #    "ert.plugins.plugin_manager.ErtRuntimePlugins",
+    #    ErtRuntimePluginsWithNoQueueOptions,
+    # ):
```
andreas-el: leftovers?

PR #12232, opened 2025-11-05
Split InstallJobConfig by usage

yngve-sk:
It is currently used for both forward model steps and workflows, disjoint by having `.to_ert_forward_model_step` and `.to_ert_executable_workflow`, making this distinction early makes it more clear what the intent of the config is.


verveerpj (approved):
LGTM!

PR #12239, opened 2025-11-05
Serialize only name and args of site-installed forward model steps

yngve-sk:
Pre-work for https://github.com/equinor/ert/pull/11513
Motivation is to not have a serialized runmodel be locked to the executable at that time. If the forward model step is from a plugin, it should upon serialization, store a reference to its plugin. If the plugin changes/updates, so should the forward model step.

Example serialization of site-installed forward model:
```
{
     "type": "site_installed",
     "name": "SITE_INSTALLED_FM",
}
```

(instead of the entire jobs.json dump)

PS: Some external plugin FM steps are still using legacy format (STEA for example) (config file for forward model, hence the `origin` arg of `forward_model_step_from_config_contents`.

### DIFF
--- a/src/ert/config/ert_config.py
+++ b/src/ert/config/ert_config.py
@@ -36,6 +36,9 @@ from .forward_model_step import (
     ForwardModelStepJSON,
     ForwardModelStepValidationError,
     ForwardModelStepWarning,
+    SiteInstalledForwardModelStep,
+    SiteOrUserForwardModelStep,
+    UserInstalledForwardModelStep,
 )
 from .gen_data_config import GenDataConfig
 from .gen_kw_config import GenKwConfig
@@ -518,18 +521,16 @@ def workflows_from_dict(
 
 def installed_forward_model_steps_from_dict(
     config_dict: ConfigDict,
-) -> dict[str, ForwardModelStep]:
+) -> dict[str, UserInstalledForwardModelStep]:
     errors: list[ErrorInfo | ConfigValidationError] = []
-    fm_steps: dict[str, ForwardModelStep] = {}
+    fm_steps: dict[str, UserInstalledForwardModelStep] = {}
     for name, (fm_step_config_file, config_contents) in config_dict.get(
         ConfigKeys.INSTALL_JOB, []
     ):
         fm_step_config_file = path.abspath(fm_step_config_file)
         try:
             new_fm_step = forward_model_step_from_config_contents(
-                config_contents,
-                name=name,
-                config_file=fm_step_config_file,
+                config_contents, name=name, config_file=fm_step_config_file
             )
         except ConfigValidationError as e:
             errors.append(e)
@@ -720,7 +721,7 @@ class ErtConfig(BaseModel):
 
     ert_templates: list[tuple[str, str]] = Field(default_factory=list)
 
-    forward_model_steps: list[ForwardModelStep] = Field(default_factory=list)
+    forward_model_steps: list[SiteOrUserForwardModelStep] = Field(default_factory=list)
     runpath_config: ModelConfig = Field(default_factory=ModelConfig)
     user_config_file: str = "no_config"
     config_path: str = Field(init=False, default="")
@@ -817,7 +818,7 @@ class ErtConfig(BaseModel):
     def with_plugins(runtime_plugins: ErtRuntimePlugins) -> type[ErtConfig]:
         class ErtConfigWithPlugins(ErtConfig):
             PREINSTALLED_FORWARD_MODEL_STEPS: ClassVar[
-                Mapping[str, ForwardModelStep]
+                Mapping[str, SiteInstalledForwardModelStep]
             ] = runtime_plugins.installed_forward_model_steps
             PREINSTALLED_WORKFLOWS = dict(runtime_plugins.installed_workflow_jobs)
             ENV_PR_FM_STEP: ClassVar[dict[str, dict[str, Any]]] = (
@@ -1365,8 +1366,10 @@ def uppercase_subkeys_and_stringify_subvalues(
 
 
 def forward_model_step_from_config_contents(
-    config_contents: str, config_file: str, name: str | None = None
-) -> ForwardModelStep:
+    config_contents: str,
+    config_file: str,
+    name: str | None = None,
+) -> UserInstalledForwardModelStep:
     if name is None:
         name = os.path.basename(config_file)
 
@@ -1390,7 +1393,7 @@ def forward_model_step_from_config_contents(
     environment = {k: v for [k, v] in content_dict.get("ENV", [])}
     default_mapping = {k: v for [k, v] in content_dict.get("DEFAULT", [])}
 
-    return ForwardModelStep(
+    return UserInstalledForwardModelStep(
         name=name,
         executable=content_dict["EXECUTABLE"],
         stdin_file=content_dict.get("STDIN"),
--- a/src/ert/config/forward_model_step.py
+++ b/src/ert/config/forward_model_step.py
@@ -1,12 +1,32 @@
 from __future__ import annotations
 
 import logging
-from typing import ClassVar, Literal, NotRequired, Self
+from typing import (
+    TYPE_CHECKING,
+    Annotated,
+    Any,
+    ClassVar,
+    Literal,
+    NotRequired,
+    Self,
+    cast,
+)
 
-from pydantic import BaseModel, Field, field_validator, model_validator
+from pydantic import (
+    BaseModel,
+    Field,
+    field_validator,
+    model_serializer,
+    model_validator,
+)
+from pydantic_core.core_schema import ValidationInfo
 from typing_extensions import TypedDict, Unpack
 
+from ..base_model_context import BaseModelWithContextSupport
 from .parsing import ConfigValidationError, ConfigWarning, SchemaItemType
+
+if TYPE_CHECKING:
+    from ert.plugins import ErtRuntimePlugins
 
 logger = logging.getLogger(__name__)
 
@@ -92,7 +112,7 @@ class ForwardModelStepDocumentation(BaseModel):
     ) = Field(default="Uncategorized")
 
 
-class ForwardModelStep(BaseModel):
+class ForwardModelStep(BaseModelWithContextSupport):
     """
     Holds information to execute one step of a forward model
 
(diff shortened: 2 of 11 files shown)


src/ert/config/forward_model_step.py
```diff
@@ -202,7 +222,62 @@ def set_stdin_file(cls, v: str | None) -> str | None:
...
+        cls, values: dict[str, Any], info: ValidationInfo
+    ) -> dict[str, Any]:
+        runtime_plugins = cast("ErtRuntimePlugins", info.context)
+        name = values["name"]
+
+        if runtime_plugins is None:
+            if set(values.keys()) == {"name", "type"}:
+                raise ValueError(
+                    f"Cannot resolve forward model step {values},"
+                    f"as it expects a the forward model {name}"
```
eivindjahren: There is a grammatical mistake with the "a the" part here. Also there is not test of the raised exceptions.

yngve-sk: fixed

src/ert/config/ert_config.py
```diff
@@ -813,7 +817,7 @@ def __eq__(self, other: object) -> bool:
     def with_plugins(runtime_plugins: ErtRuntimePlugins) -> type[ErtConfig]:
         class ErtConfigWithPlugins(ErtConfig):
             PREINSTALLED_FORWARD_MODEL_STEPS: ClassVar[
-                Mapping[str, ForwardModelStep]
+                Mapping[str, SiteOrUserForwardModelStep]
```
berland: Can UserInstalledForwardModelStep come into play here?

src/ert/config/forward_model_step.py
```diff
@@ -202,7 +222,78 @@ def set_stdin_file(cls, v: str | None) -> str | None:
         return None if v == "null" else v
 
 
-class ForwardModelStepPlugin(ForwardModelStep):
+class UserInstalledForwardModelStep(ForwardModelStep):
+    """
+    Represents a forward model step installed by a user via the ERT Config
+    forward model step format provided via the INSTALL_JOB keyword..
```
berland: double punctuation: `..` -> `.`

src/ert/config/forward_model_step.py
```diff
@@ -202,7 +222,78 @@ def set_stdin_file(cls, v: str | None) -> str | None:
...
+    Instances of this class serialize only as references to the plugin by name,
+    allowing them to dynamically update when plugins change, rather than being
+    locked to a specific executable at serialization time.
+    """
+
+    type: Literal["site_installed"] = "site_installed"
+
+    @model_serializer(mode="plain")
+    def serialize_model(self) -> _SerializedSiteInstalledForwardModelStep:
+        return {"type": "site_installed", "name": self.name}
```
berland: Would it be clearer if you used the class `_SerializedSiteInstalledForwardModelStep` to initialize? (though I assume this code just making the dict directly is equivalent)

src/ert/config/forward_model_step.py
```diff
@@ -202,7 +222,78 @@ def set_stdin_file(cls, v: str | None) -> str | None:
...
+    @model_validator(mode="before")
+    @classmethod
+    def deserialize_model(
+        cls, values: dict[str, Any], info: ValidationInfo
+    ) -> dict[str, Any]:
+        runtime_plugins = cast("ErtRuntimePlugins", info.context)
+        name = values["name"]
+
+        if runtime_plugins is None:
+            if set(values.keys()) == {"name", "type"}:
```
berland: this comparison looks odd, will it always hold that when there are no plugins, values.keys() will always be like this? Would it not be better to look for `{name}` explicitly?

src/ert/config/forward_model_step.py
```diff
@@ -202,7 +222,78 @@ def set_stdin_file(cls, v: str | None) -> str | None:
...
+
+    @model_serializer(mode="plain")
+    def serialize_model(self) -> _SerializedSiteInstalledForwardModelStep:
+        return {"type": "site_installed", "name": self.name}
+
+    @model_validator(mode="before")
+    @classmethod
+    def deserialize_model(
+        cls, values: dict[str, Any], info: ValidationInfo
+    ) -> dict[str, Any]:
```
berland, line 265: When we deserialize, should we not get a proper class, not just a dict?

src/ert/config/forward_model_step.py
```diff
@@ -202,7 +222,78 @@ def set_stdin_file(cls, v: str | None) -> str | None:
...
+                f"Expected forward model step {name} to be installed "
+                f"via plugins, but it was not found. Please check that "
+                f"your python environment has it installed."
+            )
+            raise KeyError(msg)
+        site_installed_fm = runtime_plugins.installed_forward_model_steps[name]
+
+        # Intent: copy the site installed forward model to this instance.
+        # bypassing the model_serializer
+        return {
```
berland: this is not the same as f.ex. `model_dump`? (have not checked)

src/ert/config/ert_config.py
yngve-sk: I think not, changed to only site installed.

src/ert/config/forward_model_step.py
yngve-sk: AFAIK I think the TypedDict is only for typing by convention, even though we could use it, I think just doing the dict is more clear, as serialization to json/obj is also what we will export/show outside of a python context.

src/ert/config/forward_model_step.py
yngve-sk: Reformulating it, it is only meant to raise an error if you are trying to run a runmodel depending on site plugins, outside of a plugin context.

src/ert/config/forward_model_step.py
yngve-sk: I think Pydantic `before` validators take in a dict and should return a dict, if it was an `after` validator we would already have an instance.

tests/ert/unit_tests/config/test_forward_model.py
```diff
@@ -1005,3 +1008,118 @@ def test_that_all_required_keywords_in_forward_model_are_validated():
...
+
+    with use_runtime_plugins(
+        ErtRuntimePlugins(installed_forward_model_steps={"SITE_FM": SiteForwardModel()})
+    ):
+        site_fm = SiteForwardModel()
+        # Expect only reference stored, i.e., name
+        serialized_fm_step = site_fm.model_dump(mode="json")
+        assert serialized_fm_step == {"name": "SITE_FM", "type": "site_installed"}
+
+    custom_echo_path = Path("custom_echo.sh")
```
berland, line 1090: Should this be split into two tests, where the test name of the second one mentions that we are testing that a site-installed step can be overwritten when deserialized?

src/ert/config/forward_model_step.py
yngve-sk: It is a workaround to bypass invoking `serialize_model`, in this case we want to actually copy over the site installed attrs to this instance, in dict format. model_dump will be hijacked by `serialize_model`.

We need to be able to:
(1) serialize model to only name and type (sidenote: and private args I think)
(2) deserialize name and type to actual fully specified forward model, but since the fully specified forward model is a site installed, we can't model_dump it, therefore we bypass here)

tests/ert/unit_tests/config/test_forward_model.py
```diff
@@ -1005,3 +1008,118 @@ def test_that_all_required_keywords_in_forward_model_are_validated():
...
+    ert_config = ErtConfig.from_file("config.ert")
+    user_fm = ert_config.forward_model_steps[0]
+    assert user_fm.type == "user_installed"
+    assert user_fm.executable == "echo"
+
+
+def test_that_fm_step_serializes_name_only_for_site_and_full_for_user(
+    use_tmpdir,
+):
+    Path("fm_step").write_text("EXECUTABLE echo\n", encoding="utf-8")
```
berland: Need to test for all features of user installed steps.

src/ert/config/forward_model_step.py
```diff
@@ -202,7 +222,78 @@ def set_stdin_file(cls, v: str | None) -> str | None:
...
+
+        if runtime_plugins is None:
+            if set(values.keys()) == {"name", "type"}:
+                msg = (
+                    f"Trying to find site-installed forward model step {name} "
+                    f"without site plugins. This forward model must be loaded "
+                    f"with ERT site plugins available."
+                )
+                raise KeyError(msg)
+            return values
```
berland, line 277: Do we want to return here, and not just let it be picked up below?

yngve-sk: Trying to remove, I think it is not needed.

yngve-sk: Seems it is needed at the moment, maybe there is a better way to solve this, but currently when we parse plugins, they go through here:

```

class CarefulCopyFile(ForwardModelStepPlugin):
    def __init__(self) -> None:
        super().__init__(
            name="CAREFUL_COPY_FILE",
            command=[
                str(
                    (
                        Path(__file__)
                        / "../../../resources/shell_scripts/careful_copy_file.py"
                    ).resolve()
                ),
                "<FROM>",
                "<TO>",
            ],
        )
```
then into this constructor:
```
class ForwardModelStepPlugin(SiteInstalledForwardModelStep):
    def __init__(
        self, name: str, command: list[str], **kwargs: Unpack[ForwardModelStepOptions]
    ) -> None:
        if not kwargs:
            kwargs = ForwardModelStepOptions()

        executable = command[0]
        arglist = command[1:]

        stdin_file = kwargs.get("stdin_file")
        stdout_file = kwargs.get("stdout_file")
        stderr_file = kwargs.get("stderr_file")
        start_file = kwargs.get("start_file")
        target_file = kwargs.get("target_file")
        error_file = kwargs.get("error_file")
        max_running_minutes = kwargs.get("max_running_minutes")
        environment = kwargs.get("environment", {}) or {}
        default_mapping = kwargs.get("default_mapping", {}) or {}

        super().__init__(
            name=name,
            executable=executable,
            arglist=arglist,
            stdin_file=stdin_file,
            stdout_file=stdout_file,
            stderr_file=stderr_file,
            start_file=start_file,
            target_file=target_file,
            error_file=error_file,
            max_running_minutes=max_running_minutes,
            min_arg=0,
            max_arg=0,
            required_keywords=[],
            arg_types=[],
            environment=environment,
            default_mapping=default_mapping,
            private_args={},
        )
```

and those args are fed into this model validator `SiteInstalledForwardModelStep.deserialize_model`, and they are indeed missing the `type`, so if that is missing in the  `deserialize_model` validator, it means it was a "direct" / "from-site" initialization, as opposed to a deserialization of an actually serialized model..

berland:
> PS: Some external plugin FM steps are still using legacy format (STEA for example) (config file for forward model, hence the `origin` arg of `forward_model_step_from_config_contents`.

Fixed in https://github.com/equinor/fmu-steaclient/pull/92

src/ert/config/forward_model_step.py
berland: Yes, my thinking was that since you have defined the class, you could use it for more than typing, even if you just do model_dump() or something to get the same output (probably implies a very marginal speed decrease)

tests/ert/unit_tests/config/test_forward_model.py
yngve-sk: Added `private_args` for serialization there.

tests/ert/unit_tests/config/test_forward_model.py
```diff
@@ -1005,3 +1008,145 @@ def test_that_all_required_keywords_in_forward_model_are_validated():
...
+            site_fm_with_updated_executable, SiteInstalledForwardModelStep
+        )
+        assert site_fm_with_updated_executable.executable == "custom_echo.sh"
+        assert site_fm_with_updated_executable.private_args == {
+            "hello": "hi",
+            "hi": "2",
+        }
+
+
+def test_that_user_fm_step_serializes_has_user_installed_type_and_retains_excecutable(
```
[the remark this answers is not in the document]
berland: fix typo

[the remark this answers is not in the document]
berland: agree that the name is unclear, it has two verbs: "serializes" and "has"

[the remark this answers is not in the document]
yngve-sk: The diff above is outdated, but did a fixup to the test I would have missed otherwise https://github.com/equinor/ert/pull/12239/commits/9e62d03ace2f035899bb0a3eab6c92cddf1bd03f

Issue #11393, opened 2025-07-25
Store weight, scale, type of objectives in EverestObjectivesConfig


yngve-sk:
Currently these are stored separately in a dataframe at experiment-level, instead they should be attached to their respective configs.

PR #12248, opened 2025-11-07
Store all Everest objective info in ERT storage

yngve-sk:
Resolves https://github.com/equinor/ert/issues/11393
Needed info is already stored in the response config which is more in line with ERT, and is already stored in ERT storage.

Migration is debatable, easiest is to just default the fields if missing. The case storage would potentially be off, but since it is already run, the results would be the same. If we don't migrate, older runs cannot be plotted with future ERT.

### DIFF
--- a/src/ert/storage/local_storage.py
+++ b/src/ert/storage/local_storage.py
@@ -31,7 +31,7 @@ from .realization_storage_state import RealizationStorageState
 
 logger = logging.getLogger(__name__)
 
-_LOCAL_STORAGE_VERSION = 15
+_LOCAL_STORAGE_VERSION = 16
 
 
 class _Migrations(BaseModel):
@@ -506,6 +506,7 @@ class LocalStorage(BaseMode):
             to13,
             to14,
             to15,
+            to16,
         )
 
         try:
@@ -550,6 +551,7 @@ class LocalStorage(BaseMode):
                     12: to13,
                     13: to14,
                     14: to15,
+                    15: to16,
                 }
                 for from_version in range(version, _LOCAL_STORAGE_VERSION):
                     migrations[from_version].migrate(self.path)
--- a/tests/ert/unit_tests/storage/migration/test_to16.py
+++ b/tests/ert/unit_tests/storage/migration/test_to16.py
@@ -1,0 +1,57 @@
+import json
+from pathlib import Path
+
+from ert.storage.migration.to16 import (
+    migrate_everest_objectives_config,
+    migrate_everest_objectives_in_responses_json,
+)
+
+
+def test_that_defaults_are_added_to_single_objectives_config_object():
+    original_config = {"keys": ["objective1", "objective2"]}
+
+    expected_config = {
+        "keys": ["objective1", "objective2"],
+        "weights": [None, None],
+        "scales": [None, None],
+        "objective_types": ["mean", "mean"],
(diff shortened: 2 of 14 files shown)


src/ert/storage/local_storage.py
```diff
@@ -31,7 +31,7 @@
 
 logger = logging.getLogger(__name__)
 
-_LOCAL_STORAGE_VERSION = 15
+_LOCAL_STORAGE_VERSION = 16
```
andreas-el, line 34: Needs to be bumped later when merging the other PR, or combining the PRs.

tests/ert/unit_tests/storage/migration/test_to16.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def test_that_defaults_are_added_to_single_objectives_config_object():
+    original_config = {"keys": ["objective1", "objective2"]}
+
+    expected_config = {
+        "keys": ["objective1", "objective2"],
+        "weights": [None, None],
+        "scales": [None, None],
+        "objective_types": ["mean", "mean"],
```
andreas-el, line 17: Is `mean` the only option here, or the only one we support?

tests/ert/unit_tests/storage/migration/test_to16.py
```diff
@@ -0,0 +1,57 @@
...
+)
+
+
+def test_that_defaults_are_added_to_single_objectives_config_object():
+    original_config = {"keys": ["objective1", "objective2"]}
+
+    expected_config = {
+        "keys": ["objective1", "objective2"],
+        "weights": [None, None],
+        "scales": [None, None],
```
andreas-el, line 16: I'd expect the scale to default to `1.0` ? But there probably is code somewhere that verifies if this is not `None` ?

andreas-el:
```
 scales=[1.0 if scale is None else scale for scale in objectives.scales],
```

andreas-el: Question is, why not default this?

src/ert/storage/local_storage.py
yngve-sk: yep

tests/ert/unit_tests/storage/migration/test_to16.py
yngve-sk: `mean` is the default, other (ref `objective_function_config.py` is `stddev`, migration just inserts the default. If a migration happens, the optimization is already executed.

tests/ert/unit_tests/storage/migration/test_to16.py
yngve-sk: Hm, was just mirroring the default in the everest objective config object, but probably scale 1 makes more sense.. will look upstream and see if it can be incorporated.

yngve-sk: Seems it has to be `| None` now. If set to `None`, it is defaulted to `1.0` downstream, BUT it also has to be disabled / `None` if `auto_scale` is set in the `optimization` config. Ref the `deprecate_normalization` validator in `objective_function_config.py`. Might be simplified, but suggest keeping it as a separate issue.

Issue #12222, opened 2025-11-04
Overwriting SITE_CONFIG SUBMIT_SLEEP does not work


jonathan-eq:
### What happened? (You can include a screenshot if it helps explain)

I tried running ert while having a SITE CONFIG installed, and it was using SITE_CONFIG SUBMIT_SLEEP 1, which is fine. It turned out to be a tad too slow for me, so I overwrote it with `QUEUE_OPTION LSF SUBMIT_SLEEP 0` and `SUBMIT_SLEEP 0`, but neither of them took presedence over SITE_CONFIG configuration. 

It unfortunately also logs a misleading message to the logs
```
2025-11-03 16:18:44,252 - ert.config.queue_config - MainThread - INFO - Overwriting site config setting: max_running=10000 with QUEUE_OPTION LOCAL MAX_RUNNING 50
2025-11-03 16:18:44,253 - ert.config.queue_config - MainThread - INFO - Overwriting site config setting: submit_sleep=1.0 with QUEUE_OPTION LSF SUBMIT_SLEEP 0.0

```

### What did you expect to happen?

It should be possible to overwrite configuration from SITE_CONFIG. 

### Steps to reproduce

```Python
Source the november release
Try setting `QUEUE_OPTION LSF SUBMIT_SLEEP 0`
Run poly_example and observe that submit sleep is indeed not 0, but still 1. 
```

### Environment where bug has been observed

- [x] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [x] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [x] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

jonathan-eq:
@berland, @JHolba, and I did some digging into this one. SUBMIT_SLEEP 0 defaults to 1 (or maybe site config, not 100% sure), but SUBMIT_SLEEP 0.000001 works as expected. 
I am betting my favorite coffee mug that this is an issue where we do `if not submit_sleep` instead of `if submit_sleep is not None`, and SUBMIT_SLEEP 0 shows up as falsy...

PR #12251, opened 2025-11-07
Remove unnecessary exclude_defaults for usr queue options

yngve-sk:
They should be excluded from site config, but not from user config.

**Issue**
Resolves https://github.com/equinor/ert/issues/12222

### DIFF
--- a/tests/ert/unit_tests/run_models/test_base_run_model.py
+++ b/tests/ert/unit_tests/run_models/test_base_run_model.py
@@ -12,9 +12,11 @@ from unittest.mock import AsyncMock, MagicMock, patch
 import pytest
 from pydantic import ConfigDict
 
-from ert.config import ErtConfig, GenKwConfig, ModelConfig, QueueConfig
+from ert.config import ErtConfig, GenKwConfig, ModelConfig, QueueConfig, QueueSystem
+from ert.config.queue_config import LsfQueueOptions
 from ert.ensemble_evaluator import EndEvent, EvaluatorServerConfig
 from ert.ensemble_evaluator.snapshot import EnsembleSnapshot
+from ert.plugins import ErtRuntimePlugins
 from ert.run_models.run_model import RunModel, UserCancelled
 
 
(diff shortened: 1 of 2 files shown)


tests/ert/unit_tests/run_models/test_base_run_model.py
```diff
@@ -619,3 +621,21 @@ def mock_logging(_, log_str):
 
     with patch.object(Logger, "info", mock_logging):
         rm.log_at_startup()
+
+
+def test_that_defaulted_user_submit_sleep_overrides_site_setting(use_tmpdir):
+    user_queue_options = LsfQueueOptions(submit_sleep=0)
+
+    class DummyValidationInfo:
```
berland, line 634: This mock smells a little bit of implementation details. If we refactor how site configs are injected, could it be that this test will yield false positives in the future?

yngve-sk: Hmm, I agree it does smell but it also matches the smell of the way we inject into validators, we now fetch it via the info.context, would normally do a 
```
    with use_runtime_plugins(
        ErtRuntimePlugins(queue_options=LsfQueueOptions(submit_sleep=1))
```

but we can't trigger only a single pydantic validator via that (without creating a runmodel) AFAIK. It will be easier with runmodels split into config / runtime though.. Changing the way we inject would require a change in this test too, though the test would likely just fail with an error.

tests/ert/unit_tests/run_models/test_base_run_model.py
```diff
@@ -619,3 +621,21 @@ def mock_logging(_, log_str):
 
     with patch.object(Logger, "info", mock_logging):
         rm.log_at_startup()
+
+
+def test_that_defaulted_user_submit_sleep_overrides_site_setting(use_tmpdir):
```
jonathan-eq: I wonder if we should test this with more settings than just submit_sleep 🤔

yngve-sk: Made test general for all QueueOptions settings (not for specific LSF/SLURM etc though)

berland (approved):
👍🏻

Issue #11394, opened 2025-07-25
Store all output constraint info in ERT storage


yngve-sk:
If they are to be stored into ERT storage, the experiment in ERT storage should contain the full info about what kind of constraint it is, this means including `target`, `auto_scale`, `lower_bound`, `upper_bound`, `scale` in the ERT response config too.

PR #12255, opened 2025-11-07
Store all everest constraints info in ertstorage

yngve-sk:
Resolves [#11394](https://github.com/equinor/ert/issues/11394)


tests/ert/unit_tests/storage/migration/test_to16.py
```diff
@@ -0,0 +1,63 @@
...
+
+def test_that_defaults_are_added_to_single_constraints_config_object():
+    original_config = {"keys": ["constraint1", "constraint2"]}
+
+    expected_config = {
+        "keys": ["constraint1", "constraint2"],
+        "targets": [None, None],
+        "scales": [None, None],
+        "upper_bounds": [np.inf, np.inf],
+        "lower_bounds": [np.inf, np.inf],
```
andreas-el: Why isn't the default lower bound negative `inf`?

src/ert/run_models/everest_run_model.py
```diff
@@ -652,6 +642,19 @@ def _handle_optimizer_results(self, results: tuple[Results, ...]) -> None:
...
+    def output_constraints_config(self) -> EverestConstraintsConfig | None:
+        constraints_config = next(
+            (c for c in self.response_configuration if c.type == "everest_constraints"),
+            None,
+        )
+
+        if constraints_config is None:
+            return None
+
+        assert isinstance(constraints_config, EverestConstraintsConfig)
```
andreas-el, line 649: Is this for silencing the type checking?

src/ert/config/everest_constraints_config.py
```diff
@@ -27,6 +27,10 @@ def metadata(self) -> list[ResponseMetadata]:
     type: Literal["everest_constraints"] = "everest_constraints"
     name: str = "everest_constraints"
     has_finalized_keys: bool = True
+    targets: list[float | None]
+    scales: list[float | None]
+    upper_bounds: list[float]
+    lower_bounds: list[float]
```
andreas-el, line 33: Is this a candidate for a data structure? I assume the length of all these lists have to add up and that the indices matter.

src/everest/optimizer/everest2ropt.py
```diff
@@ -144,13 +145,26 @@ def _get_control_index(name: str) -> int:
...
-        lower_bounds, upper_bounds = _get_bounds(output_constraints)
         return {
-            "lower_bounds": lower_bounds,
-            "upper_bounds": upper_bounds,
+            "lower_bounds": [
+                target if target is not None else lb
+                for lb, target in zip(
+                    output_constraints.lower_bounds,
+                    output_constraints.targets,
+                    strict=False,
```
andreas-el, line 155: Can there be a situation where the length of the input is not the same? Is it allowed to only specify either upper or lower bound for example?
Ref strict.

tests/ert/unit_tests/storage/migration/test_to16.py
yngve-sk: oops! forgot a minus on the lower bounds in the migration indeed.

src/ert/run_models/everest_run_model.py
yngve-sk: Yes, the `type` of the response config from ERT identifies (wrt the response type) it as a `EverestConstraintsConfig` for sure, but typing does not tell us this, it only knows that it is a `ParameterConfig`.

andreas-el:
Not inspected, but suspect we should also update `https://github.com/equinor/ert-testdata` with storage content from previous ERT tag

src/ert/config/everest_constraints_config.py
yngve-sk: It could be but would defer that discussion to other issue, we adhere to, and inherit from `ResponseConfig` which does this for other GEN_DATA / summary also, maintaining a list of keys and corresponding other things. The lengths all do match up yes.

src/everest/optimizer/everest2ropt.py
yngve-sk: Hm, the lengths should always be the same, but the code does not strictly enforce matching lengths of lower and upper bounds. If there is a `target` for a given index it clamps the upper and lower bounds to exactly that target and passes it on to ROPT, otherwise there is always an upper+lower bound. 

Could be a validator on `OutputConstraintConfig` for checking the len

yngve-sk:
> Not inspected, but suspect we should also update `https://github.com/equinor/ert-testdata` with storage content from previous ERT tag

https://github.com/equinor/ert/pull/12260
https://github.com/equinor/ert-testdata/pull/30

PR #12266, opened 2025-11-12
Add poll_period to the Driver class

### LINKED ISSUE #12212: Use shorter poll_period when booting everserver
When everserver is submitted to a compute cluster, it will poll until there is a StartedEvent. 

Since the default poll period is 2 seconds, there will be an extra waiting time up to 2 seconds after everserver is ready until Everest picks up it is ready.

To solve this:

* Add `poll_period` as an argument the abstract `Driver` class (so all driver must support it)
* Slurm driver already has an `squeue_timeout` argument that translates into `poll_period`. Maybe both arguments need to be supported for this driver.
* Change `detached/client.py` to a low poll period (0.1s?) for the startup of everserver.

frode-aarstad:
And allow this to be set when creating the driver

**Issue**
Resolves #12212 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/detached/client.py
+++ b/src/everest/detached/client.py
@@ -45,7 +45,7 @@ async def start_server(config: EverestConfig, logging_level: int) -> Driver:
     """
     Start an Everest server running the optimization defined in the config
     """
-    driver = create_driver(config.server.queue_system)  # type: ignore
+    driver = create_driver(config.server.queue_system, poll_period=0.1)  # type: ignore
     try:
         args = [
             "--output-dir",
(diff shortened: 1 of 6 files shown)


src/everest/detached/client.py
```diff
@@ -45,7 +45,7 @@ async def start_server(config: EverestConfig, logging_level: int) -> Driver:
     """
     Start an Everest server running the optimization defined in the config
     """
-    driver = create_driver(config.server.queue_system)  # type: ignore
+    driver = create_driver(config.server.queue_system, poll_period=0.01)  # type: ignore
```
andreas-el: This is every `10ms`, is that too often perhaps?

frode-aarstad: Yes, should be 0.1s

andreas-el (approved):
💯  Well done!

PR #12282, opened 2025-11-13
Undo migration of everest objectives.

yngve-sk:
Instead of defaulting, we will do a major breaking change. I.e., we will not support (future) use of ERT to plot Everest objectives of older storages


andreas-el (approved):
💯

PR #12284, opened 2025-11-13
Improve test name

SAKavli:


tests/ert/unit_tests/gui/simulation/test_run_dialog.py
```diff
@@ -176,12 +175,12 @@ def handle_dialog():
         qtbot.mouseClick(run_dialog.kill_button, Qt.MouseButton.LeftButton)
         assert not kill_button.isEnabled()
         assert kill_button.text() == "Terminating"
-    suggestor_termination_window = wait_for_attribute(run_dialog.fail_msg_box, qtbot)
+    wait_until(lambda: run_dialog.fail_msg_box is not None)
```
eivindjahren: This is unfortunately not quite the same due to the difference in the timeout.

tests/ert/ui_tests/gui/conftest.py
```diff
@@ -406,11 +406,6 @@ def wait_for_child(gui, qtbot: QtBot, typ: type[V], timeout=5000, **kwargs) -> V
     return get_child(gui, typ, **kwargs)
 
 
-def wait_for_attribute(gui_attribute, qtbot: QtBot, timeout=5000, **kwargs) -> V:
```
eivindjahren, line 409: The type hint should have been different and add an assert:
```python
 wait_for_attribute(gui_attribute: Callable[[], V | None], qtbot: QtBot, timeout=5000, **kwargs) -> V:
    qtbot.waitUntil(lambda: gui_attribute() is not None, timeout=timeout, **kwargs)
    t = gui_attribute()
    assert t is not None
    return t
```
in which case this method becomes quite a lot more useful as inline type hints will not complain about the possiblity of it being none.
Then you would call it like this:

```python
 wait_for_attribute(lambda: run_dialog.fail_msg_box, qtbot)
```

SAKavli: Is this something we want instead of just `wait_until` ?

eivindjahren: Probably not :)

eivindjahren (approved):

PR #12296, opened 2025-11-14
Set esmda panel active realizations string to initial active realizations string when rerun is toggled

### LINKED ISSUE #12182: ES_MDA "Active Realizations" with design matrix with  non-continous realizations wrongness

I made a poly_design case with four realizations, but not continous realization numbers.
Ran it with ensemble experiment
When moving to ES_MDA, the validation complains (rightly so), because the active realization input gets set to 0-max (screenshot)

Image

Leave blank, or fill inn the "allowed" default (all realizations)

NOTE: When re-opening ert, this does NOT happen

### Steps to reproduce

```Python
1. Make design matrix case with non-continuous realization numbers
2. Run ensemble experiment
3. Move to ES_MDA,observe that realizations 0-max is in the active reals field. This fails validation (correctly so)

Extra: Restart ert, the problem is gone
```

- [ ] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

SAKavli:
Should the active realizations be e.g 0-1, 3-4 initally and restart_run_toggled is called in esmda, the value of the active realizations box is reset to 0-4, but this yields an error as realization 2 is inactive.

This change will instead reset the value to what it was initially.

**Issue**
Resolves #12182 

Before:

https://github.com/user-attachments/assets/dcdcc8b5-5aab-48da-b987-02ac2e0b78fb

After:

https://github.com/user-attachments/assets/a1c9e27c-970f-4c29-a5a5-e3263b954dca

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/gui/simulation/multiple_data_assimilation_panel.py
+++ b/src/ert/gui/simulation/multiple_data_assimilation_panel.py
@@ -135,7 +135,7 @@ class MultipleDataAssimilationPanel(ExperimentConfigPanel):
         self._active_realizations_field.model.setValueFromMask(  # type: ignore
             active_realizations
         )
-        self.initial_active_realizations = active_realizations
+        self._initial_active_realizations = active_realizations
 
         self._ensemble_selector = EnsembleSelector(notifier)
         self._previous_ensemble_realizations_validator = EnsembleRealizationsArgument(
@@ -267,7 +267,7 @@ class MultipleDataAssimilationPanel(ExperimentConfigPanel):
                 self._new_ensemble_realizations_validator
             )
             self._active_realizations_field.model.setValueFromMask(  # type: ignore
-                self.initial_active_realizations
+                self._initial_active_realizations
             )
 
     def _createInputForWeights(self, layout: QFormLayout) -> None:


src/ert/gui/simulation/multiple_data_assimilation_panel.py
```diff
@@ -135,6 +135,8 @@ def __init__(
         self._active_realizations_field.model.setValueFromMask(  # type: ignore
             active_realizations
         )
+        self.initial_active_realizations = active_realizations
```
HakonSohoel: Could maybe be a private variable since it is only for internal use: `self.initial_active_realizations` -> `self._initial_active_realizations`

HakonSohoel (approved):

src/ert/gui/simulation/multiple_data_assimilation_panel.py
SAKavli: Good call, that makes sense!

PR #12298, opened 2025-11-17
Remove installable_jobs hook for fmsteps

yngve-sk:
Since https://github.com/equinor/scout/issues/1290 is now resolved, we can remove this. All external plugins should now use installable_forward_model_steps hook instead.


berland:
Test also in komodenv.

yngve-sk:
Tested in komodoenv, passes tests, can run poly, and everest eightcells. Can also open ERT flow example with STEA invocation

Screenshot 2025-11-17 at 09 58 49

PR #12300, opened 2025-11-17
Add required keywords specification to plugin fmsteps

yngve-sk:
Plugin authors have requested for this to be
added, in order to not manually write code to
validate that required keywords are provided.


berland:
I do not think so yet, but could there be ill-configured plugins out in the wild that now might fail?

yngve-sk:
> I do not think so yet, but could there be ill-configured plugins out in the wild that now might fail?
 
https://github.com/search?q=org%3Aequinor%20required_keywords%3D&type=code no hits so far.

berland (approved):
Lets try!

PR #12317, opened 2025-11-18
Export all well summary keys by default

### LINKED ISSUE #12316: Everest: Export summary data for all wells
Currently Everest uses the `summary` results type in its forward model specification to indicate that summary results must be exported. The code for this relies on retrieving well names from the configuration. This blocks attempts to remove the `wells` section. Instead we should just export all wells found. so the `wells` section will not be needed for this.

verveerpj:
**Issue**
Resolves #12316

**Approach**
Replace well names with `*`.

      --exec 'just rapid-tests'`)

## When applicable


frode-aarstad (approved):

src/everest/simulator/everest_to_ert.py
```diff
@@ -25,27 +24,19 @@ def extract_summary_keys(ever_config: EverestConfig) -> list[str]:
 
     data_keys = everest.simulator.DEFAULT_DATA_SUMMARY_KEYS
     field_keys = everest.simulator.DEFAULT_FIELD_SUMMARY_KEYS
-    well_sum_keys = everest.simulator.DEFAULT_WELL_SUMMARY_KEYS
+    well_keys = [
+        f"{sum_key}:*" for sum_key in everest.simulator.DEFAULT_WELL_SUMMARY_KEYS
+    ]
```
andreas-el: Can you check usage of
`_EXCLUDED_TARGET_KEYS`, `_DEFAULT_WELL_TARGET_SUMMARY_KEYS`, `DEFAULT_WELL_SUMMARY_KEYS`, `_DEFAULT_WELL_SUMMARY_KEYS` ?
Perhaps some of this can be simplified even more?

This processing done here;
```
well_keys = [
        f"{sum_key}:*" for sum_key in everest.simulator.DEFAULT_WELL_SUMMARY_KEYS
    ]
```
seems like it is also done in `test_res_initialization.py` ?

andreas-el (changes requested):
Can you check usage for defines in __init__.py ?

src/everest/simulator/everest_to_ert.py
verveerpj: I moved the function to the everest run model and simplified it a bit. The test is actually just redoing what the implementation does, so that is not a good test, I removed it. I think the required testing is already done by other tests.

PR #12318, opened 2025-11-18
Add rft internalization

eivindjahren:
image

image

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/gui/tools/plot/plottery/plots/ensemble.py
+++ b/src/ert/gui/tools/plot/plottery/plots/ensemble.py
@@ -86,7 +86,7 @@ class EnsemblePlot:
         if len(data) == 1 and not style.marker:
             style.marker = "."
 
-        if plot_config.inverted_axes:
+        if plot_config.depth_y_axis:
             x = data.to_numpy()
             y = data.index.to_numpy()
             axes.yaxis.set_inverted(True)
--- a/src/ert/gui/tools/plot/plot_window.py
+++ b/src/ert/gui/tools/plot/plot_window.py
@@ -341,7 +341,7 @@ class PlotWindow(QMainWindow):
             ):
                 plot_context.setXLabel(key.split(":")[-1])
                 plot_context.setYLabel("TVD")
-                plot_context.inverted_axes = True
+                plot_context.depth_y_axis = True
                 for ekey, data in list(ensemble_to_data_map.items()):
                     ensemble_to_data_map[ekey] = data.interpolate(
                         method="linear", axis="columns"
--- a/src/ert/gui/tools/plot/plottery/plot_config.py
+++ b/src/ert/gui/tools/plot/plottery/plot_config.py
@@ -88,7 +88,7 @@ class PlotConfig:
 
         self._std_dev_factor = 1  # sigma 1 is default std dev
 
-        self.inverted_axes = False
+        self.depth_y_axis = False
 
     def getNumberOfColors(self) -> int:
         return len(self._line_color_cycle_colors)
--- a/src/ert/gui/tools/plot/plottery/plot_context.py
+++ b/src/ert/gui/tools/plot/plottery/plot_context.py
@@ -48,12 +48,12 @@ class PlotContext:
         self._log_scale = False
 
     @property
-    def inverted_axes(self) -> bool:
-        return self._plot_config.inverted_axes
+    def depth_y_axis(self) -> bool:
+        return self._plot_config.depth_y_axis
 
-    @inverted_axes.setter
-    def inverted_axes(self, value: bool) -> None:
-        self._plot_config.inverted_axes = value
+    @depth_y_axis.setter
+    def depth_y_axis(self, value: bool) -> None:
+        self._plot_config.depth_y_axis = value
 
     def plotConfig(self) -> PlotConfig:
         return self._plot_config
--- a/src/ert/gui/tools/plot/plottery/plots/statistics.py
+++ b/src/ert/gui/tools/plot/plottery/plots/statistics.py
@@ -126,11 +126,11 @@ def _plotPercentiles(
     axes: Axes, plot_config: PlotConfig, data: DataFrame, ensemble_label: str
 ) -> None:
     style = plot_config.getStatisticsStyle("mean")
-    if plot_config.inverted_axes:
+    if plot_config.depth_y_axis:
         axes.yaxis.set_inverted(True)
 
     def xy_order(x: Any, y: Any) -> tuple[Any, Any]:
-        if plot_config.inverted_axes:
+        if plot_config.depth_y_axis:
             return (y, x)
         else:
             return (x, y)
@@ -166,7 +166,7 @@ def _plotPercentiles(
         data["std+"].values,
         data["std-"].values,
         0.5,
-        plot_config.inverted_axes,
+        plot_config.depth_y_axis,
     )
 
     style = plot_config.getStatisticsStyle("min-max")
@@ -177,7 +177,7 @@ def _plotPercentiles(
         data["Maximum"].values,
         data["Minimum"].values,
         0.5,
-        plot_config.inverted_axes,
+        plot_config.depth_y_axis,
     )
 
     style = plot_config.getStatisticsStyle("p10-p90")
@@ -188,7 +188,7 @@ def _plotPercentiles(
         data["p90"].values,
         data["p10"].values,
         0.5,
-        plot_config.inverted_axes,
+        plot_config.depth_y_axis,
     )
 
     style = plot_config.getStatisticsStyle("p33-p67")
@@ -199,7 +199,7 @@ def _plotPercentiles(
         data["p67"].values,
         data["p33"].values,
         0.5,
-        plot_config.inverted_axes,
+        plot_config.depth_y_axis,
     )
 
 


src/ert/config/rft_config.py
```diff
@@ -0,0 +1,138 @@
...
+                lambda: defaultdict(list)
+            )
+            for rft in rfts:
+                for expected in ["WELL", "DATE", "PROPERTIES"]:
+                    if expected not in rft:
+                        raise ConfigValidationError.with_context(
+                            f"Rft need to specify {expected}.", rft
+                        )
+                well = rft["WELL"]
+                props = rft["PROPERTIES"].split(",")
```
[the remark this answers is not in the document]
eivindjahren: Added support for this, but it is still a bit clunky with the ert config file format.

src/ert/dark_storage/endpoints/responses.py
```diff
@@ -133,13 +133,13 @@ def data_for_response(
         return pd.DataFrame()
 
     if response_type == "summary":
-        summary_data = ensemble.load_responses(
+        rft_data = ensemble.load_responses(
```
SAKavli: Is this change intentional? Is RFT replacing summaries?

eivindjahren: Not intentinal. Fixed now.

test-data/ert/rft_example/resources/SPE1.DATA.jinja2
```diff
@@ -0,0 +1,447 @@
...
+
+PERMX
+	100*500 100*{{100 * parameters.FIELD_PROPERTIES.X_MID_PERMEABILITY}} 100*200 /
+
+PERMY
+-- Equal to PERMX
+	100*500 100*50 100*200 /
+
+PERMZ
+	100*500 100*{{100 * parameters.FIELD_PROPERTIES.X_MID_PERMEABILITY}} 100*200 /
```
HakonSohoel: Should probably be parameters.FIELD_PROPERTIES.Z_MID_PERMEABILITY

eivindjahren: Fixed

src/ert/config/rft_config.py
```diff
@@ -0,0 +1,138 @@
...
+                        "time": [time],
+                        "depth": [fetched[well, time]["DEPTH"]],
+                        "values": [vals],
+                    }
+                )
+                for (well, time), inner_dict in fetched.items()
+                for prop, vals in inner_dict.items()
+                if prop != "DEPTH"
+            ]
+        )
```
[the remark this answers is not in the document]
eivindjahren: Added test for this

src/ert/config/rft_config.py
```diff
@@ -0,0 +1,208 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+_ALPHABET = list(set(string.punctuation) - set("_-*"))
+
+
+def _separator() -> str:
+    """Quick 'uuid' used for separator in matching rft keys"""
+    return "".join(random.choices(_ALPHABET, k=8))
```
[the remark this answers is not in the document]
eivindjahren: Changed it to "\x31"

tests/ert/unit_tests/config/test_rft_config.py
```diff
@@ -0,0 +1,253 @@
...
+    buffer.seek(0)
+
+    def mock_open(*args, **kwargs):
+        path = args[0] if args else kwargs.get("file")
+        if path is not None and str(path).endswith(filename):
+            return buffer
+        else:
+            return original_open(*args, **kwargs)
+
+    mocker.patch("builtins.open", mock_open)
```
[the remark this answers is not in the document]
eivindjahren: Fixed

src/ert/config/rft_config.py
```diff
@@ -0,0 +1,208 @@
...
+                        _translate(well),
+                        _translate(time),
+                        _props_matcher(props),
+                    )
+                )
+                + ")"
+                for well, inner_dict in self.data_to_read.items()
+                for time, props in inner_dict.items()
+            )
+        )
```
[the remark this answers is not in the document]
eivindjahren: This means we would have to consider invalidating the cache also. The computation is probably not heavy enough to consider.

src/ert/gui/tools/plot/plottery/plots/ensemble.py
```diff
@@ -86,9 +86,16 @@ def _plotLines(
...
+            x = data.index.to_numpy()
+            axes.yaxis.set_inverted(True)
+        else:
+            x = data.to_numpy()
+            y = data.index.to_numpy()
         lines = axes.plot(
-            data.index.to_numpy(),
-            data.to_numpy(),
+            y,
+            x,
```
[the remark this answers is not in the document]
eivindjahren: Fixed

src/ert/config/rft_config.py
```diff
@@ -0,0 +1,216 @@
...
+                filter_on=None,
+                finalized=self.has_finalized_keys,
+            )
+            for response_key in self.keys
+        ]
+
+    @property
+    def expected_input_files(self) -> list[str]:
+        base = self.input_files[0]
+        if base.lower().endswith(".data"):
```
berland: (this would parse faster mentally with `base.upper().endswith(".DATA")`

src/ert/config/rft_config.py
```diff
@@ -0,0 +1,216 @@
...
+            # For backwards compatibility, it is
+            # allowed to give REFCASE and ECLBASE both
+            # with and without .DATA extensions
+            base = base[:-5]
+
+        return [f"{base}.RFT"]
+
+    def read_from_file(self, run_path: str, iens: int, iter_: int) -> pl.DataFrame:
```
berland, line 54: Can we restrict `run_path` to be of `pathlib.Path` type?

src/ert/config/rft_config.py
```diff
@@ -0,0 +1,216 @@
...
+            fm_steps = config_dict.get(ConfigKeys.FORWARD_MODEL, [])
+            names = [fm_step[0] for fm_step in fm_steps]
+            simulation_step_exists = any(
+                any(sim in _name.lower() for sim in ["eclipse", "flow"])
+                for _name in names
+            )
+            if not simulation_step_exists:
+                ConfigWarning.warn(
+                    "Config contains a RFT key but no forward model "
+                    "steps known to generate rft files"
```
berland: steps -> step

src/ert/config/rft_config.py
```diff
@@ -0,0 +1,216 @@
...
+    has_finalized_keys: bool = False
+    data_to_read: dict[str, dict[str, list[str]]] = Field(default_factory=dict)
+
+    @property
+    def metadata(self) -> list[ResponseMetadata]:
+        return [
+            ResponseMetadata(
+                response_type=self.name,
+                response_key=response_key,
+                filter_on=None,
```
yngve-sk, line 37: (hope some verbosity is OK here)

Forgot to post this after in-person discussion earlier today. I recall you said something along the lines of "we have to do the split anyway because __". 

Not really seeing the benefit of encoding time into response key (similar to old deprecated pattern we used for gendata with report steps) unless it is absolutely necessary, and somehow more urgent to get this in than to think a bit more about the design.

Suggest checking some more if we can use this instead of encoding the `time` into `response_key`. With the `filter_on` it should give the same behaviour. Prop can stay and makes sense to encode into the (Y-axis is always TVD?).

Furthermore: How slow/time consuming would it be to migrate this if we add the split later? One dataset per real, so 100-200x real-world sized datasets, if the dataset is real-world sized. Can easily check by just creating some parquets and do a dummy migration w polars.

src/ert/config/rft_config.py
yngve-sk: Suggest considering that in separate PR, as now it is a `str´ in the superclass method (`ResponseConfig`)

src/ert/config/rft_config.py
```diff
@@ -0,0 +1,216 @@
...
+                }
+            )
+        matcher = re.compile(
+            "|".join(
+                "("
+                + re.escape(_SEP).join(
+                    (
+                        _translate(well),
+                        _translate(time),
+                        _props_matcher(props),
```
yngve-sk, line 99: Consider moving these two inside this method if they are only used here

src/ert/config/rft_config.py
```diff
@@ -0,0 +1,216 @@
...
+                )
+                + ")"
+                for well, inner_dict in self.data_to_read.items()
+                for time, props in inner_dict.items()
+            )
+        )
+        try:
+            with RFTReader.open(f"{run_path}/{filename}") as rft:
+                for entry in rft:
+                    for t in entry:
```
yngve-sk: It is a bit unclear what `t` is, t is a property like pressure etc?

src/ert/config/rft_config.py
```diff
@@ -0,0 +1,216 @@
...
+        return [f"{base}.RFT"]
+
+    def read_from_file(self, run_path: str, iens: int, iter_: int) -> pl.DataFrame:
+        filename = substitute_runpath_name(self.input_files[0], iens, iter_)
+        if filename.lower().endswith(".data"):
+            # For backwards compatibility, it is
+            # allowed to give REFCASE and ECLBASE both
+            # with and without .DATA extensions
+            filename = filename[:-5]
+        fetched: dict[tuple[str, date], dict[str, npt.NDArray[Any]]] = defaultdict(dict)
```
yngve-sk: The `Any` within the NDArray is truly any? Could it be something else than a float?

src/ert/dark_storage/endpoints/responses.py
```diff
@@ -151,6 +151,19 @@ def data_for_response(
...
+    if response_type == "rft":
+        return (
+            ensemble.load_responses(
+                response_key,
+                tuple(realizations_with_responses),
+            )
+            .rename({"realization": "Realization"})
+            .drop(["response_key", "time"])
+            .to_pandas()
+            .pivot(index="Realization", columns="depth", values="values")
```
yngve-sk, line 163: Could you somehow document / add a comment on the form of the dataset before and after pivot? I'm still installing OPM flow so can't just run&inspect😅, but looks like before could be one row per real, each column is a depth, and the cells hold the values

``` 
Realization | 1000 | 1020 | 1040 | ...
----------- | ---- | ---- | ---- | ----
     0      | 0.21 | 0.22 | 0.20 | ...
     1      | 0.19 | 0.20 | 0.18 | ...
     2      | 0.22 | 0.23 | 0.21 | ...
     ...    |  ... |  ... |  ... | ...
```

and after
```
index | 1000 | 1020 | 1040 | ...
----- | ---- | ---- | ---- | ----
  0   | 0.21 | 0.22 | 0.20 | ...
  1   | 0.19 | 0.20 | 0.18 | ...
  2   | 0.22 | 0.23 | 0.21 | ...
```

src/ert/config/rft_config.py
yngve-sk: Question: Will entries maybe have different properties? That would explain/justify some of this logic a bit more. Would add a comment if yes. 
```
entry 
Out[14]: <resfo_utilities._rft_reader.RFTEntry at 0x14fa4afd0>
dict(entry)
Out[15]: 
{'PRESSURE': array([101., 201.], dtype='>f4'),
 'SWAT': array([0.01, 0.01], dtype='>f4'),
 'SGAS': array([0.01, 0.01], dtype='>f4'),
 'DEPTH': array([21., 31.], dtype='>f4')}
```

eivindjahren: Yes, entries can have different properties.

src/ert/dark_storage/endpoints/responses.py
eivindjahren: I added a test for the visualization that you can put a breakpoint in

src/ert/config/rft_config.py
eivindjahren: Its going to np.float32, so I will changed it.

src/ert/config/rft_config.py
eivindjahren: It is the property like pressure, renaming to "property" would overshadow builtin so ruff doesn't like that. Renamed to "rft_property".

src/ert/config/rft_config.py
eivindjahren: Sure, I will try to put the time into filter_on instead. Note that we only have the time after we have internalized as users may put `time:*` meaning all times, whatever is in the file.

eivindjahren: Quickly looked into this and it seems that it doesn't really fit as we may have different set of days for each well so the 

https://github.com/equinor/ert/blob/4f2997bf912ec8bde752f92f9171a28f58aeead7/src/ert/gui/tools/plot/plot_api.py#L189-L191

logic doesn't work. We may have the RFT keys `WELL:2000-01-01:PRESSURE` and `WELL2:2000-02-01:SWAT` but no other keys.

src/ert/config/rft_config.py
eivindjahren: Fixed

src/ert/config/rft_config.py
eivindjahren: Done

src/ert/config/rft_config.py
```diff
@@ -0,0 +1,216 @@
...
+                {
+                    "response_key": [],
+                    "time": [],
+                    "depth": [],
+                    "values": [],
+                }
+            )
+
+        try:
+            return pl.concat(
```
yngve-sk: Suggest rewriting this: The double for loop with explode is AI reproducible and a bit hard to follow, sugest something along these lines (this passes the rft tests), also without try-catch:
```
        dfs = []

        for (well, time), inner_dict in fetched.items():
            wide = pl.DataFrame({k:pl.Series(v.astype("<f4")) for k, v in inner_dict.items()})

            if wide.columns == ["DEPTH"]:
                continue

            if "DEPTH" not in wide.columns:
                raise InvalidResponseFile(f"Could not find DEPTH in RFTFile {filename}")

            # Unpivot all columns except DEPTH
            long = wide.unpivot(
                index="DEPTH",  # keep depth as column
                # turn other prop values into response_key col
                variable_name="response_key",
                value_name="values"  # put values in own column
            ).rename({"DEPTH":"depth"})

            # Add wellname prefix to response_keys
            long = long.with_columns(
                (pl.lit(f"{well}:{time.isoformat()}:") + pl.col("response_key"))
                .alias("response_key"), pl.lit(time).alias("time")
            )

            dfs.append(long.select("response_key","time","depth","values"))

        return pl.concat(dfs)
```

src/ert/config/rft_config.py
yngve-sk: > different set of days for each well

OK then this is future work, I think ideally we would split the wells out too (also for summary), then it should work, but now we would have a mix of filtering by some cols and encoding some into the response key.

src/ert/config/rft_config.py
eivindjahren: I changed it, but its somewhat harder to follow in my opinion. But in any case it probably correctly handles the `wide.columns==["DEPTH"]` case so probably a good idea.

src/ert/dark_storage/endpoints/responses.py
```diff
@@ -151,6 +151,19 @@ def data_for_response(
...
         return data.astype(float)
 
+    if response_type == "rft":
+        return (
+            ensemble.load_responses(
+                response_key,
+                tuple(realizations_with_responses),
+            )
+            .rename({"realization": "Realization"})
+            .drop(["response_key", "time", "property", "grid_i", "grid_j", "grid_k"])
```
yngve-sk: Suggest doing a select instead of drop
`.select(["Realization","depth","values"])`

src/ert/config/rft_config.py
eivindjahren: The old code was just analogue to 

https://github.com/equinor/ert/blob/dc96f6f1b8b74d289aed02194d4b20739a76611c/src/ert/config/summary_config.py#L64-L71

src/ert/dark_storage/endpoints/responses.py
eivindjahren: Since I removed "property" and "grid_*" then it makes sense to use drop still? Changed it to select as it is more resilient to us adding columns.

src/ert/gui/tools/plot/plottery/plots/ensemble.py
```diff
@@ -86,9 +86,16 @@ def _plotLines(
         if len(data) == 1 and not style.marker:
             style.marker = "."
 
+        if plot_config.inverted_axes:
```
berland: Maybe `swap_axes` is more clear, `inverted_axes` could also mean they are reversed?

src/ert/dark_storage/endpoints/responses.py
yngve-sk: inclusion>exclusion for most of these cases yep

src/ert/gui/tools/plot/plottery/plots/ensemble.py
```diff
@@ -86,9 +86,16 @@ def _plotLines(
...
+            x = data.to_numpy()
+            y = data.index.to_numpy()
+            axes.yaxis.set_inverted(True)
```
berland, line 92: yes, `inverted` here means reverse the direction, so we should not mix usage of the term.

berland: We are reversing the direction on the y-axis because we are working subsurface, does that maybe warrant a comment?

src/ert/gui/tools/plot/plottery/plots/ensemble.py
eivindjahren: Its kind of two things in one, which is not great. We both invert the y axis and swap x and y. How about "tvd_as_y_axis"? Or split in two: `plot_config.inverted_y_axis` and `plot_config.swap_axes` ?

src/ert/config/rft_config.py
yngve-sk: Hm, that could be simplified somehow likely, the repeat of the time map series doesn't look too good 😅

I still think single for loop + unpivot is easier to follow than double for loop comprehension with an if + explode is 🤠

src/ert/gui/tools/plot/plottery/plots/ensemble.py
eivindjahren: Changed to `depth_y_axis` .

yngve-sk (approved):
Very nice! 🪨🪨🪨📊📊📊

PR #12319, opened 2025-11-18
Add action linter to pre-commit

eivindjahren:
Also Removes running actionlint as separate workflow as it will be ran in the "style" workflow (which runs pre-commit).
Also adds running pre-commit with pre-push in GHActions. This is to run the rapid-tests to ensure that it does not time out.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -51,7 +51,11 @@ jobs:
       run: shellcheck -s bash ci/testkomodo.sh
       shell: bash
 
-    - name: Run style check
+    - uses: ./.github/actions/install_dependencies_qt
+      with:
+        os: ubuntu-latest
+
+    - name: Run pre-commit
       if: ${{ always() }}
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -11,6 +11,11 @@ repos:
     -   id: trailing-whitespace
         exclude: test-data/ert/eclipse/parse/ERROR.PRT # exact format is needed for testing
 
+- repo: https://github.com/rhysd/actionlint
```
SAKavli, line 14: Is this third party? I suppose we dont mind since we are open source.

eivindjahren: It is third-party but its the same third party that previously was in github actions so this only changes which machine it runs on. It is also [listed by pre-commit](https://pre-commit.com/hooks.html) so well-known.

Issue #11357, opened 2025-07-11
Simplify everserver/experiment server booting


yngve-sk:
Overall motivation: Make dark storage boot understandable, less code, more readable, remove redundant generic. Easier to make changes -> possibly decreased boot time etc

It is now too generic and dependent on the creation of the `storage_server.json` file (which takes quite some mental gymnastics to successfully locate).

Dark storage booting is really only starting up this FastAPI server. It should be possible to replace this call (from `src/everest/detached/client.py`) 
```
driver = create_driver(config.server.queue_system)  # type: ignore
    try:
        args = [
            "--output-dir",
            str(config.output_dir),
            "--logging-level",
            str(logging_level),
            "--traceparent",
            str(get_traceparent()),
        ]
        poll_task = asyncio.create_task(driver.poll(), name="poll_task")
        await driver.submit(
            0, "everserver", *args, name=f"{Path(config.config_file).stem}-server"
        )
```

with something that just invokes the dark storage boot executable. It should only need to take in the same arguments as everserver boot does. I.e., just replace `everserver` with the name of the new dark storage executable (`ert_server`). Booting this executable should fire up the following high lvl main endpoints (same as currently in dark storage): 
(1) Readonly endpoints to look at responses/parameters/observations, to inspect experiments / ensembles
(2) Endpoints to manage running of experiments

(naming/organization of endpoints is very up for discussion here, could be deferred to another issue)
```
ert_server
ert_server/experiment_server
ert_server/storage_server
```

```
app = FastAPI(
    title="Dark Storage API",
    version="0.1.0",
    debug=True,
    default_response_class=JSONResponse,
    lifespan=lifespan,
)
```
and it must also create a file/folder with some info for how clients can connect to it.

When this is complete, new followup issue(s) should be created, to move this closer to the ideal described above.

For mocking HTTP request flakiness, we could also consider having a testing option of running in the same process, ref https://github.com/equinor/ert/pull/11781#issuecomment-3279777191

PR #12326, opened 2025-11-18
Replace StorageService with less generic port of dark storage

yngve-sk:
This PR simply ports it to a less generic version, intended to be easier to follow. It will still be subject to further rewriting after this PR. Towards simplifying dark storage / experiment server boot. 

**Issue**
Towards  #11357 

      --exec 'just rapid-tests'`)

### DIFF
--- a/src/ert/services/ert_server.py
+++ b/src/ert/services/ert_server.py
@@ -224,11 +224,13 @@ class ErtServer:
                     and storage_server_path.stat().st_size > 0
                 ):
                     with (path / "storage_server.json").open() as f:
-                        return ErtServer(
-                            storage_path=str(path),
-                            connection_info=json.load(f),
-                            logging_config=logging_config,
-                        )
+                        storage_server_content = json.load(f)
+
+                    return ErtServer(
+                        storage_path=str(path),
+                        connection_info=storage_server_content,
+                        logging_config=logging_config,
+                    )
 
                 sleep(1)
                 t += 1


src/ert/services/ert_server.py
```diff
@@ -0,0 +1,308 @@
...
+
+class ErtServer:
+    _instance: ErtServer | None = None
+
+    def __init__(
+        self,
+        storage_path: str,
+        timeout: int = 120,
+        parent_pid: int | None = None,
+        conn_info: Mapping[str, Any] | Exception | None = None,
```
berland: This should perhaps be spelt out as `connection_info`

src/ert/services/ert_server.py
```diff
@@ -0,0 +1,308 @@
...
+        verbose: bool = False,
+        logging_config: str | None = None,  # Only used from everserver
+    ) -> None:
+        if timeout is None:
+            timeout = 120
+
+        self._storage_path = storage_path
+        self._conn_info: ConnInfo = conn_info
+        self._conn_info_event = threading.Event()
```
berland: can we make this variable name a little bit more self-documenting, indicating when and/or what the purpose of the event is?

src/ert/services/ert_server.py
```diff
@@ -0,0 +1,308 @@
...
+            run_storage_main_cmd.append("--verbose")
+
+        if self._conn_info is not None:
+            if isinstance(conn_info, Mapping) and "urls" not in conn_info:
+                raise KeyError("urls not found in conn_info")
+
+            self._conn_info_event.set()
+            self._thread_that_starts_server_process = None
+        else:
+            self._thread_that_starts_server_process = _Proc(
```
berland: Hmm, I had a hope that we would be able to do without `_Proc` ? 🤷🏻

yngve-sk: We can do it some other way for sure, but suggest leaving it for a separate PR to not do too much at once. It is the current way we are communicating between the `_storage_main.py` executable and upstream, via pipes, so I think it could be a completely isolated discussion not mixed with that of this port, if that sounds OK?

src/ert/services/ert_server.py
```diff
@@ -0,0 +1,312 @@
...
+        parent_pid: int | None = None,
+        connection_info: ErtServerConnectionInfo | Exception | None = None,
+        verbose: bool = False,
+        logging_config: str | None = None,  # Only used from everserver
+    ) -> None:
+        if timeout is None:
+            timeout = 120
+
+        self._storage_path = storage_path
+        self._connection_info: ErtServerConnectionInfo | Exception | None = (
```
yngve-sk, line 55: suggest making separate issue to have this be strictly typed to only `ErtServerConnectionInfo`, handling errors outside of initialization

src/ert/services/ert_server.py
```diff
@@ -0,0 +1,310 @@
...
+
+        if parent_pid is not None:
+            run_storage_main_cmd += ["--parent_pid", str(parent_pid)]
+
+        if verbose:
+            run_storage_main_cmd.append("--verbose")
+
+        if self._connection_info is not None:
+            if isinstance(connection_info, Mapping) and "urls" not in connection_info:
+                raise KeyError("urls not found in conn_info")
```
berland: This is an error message that refers to a variable name, maybe improve it.

src/ert/services/ert_server.py
```diff
@@ -0,0 +1,310 @@
...
+            if traceparent is not None:
+                run_storage_main_cmd += ["--traceparent", traceparent]
+
+        if parent_pid is not None:
+            run_storage_main_cmd += ["--parent_pid", str(parent_pid)]
+
+        if verbose:
+            run_storage_main_cmd.append("--verbose")
+
+        if self._connection_info is not None:
```
berland: This is an `if` statment testing whether the server is already running? 

Since starting the server occurs first, it should come in the first if-block. Then maybe something can be done to make it more readable, making the mental link `_connection_info is not None` <-> `server already running` more obvious.

src/ert/services/ert_server.py
```diff
@@ -0,0 +1,310 @@
...
+    ) -> ErtServerContext:
+        try:
+            service = cls.connect(
+                project=project or Path.cwd(), timeout=0, logging_config=logging_config
+            )
+            # Check the server is up and running
+            _ = service.fetch_url()
+        except (TimeoutError, JSONDecodeError, KeyError) as e:
+            logging.getLogger(__name__).warning(
+                "Failed locating existing storage service due to "
```
berland, line 127: (I know this is inherited from _base_service..) 

When we have TimeoutError, this is normally the happy-path, and it should not trigger a warning (and avoid the word 'failed')

src/ert/services/ert_server.py
```diff
@@ -0,0 +1,310 @@
...
+            return cls.start_server(
+                project=project, timeout=timeout, logging_config=logging_config
+            )
+        except PermissionError as pe:
+            logging.getLogger(__name__).error(
+                f"{type(pe).__name__}: {pe}, cannot connect to storage service "
+                f"due to permission issues."
+            )
+            raise pe
+        return ErtServerContext(service)
```
berland: Move this `return ` to the end of first  `try`-block to make it closer to the code it applies to?

src/ert/services/ert_server.py
```diff
@@ -0,0 +1,310 @@
...
+                    f"Connecting to {url} got status: "
+                    f"{resp.status_code}, {resp.headers}, {resp.reason}, {resp.text}"
+                )
+                if resp.status_code == 200:
+                    self._url = url
+                    return str(url)
+
+            except requests.ConnectionError as ce:
+                logging.getLogger(__name__).info(
+                    f"Could not connect to {url}, but will try something else. "
```
berland, line 163: If we are at the end of the for-loop, this log message is wrong as there is nothing else to try.

src/ert/services/ert_server.py
```diff
@@ -0,0 +1,310 @@
...
+        path = Path(project)
+
+        # Wait for storage_server.json file to appear
+        try:
+            if timeout is None:
+                timeout = 240
+            t = -1
+            while t < timeout:
+                if (path / "storage_server.json").exists():
+                    with (path / "storage_server.json").open() as f:
```
berland, line 226: this pattern is probably susceptible to a race condition if the storage_server.json file is touched before written to. Should probably both check it exists and has non-zero content before returning ErtServer.

Unless we have some special meaning behind keeping the file descriptor `f` open during the construction of `ErtServer`, we should read the file first (using pathlib) and then make the ErtServer object.

src/ert/services/ert_server.py
yngve-sk: Removing the catch of the `TimeoutError` leads to hanging dark storage, we use the try-catch now to start the server. Suggest to create issue for fixing this in isolation.

src/ert/services/ert_server.py
yngve-sk: I think if there are multiple URLs it is trying to say we want to check the next, and the message is a bit vague yes. Suggest creating an issue for going over this loop and refining it.

src/ert/services/ert_server.py
yngve-sk: https://github.com/equinor/ert/issues/12457

src/ert/services/ert_server.py
yngve-sk: https://github.com/equinor/ert/issues/12458

src/ert/services/ert_server.py
yngve-sk: https://github.com/equinor/ert/issues/12459

src/ert/services/ert_server.py
yngve-sk: https://github.com/equinor/ert/issues/12460

PR #12331, opened 2025-11-19
Enforce style, typing/rapid-tests first to save CI minutes

andreas-el:
Restructure workflows such that:
- Typing and `pre-commit` style/testing are run first (in build_test.yml)
- Codecov and `build_test` workflow jobs depend on the two jobs

This could mitigate a lot of wasted CI minutes catching Style, Typing and lack of initial testing.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -20,8 +20,18 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
+  run-pre-commit:
+    uses: ./.github/workflows/run_pre_commit.yml
+
+  check-typing:
+    uses: ./.github/workflows/typing.yml
+
+  run-codspeed:
+    needs: [run-pre-commit, check-typing]
+    uses: ./.github/workflows/codspeed.yml
 
   build-wheels:
+    needs: [run-pre-commit, check-typing]
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -20,8 +20,25 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
+  check-style:
+    uses: ./.github/workflows/style.yml
+
+  check-typing:
+    uses: ./.github/workflows/typing.yml
+
+  run-rapid-tests:
```
eivindjahren: check-style now runs rapid-tests through pre-commit so need for step. Should probably rename style to `run_pre_commit` while we are at it.

eivindjahren (changes requested):
This would run rapid-tests twice. I think we should rename style -> run_pre_commit while we are at it to remove this confusion in the future.

.github/workflows/build_and_test.yml
andreas-el: Testing twice, isn't that twice as good? 🤔

.github/workflows/build_and_test.yml
```diff
@@ -20,8 +20,18 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
+  run-pre-commit:
+    uses: ./.github/workflows/style.yml
```
eivindjahren: Should rename the file also.

andreas-el: Let's do that too.

PR #12346, opened 2025-11-19
Avoid blocking IO calls in async code

### LINKED ISSUE #12325: Avoid blocking IO calls in asyncio code
The ruff rule ASYNC240: https://docs.astral.sh/ruff/rules/blocking-path-method-in-async-function/ is ignored in pyproject.toml

Using blocking IO calls in asyncio negates the effect of asyncio, and should be avoided. Solving this should start with the busiest asyncio places first. The ruff rule suggest to use anyio.Path or trio.Path.

berland:
Calling some os.path or pathlib.Path methods in an async function will block the entire event loop, preventing it from executing other tasks while waiting for the operation. This negates the benefits of asynchronous programming.

We can allow this for now in test code where performance is not critical.

**Issue**
Resolves #12325

**Approach**
Rewrite

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -12,6 +12,7 @@ from enum import StrEnum
 from pathlib import Path
 from typing import TYPE_CHECKING, assert_never
 
+import anyio
 from lxml import etree
 from opentelemetry.trace import Status, StatusCode
 
@@ -335,8 +336,12 @@ class Job:
 
         valid_checksums = [info for info in checksum.values() if "error" not in info]
 
-        # Wait for files in checksum
-        while not all(Path(info["path"]).exists() for info in valid_checksums):
+        async def all_paths_exist(paths: list[Path]) -> bool:
+            return all(
+                await asyncio.gather(*[anyio.Path(path).exists() for path in paths])
+            )
+
+        while not await all_paths_exist([info["path"] for info in valid_checksums]):
             if timeout <= 0:
                 break
             timeout -= DISK_SYNCHRONIZATION_POLLING_INTERVAL
@@ -344,17 +349,19 @@ class Job:
             await asyncio.sleep(DISK_SYNCHRONIZATION_POLLING_INTERVAL)
         async with checksum_lock:
             for info in valid_checksums:
-                file_path = Path(info["path"])
+                file_path = anyio.Path(info["path"])
                 expected_md5sum = info.get("md5sum")
-                if file_path.exists() and expected_md5sum:
-                    actual_md5sum = hashlib.md5(file_path.read_bytes()).hexdigest()
+                file_path_exists = await file_path.exists()
+                if file_path_exists and expected_md5sum:
+                    file_bytes = await file_path.read_bytes()
+                    actual_md5sum = hashlib.md5(file_bytes).hexdigest()
                     if expected_md5sum == actual_md5sum:
                         logger.debug(f"File {file_path} checksum successful.")
                     else:
                         logger.warning(
                             f"File {file_path} checksum verification failed."
                         )
-                elif file_path.exists() and expected_md5sum is None:
+                elif file_path_exists and expected_md5sum is None:
                     logger.warning(f"Checksum not received for file {file_path}")
                 else:
                     logger.error(f"Disk synchronization failed for {file_path}")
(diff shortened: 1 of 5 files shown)


tests/ert/unit_tests/scheduler/test_job.py
```diff
@@ -591,7 +591,7 @@ def true_after_delay(*args, **kwargs):
 
     # Mock st_mtime to be infinite
     mock_stat_result = SimpleNamespace(st_mtime=inf)
-    mocker.patch("pathlib.Path.stat", return_value=mock_stat_result)
+    mocker.patch("os.stat", return_value=mock_stat_result)
```
StephanDeHoop, line 594: The test pass, so I guess it works, but don't we have to patch `anyio.Path.stat` here?

src/everest/bin/everest_script.py
```diff
@@ -193,9 +195,18 @@ async def run_everest(options: argparse.Namespace) -> None:
...
         logger.info(f"Everest forward model contains job {job_name}")
 
-    if os.path.exists(options.config.simulation_dir) and any(
-        os.listdir(options.config.simulation_dir)
-    ):
+    async def directory_is_nonempty(path: Path) -> bool:
+        try:
+            async for _ in anyio.Path(path).iterdir():
+                return True
+        except OSError:  # Raised if the directory is empty
```
StephanDeHoop, line 202: I am not familiar with `anyio`, but I thought the normal `Path().iterdir()` raises `OSError` when the path doesn't exist or isn't a directory (https://docs.python.org/3/library/pathlib.html#pathlib.Path.iterdir), is this different when calling it through `anyio`?

src/ert/scheduler/job.py
```diff
@@ -346,15 +351,17 @@ async def _verify_checksum(
             for info in valid_checksums:
                 file_path = Path(info["path"])
                 expected_md5sum = info.get("md5sum")
-                if file_path.exists() and expected_md5sum:
-                    actual_md5sum = hashlib.md5(file_path.read_bytes()).hexdigest()
+                file_path_exists = await anyio.Path(file_path).exists()
```
StephanDeHoop: Nitpick: does it make sense to `file_path = anyio.Path(file_path)` and then `file_path.exists()` and `file_path.read_bytes()` later? (same below), or maybe you explicitly want to do it like this? (I am still a bit unfamiliar with async)

berland: will fix!

src/everest/bin/everest_script.py
berland: At this point we know that `path` exists, but not if it is a file. Will check what happens when it is a file.

tests/ert/unit_tests/scheduler/test_job.py
berland: They both trickle down to `os.stat` at the bottom, so we might as well just patch that instead.

src/everest/bin/everest_script.py
berland: It is caught in the pydantic validation if it is a file so it will not be possible at this point. Added a test to certify this.

src/ert/scheduler/job.py
```diff
@@ -12,6 +12,7 @@
 from pathlib import Path
 from typing import TYPE_CHECKING, assert_never
 
+import anyio
```
yngve-sk, line 15: It will "always" be installed because httpx/starlette depends on it, but if we have it explicitly used in our code we should have it in pyproject.toml?

berland: yes! Fixing.

tests/everest/entry_points/test_everest_entry.py
```diff
@@ -344,6 +345,35 @@ class ServerStatus:
...
+        "Warning from everest can only be triggered on absolute paths in config"
+    )
+    everest_config_with_defaults(
+        environment={"simulation_folder": existing_runpath.name},
+    ).dump("config.yml")
+
+    with pytest.raises(RuntimeError, match="runpath is nonempty"):
+        everest_entry(["config.yml"])
+
+    existing_runpath.cleanup()
```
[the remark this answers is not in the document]
berland: fixed.

tests/everest/entry_points/test_everest_entry.py
```diff
@@ -344,6 +345,35 @@ class ServerStatus:
...
+
+
+def test_that_everest_fails_when_runpath_is_a_file():
+    with tempfile.NamedTemporaryFile() as existing_runpath:
+        Path(existing_runpath.name).touch()
```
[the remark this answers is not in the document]
berland: fixed.

PR #12362, opened 2025-11-21
Check for migration before performing migrate step

### LINKED ISSUE #12065: Possible to open two ert instances towards the same storage

andreas-el:
**Issue**
Resolves #12065 

Check if migration is needed before trying to perform migration step

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/gui/main.py
+++ b/src/ert/gui/main.py
@@ -158,9 +158,8 @@ def _start_initial_gui_window(
     storage_path = None
     if ert_config is not None:
         try:
-            storage = open_storage(ert_config.ens_path, mode="r")
-            should_migrate = storage.check_migration_needed()
-            storage.close()
+            with open_storage(ert_config.ens_path) as read_storage:
+                should_migrate = read_storage.check_migration_needed()
 
             if should_migrate:
                 # Open in write mode to initialize the storage, so that


src/ert/storage/local_storage.py
```diff
@@ -298,13 +298,6 @@ def close(self) -> None:
 
         self._save_index()
 
-        if self.can_write:
```
berland, line 301: the docstring of this `close()` function needs an update

src/ert/storage/local_storage.py
```diff
@@ -278,12 +289,8 @@ def _acquire_lock(self) -> None:
...
     def close(self) -> None:
         """
-        Closes the storage, releasing any acquired locks and saving the index.
-
-        This method should be called to cleanly close the storage, especially
-        when it was opened in write mode. Failing to call this method may leave
-        a lock file behind, which would interfere with subsequent access to
-        the storage.
+        Closes the storage and saves the index.
+        This method should be called to cleanly close the storage.
```
yngve-sk: I think this comment should be put back to its original state, since we still release the lock.

src/ert/storage/local_storage.py
```diff
@@ -289,19 +289,13 @@ def _acquire_lock(self) -> None:
 
     def close(self) -> None:
         """
-        Closes the storage and saves the index.
-        This method should be called to cleanly close the storage.
+        Closes the storage and saves the index
```
yngve-sk: Docstring should be set back to original state.

src/ert/gui/main.py
```diff
@@ -158,9 +158,14 @@ def _start_initial_gui_window(
     storage_path = None
     if ert_config is not None:
         try:
-            # Open write to initialize the storage,so that
-            # dark storage can be mounted onto it
-            open_storage(ert_config.ens_path, mode="w").close()
+            storage = open_storage(ert_config.ens_path, mode="r")
```
yngve-sk: suggest doing a `with open_storage() as storage:` instead of explicit open/close.

src/ert/storage/local_storage.py
```diff
@@ -64,8 +64,6 @@ def __init__(
         self,
         path: str | os.PathLike[str],
         mode: Mode,
-        *,
-        ignore_migration_check: bool = False,
```
berland, line 68: Is this just an option that was never used pr. now and thus removed?

src/ert/gui/main.py
```diff
@@ -158,9 +158,14 @@ def _start_initial_gui_window(
...
+            should_migrate = storage.check_migration_needed()
+            storage.close()
+
+            if should_migrate:
+                # Open write to initialize the storage,so that
```
berland: "Open write" -> "Open in write mode"
",so" -> ", so"

src/ert/storage/local_storage.py
andreas-el: Yes. Did search for usage, found none. 0 value.

yngve-sk (commented):
Open Q: Should we do this check in the CLI also? (guessing maybe not?..) It will still auto migrate, suppose most users are using GUI though, but remember there being some bash scripts somewhere that invoke ERT, ref @berland ?

PR #12375, opened 2025-11-21
Remove get_runtime from RunModel API

### LINKED ISSUE #10974: Send ExperimentStart event instead of keeping track of start time
The gui can then respond to the event and find out what to do on its own.
ref https://github.com/equinor/ert/pull/10833#discussion_r2104523128

berland:
This allows removal of the start_time property of a RunModel, which defies serialization of the object, as it is a runtime property.

Instead of storing the starttime as a property, we send it as a StartEvent message, letting event clients be able to calculate the runtime themselves.

Logging of experiment runtime is kept, but with a bonus bugfix in that a message for every ensemble stating that the experiment is fixed, has been removed. Now we emit this log message only at the end, instead of 4 times in the case of the standard es_mda for example.

**Issue**
Resolves #10974 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/event.py
+++ b/src/ert/ensemble_evaluator/event.py
@@ -1,4 +1,5 @@
 from collections.abc import Mapping
+from datetime import datetime
 from typing import Any, Literal
 
 from pydantic import BaseModel, ConfigDict, field_serializer, field_validator
@@ -40,6 +41,11 @@ class FullSnapshotEvent(_UpdateEvent):
 
 class SnapshotUpdateEvent(_UpdateEvent):
     event_type: Literal["SnapshotUpdateEvent"] = "SnapshotUpdateEvent"
+
+
+class StartEvent(BaseModel):
+    event_type: Literal["StartEvent"] = "StartEvent"
+    timestamp: datetime
 
 
 class EndEvent(BaseModel):
--- a/src/ert/gui/simulation/run_dialog.py
+++ b/src/ert/gui/simulation/run_dialog.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+from datetime import datetime
 from pathlib import Path
 from queue import SimpleQueue
 from typing import assert_never, cast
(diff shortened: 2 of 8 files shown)


src/ert/gui/simulation/run_dialog.py
```diff
@@ -543,6 +548,9 @@ def _on_ticker(self) -> None:
     def _on_event(self, event: object) -> None:
         model = self._snapshot_model
         match event:
+            case StartEvent():
+                self._start_time = time.time()
+                print("Got StartEvent!!!")
```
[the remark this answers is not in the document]
berland: Fixed

src/ert/ensemble_evaluator/event.py
```diff
@@ -42,6 +42,10 @@ class SnapshotUpdateEvent(_UpdateEvent):
     event_type: Literal["SnapshotUpdateEvent"] = "SnapshotUpdateEvent"
 
 
+class StartEvent(BaseModel):
+    event_type: Literal["StartEvent"] = "StartEvent"
```
yngve-sk, line 47: I think we should have a timestamp attr here. If a client comes in and subscribes after the server has started, it will only have the "time of connecting to server", not the time of when the experiment started? When connecting to the events endpoint, the server will replay all old events to the client.

src/ert/run_models/run_model.py
```diff
@@ -399,9 +401,9 @@ def start_simulations_thread(
         def handle_captured_event(message: Warning | str) -> None:
             self.send_event(WarningEvent(msg=str(message)))
 
+        start_timestamp = datetime.now()
```
yngve-sk, line 404: Do we have some established agreement on what kind of way we would like to store/send dates? Would be inclined towards just doing a unix timestamp instead of this tbh.

berland: unix timestamp would give some extra accuracy potentially, but is not timezone aware. I thought we just as well could be timezone aware for these events. datetime.datetime is probably also a tad slower, but that will not matter for the StartEvent as it is sent only once

PR #12376, opened 2025-11-21
Optimize write_to_runpath for scalar parameters

xjules:
**Approach**

We do pre-load pd.Dataframe now avoid loading the file #parameters times for each realization.
There is still place to improve this though, mainly since we can pre-load all the realizations.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/run_models/_create_run_path.py
+++ b/src/ert/run_models/_create_run_path.py
@@ -16,6 +16,7 @@ from ert.config import (
     Field,
     ForwardModelStep,
     GenKwConfig,
+    ParameterCardinality,
     ParameterConfig,
     SurfaceConfig,
 )
@@ -115,16 +116,28 @@ def _generate_parameter_files(
         Returns the union of parameters returned by write_to_runpath for each
         parameter_config.
     """
+    # preload scalar parameters for this realization
+    keys = [
+        p.name
+        for p in parameter_configs
+        if p.cardinality == ParameterCardinality.multiple_configs_per_ensemble_dataset
+    ]
+    scalar_data: dict[str, float | str] = {}
+    if keys:
+        df = fs._load_scalar_keys(keys=keys, realizations=iens, transformed=True)
+        scalar_data = df.to_dicts()[0]
     exports: dict[str, dict[str, float | str]] = {}
-
     for param in parameter_configs:
         # For the first iteration we do not write the parameter
         # to run path, as we expect to read if after the forward
         # model has completed.
         if param.forward_init and iteration == 0:
             continue
-
-        export_values = param.write_to_runpath(Path(run_path), iens, fs)
+        export_values: dict[str, dict[str, float | str]] | None = None
+        if param.name in scalar_data:
+            export_values = {param.group_name: {param.name: scalar_data[param.name]}}
+        else:
+            export_values = param.write_to_runpath(Path(run_path), iens, fs)
         if export_values:
             for group, vals in export_values.items():
                 exports.setdefault(group, {}).update(vals)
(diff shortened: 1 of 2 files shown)


src/ert/run_models/_create_run_path.py
```diff
@@ -115,15 +115,22 @@ def _generate_parameter_files(
         Returns the union of parameters returned by write_to_runpath for each
         parameter_config.
     """
+    # preload scalar parameters for this realization
+    df = fs._load_scalar_keys(realizations=iens, transformed=True)
```
jonathan-eq: Can we preload scalar parameters for all realizations if we move this up another level?

src/ert/storage/local_ensemble.py
```diff
@@ -564,12 +564,13 @@ def _load_parameters_lazy(
 
     def _load_scalar_keys(
         self,
-        keys: list[str],
+        keys: list[str] | None = None,
```
jonathan-eq: What will happen here if keys is None? Will we just load the realization column?

src/ert/run_models/_create_run_path.py
xjules: I thought about that. But do we really need this level of optimization?

src/ert/storage/local_ensemble.py
xjules: no, it is all columns

src/ert/run_models/_create_run_path.py
jonathan-eq: Maybe not, this is a good step in the right direction nonetheless :)

src/ert/run_models/_create_run_path.py
```diff
@@ -115,16 +116,28 @@ def _generate_parameter_files(
...
+        for p in parameter_configs
+        if p.cardinality == ParameterCardinality.multiple_configs_per_ensemble_dataset
+    ]
+    scalar_data: dict[str, float | str] = {}
+    if keys:
+        df = fs._load_scalar_keys(keys=keys, realizations=iens, transformed=True)
+        scalar_data = df.to_dicts()[0]
     exports: dict[str, dict[str, float | str]] = {}
-
     for param in parameter_configs:
```
yngve-sk, line 130: Could we make this loop only go over the set of params that are not scalars? then just do a merge after, instead of the if inside the loop?

yngve-sk: Outline of suggestion (just to demonstrate the idea)
```python
    # preload scalar parameters for this realization
    keys = [
        p.name
        for p in parameter_configs
        if p.cardinality == ParameterCardinality.multiple_configs_per_ensemble_dataset
    ]
    exports: dict[str, dict[str, float | str]] = {}

    scalar_data: dict[str, float | str] = {}
    if keys:
        df = fs._load_scalar_keys(keys=keys, realizations=iens, transformed=True)
        scalar_data = df.to_dicts()[0]
        del scalar_data["realization"]
        exports |= scalar_data

    non_scalar_parameters = [p for p in parameter_configs if p.cardinality == ParameterCardinality.one_config_per_realization_dataset]
    for param in non_scalar_parameters:
        # For the first iteration we do not write the parameter
        # to run path, as we expect to read if after the forward
        # model has completed.
        if param.forward_init and iteration == 0:
            continue

        export_values = param.write_to_runpath(Path(run_path), iens, fs)
        if export_values:
            for group, vals in export_values.items():
                exports.setdefault(group, {}).update(vals)
        continue

    _value_export_txt(run_path, export_base_name, exports)
    _value_export_json(run_path, export_base_name, exports)
    return exports
```

xjules: scalars are still exported `{group: {key: value}}` so this won't work

yngve-sk: Can be future work, param groups does make it a bit less straight forward

yngve-sk (approved):

PR #12381, opened 2025-11-23
Make Ert warn users when overspending unrequested CPU resources

### LINKED ISSUE #12340: Display warning about overspending CPU to user in the GUI

SAKavli:
**Issue**
Resolves #12340

image

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -6,6 +6,8 @@ import threading
 import traceback
 from collections import defaultdict
 from collections.abc import Awaitable, Callable, Iterable, Sequence
+from dataclasses import dataclass
+from math import ceil
 from typing import Any, cast, get_args
 
 import zmq.asyncio
@@ -49,6 +51,13 @@ from .state import (
     ENSEMBLE_STATE_STOPPED,
 )
 
+
+@dataclass(order=True)
+class ParallelismViolation:
+    amount: float = 0
+    message: str = ""
+
+
 logger = logging.getLogger(__name__)
 
 EVENT_HANDLER = Callable[[list[SnapshotInputEvent]], Awaitable[None]]
@@ -74,6 +83,7 @@ class EnsembleEvaluator:
     # has misconfigured NUM_CPU in their config.
     ALLOWED_CPU_OVERSPENDING = 1.05
     MINIMUM_WALLTIME_SECONDS = 30  # Information is only polled every 5 sec
+    CPU_OVERSPENDING_WARNING_THRESHOLD = 1.50
 
     def __init__(
         self,
@@ -129,6 +139,7 @@ class EnsembleEvaluator:
             submit_sleep=self.ensemble._queue_config.submit_sleep,
             ens_id=self.ensemble.id_,
         )
+        self.max_parallelism_violation = ParallelismViolation()
 
     async def _publisher(self) -> None:
         heartbeat_interval = 0.1
@@ -657,6 +668,9 @@ class EnsembleEvaluator:
         """Produces a message warning about misconfiguration of NUM_CPU if
         so is detected. Returns an empty string if everything is ok."""
         allowed_overspending = self.ALLOWED_CPU_OVERSPENDING * num_cpu
+        overspending_warning_threshold = (
+            self.CPU_OVERSPENDING_WARNING_THRESHOLD * num_cpu
+        )
 
         start_time = fm_step.get(ids.START_TIME)
 
@@ -680,3 +694,18 @@ class EnsembleEvaluator:
                 f"with wall clock duration {duration:.1f} seconds, a factor of "
                 f"{parallelization_obtained:.2f}, while NUM_CPU was {num_cpu}."
             )
+            if parallelization_obtained > overspending_warning_threshold:
+                warning_msg = (
+                    "Overusage of CPUs detected!\n"
+                    f"Your experiment has used up to {ceil(parallelization_obtained)} "
+                    f"CPUs in step '{fm_step.get(ids.NAME)}', "
+                    f"while the Ert config has only requested {num_cpu}.\n"
+                    f"This means your experiment is consuming more CPU-resources than "
+                    f"requested and will slow down other users experiments.\n"
+                    f"We kindly ask you to set "
+                    f"NUM_CPU={ceil(parallelization_obtained)} in your Ert config."
+                )
+                self.max_parallelism_violation = max(
+                    self.max_parallelism_violation,
+                    ParallelismViolation(parallelization_obtained, warning_msg),
+                )
--- a/src/ert/run_models/run_model.py
+++ b/src/ert/run_models/run_model.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import asyncio
 import concurrent.futures
 import copy
-import dataclasses
 import functools
 import logging
 import os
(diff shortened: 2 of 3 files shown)


src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -129,6 +131,8 @@ def __init__(
             submit_sleep=self.ensemble._queue_config.submit_sleep,
             ens_id=self.ensemble.id_,
         )
+        self.highest_parallelization_obtained: float = 0
```
SAKavli: The idea is that this value remains 0 as long as there is no violation. When it is above 0, run_model will pick it up.

SAKavli:
I have tested it in Ert gui by hard coding violating values, but perhaps you could help me test this with a realistic use case, @berland ?

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -648,27 +655,50 @@ async def evaluate(
...
+                f"with wall clock duration {duration:.1f} seconds, a factor of "
+                f"{parallelization_obtained:.2f}, while NUM_CPU was {num_cpu}."
+            )
+            if (
+                parallelization_obtained > overspending_warning_threshold
+                and parallelization_obtained > self.highest_parallelization_obtained
+            ):
+                self.highest_parallelization_obtained = parallelization_obtained
+                self.cpu_violation_warnings_msg = (
+                    "Over usage of CPUs detected!\n"
```
berland: "Over usage" -> "Overusage" I think.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -648,27 +655,50 @@ async def evaluate(
...
+                    f"Your experiment has used up to {ceil(parallelization_obtained)} "
+                    f"CPUs in step '{fm_step.get(ids.NAME)}', "
+                    f"while the Ert config has only requested {num_cpu}.\n"
+                    f"This means your experiment is hogging resources you did not "
```
berland: I am a little bit concerned that not all users understand the word 'hogging' immediately. Maybe rewrite to "This means your experiment is consuming more CPU-resources than requested and will slow down other users experiments".

src/ert/ensemble_evaluator/evaluator.py
berland: Can't you always populate this variable? (now someone might find it strange to see it is zero in normal circumstances). If not, choose a different variable name.

src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -648,27 +655,50 @@ async def evaluate(
...
+                and parallelization_obtained > self.highest_parallelization_obtained
+            ):
+                self.highest_parallelization_obtained = parallelization_obtained
+                self.cpu_violation_warnings_msg = (
+                    "Overusage of CPUs detected!\n"
+                    f"Your experiment has used up to {ceil(parallelization_obtained)} "
+                    f"CPUs in step '{fm_step.get(ids.NAME)}', "
+                    f"while the Ert config has only requested {num_cpu}.\n"
+                    f"This means your experiment is consuming more CPU-resources than "
+                    f"requested and will slow down other users experiments\n"
```
berland: Missing a `.` at the end here.

Do we need the linebreaks, will not Qt add linebreaks?

SAKavli: Qt does add linebreaks to fit the message box. If you always want a line break, you must specify it by a newline. Example:

https://github.com/user-attachments/assets/efb480aa-9917-470f-b0b9-85f2da1ae97a

https://github.com/user-attachments/assets/52530b40-e1e6-4f47-b977-6daeb9e3ef87

I think the second one looks nice when it makes sense.

src/ert/ensemble_evaluator/evaluator.py
SAKavli: I agree.
I now renamed to `self.highest_violating_parallelization` and set it to `None` initially instead of 0 - does it make sense then?
Finding a good name for this is challenging. I am open for suggestions!
https://github.com/equinor/ert/pull/12381/commits/6a69c4698a9050d587391de513e41efd88a3aaa2

src/ert/ensemble_evaluator/evaluator.py
SAKavli: Added a punctuation 👍 What do you think of the linebreaks?

eivindjahren:
> This warning might drown in other forward model warnings. I was curious if it makes sense to implement some prioritization in the Suggestor showing these warnings?

That is a good point. I think that maybe users do not get very many warnings from run_dialog today? I think it would make sense to at least look into grouping warnings so that we don't repeat very similar ones.

Probably a good idea to gather some statistics here before we do anything.

src/ert/run_models/run_model.py
```diff
@@ -642,6 +643,16 @@ async def run_ensemble_evaluator_async(
         finally:
             await evaluator_task
 
+        if (
+            evaluator.highest_violating_parallelization is not None
```
eivindjahren: you could have
```python
self._highest_violating_evaluator_parallelization, self._cpu_violation_warning_msg = max(
  (self._highest_violating_evaluator_parallelization, self._cpu_violation_warning_msg),
  (evaluator.highest_violating_parallelization or 0, evaluator.cpu_violation_warnings_msg),
)

Comment (eivindjahren) [reply src/ert/run_models/run_model.py]: Kind of hard to read with the long names regardless. Might be easier if you use `max` instead of `highest` and `parallelism` instead of `parallelization`. And it is not really the evaluator which is in breach of the policy but the task it ran, so we might say `task` instead of `evaluator`.

It also seems a bit off to call it `cpu_violation_warnings_msg`, maybe `parallelism_violation_msg` ?

```python
self._max_violating_task_parallelism, self._parallelism_violation_msg = max(
  (self._max_violating_task_parallelism, self._parallelism_violation_msg),
  (evaluator.max_violating_parallelism or 0, evaluator.parallelism_violation_msg),
)
```

eivindjahren: If you use a 
```python
@dataclass 
class ParallelismViolation:
   amount: float
   message: str | None = None
```
with `amount <= 0` meaning no violation, then you also get:
```python
self._max_parallelism_violation = max(
  self._max_parallelism_violation,
  evaluator.max_parallelism_violation
)
```

SAKavli: I will look into this alternative. Note that I had 0 as initial value before, but I got feedback that the number is 0 when there has been some parallelization could be confusing.

src/ert/run_models/run_model.py
```diff
@@ -85,6 +86,15 @@ class ErtRunError(Exception):
     pass
 
 
+@dataclass
```
eivindjahren: `@dataclass(order=True)`

src/ert/config/__init__.py
```diff
@@ -117,6 +118,7 @@
     "ObservationSettings",
     "ObservationType",
     "OutlierSettings",
+    "ParallelismViolation",
```
eivindjahren: This should not be in `ert.config`. Probably the best place for it is `src/ert/ensemble_evaluator/evaluator.py` as any callsite fetching `EnsembleEvaluator.max_parallelism_violation` will be imported after `src/ert/ensemble_evaluator/evaluator.py` anyways.

eivindjahren (changes requested):
`ert.config` is not the right place for `ParallelismViolation`, otherwise this looks good. `ParallelismViolation` should be close to where it is used. I think the best place is `src/ert/ensemble_evaluator/evaluator.py`.

eivindjahren:
> There was a circular import issue between `run_model.py` and `evaluator.py` because both use this class. Therefore I thought it would make sense to lift it a level up. I will see what I can do!

If you want to lift it one level up then the right place for it is `src/ert/_parallel_violation.py` as putting it into `src/ert/config/` can create more problems down the line and adds to the interface of `ert.config` and its best not do that unnecessarily.

PR #12386, opened 2025-11-25
Fix merge genkw params with design matrix

xjules:
We should avoid mutating param config

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/design_matrix.py
+++ b/src/ert/config/design_matrix.py
@@ -171,26 +171,33 @@ class DesignMatrix:
 
         for param_cfg in existing_parameters:
             if isinstance(param_cfg, GenKwConfig) and param_cfg.name in design_cfgs:
-                param_cfg.input_source = DataSource(
+                del design_cfgs[param_cfg.name]
+                input_source = DataSource(
                     self.parameter_priority.get(
                         param_cfg.name, DataSource.DESIGN_MATRIX.value
                     )
                 )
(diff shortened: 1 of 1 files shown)


src/ert/config/design_matrix.py
```diff
@@ -171,26 +171,33 @@ def merge_with_existing_parameters(
 
         for param_cfg in existing_parameters:
             if isinstance(param_cfg, GenKwConfig) and param_cfg.name in design_cfgs:
-                param_cfg.input_source = DataSource(
+                del design_cfgs[param_cfg.name]
+                input_source = DataSource(
```
jonathan-eq, line 175: Is it enough to keep everything the way it was, and just do 
```
param_cfg = BaseModel.model_copy(param_cfg)
```
?

jonathan-eq: Then the else block would also add a copy of the parameter_config instance, and not actual instance. This would safe-guard us in case something downstream tries to modify it.

xjules: I see. I wanted to have a better control over it, ie. thus creating GenKw on the spot. 
Not sure what downstream can modify it though.

xjules: > Then the else block would also add a copy of the parameter_config instance, and not actual instance.

It won't work since there could be another configs than genkw. Will keep as is

Issue #11511, opened 2025-08-19
Ert crashes during update when trying to run enif on drogon_ahm.ert


larsevj:
### What happened? (You can include a screenshot if it helps explain)

Crashes with following stacktrace:
```
could not convert string to float: 'APS_input'
Traceback (most recent call last):
  File "/private/levje/ert/src/ert/run_models/run_model.py", line 361, in start_simulations_thread
    self.run_experiment(
  File "/usr/lib64/python3.11/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/private/levje/ert/src/ert/run_models/ensemble_smoother.py", line 43, in run_experiment
    posterior = self.update(prior, self.target_ensemble % 1)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/private/levje/ert/src/ert/run_models/update_run_model.py", line 117, in update
    self.update_ensemble_parameters(prior, posterior, weight)
  File "/private/levje/ert/src/ert/run_models/ensemble_information_filter.py", line 18, in update_ensemble_parameters
    enif_update(
  File "/private/levje/ert/src/ert/analysis/_enif_update.py", line 85, in enif_update
    raise e
  File "/private/levje/ert/src/ert/analysis/_enif_update.py", line 63, in enif_update
    analysis_EnIF(
  File "/private/levje/ert/src/ert/analysis/_enif_update.py", line 156, in analysis_EnIF
    X_full_scaled = X_full_scaler.fit_transform(X_full.T)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/prog/res/komodo/bleeding-20250819-0306-py311-rhel8-numpy2/root/lib64/python3.11/site-packages/sklearn/utils/_set_output.py", line 316, in wrapped
    data_to_wrap = f(self, X, *args, **kwargs)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/prog/res/komodo/bleeding-20250819-0306-py311-rhel8-numpy2/root/lib64/python3.11/site-packages/sklearn/base.py", line 894, in fit_transform
    return self.fit(X, **fit_params).transform(X)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #12388, opened 2025-11-25
Disregard unupdated parameters with EnIF

yngve-sk:
**Issue**
Resolves https://github.com/equinor/ert/issues/11511

Approach: After some discussions, it has been decided that we will first omit all `update==False` parameters from EnIF update. This implies that EnIF will completely disregard any effect unupdated parameters (including those from design matrix) have on responses, and update accordingly. 

Theoretical feasibility of [skipping only categoricals + having EnIF consider but not update design parameters](https://github.com/equinor/ert/compare/main...yngve-sk:ert:25.11.try-fixup-enif-drogonahm-skip-only-categoricals) should be investigated more closely.

### DIFF
--- a/src/ert/analysis/_enif_update.py
+++ b/src/ert/analysis/_enif_update.py
@@ -160,10 +160,14 @@ def analysis_EnIF(
         Y=S.T,
         verbose_level=5,
     )
-
+    updated_parameters = [
+        p
+        for p, config in source_ensemble.experiment.parameter_configuration.items()
+        if config.update
+    ]
     # Learn the precision matrix block-sparse over parameter groups
     Prec_u = sp.sparse.csc_matrix((0, 0), dtype=float)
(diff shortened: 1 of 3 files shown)


berland:
(Marked as draft pr)

yngve-sk:
Drogon_ahm does run without the crash, but takes forever. Now running 2days 20hrs. That is separate from this issue though, as EnIF is not suitable for large fields anyway. Will do a small dummy test with design matrix to see that it can successfully update. 

Screenshot 2025-12-01 at 10 00 42

yngve-sk:
Added test that runs ensemble experiment, then does a [manual update + evaluate ensemble] with, and one without a non-updateable parameter added, expecting responses to be the same.

yngve-sk:
Update: AHM on drogon successfully completed in 3days 11hrs

PR #12390, opened 2025-11-25
Avoid copying in substitutions and use faster substitution method

### LINKED ISSUE #12387: Optimize template substitutions for parameters
It seems that the following (down-the-line) substitution code called from `_create_runpath` function is quite inefficient:

```python
staticmethod
    def substitute_parameters(
        to_substitute: str, parameter_values: Mapping[str, Mapping[str, str | float]]
    ) -> str:
        """Applies the substitution '<param_name>' to parameter value
        Args:
            parameter_values: Mapping from parameter name to parameter value
            to_substitute: string to substitute magic strings in
        Returns:
            substituted string
        """
        for values in parameter_values.values():
            for param_name, value in values.items():
                if isinstance(value, (int, float)):
                    formatted_value = f"{value:.6g}"
                else:
                    formatted_value = str(value)
                to_substitute = to_substitute.replace(
                    f"<{param_name}>", formatted_value
                )
        return to_substitute
```
called from here
```python
result = substituter.substitute_parameters(
                    result,
                    param_data,
                )
```

This might relate to https://github.com/equinor/ert/issues/12271

eivindjahren:
Resolves #12387

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/run_models/_create_run_path.py
+++ b/src/ert/run_models/_create_run_path.py
@@ -5,6 +5,7 @@ import logging
 import os
 import time
 from collections.abc import Iterable, Mapping
+from copy import deepcopy
 from datetime import UTC, datetime
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
@@ -174,6 +175,21 @@ def _manifest_to_json(ensemble: Ensemble, iens: int, iter_: int) -> dict[str, An
     return manifest
 
 
+def _make_param_substituter(
+    substituter: Substitutions,
+    param_data: Mapping[str, Mapping[str, str | float]],
+) -> Substitutions:
+    param_substituter = deepcopy(substituter)
+    for values in param_data.values():
+        for param_name, value in values.items():
+            if isinstance(value, (int, float)):
+                formatted_value = f"{value:.6g}"
+            else:
+                formatted_value = str(value)
+            param_substituter[f"<{param_name}>"] = formatted_value
+    return param_substituter
+
+
 @log_duration(logger, logging.INFO)
 def create_run_path(
     run_args: list[RunArg],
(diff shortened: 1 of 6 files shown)


jonathan-eq:
@xjules Have we been able to reproduce the slow runpath creation issue, and verified that it is the parameter substitution that is slow?

src/ert/substitutions.py
```diff
@@ -3,13 +3,33 @@
...
+    @cached_property
+    def _matcher(self) -> Callable[[str], tuple[str, int]]:
+        pattern = "|".join(re.escape(key) for key in self)
+        if not pattern:
+            return lambda text: (text, 0)
+        regex = re.compile(f"({pattern})")
+        return lambda text: regex.subn(lambda mo: self[mo.group()], text)
+
+    def __setitem__(self, key: str, value: str) -> None:
+        with suppress(Exception):
```
xjules: we can do instead:
`if self._matcher is not None:`

src/ert/substitutions.py
```diff
@@ -3,13 +3,33 @@
...
+            del self._matcher
+        super().__setitem__(key, value)
+
+    def __delitem__(self, key: str) -> None:
+        with suppress(Exception):
```
xjules: same as above

src/ert/substitutions.py
eivindjahren: That doesn't work as you will get AttributeError and not a problem with self._matcher being None.

src/ert/substitutions.py
eivindjahren: That doesn't work as you will get `AttributeError` and not a problem with self._matcher being None.

src/ert/substitutions.py
xjules: That is interesting since we use the same pattern in `local_ensemble.py`
```python
class LocalEnsemble(BaseMode):
@cached_property
def _existing_scalars(self) -> dict[str, list[int]]:
...
```

eivindjahren: That seems incorrect according to what documentation I can find, but I haven't been able to make this del raise an AttributeError either so I am going to test a bit. The `is None` test seems very suspect as self._matcher is clearly  callable, something that None is not.

eivindjahren: In any case I had to remove the cached_property because it is a bit slower during the creation of ertconfig where substitution and setitem happens interspersed.

src/ert/run_models/_create_run_path.py
```diff
@@ -174,6 +175,21 @@ def _manifest_to_json(ensemble: Ensemble, iens: int, iter_: int) -> dict[str, An
     return manifest
 
 
+def _make_param_substituter(
+    substituter: Substitutions,
+    param_data: Mapping[str, Mapping[str, str | float]],
```
xjules, line 180: in there param_data still contains groups, but those are ignored - which is a desired behavior. Could you remove the groups? ie. `param_data: Mapping[str, str | float]`

eivindjahren: That is probably not a good idea as we would construct a lot of dictionaries then. We could make it `param_data: Iterable[Mapping[str, str | float]]` but that just makes for an unnecessary complicated interface for a small private function`.

Issue #12405, opened 2025-11-27
Manual update from ensemble experiment gives error


yngve-sk:
**Also in current stable: 25.11, works in 25.10**
### What happened? (You can include a screenshot if it helps explain)

1. Run snake oil ensemble experiment
2. Run manual update on this and see the following error: 

```
Users/yngves.kristiansen/Documents/Repositories/oppdrag/equinor/ert-tree2/src/ert/run_models/run_model.py", line 429, in start_simulations_thread
    self.run_experiment(
    ~~~~~~~~~~~~~~~~~~~^
        evaluator_server_config=evaluator_server_config,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        rerun_failed_realizations=rerun_failed_realizations,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "ert-tree2/src/ert/run_models/manual_update.py", line 48, in run_experiment
    target_experiment = self._storage.create_experiment(
        parameters=list(prior_experiment.parameter_configuration.values()),
    ...<4 lines>...
        templates=self._prior.experiment.templates_configuration,
    )
  File "ert-tree2/src/ert/storage/mode.py", line 98, in inner
    return func(self_, *args, **kwargs)
  File "ert-tree2/src/ert/storage/local_storage.py", line 346, in create_experiment
    exp = LocalExperiment.create(
        self,
    ...<7 lines>...
        templates=templates,
    )
  File "ert-tree2/src/ert/storage/local_experiment.py", line 187, in create
    shutil.copyfile(incoming_template, template_file_path)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13
```

### Steps to reproduce

```console

```

### Environment where bug has been observed

- [x] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [x] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

PR #12411, opened 2025-11-28
Fix faulty templates passthrough for manual update

yngve-sk:
**Issue**
Resolves https://github.com/equinor/ert/issues/12405

We should likely make the templates "lock in" and respect what is in storage, and fix that, this is to be considered temporary, same as for esmda restart

### DIFF
--- a/tests/ert/unit_tests/run_models/test_manual_update.py
+++ b/tests/ert/unit_tests/run_models/test_manual_update.py
@@ -1,0 +1,73 @@
+import queue
+from argparse import Namespace
+from pathlib import Path
+
+import pytest
+
+from ert.config import ErtConfig
+from ert.ensemble_evaluator import EvaluatorServerConfig
+from ert.mode_definitions import (
+    ENSEMBLE_EXPERIMENT_MODE,
+    MANUAL_ENIF_UPDATE_MODE,
+    MANUAL_UPDATE_MODE,
+)
+from ert.run_models import create_model
+from ert.storage import open_storage
+
+
+@pytest.mark.integration_test
+@pytest.mark.parametrize("mode", [MANUAL_UPDATE_MODE, MANUAL_ENIF_UPDATE_MODE])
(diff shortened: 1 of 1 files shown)


src/ert/run_models/manual_update.py
```diff
@@ -19,6 +19,7 @@
 
 class ManualUpdateConfig(UpdateRunModelConfig):
     ensemble_id: str
+    ert_templates: list[tuple[str, str]]
```
berland, line 22: Is this needed? (asking cause it was not needed in 39a32cb0829186b61c5cecc4fc32b0bd74fde524 ).

It is also defined in InitialEnsembleRunModelConfig

yngve-sk: ManualUpdate is not an `InitialEnsembleRunModelConfig`, adding it one lvl up would break the abstraction, as `EvaluateEnsemble` would then also inherit the unused prop

tests/ert/unit_tests/run_models/test_manual_update.py
```diff
@@ -0,0 +1,73 @@
...
+from ert.mode_definitions import (
+    ENSEMBLE_EXPERIMENT_MODE,
+    MANUAL_ENIF_UPDATE_MODE,
+    MANUAL_UPDATE_MODE,
+)
+from ert.run_models import create_model
+from ert.storage import open_storage
+
+
+@pytest.mark.parametrize("mode", [MANUAL_UPDATE_MODE, MANUAL_ENIF_UPDATE_MODE])
```
berland, line 19: I would guess this test takes some seconds. Should there be a marker to say it is an integration test?

yngve-sk: Marked, though they are quite fast for me locally, ~2s per case.

berland (approved):
👍🏻

PR #12437, opened 2025-12-02
Remove the detached_node_output directory

### LINKED ISSUE #12328: Move detached_node_output/endpoint.log to a more sensible place
When running everest,  logs for the communication with the "everserver" ends up in an awkwardly named path. 

I would expect the file to end up in a directory called 'logs', similar to Ert. Also, endpoint.log does not sufficiently describe the contents, this should be improved.

verveerpj:
**Issue**
Resolves #12328

**Approach**
- Move the .session directory to the everest output directory.
- Move the log to the everest logs directory.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/config/server_config.py
+++ b/src/everest/config/server_config.py
@@ -14,12 +14,7 @@ from ert.config.queue_config import (
 )
 from ert.dark_storage.client import ErtClientConnectionInfo
 
-from ..strings import (
-    CERTIFICATE_DIR,
-    DETACHED_NODE_DIR,
-    HOSTFILE_NAME,
-    SESSION_DIR,
-)
+from ..strings import CERTIFICATE_DIR, HOSTFILE_NAME, SESSION_DIR
 from .simulator_config import check_removed_config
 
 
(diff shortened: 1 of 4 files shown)


src/everest/config/server_config.py
```diff
@@ -127,7 +118,7 @@ def get_hostfile_path(output_dir: str) -> str:
     def get_session_dir(output_dir: str) -> str:
         """Return path to the session directory containing information about the
         certificates and host information"""
-        return os.path.join(ServerConfig.get_detached_node_dir(output_dir), SESSION_DIR)
+        return os.path.join(os.path.abspath(output_dir), SESSION_DIR)
```
berland, line 121: would look better with pathlib, but that should maybe be more consistent update across these functions?

verveerpj: I agree, pathlib would be better. However, we are using strings for file paths and names throughout the Everest configuration code and such, and therefore this returns a string, not a Path. It's a bit pointless to use pathlib and then convert back to a string anyway. I think it would be better to have a good look at `os.path` usage and why we use strings instead of path in the Everest configuration, maybe in a different issue.

src/everest/config/server_config.py
```diff
@@ -14,12 +14,7 @@
 )
 from ert.dark_storage.client import ConnInfo
 
-from ..strings import (
-    CERTIFICATE_DIR,
-    DETACHED_NODE_DIR,
```
berland, line 19: We can remove the definition of `DETACHED_NODE_DIR` in `strings.py` as well.

PR #12452, opened 2025-12-04
Add distance_based_localization functions to field_utils

oddvarlia:
**Issue**
Links to #12252

**Approach**
Basic functions to be used when calculating `RHO` matrix for updating of both 3D and 2D field parameters using distance-based localization method implemented in `class DistanceESMDA`.

PR #12455, opened 2025-12-04
Added distance-based localization functions to class Field

oddvarlia:
Added functions  in class Field to be used when calculating the RHO matrix for distance-based localization.
This is related to #12452 since it requires the update of the field_utils.py


src/ert/config/field.py
```diff
@@ -325,6 +328,113 @@ def ny(self) -> int:
...
+            self.ertbox_params.yinc,
+            obs_xpos,
+            obs_ypos,
+            obs_main_range,
+            obs_perp_range,
+            obs_anisotropy_angle,
+            right_handed_grid_indexing=right_handed_grid_indexing,
+        )
+
+    def transform_positions_to_local_field_coordinates(
```
xjules: Do we need this function?

src/ert/config/field.py
```diff
@@ -325,6 +328,113 @@ def ny(self) -> int:
...
+        )
+        assert self.ertbox_params.rotation_angle is not None, (
+            "Parameter for grid rotation must be defined"
+        )
+
+        return transform_positions_to_local_field_coordinates(
+            self.ertbox_params.origin, self.ertbox_params.rotation_angle, utmx, utmy
+        )
+
+    def transform_local_ellipse_angle_to_local_coords(
```
xjules: Do we need this function?

src/ert/config/field.py
oddvarlia: OK. I think we can include the transformation to be locally inside the function calculating RHO for one layer instead.
I will include it as a step in RHO calculation instead.

src/ert/config/field.py
oddvarlia: OK, we can do the coordinate transformation inside the function calculating RHO for one grid layer instead.

src/ert/config/field.py
xjules: We should try "keep-it-simple" principle in the beginning at least. I agree to have in the RHO calculation.

oddvarlia: Fixed

xjules:
Very nice @oddvarlia ! I think the commit message can be updated to account for the removal of those two functions.

oddvarlia:
Updated commit message so it is consistent with the changes. Removed reference to functions not anymore included in class Field

tests/ert/unit_tests/config/test_field.py
```diff
@@ -374,3 +374,129 @@ def test_calculate_ertbox_parameters_synthetic_grid(origin, increment, rotation,
     assert params.nx == 5
     assert params.ny == 4
     assert params.nz == 3
+
+
+@pytest.fixture
+def field():
```
xjules: can you maybe be more specific here?
suggestion:
`field_with_location_attributes`
`field_with_ertbox_params`

PR #12456, opened 2025-12-04
Distance localization update of surface field

oddvarlia:
Requires that the field_utils.py is updated. This PR adds functionality to calculate RHO matrix for distance-based localization for 2D field parameters (Surface parameters). It adds functions to the class SurfaceConfig  function to  calculate elements of the RHO matrix for distance-based localization.

This is related to https://github.com/equinor/ert/pull/12452 since it requires the update of the field_utils.py


tests/ert/unit_tests/config/test_surface_config.py
```diff
@@ -339,3 +339,105 @@ def test_surface_create_storage_datasets_raises_surface_mismatch_error_when_the_
...
+        (
+            [1050.0, 1250.0, 1250.0, 1250.0, 1000.0, 1500.0, 1000.0],  # xpos
+            [2050.0, 2300.0, 2300.0, 2300.0, 2000.0, 2000.0, 2600.0],  # ypos
+            [100.0, 300.0, 300.0, 600.0, 1000.0, 300.0, 300.0],  # main_range
+            [100.0, 100.0, 100.0, 200.0, 100.0, 10.0, 100.0],  # perp_range
+            [0.0, 35.0, 135.0, -135.0, 45.0, -45.0, -60.0],  # angle
+        ),
+    ],
+)
+def test_calc_rho_for_2d_grid_layer(
```
xjules, line 389: Btw, how long does this test take on average?

oddvarlia: This test is very quick. In Drogon case for surfaces, maybe 30seconds to 60 seconds but I don't know exact. Whole Drogon run updating 9 3D fields and 4 2D fields and all scalar parameters
 take about 5-6 minutes ( after the speed-up of scalar parameter updates was done in my script)

PR #12461, opened 2025-12-04
Have EnsembleEvaluator.listen_for_messages return early on socket issues

### LINKED ISSUE #12433: Investigate log message "Evaluator receiver closed, no new messages are received"
The last 24 hours have seen half a million occurences of "Evaluator receiver closed, no new messages are received"

jonathan-eq:
**Issue**
Resolves #12433

**Approach**
This commit makes the method return early on zmq.ENOTSOCK exception as it in one case logged
400_000 times after an experiment was cancelled. This might have
been due to a race condition of shutting down while some realizations
were starting, and the socket was being closed while jobs were
connecting. This is the easiest fix, as we do not attempt to
re-establish the socket later anyways.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/evaluator.py
+++ b/src/ert/ensemble_evaluator/evaluator.py
@@ -369,6 +369,7 @@ class EnsembleEvaluator:
                     logger.warning(
                         "Evaluator receiver closed, no new messages are received"
                     )
+                    return  # The socket is closed, and we won't re-establish it.
                 else:
                     logger.error(f"Unexpected error when listening to messages: {e}")
             except asyncio.CancelledError:
--- a/tests/ert/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
+++ b/tests/ert/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
@@ -722,3 +722,39 @@ async def test_log_forward_model_steps_with_missing_status_updates(
         f"There could be connectivity issues to evaluator running on port "
         f"{router_port} on host {evaluator_host}"
     ) in caplog.text
+
+
+@pytest.mark.timeout(5)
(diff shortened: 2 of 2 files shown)


src/ert/ensemble_evaluator/evaluator.py
```diff
@@ -369,6 +369,7 @@ async def listen_for_messages(self) -> None:
                     logger.warning(
                         "Evaluator receiver closed, no new messages are received"
                     )
+                    return  # The socket is closed, and we won't re-establish it.
                 else:
```
xjules, line 373: What about the other cases? Those having "Unexpected error when listening to messages:". Are there any entries in the logs?

jonathan-eq: No, we have never had that log entry.

tests/ert/unit_tests/ensemble_evaluator/test_ensemble_evaluator.py
```diff
@@ -739,3 +722,39 @@ async def test_log_forward_model_steps_with_missing_status_updates(
         f"There could be connectivity issues to evaluator running on port "
         f"{router_port} on host {evaluator_host}"
     ) in caplog.text
+
+
+@pytest.mark.timeout(5)
+async def test_that_evaluator_listen_for_messages_exits_if_socket_has_been_closed(
```
xjules, line 728: I'm wondering if we need this test at all when we return after logger.warning anyway

jonathan-eq: It is to make sure we return, otherwise it would hang. I think if we have this in a task with a closed socket, and log on every closed socket exception, it would never yield to the event loop. As it never yields, the task responsible for cancelling this task will never proceed.

Issue #12406, opened 2025-11-27
Distance-based localization requires batching of field parameters


oddvarlia:
To handle the large memory requirements from distance-based localizaiton, the field parameters can be updated in batches which covers a subset of the full set of field parameters. The batches must be defined in a convenient way for distance based localization. A convenient ways is to define a batch as consisting of a group of 3D grid layers since this simplifies the calculation of the RHO matrix.

An algorithm for batching can be something like this:
- Process one field parameter at a time and split that into batches:
  - Calculate available memory and use up to 50% (maybe less?) 
  - Calculate how many grid layers can be updated as one batch. The size of  a layer of a 3D field parameter is ` nx * ny`  and how many layers (up to max equal `nz`) can be handled with available memory, is calculated.
  - Number of batches will then be  `nbatches = int(nz/nmax_layer_per_batch)`  and the remaining layers `nz - (nbatches * nlayer_per_batch)`  is the size of the last batch for one 3D field parameter.
  - For each batch compute:
    - RHO matrix for the batch of layers, assign the M matrix with field parameters to be updated, calculate the updated posterior M matrix. Since the localization only cares about lateral distances, a 2D layer with scaling values can be prepared based on the observation positions and one grid layer of the 3D grid before looping into the update of a 3D parameter. The scaling values are filled into a 3D parameter corresponding to the batch using the 2D parameter of scaling values for each observation. Finally the RHO matrix corresponding to the batch is filled from the 3D parameters of scaling values for each observation. 
    - Assign the updated version of the values of current 3D field parameter to the result 3D parameter.

PR #12469, opened 2025-12-06
Update 3D field parameter with ensemble smoother with distance-based localization

oddvarlia:
Function to update 3D field parameters with the method implemented in DistanceESMDA for ensemble smoother with
distance-based localization

Linked to: #12406

**Approach**
A 3D field parameter can often not be update in one go due to large memory requirements due to the Kalman gain matrix K and the localization matrix RHO. To handle this the field parameter can be split into subsets or batches. Each batch should be a set of layers of field parameter values defined in the field parameter grid (ertbox grid). Each batch must consists of a complete set of layers, not subset of layers of field parameters. The main reason for processing the update by batches of layers of field parameters is that the calculation of scaling factors (elements in the RHO matrix) can be done layer by layer due to the assumption that only lateral distance is of importance in the definition of influence of an observation, hence the calculation of RHO is done for one layer and then copied to all the other layers of the ertbox grid.


tests/ert/unit_tests/analysis/test_es_update.py
```diff
@@ -1162,3 +1177,1546 @@ def test_update_subset_parameters(storage, uniform_parameter, obs):
...
+    X_post_stdev_3D = X_post_3D.std(axis=3)
+    # Difference between post and prior mean and stdev
+    X_diff_mean_3D = X_post_mean_3D - X_prior_mean_3D
+    X_diff_stdev_3D = X_post_stdev_3D - X_prior_stdev_3D
+
+    if write_fields:
+        # The output here is for visual inspection of the result
+        # and is not output in automatic test runs, but can be
+        # used in manual inspection of the results from the update.
+        write_mean_stdev_to_files(
```
xjules: let's remove this output functions.

oddvarlia: OK

src/ert/analysis/_es_update.py
```diff
@@ -177,6 +173,101 @@ def _calculate_adaptive_batch_size(num_params: int, num_obs: int) -> int:
     )
 
 
+def calc_max_number_of_layers_per_batch_for_distance_localization(
```
xjules, line 176: Maybe as an inspiration you can have a look at `def _calculate_adaptive_batch_size` ?

src/ert/analysis/_es_update.py
```diff
@@ -502,3 +593,205 @@ def log_warning(
...
+    nparam = nparam_per_layer * nz
+    nreal = X_prior.shape[1]
+    assert X_prior.shape[0] == nparam, (
+        f"Mismatch between X_prior dimension {X_prior.shape[0]} and nparam {nparam}"
+    )
+
+    X_prior_3D = X_prior.reshape(nx, ny, nz, nreal)
+    # No update if no observations or responses
+    if Y is None or Y.shape[0] == 0:
+        # No update of the field parameters
```
xjules, line 681: I think we can just return X_prior, yes.

xjules: you can also remove the comments, but keep the one above.  # No update if no observations or responses

src/ert/analysis/_es_update.py
```diff
@@ -502,3 +593,205 @@ def log_warning(
...
+
+    # Check memory constraints and calculate how many grid layers of
+    # field parameters is possible to update on one batch
+    max_nlayers_per_batch = (
+        calc_max_number_of_layers_per_batch_for_distance_localization(
+            nx, ny, nz, nobs, nreal, bytes_per_float=8
+        )
+    )  # Use float64
+    nlayer_per_batch = min(max_nlayers_per_batch, nz)
+    nbatch = int(nz / nlayer_per_batch)
```
xjules, line 704: if nz==10 and nlayer_per_batch==7 then nbach = 1, is this correct?

src/ert/analysis/_es_update.py
```diff
@@ -502,3 +593,205 @@ def log_warning(
...
+
+    log_msg = "Minimum number of batches required due to "
```
xjules, line 706: we can log nbatch and nlayer_per_batch and make the message concise

src/ert/analysis/_es_update.py
```diff
@@ -502,3 +593,205 @@ def log_warning(
...
+    # Initialize the X_post_3D
+    X_post_3D = X_prior_3D.copy()
+    for batch_number in range(nbatch):
+        start_layer_number = batch_number * nlayer_per_batch
+        end_layer_number = start_layer_number + nlayer_per_batch
+        log_msg = (
+            f"Batch number: {batch_number}\n"
+            f"start layer : {start_layer_number}\n"
+            f"end layer   : {end_layer_number - 1}"
+        )
```
xjules, line 743: remove the logs.

src/ert/analysis/_es_update.py
```diff
@@ -502,3 +593,205 @@ def log_warning(
...
+                print(f"Run time used: {(end_time - start_time):.4f} seconds")
+            return result
+
+        return wrapper
+
+    return decorator
+
+
+@memory_usage_decorator(enabled=False)
+def update_3D_field_with_distance_esmda(
```
xjules, line 629: Let's double-check if iterative-ensemble-smoother would be a better place to host this code.

Issue #12462, opened 2025-12-05
AttributeError: 'str' object has no attribute 'formatted_name' error in integration testing


berland:
This exception occurs in Komodo integration tests while running GUI tests:

```
Traceback (most recent call last):
  File "/tmp/f_scout_ci/actions-runner-07/_temp/test-kenv/root/lib64/python3.12/site-packages/ert/gui/simulation/experiment_panel.py", line 368, in run_experiment
    f"Queue system:\n{model.queue_config.queue_system.formatted_name}"
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'formatted_name'
```

Obviously related to https://github.com/equinor/ert/pull/12440/commits/d032a2120c97887ef56fe1e142a89c39aba2d8bc

berland:
Bug is also present in interactive sessions, so it is not caused by test-mocking.

berland:
This bug seems related to this setting:
https://github.com/equinor/ert/blob/940b8beeb718c42b3094b65148587b2d9506b66f/src/ert/config/queue_config.py#L54

berland:
Residual issue is to reapply the reverted commit after merging of #12464

berland:
> Residual issue is to reapply the reverted commit after merging of [#12464](https://github.com/equinor/ert/pull/12464)

Reapplied in #12559

PR #12474, opened 2025-12-08
Refactor Everest yaml dumper

berland:
**Issue**
Preparations for #12462 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/everest/bin/config_branch_script.py
+++ b/src/everest/bin/config_branch_script.py
@@ -5,8 +5,6 @@ from pathlib import Path
 from textwrap import dedent
 from typing import Any
 
-from ruamel.yaml import YAML
-
 from everest.bin.utils import setup_logging
 from everest.config import EverestConfig
 from everest.config_file_loader import load_yaml
@@ -139,10 +137,9 @@ def config_branch_entry(args: list[str] | None = None) -> None:
             conf_controls=yml_config["controls"], opt_controls=opt_controls
         )
 
-        yaml = YAML()
-        yaml.indent(mapping=2, sequence=4, offset=2)
-        yaml.preserve_quotes = True
-        yaml.dump(yml_config, Path(options.output_config))
+        EverestConfig.write_dict_to_file(
+            yml_config, Path(options.output_config), safe_and_pure=False
+        )
         print(f"New config file {options.output_config} created.")
 
 
--- a/src/everest/bin/everconfigdump_script.py
+++ b/src/everest/bin/everconfigdump_script.py
@@ -3,8 +3,6 @@
 import argparse
 import sys
 
-from ruamel.yaml import YAML
-
 from everest.config import EverestConfig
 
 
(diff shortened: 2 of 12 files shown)


src/everest/config/everest_config.py
```diff
@@ -1081,22 +1081,16 @@ def load_file_with_argparser(
         except ValueError as e:
             parser.error(f"Loading config file <{config_path}> failed with: {e}")
 
-    def dump(self, fname: str | None = None) -> str | None:
-        """Write a config dict to file or return it if fname is None."""
+    def dump(self, fname: str) -> None:
```
verveerpj: Why introducing the static method, can you not just rename this function to `yaml_dump` or `dump_yaml`?

The name `dump` is not very good anyway, it does not convey what it does properly.

src/everest/config/everest_config.py
```diff
@@ -1081,22 +1081,16 @@ def load_file_with_argparser(
...
         stripped_conf = self.to_dict()
-
         del stripped_conf["config_path"]
+        self.yaml_dump(stripped_conf, Path(fname))
 
+    @staticmethod
+    def yaml_dump(config: dict[str, Any], output_file: Path | TextIO) -> None:
```
verveerpj: This is basically a helper function that dumps a dictionary to YAML. Does that need to be a public static method of the configuration class? Is it going to be used for something else except dumping the config itself?

src/everest/bin/everconfigdump_script.py
```diff
@@ -29,10 +27,7 @@ def config_dump_entry(args: list[str] | None = None) -> None:
 
     config = EverestConfig.load_file(options.config_file)
 
-    yaml = YAML(typ="safe", pure=True)
-    yaml.indent = 2
-    yaml.default_flow_style = False
-    yaml.dump(config.to_dict(), sys.stdout)
+    EverestConfig.yaml_dump(config.to_dict(), sys.stdout)
```
verveerpj: Why not just `config.dump(sys.stdout)`?  (With necessary adaptions to `dump`).

src/everest/config/everest_config.py
berland: Renamed `dump` to `write_to_file`. 

Added config_branch_entry to this PR, which has the need for a dict-exporter.

src/everest/bin/everconfigdump_script.py
berland: fixed!

src/everest/bin/config_branch_script.py
```diff
@@ -139,10 +137,9 @@ def config_branch_entry(args: list[str] | None = None) -> None:
             conf_controls=yml_config["controls"], opt_controls=opt_controls
         )
 
-        yaml = YAML()
-        yaml.indent(mapping=2, sequence=4, offset=2)
-        yaml.preserve_quotes = True
-        yaml.dump(yml_config, Path(options.output_config))
+        EverestConfig.write_dict_to_file(
+            yml_config, Path(options.output_config), safe_and_pure=False
```
verveerpj, line 141: Mostly curiosity: why can't we use `pure` and `safe` in this case?

berland: the `yml_config` we get from `options.config` is of type `ruamel.yaml.CommentedMap` which ruamel cannot dump when safe and pure..

PR #12482, opened 2025-12-09
Add migration for RealizationStorageState

yngve-sk:


tests/ert/unit_tests/storage/migration/test_to17.py
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    result = migrate_realization_errors_json(input_json)
+    assert migrate_realization_errors_json(input_json) == expected_output, (
+        f"Expected {expected_output}, got {result}"
+    )
+
+
+@pytest.mark.usefixtures("use_tmpdir")
+def test_migrate():
```
xjules: do we even need this test?

PR #12500, opened 2025-12-15
Make check_migration_needed staticmethod

yngve-sk:
Opening a read storage without migrating has some implications, better to keep it to just checking the version

**Issue**
Resolves ERT gui being unable to migrate old storages

### DIFF
--- a/src/ert/storage/local_storage.py
+++ b/src/ert/storage/local_storage.py
@@ -110,7 +110,6 @@ class LocalStorage(BaseMode):
                 )
             else:
                 self._migrate(self.version)
-                self._save_index()
 
         self.refresh()
         if mode.can_write:


src/ert/gui/main.py
```diff
@@ -158,8 +163,9 @@ def _start_initial_gui_window(
     storage_path = None
     if ert_config is not None:
         try:
-            with open_storage(ert_config.ens_path) as read_storage:
-                should_migrate = read_storage.check_migration_needed()
+            should_migrate = LocalStorage.check_migration_needed(
+                Path(ert_config.ens_path)
+            )
```
andreas-el: Combine with if?

src/ert/storage/local_storage.py
```diff
@@ -101,34 +101,39 @@ def __init__(
...
-        if self.check_migration_needed():
-            self.perform_migration()
+        if self.check_migration_needed(Path(self.path)):
+            if not self.can_write:
+                raise RuntimeError(
+                    f"Cannot open storage '{self.path}' in read-only mode: "
+                    f"Storage version {self.version} is too old. "
+                    f"Run ert to initiate migration."
+                )
+            else:
```
andreas-el, line 111: If this is the only place where we want to `perform_migration`, we can inline this code.

Also the public function `perform_migration` does not check for write access anymore, this is unfortunate.

PR #12525, opened 2025-12-19
Add Latin Hypercube sampling

lucaeg:
This is a draft, do not merge.


src/ert/config/parameter_config.py
```diff
@@ -142,9 +143,7 @@ def transform_data(self) -> Callable[[float], float]:
         return lambda x: x
 
     def sample_value(
```
tommyod: Since sample_value is called in a loop, I wonder if this should be re-written to samples VALUES (plural) instead. That way we avoid the O(n) computation inside each method call, so we get O(n) overall instead of O(n^2) and a bit more clarity in the code IMO.

lucaeg: Do you mean this loop?

`datasets = [
    Ensemble.sample_parameter(
        config_node,
        realization_nr,
        random_seed=random_seed,
        num_realizations=num_realizations,
    )
    for realization_nr in active_realizations
]`

lucaeg: Do you mean this loop?

`datasets = [
                    Ensemble.sample_parameter(
                        config_node,
                        realization_nr,
                        random_seed=random_seed,
                        num_realizations=num_realizations,
                    )
                    for realization_nr in active_realizations
                ]`

tommyod: yes!

PR #12528, opened 2025-12-19
Remove redundant .name attr from response configs

yngve-sk:
Prerequisite for https://github.com/equinor/ert/pull/11513, faulty migration in `to7` before did not remove name from JSON dump of gendata, thus it has to be removed for all migrated storages.


yngve-sk:
Doing both in https://github.com/equinor/ert/pull/12529

Issue #12427, opened 2025-12-02
DESIGN_MATRIX: Fix large integers get exported in scientific notation


xjules:
### What happened? (You can include a screenshot if it helps explain)

Reported by a user:
```
Another problem they experienced was that their RMS seed number was large enough to be formatted in scientific notation when changing from DESIGN2PARAMS to DESIGN_MATRIX
```

### What did you expect to happen?

We need to make sure that the integers will remain full numbers in parameters.txt and template files when exporting values from the design matrix.

### Steps to reproduce

```console

```

### Environment where bug has been observed

- [x] Python 3.11
- [x] Python 3.12
- [x] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [x] LSF queue
- [x] Slurm queue
- [x] OpenPBS queue

PR #12535, opened 2025-12-19
Fix raw formatting for big numbers in parameter export and templating

xjules:
**Issue**
Resolves https://github.com/equinor/ert/issues/12427

**Approach**
Make sure that the formatting stays raw for integers.
Floats will use notation with `:g`  to keep the current state.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/run_models/_create_run_path.py
+++ b/src/ert/run_models/_create_run_path.py
@@ -60,15 +60,17 @@ def _value_export_txt(
     with path.open("w") as f:
         for key, param_map in values.items():
             for param, value in param_map.items():
-                if isinstance(value, (int | float)):
-                    if key == DESIGN_MATRIX_GROUP:
-                        print(f"{param} {value:g}", file=f)
-                    else:
-                        print(f"{key}:{param} {value:g}", file=f)
-                elif key == DESIGN_MATRIX_GROUP:
-                    print(f"{param} {value}", file=f)
+                if isinstance(value, int):
+                    value_str = str(value)
+                elif isinstance(value, float):
+                    value_str = f"{value:g}"
                 else:
-                    print(f"{key}:{param} {value}", file=f)
+                    value_str = str(value)
+
+                if key == DESIGN_MATRIX_GROUP:
+                    print(f"{param} {value_str}", file=f)
+                else:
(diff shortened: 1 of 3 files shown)


src/ert/run_models/_create_run_path.py
```diff
@@ -59,15 +59,17 @@ def _value_export_txt(
...
-                    if key == DESIGN_MATRIX_GROUP:
-                        print(f"{param} {value:g}", file=f)
-                    else:
-                        print(f"{key}:{param} {value:g}", file=f)
-                elif key == DESIGN_MATRIX_GROUP:
-                    print(f"{param} {value}", file=f)
+                if isinstance(value, int):
+                    value_str = str(value)
+                elif isinstance(value, float):
+                    value_str = f"{value:.6g}"
```
jonathan-eq: Are we sure we want this at .6g formatting? What will happen if the value is 1234.567464646?

xjules: I thought about that and it might get flaky without a precision; ie. `.6g` means we get always a correct rounding of the number. There are many corner cases though. Extra large and extra small numbers - not sure how realistic they are.

jonathan-eq: I am not sure this will fix the issue, as the issue is the cornercase when the number is very large. I did some querying and one number submitted by the user was ```257382775096236763020404641646935773111```

xjules: yes, but that is an integer (very likely from the design matrix), which should get a correct handling as integer.

tests/ert/ui_tests/cli/analysis/test_design_matrix.py
```diff
@@ -94,6 +94,7 @@ def test_run_poly_example_with_design_matrix_and_genkw_merge(default_values):
                 "REAL": list(range(num_realizations)),
                 "a": a_values,
                 "category": 5 * ["cat1"] + 5 * ["cat2"],
+                "big_numbers": [1e10 + i for i in range(num_realizations)],
```
oyvindeide: Perhaps rename this to big_integer for clarity?

PR #12567, opened 2026-01-05
Remove Webviz ERT and BaseService

yngve-sk:
PS: Should consider when to merge this in wrt deprecation warnings etc.

### DIFF
--- a/src/ert/services/__init__.py
+++ b/src/ert/services/__init__.py
@@ -1,3 +1,3 @@
-from .ert_server import BaseServiceExit, ErtServer, ServerBootFail
+from .ert_server import ErtServer, ErtServerExit, ServerBootFail
 
-__all__ = ["BaseServiceExit", "ErtServer", "ServerBootFail"]
+__all__ = ["ErtServer", "ErtServerExit", "ServerBootFail"]
--- a/src/ert/services/_storage_main.py
+++ b/src/ert/services/_storage_main.py
@@ -30,7 +30,7 @@ from uvicorn.supervisors import ChangeReload
 
 from ert.logging import STORAGE_LOG_CONFIG
 from ert.plugins import setup_site_logging
-from ert.services import BaseServiceExit
+from ert.services import ErtServerExit
 from ert.shared import __file__ as ert_shared_path
 from ert.shared import find_available_socket, get_machine_name
 from ert.trace import tracer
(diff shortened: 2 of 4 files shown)


src/ert/__main__.py
```diff
@@ -62,56 +62,6 @@ def run_ert_storage(args: Namespace, _: ErtRuntimePlugins | None = None) -> None
         server.wait()
 
 
-def run_webviz_ert(args: Namespace, _: ErtRuntimePlugins | None = None) -> None:
-    try:
```
yngve-sk, line 66: note2self Add this back and give the notice that it is removed and that the misfits plot is available in ERT plotter

src/ert/shared/storage/extraction.py
```diff
@@ -1,42 +0,0 @@
-from __future__ import annotations
-
-from collections.abc import Mapping
```
yngve-sk, line 3: note2self also remove misfits endpoint (assuming only used by webviz ert)

src/ert/__main__.py
berland: In this PR or in a separate PR?

yngve-sk: It is added in this PR

berland: Ok, then lets resolve this conversation.

src/ert/services/__init__.py
```diff
@@ -1,4 +1,3 @@
 from .ert_server import BaseServiceExit, ErtServer, ServerBootFail
-from .webviz_ert_service import WebvizErt
 
-__all__ = ["BaseServiceExit", "ErtServer", "ServerBootFail", "WebvizErt"]
+__all__ = ["BaseServiceExit", "ErtServer", "ServerBootFail"]
```
berland: BaseServiceExit should probably be renamed to ErtServerExit in this PR.

src/ert/services/_base_service.py
```diff
@@ -1,204 +0,0 @@
-"""
```
berland, line 1: is this file removed, or just emptied in this PR? (it looks emptied, I would suppose it should be removed)

yngve-sk: Was not, removed it now

berland (approved):
Lets cut it! 🪓 🔪

PR #12568, opened 2026-01-05
Have storage migrations done explicitly

### LINKED ISSUE #12518: Storage migrations are too implicit and should be made explicit
Currently, we migrate whenever we open a write storage on an older storage. This is too implicit and leads to bad invocation patterns that are hard to follow.

Instead, the migration should be an explicit step that is separate from opening a write storage. This does require "write access", but the invocation itself implies write access so this should be OK.

Instead of this:
```
            if LocalStorage.check_migration_needed(Path(ert_config.ens_path)):
                # Open in write mode to initialize the storage, so that
                # dark storage can be mounted onto it
                open_storage(ert_config.ens_path, mode="w").close()
```

We should have this:
```
            if LocalStorage.check_migration_needed(Path(ert_config.ens_path)):
                # Open in write mode to initialize the storage, so that
                # dark storage can be mounted onto it
                LocalStorage.migrate_storage(ert_config.ens_path)
```

This opens up for making it more explicit when and why the migration happens.

andreas-el:
**Issue**
Resolves #12518 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/storage/local_storage.py
+++ b/src/ert/storage/local_storage.py
@@ -113,7 +113,7 @@ class LocalStorage(BaseMode):
             )
 
         if not stage_for_migration:
-            self.refresh()
+            self.reload()
 
             if mode.can_write:
                 self._save_index()
@@ -139,9 +139,9 @@ class LocalStorage(BaseMode):
         if LocalStorage.check_migration_needed(path):
             with LocalStorage(path, Mode("w"), True) as storage:
                 storage._migrate(storage.version)
-                storage.refresh()
+                storage.reload()
 
-    def refresh(self) -> None:
+    def reload(self) -> None:
         """
         Reloads the index, experiments, and ensembles from the storage.
 
--- a/src/ert/dark_storage/common.py
+++ b/src/ert/dark_storage/common.py
@@ -28,7 +28,7 @@ def get_storage() -> Storage:
         except ErtStorageException as err:
             logger.exception(f"Error accessing storage: {err!s}")
             raise InternalServerError("Error accessing storage") from None
-    _storage.refresh()
+    _storage.reload()
     return _storage
 
 
--- a/src/ert/gui/ertnotifier.py
+++ b/src/ert/gui/ertnotifier.py
@@ -74,7 +74,7 @@ class ErtNotifier(QObject):
     @Slot()
     def emitErtChange(self) -> None:
         if self._storage is not None:
-            self._storage.refresh()
+            self._storage.reload()
 
         self.ertChanged.emit()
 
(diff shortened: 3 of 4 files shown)


src/ert/storage/local_storage.py
```diff
@@ -101,19 +105,18 @@ def __init__(
...
+            raise RuntimeError(
+                f"Cannot open storage '{self.path}' in read-only mode: "
+                f"Storage version {self.version} is too old. "
+                f"Run ert to initiate migration."
+            )
 
-        self.refresh()
-        if mode.can_write:
-            self._save_index()
+        if not stage_for_migration:
```
yngve-sk, line 115: Do we need this piece of logic (stage_for_migration conditional) if we anyway do migrations before opening storage? (I think the check above makes sense for sure though)

src/ert/storage/local_storage.py
```diff
@@ -123,15 +126,22 @@ def check_migration_needed(storage_dir: Path) -> bool:
...
                 f", upgrade ert to continue, or run with a different ENSPATH"
             )
 
         return version < _LOCAL_STORAGE_VERSION
 
-    def refresh(self) -> None:
+    @staticmethod
+    def perform_migration(path: Path) -> None:
+        if LocalStorage.check_migration_needed(path):
+            with LocalStorage(path, Mode("w"), True) as storage:
```
yngve-sk, line 140: This is a lot better than what is currently here, but would it be hard to get more separation and make the migration happen without opening a write storage storage?

src/ert/storage/local_storage.py
```diff
@@ -536,7 +546,7 @@ def _migrate(self, version: int) -> None:
 
                 self._index = self._load_index()
 
-                logger.info("storage backed up for version less than 6")
+                logger.info("Storage backed up for version less than 5")
```
yngve-sk, line 549: 6->5 intended?

andreas-el: Yes. This was a typo made from way back.

src/ert/storage/local_storage.py
andreas-el: Due to how LocalStorage is currently, pydantic will fail if we load non-migrated storage. This worked before since we migrated during construction of the object.

yngve-sk: I think this reduces to the "can we migrate without opening a storage" Q, so closing this as duplicate of my other comment.

src/ert/storage/local_storage.py
andreas-el: Perhaps. Regardless I think that should be part of a different PR.

yngve-sk: 👍 Can you set up a followup issue when merging this?

andreas-el: https://github.com/equinor/ert/issues/12600

PR #12593, opened 2026-01-07
A Gentle Introduction to History Matching

dafeda:
      --exec 'just rapid-tests'`)

## When applicable


docs/ert/theory/gentle_introduction_to_history_matching.rst
```diff
@@ -0,0 +1,534 @@
...
+Data that are only measured at wells are called **production data**.
+There are other types of data as well, such as **seismic data** that allows us to measure properties of the reservoir at locations between wells, but let's just focus on production data for now.
+
+We can also sample the actual rock at the well locations and measure properties such as **porosity** and **permeability**.
+The rock properties in every other position, or grid-cell :math:`x_i` are uncertain and must be estimated.
+We call these the **parameters**.
+
+So, the main question we want to answer is this:
+
+    What is the reservoir made of where direct observations can't be made?
```
dafeda, line 30: Maybe more accurate to say that the aim is to build a predictive model etc.

docs/ert/theory/gentle_introduction_to_history_matching.rst
```diff
@@ -0,0 +1,534 @@
...
+
+Again, we can only directly observe the rock at the well locations.
+The rock properties in all other positions are uncertain and must be estimated.
```
lucaeg, line 33: But observations are also uncertain? You kind of make it seem like they are not here.

docs/ert/theory/gentle_introduction_to_history_matching.rst
```diff
@@ -0,0 +1,534 @@
...
+3. Manually check the simulated results and see how well they fit data from real world sensors
+
+If the simulated values are close to the observed ones, we can be somewhat confident that our guess was good.
+On the other hand, if the simulated values do not match the observed values, we can make a different guess and repeat the process.
+You can keep on doing this until you find a set of parameter values that give you a sufficiently good match and call it a day.
+You can then use these parameters to make predictions, do well-planning etc.
+
+It's worth spending some time understanding this figure and introduce some terminology.
+Recall that :math:`w_1` and :math:`w_2` represent wells and :math:`x_i` represents the rock properties at grid-cell :math:`i`.
+The notation :math:`x_i \sim \mathcal{N}(\mu, \sigma^2)` means that the rock property in every grid-cell :math:`i` is normally distributed with mean :math:`\mu` and variance :math:`\sigma^2`.
```
lucaeg, line 53: Is it possible to fix the mathematical notation? I'm not sure if latex-notation is clear to everybody.

docs/ert/theory/gentle_introduction_to_history_matching.rst
```diff
@@ -0,0 +1,534 @@
...
+
+.. figure:: images/intro_assisted_history_matching.png
+
+The process of manually guessing parameters-values, running simulators and checking results can be automated.
+
+We start by generating many different guesses for the parameters.
+This set of guesses is called an **ensemble** and each member of an ensemble is called a **realization**.
+The number of realizations is the **ensemble size** which is usually denoted by :math:`N`.
+Note the change between manual history matching and AHM: :math:`x_i \sim \mathcal{N}(\mu_j, \sigma_j^2)`.
+The addition of the subscript :math:`j` means that every realization uses a different mean :math:`\mu_j` and a different variance :math:`\sigma_j^2`.
```
lucaeg: Is this true?? I thought we sampled from one distribution per parameter.

docs/ert/theory/gentle_introduction_to_history_matching.rst
```diff
@@ -0,0 +1,534 @@
...
+Once the ensemble is ready, we run the simulator once for each realization.
+
+The scatter plots illustrate results if we ran 3 realizations.
+At each time-step, there are 3 simulated values (blue dots) and one observed value (black dot).
+The simulated values at :math:`w_1` seem to cover the observed values, meaning that there are simulated values both above and below the observed values.
+However, the simulated values at :math:`w_2` are all above the observed values.
+This is an example of poor **coverage** and is a sign that our choice of parameter values is not good.
+If the simulated values are all above the observed values, our model is fundamentally biased.
+This should be fixed before history matching.
+
```
lucaeg, line 90: This is not intuitive to me, does "fixing" your model not simply introduce a different type of bias? What if our observations are wrong/imprecise. Can we not have that the truth lies between the simulated and the observed data?

docs/ert/theory/gentle_introduction_to_history_matching.rst
```diff
@@ -0,0 +1,534 @@
...
+That's all we need to do an **update**, which is a function we run to get a new ensemble of parameter guesses.
+The new parameter estimates are hopefully better guesses than the ones we started with.
+
+Let's introduce some math notation.
+
+:math:`X \in \mathbb{R}^{p \times N}` is a common way of defining a matrix :math:`\mathbf{X}` that consist of real-numbers (think floats) and that has :math:`p` rows (number of parameters) and :math:`N` columns (number of realizations or samples).
+Similarly, :math:`Y \in \mathbb{R}^{m \times N}` specifies a response matrix :math:`\mathbf{Y}` with :math:`m` rows (number of responses = number of observations) and :math:`N` columns.
+Note that both :math:`\mathbf{d}` and :math:`\mathbf{\epsilon}` have :math:`m` rows but only 1 column.
+The reason for this will have to wait a bit.
+
```
lucaeg, line 137: I would reformulate the last sentence.

docs/ert/theory/gentle_introduction_to_history_matching.rst
```diff
@@ -0,0 +1,534 @@
...
+What is the difference between prior and posterior?
+
+Results of updating
+~~~~~~~~~~~~~~~~~~~
+
+.. figure:: images/intro_result_of_updating.png
+
+Let's think about the update somewhat more concretely by focusing on a single parameter from a single grid-cell; grid-cell :math:`x_2` and porosity.
+
+The upper-left figure shows the distribution of :math:`x_2` and its true value.
```
lucaeg: This is the upper right figure.

docs/ert/theory/gentle_introduction_to_history_matching.rst
```diff
@@ -0,0 +1,534 @@
...
+You don't have to be able to derive the equation in order to gain an intuitive understanding of it.
+
+The following figure shows an outline of a derivation which you can safely skip.
+In essence, you start with Bayes' Theorem and assume that both the prior and likelihood are Gaussian.
+Multiplying these two Gaussians gives another Gaussian (after some algebra), and you end up with the Ensemble Smoother equations.
+Note that the assumption of a Gaussian likelihood is a strong assumption given that :math:`g(.)` is non-linear.
+That's enough derivation though!
+
+.. figure:: images/intro_bayes_derivation.png
+
```
lucaeg, line 220: A lot of UMOs (unidentified mathematical objects) in this picture that are not explained in the text. What is C_xx, C_dd? I think it would also be beneficial to explain the relationship between g and f, it is a little confusing that f suddenly pops up.

docs/ert/theory/gentle_introduction_to_history_matching.rst
```diff
@@ -0,0 +1,534 @@
...
+One example is two temperature sensors that measure the same thing.
+
+We don't want to act as if we have two independent sources of information.
+A high correlation here dampens this effect so that we don't update too strongly.
+
+Observation Error Covariance
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. figure:: images/intro_obs_error_covariance.png
```
lucaeg, line 357: I think the question in this image is a bit unmotivated and disruptive, maybe it would be better to write something like "We always assume R to be diagonal".

docs/ert/theory/gentle_introduction_to_history_matching.rst
```diff
@@ -0,0 +1,534 @@
...
+The threshold can be derived by starting with two random variables that are independent and bivariate normal.
+In this case, it can be shown that the threshold gets rid of 99.7 percent of spurious correlations.
+In reality though, we might have millions of parameters so this assumption does not hold at all, and neither does the result of removing 99.7 percent of spurious correlations.
+
+To summarize, Adaptive Localization will reduce the effects of spurious correlations, but it is an ad-hoc method without a strong theoretical foundation and has known weaknesses.
+The perhaps main weaknesses are that:
+
+- Real but weak correlations might be removed.
+- Clamping weak correlations to zero gives noisy updates which do not look good in the eyes of geologists. This could perhaps be fixed by smoothing out the results.
+
```
lucaeg, line 518: I would also add to the list of weaknesses that not all spurious correlations are removed, in fact you are keeping the worst ones.

docs/ert/theory/gentle_introduction_to_history_matching.rst
```diff
@@ -0,0 +1,534 @@
...
+The only input it needs is the correlation threshold.
+The method will also make it more obvious that including lots of parameters via fields and surfaces is perhaps not necessary,
+as it will not update most of the parameters anyway, since the correlations between most parameters and most responses is very low.
+
+Distance based localization
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. figure:: images/intro_distance_based_localization.png
+
+Distance based localization is considered the industry standard way of dealing with spurious correlations.
```
lucaeg: I would rewrite here, just to make it very clear that we are still dealing with mitigating spurious correlations. 

Distance based localization is another way of dealing with spurious correlations, and it is considered the industry standard for doing so.

docs/ert/theory/gentle_introduction_to_history_matching.rst
```diff
@@ -0,0 +1,534 @@
...
+
+The algorithm goes like this:
+
+For every parameter:
+
+- Calculate the distance between the parameter and every well.
+- Only use responses from wells that are less than :math:`r` away.
+
+One way to think of the difference between distance based,- and adaptive localization is that adaptive localization uses correlation as a measure of distance,
+while distance based localization uses euclidean distance (think actual meters) as a measure of distance.
```
lucaeg: I think this explanation is a little confusing, I would replace it with 

One way to think of the difference between distance based,- and adaptive localization is that adaptive localization uses correlations to decide which parameters to update,
while distance based localization uses euclidean distance to decide which parameters to update.

docs/ert/theory/gentle_introduction_to_history_matching.rst
```diff
@@ -0,0 +1,534 @@
...
+
+After the second update has completed, we run the final simulation step and are done.
+
+``ert`` by default runs 3 steps.
+The limiting factor is computation as running the simulation step can take days for large models.
+
+Note that ES is just a single step of ES-MDA with the scaling factor set to 1.0.
+
+.. figure:: images/intro_esmda.png
+
```
lucaeg, line 593: You captured the toolbar in this screenshot :)

docs/ert/theory/gentle_introduction_to_history_matching.rst
dafeda: What if I add this after defining **observations**:
"Note that sensors are not error-free, so observations include measurement uncertainty."

docs/ert/theory/gentle_introduction_to_history_matching.rst
dafeda: What do you mean by fixing the mathematical notation?

When the `.rst` file is rendered it looks like this:

image

docs/ert/theory/gentle_introduction_to_history_matching.rst
dafeda: Oh great catch!
That is not true no, not sure what I was thinking :D
I have pushed a proposal:

image

I need to update the notation in the figure.

docs/ert/theory/gentle_introduction_to_history_matching.rst
lucaeg: Sounds good!

docs/ert/theory/gentle_introduction_to_history_matching.rst
lucaeg: Sorry, my bad! I was looking in GitHub and the rendering was not great.

docs/ert/theory/gentle_introduction_to_history_matching.rst
dafeda: I have pushed a new proposal. Is this better?

image

docs/ert/theory/gentle_introduction_to_history_matching.rst
dafeda: I've made an attempt.

image

docs/ert/theory/gentle_introduction_to_history_matching.rst
dafeda: Good point, thanks, I think it might be best to not have a question in the figure at all.

docs/ert/theory/gentle_introduction_to_history_matching.rst
dafeda: Thanks, I added this to the list:

- Large spurious correlations can survive thresholding and still drive updates.

docs/ert/theory/gentle_introduction_to_history_matching.rst
dafeda: Fixed

docs/ert/theory/gentle_introduction_to_history_matching.rst
dafeda: That's better, thanks 👍

docs/ert/theory/gentle_introduction_to_history_matching.rst
dafeda: Looks OK here I think:

image

docs/ert/theory/gentle_introduction_to_history_matching.rst
lucaeg: This is great!

docs/ert/theory/gentle_introduction_to_history_matching.rst
lucaeg: Sorry, I forgot what I originally meant here, but I think the description you have here is good :)

docs/ert/theory/gentle_introduction_to_history_matching.rst
lucaeg: Great! :D

docs/ert/theory/gentle_introduction_to_history_matching.rst
lucaeg: Image

I meant this one. Sorry, I put the comment at the wrong place!

dafeda: Ah, that's actually on purpose to show that it's on GitHub and has some stars 😄

PR #12598, opened 2026-01-08
Fix small issues in everest docs

SAKavli:
Am reading the documentation, so I will fix any issues I find along the way.

This will be squashed.

### DIFF
--- a/docs/everest/example.rst
+++ b/docs/everest/example.rst
@@ -4,20 +4,22 @@
 Example
 *******
 
-In this section we will go through the setup and running of a simple Everest optimization case.
+In this section we will go through the setup and running of a simple Everest optimization
+case.
 
 Problem definition
 ##################
 
-The optimization problem we will solve using Everest will be to minimize the distance between an initial 3D point
-and another given target 3D point.
+The optimization problem we will solve using Everest will be to minimize the distance
+between an initial 3D point and another given target 3D point.
 
 Everest config
 ##################
 
-The main element an Everest optimization experiment requires is a configuration file. Everest supports the `YAML <https://en.wikipedia.org/wiki/YAML>`_
-format for the configuration file. The configuration file is built up of multiple sections and next we will go through it section
-by section.
+The main element an Everest optimization experiment requires is a configuration file.
+Everest supports the `YAML <https://en.wikipedia.org/wiki/YAML>`_ format for the
+configuration file. The configuration file is built up of multiple sections and next we
+will go through it section by section.
 
 Creating the optimization folder::
 
@@ -77,11 +79,12 @@ everest_config.yml::
             min: -1.0
             max: 1.0
 
-``perturbation_magnitude`` controls the size of perturbations (standard deviation of a normal distribution) of controls used to approximate the gradient.
-The value depends on the type of control and magnitude of the variables.
-For continuous controls smaller values should give a better gradient,
-whilst for more discrete controls larger values should give a better
-result. However, this is a balance as too large or too small values also cause issues.
+``perturbation_magnitude`` controls the size of perturbations (standard deviation of a
+normal distribution) of controls used to approximate the gradient. The value depends on
+the type of control and magnitude of the variables. For continuous controls smaller values
+should give a better gradient, whilst for more discrete controls larger values should give
+a better result. However, this is a balance as too large or too small values also cause
+issues.
 
 ``variables``: list of control variables.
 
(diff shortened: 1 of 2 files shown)


docs/everest/introduction.rst
```diff
@@ -9,15 +9,15 @@ Background and motivation
 
 Everest is a tool, which uses ensembles of models to deliver a robust optimization.
 
-While the result of an data assimilation process is an ensemble of models, the result of the robust optimization process is one optimal strategy, for example for well order or drainage strategy. With this optimised strategy we can, e.g., maximize net present value (NPV) with respect to the geological uncertainty in the models. Note that Everest is a generic tool that can be applied to other optimisation problems (also outside the domain of Oil & Gas).
+While the result of a data assimilation process is an ensemble of models, the result of the robust optimization process is one optimal strategy, for example for well order or drainage strategy. With this optimized strategy we can, e.g., maximize net present value (NPV) with respect to the geological uncertainty in the models. Note that Everest is a generic tool that can be applied to other optimization problems (also outside the domain of Oil & Gas).
```
berland: Equinor writes in British English, where 'optimise' is the correct spelling.

SAKavli: Then I will change it! 
It seems like "realization" is fine in British English according to my googling.

docs/everest/example.rst
```diff
@@ -4,22 +4,22 @@
...
-The optimization problem we will solve using everest will be to minimize the distance between an initial 3D point
+The optimisation problem we will solve using Everest will be to minimize the distance between an initial 3D point
 and another given target 3D point.
 
 Everest config
 ##################
 
-The main element an Everest optimization experiment requires is a configuration file. Everest supports the `YAML <https://en.wikipedia.org/wiki/YAML>`_
+The main element an Everest optimisation experiment requires is a configuration file. Everest supports the `YAML <https://en.wikipedia.org/wiki/YAML>`_
 format for the configuration file. The configuration file is build up of multiple sections and next we will go through it section
```
berland: build -> built

docs/everest/example.rst
```diff
@@ -95,18 +95,18 @@ everest_config.yml::
...
 
 
-Optimization definition
+Optimisation definition
 -----------------------
 
 everest_config.yml::
 
-    optimization:
+    optimisation:
```
berland: This we cannot fix, this is the yaml syntax which (unfortunately) is american english.

docs/everest/example.rst
```diff
@@ -320,18 +320,18 @@ distance3d.py::
...
+Now we have all the components needed to start the optimisation experiment,
 which can be done using the following command::
 
     ~/everest_example$: everest run everest_config.yml
 
 
 A successful run should provide the following information in the console::
 
-    ===================================    Optimization progress     ===================================
+    ===================================    Optimisation progress     ===================================
```
berland: This will be wrong unless there is a change also in the source code

docs/everest/introduction.rst
```diff
@@ -7,34 +7,34 @@ Introduction
 Background and motivation
 =========================
 
-Everest is a tool, which uses ensembles of models to deliver a robust optimization.
+Everest is a tool, which uses ensembles of models to deliver a robust optimisation.
 
-While the result of an data assimilation process is an ensemble of models, the result of the robust optimization process is one optimal strategy, for example for well order or drainage strategy. With this optimised strategy we can, e.g., maximize net present value (NPV) with respect to the geological uncertainty in the models. Note that Everest is a generic tool that can be applied to other optimisation problems (also outside the domain of Oil & Gas).
+While the result of a data assimilation process is an ensemble of models, the result of the robust optimisation process is one optimal strategy, for example for well order or drainage strategy. With this optimised strategy we can, e.g., maximize net present value (NPV) with respect to the geological uncertainty in the models. Note that Everest is a generic tool that can be applied to other optimisation problems (also outside the domain of Oil & Gas).
```
berland: these long lines would benefit from some line wrapping in the editor.

docs/everest/example.rst
SAKavli: Good catch

docs/everest/example.rst
SAKavli: Should we keep optimisation where we can - also in the code, or should I just scrap that commit?

SAKavli:
I thought it was a good idea to improve the docs while I went through it, but perhaps I've caused enough ruccus and should leave it as is more or less, before something breaks.

berland:
Squash all or selectively?

SAKavli:
Most commits touch the same lines - as they were huge, aka merge conflicts. Squashing all would save that headache. Can keep newlines as a separate commit to make the grammatical changes more readable.

PR #12625, opened 2026-01-12
Move makedirs_if_needed from everest to ert

berland:
This function is not used exclusively by Everest, so it should be moved to a location where it is natural to be shared between Ert and Everest.

Its logging was everest specific - which was a bug when called from the Ert side, now we do not try to state in the logging from which side we were called from

**Issue**
Resolves ert code depending on everest code.

**Approach**
🚛 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/utils/__init__.py
+++ b/src/ert/utils/__init__.py
@@ -1,8 +1,12 @@
 import logging
+import os
 import time
 from collections.abc import Callable
+from datetime import UTC, datetime
 from functools import wraps
 from typing import ParamSpec, TypeVar
+
+from _ert.utils import file_safe_timestamp
 
 P = ParamSpec("P")
 R = TypeVar("R")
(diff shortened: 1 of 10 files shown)


src/ert/utils/__init__.py
```diff
@@ -30,3 +34,19 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
         return wrapper
 
     return decorator
+
+
+def makedirs_if_needed(path: str, roll_if_exists: bool = False) -> None:
+    if os.path.isdir(path):
```
verveerpj, line 40: Do we not prefer `Pathlib`?

src/everest/util/__init__.py
```diff
@@ -1,29 +1,19 @@
...
-from datetime import UTC, datetime
 
 from ropt.version import version as ropt_version
 
 try:
     from ert.shared.version import version as ert_version
 except ImportError:
     ert_version = "0.0.0"
-from _ert.utils import file_safe_timestamp
+from _ert.utils import file_safe_timestamp as file_safe_timestamp
```
verveerpj: Am I missing something? Or is this importing as the same name?

src/ert/utils/__init__.py
berland: we do, but I am saving this for a future PR

src/everest/util/__init__.py
berland: This is an unused symbol, so it is intriguing `ruff` let this pass through. Looks like a ruff-bug, because it is only when it is imported as the same name that ruff lets it pass through.

berland: Ah, it is a feature when an import is re-exported in the same name, ruff will not touch it :)  See `ruff rule F401`

berland: (removed the line)

src/ert/utils/__init__.py
berland: Fixed in https://github.com/equinor/ert/pull/12630

Issue #12619, opened 2026-01-12
Create conversion tool for HISTORY_OBSERVATION to SUMMARY_OBSERVATION


eivindjahren:

SAKavli:
Suggestion:
- When we detect a deprecated keyword which we REQUIRE the user to fix immediately, the user should receive a Suggestor window.
- This should count as an error, the user will not be able to proceed to ert before fixing the deprecation. This helps them migrate asap instead of keeping the deprecation warning for months before the feature is removed.
- The window should not contain anything else than the migration dialogue(s), which contains a box(es) containing the correct config lines which can easily copied. 
- All migration messages should be displayed simultaneously, such that the user doesn't close the dialogue, fix one problem, then another one pops up. Then the user have no idea how many more deprecation errors they have in their config -> Frustration 😠 
- There should be a path to the file which needs to be fixed which could also be copy pasted. This does however introduce some clipboard conflict should the user close the Suggestor window. Then they only have either the path or the corrected lines copied.

I was also curious if it makes sense to have the old and new lines side by side to visualize what is deprecated and what is correct? We could have old version in a red error box with the corrected version in a green box side by side. This could be reused for future deprecations.

PR #12629, opened 2026-01-13
Migrate timemap/refcase dependent observations

yngve-sk:
(Goes with https://github.com/equinor/semeio/pull/862)

Overview of changes
================
* Remove `REFCASE`, `TIME_MAP`, and `HISTORY_OBSERVATION` from ERT
* Add tool to migrate older ERT configs so that information from `REFCASE` and `TIME_MAP` is embedded into the obs config
* `GENERAL_OBSERVATION` can now only be used with `RESTART`, usage of `DATE`, `HOURS`, `DAYS` can be migrated if there is a `REFCASE` or `TIME_MAP` in the ERT config.
* `SUMMARY_OBSERVATION` can now only be used with `DATE`, usage of `RESTART`, `HOURS`, `DAYS` can be migrated if there is a `REFCASE` or `TIME_MAP` in the ERT config.

Nice to know:
* Tests that were previously testing now deprecated behaviors have been kept by shimming the obs config migration into the test.
* Removed test case and logic for asserting that a SUMMARY_OBSERVATION's date matches up with the refcase / time map (this will be hard to pick up in migration, so suggest just dropping it)

**Issue**
Resolves https://github.com/equinor/ert/issues/12619

**Residual (IDEALLY NON BLOCKING) issues/TODOs**
Recommend resolving before next release:
* Remove FOPR observations from snake oil (i.e., remove the whole set of observations, and expect some snapshot changes) https://github.com/equinor/ert/issues/12681
* Make `_GeneralObservation.restart` non optional https://github.com/equinor/ert/issues/12682
* Reconsider the precision of the history obs https://github.com/equinor/ert/pull/12629#discussion_r2704675571 https://github.com/equinor/ert/issues/12683

Should def be fixed:
* Move validations/errors wrt observations parsing from `_create_observation_dataframes.py` to `_observations.py`. Will likely fix this in the following PR where `*Observation`s are converted to `BaseModel` objs, opening up for moving validations there (instead of doing the the `dataclasses.dataclass` way now) https://github.com/equinor/ert/issues/12680

Not urgent:
* Don't write info about the CLI tool into `_observations.py`, ref https://github.com/equinor/ert/pull/12629#discussion_r2708047641 . https://github.com/equinor/ert/issues/12685
* Port the poly template example to not use history obs, right now it creates a refcase and uses history obs to make observations based on that. Current workaround is just migrating it https://github.com/equinor/ert/issues/12684
* Replace test usages of `run_convert_observations` with just migrating the observations and taking it from there (more instances of this can be found) https://github.com/equinor/ert/issues/12686


src/ert/config/observation_config_migrations.py
```diff
@@ -0,0 +1,257 @@
...
+    summary_obs_changes = []
+    for obs in summary_observations:
+        date = obs_time_list[obs.restart]
+        declaration = (
+            f"SUMMARY_OBSERVATION {obs.name} {{\n"
+            f"    VALUE    = {obs.value};\n"
+            f"    ERROR    = {obs.error};\n"
+            f"    DATE     = {date};\n"
+            f"    KEY      = {obs.key};\n"
+            + (f"LOCATION_X={obs.location_x};\n" if obs.location_x is not None else "")
```
SAKavli: missing indent

src/ert/config/ert_config.py
oyvindeide: This should not be here?

src/ert/config/_create_observation_dataframes.py
```diff
@@ -25,59 +24,35 @@
...
     ErrorInfo,
-    HistorySource,
     ObservationConfigError,
 )
-from .refcase import Refcase
 from .rft_config import RFTConfig
 
 if TYPE_CHECKING:
-    import numpy.typing as npt
+    pass
```
oyvindeide: Can be removed

src/ert/config/_create_observation_dataframes.py
```diff
@@ -393,20 +210,17 @@ def _handle_general_observation(
...
-    if all(
-        getattr(general_observation, key) is None
-        for key in ["restart", "date", "days", "hours"]
-    ):
-        # The user has not provided RESTART or DATE, this is legal
-        # for GEN_DATA, so we default it to None
-        restart = None
-    else:
-        restart = _get_restart(general_observation, obs_key, time_map, has_refcase)
+    if general_observation.date is not None:
```
oyvindeide: Is this possible to add to the lark parser directly, so we dont even need `.date` here? Same applies for summary above

src/ert/config/_create_observation_dataframes.py
```diff
@@ -558,3 +372,22 @@ def _handle_rft_observation(
             "std": pl.Series([rft_observation.error], dtype=pl.Float32),
         }
     )
+
+
+def _handle_error_mode(
```
oyvindeide: This is just moved?

src/ert/dark_storage/endpoints/observations.py
```diff
@@ -84,6 +84,7 @@ async def get_observations_for_response(
         ensemble.experiment,
         obs_keys,
         json.loads(filter_on) if filter_on is not None else None,
+        response_type=response_type,
```
oyvindeide: Why is this needed?

test-data/ert/snake_oil/observations/observations.txt
```diff
@@ -1,4 +1,1402 @@
-HISTORY_OBSERVATION FOPR;
```
oyvindeide, line 1: Would consider just removing `FOPR` from `snake_oil`, will create a large diff with the snapshots, sure, but will reduce the size here significantly, and dont think we lose any coverage.

oyvindeide (commented):
Not a big deal, but I get one error and one warning about HISTORY_OBSERVATION

Image

If I have an existing `obs_config.old` that just gets overwritten.

oyvindeide:
Should also remove `REFCASE`/`TIME_MAP` from `snake_oil`?

src/ert/config/ert_config.py
yngve-sk: Removed

src/ert/config/_create_observation_dataframes.py
yngve-sk: Yes, will put back to make less of a diff

test-data/ert/snake_oil/observations/observations.txt
yngve-sk: I think it should be removed, but suggest keeping it and removing it in a separate commit at the end of the PR, or in a separate following PR. The snapshots NOT changing from migrating the obs config is in itself a confirmation that the migration works in that regard.

yngve-sk:
> Should also remove `REFCASE`/`TIME_MAP` from `snake_oil`?

Can do if that is most ideal? Was not sure whether we should do an inline edit to the ert config and not only the obs config.

src/ert/dark_storage/endpoints/observations.py
yngve-sk: First 2 commits of PR should have descriptions of why they are there. Core issue is that observation names may be overlapping across response types. Without these, running snake oil will error when trying to inspect observations via:

* Manage experiments->select Ensemble->view observations
* Observation named FOPR towards a response different from FOPR (Two observations with the same obs name for different response type, like rename general obs to same as one towards FOPR for example)

will expand this answer after reproducing the issue once again.

src/ert/dark_storage/endpoints/observations.py
```diff
@@ -107,10 +108,14 @@ def _get_observations(
     experiment: Experiment,
     observation_keys: list[str] | None = None,
     filter_on: dict[str, Any] | None = None,
+    response_type: str | None = None,
 ) -> list[dict[str, Any]]:
     observations = []
 
-    for response_type, df in experiment.observations.items():
+    for response_type_, df in experiment.observations.items():
+        if response_type is not None and response_type_ != response_type:
```
berland: what about naming these `stored_response_type` and `requested_response_type`?

src/ert/config/observation_config_migrations.py
```diff
@@ -0,0 +1,695 @@
...
+    for history_obs in history_observations:
+        history_obs_df = _handle_history_observation(
+            refcase, history_obs, history_obs.name, history_source, len(obs_time_list)
+        )
+        declarations = []
+        for obs_row in history_obs_df.to_dicts():
+            declaration = (
+                f"SUMMARY_OBSERVATION "
+                f"{obs_row['observation_key']} {{\n"
+                f"   VALUE    = {obs_row['observations']};\n"
```
berland, line 529: I think it is better to add `:g` to the value and error, to avoid an absolute value of `0.01` getting formatted as `0.099999123456` (it will prefer exponential format sometimes, but that is maybe fine enough here)

yngve-sk: It gives some discrepancy in `test_that_history_observations_values_are_fetched_from_refcase`, not sure anymore... Removed the commit temporarily to just have other tests pass, can add it back if we conclude it should be there.

test-data/ert/snake_oil/observations/observations.txt
oyvindeide: Separate commit or PR sounds like a good idea 👍

docs/ert/reference/configuration/observations.rst
```diff
@@ -203,14 +141,15 @@ the updating process. Therefore, it is important to have consistency
...
+The default error mode and values can be changed as follows:
 
 .. code-block:: none
 
- HISTORY_OBSERVATION GOPR:FIELD
+ SUMMARY_OBSERVATION GOPR:FIELD
  {
+    VALUE      = 0.9;
+    DATE       = 2014-09-10;
+    KEY        = FOPR;
```
oyvindeide: Minor comment, but the `KEY` should be `GOPR:FIELD` and the name of the observation something else.

src/ert/config/_create_observation_dataframes.py
yngve-sk: Fixed, only detected at the config dict parsing step, and the data will not be present in the objects that are to be serialized and passed to the runmodel side

src/ert/dark_storage/endpoints/observations.py
yngve-sk: Screenshot 2026-01-20 at 08 14 57
From disallowing duplicate obs names at ert config level (even though they are allowed downstream on current main) we get this

yngve-sk: If we allow the duplicate names, and in snake oil for example rename `GENERAL_OBSERVATION WPR_DIFF_1 ` to `GENERAL_OBSERVATION FOPR `, the GUI does not differentiate by response type and will mix them up, leading to possible fails in `Manage experiments->select Ensemble->view observations`

src/ert/config/_observations.py
```diff
@@ -15,115 +16,61 @@
...
                     pass
-                case "RESTART":
-                    date_dict.restart = validate_positive_int(value, key)
                 case "ERROR" | "ERROR_MIN":
                     float_values[str(key)] = validate_positive_float(value, key)
-                case "DAYS" | "HOURS":
-                    setattr(
-                        date_dict, str(key).lower(), validate_positive_float(value, key)
+                case "DAYS" | "HOURS" | "RESTART":
+                    raise ObservationConfigError.with_context(
```
oyvindeide, line 65: How much work is it to raise this as a specific error that we then catch higher up to add context, for example `SummaryObservationError`, the reason being that we dont need the context of the cli tool in this part of the code, that can live higher in the stack. Should still be available in the suggester in the same way as now, so should reraise it as `ObservationConfigError` higher up (or another type that brings it into the suggestor)

yngve-sk: ObservationConfigError is a ConfigValidationError, so it should show up in suggester? 
Screenshot 2026-01-20 at 12 15 26
Screenshot 2026-01-20 at 12 15 38


Screenshot 2026-01-20 at 12 29 17
Screenshot 2026-01-20 at 12 29 30

src/ert/config/_observations.py
```diff
@@ -15,115 +16,61 @@
...
+                        (
+                            "SUMMARY_OBSERVATION must use DATE to specify "
+                            "date, DAYS | HOURS is no longer allowed. "
+                            "Please run ert convert_observations "
+                            "<your_ert_config.ert> to migrate the observation config "
+                            "to use the correct format"
```
andreas-el: Would it be possible to format this somewhat?
E.g.
Please run:
ert convert_observations <your_ert_config.ert>
This will migrate the observation config to use the correct format.

src/ert/config/_observations.py
```diff
@@ -392,8 +349,18 @@ def make_observations(
...
     result: list[Observation] = []
-    error_list: list[ErrorInfo] = []
+    error_list: list[ErrorInfo | ObservationConfigError] = []
     for obs_dict in observation_dicts:
+        if obs_dict["type"] == ObservationType.HISTORY:
+            msg = (
+                "HISTORY_OBSERVATION is deprecated, and must be specified "
+                "as SUMMARY_OBSERVATION. Run"
+                " ert convert_observations <ert_config.ert> to convert your "
+                "observations automatically"
```
andreas-el, line 359: Different text than the others: "to migrate the observation config to use the correct format"

yngve-sk: Fixed

src/ert/config/_observations.py
yngve-sk: Fixed

src/ert/__main__.py
```diff
@@ -53,6 +58,67 @@
...
+    if changes is None or changes.is_empty():
+        logger.info("convert_observations did not make any changes")
+        print(
+            "No observations dependent on TIME_MAP / REFCASE found, you can "
+            "safely remove TIME_MAP / REFCASE and the "
+            "corresponding files from ERT config."
+        )
+        return
+
+    logger.info("convert_observations start")
```
andreas-el: This can probably be removed.

tests/ert/unit_tests/config/observations_generator.py
```diff
@@ -159,9 +149,12 @@ def observations(draw, ensemble_keys, summary_keys, std_cutoff, start_date):
...
         )
     if summary_keys is not None:
-        observation_generators.extend(
-            (
-                summary_observations(summary_keys, std_cutoff, unique_names, datetimes),
+        observation_generators.append(
+            summary_observations(summary_keys, std_cutoff, unique_names, datetimes),
+        )
+
+        if False:
```
andreas-el: What do you accomplish with this?

yngve-sk: Oops!! The whole block it wraps should just be removed, and is removed now.

src/ert/config/_observations.py
oyvindeide: The message is good, I was thinking about the fact that `_observations.py` knows there exists and `cli` tool, while logically that information should exist higher in the stack `_observations.py` only knows that `HISTORY_OBSERVATION` is deprecated, and we can add more context higher in the stack.

yngve-sk: Oh, I see. I think this could be a residual followup issue, likely requires some design decisions (first idea that comes to my mind is having an own observation error type, indicating that it needs migration, and taking it from there) and moving of some things.

andreas-el: This will also allow us to inject the path to the ert-file, so that the feedback message actually contains a valid command.
E.g.
`ert convert_observations test-data/ert/poly_example/poly.ert` instead of that placeholder.

PR #12648, opened 2026-01-15
Vectorize parameter sampling

lucaeg:
Vectorize the sampling rather than drawing one value at a time. This cosmetic change makes the code cleaner and easier to understand. In our view, drawing all samples for a parameter simultaneously is more natural and better aligned with standard sampling APIs (e.g., scipy).

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/parameter_config.py
+++ b/src/ert/config/parameter_config.py
@@ -142,7 +142,8 @@ class ParameterConfig(BaseModel):
         ensure reproducibility across runs.
         - active_realizations (list[int]): indices of the realizations
         to select from the sampling vector; each must satisfy 0 <= i < num_realizations.
-        - num_realizations (int): Total number of realizations.
+        - num_realizations (int): Total number of realizations. Assures stable sampling
+        for a given global_seed regardless of currently active realizations.
 
         Returns:
         - npt.NDArray[np.double]: Array of shape (len(active_realizations),


src/ert/config/parameter_config.py
```diff
@@ -126,10 +126,8 @@ def group_name(self) -> str:
...
 
-    def sample_value(
-        self,
-        global_seed: str,
-        realization: int,
+    def sample_values(
+        self, global_seed: str, active_realizations: list[int], num_realizations: int
     ) -> npt.NDArray[np.double]:
         """
         Generate a sample value for each key in a parameter group.
```
achaikou: Shouldn't the docs be updated with regards to new parameters?

src/ert/config/parameter_config.py
```diff
@@ -162,9 +160,7 @@ def sample_value(
...
-        # Advance the RNG state to the realization point
-        rng.standard_normal(realization)
-
-        # Generate a single sample
-        value = rng.standard_normal(1)
-        return np.array([value[0]])
+        # Generate samples for all active realizations
+        all_values = rng.standard_normal(num_realizations)
+        idx = np.asarray(active_realizations, dtype=int)
+        return all_values[idx]
```
achaikou: Why do you need `idx` variable to turn the list into a numpy array?
I think this should work as well
```
return all_values[active_realizations]
```

src/ert/config/parameter_config.py
```diff
@@ -162,9 +160,7 @@ def sample_value(
...
         rng = np.random.default_rng(seed)
 
-        # Advance the RNG state to the realization point
-        rng.standard_normal(realization)
-
-        # Generate a single sample
-        value = rng.standard_normal(1)
-        return np.array([value[0]])
+        # Generate samples for all active realizations
+        all_values = rng.standard_normal(num_realizations)
```
achaikou, line 159: Without the context I am quite confused about the need for `num_realizations` parameter.

From the first look it seems like `return rng.standard_normal(len(active_realizations))` should be enough.
But I suspect we want to assure realizations subsets stability as a required feature of this function?

If the point is to retrieve data in chunks in the same run

> Request 5 samples, then request new 5 samples, then other 5 samples...

stabilizing with `rng.standard_normal(num_realizations)` is a must. I see that this exact point is tested in the `test_that_sub_sample_maintains_order`

However in the code itself function is used just once and everything is always sent in one bunch.

So my best guess is that on cluster subsets of realizations can be whatever and thus subset-stability is a required feature to assure reruns on clusters are reproducible.

Is it the truth?
If so, would it make sense to document the reason for the behavior in the function itself? I feel like the reasoning for calling `standard_normal` has changed sine https://github.com/equinor/ert/commit/cad295e6d0cf7ebdea624ba65589cf43b6d2e5cf

src/ert/storage/local_ensemble.py
```diff
@@ -737,20 +737,21 @@ def load_observation_scaling_factors(
     @staticmethod
     def sample_parameter(
```
achaikou, line 738: I don't quite understand the naming thing here yet, but are we now sampling still one parameter (a list), or are those now several parameters?

src/ert/sample_prior.py
```diff
@@ -65,21 +66,18 @@ def sample_prior(
...
-                ]
-                if datasets:
-                    dataset = pl.concat(datasets, how="vertical")
+                dataset = Ensemble.sample_parameter(
+                    config_node,
+                    list(active_realizations),
+                    random_seed=random_seed,
+                    num_realizations=num_realizations,
+                )
+            if not (dataset is None or dataset.is_empty()):
```
achaikou: Are we allowed empty line before if or does our style checker forbid that?
As those are different ifs right after the other, it would make sense to separate :innocent:

src/ert/config/parameter_config.py
lucaeg: Yes, thank you, it should be plural! The sampling is still done per parameter, however.

lucaeg: I see that the rest of the docstring needs to be updated too :) I already updated the doctoring in the related PR about stratified sampling, however, I failed to migrate these changes here. Thanks for noticing this :)

src/ert/config/parameter_config.py
lucaeg: You are right, I will drop idx.

src/ert/storage/local_ensemble.py
lucaeg: We are sampling one parameter at the time. However, we are now sampling the full vector of all (active) realizations at the same time, instead of only one value at the time, as it used to be.

src/ert/config/parameter_config.py
lucaeg: Yes your observation is kind of correct I think, but more specifically: We want the sampling of each parameter to be reproducible, we therefore seed on parameter level, and to ensure that all samples are the same (regardless of which realizations are active), we need to stratify the parameter axis with respect to all realizations (hence the num_realizations parameter). I agree that this should be clear in the doc string, and I will make sure to add it there :)

lucaeg: Regarding batch sampling, and stability: I think it should all be fine with the current set up, but I will double check, this is not something I had in mind when I refactored.

src/ert/config/parameter_config.py
```diff
@@ -130,29 +130,25 @@ def sample_values(
...
         - global_seed (str): A global seed string used for RNG seed generation to ensure
         reproducibility across runs.
-        - realization (int): An integer used to advance the RNG to a specific point in
-        its sequence, effectively selecting the 'realization'-th sample from the
-        distribution.
+        - active_realizations (list[int]): indices of the realizations
+        to select from the stratified sampling vector; each must satisfy
+        0 <= i < num_realizations.
+        - num_realizations (int): Total number of realizations to generate the
+        reproducible sampling.
```
achaikou: Got somewhat confused by the wording (thinking it was "number of realizations that would be generated").

Taking words from your other comment, maybe something like

> Total number of realizations. Assures stable sampling for a given global_seed regardless of currently active realizations.

?

src/ert/config/parameter_config.py
```diff
@@ -130,29 +130,25 @@ def sample_values(
...
+        indices specified by `active_realizations` are then returned.
 
         Parameters:
         - global_seed (str): A global seed string used for RNG seed generation to ensure
         reproducibility across runs.
-        - realization (int): An integer used to advance the RNG to a specific point in
-        its sequence, effectively selecting the 'realization'-th sample from the
-        distribution.
+        - active_realizations (list[int]): indices of the realizations
+        to select from the stratified sampling vector; each must satisfy
```
achaikou: Would vector now always be stratified? From the issue I understood stratifying strategy comes in addition?

src/ert/config/parameter_config.py
lucaeg: Thanks, I have tried to make it more clear :) Also, all the commits will be squashed yes!

src/ert/config/parameter_config.py
lucaeg: Yes sorry, this is a mistake (stemming from the fact that I rewrote the doctoring of the stratified sampling function form the other PR!). Fixed now :)

Issue #12527, opened 2025-12-19
Bug in histogram plotter


frode-aarstad:
### What happened? (You can include a screenshot if it helps explain)

Got the following in the terminal:

--------------------------------------------------------------------------------
An error occurred during plotting. This stack trace is helpful for diagnosing the problem.--------------------------------------------------------------------------------
  File "/Users/FAAR/code/ert/src/ert/gui/tools/plot/plot_widget.py", line 201, in updatePlot
    self._plotter.plot(
  File "/Users/FAAR/code/ert/src/ert/gui/tools/plot/plottery/plots/histogram.py", line 36, in plot
    plotHistogram(figure, plot_context, ensemble_to_data_map)
  File "/Users/FAAR/code/ert/src/ert/gui/tools/plot/plottery/plots/histogram.py", line 129, in plotHistogram
    _plotHistogram(
  File "/Users/FAAR/code/ert/src/ert/gui/tools/plot/plottery/plots/histogram.py", line 197, in _plotHistogram
    bins = _histogramLogBins(effective_bin_count, minimum, maximum)  # type: ignore
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/FAAR/code/ert/src/ert/gui/tools/plot/plottery/plots/histogram.py", line 219, in _histogramLogBins
    minimum = log10(float(minimum))
              ^^^^^^^^^^^^^^^^^^^^^
Exception type: ValueError
math domain error

### What did you expect to happen?

Nothing in the terminal

### Steps to reproduce

```console
Run snake_oil esmda
In the plotter select ensemble_3 and SNAKE_OIL_PARAM:OP1_OFFSET
Press the logscale button in the histogram plot
```

### Environment where bug has been observed

- [ ] Python 3.11
- [ ] Python 3.12
- [x] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

xjules:
Yes, this is definitely not achievable for negative values. Not sure how to treat those, cc. @oyvindeide

N9sha18:
@jonathan-eq I've updated the pr I implemented a check in plot_widget.py that scans both the ensemble data and observations. If any values are <= 0, the Log Scale checkbox is automatically unchecked and disabled to prevent the ValueError

[...]

PR #12657, opened 2026-01-19
Fix math domain error in histogramLogBins (Issue #12527)

N9sha18:
Fixed a ValueError when plotting log-scale histograms with non-positive data. Added a check to clamp minimum values

Relates to #12527

### DIFF
--- a/src/ert/gui/tools/plot/plot_widget.py
+++ b/src/ert/gui/tools/plot/plot_widget.py
@@ -193,7 +193,7 @@ class PlotWidget(QWidget):
         logger.info(f"Plotwidget utility used: 'Log scale button' in tab '{self.name}'")
         self._log_checkbox.clicked.disconnect()  # Log only once
 
-    def updatePlot(
+   def updatePlot(
         self,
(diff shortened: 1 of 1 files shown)


src/ert/gui/tools/plot/plottery/plots/histogram.py
```diff
@@ -215,23 +215,19 @@ def _plotHistogram(
...
         else:
-            log_bin_count = bin_count
+            # If both min and max are <= 0, we cannot return log bins.
+            # Returning None or raising a clearer error is safer than crashing.
+            raise ValueError("Cannot compute log bins: all data values are <= 0")
+    # FIX END
 
-    return 10 ** np.linspace(minimum, maximum, log_bin_count)
+    minimum = log10(float(minimum))
+    maximum = log10(float(maximum))
```
jonathan-eq, line 233: You still have to return something

jonathan-eq: @N9sha18 Can you verify that the tests still pass locally?

PR #12663, opened 2026-01-19
Implement parsing of Localization objects

### LINKED ISSUE #12664: Create parsing for LOCALIZATION object
The format should be: 
```
...
LOCALIZATION {
    EAST = 10;
    NORTH = 10;
    RADIUS = 10;
};
```

SAKavli:
Resolves #12664

### DIFF
--- a/tests/ert/unit_tests/config/test_summary_config.py
+++ b/tests/ert/unit_tests/config/test_summary_config.py
@@ -1,4 +1,5 @@
 import os
+import re
 from contextlib import suppress
 from pathlib import Path
 from textwrap import dedent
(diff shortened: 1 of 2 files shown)


tests/ert/unit_tests/config/test_summary_config.py
```diff
@@ -142,26 +145,18 @@ def test_that_summary_observations_without_location_range_gets_defaulted(
...
-        "Localization for observation FOPR_1 is misconfigured.",
-        f"Only {', '.join(matches)} were provided.",
-        "ensure that both LOCATION_X and LOCATION_Y are defined",
-    ]
     with pytest.raises(ConfigValidationError) as e, tmpdir.as_cwd():
-        create_summary_observation(loc_config_lines)
-    for msg in expected_err_msgs:
-        assert msg in str(e.value)
+        create_summary_observation("LOCALIZATION {" + loc_config_lines + "};")
+    assert "Unexpected type in observations" in str(e.value)
```
eivindjahren: I think it would be a good idea to give a better error message when this happens.

PR #12667, opened 2026-01-19
Fix test failing for environments where floating point numbers are localized

eivindjahren:
This avoids an issue with resdata.geometry.Surface where floating point formatting
is erronously localized in irap files. This only applies to one test in ert so it is enough
to work around it by using xtgeo instead.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/storage/test_parameter_sample_types.py
+++ b/tests/ert/unit_tests/storage/test_parameter_sample_types.py
@@ -86,13 +86,6 @@ def test_surface_param(
 ):
     values = [[0.0, 1.0], [2.0, 3.0]]
     with tmpdir.as_cwd():
-        config = dedent(
-            """
-        JOBNAME my_name%d
-        NUM_REALIZATIONS 1
-        """
-        )
-        config += config_str
         expect_surface = RegularSurface(
             ncol=2,
             nrow=2,
@@ -107,7 +100,7 @@ def test_surface_param(
         expect_surface.to_file("surf0.irap", fformat="irap_ascii")
 
         with open("config.ert", mode="w", encoding="utf-8") as fh:
-            fh.writelines(config)
+            fh.writelines(f"NUM_REALIZATIONS 1\n{config_str}\n")
         ensemble_config, fs = create_runpath(storage, "config.ert")
         assert ensemble_config["MY_PARAM"].forward_init is expect_forward_init
         # We try to load the parameters from the forward model, this would fail if


tests/ert/unit_tests/storage/test_parameter_sample_types.py
```diff
@@ -92,13 +93,18 @@ def test_surface_param(
         """
         )
         config += config_str
-        expect_surface = Surface(
-            nx=2, ny=2, xinc=1, yinc=1, xstart=1, ystart=1, angle=0
+        expect_surface = RegularSurface(
```
achaikou, line 89: Would appreciate some grand knowledge :relaxed: 

So there are surfaces in 3 libraries:
- `resdata.Geometry.Surface` (used before for writing)
- `xtgeo.RegularSurface` (used now for writing)
- `surfio.IrapSurface` (used now for reading surface data in create_runpath - ErtConfig - EnsembleConfig - SurfaceConfig)

What is the difference between those libraries and when should each be used?
What is the reason for read-write mismatch?

tests/ert/unit_tests/storage/test_parameter_sample_types.py
```diff
@@ -126,15 +132,17 @@ def test_surface_param(
...
-            actual_surface = Surface(
-                f"simulations/realization-0/iter-{expected_iter}/surf.irap"
+            actual_surface = surface_from_file(
+                f"simulations/realization-0/iter-{expected_iter}/surf.irap",
+                fformat="irap_ascii",
             )
-            assert actual_surface == expect_surface
+            assert actual_surface.compare_topology(expect_surface)
+            actual_surface -= expect_surface
+            assert (actual_surface <= 1e-5).all()
```
achaikou: Looks somewhat weird to me.
Is this an established ert-way to compare floating points?

achaikou: This looks somewhat weird to me.

Is this an established ert-way to compare floating points regardless of values even when we deal with two surfaces with well defined integer values?

I would have expected something more straighforward, like
`assert np.array_equal(actual_surface.values, expect_surface.values)`

tests/ert/unit_tests/storage/test_parameter_sample_types.py
```diff
@@ -92,13 +93,18 @@ def test_surface_param(
...
-            nx=2, ny=2, xinc=1, yinc=1, xstart=1, ystart=1, angle=0
+        expect_surface = RegularSurface(
+            ncol=2,
+            nrow=2,
+            xinc=1.0,
+            yinc=1.0,
+            xori=1.0,
+            yori=1,
+            rotation=0.0,
+            values=values,
```
achaikou, line 97: Intellisense/pylance or something keeps screaming at me that there is a type mismatch (`values` are list of lists, when xtgeo requires just a list)

Another place in the test it also complains about who-knows-what.

So my question is: how strict should we be to such warnings? I see those pop here and there.
Given all that "mypy" type checking I thought we cared?

tests/ert/unit_tests/storage/test_parameter_sample_types.py
eivindjahren: The two first versions were just developed by two people independently. The last one we are trying to replace the two others with behind the scenes but we haven't finished that yet.

tests/ert/unit_tests/storage/test_parameter_sample_types.py
eivindjahren: For an approximate comparison I would prefer `actual_surface.values == pytest.approx(values)`, but I was trying to just use the api of xtgeo which didn't allow that. I was not sure whether we could use exact comparison because of the roundtrip with writing ascii, but I will have a look whether we can do that.

tests/ert/unit_tests/storage/test_parameter_sample_types.py
eivindjahren: Yes, this is unfortunately a bug in the type hint of xtgeo and the documentation says this is supported.

tests/ert/unit_tests/storage/test_parameter_sample_types.py
eivindjahren: I changed it to an exact comparison

PR #12675, opened 2026-01-20
Move some validation errors to _observations.py

yngve-sk:
Prep for attaching observation declarations to runmodel config, but still getting most of the validations in the ERT config parsing step (even though `_create_observation_dataframes` is no longer to be called in the ERT config parsing step)

### DIFF
--- a/src/ert/config/_observations.py
+++ b/src/ert/config/_observations.py
@@ -228,6 +228,12 @@ class GeneralObservation(_GeneralObservation):
                 observation_dict["name"],
             )
 
+        if output.value is None and output.error is None and output.obs_file is None:
+            raise ObservationConfigError.with_context(
+                "GENERAL_OBSERVATION must contain either VALUE and ERROR or OBS_FILE",
+                context=observation_dict["name"],
+            )
+
         # Bypass pydantic discarding context
         # only relevant for ERT config surfacing validation errors
         # irrelevant for runmodels etc.
--- a/src/ert/config/_create_observation_dataframes.py
+++ b/src/ert/config/_create_observation_dataframes.py
@@ -144,16 +144,6 @@ def _handle_general_observation(
     restart = 0 if restart is None else restart
 
     if (
-        general_observation.value is None
-        and general_observation.error is None
-        and general_observation.obs_file is None
-    ):
-        raise ObservationConfigError.with_context(
-            "GENERAL_OBSERVATION must contain either VALUE and ERROR or OBS_FILE",
-            context=obs_key,
-        )
-
-    if (
         general_observation.value is not None
         and general_observation.error is not None
         and general_observation.obs_file is not None


src/ert/config/_observations.py
```diff
@@ -99,24 +97,28 @@ def from_obs_dict(
...
+            value=float_values["VALUE"],
+            location_x=localization_values.get("x"),
+            location_y=localization_values.get("y"),
+            location_range=localization_values.get("range"),
+            date=date,
+        )
+        # Bypass pydantic discarding context
+        # only relevant for ERT config surfacing validation errors
+        # irrelevant for runmodels etc.
+        instance.name = observation_dict["name"]
```
berland, line 156: This seems a little bit smelly. Will take a second look on this.

src/ert/config/_observations.py
```diff
@@ -107,7 +132,7 @@ def from_obs_dict(
             location_x=localization_values.get("x"),
             location_y=localization_values.get("y"),
             location_range=localization_values.get("range"),
-            date=date,
+            date=standardized_date,
```
berland, line 151: You can do `parsed_date.date().isoformat()` instead, always better to avoid `strftime`.

src/ert/config/_observations.py
```diff
@@ -97,6 +119,9 @@ def from_obs_dict(
             raise _missing_value_error(observation_dict["name"], "ERROR")
 
         assert date is not None
+        # Raise errors if the date is off
+        parsed_date = _parse_date(date)
```
berland: makes sense to type this assignment for readability:
`parsed_date: datetime = _parse_date(date)`

src/ert/config/_observations.py
yngve-sk: 👍 It is smelly, but I believe it must be there to keep the FileContext stuff attached to the name (which is used in ERT config to pinpoint where errors/warnings are from). Piping it through `dataclasses.dataclass` preserved the str as a ContextString, but pydantic will drop the context, so it must be forcefully applied afterwards.

src/ert/config/_observations.py
```diff
@@ -135,6 +135,11 @@ def from_obs_dict(
                 case default:
                     assert_never(default)
 
+        if error <= 0:
```
berland, line 138: This construct is somewhat misleading, as we have already done `_validate_positive_float`, so negative values cannot occur here (but this is not important for this change which is purely a move)

tests/ert/unit_tests/config/test_observation_declaration.py
```diff
@@ -150,7 +150,7 @@ def test_make_observations():
     assert isinstance(wopr, SummaryObservation)
     assert wopr.value == 0.1
     assert wopr.key == "WOPR:OP1"
-    assert wopr.date == "2010-03-31"
+    assert wopr.date == "2010-03-31T00:00:00"
```
berland: If you include `date()` in `parsed_date.date().isoformat()` you avoid this. But I suppose we do support datetimes also here. 🤔

yngve-sk: Set it to `parsed_date.date().isoformat()`. Suggest leaving up to future to make more specific decisions on this. I think all we need is for the string on the obs declaration to be a date we can parse on the other side.

src/ert/config/_observations.py
yngve-sk: Edit: Leaving this for future work / separate issue, ref https://github.com/equinor/ert/issues/12693

berland (approved):
Lets go. Remember to squash the fixup.

src/ert/config/_observations.py
```diff
@@ -99,24 +115,51 @@ def from_obs_dict(
...
+        if error_mode is not None:
+            match error_mode:
+                case ErrorModes.ABS:
+                    error = np.abs(input_error)
+                case ErrorModes.REL:
+                    error = np.abs(value) * input_error
+                case ErrorModes.RELMIN:
+                    error = np.maximum(np.abs(value) * input_error, error_min)
+                case default:
+                    assert_never(default)
```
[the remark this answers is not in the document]
berland: This one looks correct

src/ert/config/_observations.py
berland: PS: Dropping this does not fail any test.

PR #12715, opened 2026-01-23
Fix rounding logic in test_field_parameter

### LINKED ISSUE #12713: tests/ert/ui_tests/cli/test_field_parameter.py::test_field_param_update_using_heat_equation_enif_snapshot not working
If you run

`pytest tests/ert/ui_tests/cli/test_field_parameter.py::test_field_param_update_using_heat_equation_enif_snapshot --snapshot-update
`
it will fail with an error message that the snapshot has changed which should not be the case on `main`.

This has been like this since commit 9d1325f05c430d4ba430e4bdc8b2802dbf584ea5 (from git bisect)

This culprit commit exposed the error, as the test was faulty prior to that.

berland:
The previous code was called with outlier_threshold having a typical value of 0.01, which makes polars round to 100 decimal places. This causes almost certain failures when trying to update the snapshot

After the change, an outlier_threshold at 0.01 means that we round to 2 decimal place.

The snapshot is updated. This snapshot was never valid, as the bug fixed in 9d1325f05c430d4ba430e4bdc8b2802dbf584ea5 hid it.

**Issue**
Resolves #12713

**Approach**
🧠 

      --exec 'just rapid-tests'`)

## When applicable


tests/ert/ui_tests/cli/test_field_parameter.py
```diff
@@ -136,7 +137,7 @@ def _compare_ensemble_params(
 
     def round_and_cast_(df: pl.DataFrame) -> pl.DataFrame:
         return df.cast(_schema).with_columns(
-            pl.col(pl.Float64).round(int(1 / outlier_threshold))
+            pl.col(pl.Float64).round(int(-math.log10(outlier_threshold)))
```
yngve-sk, line 140: Nitpick: outlier threshold is always <1 so log10 always negative, but perhaps still do `abs` instead of `-`?

berland: Using `-` makes it valid without the assumption that `outlier_threshold < 1` so I think keeping `-` makes the code clearer.

PR #12719, opened 2026-01-23
Add snapshot_test marker for snapshot tests

yngve-sk:
Allows for running only snapshot tests, can consider later adding a justfile command for this


justfile
```diff
@@ -71,3 +71,15 @@ ert-tests:
 
 check-all:
     parallel -j8 ::: 'just ert-gui-tests' 'just ert-cli-tests' 'just ert-unit-tests' 'just ert-doc-tests' 'just everest-tests' 'just check-types' 'just build-everest-docs' 'just build-ert-docs'
+
+ert-snapshot-tests:
```
oyvindeide: Should be renamed to something like: `update-snapshots` to make it clear what it does

PR #12722, opened 2026-01-23
Fix EVEREST name to uppercase

roliveira:
**Issue**
After the open source of EVEREST it was agreed to use a trademark:

> EVEREST™ should have a trademark symbol the first time it appears in a publication.

but also to refer to it as an all caps, instead of the originally named EVEReST (Ensemble based Value Evaluation for Reservoir Simulation Tool), or other variations of the name.

Resolves #my_issue

**Approach**
Changed the ocurrances of `Everest` to `EVEREST` in docs, docstrings and a few comments. 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/everest/test_everest_client.py
+++ b/tests/everest/test_everest_client.py
@@ -244,18 +244,16 @@ def test_that_multiple_everest_clients_can_connect_to_server(
     assert all(first == other for other in client_event_lists[1:])
 
     everest_event_types = (EverestStatusEvent, EverestBatchResultEvent)
-    
+
     first_everevents = [
         e.event_type for e in first if isinstance(e, everest_event_types)
     ]
-    assert (len(first_everevents) > 0)
-    
+    assert len(first_everevents) > 0
+
     server_everevents = [
-        e.event_type
-        for e in server_events_list
-        if isinstance(e, everest_event_types)
+        e.event_type for e in server_events_list if isinstance(e, everest_event_types)
     ]
-    assert (len(server_everevents) > 0)
+    assert len(server_everevents) > 0
 
     # Compare only everest events, as the events from the forward model
     # are (at time of writing) not deterministic enough to expect equality


docs/everest/conf.py
```diff
@@ -173,15 +173,15 @@
 # (source start file, target name, title,
 #  author, documentclass [howto, manual, or own class]).
 latex_documents = [
-    (master_doc, "Everest.tex", "Everest Documentation", "Equinor \\& TNO", "manual"),
+    (master_doc, "EVEREST.tex", "EVEREST Documentation", "Equinor \\& TNO", "manual"),
```
andreas-el: Is the intent that we want uppercase names? Can be done, but this does not immediately sit right with me.

docs/everest/conf.py
```diff
@@ -173,15 +173,15 @@
...
 ]
 
 
 # -- Options for manual page output ------------------------------------------
 
 # One entry per manual page. List of tuples
 # (source start file, name, description, authors, manual section).
-man_pages = [(master_doc, "everest", "Everest Documentation", [author], 1)]
+man_pages = [(master_doc, "everest", "EVEREST Documentation", [author], 1)]
```
andreas-el, line 184: This somewhat contradicts the previous entry, as the filename now is lowercase?

docs/everest/conf.py
```diff
@@ -192,10 +192,10 @@
 texinfo_documents = [
     (
         master_doc,
-        "Everest",
-        "Everest Documentation",
+        "EVEREST",
```
andreas-el: And uppercase file again. I think we should try to pick one of these styles, and stick with that.

tests/everest/test_everest_client.py
```diff
@@ -242,9 +242,9 @@ def everserver_is_running():
     first = client_event_lists[0]
     assert all(first == other for other in client_event_lists[1:])
 
-    first_everevents = [e.event_type for e in first if "Everest" in e.event_type]
+    first_everevents = [e.event_type for e in first if "EVEREST" in e.event_type]
     server_everevents = [
-        e.event_type for e in server_events_list if "Everest" in e.event_type
```
andreas-el, line 247: Could you add an assertion to this test also? Verifying that the two lists are not empty?

src/ert/run_models/everest_run_model.py
```diff
@@ -588,7 +588,7 @@ def start_simulations_thread(
         error_messages: MutableSequence[str] = []
         traceback_str: str | None = None
         try:
-            logger.debug("Starting Everest simulations thread")
+            logger.debug("Starting EVEREST simulations thread")
```
andreas-el, line 617: We'd have to verify that any kusto queries handle case insensitive naming, otherwise the grafana dashboards could show wrong statistics after this.

andreas-el: After checking I found that the `contains` operator in `kusto` is case insensitive. 🎉

tests/everest/test_everest_client.py
roliveira: Great catch @andreas-el! This was actually wrong, but the test still passed (assert first_everevents == server_everevents --> 0 == 0)

I've added your suggestion to check if both event lists are not empty but also changed the test to check for the class instance itself instead of a string casting.

Issue #12538, opened 2025-12-21
Random seeds are reused between experiments in same Ert instance


SAKavli:
I added:
```py
print(random_seed)
```
to `sample_prior()` in `sample_prior.py`.

When I run multiple experiments using the same Ert instance, they have the same random seed.
I would expect the sampled values between experiments to be different, but they are the same.

This can be recreated by running a test-data example without hardcoded random seed twice.

https://github.com/user-attachments/assets/5fac95a0-f41b-4a03-9aab-e2351b208f14

I was curious if this is a bug or not, hence `needs-discussion` label.

oyvindeide:
Agree that this needs to be fixed, we just need to make sure that:
1.  if the seed has been set by the user, we always sample the same, regardless of how many times they run the experiment
2. we write the seed out to the user (in the log file), so that they can reproduce the experiment

PR #12723, opened 2026-01-23
Fix random seed ensemble experiments

ajaust:
**Issue**
Resolves https://github.com/equinor/ert/issues/12538

**Approach**
Makes the `random_seed` an computed property. The `ErtConfig` object is able to track if a random seed is user-defined or not and "advance" a seed accordingly. The random seed must be advance before an ensemble experiment is created to ensure that the random seed value is correct.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/ui_tests/cli/test_setting_random_seeds.py
+++ b/tests/ert/ui_tests/cli/test_setting_random_seeds.py
@@ -1,0 +1,70 @@
+import logging
+from pathlib import Path
+from textwrap import dedent
+
+import pytest
+
+from ert.mode_definitions import (
+    ENSEMBLE_EXPERIMENT_MODE,
+    TEST_RUN_MODE,
+)
+
+from .run_cli import run_cli
+
+
+@pytest.mark.parametrize("experiment_type", [TEST_RUN_MODE, ENSEMBLE_EXPERIMENT_MODE])
+def test_that_cli_uses_config_random_seed_when_specified(
+    caplog, use_tmpdir, experiment_type
+):
+    config_text = dedent(
+        """
+        NUM_REALIZATIONS 1
+        RANDOM_SEED 12345
+        """
+    )
+    Path("config.ert").write_text(config_text, encoding="utf-8")
+
+    with caplog.at_level(logging.INFO):
+        run_cli(
+            experiment_type,
+            "--disable-monitoring",
+            "config.ert",
+        )
+
+    seed_logs = [line for line in caplog.text.splitlines() if "'random_seed':" in line]
+    assert len(seed_logs) == 1
+    assert "'random_seed': 12345" in seed_logs[0]
+
+
+@pytest.mark.parametrize("experiment_type", [TEST_RUN_MODE, ENSEMBLE_EXPERIMENT_MODE])
+def test_that_cli_generates_different_seeds_for_consecutive_runs(
+    caplog, use_tmpdir, experiment_type
+):
+    config_text = dedent(
+        """
+        NUM_REALIZATIONS 1
+        """
+    )
+    Path("config.ert").write_text(config_text, encoding="utf-8")
+
+    with caplog.at_level(logging.INFO):
+        run_cli(
+            experiment_type,
+            "--disable-monitoring",
+            "config.ert",
+        )
+        seed_logs = [line for line in caplog.text.splitlines() if "RANDOM_SEED" in line]
+        first_seed_from_log = seed_logs[-1]
+
+        run_cli(
+            experiment_type,
+            "--disable-monitoring",
+            "config.ert",
+        )
+        seed_logs = [line for line in caplog.text.splitlines() if "RANDOM_SEED" in line]
+        second_seed_from_log = seed_logs[-1]
+
+    assert first_seed_from_log != second_seed_from_log
+
+    seed_logs = [line for line in caplog.text.splitlines() if "'random_seed':" in line]
+    assert len(seed_logs) == 2
--- a/tests/ert/ui_tests/gui/test_setting_random_seeds.py
+++ b/tests/ert/ui_tests/gui/test_setting_random_seeds.py
@@ -1,0 +1,66 @@
+import logging
+import shutil
+from pathlib import Path
+from textwrap import dedent
+
+import pytest
+
+from ert.run_models import EnsembleExperiment, SingleTestRun
+
+from .conftest import _open_main_window
+
+
+@pytest.mark.parametrize("experiment_type", [SingleTestRun, EnsembleExperiment])
+def test_that_gui_uses_config_random_seed_when_specified(
+    run_experiment, use_tmpdir, qtbot, caplog, experiment_type
+):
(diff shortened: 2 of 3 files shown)


src/ert/config/ert_config.py
```diff
@@ -1134,6 +1148,12 @@ def from_dict(cls, config_dict: ConfigDict) -> Self:
                     ensemble_config.response_configs.get("rft", None),
                 ),
             )
+
+            cls_config._user_defined_random_seed = config_dict.get(
+                ConfigKeys.RANDOM_SEED
+            )
+            if cls_config._user_defined_random_seed is not None:
+                cls_config.random_seed = cls_config._user_defined_random_seed
```
SAKavli: To me it looks like `random_seed` and `_user_defined_random_seed` are quite correlated, do we need both?

ajaust: Yes, they are indeed. 

I think we need two variables to solve this:

- One variable that saves the random seed as methods may query the configuration for the currently used random seed.
- Some variable that tells us whether advancing the random seed will update the seed or if the seed was user defined and thus will not be updated.

From my understanding, the question is whether we want to store two random seed variables or if it would be better to just keep a single variable for the random seed and another variable (Bool?) that states whether the random seed is user defined. I have no strong opinion on this. I went with the current solution because it looked natural to me to reuse `_seed_sequence`. 

What is your feeling/opinion about this?

ajaust: Here is some idea how it could look using `bool`: https://github.com/equinor/ert/commit/d5ce91db7fa6f9222e462c9e12fd7fa4f5b604c0

Some tests are failing so this would not be the final solution yet.

SAKavli: Ooo, I like the bool version, then it's less confusing to me

src/ert/config/ert_config.py
```diff
@@ -742,6 +743,20 @@ def observations(self) -> dict[str, pl.DataFrame]:
             )
         return self._observations
 
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def random_seed(self) -> int:
+        return self._random_seed
+
+    @random_seed.setter
+    def random_seed(self, value: int) -> None:
```
eivindjahren: This has become rather hard to read for me now. I think it has just gone through too many iterations of additions. I think we want this:

```python
@dataclass
class RandomSeed:
    user_defined_seed: int | None

    def advance(self) -> int:
        if self.user_defined_seed is not None:
            return self.user_defined_seed

        int_seed = SeedSequence().entropy
        logger.info(
            "To repeat this experiment, "
            "add the following random seed to your config file:\n"
            f"RANDOM_SEED {int_seed}"
        )
        assert isinstance(int_seed, int)
        return int_seed
```

ajaust: Do you consider the `_seed_sequence` the main issue or the definition of `random_seed` as a property? The comment appears in the `random_seed` definition, but the suggested code seems to concern `_seed_sequence`.

Do you think something like this would be better: https://github.com/equinor/ert/compare/main...ajaust:ert:fix/random_seed_ensemble_experiments2?diff=split&w#diff-51c680229ac06ac420b90d8ae0e7567679b186f2a1e470a490e42a0cd98b0911

This would replace the `_seed_sequence` with the dataclass that you suggest which would simplify some of the code. 

If `ert_config.random_seed = VALUE` should imply a user-defined seed, then I don't see any good solution without the property and computed fields at the moment. The pattern `ert_config.random_seed = VALUE` is used in some tests at least.

eivindjahren: Oh, I was thinking we would just do 
```
random_seed_generator: RandomSeedGenerator = Field(default_factory=RandomSeedGenerator )
```

Then we fix up all the call sites so that instead of

```
seed = config.random_seed
```
you would do

```
seed = config.random_seed_generator.advance()
```
and instead of
ErtConfig(
  ...,
  random_seed=seed
)
we do
```
ErtConfig(
  ...,
  random_seed_generator=RandomSeedGenerator(seed)
)
```

If the random seed is ever set we would replace it with a call to

```python
def set_random_seed(self, seed: int) -> None:
   self.random_seed_generator.seed = seed
```

I would just rather avoid `a = config.random_seed` meaning "advance random seed" rather than getting some static value.

ajaust: Ah, I see. I agree that something like this is probably nicer. I cooked something along these lines (without test) here: https://github.com/equinor/ert/compare/58f2ed394de67b03d5d02100861c812ac5cd89be~..795e9d7cc707ee2b2d0bf928beb13601640a6a15

I also added a comment to the main issue for the discussion which part of the code should be responsible to update/advance the random seed.

src/ert/config/ert_config.py
```diff
@@ -695,7 +702,7 @@ class ErtConfig(BaseModel):
     ensemble_config: EnsembleConfig = Field(default_factory=EnsembleConfig)
     ens_path: str = DEFAULT_ENSPATH
     env_vars: dict[str, str] = Field(default_factory=dict)
-    random_seed: int = Field(default_factory=lambda: _seed_sequence(None))
+    random_seed: int = 0
```
eivindjahren: I think you can replace this with

```python

@property
def random_seed(self):
    return self.random_seed_generator.seed
```

ajaust: I added the `random_seed` as property as a fixup commit in order to check if this would pass the full test pipeline. 🙂

ajaust: Tests pass. ✅ 

Iadded a property setter for the random seed to avoid updating all the tests that set the random seed to a fixed value while evaluating the approach using `@property`. We can also drop this setter and simply update the tests. Do you have any opinions on this? I am happy with using the approach using a property as long as we are convinced that serialisation is not an issue and that people are not confused that `ErtConfig.random_seed` may return a different value every time they call the property if there is no user-defined random seed set.

src/ert/config/ert_config.py
```diff
@@ -94,10 +94,12 @@ class RandomSeedGenerator:
 
     @property
     def seed(self) -> int:
-        if self.user_defined_seed is not None:
-            return self.user_defined_seed
+        int_seed = (
```
SAKavli: I am a bit confused, does this change correspond to the commit message? It's of no importance if this is to be squashed either way.

ajaust: Good catch! 

I think the commit made more sense in an earlier version of this PR. At some point, ERT was always creating a random seed on startup, even when there was a user-specified seed. In that case, it looked useful to indicate that the user-specified seed overwrites he initially created (and logged) seed. 

In the current version of the PR, we don't create this initial random seed anymore. Therefore, my suggestion would be to drop this commit from the PR as it is not needed anymore. What do you think?

SAKavli: I know what you mean! 😄 After a handful of rebases theres always some stray code changes laying around on my branches.

PR #12737, opened 2026-01-27
Add validation of zone for rft observations

eivindjahren:
**Issue**
Resolves #12700 

Currently shows you a wall of warnings when there are a lot of mismatches:

image

Trying to combine warnings across realizations and iterations is somewhat difficult so this might have to do for a first iteration.

- [x] PR title captures the intent of the changes, and is fitting for release notes.
- [x] Added appropriate release note label
- [x] Commit history is consistent and clean, in line with the [contribution guidelines](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md).
- [x] Make sure unit tests pass locally after every commit (`git rebase -i main
      --exec 'just rapid-tests'`)

## When applicable
- [ ] **When there are user facing changes**: Updated documentation
- [ ] **New behavior or changes to existing untested code**: Ensured that unit tests are added (See [Ground Rules](https://github.com/equinor/ert/blob/main/CONTRIBUTING.md#ground-rules)).
- [ ] **Large PR**: Prepare changes in small commits for more convenient review
- [ ] **Bug fix**: Add regression test for the bug
- [ ] **Bug fix**: Add backport label to latest release (format: 'backport release-branch-name')

PR #12692, opened 2026-01-21
Create observation dataframes on runmodel initialization

yngve-sk:
Resolves https://github.com/equinor/ert/issues/12708

Followup issues should be stop using ErtConfig.observations, main usages are here. Considering resolving some of these in this PR but the rest should be left as (immediate to be picked up by others) future work:
* Stop using ErtConfig.observations in test (could likely do one PR per file if it is too much to go over all at once)
...

PR #12752, opened 2026-01-29
Add obs dataframe generation back to ErtConfig

yngve-sk:
fix failing tests after silent merge conflict https://github.com/equinor/ert/pull/12737 + https://github.com/equinor/ert/pull/12692

### DIFF
--- a/src/ert/config/ert_config.py
+++ b/src/ert/config/ert_config.py
@@ -1167,6 +1167,15 @@ class ErtConfig(BaseModel):
                     zonemap=cls_config.zonemap,
                 )
 
+            # PS:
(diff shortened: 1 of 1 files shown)


src/ert/config/ert_config.py
```diff
@@ -1167,6 +1167,15 @@ def from_dict(cls, config_dict: ConfigDict) -> Self:
                     zonemap=cls_config.zonemap,
                 )
 
+            # PS:
+            # This mutates the rft config and is necessary for the moment
```
xjules, line 1171: Put a comment that this should be refactored. It is quite ugly.

Issue #12651, opened 2026-01-16
`unused_tcp_port` has race condition


eivindjahren:
The `unused_tcp_port` fixture from `pytest-asyncio` has a race condition where the socket might get bound by a different process between the time it was found and the test has a chance to bind it. This is an inherent flaw in its design and we probably need to automatically add `pytest.mark.flaky(rerun=3)` to all tests that use it.

This was observed to be very flaky on at least one developer machine when it was very busy with some other network task.

PR #12754, opened 2026-01-29
Mark tests using TCP ports as flaky

ajaust:
The unused_tcp_port fixture from pytest-asyncio has a race condition where the socket might get bound by a different process between the time it was found and the test has a chance to bind it.

**Issue**
Resolves https://github.com/equinor/ert/issues/12651

**Approach**
Marking the tests as flaky was suggested in https://github.com/equinor/ert/issues/12651.

The marker `pytest.mark.rerun(3)` is added to all tests requesting a TCP port via the `unused_tcp_port` fixture by configuring the marker in `conftest.py`. This way, we do not need to manually mark the tests as flaky.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -42,6 +42,8 @@ def pytest_collection_modifyitems(config, items):
         fixtures = getattr(item, "fixturenames", ())
         if "qtbot" in fixtures or "qapp" in fixtures or "qtmodeltester" in fixtures:
             item.add_marker("requires_window_manager")
+        if "unused_tcp_port" in fixtures:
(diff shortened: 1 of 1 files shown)


ajaust:
Is there any appropriate label to add to this PR? It does not fix a the flakiness, I don't think that is something to backport anywhere and it isn't user facing (so I guess it's not relevant for the release notes?!).

ajaust:
It was discussed that the `flaky-test-fix` label is most appropriate. The source of flakiness is hard (impossible?) to fix so marking the test as flaky seems the best solution.

PR #12761, opened 2026-02-02
Refactor observation tests to stop using ErtConfig.observations

### LINKED ISSUE #12751: Stop using ErtConfig.observations
Residual issue from https://github.com/equinor/ert/pull/12692 , it is no longer used outside of tests. Instead, we should use ErtConfig.observation_declarations. If we do need to do assertions on the polars dataframes, we should go through runmodel -> storage.

eqbech:
**Issue**
Resolves #12751 

**Approach**
Refactored a couple of tests to utilize `make_observation` and `create_observation_dataframe` instead of the `ErtConfig.observations` property. The helper function `make_observation` inside `test_observations.py` has been renamed to `ert_config_from_parser` and changed to now only parse obs config via `parse_observations`.

### DIFF
--- a/tests/ert/unit_tests/config/test_observations.py
+++ b/tests/ert/unit_tests/config/test_observations.py
@@ -3,6 +3,7 @@ from contextlib import ExitStack as does_not_raise
 from datetime import datetime, timedelta
 from pathlib import Path
 from textwrap import dedent
+from typing import cast
 
 import hypothesis.strategies as st
 import polars as pl
@@ -18,17 +19,20 @@ from ert.config import (
     ConfigWarning,
     ErtConfig,
 )
+from ert.config._create_observation_dataframes import create_observation_dataframes
+from ert.config._observations import make_observations
 from ert.config.parsing import parse_observations
 from ert.config.parsing.observations_parser import (
     ObservationConfigError,
     ObservationType,
 )
+from ert.config.rft_config import RFTConfig
 from ert.namespace import Namespace
 
 pytestmark = pytest.mark.filterwarnings("ignore:Config contains a SUMMARY key")
 
 
-def make_observations(obs_config_contents, parse=True):
+def ert_config_from_parser(obs_config_contents):
     obs_config_file = "obs_config"
     return ErtConfig.from_dict(
         {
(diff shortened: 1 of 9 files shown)


tests/ert/unit_tests/config/test_observations.py
```diff
@@ -144,24 +144,40 @@ def test_that_when_history_source_is_simulated_the_summary_vector_is_used():
...
+                [
+                    (
+                        {
+                            "type": ObservationType.SUMMARY,
+                            "name": "FOPR",
+                            "KEY": "FOPR",
+                            "VALUE": "1",
+                            "ERROR": "1",
+                            "DATE": datestring,
+                        }
```
andreas-el: This dict, the tuple or even the whole list can be extracted and rather referenced instead of being duplicated here.

tests/ert/unit_tests/config/test_observations.py
```diff
@@ -521,6 +542,7 @@ def test_that_error_must_be_greater_than_zero_in_summary_observations(std):
         ConfigValidationError, match=r"must be given a strictly positive value"
     ):
         make_observations(
+            "",
```
andreas-el, line 534: Could you look into if this argument could be defaulted instead?
Meaning if you do not provide this, the default value should be `""`.
There is only one instance outside this file using that function.

andreas-el: You'd have to change the order, or allow defaults for both arguments perhaps.
At least something you could look into.

tests/ert/unit_tests/config/test_observations.py
eqbech: Good catch! Extracted out the list.

tests/ert/unit_tests/config/test_observations.py
eqbech: Could swap the order and default the `directory` argument. The directory argument seems to only be used for the `from_obs_dict` method implemented by the three `ObservationType` classes we have, where `SummaryObservation` does not use it at all, but the others do. Does it make sense in this context to default a directy arg to an empty string or should it be nullable and add additional checks?

eqbech: After consulting with Yngve, it maybe does not make sense to make changes to `make_observation` because of how it is used in the tests? The `directory` parameter will always be a real value when used elsewhere. If we do decide to make changes to the method, maybe it should be a separate issue? What do you think?

andreas-el: If you could write an issue for reducing the necessity of providing all these `""` paths in the tests, -- that would be nice.
Agreed that this is a bit to the side of this PR. 👍

tests/ert/unit_tests/config/test_observations.py
```diff
@@ -28,7 +30,7 @@
 pytestmark = pytest.mark.filterwarnings("ignore:Config contains a SUMMARY key")
 
 
-def make_observations(obs_config_contents, parse=True):
+def ert_config_from_parser(obs_config_contents):
```
yngve-sk, line 35: Nitpick: suggest rename `ert_config_with_obs_config_contents`

PR #12780, opened 2026-02-03
Add hook for plugins to provide `get_ip_address()`

### LINKED ISSUE #12782: PRIORITIZE_PRIVATE_IP_ADDRESS does not work

A user sent us a message explaining ert was not communicating with the jobs on the cluster. After examining the logs and jobs.json, it turns out Ert picked a private ip address even when it was prioritize_private_ip_address=False. 0e1fa0523069f37f3289a8dcfec7736b5de3c434 needs to be reverted, and a better fix is needed, maybe passing a get_ip_address function from site_config and defaulting to the old function.

_No response_

### Steps to reproduce

```console

```

- [ ] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

jonathan-eq:
This commit fixes the PRIORITIZE_PRIVATE_IP_ADDRESS site config setting by using the old implementation when it is False, and the one introduced in 0e1fa0523069f37f3289a8dcfec7736b5de3c434 if it is set to True.

**Issue**
Resolves #12782 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/plugins/hook_specifications/net_utils.py
+++ b/src/ert/plugins/hook_specifications/net_utils.py
@@ -1,0 +1,13 @@
+from ert.plugins.plugin_manager import hook_specification
+
+
+@hook_specification
+def get_ip_address() -> str:  # type: ignore
+    """Ert uses network communication over TCP/IP and needs to provide potential
+    network clients with which IP address is to be used to contact the main Ert
+    process. By default, Ert will check the operating system routing table to
+    get the IP address in use for non-localhost connections.
+
+    On machines exposing several IP addresses, the correct IP is non-trivial
+    to pick, and by specifying this hook in an installed plugin, any custom
+    code can be injected in order to pick the correct IP."""
--- a/docs/ert/getting_started/howto/plugin_system.rst
+++ b/docs/ert/getting_started/howto/plugin_system.rst
@@ -254,3 +254,20 @@ Minimal example to log to a new file:
         formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
         fh.setFormatter(formatter)
         return fh
+
+
+IP address configuration
+~~~~~~~~~~~~~~~~~~~~~~~~~
+The way Ert chooses IP addresses for communicating with cluster jobs can be configured by plugins.
+
+.. autofunction:: ert.plugins.hook_specifications.net_utils.get_ip_address
+
+Minimal example to override how Ert chooses the IP address:
+
+.. code-block:: python
+
+   import ert
+
+    @ert.plugin(name="my_plugin")
+    def get_ip_address():
+        return "127.0.0.1"
--- a/src/ert/ensemble_evaluator/config.py
+++ b/src/ert/ensemble_evaluator/config.py
@@ -4,7 +4,7 @@ import warnings
 
 import zmq
 
-from ert.shared import get_ip_address
+from ert.plugins.plugin_manager import get_ip_address
 from ert.shared import get_machine_name as ert_shared_get_machine_name
 from ert.shared.constants import PORT_RANGE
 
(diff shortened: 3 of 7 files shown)


jonathan-eq:
@berland Something like this would be the easiest quickfix, but we would have to maintain both implementations, and there is no guarantee the one for private_ip_address even works. As you said, having the get_ip_address() as a site_config would be better, but that requires additional work.

src/ert/config/ert_config.py
```diff
@@ -870,9 +874,10 @@ class ErtConfigWithPlugins(ErtConfig):
             )
             ENV_VARS = dict(runtime_plugins.environment_variables)
             QUEUE_OPTIONS = runtime_plugins.queue_options
-            PRIORITIZE_PRIVATE_IP_ADDRESS = (
-                runtime_plugins.prioritize_private_ip_address
-            )
+
+            @property
+            def get_ip_address_func(self) -> Callable[[], str] | None:
+                return runtime_plugins.get_ip_address_func
```
yngve-sk: Why do we need to get this in the ERT config? I think we should not need it there. When it exists in plugins, we could just hook into runtime plugins where we need it, i.e., when we create the EvaluatorServerConfig

yngve-sk:
Is there a fmu-site-configurations PR associated to this, where we add whatever is default for ERT?

src/ert/cli/main.py
```diff
@@ -133,7 +133,7 @@ def run_cli(args: Namespace, runtime_plugins: ErtRuntimePlugins | None = None) -
         if args.port_range is None
         else (min(args.port_range), max(args.port_range) + 1),
         use_ipc_protocol=using_local_queuesystem,
-        prioritize_private_ip_address=ert_config.prioritize_private_ip_address,
+        get_ip_address_func=ert_config.get_ip_address_func,
```
yngve-sk: Get directly from site plugins

berland:
Can you split this into one commits that removes anything related to prioritize private addresses, and then another commit that adds a plugin feature for overriding get_public_address? In case we need to revert any of these, the other change should not follow along.

jonathan-eq:
> Can you split this into one commits that removes anything related to prioritize private addresses, and then another commit that adds a plugin feature for overriding get_public_address? In case we need to revert any of these, the other change should not follow along.

It is ready 🫡

src/ert/dark_storage/endpoints/experiment_server.py
```diff
@@ -317,9 +317,10 @@ async def run(self) -> None:
             simulation_future = loop.run_in_executor(
                 None,
                 lambda: run_model.start_simulations_thread(
-                    EvaluatorServerConfig()
-                    if run_model.queue_config.queue_system == QueueSystem.LOCAL
-                    else EvaluatorServerConfig(use_ipc_protocol=False)
+                    EvaluatorServerConfig(
+                        use_ipc_protocol=run_model.queue_config.queue_system
+                        == QueueSystem.LOCAL,
+                    )
```
berland: (this change is unrelated to the commit message)

src/ert/plugins/hook_specifications/net_utils.py
```diff
@@ -0,0 +1,6 @@
+from ert.plugins.plugin_manager import hook_specification
+
+
+@hook_specification
+def get_ip_address() -> str:  # type: ignore
```
berland, line 5: We should probably leave a docstring here informing about the intention of this hook, something like:

"Ert uses network communication over TCP/IP and needs to provide potential network clients
with which IP address is to be used to contact the main Ert process. By default, Ert will check
the operating system routing table to get the IP address in use for non-localhost connections

On machines exposing several IP addresses, the correct IP is non-trivial to pick, and by
specifying this hook in an installed plugin, any custom code can be injected in order to
pick the correct IP."

PR #12788, opened 2026-02-04
Merge Everest storage into ERT storage

yngve-sk:
**Issue**
Resolves https://github.com/equinor/ert/issues/11129

**Approach**
Store Everest data frames directly into ERT storage, so they can be exposed through dark storage for plotting
* remove realization_controls.parquet (info already exists in parameters dataset)
* remove realization_objectives.parquet (info already exists in everest_objectives responses dataset)
* remove realization_constraints.parquet (info already exists in everest_constraints responses dataset)
* remove perturbation_objectives.parquet (info already exists in everest_objectives responses dataset)
* remove perturbation_constraints.parquet (info already exists in everest_constraints responses dataset)
* merge batch.json contents into index.json

### DIFF
--- a/src/ert/storage/local_ensemble.py
+++ b/src/ert/storage/local_ensemble.py
@@ -1537,15 +1537,11 @@ class LocalEnsemble(BaseMode):
             )
         )
 
-    @cached_property
+    @property
     def is_improvement(self) -> bool:
         return bool(self._index.is_improvement)
 
     def update_improvement_flag(self, is_improvement: bool) -> None:
-        # Clear the cached prop for the new value to take place
-        if "is_improvement" in self.__dict__:
-            del self.is_improvement
-
         self._index.is_improvement = is_improvement
         self._storage._write_transaction(
             self._path / "index.json",


src/ert/run_models/everest_run_model.py
```diff
@@ -697,8 +733,10 @@ def run_experiment(
         # Initialize the ropt optimizer:
         optimizer, initial_guesses = self._create_optimizer()
 
+        # ROPT expects this folder to exist wrt stdout/stderr redirect files
+        os.makedirs(self.optimization_output_dir, exist_ok=True)
```
berland: `Path(self.optimization_output_dir).mkdir(exist_ok=True)`

src/ert/run_models/everest_run_model.py
```diff
@@ -631,19 +631,55 @@ def start_simulations_thread(
 
     def _handle_optimizer_results(self, results: tuple[Results, ...]) -> None:
         assert self._ever_storage is not None
-        self._ever_storage.unpack_ropt_results(results)
+        assert self._experiment is not None
+
+        batch_dataframes = self._ever_storage.unpack_ropt_results(results)
```
berland: (this looks erroneous commit-by-commit, as you are asserting self._experiment is not None but accessing self._ever_storage - but I see this is changed in a later commit)

src/everest/bin/visualization_script.py
```diff
@@ -57,18 +57,20 @@ def visualization_entry(args: list[str] | None = None) -> None:
...
+        print(
+            f"No data found in storage at {storage._storage.path}. "
+            f"Please try again later"
+        )
+        return
 
-        pm = EverestPluginManager()
-        pm.hook.visualize_data(api=EverestDataAPI(options.config))
+    pm = EverestPluginManager()
+    pm.hook.visualize_data(api=EverestDataAPI(options.config))
```
berland: This is now falling out of the `setup_logging` ctx manager? Intended?

src/ert/run_models/everest_run_model.py
```diff
@@ -630,20 +629,52 @@ def start_simulations_thread(
...
+
+        batch_dataframes = EverestStorage.unpack_ropt_results(results)
+
+        for batch_id, batch_dict in batch_dataframes.items():
+            target_ensemble = self._experiment.get_ensemble_by_name(f"batch_{batch_id}")
+            target_ensemble.save_batch_dataframes(dataframes=batch_dict)
+
+            with open(
+                target_ensemble._path / "batch.json",
+                "w+",
```
berland: Why the `+` here? The file will not be valid json if we just append another "{}'"-construct to the file.

src/ert/run_models/everest_run_model.py
```diff
@@ -630,20 +629,52 @@ def start_simulations_thread(
...
+            (target_ensemble._path / "batch.json").write_text(
+                json.dumps(
+                    {
+                        "batch_id": batch_id,
+                        "is_improvement": False,
+                    },
+                ),
+                encoding="utf-8",
+            )
+            # -----------------------
```
berland: useless line

src/ert/storage/local_ensemble.py
```diff
@@ -79,6 +95,21 @@ class LocalEnsemble(BaseMode):
     parameters and responses.
     """
 
+    BATCH_DATAFRAMES: ClassVar[list[str]] = [
+        "realization_controls",
```
berland: I think this can be simplified with 

`BATCH_DATAFRAMES: ClassVar[list[str]] = BatchDataframes.__annotations__.keys()
`

berland: Rename to `BATCH_DATAFRAME_NAMES`?

src/ert/run_models/everest_run_model.py
yngve-sk: I think the experiment assert should also be there as it also accesses the experiment/ensemble. It is a step in the move/port though, using both everest storage and the experiment.

src/ert/run_models/everest_run_model.py
yngve-sk: This looks like a rebase error with double up of some logic, will fix

yngve-sk: Only need one save of batch.json here. (Will add commit where batch.json contents is merged into index.json of the experiment)

src/everest/bin/visualization_script.py
yngve-sk: Oops, nope, fixed

src/ert/storage/local_ensemble.py
yngve-sk: Will do after moving/porting on what remains, but leave alone in the commits added up to this point

tests/everest/utils/optimal_result.py
```diff
@@ -39,6 +38,8 @@ def get_optimal_result(output_dir: str) -> OptimalResult | None:
             ]
         ).to_dicts()[0]
 
+        experiment._storage.close()
+
```
andreas-el, line 42: This won't close the storage if not `matching_batches` ?
Perhaps this (close()) can be done implicitly on teardown instead?

tests/everest/snapshots/test_api_snapshots/test_api_snapshots/config_advanced.yml/snapshot.json
```diff
@@ -232,91 +232,91 @@
       "batch": 0,
       "function": "distance",
       "realization": 0,
-      "scale": 1.0,
+      "scale": null,
       "simulation": 0,
       "value": -6.1875,
-      "weight": 1.0
+      "weight": null
```
andreas-el: I'm a bit worried regards to `scale` and `weight` now being set to `null` instead of the presumably default  `1.0`. Thoughts?

tests/everest/utils/optimal_result.py
yngve-sk: Nice catch, adding a close if there are no matching batches. Can make EverestStorage.from_storage_path a context manager in a later commit, but for now it is just a port, replacing one `EverestStorage` instance with one `LocalExperiment` instance.

tests/everest/snapshots/test_api_snapshots/test_api_snapshots/config_advanced.yml/snapshot.json
yngve-sk: will fix, data flow changed from 
`EverestDataAPI <- EverestStorage <- ROPT <- ERT Storage`
to
`EverestDataApi <- Ert Storage`
will tweak to make snapshots identical (ideally)

src/ert/storage/local_ensemble.py
```diff
@@ -1216,6 +1247,26 @@ def save_everest_realization_info(
...
+    def save_batch_dataframes(self, dataframes: BatchDataframes) -> None:
+        for df_name in self.BATCH_DATAFRAMES:
+            df = dataframes.get(df_name)
+            if isinstance(df, pl.DataFrame):
+                df.write_parquet(self._path / f"{df_name}.parquet")
+
+    def save_batch_metadata(self, is_improvement: bool) -> None:
+        with open(
+            self._path / "batch.json",
+            "w+",
```
andreas-el: Just make sure that `w+` is what you want.

tests/everest/snapshots/test_api_snapshots/test_api_snapshots/config_multiobj.yml/snapshot.json
```diff
@@ -111,16 +111,16 @@
       "scale": 0.66666667,
       "simulation": 0,
       "value": -0.75,
-      "weight": 0.66666667
+      "weight": 0.5
```
andreas-el, line 114: Seems like all/most weights suddenly are slightly smaller than before. As if divided by n + 1 instead of n.
2 / 3 became 2 / 4 etc.

src/ert/storage/local_experiment.py
```diff
@@ -509,3 +516,507 @@ def _update_response_keys(
...
+            output_constraint_violation = (
+                0.0
+                if b.batch_output_constraint_violations is None
+                else (
+                    b.batch_output_constraint_violations.drop("batch_id")
+                    .to_numpy()
+                    .min()
+                    .item()
+                )
+            )
```
andreas-el: Create a helper function for this, that takes the constraints as input

src/ert/storage/local_experiment.py
```diff
@@ -509,3 +516,507 @@ def _update_response_keys(
...
+            for b in self.batches
+            if b.has_gradient_results
+        ]
+
+    def on_optimization_finished(self) -> None:
+        logger.debug("Storing final results Everest storage")
+
+        # This a somewhat arbitrary threshold, this should be a user choice
+        # during visualization:
+        CONSTRAINT_TOL = 1e-6
```
andreas-el: Move to top of file?

src/ert/storage/local_experiment.py
```diff
@@ -509,3 +516,507 @@ def _update_response_keys(
...
+            batch_id: int, *dfs: pl.DataFrame | None
+        ) -> None:
+            for df_ in dfs:
+                _try_append_df(batch_id, df_, realization_dfs_to_join)
+
+        def try_append_perturbation_dfs(
+            batch_id: int, *dfs: pl.DataFrame | None
+        ) -> None:
+            for df_ in dfs:
+                _try_append_df(batch_id, df_, perturbation_dfs_to_join)
```
andreas-el: These all perform the same action, please merge into one generic function?

src/ert/storage/local_experiment.py
andreas-el: Could be a nice issue also

src/ert/storage/local_experiment.py
andreas-el: Could be a nice issue also

src/ert/storage/local_experiment.py
andreas-el: Could be a nice issue also

src/ert/storage/local_ensemble.py
yngve-sk: does not already exist, so will remove

tests/everest/snapshots/test_api_snapshots/test_api_snapshots/config_advanced.yml/snapshot.json
```diff
@@ -343,26 +343,31 @@
     {
       "accepted": 1,
       "batch": 0,
+      "distance": -1.6875,
       "objective": -1.6875
```
andreas-el: Distance and objective just happens to be the same?

tests/everest/snapshots/test_api_snapshots/test_api_snapshots/config_multiobj.yml/snapshot.json
```diff
@@ -179,22 +179,22 @@
...
+      "distance_p": -0.57423449,
+      "distance_q": -1.17591,
       "objective": -2.07831065
     },
     {
       "accepted": 1,
       "batch": 2,
-      "distance_p": -0.50778502,
-      "distance_q": -1.49226332,
+      "distance_p": -0.38083877,
```
yngve-sk, line 196: "duplicate" of https://github.com/equinor/ert/pull/12788#discussion_r2782271675
This is a consequence of the weights being different, ref on main:
```
        for o in objectives.to_dicts():
            batch_datas = batch_datas.with_columns(
                pl.col(o["objective_name"]) * o["weight"] / o["scale"]
            )
```

vs here:
```
        for name, weight, scale in zip(
            objectives.keys, objectives.weights, objectives.scales, strict=False
        ):
            batch_datas = batch_datas.with_columns(pl.col(name) * weight / scale)
```

tests/everest/snapshots/test_api_snapshots/test_api_snapshots/config_advanced.yml/snapshot.json
yngve-sk: Fixed, seems we only do total objective value in the everest data API if there is only one objective

tests/everest/snapshots/test_api_snapshots/test_api_snapshots/config_multiobj.yml/snapshot.json
yngve-sk: This comes out from ROPT, and differs from what directly exists in the everest config (which is what is "active" in this PR). Edit: ROPT uses normalized weights, thus we will store the weights unnormalized as they are metadata.

src/ert/run_models/everest_run_model.py
```diff
@@ -765,7 +765,7 @@ def objectives_config(self) -> EverestObjectivesConfig:
         assert isinstance(obj_config, EverestObjectivesConfig)
         return obj_config
 
-    def on_optimization_finished(self) -> None:
+    def _update_ensemble_improvement_flags(self) -> None:
         assert self._experiment is not None
         logger.debug("Storing final results Everest storage")
```
berland: This debug log is maybe also misleading if the function name is changing?

yngve-sk: The logging statement is no longer true either, I think removing this makes the most sense. We do have a logging statement for experiment end. The overall logging in here could be revised a bit but suggest leaving that for a separate issue.

src/ert/storage/local_ensemble.py
```diff
@@ -1553,17 +1539,18 @@ def perturbation_constraints(self) -> pl.DataFrame | None:
 
     @cached_property
     def is_improvement(self) -> bool:
-        info = json.loads((self._path / "batch.json").read_text(encoding="utf-8"))
-        return bool(info["is_improvement"])
+        return bool(self._index.is_improvement)
```
berland, line 1542: We should probably remove the `cached_property` decorator when the function becomes as simple as this.

src/ert/storage/local_ensemble.py
```diff
@@ -1553,17 +1539,18 @@ def perturbation_constraints(self) -> pl.DataFrame | None:
...
 
     def write_metadata(self, is_improvement: bool) -> None:
         # Clear the cached prop for the new value to take place
         if "is_improvement" in self.__dict__:
```
berland: Is this and the following line doing anything now?

yngve-sk: nope, removing

berland: ah, should be deleted when `@cached_property` decorator is removed

PR #12789, opened 2026-02-04
Limit capturing of warnings from forward model steps

### LINKED ISSUE #12781: Excessive amounts of warnings from fm step logged
Operation id 0334c6eab6fb09431665c90497d90c0a has logged around 230K times. 

Most of these seem to be variants of

`Realization 1 step SUMO_UPLOAD.200 warned 1 time(s) in stdout: 2026-01-27 00:40:21,411:WARNING:fmu.sumo.uploader:Metadata upload unsuccessful, returning 404`

This logging should be capped in some way.

berland:
Leaving this uncapped can give an arbitrary amount of log messages, which will be useless the user seeing this in the GUI, and will also clog the logging system.

**Issue**
Resolves #12781 

**Approach**
⛔ 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -581,4 +581,8 @@ async def log_warnings_from_forward_model(
                     )
             if timeout_seconds <= 0:
                 break
+    if log_count >= max_logged_warnings:
+        logger.warning(
+            "Reached maximum number of forward model step warnings to extract"
+        )
     return timeout_seconds
--- a/tests/ert/unit_tests/scheduler/test_job.py
+++ b/tests/ert/unit_tests/scheduler/test_job.py
@@ -483,6 +483,7 @@ async def test_log_warnings_from_forward_model(
             warnings.simplefilter("error")
             await log_warnings_from_forward_model(realization, start_time - 1)
         assert emitted_warning_str not in caplog.text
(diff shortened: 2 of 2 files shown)


src/ert/scheduler/job.py
```diff
@@ -482,6 +482,7 @@ async def log_warnings_from_forward_model(
     real: Realization,
     job_submission_time: float,
     timeout_seconds: int = Job.DEFAULT_FILE_VERIFICATION_TIMEOUT,
+    max_logged_warnings: int = 200,
```
andreas-el, line 485: Do you think this will be high enough? Since this is the sum of all warnings, others might be overlooked. This is per realization?

berland: This is up for discussion. 200 is not very much, agree, but still more than what the user is able to take in when these are flashed in the GUI. 

Some warnings will for sure be overlooked, and then will be missing in our logging system if we search for specific warnings, but that will be the case for any picked number.

andreas-el: Would it be a bad idea to log a certain message every time we hit this cap?

berland: Added

PR #12791, opened 2026-02-04
Fix inconsistent adaptive localization correlation threshold

### LINKED ISSUE #12155: Misleading "Adaptive localization correlation threshold" parameter in AnalysisConfig

I Tried to use adaptive localization in ERT, and noticed that the correlation threshold changed depending on the "max realizations" (either from design matrix, from active realizations lenght etc). So if I changed a design matrix realization 100 it would change this number a lot, even if I have the same number of realizations.

However, when debugging the code (after hints from @larsevj) it seems like, unless the user have touched this input, it will be re-calculated in iterative_ensemble_smoother (based on the correct number of realizations) 

Image

I would expect the number to be the number that is used.
https://github.com/equinor/iterative_ensemble_smoother/blob/8dcb3dabfe28876a03aa67ae939911719c1d5526/src/iterative_ensemble_smoother/experimental.py#L313

### Steps to reproduce

```Python
1. Make a design matrix with some realizations
2. Go to e.g. es_mda, and click the analysis view
3. Quit, and change the design matrix to have a much higher number in the last realization
4. Open ert again, go to es_mda, click the analysis view, and observe that the correlation threshold has changed.

So it should not change, and the one that is there should be used? To see this last part, debugging is probably needed.
```

- [ ] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

erlenlh:
**Issue**
Resolves #12155 

**Approach**
Change input of threshold calculation to only consider active realizations

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/config/analysis_module.py
+++ b/src/ert/config/analysis_module.py
@@ -30,6 +30,18 @@ es_description = """
         U, w, V.T = svd(D_delta) then we assume that U @ U.T = I.
     """
 
+loc_description = """
+    The default adaptive localization correlation threshold
+    is computed as 3/sqrt(ensemble_size), where ensemble_size
+    is the number of active realizations in the ensemble.
+
+    You can override this value by setting a custom threshold here or in the config.
+    """
+
+cust_loc_thresh_description = """
+    Adaptive localization correlation threshold:
+    """
+
 
 class ESSettings(BaseModel):
     model_config = ConfigDict(extra="forbid", validate_assignment=True)
@@ -40,13 +52,16 @@ class ESSettings(BaseModel):
     inversion: Annotated[
         InversionTypeES, Field(title="Inversion algorithm", description=es_description)
     ] = "EXACT"
-    localization: Annotated[bool, Field(title="Adaptive localization")] = False
+    localization: Annotated[
+        bool, Field(title="Enable adaptive localization", description=loc_description)
+    ] = False
     localization_correlation_threshold: Annotated[
         float | None,
         Field(
             ge=0.0,
             le=1.0,
-            title="Adaptive localization correlation threshold",
+            title="Custom adaptive localization correlation threshold",
+            description=cust_loc_thresh_description,
         ),
     ] = None
     distance_localization: Annotated[
--- a/src/ert/gui/ertwidgets/analysismodulevariablespanel.py
+++ b/src/ert/gui/ertwidgets/analysismodulevariablespanel.py
@@ -14,6 +14,7 @@ from PyQt6.QtWidgets import (
     QHBoxLayout,
     QLabel,
     QLayout,
+    QVBoxLayout,
     QWidget,
 )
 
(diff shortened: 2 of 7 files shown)


src/ert/config/analysis_module.py
```diff
@@ -40,16 +48,30 @@ class ESSettings(BaseModel):
...
         InversionTypeES, Field(title="Inversion algorithm", description=es_description)
     ] = "EXACT"
-    localization: Annotated[bool, Field(title="Adaptive localization")] = False
+    localization: Annotated[
+        bool, Field(title="Enable adaptive localization", description=loc_description)
+    ] = False
+    custom_localization_threshold: Annotated[
+        bool, Field(title="Custom localization correlation threshold")
+    ] = False
     localization_correlation_threshold: Annotated[
```
xjules, line 58: Not sure if I understand why we need to have both; ie. `localization` and `custom_localization_threshold`?

erlenlh: re-wrote this to remove the new flag.

src/ert/gui/ertwidgets/analysismodulevariablespanel.py
```diff
@@ -61,25 +62,45 @@ def __init__(self, analysis_module: AnalysisModule, ensemble_size: int) -> None:
...
-
-        metadata = AnalysisModule.model_fields["localization_correlation_threshold"]
-        local_checkbox = QCheckBox(metadata.title)
-        local_checkbox.setObjectName("localization")
-        local_checkbox.clicked.connect(
+        localization_frame.setLayout(QVBoxLayout())
+        main_loc_layout = cast(QLayout, localization_frame.layout())
+        main_loc_layout.setContentsMargins(0, 0, 0, 0)
+
+        row1_frame = QFrame()
```
xjules: `row1` - is there any reason for `1` if both variables? Maybe we can be explicit
`row_checkbox_layout`
`row_threshold_layout`

or something similar

erlenlh: Agree, renamed

tests/ert/unit_tests/gui/experiments/test_multiple_data_assimilation_panel.py
```diff
@@ -128,3 +137,58 @@ def test_multiple_data_assimilation_panel_sets_active_realizations_to_initial_ac
     assert mda_panel._active_realizations_field.text() == active_realizations_string
     mda_panel.restart_run_toggled()
     assert mda_panel._active_realizations_field.text() == active_realizations_string
+
+
+def _open_and_capture_threshold(panel, qtbot):
+    captured = {"value": None}
```
jonathan-eq: I think this shouldn't be a dictionary, maybe just a variable

tests/ert/unit_tests/gui/experiments/test_multiple_data_assimilation_panel.py
```diff
@@ -128,3 +137,58 @@ def test_multiple_data_assimilation_panel_sets_active_realizations_to_initial_ac
...
+
+    def inspect_and_close_dialog() -> None:
+        for w in QApplication.topLevelWidgets():
+            if isinstance(w, QDialog) and w.windowTitle() == "Edit variables":
+                spinner = w.findChild(QDoubleSpinBox, name="localization_threshold")
+                assert spinner is not None
+                captured["value"] = spinner.value()
+                w.accept()
+                break
```
jonathan-eq: I dont think we need to test the spinner, and there might be an easier way to do this. I am not sure what QApplication.topLevelWidgets() is.

tests/ert/unit_tests/gui/simulation/test_multiple_data_assimilation_panel.py
```diff
@@ -128,3 +137,58 @@ def test_multiple_data_assimilation_panel_sets_active_realizations_to_initial_ac
...
+    def inspect_and_close_dialog() -> None:
+        for w in QApplication.topLevelWidgets():
+            if isinstance(w, QDialog) and w.windowTitle() == "Edit variables":
+                spinner = w.findChild(QDoubleSpinBox, name="localization_threshold")
+                assert spinner is not None
+                captured["value"] = spinner.value()
+                w.accept()
+                break
+
+    QTimer.singleShot(0, inspect_and_close_dialog)
```
jonathan-eq: How can we inspect the dialog before we press the tool_button?
Can we try adding qt sleep and running the test with --show-gui so we can see what it does?

erlenlh: bilde

putting a sleep on the line before the `.accept()` gives me just this window!

tests/ert/unit_tests/gui/experiments/test_multiple_data_assimilation_panel.py
erlenlh: agree!

tests/ert/unit_tests/gui/simulation/test_multiple_data_assimilation_panel.py
erlenlh: With this code around. Note sure how I´m spawning the window tbh
```
def _open_and_capture_threshold(panel, qtbot):
    captured_value = None

    def inspect_and_close_dialog() -> None:
        nonlocal captured_value
        dialog = QApplication.activeModalWidget()
        if isinstance(dialog, QDialog) and dialog.windowTitle() == "Edit variables":
            spinner = dialog.findChild(QDoubleSpinBox, name="localization_threshold")
            if spinner is not None:
                captured_value = spinner.value()
            time.sleep(5)
            dialog.accept()

    QTimer.singleShot(0, inspect_and_close_dialog)
```

jonathan-eq: Hmmm, it seems like the popup is an ert closabledialog, which is a QDialog. I think something like this could work (chatgpt)
```
def test_button_opens_dialog(qtbot, main_window):
    # 1. Click the button
    qtbot.mouseClick(main_window.my_button, Qt.LeftButton)

    # 2. Wait until a QDialog child is visible
    dialog = None
    def dialog_shown():
        nonlocal dialog
        dialog = main_window.findChild(QDialog)
        return dialog is not None and dialog.isVisible()

    qtbot.waitUntil(dialog_shown, timeout=3000)  # wait up to 3 seconds

    # 3. Assert dialog is shown
    assert dialog.isVisible()
```

erlenlh: if my understanding is correct we cant/shouldnt use mainwindow in unit tests? seems like those fixtures live in gui tests, and that makes sense i guess. Pushed a new and cleaner version now, look again pretty please? :)

jonathan-eq (approved):
There does not seem to be a better way to get the active pop up except for QApplication.activeModalWidget, so we will keep it as is. Very good! 🥇

PR #12792, opened 2026-02-04
Refactor update_from_event in ensemble evaluator with match case

### LINKED ISSUE #10038: Use match-case in ensemble_evaluator/snapshot.py
Python 3.10's match-case structure can be applied in ensemble_evaluator/snapshot.py as pr the suggestion in 
https://github.com/equinor/ert/pull/10025/files/105f8917cd1d0f26b7500aa13aad38725d3b4678#r1950652107

achaikou:
**Issue**
Resolves #10038 

**Approach**
Make ensemble evaluator code use match case.
There were both upsides and downsides with using match, but general first response was positively for it, so it goes in :slightly_smiling_face: 

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/ensemble_evaluator/snapshot.py
+++ b/src/ert/ensemble_evaluator/snapshot.py
@@ -4,7 +4,7 @@ import logging
 from collections import Counter, defaultdict
 from collections.abc import Mapping
 from datetime import datetime
-from typing import Any, TypeVar, cast, get_args
+from typing import Any, TypeVar, cast
 
 from typing_extensions import TypedDict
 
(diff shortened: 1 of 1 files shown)


src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -382,57 +361,85 @@ def update_from_event(
...
+                if not isinstance(ensemble_event, EnsembleEvaluationWarning):
+                    self._ensemble_state = _ENSEMBLE_TYPE_EVENT_TO_STATUS[
+                        type(ensemble_event)
+                    ]
+            case EESnapshotUpdate():
+                self.merge_snapshot(EnsembleSnapshot.from_nested_dict(event.snapshot))
+            case EESnapshot():
+                return EnsembleSnapshot.from_nested_dict(event.snapshot)
+            case _:
+                raise ValueError(f"Unknown type: {e_type}")
```
eivindjahren, line 438: If the match is exhaustive according to the type then we might want to consider using `case default: assert_never(default)` here.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -382,57 +361,85 @@ def update_from_event(
...
+        fm.update(fm_data)
+        fm = _filter_nones(fm)
 
-                if type(event) is ForwardModelStepFailure:
-                    error = event.error_msg or ""
-                else:
-                    # Make sure error msg from previous failed run is replaced
-                    error = ""
+        self.update_fm_step(
+            event.real,
```
eivindjahren: I think ruff would have all of this on one line if you remove the comma on line 410, and to me this is more readable here.

src/ert/ensemble_evaluator/snapshot.py
```diff
@@ -382,57 +361,85 @@ def update_from_event(
...
-            return EnsembleSnapshot.from_nested_dict(event.snapshot)
-        else:
-            raise ValueError(f"Unknown type: {e_type}")
+        match event:
+            case realization_event if isinstance(realization_event, RealizationEvent):
+                self.update_from_realization_event(realization_event, source_snapshot)
+            case fm_event if isinstance(fm_event, FMEvent):
+                self.update_from_fm_event(fm_event)
+            case ensemble_event if isinstance(ensemble_event, EnsembleEvent):
+                if not isinstance(ensemble_event, EnsembleEvaluationWarning):
```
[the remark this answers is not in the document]
eivindjahren: I do not think copilot is correct here, this works fine for python 3.10 and later.

src/ert/ensemble_evaluator/snapshot.py
achaikou: It would be a nice thing to do as that should help codecov too.

Problem is that `match` doesn't work nicely with union types like `RealizationEvent` and so on. I had to bypass that with conditional `isinstance`.
However that makes linter to think that not all options are exhausted and complain when I do `assert_never(default)`

Don't know who I should aim to please more - Codecov or Pylance 😄 I'll go with Pylance for now.

Issue #11159, opened 2025-06-18
Move magic number 51820 to a common place for Ert


berland:
The port number for Ert is mentioned several places in the source code:

https://github.com/equinor/ert/blob/7ae692029e4a7574e01f778ee74f2a715d872e36/src/ert/ensemble_evaluator/config.py#L39

https://github.com/equinor/ert/blob/7ae692029e4a7574e01f778ee74f2a715d872e36/src/ert/shared/net_utils.py#L50

At the same place, a mnenomic should be made for dark_storage ports:

https://github.com/equinor/ert/blob/7ae692029e4a7574e01f778ee74f2a715d872e36/src/ert/services/_storage_main.py#L107

The number 51820 is not by coincidence, and is computed as:

```python
>>> def idx(x: str):
...     str(return ord(x) - ord("a") + 1)
>>> "".join([idx(x) for x in "ert"])
'51820'
```

PR #12794, opened 2026-02-05
Make port range constant and replaced magic numbers used with it

eqbech:
**Issue**
Resolves https://github.com/equinor/ert/issues/11159

Moved the magic number into a constant and applied it where the magic number previously appeared. Also updated the way the ports are handled for consistency.


src/ert/shared/net_utils.py
```diff
@@ -49,7 +51,7 @@ def get_machine_name() -> str:
 
 def find_available_socket(
     host: str | None = None,
-    port_range: range = range(51820, 51840 + 1),
+    port_range: tuple[int, int] = PORT_RANGE,
```
berland: Is this change necesseraly tied to making the port range a constant? Can you make an argument for this type change and put it in the git commit body? 

(and if it is not tied to the constant-thingie, it could have been in a separate commit)

eqbech: Updated commits to reflect this change!

eqbech: Removed type change from this pr, now it only includes the added port range constant.

src/ert/ensemble_evaluator/config.py
```diff
@@ -23,7 +24,7 @@ def get_machine_name() -> str:
 class EvaluatorServerConfig:
     def __init__(
         self,
-        port_range: tuple[int, int] | None = None,
+        port_range: tuple[int, int] | None = PORT_RANGE,
```
andreas-el, line 27: Perhaps we should disallow `None` here? Why would we want this here?

eqbech: Not sure, but its called with nullable ports at least once, here's an example.
https://github.com/equinor/ert/blob/0fc0d1958f76ab498db33f6bc7b533396f3802df/src/ert/cli/main.py#L131-L136

andreas-el: Okay, so this is to allow custom port range from CLI.

```
cli_parser.add_argument(
            "--port-range",
            type=valid_port_range,
            required=False,
            help="Port range [a,b] to be used by the evaluator. Format: a-b",
        )
```

andreas-el (approved):
Nice! 🍖

PR #12807, opened 2026-02-05
Fix flaky `test_event_reporter.py::test_event_reporter_does_not_hang_…

### LINKED ISSUE #12799: flaky test: tests/ert/unit_tests/forward_model_runner/test_event_reporter.py::test_event_reporter_does_not_hang_after_failed[failed_connect]
https://github.com/equinor/ert/actions/runs/21706842965/job/62601458279#step:11:5684
https://github.com/equinor/ert/actions/runs/21509660890/job/61973885375
https://github.com/equinor/ert/actions/runs/21482137582/job/61881345255

jonathan-eq:
…after_failed`

**Issue**
Resolves #12799

**Approach**

This commit fixes the flaky test by adding a helper function in the event reporter for re-raising the exception in publisher thread. The flakiness was introduced in 3f4fcfcd3345834073fa02704321ee23d2bbf752, and was caused by the reporter publisher thread exiting very fast, and setting the exception before we had even sent all messages in the test. The final reporter.report(...) call would reraise the exception and cause the test to fail.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/forward_model_runner/test_event_reporter.py
+++ b/tests/ert/unit_tests/forward_model_runner/test_event_reporter.py
@@ -1,4 +1,6 @@
 import os
+import time
+from contextlib import suppress
 
 import pytest
 
@@ -253,12 +255,10 @@ def test_event_reporter_does_not_hang_after_failed(
         )
 
         reporter.report(Init([fmstep1], 1, 19, ens_id="ens_id", real_id=0))
-        if mocked_server_signal == MockZMQServerSignal.FAIL_ACK_BUT_STORE_EVENTS:
+        # The reporter will raise this error if the previous event has already been
+        # processed and failed, but we are not interested in this error in this test.
+        with suppress(ClientConnectionError):
             reporter.report(Start(fmstep1))
-        elif (
-            mocked_server_signal
-            == MockZMQServerSignal.NORMAL_OPERATION_BUT_FAIL_ACK_DISCONNECT
-        ):
             reporter.report(Finish())
 
         reporter._event_publisher_thread.join(timeout=10)
@@ -269,3 +269,39 @@ def test_event_reporter_does_not_hang_after_failed(
             "Expected ClientConnectionError in event publisher thread"
         )
     assert expected_message in caplog.text
+
+
+@pytest.mark.integration_test
+def test_that_event_reporter_raises_exception_on_following_report_after_connection_failure(  # noqa: E501
+    monkeypatch,
+):
+    monkeypatch.setattr(
+        "_ert.forward_model_runner.reporting.event.Client.DEFAULT_MAX_RETRIES", 0
+    )
+    reporter = Event(
+        evaluator_url="ipc:///no_server_running_here:1337",
+        ack_timeout=0.01,
+        max_retries=0,
+    )
+    fmstep1 = ForwardModelStep(
+        {"name": "fmstep1", "stdout": "stdout", "stderr": "stderr"}, 0
+    )
+
+    reporter.report(Init([fmstep1], 1, 19, ens_id="ens_id", real_id=0))
+    # The reporter will raise this error if the previous event has
+    # already been processed and sending failed
+    for _ in range(100):
+        if reporter._reporter_exception is not None:
+            break
+        time.sleep(0.05)
+    else:
+        raise AssertionError(
+            "ClientConnectionError was not set in reporter after connection failure"
+        )
+    with pytest.raises(ClientConnectionError, match=r"Failed to send b'CONNECT'"):
+        reporter.report(Start(fmstep1))
+
+    reporter._event_publisher_thread.join(timeout=10)
+    assert not reporter._event_publisher_thread.is_alive(), (
+        "Event publisher thread is hanging"
+    )


src/_ert/forward_model_runner/reporting/event.py
```diff
@@ -180,9 +180,18 @@ async def publisher() -> None:
     def report(self, msg: Message) -> None:
         if self._reporter_exception is not None:
             self.stop()
-            raise self._reporter_exception
+            self._propagate_reporter_exception()
         self._statemachine.transition(msg)
 
+    def _propagate_reporter_exception(self) -> None:
+        """This is mainly for mocking purposes in tests
```
eivindjahren: This is not a useful comment for when reading the code.

tests/ert/unit_tests/forward_model_runner/test_event_reporter.py
```diff
@@ -224,6 +224,10 @@ def test_event_reporter_does_not_hang_after_failed(
         "_ert.forward_model_runner.reporting.event.Client.DEFAULT_MAX_RETRIES", 0
     )
 
+    monkeypatch.setattr(
+        "_ert.forward_model_runner.reporting.event.Event._propagate_reporter_exception",
+        lambda self: None,
```
eivindjahren: It does not seem like a good idea to use a mock to change the behavior of the code. The behavior being tested is no longer representative of what the code does.

tests/ert/unit_tests/forward_model_runner/test_event_reporter.py
```diff
@@ -224,6 +224,10 @@ def test_event_reporter_does_not_hang_after_failed(
...
+    )
```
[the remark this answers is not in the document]
eivindjahren: This is exactly my objection also, but I do not believe adding another test is the correct route. The behavior being tested for is not the actual behavior and we should either change the test or fix the code which is behaving incorrectly.

tests/ert/unit_tests/forward_model_runner/test_event_reporter.py
```diff
@@ -233,11 +234,15 @@ def test_event_reporter_does_not_hang_after_failed(
         )
 
         reporter.report(Init([fmstep1], 1, 19, ens_id="ens_id", real_id=0))
-        reporter.report(Start(fmstep1))
-        reporter.report(Finish())
+        if mocked_server_signal != 5:
```
jonathan-eq: Either this, suppress it, or make a new test for the case with failing initial connection. 
We can also do a final reporter.report(...) after the exception has occured to make sure it raises it, but thats a different test.

tests/ert/unit_tests/forward_model_runner/test_event_reporter.py
```diff
@@ -233,11 +234,15 @@ def test_event_reporter_does_not_hang_after_failed(
...
+            reporter.report(Start(fmstep1))
+            reporter.report(Finish())
```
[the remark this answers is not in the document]
jonathan-eq: This goes for all tests using MockZMQServer. Maybe it would be better to suppress the initial exceptions, and assert it at the end.

tests/ert/unit_tests/forward_model_runner/test_event_reporter.py
eivindjahren: If you write it as 

```python
FAILED_CONNECT_SIGNAL = 5
...
    if mocked_server_signal != FAILED_CONNECT_SIGNAL:
```
then it is a bit easier to understand what is going on. Also, why does the initial connection fail? Should that be possible here? Could the test be written so that the initial connection does not fail?

jonathan-eq: That is the test case. We could split it out into a separate test to avoid the if check in the middle of it.

eivindjahren: Well the test assert that it does get messages, while it will not get those if the connection fails early. The easiest to reason about later is that which messages you get is not dependent on when the connection fails.

tests/ert/unit_tests/forward_model_runner/test_event_reporter.py
```diff
@@ -210,11 +217,24 @@ def test_report_with_reconnected_reporter_but_finished_jobs():
 @pytest.mark.parametrize(
     ("mocked_server_signal", "ack_timeout", "expected_message"),
     [
-        pytest.param(5, 0.01, "No ack for dealer connection", id="failed_connect"),
         pytest.param(
-            4, 0.25, "No ack for dealer disconnection", id="failed_disconnect"
+            MockZMQServerSignal.ACK_NOTHING,
+            0.10,
+            "No ack for dealer connection",
+            id="failed_connect",
```
xjules, line 226: You made a change to this one @jonathan-eq to speed things up:
```
- pytest.param(4, "No ack for dealer disconnection", id="failed_disconnect"),
+ pytest.param(5, 0.01, "No ack for dealer connection", id="failed_connect"),
```
Could this be related?
I don't think the changes you made are related to the flakiness

jonathan-eq: Yes, that is the cause of it being flaky, but the test is still very timing dependent.

jonathan-eq: The flakiness is due to a side effect in reporter.report(...) that is completely unrelated to what we testing here. We could continue down that route, and add asserts for it, or move it to a different test.

eivindjahren: So either we

* make the test take into account the side-effect of `reporter.report()`,
* make the tested behavior consistent regardless of what `reporter.report()` does, or
* mock reporter.report() so that you get consistent behavior.

PR #12809, opened 2026-02-05
Use unique dates for lookup in refcase test

### LINKED ISSUE #12721: Check hypothesis failure for testing date keyword sets report step by looking up refcase
See https://github.com/equinor/ert/pull/12720/changes/4292c5f17f72094d3aeb88d396945c4cb515715a

eivindjahren:
After changing to migrating observation configs for users a corner case was introduced where it may be ambiguous which report step was ment as we round to seconds.

This changes the test to only apply to refcases with unique dates. There is still the question of whether it is good enough for the migration to work when the refcase has sufficiently distinct report step times.

Resolves #12721

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/config/test_observations.py
+++ b/tests/ert/unit_tests/config/test_observations.py
@@ -1,6 +1,6 @@
 import logging
 from contextlib import ExitStack as does_not_raise
-from datetime import datetime, timedelta
+from datetime import date, datetime, timedelta
 from pathlib import Path
 from textwrap import dedent
 from typing import cast
@@ -473,14 +473,29 @@ def test_that_the_date_keyword_sets_the_general_index_by_looking_up_time_map():
     assert observations["gen_data"].to_dicts()[0]["report_step"] == restart
 
 
-@given(summary=summaries(), data=st.data())
+@given(data=st.data())
 @pytest.mark.integration_test
 def test_that_the_date_keyword_sets_the_report_step_by_looking_up_refcase(
-    tmp_path_factory: pytest.TempPathFactory, summary, data
+    tmp_path_factory: pytest.TempPathFactory, data
 ):
+    start_date, *times = sorted(
+        data.draw(
+            st.lists(
+                st.dates(min_value=date(1969, 1, 1), max_value=date(2100, 1, 1)),
+                min_size=2,
+                unique=True,
+            )
+        )
+    )
+    times = [(t - start_date).total_seconds() / 3600 for t in times]
+    smspec, unsmry = data.draw(
+        summaries(
+            start_date=st.just(datetime.combine(start_date, datetime.min.time())),
+            time_deltas=st.just(times),
+        )
+    )
     with pytest.MonkeyPatch.context() as patch:
         patch.chdir(tmp_path_factory.mktemp("history_observation_values_are_fetched"))
-        smspec, unsmry = summary
         smspec.to_file("ECLIPSE_CASE.SMSPEC")
         unsmry.to_file("ECLIPSE_CASE.UNSMRY")
         start_date = smspec.start_date.to_datetime()


tests/ert/unit_tests/config/test_observations.py
```diff
@@ -473,14 +473,29 @@ def test_that_the_date_keyword_sets_the_general_index_by_looking_up_time_map():
     assert observations["gen_data"].to_dicts()[0]["report_step"] == restart
 
 
-@given(summary=summaries(), data=st.data())
+@given(data=st.data())
 @pytest.mark.integration_test
 def test_that_the_date_keyword_sets_the_report_step_by_looking_up_refcase(
```
achaikou, line 478: I understand that `DATE` became deprecated for `GENERAL_OBSERVATION` (though I don't find any information about deprecation in docs).
Are new ERT versions supposed to be backwards compatible with old configs?
- If so, is there a policy for how long? Forever? It seems ert configs are not versioned, so I wonder how it is possible to match correctly ert version with ert config version to ever remove those.
- If not, should we keep testing deprecated functionality?

tests/ert/unit_tests/config/test_observations.py
```diff
@@ -473,14 +473,29 @@ def test_that_the_date_keyword_sets_the_general_index_by_looking_up_time_map():
...
+    start_date, *times = sorted(
+        data.draw(
+            st.lists(
+                st.dates(min_value=date(1969, 1, 1), max_value=date(2100, 1, 1)),
+                min_size=2,
+                unique=True,
+            )
+        )
+    )
+    times = [(t - start_date).total_seconds() / 3600 for t in times]
```
achaikou, line 490: Don't understand what you are doing in that line.
Assuming start_date is January 1st 2026 and t is January 2nd 2026, `t - start_date` is timedelta of a day. 
`1day.total_seconds() / 3600 = 24 `

So basically you multiply every drawn day difference by 24.

I think because of this Hypothesis also prints some very-far-in-the-future years, while I thought that we want to generally limit to 1969-2100 range as `data.draw` range.

`time_map: [datetime.datetime(1999, 7, 17, 0, 0), datetime.datetime(3112, 7, 14, 0, 0), datetime.datetime(3193, 8, 14, 0, 0), datetime.datetime(3303, 3, 23, 0, 0)]`

Was that intended? If so - why?

tests/ert/unit_tests/config/test_observations.py
```diff
@@ -473,14 +473,29 @@ def test_that_the_date_keyword_sets_the_general_index_by_looking_up_time_map():
...
+    smspec, unsmry = data.draw(
+        summaries(
+            start_date=st.just(datetime.combine(start_date, datetime.min.time())),
+            time_deltas=st.just(times),
```
achaikou, line 494: We got into this problem because timedeltas (which are days or hours) are too close floats, so our datetimes without nanoseconds were indistinguishable. 
`2500.0, 2499.9999999999995`

Now, `times` are dates, not datetimes. Then `times` becomes some integer number, which in `.summaries` will be treated either as days or hours (hours seem to be x24, so full days anyway).

So by this change we drop a whole range of possible `DATE` values, always ending up at midnight.
Is there a reason for limiting us to the full days and dropping the time part?

tests/ert/unit_tests/config/test_observations.py
eivindjahren: It was deprecated, but for now we are migrating the config for the user. This should be a very rare occurrence, if it ever happens. We will have too look in our logs to see when this can be removed.

tests/ert/unit_tests/config/test_observations.py
eivindjahren: Oh, this is my bad, I did this incorrectly.

tests/ert/unit_tests/config/test_observations.py
eivindjahren: This would be the most common case, and therefore we made this slightly more coarse than necessary. From what I have seen most of the time, users will only have dates in their observations so they end up at midnight.

PR #12819, opened 2026-02-09
Move some everestconfig defaulting

yngve-sk:
Make defaulting happen before creating the EverestRunModel and corresponding ERT response configs

### DIFF
--- a/src/everest/config/everest_config.py
+++ b/src/everest/config/everest_config.py
@@ -1123,8 +1123,14 @@ to read summary data from forward model, do:
         return EverestObjectivesConfig(
             keys=objective_names,
             input_files=objective_names,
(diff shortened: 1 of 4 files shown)


src/everest/config/everest_config.py
```diff
@@ -1123,8 +1123,8 @@ def create_ert_objectives_config(self) -> EverestObjectivesConfig:
         return EverestObjectivesConfig(
             keys=objective_names,
             input_files=objective_names,
-            weights=[o.weight for o in self.objective_functions],
-            scales=[o.scale for o in self.objective_functions],
+            weights=[o.weight or 1.0 for o in self.objective_functions],
```
andreas-el: This uses the truthy evaluation, will `0` then also result in `1.0` ?

PR #12833, opened 2026-02-10
Autoscaling tests

dafeda:
      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
+++ b/tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
@@ -2,6 +2,7 @@ import numpy as np
 import pytest
 
 from ert.analysis.misfit_preprocessor import (
+    cluster_responses,
     get_nr_primary_components,
     get_scaling_factor,
     main,
@@ -142,34 +143,33 @@ def test_that_correlated_and_independent_observations_are_grouped_separately(
     np.testing.assert_equal(obs_errors, obs_error_copy)
 
 
-@pytest.mark.parametrize("nr_observations", [0, 1, 2])
-def test_edge_cases_with_few_observations_return_default_values(nr_observations):
-    """Test that edge cases with 0-2 observations return default scaling values.
-    We do not know why this is the case.
+def test_edge_cases_with_few_observations_return_default_values():
+    """
+    Test that edge cases with 2 observations return default scaling values.
+    We create an example of a response matrix where all rows are perfectly correlated,
+    which should lead to a single cluster with 1 component and a scaling factor of
+    sqrt(num_observations / num_components) = sqrt(2).
+    However, since the number of observations is <= 2, the function should skip the
+    clustering and PCA and return default values of 1.0 for all scaling factors
     """
     nr_realizations = 1000
+    nr_observations = 2
     Y = np.ones((nr_observations, nr_realizations))
 
     rng = np.random.default_rng(1234)
     parameters_a = rng.normal(10, 1, nr_realizations)
 
+    # create response matrix
     for i in range(nr_observations):
         Y[i] = (i + 1) * parameters_a
 
-    scale_factors, clusters, nr_components = main(Y, Y.mean(axis=1))
+    # Add observation errors
+    obs_errors = np.ones(nr_observations) * 0.1
+
+    scale_factors, *_ = main(Y, obs_errors)
 
     np.testing.assert_equal(
         scale_factors,
-        np.array(nr_observations * [1.0]),
-    )
-
-    np.testing.assert_equal(
-        clusters,
-        np.array(nr_observations * [1.0]),
-    )
-
-    np.testing.assert_equal(
-        nr_components,
         np.array(nr_observations * [1.0]),
     )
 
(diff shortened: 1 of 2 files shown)


tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
```diff
@@ -340,3 +341,192 @@ def test_autoscale_clusters_observations_by_correlation_pattern_ignoring_sign(
...
+def test_that_error_scaling_discards_noisy_observations_in_pca():
+    """
+    Documents current behavior: when scaling responses by observation error
+    (Y / obs_errors), precise observations dominate and noisy observations
+    are effectively discarded by PCA truncation.
+
+    Scenario: 500 noisy seismic amplitudes + 20 precise well pressures,
+    where seismic responds to a shallow parameter and pressure responds
+    to an independent deep parameter.
+
```
lucaeg: Is there a reason it is called shallow and deep parameters, or is the point just that the seismic are correlated with the seismic and the pressures are correlated with the pressures?

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
```diff
@@ -340,3 +341,192 @@ def test_autoscale_clusters_observations_by_correlation_pattern_ignoring_sign(
...
+    # When D-E merges first, A-B should not (they stay separate)
+    if expect_de_merged:
+        assert not is_AB_merged, failure_msg
+
+
+def test_that_error_scaling_discards_noisy_observations_in_pca():
+    """
+    Documents current behavior: when scaling responses by observation error
+    (Y / obs_errors), precise observations dominate and noisy observations
+    are effectively discarded by PCA truncation.
```
lucaeg: Should we explain what this means in terms of the clustering algorithm? I think this explanation is quite technical and will only make sense to people who have very detailed knowledge about the implementation. :) 

I'm not sure who the explanation is intended for.

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
```diff
@@ -340,3 +341,192 @@ def test_autoscale_clusters_observations_by_correlation_pattern_ignoring_sign(
...
+
+    With error-scaling (divide by obs_errors):
+    - Precise pressure measurements get amplified
+    - Noisy seismic measurements get suppressed
+    - Result: 1 PC needed (pressure dominates, seismic info lost)
+
+    With StandardScaler (z-score, unit variance per observation):
+    - Both observation types are treated equally
+    - Result: 2 PCs needed (both independent directions preserved)
+    """
```
lucaeg, line 512: I think the two sections above make sense from a theoretical point of view, but maybe not within the framework of autoscaling. We don't actually use the directions of the PCs anywhere right? However, the number noisy seismic amplitudes vs. precise well pressures does impact the scaling factor in the update algorithm. Maybe we should explain that here.

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
```diff
@@ -340,3 +341,192 @@ def test_autoscale_clusters_observations_by_correlation_pattern_ignoring_sign(
...
+    # Seismic: sensitive to shallow param
+    seismic_sensitivity = rng.uniform(0.5, 1.5, size=(1, n_seismic))
+    seismic_responses = param_shallow @ seismic_sensitivity
+    seismic_responses += rng.normal(0, 0.1, size=(n_realizations, n_seismic))
+
+    # Pressure: sensitive to deep param (independent of seismic)
+    pressure_sensitivity = rng.uniform(0.5, 1.5, size=(1, n_pressure))
+    pressure_responses = param_deep @ pressure_sensitivity
+    pressure_responses += rng.normal(0, 0.1, size=(n_realizations, n_pressure))
+
```
lucaeg, line 547: I think the initialization is a little complicated, I had to read it many times, but it might just be me that is stupid! :D Can the same be achieved in a simpler way?

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
```diff
@@ -340,3 +341,192 @@ def test_autoscale_clusters_observations_by_correlation_pattern_ignoring_sign(
...
+    responses = np.hstack([seismic_responses, pressure_responses])
+
+    # Observation errors: seismic is NOISY, pressure is PRECISE
+    seismic_errors = np.full(n_seismic, 2.0)
+    pressure_errors = np.full(n_pressure, 0.05)
+    obs_errors = np.hstack([seismic_errors, pressure_errors])
+
```
lucaeg, line 554: Good! Very clear :D

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
```diff
@@ -340,3 +341,192 @@ def test_autoscale_clusters_observations_by_correlation_pattern_ignoring_sign(
...
+    assert n_components_error_scaling == 1, (
+        f"Error scaling should yield 1 PC (pressure dominates), "
+        f"got {n_components_error_scaling}"
+    )
+
+    # With standard scaling, both independent directions are visible → 2 PCs needed
+    assert n_components_standard_scaling == 2, (
+        f"Standard scaling should yield 2 PCs (both groups visible), "
+        f"got {n_components_standard_scaling}"
+    )
```
lucaeg: In this last part (from line 510 and onwards), I would try to illustrate the direct behavior of the autoscaler instead, by calling main() and reviewing the results. I think we should make a point out of why it is better for the autoscaler if there are two clusters rather than one.

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
dafeda: My thinking was that "shallow" refers to parameter closer to the top-surface of the reservoir since seismic waves don't penetrate deep into the reservoir. This is of course based on my limited knowledge of these things :)

A pressure sensor can however be placed deep down some well.

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
dafeda: I agree that this might be too technical. Proposals very welcome :)

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
dafeda: I just had to read it a few times myself and I wrote it 😄 
Would it help if we use "*" instead of "@" and add a comment such as this:

I'm just trying to create responses that lead to 2 principal components really.
There's probably a better way to do this...

```python
    # Generate responses using broadcasting to create the correct structure:
    # - Seismic responses = param_shallow * sensitivities creates within-group correlation
    #   (all 500 seismic obs are linear combinations of the same parameter)
    # - Pressure responses = param_deep * sensitivities creates within-group correlation
    #   (all 20 pressure obs are linear combinations of a different parameter)
    # - Since param_shallow and param_deep are independent, this results in exactly
    #   2 principal components needed to explain the variance (one per parameter)

    # Seismic: sensitive to shallow param
    seismic_responses = param_shallow * rng.uniform(0.5, 1.5, size=n_seismic)
    seismic_responses += rng.normal(0, 0.1, size=(n_realizations, n_seismic))

    # Pressure: sensitive to deep param (independent of seismic)
    pressure_responses = param_deep * rng.uniform(0.5, 1.5, size=n_pressure)
    pressure_responses += rng.normal(0, 0.1, size=(n_realizations, n_pressure))
```

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
dafeda: I thought of this as a unit test of `get_nr_primary_components` but we can definitely turn it into more of an integration test if you like.

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
lucaeg: I see! I think if this interplays well with the intuition of reservoir people, then it is nice to have it :)

dafeda: I posted this to a channel with reservoir people and they did not complain :)

lucaeg: great! :D

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
lucaeg: Let me make a PR :)

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
lucaeg: Lets keep the unit test structure then :) I will write the intro in my pr accordingly.

tests/ert/unit_tests/analysis/test_misfit_preprocessor.py
lucaeg: I'm not entirely happy, I will add a suggestion:)

Issue #12633, opened 2026-01-13
Make ErtServer.connect a module function


berland:
This function does not need to be a classmethod of the ErtServer class, as it confuses the responsibility of the ErtServer class. By making it a separate module function, the remainder ErtServer class can be easier named.

https://github.com/equinor/ert/blob/5fe1b2508f508cee5f87c38fc780ffa58b1a57dd/src/ert/services/ert_server.py#L200-L244

PR #12839, opened 2026-02-11
Make Ertserver.connect a module function

eqbech:
Remove the connect classmethod and replace it with a module function instead

**Issue**
Resolves #https://github.com/equinor/ert/issues/12633

### DIFF
--- a/tests/ert/unit_tests/services/test_ert_server.py
+++ b/tests/ert/unit_tests/services/test_ert_server.py
@@ -225,6 +225,28 @@ def test_singleton_start(monkeypatch, server_script, tmp_path):
     assert not (tmp_path / _ERT_SERVER_CONNECTION_INFO_FILE).exists()
 
 
+@pytest.mark.script(
+    """\
+time.sleep(1)
+os.write(fd, b'{"authtoken": "test123", "urls": ["url"]}')
+os.close(fd)
+"""
+)
+def test_that_connect_logs_permission_error(tmp_path, caplog):
+    tmp_path.chmod(0o000)
+    caplog.clear()
+    caplog.set_level("ERROR")
+    with pytest.raises(PermissionError):
+        connect(project=tmp_path, timeout=30)
+
+    tmp_path.chmod(0o755)
+
+    assert len(caplog.records) == 1
+    assert (
+        "cannot connect to ert server service due to permission issues." in caplog.text
+    )
+
+
 @pytest.mark.integration_test
 @pytest.mark.script(
     """\


tests/ert/unit_tests/services/test_ert_server.py
```diff
@@ -224,19 +225,17 @@ def test_singleton_start(monkeypatch, server_script, tmp_path):
...
 os.write(fd, b'{"authtoken": "test123", "urls": ["url"]}')
 os.close(fd)
 """
 )
-def test_singleton_connect(monkeypatch, tmp_path, server_script):
-    monkeypatch.setattr(ert_server, "_ERT_SERVER_EXECUTABLE_FILE", server_script)
-    with _DummyService().start_server(".", timeout=10) as server:
-        client = _DummyService.connect(project=tmp_path, timeout=30)
-        assert server is client
+def test_that_connect_raises_permission_error(tmp_path):
```
berland: Testing that chmod(0) gives a PermissionError is almost the same as testing Python itself. The reason for catching the PermissionError in the code is to log the exception, so the test can be changed to asserting on the log statement (and the name of the test changed accordingly).

eqbech: Updated test to now check the logs.

tests/ert/unit_tests/services/test_ert_server.py
```diff
@@ -224,19 +225,24 @@ def test_singleton_start(monkeypatch, server_script, tmp_path):
...
+    tmp_path.chmod(0o000)
+    caplog.clear()
+    caplog.set_level("ERROR")
+    with pytest.raises(PermissionError):
+        connect(project=tmp_path, timeout=30)
+
+    assert len(caplog.records) == 1
+    assert (
+        "cannot connect to ert server service due to permission issues." in caplog.text
+    )
```
berland, line 247: Should probably also reset the permission to what they were before this test on `tmp_path` to allow pytest to clean up/garbage collect the test directories.

src/ert/services/ert_server.py
```diff
@@ -474,3 +428,41 @@ def create_ertserver_client(project: Path, timeout: int | None = None) -> Client
...
         )
     )
+
+
+def connect(
+    *,
+    project: os.PathLike[str],
+    timeout: int | None = None,
+    logging_config: str | None = None,
+) -> ErtServer:
```
andreas-el: Not your doing, but I feel now that we move from `ErtServer.connect` to just `connect` the name is a bit misleading since we expect to generate an `ErtServer` object.
Does it make sense to rename this to something else?
Perhaps `obtain_ert_server` or something along those lines?

eqbech: I agree, what about `connect_to_ert_server` ?

berland: The function name is misleading yes, but can be fixed in a separate issue. The function name should describe what is actually doing, which is either to connect to an existing, or start a new one if needed.

andreas-el: https://github.com/equinor/ert/issues/12882

PR #12840, opened 2026-02-11
Migrate surfaces over to binary format

### LINKED ISSUE #12824: Writing surfaces in create_run_path(...) can we very slow

Image
Writing irap files can take upwards of 10 minutes. 
Relates to #12766

This is probably a low hanging fruit for some speedup. 

### Steps to reproduce

```console

```

- [ ] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

jonathan-eq:
**Issue**
Resolves #12824 

**Approach**
Before this commit, surfaces are read by ert in ascii format. This commit changes the default to be binary, with the option to override by setting `SURFACE .... FORMAT:ASCII` in config. This will retain the SURFACE old behavior.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/unit_tests/config/test_surface_config.py
+++ b/tests/ert/unit_tests/config/test_surface_config.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import warnings
 from pathlib import Path
 
@@ -11,6 +12,11 @@ from surfio import IrapHeader, IrapSurface
 from ert.config import ConfigValidationError, ConfigWarning, ErtConfig, SurfaceConfig
 from ert.config.parameter_config import InvalidParameterFile
 from ert.config.surface_config import ASCII_SURFACE_WARNING_MESSAGE
+from ert.run_arg import create_run_arguments
+from ert.run_models._create_run_path import create_run_path
+from ert.runpaths import Runpaths
+from ert.sample_prior import sample_prior
+from ert.storage import open_storage
 
 
 @pytest.fixture
@@ -57,7 +63,6 @@ def test_runpath_roundtrip(tmp_path, storage, surface, surface_format, caplog):
         tmp_path / "input_0",
         fformat="irap_ascii" if surface_format == "ascii" else "irap_binary",
     )
-
     # run_path -> storage
     ds = config.read_from_runpath(tmp_path, 0, 0)
     ensemble.save_parameters(ds, config.name, 0)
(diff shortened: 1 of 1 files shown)


src/ert/config/surface_config.py
```diff
@@ -148,7 +168,11 @@ def read_from_runpath(
                 f"'{self.name}' in file {file_name}: "
                 "File not found\n"
             )
-        surface = IrapSurface.from_ascii_file(file_path)
+        surface = (
+            IrapSurface.from_ascii_file(file_path)
+            if self._use_ascii
+            else IrapSurface.from_binary_file(file_path)
+        )
```
jonathan-eq: I added the PrivateAttr with _use_ascii for this. In case the user wants to load from runpath a surface that has ascii format, we should support it.

src/ert/config/surface_config.py
```diff
@@ -59,6 +67,8 @@ class SurfaceConfig(ParameterConfig):
     output_file: Path
     base_surface_path: str
 
+    _use_ascii: bool = PrivateAttr(default=False)
```
jonathan-eq: We need this to support ascii when writing/reading from runpath. It should not be exported to json (I assume) @xjules?.

xjules: I can imagine this being useful when it can be set to ascii. I'd call it differently though: `file_format` or something similar?

src/ert/config/surface_config.py
```diff
@@ -110,14 +128,21 @@ def from_config_list(cls, config_list: list[str | dict[str, str]]) -> Self:
...
+        file_format = SurfaceFileFormat.BINARY
         try:
-            surf = IrapSurface.from_ascii_file(Path(base_surface))
+            try:
+                surf = IrapSurface.from_binary_file(Path(base_surface))
+                logger.info("Loaded surface in binary format")
+            except Exception:
+                surf = IrapSurface.from_ascii_file(Path(base_surface))
+                ConfigWarning.warn(ASCII_SURFACE_WARNING_MESSAGE)
+                file_format = SurfaceFileFormat.ASCII
```
jonathan-eq: The idea here is that if you have binary on the initial validation, then you stick with the binary version when writing/reading from runpath. If you still use ascii, you get the warning, but ert will continue as before, thus not breaking any forward models that still read/write ascii surfaces

src/ert/config/surface_config.py
```diff
@@ -59,6 +75,8 @@ class SurfaceConfig(ParameterConfig):
     output_file: Path
     base_surface_path: str
 
+    _file_format: SurfaceFileFormat = PrivateAttr(default=SurfaceFileFormat.ASCII)
```
jonathan-eq: This defaults to ASCII, as some tests do not create surface_config via `from_config_list` but directly instead

src/ert/config/surface_config.py
```diff
@@ -44,6 +47,19 @@ def __init__(
         super().__init__(msg)
 
 
+ASCII_SURFACE_WARNING_MESSAGE = (
```
frode-aarstad, line 49: We should probably add something about this in the docs aswell

tests/ert/unit_tests/config/test_surface_config.py
```diff
@@ -372,3 +375,44 @@ def surface_for_dl():
         output_file=Path("dummy.txt"),
         base_surface_path="dummy.txt",
     )
+
+
+@pytest.mark.parametrize("is_ascii_surface", [True, False])
```
frode-aarstad, line 396: Could we add a test that tests if we read ascii we also write ascii

jonathan-eq: Yes, added 👌

src/ert/config/surface_config.py
```diff
@@ -44,6 +47,20 @@ def __init__(
...
+    "From this release, the default format for surfaces is changing from ASCII "
```
xjules: I'd remove "From this release" :)
And refactor the text a bit

src/ert/config/surface_config.py
```diff
@@ -44,6 +47,20 @@ def __init__(
         super().__init__(msg)
 
 
+ASCII_SURFACE_WARNING_MESSAGE = (
+    "The default format for surfaces is changing from ASCII "
+    "to binary, but Ert will temporarily still support ASCII surfaces. Binary "
+    "surfaces are recommended, as it is faster to write and read from runpath."
```
xjules: Let's add a sentence that the forward model creating the surfaces need to be changed too to export surfaces in binary format.

jonathan-eq:
image

src/ert/config/surface_config.py
```diff
@@ -133,6 +162,8 @@ def from_config_list(cls, config_list: list[str | dict[str, str]]) -> Self:
             base_surface_path=base_surface,
             update=update_parameter,
         )
+        instance._file_format = file_format
```
oyvindeide: Dont think this will work, because when we read from storage this will never be set, as an example, if you change: https://github.com/equinor/ert/blob/4c1d47412a603ea1bdcf05919ef472acc83f0c09/tests/ert/ui_tests/cli/test_parameter_sample_types.py#L44 to `irap_ascii` I would expect this setting to be: `ascii`, but it is `binary` when we go to read the file from runpath.

src/ert/config/surface_config.py
```diff
@@ -59,6 +78,8 @@ class SurfaceConfig(ParameterConfig):
     output_file: Path
     base_surface_path: str
```
oyvindeide: These file names should be mentioned explicitly in the warning message.

src/ert/config/surface_config.py
xjules: No sure I understand. Base surface creates `surf.irap` that is `binary` and we still gonna read it from runpath as `binary` (by default). So binary all the way.

jonathan-eq: I think this needs more discussion @xjules @oyvindeide 
If I set the default to ASCII, then old storages will be handled. If there is no key for file_format in parameters.json, then it will use ASCII and write file_format: ASCII to parameters.json and that will be picked up on the next reading. If there is a new field already in binary format, that will write file_format: BINARY to parameters.json, and that will also work downstream.

xjules:
Remember to squash the commits.

PR #12844, opened 2026-02-11
Rename ErtServer to ErtServerController

### LINKED ISSUE #12621: Rename ErtServer class
The ErtServer class should be renamed, because its name indicates this class is the server, while it is not. It is a wrapper that can act as a connection  to the server, or start the server if it is not running. This behaviour should be reflected in the class name.

Suggestions:
* ErtServerConnection
* ErtServerHandle

https://github.com/equinor/ert/blob/87e6aa7b745b8690f7ce36ed9245ded6f08fdf0e/src/ert/services/ert_server.py#L39-L97

eqbech:
Rename the ErtServer class to ErtServerController
**Issue**
Resolves #12621


andreas-el (approved):
🥓

PR #12860, opened 2026-02-12
Strategy pattern for updates

### LINKED ISSUE #12851: How to make update parameters customizable
In the long term, we would like to support parameter-wise selection of update algorithms. That is, users should be able to choose different update methods for different parameter types.

For example:
 - For scalar parameters, a user might choose between ENIF, adaptive localization, or ESMDA.
 - For fields and surfaces, a user might choose distance-based localization or adaptive localization.

To move in this direction, we need to revisit both the iterative_ensemble_smoother (IES) package and the `_es_update` implementation in ERT. The goal is to determine the most natural architecture that makes such flexibility straightforward to implement.

In particular:
 - The update-related API should remain as simple and clean as possible.
 - All batching operations should be considered an implementation detail of IES and handled internally within the IES package.

On the other hand, the computation and storage of algorithm-specific auxiliary data (e.g., the H matrix for ENIF, the RHO matrix for distance-based methods, etc.) are currently implemented inside IES. However, maybe we would prefer to move these into ERT(?), wherein we would like to have them as `storage/update-data`.

Thus IES would focus on numerical update mechanics and ERT would manage the update configuration and algorithm-specific data (?).

dafeda:
**Issue**
Resolves #12851 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/analysis/_es_update.py
+++ b/src/ert/analysis/_es_update.py
@@ -4,29 +4,13 @@ import logging
 import re
 import time
 import warnings
-from collections.abc import Callable, Iterable, Sequence
-from typing import TYPE_CHECKING, Generic, Self, TextIO, TypeVar
+from collections.abc import Callable, Iterable
+from typing import TYPE_CHECKING, TextIO
 
-import iterative_ensemble_smoother as ies
 import numpy as np
 import polars as pl
-import psutil
-import scipy
-from iterative_ensemble_smoother.experimental import AdaptiveESMDA, DistanceESMDA
-from iterative_ensemble_smoother.utils import calc_rho_for_2d_grid_layer
 
-from ert.config import (
-    ESSettings,
-    Field,
-    GenKwConfig,
-    ObservationSettings,
-    SurfaceConfig,
-)
-from ert.field_utils import (
-    AxisOrientation,
-    transform_local_ellipse_angle_to_local_coords,
-    transform_positions_to_local_field_coordinates,
-)
+from ert.config import Field, ObservationSettings, SurfaceConfig
 
 from ._update_commons import (
     ErtAnalysisError,
@@ -35,12 +19,19 @@ from ._update_commons import (
     _preprocess_observations_and_responses,
     noop_progress_callback,
 )
+from ._update_strategies import (
+    AdaptiveLocalizationUpdate,
+    DistanceLocalizationUpdate,
+    ObservationContext,
+    ObservationLocations,
+    StandardESUpdate,
+    UpdateStrategy,
+)
 from .event import (
     AnalysisCompleteEvent,
     AnalysisErrorEvent,
     AnalysisEvent,
     AnalysisStatusEvent,
-    AnalysisTimeEvent,
     DataSection,
 )
 from .snapshots import ObservationStatus, SmootherSnapshot
@@ -48,137 +39,15 @@ from .snapshots import ObservationStatus, SmootherSnapshot
 if TYPE_CHECKING:
     import numpy.typing as npt
 
+    from ert.config import ESSettings
     from ert.storage import Ensemble
 
 logger = logging.getLogger(__name__)
 
-T = TypeVar("T")
 
-# When running adaptive localization we reserve some resources for
-# the GUI and other applications that might be running
-RESERVED_CPU_CORES = 2
-# Testing on drogon seems to indicate that this is a
-# reasonable 'default' value for the parallel config of joblib
-NUM_JOBS_ADAPTIVE_LOC = max(
-    1, ((psutil.cpu_count(logical=False) or 1) - RESERVED_CPU_CORES)
-)
-
-
-class TimedIterator(Generic[T]):
-    SEND_FREQUENCY = 1.0  # seconds
-
-    def __init__(
-        self, iterable: Sequence[T], callback: Callable[[AnalysisEvent], None]
-    ) -> None:
-        self._start_time = time.perf_counter()
-        self._iterable = iterable
-        self._callback = callback
-        self._index = 0
-        self._last_send_time = 0.0
-
-    def __iter__(self) -> Self:
-        return self
-
-    def __next__(self) -> T:
-        try:
-            result = self._iterable[self._index]
-        except IndexError as e:
-            raise StopIteration from e
-
-        if self._index != 0:
-            elapsed_time = time.perf_counter() - self._start_time
-            estimated_remaining_time = (elapsed_time / (self._index)) * (
-                len(self._iterable) - self._index
-            )
-            if elapsed_time - self._last_send_time > self.SEND_FREQUENCY:
-                self._callback(
-                    AnalysisTimeEvent(
-                        remaining_time=estimated_remaining_time,
-                        elapsed_time=elapsed_time,
-                    )
-                )
-                self._last_send_time = elapsed_time
-
-        self._index += 1
-        return result
-
-
-def _split_by_batchsize(
-    arr: npt.NDArray[np.int_], batch_size: int
-) -> list[npt.NDArray[np.int_]]:
-    """
-    Splits an array into sub-arrays of a specified batch size.
-
-    Examples
-    --------
-    >>> num_params = 10
-    >>> batch_size = 3
-    >>> s = np.arange(0, num_params)
-    >>> _split_by_batchsize(s, batch_size)
-    [array([0, 1, 2, 3]), array([4, 5, 6]), array([7, 8, 9])]
-
-    >>> num_params = 10
-    >>> batch_size = 10
-    >>> s = np.arange(0, num_params)
-    >>> _split_by_batchsize(s, batch_size)
-    [array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])]
-
-    >>> num_params = 10
-    >>> batch_size = 20
-    >>> s = np.arange(0, num_params)
-    >>> _split_by_batchsize(s, batch_size)
-    [array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])]
-    """
-    sections = 1 if batch_size > len(arr) else len(arr) // batch_size
-    return np.array_split(arr, sections)
-
-
-def _calculate_adaptive_batch_size(num_params: int, num_obs: int) -> int:
-    """Calculate adaptive batch size to optimize memory usage during Adaptive
-    Localization. Adaptive Localization calculates the cross-covariance between
-    parameters and responses. Cross-covariance is a matrix with shape num_params
-    x num_obs which may be larger than memory. Therefore, a batching algorithm is
-    used where only a subset of parameters is used when calculating cross-covariance.
-    This function calculates a batch size that can fit into the available memory,
-    accounting for a safety margin.
-
-    Derivation of formula:
-    ---------------------
-    available_memory = (amount of available memory on system) * memory_safety_factor
-    required_memory = num_params * num_obs * bytes_in_float32
-    num_params = required_memory / (num_obs * bytes_in_float32)
-    We want (required_memory < available_memory) so:
-    num_params < available_memory / (num_obs * bytes_in_float32)
-
-    The available memory is checked using the `psutil` library, which provides
-    information about system memory usage.
-    From `psutil` documentation:
-    - available:
-        the memory that can be given instantly to processes without the
-        system going into swap.
-        This is calculated by summing different memory values depending
-        on the platform and it is supposed to be used to monitor actual
-        memory usage in a cross platform fashion.
-    """
-    available_memory_in_bytes = psutil.virtual_memory().available
-    memory_safety_factor = 0.8
-    # Fields are stored as 32-bit floats.
-    bytes_in_float32 = 4
-    return min(
-        int(
-            np.floor(
-                (available_memory_in_bytes * memory_safety_factor)
-                / (num_obs * bytes_in_float32)
-            )
-        ),
-        num_params,
-    )
-
-
-def analysis_ES(
+def perform_ensemble_update(
     parameters: Iterable[str],
     observations: Iterable[str],
-    rng: np.random.Generator,
     module: ESSettings,
     observation_settings: ObservationSettings,
     global_scaling: float,
@@ -187,16 +56,46 @@ def analysis_ES(
     source_ensemble: Ensemble,
     target_ensemble: Ensemble,
     progress_callback: Callable[[AnalysisEvent], None],
+    strategy_map: dict[str, UpdateStrategy],
 ) -> None:
+    """
+    Orchestrate ensemble-based parameter updates using configurable strategies.
+
+    This function coordinates the parameter update process using a strategy pattern
+    that supports multiple analysis algorithms (ES, IES, etc.). The workflow:
+    1. Preprocessing observations and responses
+    2. Preparing strategies with context data
+    3. Delegating parameter updates to the mapped strategies
+    4. Saving updated parameters to the target ensemble
+
+    Parameters
+    ----------
+    parameters : Iterable[str]
+        Names of parameter groups to update.
+    observations : Iterable[str]
+        Names of observations to use.
+    module : ESSettings
+        ES settings controlling update behavior.
+    observation_settings : ObservationSettings
+        Settings for observation handling.
+    global_scaling : float
+        Global scaling factor for observations.
+    smoother_snapshot : SmootherSnapshot
+        Snapshot object for storing results.
+    ens_mask : npt.NDArray[np.bool_]
+        Boolean mask for active realizations.
+    source_ensemble : Ensemble
+        Source ensemble to read parameters from.
+    target_ensemble : Ensemble
+        Target ensemble to save updated parameters to.
+    progress_callback : Callable[[AnalysisEvent], None]
+        Callback for reporting progress.
+    strategy_map : dict[str, UpdateStrategy]
+        Mapping from parameter group names to update strategies.
+    """
     iens_active_index = np.flatnonzero(ens_mask)
 
-    ensemble_size = ens_mask.sum()
-
-    def adaptive_localization_progress_callback(
-        iterable: Sequence[T],
-    ) -> TimedIterator[T]:
-        return TimedIterator(iterable, progress_callback)
-
+    # Preprocess observations and responses
     preprocessed_data = _preprocess_observations_and_responses(
         ensemble=source_ensemble,
         outlier_settings=observation_settings.outlier_settings,
@@ -211,7 +110,7 @@ def analysis_ES(
         pl.col("status") == ObservationStatus.ACTIVE
     )
 
-    S = filtered_data.select([*map(str, iens_active_index)]).to_numpy(order="c")
+    responses = filtered_data.select([*map(str, iens_active_index)]).to_numpy(order="c")
     observation_values = filtered_data["observations"].to_numpy()
     observation_errors = filtered_data[_OutlierColumns.scaled_std].to_numpy()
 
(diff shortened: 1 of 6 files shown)


src/ert/analysis/_es_update.py
```diff
@@ -188,15 +51,45 @@ def analysis_ES(
...
 ) -> None:
+    """
+    Orchestrate ensemble-based parameter updates using configurable strategies.
+
+    This function coordinates the parameter update process using a strategy pattern
+    that supports multiple analysis algorithms (ES, IES, etc.). The workflow:
+    1. Preprocessing observations and responses
+    2. Creating appropriate update strategies via the factory
+    3. Delegating parameter updates to the selected strategies
+    4. Saving updated parameters to the target ensemble
```
xjules, line 69: Each parameter has an update `update=True|False` flag. Maybe this can be changed to `update=[None,UpdateAlg1, UpdateAlg2, etc.]` which could be used to decide on the initial strategy.

dafeda: Are you envisioning something like this?

```
FIELD COND PARAMETER cond.bgrdecl INIT_FILES:cond.bgrdecl FORWARD_INIT:True UPDATE:distance
GEN_KW INIT_TEMP_SCALE init_temp_prior.txt UPDATE:adaptive
```

xjules: yes, with some reasonable defaults ofc. But it should be update-able from the GUI as well.

dafeda: This might be a good idea, but definitely outside the scope of this PR.

src/ert/analysis/_update_strategies/_factory.py
```diff
@@ -0,0 +1,188 @@
...
+
+        if not self._settings.distance_localization:
+            return None
+
+        # Filter to observations with location data
+        has_location = (
+            filtered_data["east"].is_not_null() & filtered_data["north"].is_not_null()
+        ).to_numpy()
+
+        self._obs_locations = ObservationLocations(
```
xjules: isn't `has_location` as eg. `obs_with_location_idx` enough to store instead the full dataframe?

src/ert/analysis/_es_update.py
```diff
@@ -245,84 +139,31 @@ def adaptive_localization_progress_callback(
...
-            observations=obs_values_with_loc,
-            alpha=1,
-            seed=rng,
-        )
-    elif module.localization:
-        smoother_adaptive_es = AdaptiveESMDA(
-            covariance=observation_errors**2,
-            observations=observation_values,
-            seed=rng,
+        factory.extract_observation_locations(
```
xjules: This can be done regardless; even without the `distance_localization` flag.

xjules: Basically this is all that is needed, which is not a biggy:
```
has_location = (
            filtered_data["east"].is_not_null() & filtered_data["north"].is_not_null()
        ).to_numpy()
```

src/ert/analysis/_es_update.py
```diff
@@ -245,84 +139,31 @@ def adaptive_localization_progress_callback(
...
     )
-    truncation = module.enkf_truncation
 
-    obs_xpos: npt.NDArray[np.float64] | None = None
-    obs_ypos: npt.NDArray[np.float64] | None = None
-    obs_main_range: npt.NDArray[np.float64] | None = None
-    S_with_loc: npt.NDArray[np.float64] | None = None
-    smoother_distance_es: DistanceESMDA | None = None
+    # Create strategy factory and strategies
+    factory = UpdateStrategyFactory(module, smoother_snapshot)
```
xjules: Can we do UpdateStrategyFactory(context, smoother_snapshot) instead? it is anyway supplied in the create_strategies(...)

src/ert/analysis/_update_strategies/_factory.py
```diff
@@ -0,0 +1,188 @@
...
+            List of strategy instances. The order matters - strategies
+            are checked in order for can_handle().
+
+        Raises
+        ------
+        RuntimeError
+            If distance localization is enabled but observation locations
+            were not extracted first.
+        """
+        strategies: list[UpdateStrategy] = []
```
xjules: I think we can provide all strategies as available, the filtering should be via can_handle....
The strategies are obtained from the "config" settings during the actual update, which copies the current ert update scheme - that's what we have today and currently is fine.
The idea here is that either we can just ignore that config and instead provide all strategies, wherein we leave it up the the actual parameter specification (currently not this place but I can see having something like `paramer.update==DistanceESMDA`) and then strategy.can_handle can reason if this makes sense.
Since parameter.update==alg.type is not in place we can use the `settings.distance_localization`, etc. instead.

src/ert/analysis/_update_strategies/_factory.py
```diff
@@ -0,0 +1,188 @@
...
+        UpdateStrategy
+            The first strategy that can handle this parameter type.
+
+        Raises
+        ------
+        ValueError
+            If no strategy can handle the parameter type.
+        """
+        for strategy in strategies:
+            if strategy.can_handle(param_config):
```
xjules: When populating this list of available update alg. for arbitrary parameter, this can be a validator what to list in the selector.
Note: this is more a comment for the UX design (not applicable now!) to populate this:
image

src/ert/analysis/_es_update.py
```diff
@@ -350,161 +190,19 @@ def adaptive_localization_progress_callback(
...
-            param_ensemble_array[non_zero_variance_mask] @= T.astype(
-                param_ensemble_array.dtype
-            )
+        # Delegate update to strategy
+        param_ensemble_array = strategy.update(
+            param_group,
+            param_ensemble_array,
+            param_cfg,
+            non_zero_variance_mask,
+            context,
```
xjules: context is already provided when strategies are created in here:.
To clarify: context is injected here:
```python
strategies = factory.create_strategies(context)
``` 
So I was wondering if we need to provide as an argument here too?

src/ert/analysis/_update_strategies/_distance.py
```diff
@@ -0,0 +1,247 @@
...
+            f"in {(time.time() - start) / 60} minutes"
+        )
+
+        return param_ensemble
+
+
+class DistanceLocalizationFieldUpdate(_DistanceLocalizationBase):
+    """Distance-based localization for Field parameters."""
+
+    def can_handle(self, param_config: ParameterConfig) -> bool:
```
xjules: `if (isinstance(param_config, Field) and param_config.extbox_params.origin is not None)`

src/ert/analysis/_update_strategies/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+    # Create strategies
+    strategies = factory.create_strategies()
+
+    # Initialize all strategies
+    for strategy in strategies:
+        strategy.initialize(context)
+
+    # Update each parameter group
+    for param_group in parameters:
+        strategy = factory.get_strategy_for(param_config, strategies)
```
tommyod: i dont quite understand this part. the docstrings say we get the first available strategy? how is user preference/settings accounted for? shouldnt there be a one-to-one between config and available strategies?

src/ert/analysis/_update_strategies/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+    # For distance localization, extract observation locations first
+    if es_settings.distance_localization:
+        factory.extract_observation_locations(filtered_data, responses)
+
+    # Create strategies
+    strategies = factory.create_strategies()
+
+    # Initialize all strategies
+    for strategy in strategies:
+        strategy.initialize(context)
```
tommyod: why do we need to initialize all strategies? if the user wants to apply strategy A, why not just init A?

src/ert/analysis/_update_strategies/_standard.py
```diff
@@ -0,0 +1,149 @@
...
+
+        # Add identity in place for efficient computation: T = I + K @ H
+        np.fill_diagonal(self._T, self._T.diagonal() + 1)
+
+    def can_handle(self, param_config: ParameterConfig) -> bool:
+        """Check if this strategy handles the parameter type.
+
+        Parameters
+        ----------
+        param_config : ParameterConfig
```
tommyod: i personally dont like docstrings that say nothing. we know the type from the signature. we know that "param_config" is a "Configuration for the parameter." so i would try to cut this down, but it depends on the style elsewhere too of course.

src/ert/analysis/_update_strategies/_distance.py
```diff
@@ -0,0 +1,247 @@
...
+            self._obs_loc.ypos,
+        )
+
+        ellipse_rotation = transform_local_ellipse_angle_to_local_coords(
+            ertbox.rotation_angle,
+            np.zeros_like(self._obs_loc.main_range, dtype=np.float64),
+        )
+
+        rho_matrix = calc_rho_for_2d_grid_layer(
+            ertbox.nx,
```
tommyod, line 127: i suggest calling functions with many arguments with explict parameters names. makes it more clear and reduces the probably of copy paste errors etc

src/ert/analysis/_update_strategies/_adaptive.py
```diff
@@ -0,0 +1,226 @@
...
+    list[npt.NDArray[np.int_]]
+        List of sub-arrays.
+
+    Examples
+    --------
+    >>> num_params = 10
+    >>> batch_size = 3
+    >>> s = np.arange(0, num_params)
+    >>> _split_by_batchsize(s, batch_size)
+    [array([0, 1, 2, 3]), array([4, 5, 6]), array([7, 8, 9])]
```
tommyod, line 60: the desired batch size is 3, but an array of length 4 is returned. that seems strange to me. shouldnt desired batch size be the largets possible? i would expect sizes: 3, 3, 3, 1 (remaining)

src/ert/analysis/_update_strategies/_adaptive.py
```diff
@@ -0,0 +1,226 @@
...
+        Number of observations.
+
+    Returns
+    -------
+    int
+        Batch size that fits in available memory.
+    """
+    available_memory_in_bytes = psutil.virtual_memory().available
+    memory_safety_factor = 0.8
+    bytes_in_float32 = 4
```
tommyod, line 88: dont we use float64?

src/ert/analysis/_update_strategies/__init__.py
dafeda: This is an outdated docstring unfortunately. I've moved initialization into the `__init__` of each strategy.

src/ert/analysis/_update_strategies/__init__.py
dafeda: This PR is not indented to change the behaviour of `ert` in any way, and `ert` does not currently let users choose different algorithms for different parameters.
However, the strategy pattern should make this easier to do down the line.

src/ert/analysis/_update_strategies/_adaptive.py
dafeda: Fields and surfaces in `ert` use float32. Scalars use float64.

src/ert/analysis/_update_strategies/_adaptive.py
dafeda: This does seem a bit strange I agree. We should look into this, but this is "old" code and not that relevant to the refactoring done in this PR I think.

src/ert/analysis/_update_strategies/_factory.py
dafeda: Not sure I understand the question. Could you give me some more info.

src/ert/analysis/_es_update.py
dafeda: Not sure I understand. Are you suggesting we run `extract_observation_locations` even if `module.distance_localization` is `False`?

xjules: yes, exactly.

src/ert/analysis/_update_strategies/_factory.py
xjules: yes, it should be enough to store the indices of observations with locations rather then the full dataframe.

src/ert/analysis/_update_strategies/_distance.py
```diff
@@ -0,0 +1,262 @@
...
+
+        return smoother.update_params(
+            X=param_ensemble,
+            Y=self._obs_loc.responses_with_loc,
+            rho_input=rho_matrix,
+            nz=ertbox.nz,
+        )
+
+
+class DistanceLocalizationSurfaceUpdate(_DistanceLocalizationBase):
```
xjules: Still not sure if we need this level of granularity; ie. DistanceLocalizationSurfaceUpdate and DistanceLocalizationFieldUpdate

src/ert/analysis/_update_strategies/_standard.py
```diff
@@ -0,0 +1,169 @@
...
+    ) -> None:
+        self._smoother_snapshot = smoother_snapshot
+        self._settings = settings
+        self._rng = rng
+        self._progress_callback = progress_callback
+        self._T: npt.NDArray[np.float64] | None = None
+        self._ensemble_size: int = 0
+        self._num_obs: int = 0
+
+    def prepare(self, obs_context: ObservationContext) -> None:
```
oyvindeide, line 78: I think it would be better to explicitly have the needed arguments here instead of `ObservationContext` as we are immediately unpacking it.

src/ert/analysis/_update_strategies/_standard.py
```diff
@@ -0,0 +1,169 @@
...
+    ErtAnalysisError
+        If computing the transition matrix fails due to singular matrix.
+    """
+
+    def __init__(
+        self,
+        smoother_snapshot: SmootherSnapshot,
+        settings: ESSettings,
+        rng: np.random.Generator,
+        progress_callback: Callable[[AnalysisEvent], None],
```
oyvindeide, line 68: Could be a dataclass?

src/ert/analysis/_es_update.py
```diff
@@ -246,84 +145,34 @@ def adaptive_localization_progress_callback(
...
-            raise ErtAnalysisError(msg) from err
-        # Add identity in place for fast computation
-        np.fill_diagonal(T, T.diagonal() + 1)
+    # Prepare each unique strategy once (multiple params may share the same instance)
+    for strategy in set(strategy_map.values()):
+        strategy.prepare(obs_context)
 
+    # Update each parameter group
     for param_group in parameters:
         param_cfg = source_ensemble.experiment.parameter_configuration[param_group]
```
oyvindeide, line 177: Did you look into saving the strategies on the parameter configuration instead? Then we dont need to map them, and can either, just send in parameter configuration instead of a list of parameters, or, just load them from the ensemble.

src/ert/analysis/_update_strategies/_standard.py
```diff
@@ -0,0 +1,169 @@
...
+            raise RuntimeError("prepare() must be called before update()")
+
+        log_msg = (
+            f"There are {self._num_obs} responses "
+            f"and {self._ensemble_size} realizations."
+        )
+        logger.info(log_msg)
+        self._progress_callback(AnalysisStatusEvent(msg=log_msg))
+
+        # In-place multiplication is not yet supported, therefore avoiding @=
```
oyvindeide: Out of date comment?

src/ert/analysis/_update_strategies/_adaptive.py
```diff
@@ -0,0 +1,249 @@
...
+        Pre-computed response covariance matrix (set after prepare()).
+    _D : npt.NDArray[np.float64] | None
+        Perturbed observations matrix (set after prepare()).
+    _num_obs : int
+        Number of observations.
+    """
+
+    def __init__(
+        self,
+        settings: ESSettings,
```
oyvindeide: Only used for threshold, suggest adding that explicitly instead of passing in an object. Could be consider cases where you want different threshold per parameter?

src/ert/analysis/_update_strategies/_protocol.py
```diff
@@ -0,0 +1,199 @@
...
+        obs_context : ObservationContext
+            Preprocessed observation and response data.
+        """
+        ...
+
+    def update(
+        self,
+        param_group: str,
+        param_ensemble: npt.NDArray[np.float64],
+        param_config: ParameterConfig,
```
oyvindeide, line 177: This also contains `param_group`, so dont need both.

src/ert/analysis/_update_strategies/_standard.py
dafeda: Different strategies don't necessarily need the same data though. The `_distance` strategy needs `observation_locations` while the others don't for example.

src/ert/analysis/_update_strategies/_standard.py
dafeda: Are you suggesting we convert all strategies into dataclasses?
If so, what would we gain?

src/ert/analysis/_es_update.py
dafeda: Have not looked into this yet. Could write an issue and do some design etc.

src/ert/analysis/_update_strategies/_standard.py
dafeda: Seems like it yes.

src/ert/analysis/_update_strategies/_adaptive.py
dafeda: Done. Not sure about different thresholds per parameter. Maybe :)

src/ert/analysis/_update_strategies/_protocol.py
dafeda: Nice catch, fixed.

src/ert/analysis/_update_strategies/_standard.py
oyvindeide: > Are you suggesting we convert all strategies into dataclasses?
If so, what would we gain?

Yes, suggest changing them all as there is no logic in `__init__`, so that just complicates it.

dafeda: Not sure if that is a good idea as these strategies are not really ready until we run `prepare()`. To me, dataclasses should "be ready" at construction time if you know what I mean.

src/ert/analysis/_update_strategies/_standard.py
oyvindeide: Which I think would be easier to see if we add them explicitly as arguments, otherwise `ObservationContext` becomes as super-set of all things needed for all future strategies, which I think will complicate the logic.

src/ert/analysis/_update_strategies/_standard.py
oyvindeide: I see your point, but that was not obvious to me when reading the code, then it just read as a unneeded complication.

src/ert/analysis/_update_strategies/_standard.py
dafeda: Strategies implement a protocol though:

```python
class UpdateStrategy(Protocol):
    """Protocol for parameter update strategies.

    Each strategy implements a specific algorithm for updating ensemble
    parameters based on observations. Strategies receive runtime dependencies
    (rng, settings, progress_callback) at construction, and observation data
    via prepare().

    Lifecycle:
        1. Create strategy with dependencies (rng, settings, progress_callback)
        2. Call prepare(obs_context) to initialize with observation data
        3. Call update() for each parameter group
    """

    def prepare(self, obs_context: ObservationContext) -> None:
        """Initialize the strategy with observation data.

        Called once before any update() calls. Performs any expensive
        pre-computation (e.g., computing transition matrices).

        Parameters
        ----------
        obs_context : ObservationContext
            Preprocessed observation and response data.
        """
        ...
```

We don't have to do this of course.

oyvindeide: This could be changed in the future of course, and there is a legitimate question as the whether we should introduce `ObservationContext` and 
```python
class ObservationLocationContext(ObservationContext):
```

or some kind of hierarchy, which also comes with some complexity.

Issue #3359, opened 2022-05-10
Leftover Title Case


berland:
At least this file contains some leftover Title Case strings, should have been removed in #3190 

https://github.com/equinor/ert/blob/7d27de65e4b6de241c30b730f6984a6b2581964a/ert_gui/tools/workflows/run_workflow_widget.py#L39

PR #12867, opened 2026-02-13
Add everest objectives/constraints over iteration plot

yngve-sk:
Tentative PR with Everest plots (naming etc is subject to improvement)

How to test manually:
```
everest run test-data/everest/math_func/config_advanced.yml 
```

Add absolute enspath to `poly.ert`: add the absolute path to `test-data/everest/math_func/everest_output/simulation_results`, by running for example `realpath test-data/everest/math_func/everest_output/simulation_results ` in the terminal.
Then open poly with `ert gui` and open the plotter.

Example poly.ert config:
```
...
ENSPATH /Users/someone/Project/Repositories/ert/test-data/everest/math_func/everest_output/simulation_results
... 
```

Screenshot 2026-02-13 at 14 18 13

controls:
Screenshot 2026-02-13 at 15 08 42

### DIFF
--- a/src/ert/gui/tools/plot/plot_api.py
+++ b/src/ert/gui/tools/plot/plot_api.py
@@ -233,6 +233,14 @@ class PlotApi:
         response_key: str,
         filter_on: dict[str, Any] | None = None,
     ) -> pd.DataFrame:
+        key_def = next(
+            (k for k in self.responses_api_key_defs if k.key == response_key), None
+        )
+        is_everest = key_def is not None and key_def.metadata.get("data_origin") in {
+            "everest_objectives",
+            "everest_constraints",
+        }
+
         if "@" in response_key:
             response_key = response_key.split("@", maxsplit=1)[0]
         with create_ertserver_client(self.ens_path) as client:
@@ -248,6 +256,21 @@ class PlotApi:
 
             stream = io.BytesIO(http_response.content)
             df = pd.read_parquet(stream)
+
+            if is_everest:
+                assert {"batch_id", "realization"}.issubset(df.columns)
+
+                float_columns = [
+                    col for col in df.columns if col not in {"batch_id", "realization"}
+                ]
+
+                return df.astype(
+                    dict.fromkeys(float_columns, float)
+                    | {
+                        "batch_id": int,
+                        "realization": int,
+                    }
+                )
 
             try:
                 df.columns = pd.to_datetime(df.columns, format="%Y-%m-%d %H:%M:%S")
@@ -273,6 +296,9 @@ class PlotApi:
 
             stream = io.BytesIO(http_response.content)
             df = pd.read_parquet(stream)
+
+            if {"batch_id", "realization"}.issubset(df.columns):
+                return df
 
             try:
                 df.columns = pd.to_datetime(df.columns, format="%Y-%m-%d %H:%M:%S")
--- a/src/ert/gui/tools/plot/plot_window.py
+++ b/src/ert/gui/tools/plot/plot_window.py
@@ -43,6 +43,7 @@ from .plottery.plots import (
     MisfitsPlot,
     StatisticsPlot,
     StdDevPlot,
+    ValuesOverIterationsPlot,
 )
 
 CROSS_ENSEMBLE_STATISTICS = "Cross ensemble statistics"
@@ -53,6 +54,8 @@ HISTOGRAM = "Histogram"
 STATISTICS = "Statistics"
 STD_DEV = "Std Dev"
 MISFITS = "Misfits"
+EVEREST_RESPONSES_PLOT = "Batch responses"
+EVEREST_CONTROLS_PLOT = "Batch controls"
 
 RESPONSE_DEFAULT = 0
 GEN_KW_DEFAULT = 3
@@ -174,6 +177,12 @@ class PlotWindow(QMainWindow):
                 self._key_definitions = []
             QApplication.restoreOverrideCursor()
 
+            is_everest = any(
+                k.metadata.get("data_origin")
+                in {"everest_parameters", "everest_objectives", "everest_constraints"}
+                for k in self._key_definitions
+            )
+
             self._plot_customizer = PlotCustomizer(self, self._key_definitions)
             self._plot_customizer.settingsChanged.connect(self.keySelected)
             self._central_tab = QTabWidget()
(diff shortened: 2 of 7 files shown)


berland:
I assume the greyed out tabs will appear on all normal Ert plot GUIs as well with this PR? We should probably not include them in order to not introduce visual noise for Ert users.

src/ert/gui/tools/plot/plottery/plots/values_over_iteration_plot.py
```diff
@@ -0,0 +1,77 @@
...
+                data[value_col],
+                "-o",
+                label=f"Realization {int(realization)}",
+                color=color,
+                markersize=4,
+            )
+
+        axes.set_xlabel("Iteration")
+        axes.set_ylabel(value_col)
+        axes.set_title(f"{value_col} over iterations")
```
andreas-el: Could the title be used to provide a stronger correlation between what you click, and what you see?
I kind of assume that certain values here yield a specific graph, e.g. distance -> "Objective function values" instead of "distance over iterations" ? The y over x information is perhaps a bit redundant.

src/ert/gui/tools/plot/plot_api.py
```diff
@@ -249,6 +257,21 @@ def data_for_response(
             stream = io.BytesIO(http_response.content)
             df = pd.read_parquet(stream)
 
+            if is_everest:
+                assert {"batch_id", "realization"}.issubset(df.columns)
+
+                float_columns = [
+                    col for col in df.columns if col not in {"batch_id", "realization"}
+                ]
```
andreas-el, line 265: Could be made helper function?

src/ert/gui/tools/plot/plot_api.py
```diff
@@ -249,6 +257,21 @@ def data_for_response(
             stream = io.BytesIO(http_response.content)
             df = pd.read_parquet(stream)
 
+            if is_everest:
+                assert {"batch_id", "realization"}.issubset(df.columns)
```
andreas-el, line 261: {"batch_id", "realization"} this set is repeated many places, could extract as constant and allow re-use

src/ert/gui/tools/plot/plot_api.py
```diff
@@ -274,6 +297,9 @@ def data_for_parameter(self, ensemble_id: str, parameter_key: str) -> pd.DataFra
             stream = io.BytesIO(parameter.content)
             df = pd.read_parquet(stream)
 
+            if "batch_id" in df.columns and "realization" in df.columns:
```
andreas-el: if {"batch_id", "realization"}.issubset(df.columns) ?

src/ert/gui/tools/plot/plottery/plots/values_over_iteration_plot.py
yngve-sk: I think I agree, the title can be removed, as the axis labels already say the same information. The user will know whether it is an objective or a control / constraint, we could annotate that into the chart somehow but suggest leaving this a future maybe TODO.

src/ert/gui/tools/plot/plot_api.py
yngve-sk: Hmm. I find 3 usages (after applying proposed fix below) in plot_api.py and one in the values over iteration plot, I don't see the DRY + extra indirection making it more readable in this case to be honest, if we add more usages or somehow specify the "schemas" uniformly somehow here, we might do something.

src/ert/gui/tools/plot/plot_api.py
yngve-sk: Hmm, suggest deferring these kinds of refactorings until more plots have been added and there is a more targeted cleanup at the GUI/plotter code, I think adding a function adds one layer of indirection, making it slightly more convoluted right now. Only other usage of this comprehension is in `values_over_iteration_plot.py`

src/ert/gui/tools/plot/plot_window.py
```diff
@@ -53,6 +54,8 @@
 STATISTICS = "Statistics"
 STD_DEV = "Std Dev"
 MISFITS = "Misfits"
+EVEREST_RESPONSES_PLOT = "Batch Responses"
```
berland: Suggest "Batch responses" and "Batch controls" (ref https://github.com/equinor/ert/issues/3359 )

src/ert/gui/tools/plot/plottery/plots/values_over_iteration_plot.py
```diff
@@ -0,0 +1,76 @@
...
+    import numpy as np
+    import numpy.typing as npt
+    from matplotlib.figure import Figure
+
+    from ert.gui.tools.plot.plot_api import EnsembleObject, PlotApiKeyDefinition
+    from ert.gui.tools.plot.plottery import PlotContext
+
+
+class ValuesOverIterationsPlot:
+    def __init__(self) -> None:
```
berland, line 31: A docstring describing the intention of the plot, and its assumptions on data?

See e.g. pots/misfits.py.

berland:
I think good to go when the png diffs are gone from the diff.

src/ert/gui/tools/plot/plottery/plots/values_over_iteration_plot.py
andreas-el: Seeing this is experimental I guess this is ok. The problem I had was trying to understand what to select to get the same behaviour as in everviz. E.g. how do I view "Objective function values" etc.

src/ert/gui/tools/plot/plot_window.py
```diff
@@ -189,25 +198,40 @@ def __init__(
...
             self._prev_key_dimensionality = -1
-            self._prev_tab_widget_index_map: dict[int, int] = {
-                2: RESPONSE_DEFAULT,
-                1: GEN_KW_DEFAULT,
-                3: STD_DEV_DEFAULT,
-            }
+            self._prev_tab_widget_index_map: dict[int, int] = {}
+            if is_everest:
+                self._prev_tab_widget_index_map = {
+                    1: 0,
```
berland, line 225: Use enums/mnenomics for these, as for `not is_everest`?

yngve-sk: Suggest leaving this for separate issue, it is not immediately clear what the purpose of the things are, but it at least works for the moment.

PR #12874, opened 2026-02-13
Limit ensembles available for restart in ES-MDA panel

### LINKED ISSUE #12704: Cannot restart ES-MDA from last iteration

I noticed that ERT is hanging when trying to rerun ES-MDA from specifically the last iteration.
To recreate, run poly example with es-mda, then try to run a restart-run from the last iteration.

https://github.com/user-attachments/assets/1b0a024b-aa4c-4b51-b44e-7b97ddcc7c25

For it to run as normal, or give some warning if it's not supposed to be possible to restart from last iteration.

### Steps to reproduce

```console

```

- [ ] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

achaikou:
**Issue**
Resolves #12704

**Approach**
Issues is fixed more generally. We do not just prevent experiment from starting, we remove all illegal entries from the restart select.

Notes:
- I believe storage migration is unnecessary, but might be wrong.
- Don't think PR warrants user-impact label as users apparently don't reuse old storage and likely no-one tried to restart from removed entries.
- Also don't consider that a bug, rather improvement.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/gui/experiments/multiple_data_assimilation_panel.py
+++ b/src/ert/gui/experiments/multiple_data_assimilation_panel.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+from collections.abc import Callable
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
 
@@ -26,6 +27,7 @@ from ert.gui.ertwidgets import (
 )
 from ert.mode_definitions import ES_MDA_MODE
 from ert.run_models import MultipleDataAssimilation, MultipleDataAssimilationConfig
+from ert.storage.local_experiment import ExperimentType
 from ert.storage.realization_storage_state import RealizationStorageState
 from ert.validation import (
     ExperimentValidation,
@@ -40,8 +42,11 @@ from ._design_matrix_panel import DesignMatrixPanel
 from .experiment_config_panel import ExperimentConfigPanel
 
 if TYPE_CHECKING:
+    from collections.abc import Iterable
+
     from ert.config import AnalysisConfig
     from ert.gui.ertwidgets import ValueModel
+    from ert.storage import Ensemble
 logger = logging.getLogger(__name__)
 
 
@@ -140,7 +145,40 @@ class MultipleDataAssimilationPanel(ExperimentConfigPanel):
         )
         self._initial_active_realizations = active_realizations
 
-        self._ensemble_selector = EnsembleSelector(notifier)
+        def get_ensembles_that_are_not_last_es_mda_iteration(
+            ensembles: Iterable[Ensemble],
+        ) -> Iterable[Ensemble]:
+            """
+            Only non-leafs of ES-MDA experiment are eligible for restart.
+            Easiest way to get those is to compare ensemble iteration with total
+            number of ES-MDA iterations found via relative weights list length."""
+            return (
+                ensemble
+                for ensemble in ensembles
+                if ensemble.relative_weights
+                and ensemble.iteration < ensemble.relative_weights.count(",") + 1
+            )
+
+        def get_ensembles_of_ensemble_experiment_type(
+            ensembles: Iterable[Ensemble],
+        ) -> Iterable[Ensemble]:
+            """Ensemble experiment type, which consists just from one iteration,
+            is always eligible for MDA "restart". Used to spare some computing
+            time if users decide to run ES-MDA based on Ensemble Experiment
+            results."""
+            return (
+                ensemble
+                for ensemble in ensembles
+                if ensemble.experiment.experiment_type
+                == ExperimentType.ENSEMBLE_EXPERIMENT
+            )
+
+        filters: list[Callable[[Iterable[Ensemble]], Iterable[Ensemble]]] = [
+            get_ensembles_that_are_not_last_es_mda_iteration,
+            get_ensembles_of_ensemble_experiment_type,
+        ]
+
+        self._ensemble_selector = EnsembleSelector(notifier, filters=filters)
         self._previous_ensemble_realizations_validator = EnsembleRealizationsArgument(
             lambda: self._ensemble_selector.selected_ensemble,
             required_realization_storage_states=[
--- a/tests/ert/ui_tests/gui/test_main_window.py
+++ b/tests/ert/ui_tests/gui/test_main_window.py
@@ -744,12 +744,10 @@ def test_right_click_plot_button_opens_external_plotter(qtbot, use_tmpdir, monke
     gui.close()
 
 
-@pytest.mark.usefixtures("copy_poly_case")
-def test_that_es_mda_restart_run_box_is_disabled_when_there_are_no_cases(qtbot):
-    args = Mock()
-    args.config = "poly.ert"
-    gui, *_ = ert.gui.main._start_initial_gui_window(args, GUILogHandler())
-    assert gui.windowTitle().startswith("ERT - poly.ert")
+def test_that_es_mda_restart_run_box_is_disabled_when_there_are_no_valid_cases(
+    qtbot, opened_main_window_minimal_realizations, run_experiment
+):
+    gui = opened_main_window_minimal_realizations
 
     combo_box = get_child(gui, QComboBox, name="experiment_type")
     qtbot.mouseClick(combo_box, Qt.MouseButton.LeftButton)
@@ -766,12 +764,17 @@ def test_that_es_mda_restart_run_box_is_disabled_when_there_are_no_cases(qtbot):
 
     assert restart_button
 
-    assert len(ensemble_selector._ensemble_list()) == 0
+    assert ensemble_selector.count() == 0
     assert not restart_button.isEnabled()
 
     add_experiment_manually(qtbot, gui, ensemble_name="test_ensemble")
-    assert len(ensemble_selector._ensemble_list()) == 1
 
+    assert ensemble_selector.count() == 0
+    assert not restart_button.isEnabled()
+
+    run_experiment(EnsembleExperiment, gui)
+
+    assert ensemble_selector.count() == 1
     assert restart_button.isEnabled()
 
 
--- a/tests/ert/ui_tests/gui/test_restart_esmda.py
+++ b/tests/ert/ui_tests/gui/test_restart_esmda.py
@@ -3,7 +3,7 @@ from PyQt6.QtWidgets import QCheckBox, QComboBox, QWidget
 
 from ert.gui.ertwidgets import StringBox
 from ert.gui.experiments import ExperimentPanel, RunDialog
-from ert.run_models import MultipleDataAssimilation, SingleTestRun
+from ert.run_models import EnsembleExperiment, MultipleDataAssimilation
 
 from .conftest import get_child
 
@@ -39,19 +39,23 @@ def test_restart_esmda(ensemble_experiment_has_run_no_failure, qtbot):
     )
 
 
-def test_active_realizations_esmda(opened_main_window_poly, qtbot):
-    """This runs a single test run and then verifies that this does
-    not interfere with the activate realizations in the es_mda panel
-    unless the restart from that specific ensemble is checked.
+def test_that_esmda_active_realizations_are_set_only_when_restart_is_checked(
+    opened_main_window_poly, qtbot
+):
+    """This runs a experiment and then verifies that this does
+    not interfere with the activate realizations in the es_mda panel unless the
+    restart from that specific ensemble is checked.
     """
     gui = opened_main_window_poly
 
     experiment_panel = get_child(gui, ExperimentPanel)
     simulation_mode_combo = get_child(experiment_panel, QComboBox)
-    simulation_mode_combo.setCurrentText(SingleTestRun.display_name())
+    simulation_mode_combo.setCurrentText(EnsembleExperiment.display_name())
 
-    single_test_run_panel = gui.findChild(QWidget, name="Single_test_run_panel")
-    assert single_test_run_panel
+    ensemble_run_panel = gui.findChild(QWidget, name="Ensemble_experiment_panel")
+    assert ensemble_run_panel
+    ensemble_run_panel._active_realizations_field.setText("0-1")
+
     run_experiment = experiment_panel.findChild(QWidget, name="run_experiment")
     qtbot.mouseClick(run_experiment, Qt.MouseButton.LeftButton)
     qtbot.waitUntil(lambda: gui.findChild(RunDialog) is not None)
@@ -72,7 +76,7 @@ def test_active_realizations_esmda(opened_main_window_poly, qtbot):
     assert restart_checkbox
     assert not restart_checkbox.isChecked()
     restart_checkbox.click()
-    assert active_reals.text() == "0"
+    assert active_reals.text() == "0-1"
     restart_checkbox.click()
     assert active_reals.text() == "0-9"
 
--- a/tests/ert/unit_tests/gui/experiments/conftest.py
+++ b/tests/ert/unit_tests/gui/experiments/conftest.py
@@ -3,7 +3,7 @@ from uuid import uuid4
 
 from ert.storage.local_ensemble import LocalEnsemble
 from ert.storage.local_ensemble import _Index as _EnsembleIndex
-from ert.storage.local_experiment import LocalExperiment
+from ert.storage.local_experiment import ExperimentType, LocalExperiment
 from ert.storage.local_experiment import _Index as _ExperimentIndex
 from ert.storage.local_storage import LocalStorage
 from ert.storage.realization_storage_state import RealizationStorageState
@@ -21,12 +21,14 @@ REALIZATION_FAILED_DURING_EVALUATION = {
 
 
 class MockEnsemble(LocalEnsemble):
-    def __init__(self, ensemble_name, experiment_id, storage_states, storage) -> None:
+    def __init__(
+        self, ensemble_name, experiment_id, storage_states, storage, iteration=0
+    ) -> None:
         self._index = _EnsembleIndex(
             id=uuid4(),
             experiment_id=experiment_id,
             ensemble_size=len(storage_states),
-            iteration=0,
+            iteration=iteration,
             name=ensemble_name,
             prior_ensemble_id=None,
             started_at=datetime.now(),
@@ -39,12 +41,19 @@ class MockEnsemble(LocalEnsemble):
 
 
 class MockExperiment(LocalExperiment):
-    def __init__(self, experiment_name) -> None:
+    def __init__(self, experiment_name, experiment_type) -> None:
         self._index = _ExperimentIndex(id=uuid4(), name=experiment_name, ensembles=[])
+        self._type = experiment_type
 
     @property
     def relative_weights(self) -> str:
-        return "4, 2, 1"
+        if self._type in {ExperimentType.ES_MDA, ExperimentType.UNDEFINED}:
+            return "4, 2, 1"
+        return ""
+
+    @property
+    def experiment_type(self) -> ExperimentType:
+        return self._type
 
 
 class MockStorage(LocalStorage):
@@ -53,14 +62,21 @@ class MockStorage(LocalStorage):
         self._experiments = {}
 
     def _setup_mocked_run(
-        self, ensemble_name, experiment_name, ensemble_states
+        self,
+        ensemble_name,
+        experiment_name,
+        ensemble_states,
+        *,
+        iteration=0,
+        experiment_type=ExperimentType.UNDEFINED,
     ) -> None:
-        mock_experiment = MockExperiment(experiment_name)
+        mock_experiment = MockExperiment(experiment_name, experiment_type)
         mock_ensemble2 = MockEnsemble(
             ensemble_name,
             experiment_id=mock_experiment.id,
             storage_states=ensemble_states,
             storage=self,
+            iteration=iteration,
         )
         self._ensembles[mock_ensemble2.id] = mock_ensemble2
         self._experiments[mock_experiment.id] = mock_experiment
--- a/tests/ert/unit_tests/gui/experiments/test_multiple_data_assimilation_panel.py
+++ b/tests/ert/unit_tests/gui/experiments/test_multiple_data_assimilation_panel.py
@@ -1,4 +1,5 @@
 import math
+from dataclasses import dataclass
 from unittest.mock import MagicMock
 
 import pytest
@@ -19,6 +20,7 @@ from ert.gui.ertwidgets import EnsembleSelector, StringBox
 from ert.gui.experiments.multiple_data_assimilation_panel import (
     MultipleDataAssimilationPanel,
 )
+from ert.storage.local_experiment import ExperimentType
 
 from .conftest import (
     REALIZATION_FINISHED_SUCCESSFULLY,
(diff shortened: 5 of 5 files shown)


src/ert/storage/local_experiment.py
```diff
@@ -52,6 +52,19 @@ class ExperimentState(StrEnum):
     never_run = auto()
 
 
+class ExperimentType(StrEnum):
+    UNDEFINED = "Undefined"
+    TEST = "Single Test Run"
```
SAKavli: I think it would be an improvement for readability if the enum names were more elaborative. 
For example, instead of `ExperimentType.ENSEMBLE` -> `ExperimentType.ENSEMBLE_EXPERIMENT`.
The words `ENSEMBLE`, `TEST` and `EVALUATE` have other meanings depending on the context. 
I think it is just fine to use the full name of experiment types.

src/ert/storage/local_experiment.py
```diff
@@ -52,6 +52,19 @@ class ExperimentState(StrEnum):
...
+    ENSEMBLE = "Ensemble Experiment"
+    EVALUATE = "Evaluate Ensemble"
+    ES_MDA = "Multiple Data Assimilation"
+    ES = "Ensemble Smoother"
+    EnIF = "Ensemble Information Filter"
+    MANUAL_UPDATE = "Manual Update"
+    MANUAL = "Manual"
```
SAKavli, line 64: Is this the same as `Manual Update`?

src/ert/storage/local_experiment.py
```diff
@@ -52,6 +52,19 @@ class ExperimentState(StrEnum):
     never_run = auto()
 
 
+class ExperimentType(StrEnum):
+    UNDEFINED = "Undefined"
+    TEST = "Single Test Run"
+    ENSEMBLE = "Ensemble Experiment"
+    EVALUATE = "Evaluate Ensemble"
+    ES_MDA = "Multiple Data Assimilation"
```
SAKavli, line 60: The short for of ES_MDA is just fine, that is the name we usually use for this run model.

src/ert/gui/ertwidgets/ensembleselector.py
```diff
@@ -22,30 +22,38 @@
...
     ensemble_populated = Signal()
 
     def __init__(
         self,
         notifier: ErtNotifier,
         update_ert: bool = True,
-        show_only_undefined: bool = False,
-        show_only_no_children: bool = False,
-        show_only_with_response_data: bool = False,
+        filters: list[Callable[[Iterable[Ensemble]], Iterable[Ensemble]]] | None = None,
```
SAKavli: Would it work if filters defaults to `= Field(default_factory=list)`?

src/ert/gui/ertwidgets/ensembleselector.py
```diff
@@ -108,28 +116,17 @@ def populate(self) -> None:
...
-            ]
-        return self.sort_ensembles(ensemble_list)
+        if not self._or_filters:
+            return self.sort_ensembles(self.notifier.storage.ensembles)
+
+        all_ensembles = list(self.notifier.storage.ensembles)
+        filtered_ensembles = []
+
+        for filter_func in self._or_filters:
+            filtered_ensembles.extend(list(filter_func(all_ensembles)))
```
SAKavli, line 123: That's a really clean improvement 😎

src/ert/gui/experiments/evaluate_ensemble_panel.py
```diff
@@ -43,7 +44,24 @@ def __init__(self, run_path: str, notifier: ErtNotifier) -> None:
...
+            parents = [
+                ens.parent for ens in self.notifier.storage.ensembles if ens.parent
+            ]
+            return (ensemble for ensemble in ensembles if ensemble.id not in parents)
+
+        # Filter out any ensembles which have children.
+        # One use case is if a user wants to rerun because of failures
+        # not related to parameterization. We can allow that, but only
+        # if the ensemble has not been used in an update, as that would
+        # invalidate the result
```
SAKavli, line 60: Will this not allow reruns from any iterations after the first? 
My impression of the issue was initially that we should only disallow runs from the last iteration. 
But perhaps it is me who have misunderstood something.

tests/ert/unit_tests/gui/ertwidgets/test_ensembleselector.py
```diff
@@ -110,28 +110,49 @@ def test_ensembles_are_sorted_failed_first_then_by_start_time(storage):
...
+                lambda ensembles: (e for e in ensembles if e.parent is None),
+            ],
+            1,
+            id="duplicating filters",
+        ),
     ],
 )
-def test_show_only_no_parent(
-    qtbot, notifier, storage, uniform_parameter, response, flag, expected
+def test_that_filters_are_applied(
```
SAKavli: We should elaborate this test title more, even if the original title was not great.

tests/ert/unit_tests/gui/experiments/test_evaluate_ensemble_panel.py
```diff
@@ -19,3 +27,48 @@ def test_that_the_run_experiments_button_is_disabled(qtbot):
     experiment_panel = get_child(gui, ExperimentPanel)
     evaluate_ensemble_panel = get_child(experiment_panel, EvaluateEnsemblePanel)
     assert not evaluate_ensemble_panel.isConfigurationValid()
+
+
+def test_that_ensemble_select_contains_only_leaf_ensembles(
```
SAKavli, line 32: I am a little confused by this test, isn't the expected behavior that all ensembles except the leaf (aka iteration 3) are possible to rerun from?
I interpret this test tests that the selector only contains the leaf.
But the behavior is correct in the PR. Perhaps you could explain for me.

src/ert/gui/tools/manage_experiments/storage_widget.py
```diff
@@ -190,6 +191,7 @@ def _addItem(self) -> None:
                                 for d in self._ert_config.observation_declarations
                             ],
                             "ert_templates": self._ert_config.ert_templates,
+                            "experiment_type": ExperimentType.MANUAL,
```
SAKavli, line 194: Could you explain why we add `MANUAL` here?

src/ert/gui/simulation/multiple_data_assimilation_panel.py
```diff
@@ -137,7 +142,38 @@ def __init__(
         )
         self._initial_active_realizations = active_realizations
 
-        self._ensemble_selector = EnsembleSelector(notifier)
+        def get_ensembles_that_are_not_last_es_mda_iteration(
+            ensembles: Iterable[Ensemble],
+        ) -> Iterable[Ensemble]:
+            """
+            Only leafs of ES-MDA experiment are eligible for restart. Easiest
```
SAKavli: Is it the other way around, that leaves are the only not eligible for restart?

src/ert/gui/experiments/multiple_data_assimilation_panel.py
```diff
@@ -137,7 +142,38 @@ def __init__(
...
+            ensembles: Iterable[Ensemble],
+        ) -> Iterable[Ensemble]:
+            """
+            Only leafs of ES-MDA experiment are eligible for restart. Easiest
+            way to get those is to compare ensemble iteration with total number
+            of ES-MDA iterations."""
+            return (
+                ensemble
+                for ensemble in ensembles
+                if ensemble.relative_weights
```
SAKavli, line 158: Should the relative_weights also be mentioned in the docstring? Although, what is this trying to check? Out of ensembles 0-3, which has weights and which doesn't?

src/ert/gui/simulation/multiple_data_assimilation_panel.py
```diff
@@ -137,7 +142,38 @@ def __init__(
...
+            way to get those is to compare ensemble iteration with total number
+            of ES-MDA iterations."""
+            return (
+                ensemble
+                for ensemble in ensembles
+                if ensemble.relative_weights
+                and ensemble.iteration < ensemble.relative_weights.count(",") + 1
+            )
+
+        def get_ensembles_of_ensemble_type(
```
SAKavli: If we change the StrEnum, this should probably change too.
`get_ensembles_of_ensemble_experiment_type` or along those lines

tests/ert/ui_tests/gui/test_main_window.py
```diff
@@ -745,12 +745,10 @@ def right_click_plotter_button() -> None:
...
-@pytest.mark.usefixtures("copy_poly_case")
-def test_that_es_mda_restart_run_box_is_disabled_when_there_are_no_cases(qtbot):
-    args = Mock()
-    args.config = "poly.ert"
-    gui, *_ = ert.gui.main._start_initial_gui_window(args, GUILogHandler())
-    assert gui.windowTitle().startswith("ERT - poly.ert")
+def test_that_es_mda_restart_run_box_is_disabled_when_there_are_no_valid_cases(
+    qtbot, opened_main_window_poly, run_experiment
+):
+    gui = opened_main_window_poly
```
SAKavli: It's really great that you find and use these fixtures! 🫶

tests/ert/ui_tests/gui/test_main_window.py
```diff
@@ -771,6 +769,11 @@ def test_that_es_mda_restart_run_box_is_disabled_when_there_are_no_cases(qtbot):
     assert not restart_button.isEnabled()
 
     add_experiment_manually(qtbot, gui, ensemble_name="test_ensemble")
+
+    assert len(ensemble_selector._ensemble_list()) == 0
+    assert not restart_button.isEnabled()
+
+    run_experiment(EnsembleExperiment, gui)
```
SAKavli, line 775: This test used to take ~1.5 seconds, but now it's 18 seconds (locally on my computer).
It is likely due to running 100 realizations for a whole experiment.
I think we should do some tricks here to avoid the time increase, perhaps mocking ensembles.

tests/ert/unit_tests/gui/simulation/test_multiple_data_assimilation_panel.py
```diff
@@ -128,3 +133,101 @@ def test_multiple_data_assimilation_panel_sets_active_realizations_to_initial_ac
...
+            [
+                "experiment_exp : ensemble_ens",
+                "mda_exp : a_ens",
+                "mda_exp : b_ens",
+            ],
+            id="both ensemble runs and old-storage ES_MDA non-leafs",
+        ),
+    ],
+)
+def test_that_restart_ensemble_select_contains(
```
SAKavli: Is this test title cut short?

src/ert/gui/ertwidgets/ensembleselector.py
SAKavli: Aka farther down it defaults to `[]` if `filters is None`

achaikou: Thanks, never seen this one before!

Though apparently `Field` is supposed to work only in relation to pydantic `BasicModel` and I am honesty confused yet when and how we are supposed to use pydantic 😄 
So I think I can't use what you suggest here, because it is just a function parameter list, not `BasicModel` or `dataclass`. At least something complains 😄 

`None`, as the most standard solution for "Mutable Default Arguments" problem, originated from my earliest design where I still modified filters. But a tuple instead of list should work now because I am just iterating over filters, not modifying them. 
So changed to `()`

src/ert/gui/experiments/evaluate_ensemble_panel.py
achaikou: Note that this is the comment for `EvaluateEnsemblePanel` that I moved here from `EnsembleSelector` init. It has nothing to do with the issue or ES-MDA, just with refactoring 🙂 

I think for "evaluate ensemble" ensemble should not be re-evaluated if any other ensemble depends on it, so all good.

tests/ert/unit_tests/gui/ertwidgets/test_ensembleselector.py
achaikou: Do you have any title suggestions? Because I can't find a better one :smile: 

I was struggling with this and decided that the only thing this test does is, well, test that whatever filters you provide to ensemble selector are applied inside of it.
As we are already in the `test_ensemble_selector` file, `test_that_filters_are_applied` made most sense to me. 🤷‍♀️ 

This is a different test from `test_show_only_no_parent` btw, it just lies in place of the old test.

tests/ert/unit_tests/gui/experiments/test_evaluate_ensemble_panel.py
achaikou: Surely! 🐱 
Note that this is the test inside `test_evaluate_ensemble_panel`, so it has nothing to do with ES-MDA.

This is actually a follower of the `test_show_only_no_parent` from `test_ensembleselector.py` . As I removed `show_only_no_children` parameter, I had to move the test to actual panel tests as filter-under-test became integral part of the panel. Necessary rewrite to keep the idea of the test had to follow.

----

Btw, in case it added to confusion. I think previous naming scheme deciphers to:
- "show_only_no_children" means "show only entries with no children".
- "show_only_no_parent" means "show only entries that are not parents"

Both mean the same thing.

src/ert/gui/tools/manage_experiments/storage_widget.py
achaikou: For consistency, every experiment that we create in the production storage, should have a type.
All experiments that stem from `RunModel` will have it defined.

For whatever reason, "Manual" experiment doesn't belong to the hierarchy, so configuration that is serialized by `RunModel` must be added here manually.

src/ert/storage/local_experiment.py
achaikou: No.
This is that weird experiment that you can create by going to `Manage Experiments` and clicking `+` button at the bottom.

src/ert/storage/local_experiment.py
achaikou: I think I made those shorter because during my adventure with this PR long names lead to uglier formatting which I didn't like 😄 .

But OK, updating the names and pretending with all my strength that I do not care about resulting line-breaks 😄

tests/ert/unit_tests/gui/simulation/test_multiple_data_assimilation_panel.py
achaikou: No, I think it is fine.
I used here an advice from Eivind (at least how I understood it): test name + id should make something nice. In this case

> 
> test_that_restart_ensemble_select_contains[ES-MDA which are not leafs in current ES_MDA run]
> test_that_restart_ensemble_select_contains[both ensemble runs and old-storage ES_MDA non-leafs]

src/ert/gui/simulation/multiple_data_assimilation_panel.py
achaikou: 🤦‍♀️ Of course.
Thanks for catching that!

src/ert/gui/experiments/multiple_data_assimilation_panel.py
achaikou: Updated docs for clarity (though not sure if it actually clarifies enough?..).

Yes and no.
- Only ES_MDA experiments have weights, so non-empty relative weights define experiment as ES-MDA.
- `ensemble.relative_weights` is a shortcut for `ensemble.experiment.relative_weights`, so all ensembles in the same ES-MDA experiment will have the same weights.
- Weights are always written as `a, b, c, d, e, f`, so from weights list we can find total number of iterations performed.
- Comparing ensemble iteration number with total number of iterations we can find out if that is a leaf or not. Just checking if ensemble is a parent does not help because leaf ES-MDA can still be a parent in another experiment.

tests/ert/ui_tests/gui/test_main_window.py
achaikou: Ouch, you are right, thanks. 👍  I will try to remember checking test time in the future.

Though there are just 10 realizations running, for me execution time also goes from 1 to 7 seconds.

So I see two options:

1. use a different fixture: `opened_main_window_minimal_realizations`. Then using `run_experiment(EnsembleExperiment, gui)` takes test execution time down to around 4 seconds for me. This is a cleaner solution.

2. Manually update the storage and notify. This brings test performance back to where it was, but forces gui test dig into code internals, which should be in my opinion avoided as much as possible - 👎 .

```
    config = ErtConfig.from_file("poly.ert")
    with open_storage(config.ens_path, mode="w") as storage:
        storage.create_experiment(
            experiment_config={
                "experiment_type": "Ensemble Experiment"
            },
            name="my-experiment",
        ).create_ensemble(
            ensemble_size=config.runpath_config.num_realizations,
            name="my-ensemble",
        )
    gui.notifier.emitErtChange()

```
So I am torn. Faking storage here might be acceptable because it is an unimportant prerequisite, but it is supposed to be a ui test! It is testing main_window, so it closest thing to e2e that we have...

So do we have guidance about the balance between "fidelity" and "tests speed" that we are striving to achieve in `ui_tests` (what Eivind was mentioning recently)?

I for now went with option 1.

tests/ert/unit_tests/gui/simulation/test_multiple_data_assimilation_panel.py
eivindjahren: Maybe to make both read ok we can have 
```python
test_that_restart_ensemble_select_contains_the_element[ES-MDA ensemble when not a leaf in the current ES_MDA run]
```

tests/ert/unit_tests/gui/ertwidgets/test_ensembleselector.py
eivindjahren: Quite often a title becomes clear when we write out the parametrize part:

```python
@pytest.fixture
def storage_with_three_ensembles(storage, notifier):
    ensemble_a = storage.create_experiment().create_ensemble(name="a", ensemble_size=1)
    ensemble_b = storage.create_experiment().create_ensemble(
        name="b", ensemble_size=1, prior_ensemble=ensemble_a
    )
    storage.create_experiment().create_ensemble(
        name="c", ensemble_size=1, prior_ensemble=ensemble_b
    )
    notifier.set_storage(str(storage.path))

def test_that_when_ensemble_selector_filters_is_none_then_all_ensembles_are_selected(
    qtbot, notifier, storage_with_three_ensembles
):
    widget = EnsembleSelector(notifier, filters=None)
    qtbot.addWidget(widget)
    assert widget.count() == 3

def test_that_when_ensemble_selector_filters_are_empty_then_all_ensembles_are_selected(
    qtbot, notifier, storage_with_three_ensembles
):
    widget = EnsembleSelector(notifier, filters=[])
    qtbot.addWidget(widget)
    assert widget.count() == 3

def test_that_ensemble_selector_filters_ensembles_from_storage(
    qtbot, notifier, storage_with_three_ensembles
):
    widget = EnsembleSelector(notifier, filters=[
                lambda ensembles: (e for e in ensembles if e.name == "b"),
                lambda ensembles: (e for e in ensembles if e.name == "c"),
            ]
    )
    qtbot.addWidget(widget)
    assert widget.count() == 2
```

This is just a suggestion, so feel free to just keep it the way it is :)

achaikou: Thanks! That's a new way to look at it.
Agree that names give clarity. 👍 
Don't really like some duplication and decoupling (it is way easier to see that tests actually test the same feature when they are parametrized), but I think in this particular case it is worth it.

src/ert/gui/ertwidgets/ensembleselector.py
SAKavli: Ah, I rarely see the default tuple value, but that might work! What do you think of:
`tuple()` instead of `()` for readability?

tests/ert/unit_tests/gui/ertwidgets/test_ensembleselector.py
SAKavli: I like the tests the way you made them now! 🎉 
I think its nice when a test break that you get an elaborative title of the specific use case instead of a more general parameterized test, where one of the test cases failed but the title doesn't give a hint of which.

src/ert/gui/experiments/multiple_data_assimilation_panel.py
SAKavli: Ah, so all ensembles in an ES-MDA experiment are always loaded? Then I think it makes sense! 👍

src/ert/gui/simulation/multiple_data_assimilation_panel.py
```diff
@@ -137,7 +142,40 @@ def __init__(
...
+            Easiest way to get those is to compare ensemble iteration with total
+            number of ES-MDA iterations found via relative weights list length."""
+            return (
+                ensemble
+                for ensemble in ensembles
+                if ensemble.relative_weights
+                and ensemble.iteration < ensemble.relative_weights.count(",") + 1
+            )
+
+        def get_ensembles_of_ensemble__experiment_type(
```
SAKavli: Double underscore

tests/ert/ui_tests/gui/test_main_window.py
SAKavli: I agree that e2e tests should not skip important steps.
But I would argue that this test with title
`test_that_es_mda_restart_run_box_is_disabled_when_there_are_no_valid_cases`
has a scope which begins in `storage with no valid cases` and ends in `Assert rerun box is disabled`.
I think we (should) have own tests testing that es-mda produces expected results, and tests like this shouldn't worry about experiment execution.

We could always discuss this with Eivind if we still disagree, these are useful discussions!

tests/ert/unit_tests/gui/simulation/test_multiple_data_assimilation_panel.py
SAKavli: I see!

tests/ert/unit_tests/gui/ertwidgets/test_ensembleselector.py
eivindjahren: I find that you have to strike a balance with the `parametrize` option. You can do
```python
@pytest.mark.parametrize(...)
def test_...(computed, expected):
   assert computed == expected
```
which removes all duplication but is hiding all the relevant details. However, if every detail is elaborated on then its hard to see what is relevant (see for instance `test_fm_dispatch_run_subset_specified_as_parameter`). What we are after is essentially pieces that fit inside a humans head. 

It's like looking for your gloves in the drawers of a commode, too small compartments then you can't remember which one it is in, if there is only one drawer its one big mess. We want a reasonable amount of drawers with good labels on them.

eivindjahren:
I think we can aim for high fidelity and keep things at a high level inside `tests/ert/ui_tests`, so I wouldn't worry too much about tens of seconds there. Rather, I would that we have a relatively small amount of high fidelity tests there and a large amount of fast tests in `tests/ert/unit_tests`.

src/ert/storage/local_experiment.py
SAKavli: Ahaaa, never seen this before 😆 👍

src/ert/gui/ertwidgets/ensembleselector.py
achaikou: Someone is not happy 😄 

```
Unnecessary `tuple()` call (rewrite as a literal)Ruff[C408]
```

So leaving as is.

tests/ert/ui_tests/gui/test_main_window.py
achaikou: I'll follow Eivind's advice and keep it like it is for now.
This test wasn't written by me (I am just adjusting it to my changes), so as in doubt I think I can leave it where it is.

But in the future would pay attention to limiting `ui_tests` to main flows only. :+1:

PR #12881, opened 2026-02-16
Fix eclbase name

### LINKED ISSUE #12823: Fix eclbase_format in ModelConfig to allow to be None
Currently even-though we don't use eclipse it is still defaulted to:
https://github.com/equinor/ert/blob/975e418be4a8d4f9662e0e53c093ca09d200bf6a/src/ert/config/model_config.py#L38

 - [ ] variable `eclbase_format_string` should be renamed `summary_file_base_name`
 - [ ] it should default to None

The config option should still use ECLBASE though for reading the value.
The relates to the usage of `jobname_format_string` which should be updated accordingly

frode-aarstad:
**Issue**
Resolves #12823 

**Approach**

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/run_models/everest_run_model.py
+++ b/src/ert/run_models/everest_run_model.py
@@ -424,7 +424,7 @@ class EverestRunModel(RunModel, EverestRunModelConfig):
             "<CONFIG_FILE>": Path(everest_config.config_file).stem,
         }
 
-        if runpath_config.summary_file_base_name:
+        if runpath_config.summary_file_base_name is not None:
             substitutions.update(
                 {
                     "<ECL_BASE>": runpath_config.summary_file_base_name,
--- a/src/ert/config/ert_config.py
+++ b/src/ert/config/ert_config.py
@@ -906,7 +906,7 @@ class ErtConfig(BaseModel):
             runpath = model_config.runpath_format_string
             summary_file_base_name = model_config.summary_file_base_name
             substitutions["<RUNPATH>"] = runpath
-            if summary_file_base_name:
+            if summary_file_base_name is not None:
                 substitutions["<ECL_BASE>"] = summary_file_base_name
                 substitutions["<ECLBASE>"] = summary_file_base_name
             else:


src/ert/config/model_config.py
```diff
@@ -112,9 +114,9 @@ def from_dict(cls, config_dict: ConfigDict) -> ModelConfig:
                     config_dict.get(ConfigKeys.ECLBASE, DEFAULT_JOBNAME_FORMAT)
                 ),
             ),
-            eclbase_format_string=config_dict.get(
+            summary_file_base_name=config_dict.get(
                 ConfigKeys.ECLBASE,
-                config_dict.get(ConfigKeys.JOBNAME, DEFAULT_ECLBASE_FORMAT),
+                config_dict.get(ConfigKeys.JOBNAME),
             ),
```
frode-aarstad, line 120: Not sure if this is correct. Should it still default to DEFAULT_ECLBASE_FORMAT if there are no JOBNAME or should we use None then

jonathan-eq: Hmm, good question. I think the issue wants us to use None.

src/ert/run_models/everest_run_model.py
```diff
@@ -419,13 +419,26 @@ def create(
...
             "<RUNPATH_FILE>": str(runpath_file),
             "<RUNPATH>": runpath_config.runpath_format_string,
-            "<ECL_BASE>": runpath_config.eclbase_format_string,
-            "<ECLBASE>": runpath_config.eclbase_format_string,
             "<NUM_CPU>": str(queue_config.queue_options.num_cpu),
             "<CONFIG_PATH>": everest_config.config_directory,
             "<CONFIG_FILE>": Path(everest_config.config_file).stem,
         }
 
+        if runpath_config.summary_file_base_name:
```
jonathan-eq: I am not sure what will happen if we pass an empty string (still falsy) here. Is that a possibility?

src/ert/run_models/everest_run_model.py
```diff
@@ -1229,7 +1242,7 @@ def _get_run_args(
             runpath_format=self.runpath_config.runpath_format_string,
             filename=str(self.runpath_file),
             substitutions=substitutions,
-            eclbase=self.runpath_config.eclbase_format_string,
+            eclbase=self.runpath_config.summary_file_base_name,
```
jonathan-eq, line 1245: Can we change the eclbase keyword here to summary?

jonathan-eq: (if that is not outside of the scope of this pr)

jonathan-eq (commented):
Some small questions, then we will be ready to push to main 🫨

src/ert/run_models/everest_run_model.py
frode-aarstad: Good point. Ill fix to check for not None

src/ert/run_models/everest_run_model.py
frode-aarstad: I will leave this for another PR

PR #12887, opened 2026-02-16
Rename connect function

### LINKED ISSUE #12882: Rename `connect` function name to reflect what the function does
```
def connect(
    *,
    project: os.PathLike[str],
    timeout: int | None = None,
    logging_config: str | None = None,
) -> ErtServer:
```

eqbech:
connect function has been renamed to imply that it returns a new instance of an ErtServerController

**Issue**
Resolves #12882

### DIFF
--- a/src/ert/services/ert_server.py
+++ b/src/ert/services/ert_server.py
@@ -277,12 +277,12 @@ class ErtServerController:
         logging_config: str | None = None,
     ) -> ErtServerContext:
         try:
-            service = connect(
+            controller = create_ert_server_controller(
                 project=project or Path.cwd(), timeout=0, logging_config=logging_config
             )
             # Check the server is up and running
-            _ = service.fetch_url()
-            return ErtServerContext(service)
+            _ = controller.fetch_url()
+            return ErtServerContext(controller)
         except (TimeoutError, json.JSONDecodeError, KeyError) as e:
             logging.getLogger(__name__).warning(
                 "Failed locating existing storage service due to "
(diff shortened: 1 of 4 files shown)


src/ert/services/ert_server.py
berland: This is currently the best description of what this function does. At best, we are able to encode all or almost all information in this sentence into the function name, and then we can drop the docstring.

berland: Or, take a new look, is this docstring even correct here?

eqbech: Ok, let me know if any of these options sounds good:

- `connect_or_start_ert_server_controller`
- `connect_or_create...`
- `get_or_start...`
- `get_or_create...`
- `ensure...`

tests/ert/unit_tests/services/test_ert_server.py
```diff
@@ -269,7 +269,7 @@ class ClientThread(threading.Thread):
         def run(self):
             start_event.set()
             try:
-                self.client = connect(project=tmp_path, timeout=30)
+                self.client = create_ert_server_controller(project=tmp_path, timeout=30)
```
andreas-el: We should perhaps create a new issue here. Now we are mixing `client` with `server_controller`.

src/ert/services/ert_server.py
```diff
@@ -277,7 +277,7 @@ def init_service(
         logging_config: str | None = None,
     ) -> ErtServerContext:
         try:
-            service = connect(
+            service = create_ert_server_controller(
```
andreas-el: This is a service, or server_controller?

src/ert/services/ert_server.py
```diff
@@ -419,7 +419,7 @@ def wait(self) -> None:
 
 def create_ertserver_client(project: Path, timeout: int | None = None) -> Client:
     """Read connection info from file in path and create HTTP client."""
-    connection = connect(timeout=timeout, project=project)
+    connection = create_ert_server_controller(timeout=timeout, project=project)
```
andreas-el: Not a service? Now a connection. 🤔

src/ert/services/ert_server.py
eqbech: > Or, take a new look, is this docstring even correct here?

No, the docstring is outdated from when it was a classmethod, it no longer recieves self, and checks if it exists, it now only creates a new instance of ErtServerController.

tests/ert/unit_tests/services/test_ert_server.py
andreas-el: And `connection`, and `service`. This is totally random!

src/ert/services/ert_server.py
andreas-el: The docstring should go away imo.

The code definitely creates, obtains or generates something. Regardless of prior existence or not.

tests/ert/unit_tests/services/test_ert_server.py
eqbech: Updated, now to share same semantics by calling it a controller.

PR #12892, opened 2026-02-16
Add everest gradient plots

yngve-sk:
Approach:
Screenshot 2026-02-16 at 15 42 48

### DIFF
--- a/src/ert/gui/tools/plot/plot_window.py
+++ b/src/ert/gui/tools/plot/plot_window.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+from concurrent.futures import ThreadPoolExecutor
 from typing import TYPE_CHECKING, cast
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


src/ert/dark_storage/endpoints/responses.py
```diff
@@ -91,6 +91,90 @@ async def get_response(
...
+async def get_gradient(
+    *,
+    storage: Storage = DEFAULT_STORAGE,
+    ensemble_id: UUID,
+    key: str,
+    accept: Annotated[str | None, Header()] = None,
+) -> Response:
+    try:
+        ensemble = storage.get_ensemble(ensemble_id)
+    except KeyError as e:
```
berland, line 120: I added https://github.com/equinor/ert/pull/12934 on which you can cherry-pick rebase. 

The `data_for_gradient` should probably also be wrapped in the context manager for consistency with `get_response`

src/ert/gui/tools/plot/plot_window.py
```diff
@@ -304,12 +339,21 @@ def updatePlot(self, layer: int | None = None) -> None:
                         key_def.parameter.type
                         in {"gen_kw", "everest_parameters", "everest_objective"}
                     ):
-                        ensemble_to_data_map[ensemble] = self._api.data_for_parameter(
+                        data = self._api.data_for_parameter(
                             ensemble_id=ensemble.id,
```
berland, line 343: (an alternative to using threads could be to supply a list of ensemble id's here - changing the api as needed)

berland: No, probably not a very good idea.

src/ert/dark_storage/endpoints/responses.py
```diff
@@ -91,6 +91,90 @@ async def get_response(
...
+        logger.error(e)
+        raise HTTPException(status_code=404, detail="Ensemble not found") from e
+
+    unquoted_key = unquote(key)
+    dataframe = data_for_gradient(ensemble, unquoted_key)
+
+    media_type = accept if accept is not None else "text/csv"
```
andreas-el, line 127: I take it that `accept` are acceptable media types? Not very intuitive argument perhaps?

yngve-sk: Yep, suggest leaving rename / maybe a small refactor for separate issue/PR though, this is also in `parameters.py`

berland: I think this can be made nice with a `match` construct.

berland (approved):
Let's go!

PR #12901, opened 2026-02-17
Rename simulations to experiments in GUI

eivindjahren:
The gui would use the word "simulation" where "experiment" would be more appropriate.

This tries to make this consistent with the following usage

* A realization is simulated
* An experiment may include several ensembles and updates.

image

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/docs/ert/reference/configuration/keywords.rst
+++ b/docs/ert/reference/configuration/keywords.rst
@@ -716,8 +716,8 @@ knows if the DATA file is to be executed in parallel.
 
 .. _keywords_controlling_the_simulations:
 
-Keywords controlling the simulations
-------------------------------------
+Keywords controlling the experiment
+-----------------------------------
 
 .. _min_realizations:
 
@@ -725,7 +725,7 @@ MIN_REALIZATIONS
 ----------------
 
 MIN_REALIZATIONS is the minimum number of realizations that
-must have succeeded for the simulation to be regarded as a
+must have succeeded for the ensemble to be regarded as a
 success.
 
 MIN_REALIZATIONS can also be used in combination with
@@ -748,7 +748,7 @@ NUM_REALIZATIONS
 The MIN_REALIZATIONS key is optional, but if it has not been
 set *all* the realisations must succeed.
 
-Please note that MIN_REALIZATIONS = 0 means all simulations must succeed
+Please note that MIN_REALIZATIONS = 0 means all realizations must succeed
 (this happens to be the default value). Note MIN_REALIZATIONS is rounded up
 e.g. 2% of 20 realizations is rounded to 1.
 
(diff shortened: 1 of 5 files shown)


docs/ert/reference/configuration/keywords.rst
```diff
@@ -716,16 +716,16 @@ knows if the DATA file is to be executed in parallel.
 
 .. _keywords_controlling_the_simulations:
 
-Keywords controlling the simulations
-------------------------------------
+Keywords controlling the realizations
```
berland: "experiment" is maybe better fitted here?

docs/ert/reference/workflows/configuring_jobs.rst
```diff
@@ -46,10 +46,10 @@ executable:
 NB: note that relative paths are resolved from the location of the job
 configuration file, not the configuration file provided to ert.
 
-Stop Ert execution upon job failure
------------------------------------
-By default, failing jobs (both internal and external) will not stop the entire ert simulation.
-In some cases it is best to cancel the entire simulation if a job fails.
+Stop Workflow execution upon job failure
```
berland: Should not this also be experiment? The workflow is just the thing that can stop the experiment. If a workflow stops, it is stopped irrespective of this setting.

docs/ert/reference/configuration/keywords.rst
eivindjahren: Not, sure as the keywords in question are
* `MIN_REALIZATIONS`
* `SUBMIT_SLEEP`
* `STOP_LONG_RUNNING`
* `MAX_RUNNING`, and
* `MAX_RUNTIME`

I wonder if "simulations" was actually the correct word here. Its a weird collection of keywords to lump together. Should we go for "Keywords setting bounds for execution" ? Any other suggestions?

eivindjahren: I changed it to experiment as suggested.

docs/ert/reference/workflows/configuring_jobs.rst
eivindjahren: Workflow can run in several settings and not just inside an experiment, so I think this is correct.

src/ert/gui/__init__.py
```diff
@@ -1,11 +1,28 @@
+"""The Qt GUI for ert.
```
berland, line 1: Towards RUF067, nice!

src/ert/warnings/_warnings.py
```diff
@@ -2,5 +2,5 @@ class ErtWarning(Warning):
     """Base class for warnings in this module."""
 
 
-class PostSimulationWarning(ErtWarning):
-    """Warnings to be shown in GUI after simulation has finished."""
+class PostExperimentWarning(ErtWarning):
```
berland, line 5: This could easily trip some Grafana visualizations. I have checked that the left and right Scout dashboards do not contain this string (and a few others)

berland: Maybe @andreas-el has a hunch of whether it is in use anywhere.

eivindjahren: I am not sure whether it would effect any log message, if so, just indirectly.

PR #12919, opened 2026-02-18
Update glossary on ensembles and experiments

oyvindeide:
      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/docs/ert/reference/glossary.rst
+++ b/docs/ert/reference/glossary.rst
@@ -45,6 +45,11 @@ It is not about being correct, it is about being relevant and coherent.
         :code:`x` consists of only deterministic variables and is hence much easier to
         simulate and reason about than the template model.
 
+    experiment
+        An experiment consists of one or more ensembles, which may be related via
+        zero or more update steps. For example you might have a prior ensemble that
+        results in the posterior ensemble by running IES.
+
     ensemble
         As the name hints towards, ERT's approach to studying a template model, is to
         create an ensemble of realisations. In particular, an ensemble is a list of
@@ -84,11 +89,6 @@ It is not about being correct, it is about being relevant and coherent.
         way to obtain a merged ensemble response where all realisations that was
         successful in at least one the two ensembles evaluations will be successful in
         the merged ensemble response.
-
-    experiment
-        An experiment consists of one or more ensembles, which may be related via
-        zero or more update steps. For example you might have a prior ensemble that
-        results in the posterior ensemble by running IES.
 
     simulation
         There are two uses of the word simulation:
--- a/docs/ert/reference/glossary.rst
+++ b/docs/ert/reference/glossary.rst
@@ -54,7 +54,10 @@ It is not about being correct, it is about being relevant and coherent.
     experiment
         An experiment consists of one or more ensembles, which may be related via
         zero or more update steps. For example you might have a prior ensemble that
-        results in the posterior ensemble by running IES.
+        results in the posterior ensemble by running IES. We define an experiment
+        as a set of parameter definitions and a set of observation values. That means
+        tha the parameter definitions and the observation values as shared by all
+        ensembles in an experiment.
 
     evaluating an ensemble
         Recall that a realisation consists of a forward model :code:`fm` together with input
--- a/docs/ert/reference/glossary.rst
+++ b/docs/ert/reference/glossary.rst
@@ -51,9 +51,9 @@ It is not about being correct, it is about being relevant and coherent.
         realisations. The idea is that with a sufficiently large ensemble the
         uncertainty of the template model is represented by the ensemble. An ensemble
         consists of input parameters and the corresponding responses of the forward model, and
-        each ensemble can trace it´s lineage to its parent ensemble and the experiment.
+        each ensemble can trace its lineage to its parent ensemble and the experiment.
         When we talk about about prior and posterior ensembles this is in the context of
-        and update, where we load the responses of a prior ensemble, and use it together
+        an update, where we load the responses of a prior ensemble and use it together
         with the observation values to create a posterior ensemble. An ensemble without
         a parent has a special meaning, as that has sampled parameters which are not the
         result of an update. The lineage in multiple data assimilation can be shown as:
--- a/docs/ert/reference/glossary.rst
+++ b/docs/ert/reference/glossary.rst
@@ -70,7 +70,7 @@ It is not about being correct, it is about being relevant and coherent.
         zero or more update steps. For example you might have a prior ensemble that
         results in the posterior ensemble by running IES. We define an experiment
         as a set of parameter definitions and a set of observation values. That means
-        tha the parameter definitions and the observation values as shared by all
+        that the parameter definitions and the observation values as shared by all
         ensembles in an experiment.
 
     evaluating an ensemble
--- a/docs/ert/reference/glossary.rst
+++ b/docs/ert/reference/glossary.rst
@@ -45,16 +45,16 @@ It is not about being correct, it is about being relevant and coherent.
         :code:`x` consists of only deterministic variables and is hence much easier to
         simulate and reason about than the template model.
 
-    experiment
-        An experiment consists of one or more ensembles, which may be related via
-        zero or more update steps. For example you might have a prior ensemble that
-        results in the posterior ensemble by running IES.
-
     ensemble
         As the name hints towards, ERT's approach to studying a template model, is to
         create an ensemble of realisations. In particular, an ensemble is a list of
         realisations. The idea is that with a sufficiently large ensemble the
         uncertainty of the template model is represented by the ensemble.
+
+    experiment
+        An experiment consists of one or more ensembles, which may be related via
+        zero or more update steps. For example you might have a prior ensemble that
+        results in the posterior ensemble by running IES.
 
     evaluating an ensemble
         Recall that a realisation consists of a forward model :code:`fm` together with input
--- a/docs/ert/reference/glossary.rst
+++ b/docs/ert/reference/glossary.rst
@@ -49,7 +49,21 @@ It is not about being correct, it is about being relevant and coherent.
         As the name hints towards, ERT's approach to studying a template model, is to
         create an ensemble of realisations. In particular, an ensemble is a list of
         realisations. The idea is that with a sufficiently large ensemble the
-        uncertainty of the template model is represented by the ensemble.
+        uncertainty of the template model is represented by the ensemble. An ensemble
+        consists of input parameters and the corresponding responses of the forward model, and
+        each ensemble can trace it´s lineage to its parent ensemble and the experiment.
+        When we talk about about prior and posterior ensembles this is in the context of
+        and update, where we load the responses of a prior ensemble, and use it together
+        with the observation values to create a posterior ensemble. An ensemble without
+        a parent has a special meaning, as that has sampled parameters which are not the
+        result of an update. The lineage in multiple data assimilation can be shown as:
+
+        .. code-block:: python
+
+           prior = sample_parameters()
+           for iteration, weight in enumerate(weights):
+               posterior = update(prior, weight)
+               prior = posterior
 
     experiment
         An experiment consists of one or more ensembles, which may be related via
--- a/docs/ert/reference/glossary.rst
+++ b/docs/ert/reference/glossary.rst
@@ -66,12 +66,12 @@ It is not about being correct, it is about being relevant and coherent.
                prior = posterior
 
     experiment
-        An experiment consists of one or more ensembles, which may be related via
-        zero or more update steps. For example you might have a prior ensemble that
-        results in the posterior ensemble by running IES. We define an experiment
-        as a set of parameter definitions and a set of observation values. That means
-        that the parameter definitions and the observation values as shared by all
-        ensembles in an experiment.
+        An experiment consists of a set of ensembles, which are related via update
+        steps. For example you might have a prior ensemble that results in a
+        posterior ensemble by running the ensemble smoother algorithm.
+        We define an experiment as a set of parameter definitions and a set of
+        observation values. That means that the parameter definitions and the
+        observation values as shared by all ensembles in an experiment.
 
     evaluating an ensemble
         Recall that a realisation consists of a forward model :code:`fm` together with input
--- a/docs/ert/reference/glossary.rst
+++ b/docs/ert/reference/glossary.rst
@@ -72,7 +72,8 @@ It is not about being correct, it is about being relevant and coherent.
         experiment consists of one or more ensembles, which are related via
         zero or more update steps. In the case of ensemble smoother you will have one
         prior ensemble and one posterior ensemble, while for multiple data assimilation
-        there will be multiple posterior ensembles.
+        there will typically be four ensembles, the sampled prior, the posterior after the first update,
+        the posterior after the second update and finally the posterior after the third update.
 
     evaluating an ensemble
         Recall that a realisation consists of a forward model :code:`fm` together with input
--- a/docs/ert/reference/glossary.rst
+++ b/docs/ert/reference/glossary.rst
@@ -46,17 +46,16 @@ It is not about being correct, it is about being relevant and coherent.
         simulate and reason about than the template model.
 
     ensemble
-        As the name hints towards, ERT's approach to studying a template model, is to
-        create an ensemble of realisations. In particular, an ensemble is a list of
-        realisations. The idea is that with a sufficiently large ensemble the
-        uncertainty of the template model is represented by the ensemble. An ensemble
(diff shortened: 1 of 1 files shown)


docs/ert/reference/glossary.rst
```diff
@@ -45,11 +45,33 @@ It is not about being correct, it is about being relevant and coherent.
         :code:`x` consists of only deterministic variables and is hence much easier to
         simulate and reason about than the template model.
 
+    experiment
+        An experiment consists of one or more ensembles, which may be related via
+        zero or more update steps. For example you might have a prior ensemble that
+        results in the posterior ensemble by running IES. We define an experiment
+        as a set of parameter definitions and a set of observation values. That means
+        tha the parameter definitions and the observation values as shared by all
```
andreas-el: Minor typo "tha"

docs/ert/reference/glossary.rst
```diff
@@ -45,11 +45,33 @@ It is not about being correct, it is about being relevant and coherent.
...
+        ensembles in an experiment.
+
     ensemble
         As the name hints towards, ERT's approach to studying a template model, is to
         create an ensemble of realisations. In particular, an ensemble is a list of
         realisations. The idea is that with a sufficiently large ensemble the
-        uncertainty of the template model is represented by the ensemble.
+        uncertainty of the template model is represented by the ensemble. An ensemble
+        consists of input parameters and the corresponding responses of the forward model, and
+        each ensemble can trace it´s lineage to its parent ensemble and the experiment.
```
andreas-el: Should remove apostrophe.

docs/ert/reference/glossary.rst
```diff
@@ -45,11 +45,33 @@ It is not about being correct, it is about being relevant and coherent.
         :code:`x` consists of only deterministic variables and is hence much easier to
         simulate and reason about than the template model.
 
+    experiment
+        An experiment consists of one or more ensembles, which may be related via
+        zero or more update steps. For example you might have a prior ensemble that
```
andreas-el: Should this section have been underneath `ensemble` ? This to avoid referencing keywords not yet described?

docs/ert/reference/glossary.rst
```diff
@@ -45,11 +45,33 @@ It is not about being correct, it is about being relevant and coherent.
...
     ensemble
         As the name hints towards, ERT's approach to studying a template model, is to
         create an ensemble of realisations. In particular, an ensemble is a list of
         realisations. The idea is that with a sufficiently large ensemble the
-        uncertainty of the template model is represented by the ensemble.
+        uncertainty of the template model is represented by the ensemble. An ensemble
+        consists of input parameters and the corresponding responses of the forward model, and
+        each ensemble can trace it´s lineage to its parent ensemble and the experiment.
+        When we talk about about prior and posterior ensembles this is in the context of
+        and update, where we load the responses of a prior ensemble, and use it together
```
andreas-el: an(d) update?

Consider removing last comma?

docs/ert/reference/glossary.rst
achaikou: "may be related" confuses me.
Can you prodive an example of experiment which consists of more than one ensemble that are NOT related via update steps?

docs/ert/reference/glossary.rst
```diff
@@ -48,7 +48,10 @@ It is not about being correct, it is about being relevant and coherent.
     experiment
         An experiment consists of one or more ensembles, which may be related via
         zero or more update steps. For example you might have a prior ensemble that
-        results in the posterior ensemble by running IES.
+        results in the posterior ensemble by running IES. We define an experiment
```
berland: the -> a (?)

docs/ert/reference/glossary.rst
```diff
@@ -45,11 +45,33 @@ It is not about being correct, it is about being relevant and coherent.
         :code:`x` consists of only deterministic variables and is hence much easier to
         simulate and reason about than the template model.
 
+    experiment
+        An experiment consists of one or more ensembles, which may be related via
+        zero or more update steps. For example you might have a prior ensemble that
+        results in the posterior ensemble by running IES. We define an experiment
```
achaikou: `IES` is not a glossary term.
Also - why "for example"? Are there other options other than "prior" and "posterior" of how ensembles in the experiment can be related to each other?

docs/ert/reference/glossary.rst
```diff
@@ -45,11 +45,33 @@ It is not about being correct, it is about being relevant and coherent.
...
+        as a set of parameter definitions and a set of observation values. That means
```
achaikou: I think it is more than a set of parameter definitions and observation values.
Experiment is supposed to be all information provided by the user, including forward model steps, workflows, chosen seed, etc AND, as far as I can tell, the final and intermediate results of the whole run.

Experiment is one big main entity that contains it all and I miss that emphasis in the definition.

docs/ert/reference/glossary.rst
```diff
@@ -45,11 +45,33 @@ It is not about being correct, it is about being relevant and coherent.
...
+        that the parameter definitions and the observation values as shared by all
+        ensembles in an experiment.
+
     ensemble
         As the name hints towards, ERT's approach to studying a template model, is to
         create an ensemble of realisations. In particular, an ensemble is a list of
         realisations. The idea is that with a sufficiently large ensemble the
```
achaikou: First two sentences confuse me quite a bit.
How does name "hint towards" and why is that important? Why is that "in particular"?
And "template model" btw is used in glossary but never defined, so what means to "study a template model" is also shady.

Wouldn't something like

> An ensemble is a list of realisations, i.e. a set of execution results of the same iteration of the same forward model for various input values (parameters).

be more informative?

Though I have to note: Model Update also creates an ensemble and it does NOT contain realizations. It only contains inputs. So this makes me stuck a bit. By that definition should start with something like

> An ensemble is a set of currently chosen parameters for one iteration of the forward model and its run's results, if they are available.

docs/ert/reference/glossary.rst
```diff
@@ -45,11 +45,33 @@ It is not about being correct, it is about being relevant and coherent.
         :code:`x` consists of only deterministic variables and is hence much easier to
         simulate and reason about than the template model.
 
+    experiment
+        An experiment consists of one or more ensembles, which may be related via
```
achaikou: For the moment, this sentence should go after definition of what experiment is ("we define an experiment as"...) because now this is an additional information, not definition.

Current wording makes it seem like an experiment is supposed to be considered mainly as a set of ensembles with little additional meaning, while we want to define it as one grand execution of all forward model steps and updates on the set of input parameters.

docs/ert/reference/glossary.rst
```diff
@@ -45,11 +45,33 @@ It is not about being correct, it is about being relevant and coherent.
...
         create an ensemble of realisations. In particular, an ensemble is a list of
         realisations. The idea is that with a sufficiently large ensemble the
-        uncertainty of the template model is represented by the ensemble.
+        uncertainty of the template model is represented by the ensemble. An ensemble
+        consists of input parameters and the corresponding responses of the forward model, and
+        each ensemble can trace it´s lineage to its parent ensemble and the experiment.
+        When we talk about about prior and posterior ensembles this is in the context of
+        and update, where we load the responses of a prior ensemble, and use it together
+        with the observation values to create a posterior ensemble. An ensemble without
+        a parent has a special meaning, as that has sampled parameters which are not the
```
achaikou: The order looks very confusing here as you first talk about update and then about something with a "special meaning".
In all the runs we always first have an ensemble without a parent and only then definitions or prior and posterior come it.

So I think something like that is better:

> All ensembles can be organized into a forest that may span multiple experiments.
> Only the first ensemble in the experiment may be a root (have no parent). Its input parameters are then chosen randomly.
> Other ensembles must have a parent due to model update. Parent and child ensembles don't have to belong to the same experiment. Parent ensemble is called "prior", when child one is called "posterior". In update we load the response of a prior ensemble and use it together with observation values to create a posterior ensemble. For example, the lineage in multiple data assimilation experiment can be show as:

P.S. Not sure how parameters are chosen in this "manual" ensemble/experiment that can be created via Manage Experiments and if it fits here or is a special snowflake.

docs/ert/reference/glossary.rst
oyvindeide: Maybe, my reasoning is that experiment is above ensemble in the hierarchy, but see you point

achaikou (commented):
Sorry for many comments, but during my onboarding into ERT I found our docs quite confusing, so I have _opinions_. :smiling_imp:

docs/ert/reference/glossary.rst
oyvindeide: Not glossary, and not even relevant anymore 😅 It refers to Iterative Ensemble Smoother, an algorithm we deleted. Will rewrite.

> Are there other options other than "prior" and "posterior" of how ensembles in the experiment can be related to each other

No, not that I can think of

docs/ert/reference/glossary.rst
oyvindeide: You are right, I will rewrite this section

docs/ert/reference/glossary.rst
oyvindeide: > as far as I can tell, the final and intermediate results of the whole run.

Not sure what you mean here, the results are stored in the ensembles, but encompassed by the experiment?

Otherwise I agree, I will reword to expand the definition.

achaikou: Yes, this is what I mean :cat: 

But your question leads to another interesting point:
- Do you understand experiment as a directory which stores the files? (then yes, it is relevant that results are actually stored in a different "ensembles" directory and thus do not belong to the "experiment" directory)
- Or do you understand experiment as user sees it - the big task? (then directories are irrelevant. Results are part of the experiment. That results are also part of the ensemble and are stored in the ensembles directory is just a detail.)

To me "experiment" definition is that facade root task, same as for the user.  An "ensemble" is just one step in its execution. How it is structured on the disk is a totally separate issue.

oyvindeide: Agree with you on that, I also understand it as:
> Or do you understand experiment as user sees it - the big task? (then directories are irrelevant. Results are part of the experiment. That results are also part of the ensemble and are stored in the ensembles directory is just a detail.)

docs/ert/reference/glossary.rst
oyvindeide: I agree with your wording, but am a bit confused by: 

> Though I have to note: Model Update also creates an ensemble and it does NOT contain realizations. It only contains inputs. So this makes me stuck a bit. By that definition should start with something like

Do you mean `Manual update`? That creates realizations

achaikou: Yes, Manual Update (EnIF or ES).

As I think about it, you are right, realizations are not gone in Manual Update. :+1: 

---

But if so - does current definition of ensemble include Manual Update ensembles? I think it talks only about Evaluation step.

This confusion makes me wonder if the thing we now call ensemble might actually be two different entities in the new model:
- one entity dealing with running forward models (evaluation/iteration step)
- one entity dealing with input parameters and observations (sampling or updating)

---

That also leads to new ~wall of text~  thoughts (sorry 😅 )

1. If we follow the definition of realization from the glossary (which I am _not_ happy about too), it talks something about "consisting of forward model with input variables", which in not a part of update as long as I understand.

2. If we follow what we see in storage, it also is not clear.
- It creates realization directories if it has something to copy there - like `error.json`
- It doesn't create any realization directories if all is good. There is just, for example, `SCALAR.parquet` file present. Realizations are still there, they are just "hidden".

So now it all depends on proper definition of "realization" 😄 

I think I considered realizations to relate only to the forward model. That was a confusing part - in my mental model realizations applied only to evaluation step, which you showed me is not true.

If we follow the current _folder_ structure, then we have 3 kinds of ensembles:
- forward model results only (ensemble created via `Ensemble Experiment`)
- update results only (ensemble created via `Manual Update`)
- forward model and update result mixed together (ensembles created in one step of `ES-MDA`)

So to me if we define realization path as the line of steps (forward-model-runs and updates) and their results, stemming deterministically (on success) from chosen input parameters, then currently ensemble is some kind of intersection between one abstract step (or two steps) and all realization paths.

docs/ert/reference/glossary.rst
```diff
@@ -66,12 +66,14 @@ It is not about being correct, it is about being relevant and coherent.
...
-        observation values. That means that the parameter definitions and the
-        observation values as shared by all ensembles in an experiment.
+        We define an experiment as a complete specification for running an ensemble of
+        models, which includes the parameter definitions and a set of observation values and
+        their responses. It also includes the specification of the forward model, and the
+        update method to be applied on the ensemble as well as the update settings. An
+        experiment consists of one or more ensembles, which are related via
+        zero or more update steps. In the case of ensemble smoother you will have one
+        prior ensemble and one posterior ensemble, while for multiple data assimilation
+        there will be multiple posterior ensembles.
```
achaikou: ... and also multiple prior ones?

(If I were reading this definition for the first time, I would assume I misunderstood something about prior-posterior, because I read this line as "ES: 1 prior, 1 posterior. ES-MDA: still 1 prior, many posteriors")

Not sure if talk about prior-posterior is not confusing here.

docs/ert/reference/glossary.rst
```diff
@@ -66,12 +66,14 @@ It is not about being correct, it is about being relevant and coherent.
...
-        posterior ensemble by running the ensemble smoother algorithm.
-        We define an experiment as a set of parameter definitions and a set of
-        observation values. That means that the parameter definitions and the
-        observation values as shared by all ensembles in an experiment.
+        We define an experiment as a complete specification for running an ensemble of
+        models, which includes the parameter definitions and a set of observation values and
+        their responses. It also includes the specification of the forward model, and the
+        update method to be applied on the ensemble as well as the update settings. An
+        experiment consists of one or more ensembles, which are related via
+        zero or more update steps. In the case of ensemble smoother you will have one
```
achaikou, line 73: Interesting:

> ensembles are related via update steps

Does it mean that you do not consider update steps to be ensembles?

Currently an experiment can be just one ensemble which is an update step (Manual Update).
(And I like it because it challenges our mental model of what ensemble and experiment are.)

achaikou (commented):
Defining things is difficult, I probably can fight you for a few more review iterations, so will approve whenever you wish 😄

docs/ert/reference/glossary.rst
oyvindeide: > Does it mean that you do not consider update steps to be ensembles?

Correct, though a manual update will create a new ensemble because the update outputs parameters which we insert into the new ensemble.

> Currently an experiment can be just one ensemble which is an update step (Manual Update).
(And I like it because it challenges our mental model of what ensemble and experiment are.)

Yes, but there is something lacking in our data model here, we could solve that with having experiments be related I think?

docs/ert/reference/glossary.rst
oyvindeide: There will only be one prior for each ensemble, but a prior could potentially spawn multiple posteriors I guess? See that this is not covered by the datamodel, so need to improve this.

docs/ert/reference/glossary.rst
achaikou: I think this discussion means:
- we need a separate definition of an "update step"
- distinction that "ensemble" is a totally separate entry from "update" (ensemble can exist without updates, update connects two ensembles). Now ensemble definition talks too much about updates.
- experiment is a complete specification defining how to create ensembles and update them.

----

We do not need experiments to be related, I think.

But maybe we need to introduce glossary concept objects like"update step"/"prediction step"/"sampling step"/"simulation step".

Then sampling settings become property of "sampling step", forward model setups become property of "simulation step", update settings and observations become properties of "update step".

Experiment execution consists of steps. Experiment result consists of step outputs.
Outputs of "sampling step" and "update step" are ensemble input parameters.
Outputs of "simulation steps" are ensembles.

Then steps and results become related instead of experiments:
- ensemble A is the result of  execution of "sampling step" + "simulation step1".
- ensemble B is the result of execution of
    - "sampling step" + "simulation step1" + "update step" + "simulation step2", or
    -  "simulation step1" + "update step" + "simulation step2" applied to parameters1
    -  "update step" + "simulation step2" applied to ensemble A, or
    -  "simulation step2" applied to parameters2.

docs/ert/reference/glossary.rst
achaikou: > There will only be one prior for each ensemble,


Yes, but we are talking in the context of ES-MDA experiment. And in this context you contradict what is written in "ensemble" definition.

According to that definition when you have prior `A` and posterior `B`, posterior `B` becomes prior `B` in the context of new update to `C`.

And here it sounds like you want to name only sampling ensemble "prior".

PR #12933, opened 2026-02-18
Squash logs

### LINKED ISSUE #12921: Reduce logging by combining parameter sampling messages
We are logging quite the few instances (~650.000 last 30d) of messages when processing parameters during sampling.

I've come across these two, where combining them with others would reduce the number of messages needed, but also reduce repetetive information for all of them.

https://github.com/equinor/ert/blob/822e61c78afb3e3243270c66c775ae730eef07ea/src/ert/sample_prior.py#L49-L52

https://github.com/equinor/ert/blob/822e61c78afb3e3243270c66c775ae730eef07ea/src/ert/sample_prior.py#L65-L68

Image
Image

Kusto Query:
https://portal.azure.com#@3aa4a235-b6e2-48d5-9195-7fcf05b459b0/blade/Microsoft_OperationsManagementSuite_Workspace/Logs.ReactView/resourceId/%2Fsubscriptions%2Fe823d00d-e01a-4af3-b71b-61df62a8e3ea%2Fresourcegroups%2Ffmu-logging%2Fproviders%2Fmicrosoft.operationalinsights%2Fworkspaces%2Ffmu-logs/source/LogsBlade.AnalyticsShareLinkToQuery/q/H4sIAAAAAAAAA3XOsQrCMBCH8T1PcXRqpwruglNxEAf7AmfzJwbMJVyuFMWHlwwugvuPj%252B9Yyqy8oLo3bXcoaI4JEwTKBk8H4pD7%252Fc4Pbhy%252F5IxaOYCWLMZRKnUTzKIEKqycYNDO%252FcdXTuXxo1u8rimxxlezq1g%252F0O1Jl9JOYpaT%252FwC%252FsnEvrAAAAA%253D%253D/limit/15000

eqbech:
**Issue**
Resolves #12921

### DIFF
--- a/src/ert/sample_prior.py
+++ b/src/ert/sample_prior.py
@@ -6,6 +6,7 @@ from pathlib import Path
 
 import polars as pl
 
+from ert.config.parameter_config import ParameterConfig
 from ert.utils import log_duration
 
 from .config import DataSource, GenKwConfig
@@ -35,6 +36,9 @@ def sample_prior(
     if parameters is None:
         parameters = list(parameter_configs.keys())
     complete_dataset: pl.DataFrame | None = None
+
+    log_params(parameter_configs, design_matrix_df, active_realizations)
+
     for parameter in parameters:
         config_node = parameter_configs[parameter]
         if config_node.forward_init:
@@ -46,10 +50,6 @@ def sample_prior(
                 config_node.input_source == DataSource.DESIGN_MATRIX
                 and design_matrix_df is not None
             ):
-                logger.info(
-                    f"Getting parameter {config_node.name} "
-                    f"from design matrix for realizations {active_realizations}"
-                )
                 cols = {"realization", config_node.name}
                 missing = cols - set(design_matrix_df.columns)
                 if missing:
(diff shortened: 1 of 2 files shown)


src/ert/sample_prior.py
```diff
@@ -35,6 +36,40 @@ def sample_prior(
...
+
+    sample_params: list[str] = [
+        parameter_configs[p].name
+        for p in parameters
+        if (
+            isinstance(parameter_configs[p], GenKwConfig)
+            and cast(GenKwConfig, parameter_configs[p]).input_source
+            == DataSource.SAMPLED
+        )
+    ]
```
andreas-el: You could also collect the names where the original logger statements where instead of doing this up front. But perhaps that was something you wanted to have?

eqbech: Both works, depends on whether you want the logging before or after the operation, in the case after, nothing will be logged if a `KeyError` is raised.

src/ert/sample_prior.py
```diff
@@ -35,6 +36,40 @@ def sample_prior(
...
+    if len(sample_params) > 0:
+        logger.info(
+            f"Sampling parameters {sample_params} "
+            f"for realizations {active_realizations}"
+        )
+    if len(get_params) > 0:
+        logger.info(
+            f"Getting parameters {get_params} "
+            f"from design matrix for realizations {active_realizations}"
+        )
```
andreas-el: I've dug a bit in the logs, and for both of these cases we have simulations that have in excess of 5000 parameters. Logging all of those might not be possible in one go?
Perhaps we would need to log multiple batches of 1000 parameters each, to accommodate?

eqbech: Do you know if this is restriction is from the python logger or azure/grafana?

andreas-el: The maximum payload for a logger statement is 32,768 characters (in Azure). We've had issues with reaching this limit before, so splitting into multiple messages would avoid going beyond this limit.

PR #12934, opened 2026-02-18
Use context manager and match-case in endpoints

berland:
**Issue**
Readability.

**Approach**
contextlib.

      --exec 'just rapid-tests'`)

## When applicable


src/ert/dark_storage/endpoints/responses.py
```diff
@@ -23,6 +24,21 @@
 DEFAULT_BODY = Body(...)
 
 
+@contextmanager
+def reraise_as_http_errors() -> Iterator[None]:
```
yngve-sk: I think this can be used for most other endpoints also? parameters, observations, experiments, etc?

src/ert/dark_storage/endpoints/parameters.py
```diff
@@ -45,61 +45,38 @@ async def get_parameter(
...
+                media_type="application/x-parquet",
+            )
+        case "application/json":
+            return Response(dataframe.to_json(), media_type="application/json")
+        case "text/csv" | None:
+            return Response(
+                content=dataframe.to_csv().encode(),
+                media_type="text/csv",
+            )
+    raise HTTPException(status_code=status.HTTP_415_UNSUPPORTED_MEDIA_TYPE)
```
eqbech: This could fall under the match statement under `case _:` ?

berland: Yes it could. Fixed!

src/ert/dark_storage/endpoints/parameters.py
```diff
@@ -45,61 +45,39 @@ async def get_parameter(
...
+            )
+        case "application/json":
+            return Response(dataframe.to_json(), media_type="application/json")
+        case "text/csv" | None:
+            return Response(
+                content=dataframe.to_csv().encode(),
+                media_type="text/csv",
+            )
+        case _:
+            raise HTTPException(status_code=status.HTTP_415_UNSUPPORTED_MEDIA_TYPE)
```
andreas-el: Can this be extracted as an helper function to avoid code-dedupe? The raise could be left to be raised on `return None` to maintain the stacktrace?

berland: Right, those were more equal than I anticipated. Added a helper function. Kept the `raise` in the helper function, the raise should not happen from our own cde.

andreas-el (approved):
Well done! 🥓

PR #12951, opened 2026-02-20
Provide distance base localization into the update status

xjules:
**Issue**
Currently the message is misleading even we run distance localization

image

**Approach**
update

      --exec 'just rapid-tests'`)

## When applicable

the new report

image

### DIFF
--- a/src/ert/analysis/_es_update.py
+++ b/src/ert/analysis/_es_update.py
@@ -335,14 +335,6 @@ def analysis_ES(
         # Create mask for non-zero variance parameters
         non_zero_variance_mask = ~np.isclose(param_variance, 0.0)
 
-        log_msg = (
-            f"Updating {np.sum(non_zero_variance_mask)} parameters "
-            f"{'with' if module.localization else 'without'} "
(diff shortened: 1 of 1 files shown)


src/ert/analysis/_es_update.py
oddvarlia: Maybe write adaptive localization where that is used instead of only localization?

src/ert/analysis/_es_update.py
```diff
@@ -503,6 +499,13 @@ def adaptive_localization_progress_callback(
             module.distance_localization is False
             or (module.distance_localization and isinstance(param_cfg, GenKwConfig))
         ) and module.localization is False:
+            log_msg = (
+                f"Updating {np.sum(non_zero_variance_mask)} parameters "
+                f"{'with' if module.localization else 'without'} "
```
oddvarlia: Since module.localization always is False, maybe simplify this and remove the if test?

oddvarlia (approved):
I suggested a couple of changes like using adaptive localization instead of localization where adaptive localization is used

PR #12961, opened 2026-02-23
squash logs in ErtServer fetch_url

### LINKED ISSUE #12460: Revise ErtServer.fetch_url logger message emits
When it loops through all the URLs, trying to connect, we should aggregate up all the errors and give one logging statement, instead of emitting one logger statement per failed try.

eqbech:
**Issue**
Resolves #12460

The connection attempts are kept in a dictionary grouped by the status code of the request, and the attempts will be logged if it fails, but if one succeeds it will only log that one. Let me know if we want it in any other format.

### DIFF
--- a/src/ert/services/ert_server.py
+++ b/src/ert/services/ert_server.py
@@ -302,6 +302,7 @@ class ErtServerController:
         """Returns the url. Blocks while the server is starting"""
         if self._url is not None:
             return self._url
+        logs: list[tuple[str | int, str]] = []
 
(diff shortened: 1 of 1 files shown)


eqbech:
> Could a list (of tuples?) be more natural than a dict, to signify that there is an explicit ordering of each "sublog" entry?
> 
> The commit header could also be expanded slightly, put as much information as possible into the 50 characters that we can use there (and capitalize the first word).

Sure, updated it to be a list of tuples and updated commit message.

PR #12977, opened 2026-02-26
Have scheduler kill hanging cluster jobs

### LINKED ISSUE #12965: Ert should handle forward model steps spawning never-ending processes
Some users have custom forward model steps that spawn new processes that live beyond the main process. This will cause Ert to have all forward model steps marked as finished, while the cluster reports something is still running, so it will be completely stuck without any info. 

Some options:
* Automatically kill cluster jobs where it has been more than 10 minutes since all forward models steps were reported finished (and checksum has arrived back)
* Give a pop up when it has been hanging for some time
* A new color in the driver status circles (orange - warning), and a tooltip that explains the situation
* An alternative override button for terminate experiment (terminate experiment - manually verified that it succeeded so the next iteration can start)
* fm_dispatch should automatically kill everything on both failure and success

**Additional context**
Reproducible by changing poly_eval.py to this
```
#!/usr/bin/env python3
import json
from pathlib import Path


def _load_coeffs(filename):
    with open(filename, encoding="utf-8") as f:
        return json.load(f)


def _evaluate(coeffs, x):
    return coeffs["a"]["value"] * x**2 + coeffs["b"]["value"] * x + coeffs["c"]["value"]


if __name__ == "__main__":
    import subprocess
    subprocess.Popen(["sleep", "99999"])
    Path("poly.out").write_text("\n".join(map(str, output)), encoding="utf-8")

```
Applicable for all drivers.

jonathan-eq:
**Issue**
Resolves #12965

**Approach**
This commit makes scheduler kill jobs running on the cluster if the evaluator has received a message that the job is finished, but the driver disagrees. This can be due to one of the forward model steps spawning new processes that Ert does not see. This fixes the issue where the orphan processes caused the drivers to report jobs as still running, and hanging the run.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/scheduler/scheduler.py
+++ b/src/ert/scheduler/scheduler.py
@@ -62,6 +62,9 @@ class SubmitSleeper:
 class Scheduler:
     BATCH_KILLING_INTERVAL = 1.5
 
+    # Seconds for driver to report job as finished after checksum arrival
+    ORPHAN_KILL_DELAY = 60
+
     def __init__(
         self,
         driver: Driver,
@@ -110,6 +113,7 @@ class Scheduler:
         self._realization_ids_to_kill: list[int] = []
         self._kill_task: asyncio.Task[None] | None = None
         self._stop_kill_task = asyncio.Event()
+        self._watchdog_tasks: list[asyncio.Task[None]] = []
 
     async def kill_all_jobs(self) -> None:
         await self.cancel_all_jobs()
@@ -197,7 +201,30 @@ class Scheduler:
             event = await self._manifest_queue.get()
             if type(event) is ForwardModelStepChecksum:
                 self.checksum.update(event.checksums)
+                iens = int(event.real)
+                task = asyncio.create_task(
+                    self._kill_if_driver_lags(iens),
+                    name=f"orphan-watchdog-{iens}",
+                )
+                self._watchdog_tasks.append(task)
             self._manifest_queue.task_done()
+
+    async def _kill_if_driver_lags(self, iens: int) -> None:
+        """Kill realization *iens* if the driver hasn't reported it
+        finished within the grace period after checksum arrival."""
+        job = self._jobs[iens]
+        try:
+            await asyncio.wait_for(
+                asyncio.shield(job.returncode), timeout=self.ORPHAN_KILL_DELAY
+            )
+        except TimeoutError:
+            logger.warning(
+                f"Realization {iens}: checksum received but driver has not "
+                f"reported the job as finished after {self.ORPHAN_KILL_DELAY}s. "
+                "Assuming orphaned child processes are keeping the cluster job "
+                "alive. Sending a signal to the queue system to kill the job now"
+            )
+            await self.schedule_kill(iens)
 
     async def _publisher(self) -> None:
         if self._ensemble_evaluator_queue is None:
@@ -335,6 +362,9 @@ class Scheduler:
             )
             if self._kill_task:
                 await self._kill_task
+            for watchdog_task in self._watchdog_tasks:
+                watchdog_task.cancel()
+            await asyncio.gather(*self._watchdog_tasks, return_exceptions=True)
         if self._cancelled:
             logger.debug("Scheduler has been cancelled, jobs are stopped.")
             return False
--- a/tests/ert/unit_tests/scheduler/test_scheduler.py
+++ b/tests/ert/unit_tests/scheduler/test_scheduler.py
@@ -12,6 +12,7 @@ from unittest.mock import AsyncMock, MagicMock
 import pytest
 
 from _ert.events import (
+    ForwardModelStepChecksum,
     RealizationFailed,
     RealizationStoppedLongRunning,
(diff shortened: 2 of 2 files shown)


src/ert/scheduler/scheduler.py
```diff
@@ -110,6 +113,7 @@ def __init__(
         self._realization_ids_to_kill: list[int] = []
         self._kill_task: asyncio.Task[None] | None = None
         self._stop_kill_task = asyncio.Event()
+        self._watchdog_tasks: list[asyncio.Task[None]] = []
```
jonathan-eq, line 116: This should probably have a better name. Any suggestions?

src/ert/scheduler/scheduler.py
```diff
@@ -197,8 +201,31 @@ async def _checksum_consumer(self) -> None:
...
+                self._watchdog_tasks.append(task)
             self._manifest_queue.task_done()
 
+    async def _kill_if_driver_lags(self, iens: int) -> None:
+        """Kill realization *iens* if the driver hasn't reported it
+        finished within the grace period of checksum arrival."""
+        job = self._jobs[iens]
+        try:
+            await asyncio.wait_for(
+                asyncio.shield(job.returncode), timeout=self.ORPHAN_KILL_DELAY
```
jonathan-eq, line 218: This has to be shielded, otherwise it will cancel the job.returncode Future on timeout.

src/ert/scheduler/scheduler.py
```diff
@@ -335,6 +362,9 @@ async def execute(
             )
             if self._kill_task:
                 await self._kill_task
+            for watchdog_task in list(self._watchdog_tasks):
```
jonathan-eq: This is already a list

xjules:
hm, was there a bug reported that this PR solves?

src/ert/scheduler/scheduler.py
erlenlh: `tracked_tasks` or `monitored_tasks` maybe?

src/ert/scheduler/scheduler.py
```diff
@@ -62,6 +62,9 @@ async def sleep_until_we_can_submit(self) -> None:
 class Scheduler:
     BATCH_KILLING_INTERVAL = 1.5
 
+    # Seconds for driver to report job as finished after checksum arrival
+    ORPHAN_KILL_DELAY = 60
```
berland, line 66: (we are sure that the fm_dispatch process has finished and is not hanging also?  - if fm_dispatch is still running (waiting?) then these are technically not orphans)

jonathan-eq: Yes, we get "runner exiting" in the logs, and checksum is the final event that is sent.

src/ert/scheduler/scheduler.py
```diff
@@ -197,8 +201,31 @@ async def _checksum_consumer(self) -> None:
...
+                task = asyncio.create_task(
+                    self._kill_if_driver_lags(iens),
+                    name=f"orphan-watchdog-{iens}",
+                )
+                self._watchdog_tasks.append(task)
             self._manifest_queue.task_done()
 
+    async def _kill_if_driver_lags(self, iens: int) -> None:
+        """Kill realization *iens* if the driver hasn't reported it
+        finished within the grace period of checksum arrival."""
```
berland, line 214: of -> after

src/ert/scheduler/scheduler.py
```diff
@@ -197,8 +201,31 @@ async def _checksum_consumer(self) -> None:
...
+        try:
+            await asyncio.wait_for(
+                asyncio.shield(job.returncode), timeout=self.ORPHAN_KILL_DELAY
+            )
+        except TimeoutError:
+            logger.warning(
+                f"Realization {iens}: checksum received but driver has not "
+                f"reported the job as finished after {self.ORPHAN_KILL_DELAY}s. "
+                "Assuming orphaned child processes are keeping the cluster job "
+                "alive. Killing the job now."
```
berland, line 225: "Killing the job now" -> "Sending a signal to the queue system to kill the job now"

src/ert/scheduler/scheduler.py
```diff
@@ -197,8 +201,31 @@ async def _checksum_consumer(self) -> None:
             event = await self._manifest_queue.get()
             if type(event) is ForwardModelStepChecksum:
                 self.checksum.update(event.checksums)
+                iens = int(event.real)
```
xjules, line 204: I often see that `ForwardModelStepCheckSum` is very flaky, like frequent error saying Checksum not received; ie. then this would not do anything. Maybe it is ok

jonathan-eq: Hmm, you are right! But why is checksum flaky? I did a random check, and one of the runs that had a lot of them also contained `
bjobs gave returncode 255 and error
`

jonathan-eq: It has been 34 different OperationIds with this error since january 1st. Maybe that is okay.

PR #12993, opened 2026-03-02
Make all batches selected by default in plot

### LINKED ISSUE #12957: Some Everest plots should have all ensembles/batches selected by default
When plotting e.g. "Batch responses", it does not make much sense that the default is that only the last batch is selected. All batches should be selected by default

Image

eqbech:
**Issue**
Resolves #12957

### DIFF
--- a/src/ert/gui/tools/plot/plot_ensemble_selection_widget.py
+++ b/src/ert/gui/tools/plot/plot_ensemble_selection_widget.py
@@ -37,11 +37,14 @@ class EnsembleSelectionWidget(QWidget):
     ensembleSelectionChanged = Signal()
 
     def __init__(
-        self, ensembles: list[EnsembleObject], number_of_plot_colors: int
+        self,
+        ensembles: list[EnsembleObject],
+        number_of_plot_colors: int,
+        is_everest: bool = False,
     ) -> None:
         QWidget.__init__(self)
         self._selected_ensembles = EnsembleSelectListWidget(
-            ensembles, number_of_plot_colors
+            ensembles, number_of_plot_colors, is_everest=is_everest
         )
 
         self._ensemble_layout = QVBoxLayout()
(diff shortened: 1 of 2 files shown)


src/ert/gui/tools/plot/plot_ensemble_selection_widget.py
```diff
@@ -87,9 +87,9 @@ def __init__(
             it.setData(EnsembleSelectListWidgetItemDataRole.ENSEMBLE, ensemble)
             it.setData(
                 EnsembleSelectListWidgetItemDataRole.COLOR_INDEX,
-                self.assign_available_color(None) if i == 0 else None,
+                self.assign_available_color(None) if i < 5 else None,
```
andreas-el: This should be replaced with `< MAXIMUM_SELECTED` instead.

andreas-el: Not sure regarding the impact for `ert`, this could be ok but could also be quite annoying that it loads all 5 by default.
Is there a way of determining if the plotter was triggered from `everest` or not?
Perhaps a default (selection count) could be passed on into the plotter?

eqbech: Good spot, will change it to use the consts instead. Im not sure, we did mention using a Everest flag for all the everest specific modifications, but im not sure this should be introduced in this pr?

andreas-el: Can you consider serving the everest flag into this `EnsembleSelectionWidget` object?
I still worry that the impact of autoselecting 5 plots for all art-users might be unwelcome.

The flag is present where the object is created, e.g.
https://github.com/equinor/ert/blob/6588eab3946a089562bf2162a3001684a7c35c0f/src/ert/gui/tools/plot/plot_window.py#L185-L260

eqbech: added 👍

PR #13023, opened 2026-03-04
Extract build_strategy_map from smoother_update

dafeda:
This breaks `semeio` since `smoother_update` is a public API.
I've created a PR in `semeio` to use the new `smoother_update`: https://github.com/equinor/semeio/pull/867

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/analysis/_es_update.py
+++ b/src/ert/analysis/_es_update.py
@@ -10,7 +10,7 @@ from typing import TYPE_CHECKING, TextIO
 import numpy as np
 import polars as pl
 
-from ert.config import Field, ObservationSettings, SurfaceConfig
+from ert.config import ESSettings, Field, ObservationSettings, SurfaceConfig
 
 from ._update_commons import (
     ErtAnalysisError,
@@ -352,13 +352,24 @@ def smoother_update(
     posterior_storage: Ensemble,
     observations: Iterable[str],
     update_settings: ObservationSettings,
-    strategy_map: dict[str, UpdateStrategy],
+    strategy_map: dict[str, UpdateStrategy] | None = None,
     progress_callback: Callable[[AnalysisEvent], None] | None = None,
     global_scaling: float = 1.0,
     active_realizations: list[bool] | None = None,
 ) -> SmootherSnapshot:
     if not progress_callback:
         progress_callback = noop_progress_callback
+
+    if strategy_map is None:
+        settings = ESSettings()
+        experiment = prior_storage.experiment
+        strategy_map = build_strategy_map(
+            parameters=experiment.update_parameters,
+            param_configs=experiment.parameter_configuration,
+            inversion=settings.inversion,
+            enkf_truncation=settings.enkf_truncation,
+            progress_callback=progress_callback,
+        )
 
     ens_mask = prior_storage.get_realization_mask_with_responses()
     ens_mask = _create_combined_ensemble_mask(ens_mask, active_realizations)
--- a/tests/ert/unit_tests/analysis/test_es_update.py
+++ b/tests/ert/unit_tests/analysis/test_es_update.py
@@ -301,20 +301,11 @@ def test_update_handles_precision_loss_in_std_dev(tmp_path):
         )
         events = []
 
-        es_settings = ESSettings()
-        strategy_map = build_strategy_map(
-            parameters=["coeff_0"],
-            param_configs=prior.experiment.parameter_configuration,
-            inversion=es_settings.inversion,
-            enkf_truncation=es_settings.enkf_truncation,
-            progress_callback=events.append,
-        )
         ss = smoother_update(
             prior,
             posterior,
             experiment.observation_keys,
             ObservationSettings(auto_scale_observations=[["OBS*"]]),
-            strategy_map,
             progress_callback=events.append,
         )
 
@@ -648,20 +639,11 @@ def test_update_only_using_subset_observations(
     )
     events = []
 
-    es_settings = ESSettings()
-    strategy_map = build_strategy_map(
-        parameters=ert_config.ensemble_config.parameters,
-        param_configs=prior_ens.experiment.parameter_configuration,
-        inversion=es_settings.inversion,
-        enkf_truncation=es_settings.enkf_truncation,
-        progress_callback=events.append,
-    )
     smoother_update(
         prior_ens,
         posterior_ens,
         ["WPR_DIFF_1"],
         ObservationSettings(),
-        strategy_map,
         progress_callback=events.append,
     )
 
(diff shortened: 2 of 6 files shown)


src/ert/analysis/_es_update.py
```diff
@@ -353,30 +399,36 @@ def log_warning(
...
-        progress_callback(
-            AnalysisErrorEvent(
-                error_msg=str(e),
-                data=DataSection(
-                    header=smoother_snapshot.header,
-                    data=smoother_snapshot.csv,
-                    extra=smoother_snapshot.extra,
-                ),
+        if smoother_snapshot is not None:
+            progress_callback(
```
eivindjahren, line 422: There doesn't seem to be a test for this block (exception occurred and `smoother_snapshot is not None`. Could be nice to add one to make further refactoring more safe.

src/ert/analysis/_es_update.py
```diff
@@ -244,50 +250,68 @@ def perform_ensemble_update(
...
     )
 
+    return smoother_snapshot
 
-def smoother_update(
-    prior_storage: Ensemble,
-    posterior_storage: Ensemble,
-    observations: Iterable[str],
+
+def build_strategy_map(
```
eivindjahren, line 254: Maybe call this just `strategy_map`?

src/ert/analysis/_es_update.py
```diff
@@ -170,12 +176,12 @@ def perform_ensemble_update(
...
+    # Extract observation locations when location data is available
     observation_locations: ObservationLocations | None = None
-    if module.distance_localization:
-        has_location = (
-            filtered_data["east"].is_not_null() & filtered_data["north"].is_not_null()
-        ).to_numpy()
+    has_location = (
+        filtered_data["east"].is_not_null() & filtered_data["north"].is_not_null()
+    ).to_numpy()
+    if has_location.any():
```
eivindjahren, line 182: I guess this is more than "extract build_strategy_map", and changes behavior from "You have to enable distance based localization" to "distance based localization is always used when location is available" ?

tests/ert/unit_tests/analysis/test_es_update.py
```diff
@@ -1029,13 +1092,21 @@ def test_gen_data_missing(storage, uniform_parameter, obs):
     )
     events = []
 
+    es_settings = ESSettings()
+    strategy_map = build_strategy_map(
```
eivindjahren: I guess these changes to the tests would not be necessary if the implicit default value in the strategy map was 
```
settings = ESSettings()
standard_strategy = StandardESUpdate(
        settings.inversion, settings.enkf_truncation,
        np.random.default_rng(), noop_progress_callback,
)
```
then we could have the default values `strategy_map={}, observation_settings=ObservationSettings()`. 

Currently we instead have the failure condition of the parameter name might not be in the dictionary. That could be solved by not giving a list of strings for the parameters, but instead using `parameters = list(strategy_map.keys())`.

In this particular case you would then instead have: 
```python
    update_snapshot = smoother_update(
        prior,
        posterior_ens,
        ["OBSERVATION"],
        {"KEY_1": StandardESUpdate(
            settings.inversion, settings.enkf_truncation,
            np.random.default_rng(), noop_progress_callback,
        )},
        progress_callback=events.append,
    )
```

src/ert/analysis/_es_update.py
```diff
@@ -353,30 +399,36 @@ def log_warning(
...
+            smoother_snapshot = perform_ensemble_update(
                 parameters,
                 observations,
-                es_settings,
                 update_settings,
                 global_scaling,
-                smoother_snapshot,
                 ens_mask,
                 prior_storage,
                 posterior_storage,
```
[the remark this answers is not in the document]
eivindjahren: This is related to the suggestion I had about avoiding giving a separate list of parameters.

src/ert/analysis/_es_update.py
```diff
@@ -353,30 +399,36 @@ def log_warning(
                     )
 
             warnings.showwarning = log_warning
-            perform_ensemble_update(
+            smoother_snapshot = perform_ensemble_update(
```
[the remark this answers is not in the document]
eivindjahren: This seems like a good catch.

src/ert/analysis/_es_update.py
```diff
@@ -353,30 +399,36 @@ def log_warning(
...
+            )
+        else:
+            progress_callback(
+                AnalysisErrorEvent(
+                    error_msg=str(e),
+                    data=DataSection(header=[], data=[]),
+                )
             )
-        )
         raise e
```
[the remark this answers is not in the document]
eivindjahren: This is unrelated to this change.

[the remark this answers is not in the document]
dafeda: I agree it's unrelated, but kinda curious as to why Copilot things this is a good change.
I have feeling this is not that relevant with newer versions of Python, but `ert` supports older versions as well so maybe that's why.

src/ert/analysis/_es_update.py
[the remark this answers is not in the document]
dafeda: I thought that `log_warning` is replaced with the original handler when the `with` block exists.
Is this not the case?

src/ert/analysis/_es_update.py
dafeda: Seems like the test `test_update_raises_on_singular_matrix` covers this path.

src/ert/analysis/_es_update.py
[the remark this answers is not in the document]
dafeda: This is a good idea 👍

tests/ert/unit_tests/analysis/test_es_update.py
dafeda: Thanks, I think having some good defaults makes sense. I have pushed a proposal.

src/ert/analysis/_es_update.py
dafeda: `ObservationLocations` is computed whenever location data exists in the observations, but that does not mean distance-based localization is used.

Which strategy actually runs is determined by the `strategy_map` which is built by the caller based on parameter type and user configuration, not whether locations happen to exist.

Does this make sense?

src/ert/analysis/_es_update.py
dafeda: I personally prefer that it starts with a verb but I will not die on this hill :D

src/ert/analysis/_es_update.py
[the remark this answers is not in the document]
eivindjahren: Ah, you are right, I didn't see the entire context.

src/ert/analysis/_es_update.py
eivindjahren: Not a problem :)

src/ert/analysis/_es_update.py
eivindjahren: Makes sense!

src/ert/analysis/_es_update.py
eivindjahren: Coverage data says its not hit, and on my machine when I put a breakpoint on line 414 I get `smoother_snapshot == None`.

dafeda: My bad you are right. Not sure how to write a test for this though.

PR #13048, opened 2026-03-06
Add config for update strategy for parameter types

### LINKED ISSUE #13027: Add configuration for default update strategy per parameter type
Allow users to define a default update strategy based on the parameter type. This default should apply automatically to all parameters of that type, which later can be either override at the parameter level (ideally from the GUI - but currently not in place).
For example:
```
ANALYSIS_SET_VAR PARAMETERS FIELD DISTANCE
ANALYSIS_SET_VAR PARAMETERS GEN_KW ADAPTIVE
```

frode-aarstad:
**Issue**
Resolves #13027 

* Deprecated *


docs/ert/reference/configuration/keywords.rst
```diff
@@ -1694,6 +1694,25 @@ proportional to the identity matrix.
...
+DISTANCE
+
+ADAPTIVE
+
+
+::
+        ANALYSIS_SET_VAR PARAMETERS FIELD DISTANCE
+        ANALYSIS_SET_VAR PARAMETERS GEN_KW ADAPTIVE
+
+
```
frode-aarstad, line 1741: I need some help filling out the docs here

xjules:
We will need to link it to the storage (maybe parameter configs?) and currently we don't have a consensus on whether it should be the full analysis config info (and thus dropping the analysis config entirely) or just to have a name of the `UpdateStrategy`.
Related issue: https://github.com/equinor/ert/issues/13049

docs/ert/reference/configuration/keywords.rst
```diff
@@ -1698,6 +1698,25 @@ proportional to the identity matrix.
...
+The PARAMETERS keyword allows users to define a default update strategy for parameters based on the parameter type.
+
+DISTANCE
+
+ADAPTIVE
+
+
+::
+        ANALYSIS_SET_VAR PARAMETERS FIELD DISTANCE
+        ANALYSIS_SET_VAR PARAMETERS GEN_KW ADAPTIVE
```
xjules, line 1728: And the default classical ensemble smoother? Or that will be by default then?

docs/ert/reference/configuration/keywords.rst
```diff
@@ -1698,6 +1698,25 @@ proportional to the identity matrix.
...
+
+
+
+
+
+
+
```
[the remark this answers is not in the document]
xjules: "<parameter-type> is the parameter group " is ambigous given that our parameter group concept is something else. I'd say just type instead.

xjules:
I see that you modify `to27.py` migration. The only issue there is that we do have users (testing the DL) using bleeding and thus they would not be able to reopen the storage when using newer bleeding builds.

docs/ert/reference/configuration/keywords.rst
frode-aarstad: What do you mean?

frode-aarstad:
Updated the docs and created new migration version

docs/ert/reference/configuration/keywords.rst
xjules: running classical ES-MDA

docs/ert/reference/configuration/keywords.rst
```diff
@@ -903,9 +903,10 @@ Field parameters (e.g. porosity, permeability or Gaussian Random Fields from APS
   If not specified, the global grid from the :ref:`GRID<grid>` keyword will be used.
 
 - **UPDATE** (Optional)
-  Specifies whether this field parameter should be included during the history matching
-  update step. Must be set to either ``TRUE`` or ``FALSE``.
-  Defaults to ``TRUE``.
+  Specifies the update strategy for this field parameter during the history matching
```
xjules, line 906: We should then have an option for `NO_LOC` update version; suggestion is to call it: `STANDARD`
``ADAPTIVE``, ``DISTANCE``, ``STANDARD`` or ``NONE``

docs/ert/reference/configuration/keywords.rst
```diff
@@ -1396,7 +1400,7 @@ format. The surface keyword is configured like this:
 
 ::
 
-        SURFACE  ID  OUTPUT_FILE:surf.irap  INIT_FILES:Surfaces/surf<IENS>.irap  BASE_SURFACE:Surfaces/surf0.irap  FORWARD_INIT:True  UPDATE:TRUE
+        SURFACE  ID  OUTPUT_FILE:surf.irap  INIT_FILES:Surfaces/surf<IENS>.irap  BASE_SURFACE:Surfaces/surf0.irap  FORWARD_INIT:True  UPDATE:ADAPTIVE
```
xjules, line 1403: We should still support old format here; ie. UPDATE:TRUE | FALSE

docs/ert/reference/configuration/keywords.rst
```diff
@@ -1417,9 +1421,12 @@ format. The surface keyword is configured like this:
   Defaults to ``False``. See the section on initializing from the forward model below.
 
 - **UPDATE** (Optional)
-  Specifies whether this surface parameter should be included during the history matching
-  update step. Must be set to either ``TRUE`` or ``FALSE``.
-  Defaults to ``TRUE``.
+  Specifies the update strategy for this surface parameter during the history matching
```
xjules, line 1424: Revert

xjules:
To clarify:
each PARAM config type GEN_KW, FIELD, SURFACE will remain as it is now. 
This only reflects the internal update attribute, which previously was TRUE | FALSE and now it becomes 
```python
``ADAPTIVE``, ``DISTANCE``, ``STANDARD`` or ``NONE``
```
wherein for example:
```
GEN_KW COEFFS coeff_priors UPDATE:FALSE
``` 
would yield:
`param_cfg.update = None`

PR #13058, opened 2026-03-06
Updated the default name value for TargetEnsembleModel

### LINKED ISSUE #12829: Strange naming of ensembles in repeated es-mda runs

Running multiple esmda runs after each other have naming conventions to avoid duplicate names. However, it seems strange that suffixes from previous experiments carries over to the next experiment. See attached picture:

Image

I expected to have a "minimal" naming convention which still ensures unique ensemble names

I am sure there is a good reason it is the way it is, but perhaps we can find an even better solution

eilskra:
**Issue**
Resolves #12829

**Approach**
Updated the TargetEnsembleModel default name value to be `iter-%d` and not use ensemble name

image


eilskra:
Made changes to some tests and images, re-requesting review to ensure correctly

Issue #12787, opened 2026-02-04
Difficult to understand why an observation was disabled


eivindjahren:
If one realization is missing say the FOPR keyword in its summary and you have a SUMMARY_OBSERVATION for that key, then the observation will be disabled in the update step. However, it is not obvious why the observation was disabled, the report only says MISSING_RESPONSE but not in which realization(s) the response was missing.

There are at least two cases where we have much more information about why it was missing

1. The RFT observation did not fall within the zone. This currently shows one warning at the end of e.g. ESMDA, while the report during the experiment only shows. 

https://github.com/equinor/ert/blob/c58ee6bcff44e11474cb033a45d892a950a400d5/src/ert/config/rft_config.py#L122-L127

<img width="1419" height="1054" alt="Image" src="https://github.com/user-attachments/assets/0a8b05f3-f4c5-4219-bfdc-c55ca85b8bfc" />

2. The summary observation is either not in the summary file or not defined for the time step in the observation:

https://github.com/equinor/ert/blob/c58ee6bcff44e11474cb033a45d892a950a400d5/src/ert/config/summary_config.py#L40-L57

This should instead show exactly why the observation as deactivated in the report:

<img width="2295" height="940" alt="Image" src="https://github.com/user-attachments/assets/31c91dab-cc2e-4147-9995-ca8bb87b00d7" />

currently is only says: `Deactivated, missing responses`, but we could make that clickable and include information about why it was deactivated there. (Or use a similar solution). It needs to be quite obvious that it is clickable in that case, eg. use "external link" styling.

Currently, this check for missing responses is done here: 
https://github.com/equinor/ert/blob/7fc50c7d1ff401e684a04bb8b8b6cd1f62d08a60/src/ert/analysis/snapshots.py#L37

If we check this info from just the `ObservationStatus` enum to include a message:

```python
@dataclass
class MissingResponseEvent:
     ...
```
Then we can display that in the GUI. To send the message we can use a separate warning that can then go through the `send_event`:

https://github.com/equinor/ert/blob/c58ee6bcff44e11474cb033a45d892a950a400d5/src/ert/run_models/run_model.py#L394-L397

```python
class MissingResponseWarning(ErtWarning):
    """Warning about response missing for an observation"""

class MissingRFTResponseWarning(MissingResponseWarning):
    property: str
    time: datetime
    location: tuple[float, float, float]
    zone: str | None

```

```python

class RunModel(RunModelConfig, ABC):
        ...

[...]

PR #13066, opened 2026-03-09
Show which realizations deactivated observation

achaikou:
**Issue**
Relates to #12787 but does not fully resolve it as it does nothing to give users the reason of _why_ observation was disabled.

**Approach**
When observation is deactivated ("nan" status), clicking on the status opens a dialog window with missing realizations.

image

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/gui/experiments/view/update.py
+++ b/src/ert/gui/experiments/view/update.py
@@ -28,6 +28,7 @@ from PyQt6.QtWidgets import (
 from typing_extensions import override
 
 from ert.analysis.event import DataSection
+from ert.analysis.snapshots import ObservationStatus
 from ert.ensemble_evaluator import state
 from ert.run_models import (
     RunModelEvent,
@@ -79,6 +80,64 @@ class UpdateLogTable(QTableWidget):
                 )
         else:
             super().keyPressEvent(e)
+
+
+class ReportLogTable(UpdateLogTable):
+    def __init__(self, data: DataSection, parent: QWidget | None = None) -> None:
+        super().__init__(data, parent)
+
+        self.data = data
+
+        if "status" not in data.header:
+            raise RuntimeError("'status' column should be present in the report table")
+        if "missing_realizations" not in data.header:
+            raise RuntimeError(
+                "'missing_realizations' column should be present in the report table"
+            )
+
+        self.status_column_index = data.header.index("status")
+        self.missing_realizations_col_index = data.header.index("missing_realizations")
+
+        self.hideColumn(self.missing_realizations_col_index)
+
+        self.itemClicked.connect(self._handle_item_click)
+        self._underline_missing_realization_status()
+
+    def _underline_missing_realization_status(self) -> None:
+        """
+        Underline 'status' of observations with missing responses to indicate
+        that they are clickable.
+        """
+        for i, row in enumerate(self.data.data):
+            str_val = str(row[self.status_column_index])
+            if str_val == ObservationStatus.MISSING_RESPONSE:
+                item = self.item(i, self.status_column_index)
+                if item is not None:
+                    font = item.font()
+                    font.setUnderline(True)
+                    item.setFont(font)
+
+    @Slot(QTableWidgetItem)
+    def _handle_item_click(self, item: QTableWidgetItem) -> None:
+        if (
+            self.status_column_index == item.column()
+            and item.text() == ObservationStatus.MISSING_RESPONSE
+        ):
+            hidden_item = self.item(item.row(), self.missing_realizations_col_index)
+            assert hidden_item is not None, (
+                "all items in the table should have been initialized"
+            )
+
+            missing_realizations = hidden_item.text()
+            reasoning = "Missing responses from active realizations: " + str(
+                missing_realizations
+            )
+            QMessageBox.information(
+                self,
+                "Observation deactivated",
+                reasoning,
+                QMessageBox.StandardButton.Ok,
+            )
 
 
 class UpdateWidget(QWidget):
@@ -138,12 +197,13 @@ class UpdateWidget(QWidget):
         self,
         name: str,
         data: DataSection,
+        table_type: type[UpdateLogTable] = UpdateLogTable,
     ) -> None:
         widget = QWidget()
         layout = QVBoxLayout()
         widget.setLayout(layout)
 
-        table = UpdateLogTable(data)
+        table = table_type(data)
         table.setObjectName("CSV_" + name)
         layout.addWidget(table)
 
@@ -158,6 +218,12 @@ class UpdateWidget(QWidget):
             layout.addLayout(grid_layout)
 
         self._tab_widget.setCurrentIndex(self._tab_widget.addTab(widget, name))
+
+    def _insert_report_tab(
+        self,
+        data: DataSection,
+    ) -> None:
+        self._insert_table_tab("Report", data, ReportLogTable)
 
     @Slot(RunModelUpdateBeginEvent)
     def begin(self, event: RunModelUpdateBeginEvent) -> None:
@@ -177,7 +243,7 @@ class UpdateWidget(QWidget):
         self._progress_bar.setMaximum(1)
         self._progress_bar.setValue(1)
 
-        self._insert_table_tab("Report", event.data)
+        self._insert_report_tab(event.data)
 
     @Slot(RunModelDataEvent)
     def add_table(self, event: RunModelDataEvent) -> None:
--- a/tests/ert/ui_tests/gui/test_update_runs.py
+++ b/tests/ert/ui_tests/gui/test_update_runs.py
@@ -185,3 +185,49 @@ def test_that_autoscale_tab_is_displayed_in_es_update(
 
     for i, name in enumerate(expected_columns):
         assert update_log_table.horizontalHeaderItem(i).text() == name
+
+
+@pytest.fixture
+def poly_case_with_one_missing_response(source_root, tmp_path, run_experiment):
+    """
+    Sets up a poly case with missing response in realization 0 at indices
+    corresponding to observations 2 and 4.
+    """
+    _new_poly_example(source_root, tmp_path, 2)
+    config_path = tmp_path / "poly.ert"
+
+    with open_gui_with_config(config_path) as gui:
+        run_experiment(EnsembleExperiment, gui)
+        # observations exist at response indices 0, 2, 4, 6, 8
+        remove_responses_in_realization0(gui, indices_to_drop=[4, 8])
+        yield gui
+
+
+def test_that_report_table_is_displayed_on_missing_responses(
+    qtbot, poly_case_with_one_missing_response, run_experiment
+):
+    gui = poly_case_with_one_missing_response
+
+    run_experiment(ManualUpdate, gui, check_realizations=False)
+
+    run_dialog = gui.findChildren(RunDialog)[-1]
+    update_widget = run_dialog.findChild(UpdateWidget)
+    assert update_widget._tab_widget.count() == 2
+    assert update_widget._tab_widget.tabText(0) == "Status"
+    assert update_widget._tab_widget.tabText(1) == "Report"
+
+    update_log_table = update_widget._tab_widget.widget(1).findChild(UpdateLogTable)
+
+    status_index = update_log_table.data.header.index("status")
+    missing_realizations_index = update_log_table.data.header.index(
+        "missing_realizations"
+    )
+
+    expected_disabled_observations = [2, 4]
+    for index, row in enumerate(update_log_table.data.data):
+        if index in expected_disabled_observations:
+            assert row[status_index] == "nan"
+            assert row[missing_realizations_index] == "0"
+        else:
+            assert row[status_index] == "Active"
+            assert not row[missing_realizations_index]
--- a/tests/ert/unit_tests/gui/run_analysis/test_update_widget.py
+++ b/tests/ert/unit_tests/gui/run_analysis/test_update_widget.py
@@ -1,11 +1,15 @@
 from uuid import uuid4
 
 import numpy as np
-from PyQt6.QtWidgets import QTableWidget
+import pytest
+from PyQt6.QtCore import Qt, QTimer
+from PyQt6.QtWidgets import QMessageBox, QTableWidget
 from pytestqt.qtbot import QtBot
 
 from ert.analysis.event import DataSection
+from ert.analysis.snapshots import ObservationStatus
 from ert.gui.experiments.view import UpdateWidget
+from ert.gui.experiments.view.update import ReportLogTable
 from ert.run_models.event import RunModelDataEvent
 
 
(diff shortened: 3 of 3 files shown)


tests/ert/ui_tests/gui/test_docs_screenshots.py
```diff
@@ -292,8 +292,8 @@ def ignore_pngs(src, files):
 
         Path(config_file).write_text(combined_content, encoding="utf-8")
 
-    gui_generator = open_gui_with_config(tmp_path / config_file)
-    return next(gui_generator)
+    with open_gui_with_config(tmp_path / config_file) as gui:
+        return gui
```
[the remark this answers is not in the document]
achaikou: Hm. Apparently our code works fine even on closed gui.
Didn't want to introduce additional level of indentation in tests, so solved that via fixture.

tests/ert/ui_tests/gui/test_update_runs.py
```diff
@@ -0,0 +1,232 @@
...
+        dst_path = tmp_path
+        shutil.copytree(src_path, dst_path, dirs_exist_ok=True)
+
+    return _copy_case
+
+
+def remove_responses_in_realization0(gui, indices_to_drop):
+    """Removes responses in the first realization of the only ensemble.
+    indices_to_drop are expected to correspond to observation indices that must
+    be disabled.."""
```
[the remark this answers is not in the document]
achaikou: Will fix along with some other comments.

tests/ert/unit_tests/gui/run_analysis/test_update_widget.py
```diff
@@ -25,3 +29,160 @@ def test_update_widget(qtbot: QtBot):
...
+                message_box.button(QMessageBox.StandardButton.Ok),
+                Qt.MouseButton.LeftButton,
+            )
+
+    QTimer.singleShot(
+        500,
+        lambda: handle_disabled_observations_blocking_dialog(
+            qtbot, report_table, nan_message
+        ),
+    )
```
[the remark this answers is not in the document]
achaikou: I still have a bad understanding about what happens here, but likely no.

From what I understood, `QMessageBox` blocks the main event loop, so we need something that would work concurrently and dismiss the dialog. `singleShot` seems to be the proper solution

src/ert/analysis/_es_update.py
```diff
@@ -419,25 +415,11 @@ def log_warning(
...
-                    error_msg=str(e),
-                    data=DataSection(header=[], data=[]),
-                )
-            )
-        raise e
+        data = None
+        if isinstance(e, ErtAnalysisError):
+            data = e.data
+        progress_callback(AnalysisErrorEvent(error_msg=str(e), data=data))
+        raise
```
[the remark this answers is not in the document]
achaikou: Don't understand that comment, but probably wrong?

The point of one of the commits here was to assure that Report tab is displayed on failure when data is present.
The only chance for `smoother_snapshot` not being None is when execution completes without error.

src/ert/analysis/_enif_update.py
```diff
@@ -71,18 +70,11 @@ def enif_update(
...
-                    extra=smoother_snapshot.extra,
-                ),
-            )
-        )
-        raise e
+        data = None
+        if isinstance(e, ErtAnalysisError):
+            data = e.data
+        progress_callback(AnalysisErrorEvent(error_msg=str(e), data=data))
+        raise
```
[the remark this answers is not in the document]
achaikou: Probably wrong.
Preserving errors was exactly the point. Assuming that `ErtAnalysisError` is thrown only from inside of the library, we should be good?

tests/ert/ui_tests/gui/test_update_runs.py
```diff
@@ -0,0 +1,236 @@
...
+from ert.run_models.manual_update import ManualUpdate
+from tests.ert.ui_tests.gui.conftest import (
+    _new_poly_example,
+    get_child,
+    open_gui_with_config,
+)
+
+
+@pytest.mark.usefixtures("copy_poly_case")
+def test_no_updateable_parameters(qtbot):
```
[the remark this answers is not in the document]
achaikou: Not a part of this PR.
Test wasn't touched, only the file was renamed.

src/ert/gui/experiments/view/update.py
```diff
@@ -81,6 +82,61 @@ def keyPressEvent(self, e: QKeyEvent | None) -> None:
...
+        if (
+            self.status_column_index == item.column()
+            and item.text() == ObservationStatus.MISSING_RESPONSE
+        ):
+            missing_realizations = self.data.data[item.row()][
+                self.missing_realizations_col_index
+            ]
+            reasoning = "Missing responses from active realizations: " + str(
+                missing_realizations
+            )
```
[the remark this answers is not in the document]
achaikou: Very good point!
Fixed.

SAKavli:
Actually it might just me being too quick here, sorry.
Let me review this properly 👍

tests/ert/ui_tests/gui/conftest.py
```diff
@@ -265,7 +265,7 @@ def _ensemble_experiment_has_run(
 
 @pytest.fixture(name="run_experiment", scope="module")
 def run_experiment_fixture(request):
-    def func(experiment_mode, gui, click_done=True):
+    def func(experiment_mode, gui, wait_done=True, check_realizations=True):
```
SAKavli: To me, the `wait_done` and `check_realization` is still a bit confusing without reading the code, perhaps a docstring could be helpful here?

tests/ert/ui_tests/gui/conftest.py
```diff
@@ -299,27 +299,24 @@ def handle_dialog():
...
-            realization_widget = run_dialog._tab_widget.currentWidget()
-            assert isinstance(realization_widget, RealizationWidget)
-            list_model = realization_widget._real_view.model()
-            expected_num_realizations = (
-                experiment_panel.config.runpath_config.num_realizations
-            )
-            if experiment_mode.name() == "Single realization test-run":
-                expected_num_realizations = 1
-            assert list_model.rowCount() == expected_num_realizations
+            if check_realizations:
```
SAKavli, line 326: What was the motivation behind this conditioning?

tests/ert/ui_tests/gui/conftest.py
```diff
@@ -283,6 +288,12 @@ def func(experiment_mode, gui, wait_done=True, check_realizations=True):
         if hasattr(simulation_settings, "_ensemble_name_field"):
             simulation_settings._ensemble_name_field.setText("iter-0")
 
+        if experiment_mode.name() == "Manual update":
+            update_method = "ES Update"
```
SAKavli: If `kwargs` defaults to `{}` I believe this could be inlined as `kwargs.get("update_method", "ES Update")`?

Image

tests/ert/ui_tests/gui/test_missing_parameters_to_update.py
```diff
@@ -1,39 +0,0 @@
-import fileinput
```
SAKavli, line 1: I think it would be nice to rename in a separate commit, such that I don't accidentally start reviewing old code which seems to be new

SAKavli: aka `test_no_updateable_parameters` has not changed as far as I know

tests/ert/ui_tests/gui/test_update_runs.py
```diff
@@ -0,0 +1,146 @@
...
+    df = pl.read_parquet(realization0_response_file_path)
+
+    df = df.filter(~pl.Series(range(len(df))).is_in(indices_to_drop))
+    df.write_parquet(realization0_response_file_path)
+
+    gui.notifier.emitErtChange()
+
+
+@pytest.fixture(scope="module")
+def poly_example_with_missing_responses_dir(tmp_path_factory):
```
SAKavli, line 80: Could you explain to me what the benefit of this fixture is?

tests/ert/ui_tests/gui/test_update_runs.py
```diff
@@ -0,0 +1,146 @@
...
+    run_dialog = gui.findChildren(RunDialog)[-1]
+
+    update_widget = run_dialog.findChild(UpdateWidget)
+    assert update_widget._tab_widget.count() == 2
+    assert update_widget._tab_widget.tabText(0) == "Status"
+    assert update_widget._tab_widget.tabText(1) == "Report"
+
+    status_list = update_widget._tab_widget.widget(0).findChild(QListWidget)
+
+    count = 0
```
SAKavli: I believe this is possible to inlined as 
`count = sum(1 for i in range(status_list.count()) if "No active observations for update step" in status_list.item(i).text())`
but I respect if you think its as nice or nicer to write it out with indented for loops.

tests/ert/ui_tests/gui/test_update_runs.py
```diff
@@ -144,3 +144,44 @@ def test_that_report_table_is_displayed_on_no_active_observations(
...
+
+    with open(config_path, "a", encoding="utf-8") as f:
+        f.write("\nANALYSIS_SET_VAR OBSERVATIONS AUTO_SCALE *\n")
+
+    with open_gui_with_config(config_path) as gui:
+        run_experiment(EnsembleExperiment, gui)
+        yield gui
+
+
+def test_that_autoscale_tab_is_displayed_in_es_update(
```
SAKavli, line 163: Good catch, thanks for cleaning up after us 🧹

src/ert/analysis/_update_commons.py
```diff
@@ -267,17 +267,30 @@ def _compute_observation_statuses(
...
      * status of (the responses of) each observation,
        corresponding to ObservationStatus
+     * missing realizations that lead to deactivation
     """
 
     df_with_status = observations_and_responses
 
-    obs_has_null_response_ = pl.any_horizontal(
-        [pl.col(c).is_nan() | pl.col(c).is_null() for c in active_realizations]
+    missing_realizations_expr = (
```
SAKavli, line 280: This section was hard for me to understand. Might be a me-issue. Not sure how to make it more understandable, but thought I would leave a comment after starting for multiple minutes.

src/ert/gui/experiments/view/update.py
```diff
@@ -81,6 +82,64 @@ def keyPressEvent(self, e: QKeyEvent | None) -> None:
             super().keyPressEvent(e)
 
 
+class ReportLogTable(UpdateLogTable):
```
SAKavli, line 85: For this last commit, could you explain what problem you are solving for, since you mentioned you didn't solve the whole issue? And how to recreate this locally so I can inspect gui?

tests/ert/ui_tests/gui/conftest.py
achaikou: The original motivation is a mystery to me. Figuring it out and deciding if we can drop the check altogether / move it to kwargs is an issue on its own...

I was only interested in reusing this fixture in my new tests and those are on Manual Update, which does not have realization widget, so I had to split that out at least...

tests/ert/ui_tests/gui/test_missing_parameters_to_update.py
achaikou: Good point. 👍  This time git doesn't figure out this is the same file.
Could extract once we are closer to merging in order not to force-push? But maybe it would be too late by then 😄

tests/ert/ui_tests/gui/test_update_runs.py
achaikou: Maybe I don't see some simpler solution, but logic was as following:

Basically I want to create a directory somewhere that is used by two tests
`test_that_report_table_is_displayed_on_no_active_observations` for ES Update and EnIF Update.

This would save a few seconds so that prerequisite experiment with missing responses is not re-created in both tests, but is created just once before all the tests run.

It needs to be created in non-test general directory in the pytest folder - hence `tmp_path_factory.mktemp` and then used in tests.

So I could solve it either by creating this small directory-creation fixture that I next can use both in `setup_poly_case_with_missing_response_for_each_observation` fixture and in the test, or I had to manually combine `tmp_path_factory` and folder name both in fixture and test to get the final result.

Fixture seemed simpler 🤷‍♀️ .

tests/ert/ui_tests/gui/test_update_runs.py
achaikou: Hm... `sum` function exist. Who would have thought 😄 
No, your suggestion is nicer as it is more focused 🙂

tests/ert/ui_tests/gui/test_update_runs.py
achaikou: There were some autoscale tests already.
It was just that I made wrong decision in the next commit and didn't notice I broke autoscale UI, so I had to add more tests 🙂

src/ert/analysis/_update_commons.py
achaikou: I only did it with the help of the AI, so I get it.

I am not in favor of what-comments, but here they are warranted then. Adding something 🙂 .

src/ert/gui/experiments/view/update.py
achaikou: This PR only shows which realizations disabled observation. Eivind said showing which realization disabled observation is already important information for the user.
If we get more information later, we would be able to add it there too.

You manually recreate it next way:
- have a `.config` with at least two observations. You want one to be untouched - Active, so that update doesn't fail with "no active observations found" (unless you want to test that I didn't screw up Report table for that case - related to [Assure AnalysisErrorEvent is sent once in updates](https://github.com/equinor/ert/pull/13066/changes/d5d55727930cf0aadca94561cc663d50403a1bd7) commit )
- run Ensemble Experiment
- go to some realization directories - the ones you want to be show as missing - and open .parquet file with the responses.
- delete responses which correspond to observation that you want to disable and save the file back
- run Manual Update based on this Ensemble Experiment - you should then see `nan` in the status column

tests/ert/ui_tests/gui/conftest.py
achaikou: Would fixup to a different commit that already has docs 👍

achaikou (commented):
Thank you sander for taking up this review 😻

tests/ert/ui_tests/gui/test_missing_parameters_to_update.py
SAKavli: Yes, I don't think git is able to detect this. In my previous team we had a policy of keeping renaming of files to own atomic commits.

tests/ert/ui_tests/gui/test_update_runs.py
SAKavli: Ah, so have you verified that this will reuse the same `tmp_path`? Then it makes a lot sense!

src/ert/analysis/_update_commons.py
SAKavli: Perhaps comment, but I was thinking whether we could split the operations into separate parts and store the results in descriptive variables. Then there might not be a need for comments.

Could also consider extracting this to a method such that we could have some tests clearly describing input/output expectation of this unit.

achaikou: Tried to split, failed.
It is difficult to explain what this code produces, so I can't name it helpfully.

As for splitting, I can extract
```
  pl.when(pl.col(c).is_nan() | pl.col(c).is_null())
  .then(pl.lit(c))
  .otherwise(pl.lit(None))
  for c in active_realizations
```

but I have no clue what that really is. This is some kind of rule that will set value in the cell either to column name or to None. But how does one name it?..

And this whole expression is some kind of rule about how to deal with data. Before actual dataframe is provided, it is pointless.
So to test it, you need a dataframe, as far as I can tell...
So the closest unit tests that make sense to me are the ones in `test_es_update`

So I see comment as my best option. :woman_shrugging:

tests/ert/ui_tests/gui/test_update_runs.py
achaikou: Yes, the content of the pytest directory when running this test is
```

poly_example_with_missing_responses_for_each_observation0
test_that_report_table_is_disp1
poly_example_with_missing_responses_for_each_observationcurrent
test_that_report_table_is_dispcurrent
test_that_report_table_is_disp0
```

with contents of `poly_example_with_missing_responses_for_each_observation0` copied into each test directory.

PR #13067, opened 2026-03-09
Remove timestamp from experiment name

### LINKED ISSUE #12954: Batch number not visible in GUI
The actual number of the batch is not visible in this GUI, only the constant string `EnOpt@2026-02-18T12:24:44` which is not as interesting as the batch number:

Image

eqbech:
Removes the same timestamp from from the gui side if `is_everest` is true else it keeps the orginal format.

**Issue**
Resolves #12954

new proposed changes on the ui looks like this:
Screenshot from 2026-03-11 08-08-38


andreas-el:
Please add a screenshot how this looks like

PR #13079, opened 2026-03-10
Improve experiment_config typing

### LINKED ISSUE #12803: Make LocalExperiment.experiment_config typed
This should be typed to improve readability and support future API doc generation. Split out experiment configs into own file so they can be imported from ERT storage and run models without introducing circular imports

jonathan-eq:
**Issue**
Resolves https://github.com/equinor/ert/issues/12803

**Approach**
This commit changes LocalExperiment.experiment_config from dict[str,Any]
to a typeddict - ExperimentConfig which is a dict with all the potential
keys possible. This commit also makes the experiment config models
export to this typeddict, instead of doing model_dump(...).

      --exec 'just rapid-tests'`)

## When applicable


src/ert/storage/local_experiment.py
jonathan-eq, line 493: I don't get this. Why are we getting observations from self.experiment_config, when not all run model configs have observations from the start (like EverestRunModelConfig). Should this be self.experiment or self.stored_experiment instead of experiment_config? It wasnt a problem when using a TypedDict that may contain all the possible keys, but if I wanted to use a union type of all the RunModelConfigs, this was more complex.

src/ert/run_models/ensemble_smoother.py
```diff
@@ -42,7 +39,7 @@ def run_experiment(
         self.run_workflows(fixtures=PreExperimentFixtures(random_seed=self.random_seed))
 
         experiment_storage = self._storage.create_experiment(
-            experiment_config=self.model_dump(mode="json"),
+            experiment_config=self.to_experiment_config(),
```
jonathan-eq, line 43: model_dump() returns a dict[str,Any]. I changed it to `to_experiment_config()` which returns the typeddict ExperimentConfig.

src/ert/experiment_configs.py
```diff
@@ -0,0 +1,293 @@
+from __future__ import annotations
```
frode-aarstad, line 1: Perhaps run_model_configs.py is a better name for this file? or move it into ert.config ?

jonathan-eq: I can change it to run_model_configs.py 👌

tests/ert/unit_tests/run_models/test_experiment_serialization.py
```diff
@@ -56,7 +57,6 @@
     MultipleDataAssimilation,
     create_model,
 )
```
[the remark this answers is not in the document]
jonathan-eq: Good idea!

jonathan-eq:
> I think this one needs a dependency graph.

What is that?

tests/ert/unit_tests/test_run_path_creation.py
```diff
@@ -1196,9 +1212,10 @@ def test_that_generate_parameter_files_raises_when_end_event_is_set(storage):
     )
     experiment_id = storage.create_experiment(
         experiment_config={
-            "parameter_configuration": (
-                ert_config.ensemble_config.parameter_configuration
-            ),
+            "parameter_configuration": [
```
xjules, line 1215: I don't understand; ie. why is this necessary?

jonathan-eq: pydantic basemodel validates that this should be a dict[str,Any], but in the test it is list[ParameterConfig]. This worked before as it was type Any.

jonathan-eq: We had a discussion earlier where we found out it accepted both previously, and that was due to the very relaxed typing.

src/ert/run_models/run_model_configs.py
```diff
@@ -0,0 +1,293 @@
...
+    hooked_workflows: defaultdict[HookRuntime, list[Workflow]]
+    active_realizations: list[bool]
+    log_path: Path
+    random_seed: int
+    start_iteration: int = 0
+    minimum_required_realizations: int = 0
+    supports_rerunning_failed_realizations: ClassVar[bool] = False
+
+
+class InitialEnsembleRunModelConfig(RunModelConfig):
```
xjules, line 116: Should this one implement `to_experiment_config` too?

jonathan-eq: It already implements `_initial_ensemble_experiment_config`, which is called in the class EnsembleInformationFilterConfig(
    InitialEnsembleRunModelConfig, UpdateRunModelConfig
) to_experiment_config method. We could maybe implement it as `to_experiment_config() here as well and call super().to_experiment_config(), but pydantic blows up if we try it.

xjules (approved):
Nice work @jonathan-eq ! Don't forget to squash commits.

PR #12445, opened 2025-12-03
Add plots for summary and gen data misfits

yngve-sk:
**Issue**
Resolves [11074](https://github.com/equinor/ert/issues/11074)

Snake oil, 201reals gendata
Screenshot 2025-12-12 at 14 21 00

Snake oil, 201reals, FOPR with history obs
Screenshot 2025-12-12 at 14 21 42

Snake oil, 201reals, fewer observations
Screenshot 2025-12-12 at 14 22 10

Snake oil, 101reals, gendata
Screenshot 2025-12-12 at 14 26 13

Snake oil, 101reals, FOPR with history obs
Screenshot 2025-12-12 at 14 26 35

Snake oil, 101reals, fewer observations
Screenshot 2025-12-12 at 14 26 56

Snake oil, 50 reals, gendata
Screenshot 2025-12-12 at 14 29 31

Snake oil, 50 reals, FOPR with history obs
Screenshot 2025-12-12 at 14 30 05

Snake oil, 50 reals, fewer observations
Screenshot 2025-12-12 at 14 30 22

Issue #11074, opened 2025-06-10
Add some webviz_ert plots to ERT plotter (then remove webviz ert)


yngve-sk:
Following plots should be moved to the ert plotter:

Misfits: (univariate)
<img width="1589" alt="Image" src="https://github.com/user-attachments/assets/10abd8fd-5533-40cb-97a3-9a2e9d387838" />

Misfits (summary)
<img width="1571" alt="Image" src="https://github.com/user-attachments/assets/1fc326e6-afed-46b3-a557-fcc39cf1a6c9" />

Following will not be ported(?):
Response correlation views
Parameter correlation views

PR #13080, opened 2026-03-10
Unify multiple panels for misfit plots

SAKavli:
I've worked with the misfit plotters, and come to the conclusion that having two separate plotters should not be necessary. I've kept the one which looks the nicest.

Old:

image

Also old:

image

New behaviour for all misfit plots:

image

image


src/ert/gui/tools/plot/plottery/plots/misfits.py
achaikou, line 180: Irrelevant for this PR, but may be relevant for some other task:
I think if this results in an empty dictionary, the plot widget becomes empty and exception gets logged (because of later calculations).

Cannot find when "Misfits" is supposed to be greyed-out, so can't see where the difference is :woman_shrugging: .

src/ert/gui/tools/plot/plottery/plots/misfits.py
```diff
@@ -158,21 +157,13 @@ def plot(
...
-                plot_context,
-            )
-
-        elif response_type in {"gen_data", "rft"}:
-            self._plot_gendata_misfits(
-                figure,
-                data_with_misfits,
-                plot_context,
-            )
+        self._misfit_boxplot(
```
achaikou, line 160: I struggle with confirming validity of this change.
The commit message does not tell me enough about why it is fine. 

Both plots were introduced in the same PR #12445 done by Yngve.
This leads me to assume that plots were different for a reason.
In the PR he shows many nice different plots and what they look like (and for some reason focusing on number of realizations).

On your PR only the basic plots are shown.
I myself can't reproduce all nice plots from Yngve (am I supposed to add FOPR observations to snake_oil manually myself?)

But from the look of it, in the previous Summary one it seems that the point was to create a real-world timeline (ticks are 1st Jan 2010, 1st Jan 2011, 1st Jan 2012...) and see how misfits changed for each assemble over the timeline.

Image


If the plots are merged, dates become just indices on the timeline and all ensembles are displayed together. Also the values for the last observation seem vastly different, for some reason.
So it creates a different picture.

Image

That leads me to a question if this change doesn't destroy something useful. 

Can we find out what the original point of two separate plots was and confirm if it no longer applies?

src/ert/gui/tools/plot/plottery/plots/misfits.py
```diff
@@ -194,13 +192,11 @@ def _misfit_boxplot(
             sorted_ensemble_keys=sorted_ensemble_keys,
         )
 
-        # Create subplot grid (2 rows, N columns)
-        axes = figure.subplots(
-            nrows=2, ncols=num_gendata_index, sharex="col", sharey=True
-        )
+        num_cols = len(distinct_gendata_index)
```
achaikou, line 195: If you are changing these to support more than just gendata (but breakthrough too?) - maybe all "gendata" references should be removed from here?

Also - I see that this is used as a number of columns, but I struggle to understand what that actually is. Number of ensembles to display for each index? Number of different index points for which observations are available?

So would prefer a more telling name :smirk_cat:

src/ert/gui/tools/plot/plottery/plots/misfits.py
```diff
@@ -273,38 +251,74 @@ def _misfit_boxplot(
         )
         figure.tight_layout(rect=(0.02, 0.02, 0.98, 0.88))
 
-    def _style_boxplots(
+    def _create_scatterplot(
+        self, ax: Axes, color: str, mis_vals: Any, x_center: Any
+    ) -> None:
+        """This really creates a stripplot (1D scatterplot) with 'jitter' (spread
+        accross 2D), but I think it will be less confusing for users and developers if
```
achaikou, line 258: I think declarative style is more preferred in the docs/comments than using I/we, especially when product ownership can be passed from one team to another.

Also per seaborn docs stripplot is a categorical scatterplot, so if we are talking only in the context of misfits, there are no other scatterplots, so we can safely use the scatterplot name without excuses 😄

src/ert/gui/tools/plot/plottery/plots/misfits.py
```diff
@@ -273,38 +251,74 @@ def _misfit_boxplot(
...
+            ax.grid(True, axis="y", linestyle=":", alpha=0.4)
+
+            is_first_col = c_idx == 0
+
+            # Visualize only left-most and bottom-most borders
+            ax.spines["top"].set_visible(False)
+            ax.spines["right"].set_visible(False)
+            ax.spines["left"].set_visible(is_first_col)
+
+            # Disable ticks on other than left-most and bottom-most borders
```
achaikou, line 320: So you also remove ticks for indices because all ensembles in the group are supposed to be related to the same index and thus ticks become confusing?
Me likes it! 🐱 

But you probably need to update the comment here :)

src/ert/gui/tools/plot/plottery/plots/misfits.py
```diff
@@ -221,46 +226,19 @@ def _misfit_boxplot(
...
-                )
-                ax_top.plot(x_center, np.mean(mis_vals), "o", markersize=4, color=color)
-
-                num_points = len(mis_vals)
-
-                if num_points >= 200:
-                    marker_size = 2
-                elif num_points >= 100:
-                    marker_size = 3
+                if plot_context.misfit_boxplot:
```
achaikou, line 229: I am not sure if this change is what we want (as from the issue #11074 we see that webviz-ert for some reason explicitly had those two side-by-side), but if we can confirm that we do, I really like it 🐱

src/ert/gui/tools/plot/plottery/plot_context.py
```diff
@@ -46,6 +46,7 @@ def __init__(
         self._x_axis: str | None = None
         self._y_axis: str | None = None
         self._log_scale = False
+        self.misfit_boxplot = True
```
achaikou, line 49: Should we follow the implicit class guidelines?
Log scale, thing similar to `mistfit_boxplot`, is a private attribute with public access via properties. So are other attributes.

Should we do the same for `misfit_boxplot`?

(Just wondering, not insisting 😄  )

src/ert/gui/tools/plot/plottery/plots/misfits.py
```diff
@@ -32,6 +34,10 @@ def __init__(self) -> None:
         self.dimensionality = 2
         self.requires_observations = True
 
+        self._toggle_boxplot = QRadioButton()
+        self._toggle_boxplot.setChecked(True)
+        self._toggle_scatterplot = QRadioButton()
```
achaikou, line 39: Are you using those? vscode is unable to find references.

And without those this class is nicely qt-independent.

src/ert/gui/tools/plot/plot_widget.py
```diff
@@ -162,12 +164,25 @@ def __init__(
...
+        self._misfit_boxplot_button.setVisible(False)
+        self._misfit_scatterplot_button.setVisible(False)
+        self._misfit_boxplot_button.setChecked(True)
+        self._misfit_boxplot_button.toggled.connect(self.requestUpdatePlot.emit)
+        self._misfit_scatterplot_button.toggled.connect(self.requestUpdatePlot.emit)
+
+        button_row = QHBoxLayout()
+        button_row.addWidget(self._log_checkbox)
+        button_row.addWidget(self._misfit_boxplot_button)
+        button_row.addWidget(self._misfit_scatterplot_button)
```
achaikou, line 181: With current design where buttons parent widget is `button_row` I am wondering if we shouldn't from the start put those buttons into their own `QButtonGroup`, for future safety.

> Radio buttons are autoExclusive by default. If auto-exclusive is enabled, radio buttons that belong to the same parent widget behave as if they were part of the same exclusive button group. If you need multiple exclusive button groups for radio buttons that belong to the same parent widget, put them into a [QButtonGroup](https://doc.qt.io/qtforpython-6/PySide6/QtWidgets/QButtonGroup.html#PySide6.QtWidgets.QButtonGroup)


Or better - add buttons to `button_row` conditionally, depending on the plotter type. Then misfit would have only misfit buttons, the "log" ones - the log button and everything else - nothing.

tests/ert/ui_tests/gui/test_scatterplot_visualization.py
```diff
@@ -0,0 +1,16 @@
...
+)
+
+plot_figure = create_breakthrough_figure(MISFITS, select_scatterplot=True)
+
+
+@pytest.mark.mpl_image_compare(tolerance=10.0)
+@pytest.mark.skip_mac_ci
+@pytest.mark.snapshot_test
+@pytest.mark.filterwarnings("ignore:Config contains a SUMMARY key")
+def test_that_scatterplot_visualization_matches_snapshot(plot_figure):
```
achaikou, line 15: I wonder if we need this yet another GUI test (which, if I understand correctly only checks that something renders).
Could we do with a short `PlotWindow` unit-test instead, that would also check that something returns without errors for various radio-buttons?

Or is directly setting experiment data or something too ugly and thus unit-testing is too messy here?

(I, btw, prefer e2e-like tests over unit ones, so I definitely won't mind if this stays a GUI-test :smile_cat:  )

src/ert/gui/tools/plot/plot_widget.py
```diff
@@ -198,6 +209,11 @@ def _sync_log_checkbox(self) -> None:
         else:
             self._log_checkbox.setVisible(False)
 
+    def _sync_misfit_radio_buttons(self) -> None:
```
achaikou, line 212: We need to sync log checkbox because the data could be negative.
For misfits we want to display those buttons unconditionally.

So shouldn't these buttons just be added to the footer and be visible from the start in `init` when
` if plotter_type == "MisfitsPlot":`

Then syncing won't be needed.

src/ert/gui/tools/plot/plottery/plots/misfits.py
SAKavli: You are very correct, thank you for pointing this out!

src/ert/gui/tools/plot/plottery/plots/misfits.py
SAKavli: I believe it's the flag in `misfits.py` : `self.requires_observations = True` which determines whether the window should be active or not.

src/ert/gui/tools/plot/plottery/plots/misfits.py
SAKavli: Indeed, thanks @achaikou !
It confused me quite a bit also which is why I wanted to do some simplifications to these plots.

All misfit DataFrames are identical, so I thought it would make sense to have one misfit plotter to support DataFrames of misfit values.

I changed these to `distinct_observations` and `num_plots`.

My change removes the row concept, which makes `num_cols` slightly confusing.

src/ert/gui/tools/plot/plottery/plots/misfits.py
SAKavli: Ahhhh, I am afraid you are right.
We will see if this will be scrapped.

src/ert/gui/tools/plot/plottery/plots/misfits.py
SAKavli: Thanks for pointing this out! I will remove this 🥸

src/ert/gui/tools/plot/plottery/plots/misfits.py
SAKavli: I've been adding some more understandable comments to the existing code. The behavior have not changed, here is with vs without this code:

Image

Image

The comment could still be a little more understandable, I changed it to:
`# Enable ticks only for left-most and bottom-most borders`

src/ert/gui/tools/plot/plottery/plot_context.py
SAKavli: Good point!

src/ert/gui/tools/plot/plot_widget.py
SAKavli: I like the way you think, I will throw them into a `QButtonGroup` for now!

SAKavli: Seems to work, at first I thought I would need to add the buttongroup to the QHBoxLayout, but apparently you just need to add them to the button group without anything else for it to work.

https://stackoverflow.com/questions/52880441/in-qt-designer-qbuttongroup-is-not-available-in-the-widget-box

Image

Issue #13101, opened 2026-03-11
Test for setting of random seed via GUI is flaky on Komodo


ajaust:
[This test](https://github.com/equinor/ert/blob/fd15e0160540a4879465f103140498d27300114f/tests/ert/ui_tests/gui/test_setting_random_seeds.py#L36-66) tends to fail sometimes on the Komodo test suite.

Some observations/notes:

* The test often fails when run as part of the Komodo tests, but seems to be fine in the ERT repo.
* The test tends to fail much more often on the Azure runner. `onprem` works most of the time. Tests indicate that the tests become especially flaky if there is little compute power available, e.g., when there is high load on the runner. He could also reproduce the issue on his local machine with `nice pytest tests/ert/ui_tests/gui/test_setting_random_seeds.py::test_that_gui_generates_different_seeds_for_consecutive_runs --count 200 -n 36`.
* The test sometimes passes.
* There test fails with different types of failures, but they all seem to be related to reaching some timeout. The error message seems to mainly vary due to timing when the test is killed.
* Type of failures 
    * [Detected differences in environment](https://github.com/equinor/komodo-releases/actions/runs/22836605164/job/66235706803). This seems to be the most common error in recent failures.
    * [Fatal Python Error](https://github.com/equinor/komodo-releases/actions/runs/22747068633/job/65974004596)
    * [QtBot timeout reached](https://github.com/equinor/komodo-releases/actions/runs/22719407480/job/65886043816)
* The test sometimes fails due to timeout already on the first `run_experiment`, e.g., [in this test run](https://github.com/equinor/komodo-releases/actions/runs/22719407480/job/65886043816). It indicates it not necessarily an issue with running the experiment twice.
* On my laptop the tests usually pass within a few seconds (<10s) so the test itself should not be heavy in terms of computations.

PR #13107, opened 2026-03-11
Increase timeout of flaky GUI experiment test

ajaust:
We observed flaky behavior when this test is run during a Komodo build. The test is reproducible when putting load on a developer machine, but in that case it is possible to hit almost any timeout. This might be partially related to the fact that `fm_dispatch` sets the process nice value to 19 (=low priority).

We move some of the logic to run the test into the test such that we can increase the timeout for individual test runs. The test runs two experiments with individual timeouts increased to 250 s. The global pytest timeout is 360 s so we increase it to 600 s.

**Issue**
Resolves https://github.com/equinor/ert/issues/13101

**Approach**
Extend the test timeout to see if this reduces flakiness.

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/tests/ert/ui_tests/gui/test_setting_random_seeds.py
+++ b/tests/ert/ui_tests/gui/test_setting_random_seeds.py
@@ -5,9 +5,10 @@ from textwrap import dedent
 
 import pytest
 
+from ert.gui.experiments import RunDialog
 from ert.run_models import EnsembleExperiment, SingleTestRun
 
-from .conftest import _open_main_window
+from .conftest import _open_main_window, get_children
 
 
 @pytest.mark.parametrize("experiment_type", [SingleTestRun, EnsembleExperiment])
@@ -33,6 +34,7 @@ def test_that_gui_uses_config_random_seed_when_specified(
     assert "'random_seed': 12345" in seed_logs[0]
 
 
+@pytest.mark.timeout(700)
 @pytest.mark.parametrize("experiment_type", [SingleTestRun, EnsembleExperiment])
 def test_that_gui_generates_different_seeds_for_consecutive_runs(
     run_experiment, use_tmpdir, qtbot, caplog, experiment_type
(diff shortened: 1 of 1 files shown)


tests/ert/ui_tests/gui/test_setting_random_seeds.py
```diff
@@ -45,18 +47,34 @@ def test_that_gui_generates_different_seeds_for_consecutive_runs(
...
+            lambda: run_dialog._tab_widget.currentWidget() is not None, timeout=10000
+        )
+
     with (
         caplog.at_level(logging.INFO),
         _open_main_window("config.ert") as (gui, _, _),
     ):
-        run_experiment(experiment_type, gui)
+        qtbot.addWidget(gui)
+        run_experiment(experiment_type, gui, click_done=False)
```
achaikou, line 66: Sorry, `click_done` does not exist in the conftest anymore 😇
So with current code old timeouts still would apply.

Try `wait_done=False` instead.
Or it might be worth combining `wait_for_experiment_completion` with `run_experiment` via adding timeout to `kwargs` to not multiply the logic.

tests/ert/ui_tests/gui/test_setting_random_seeds.py
```diff
@@ -33,6 +34,7 @@ def test_that_gui_uses_config_random_seed_when_specified(
     assert "'random_seed': 12345" in seed_logs[0]
 
 
+@pytest.mark.timeout(700)
 @pytest.mark.parametrize("experiment_type", [SingleTestRun, EnsembleExperiment])
 def test_that_gui_generates_different_seeds_for_consecutive_runs(
```
achaikou, line 39: Note that I don't think we know for 100% which tests fails. In the runs you shown both are present.
It might be just `test_that_gui_uses_config_random_seed_when_specified` or both.
https://github.com/equinor/komodo-releases/actions/runs/22909021734/job/66475799672

tests/ert/ui_tests/gui/test_setting_random_seeds.py
```diff
@@ -33,6 +34,7 @@ def test_that_gui_uses_config_random_seed_when_specified(
     assert "'random_seed': 12345" in seed_logs[0]
 
 
+@pytest.mark.timeout(700)
```
achaikou, line 37: I am not sure I buy that. How is it supposed to help? Do we expect this test to take THAT long?

PR #13116, opened 2026-03-12
Update total progress to account for inactive realizations

### LINKED ISSUE #13018: Total progress label starts at weird number on Ensemble Experiment

When having NUM_REALIZATIONS in config and specifying 0-4 when running EnsembleExperiment, the totalprogress at the top of the GUI starts at 95%. It seems it counts the final 95 realizations that were unselected as already completed. This does not seem so intuitive. 

We are only running five realizations, so it should go 1-5. The other number maybe makes sense when running esmda, and the previous iteration had some realizations failing (and not running anymore). 

### Steps to reproduce

```console
Modify poly_eval.py to time.sleep(30). Select 0-4 realizations. Run ensemble experiment.
```

- [ ] Python 3.11
- [ ] Python 3.12
- [ ] macOS
- [ ] RHEL8
- [ ] RHEL9
- [ ] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

erlenlh:
**Issue**
Resolves #13018 

**Approach**
Instead of progress of all realiztions in the setup, we instead check progress compared only to active realizations
Meaning if you had NUM_REALIZATIONS set to 100, but use a custom mask of 0-4 reals, then you would start at 95% complete with the old way. Pictures before and after of said experiment: 

BEFORE:
bilde

AFTER:
bilde

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/run_models/run_model.py
+++ b/src/ert/run_models/run_model.py
@@ -538,29 +538,74 @@ class RunModel(RunModelConfig, ABC):
 
     def calculate_current_progress(self) -> float:
         current_iter = max(list(self._iter_snapshot.keys()))
-        done_realizations = self.active_realizations.count(False)
         all_realizations = self._iter_snapshot[current_iter].reals
         current_progress = 0.0
 
+        filtered_active_realizations = [
+            active
+            for active, initial in zip(
+                self.active_realizations,
+                self._initial_realizations_mask,
+                strict=False,
+            )
+            if initial
+        ]
+
+        initial_active_indices = np.flatnonzero(self._initial_realizations_mask)
+        total_initial = len(initial_active_indices)
+        if total_initial == 0:
+            return current_progress
+
         if all_realizations:
-            for real in all_realizations.values():
-                if real["status"] in {
-                    REALIZATION_STATE_FINISHED,
-                    REALIZATION_STATE_FAILED,
-                }:
-                    done_realizations += 1
+            initial_id_str = {str(i) for i in initial_active_indices}
+            done_realizations = [
+                (real_id in initial_id_str)
+                and (
+                    real.get("status")
+                    in {REALIZATION_STATE_FINISHED, REALIZATION_STATE_FAILED}
+                )
+                for real_id, real in all_realizations.items()
+            ].count(True)
 
-            realization_progress = float(done_realizations) / len(
-                self.active_realizations
-            )
+            if self._total_iterations != 1:
+                start_iter = min(self._iter_snapshot)
 
-            current_it_offset = current_iter - min(list(self._iter_snapshot.keys()))
+                completed_realizations = 0
+                for it in range(start_iter, current_iter):
+                    iter_reals = self._iter_snapshot[it].reals
+                    if iter_reals:
+                        completed_realizations += [
+                            (real_id in initial_id_str) for real_id in iter_reals
+                        ].count(True)
+                    else:
+                        completed_realizations += total_initial
 
-            current_progress = (
-                (current_it_offset + realization_progress) / self._total_iterations
-                if self._total_iterations != 1
-                else realization_progress
-            )
+                current_iter_attempted = [
+                    (real_id in initial_id_str) for real_id in all_realizations
+                ].count(True)
+
+                current_iter_active = [
+                    (real_id in initial_id_str)
+                    and (real.get("status") != REALIZATION_STATE_FAILED)
+                    for real_id, real in all_realizations.items()
+                ].count(True)
+
+                remaining_iters = start_iter + self._total_iterations - current_iter - 1
+                total_realizations = (
+                    completed_realizations
+                    + current_iter_attempted
+                    + remaining_iters * current_iter_active
+                )
+
+                if total_realizations == 0:
+                    current_progress = 1.0
+                else:
+                    current_done = completed_realizations + done_realizations
+                    current_progress = float(current_done) / float(total_realizations)
+            else:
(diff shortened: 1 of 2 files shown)


tests/ert/unit_tests/run_models/test_base_run_model.py
```diff
@@ -536,8 +543,18 @@ async def send_terminate(end_event) -> bool:
...
             id="progress_with_another_extended_offset_iterations",
         ),
+        pytest.param(
+            {"0": "Running", "1": "Finished", "2": "Finished", "3": "Finished"},
+            0,
+            0,
+            1,
+            0.5,
+            [True, True, False, False],
+            id="progress_uses_initial_active_mask_ignores_non_active_reals",
```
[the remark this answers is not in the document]
erlenlh: ignoring this, name is in convention with neighbours

src/ert/run_models/run_model.py
```diff
@@ -538,22 +538,28 @@ def get_memory_consumption(self) -> int:
...
-        done_realizations = self.active_realizations.count(False)
         all_realizations = self._iter_snapshot[current_iter].reals
         current_progress = 0.0
 
+        initial_active_indices = [
+            i
+            for i, is_active in enumerate(self._initial_realizations_mask)
+            if is_active
+        ]
+        if (total_initial := len(initial_active_indices)) == 0:
```
erlenlh: Usikker på return her, og om man trenger denne sanity-checken i det hele tatt. Vet ikke om det er mulig å komme hit uten _initial_realization_mask

jonathan-eq: I think this reads a bit better without the walrus operator.
```
total_initial = len(initial_active_indices)
if total_initial == 0:
   return current_progress
```

src/ert/run_models/run_model.py
```diff
@@ -538,22 +538,28 @@ def get_memory_consumption(self) -> int:
 
     def calculate_current_progress(self) -> float:
         current_iter = max(list(self._iter_snapshot.keys()))
-        done_realizations = self.active_realizations.count(False)
         all_realizations = self._iter_snapshot[current_iter].reals
         current_progress = 0.0
 
+        initial_active_indices = [
```
xjules:
```suggestion
        initial_active_indices = np.flatnonzero(self._initial_realizations_mask)
```

src/ert/run_models/run_model.py
```diff
@@ -538,22 +538,28 @@ def get_memory_consumption(self) -> int:
...
-            realization_progress = float(done_realizations) / len(
-                self.active_realizations
+            initial_id_str = {str(i) for i in initial_active_indices}
+            done_realizations = sum(
+                1
+                for real_id, real in all_realizations.items()
+                if real_id in initial_id_str
+                and real.get("status")
+                in {REALIZATION_STATE_FINISHED, REALIZATION_STATE_FAILED}
             )
```
jonathan-eq: Hmm, summing 1 for each active realization reads a bit difficult...
Can we do 
```
[
real_id in initial_id_str 
and real.get("status") in {REALIZATION_STATE_FINISHED, REALIZATION_STATE_FAILED} 
for real_id,real in all_realizations.items()
].count(True)
```
I am not sure, but for me that reads a bit better

src/ert/run_models/run_model.py
erlenlh: Agree, always a tradeoff between "efficient" and readable code!

src/ert/run_models/run_model.py
erlenlh: thanks, much cleaner 👍

src/ert/run_models/run_model.py
erlenlh: Agree, still a mouthfull but better indeed

src/ert/run_models/run_model.py
```diff
@@ -537,30 +537,65 @@ def get_memory_consumption(self) -> int:
...
+                )
+                for real_id, real in all_realizations.items()
+            ].count(True)
+
+            realization_progress = float(done_realizations) / float(total_initial)
+
+            if self._total_iterations != 1:
+                start_iter = min(self._iter_snapshot)
+
+                completed_work_units = 0
```
jonathan-eq: Is this the total count of realizations?

erlenlh: there is "total_work" on line 586 (as i type this, i realize it should probably be named total_realizations or something to that effect) that estimates how many total realizations that is expected to be run. Based on already completed realizations and how many iterations are left. 

Completed_work_units is a measure of how many realizations have already been completed (success/fail)

jonathan-eq: Okay, very good :)

src/ert/run_models/run_model.py
```diff
@@ -538,29 +538,74 @@ def get_memory_consumption(self) -> int:
...
+        if total_initial == 0:
+            return current_progress
 
-            current_progress = (
-                (current_it_offset + realization_progress) / self._total_iterations
-                if self._total_iterations != 1
-                else realization_progress
-            )
+        if all_realizations:
+            initial_id_str = {str(i) for i in initial_active_indices}
```
xjules, line 560: Is there a reason that we need this as str?

erlenlh: Its because we use "real_id" for comparison with "all-realizations.items()" on line 567, where every entry is a string. "real_id" is derived from "initial_id_str". But I'm open to other ways of doing this!

PR #13137, opened 2026-03-13
Do not emit config warnings from the everest experiment server

### LINKED ISSUE #13124: Deprecation warnings are printed twice when starting Everest

Deprecation warnings that are raised during parsing of the Everest configuration are printed twice. This is likely because the configuration is parsed twice. At least once when staring Everest, then again when the experiment server is started.

It should be printed only once, at the earliest possible opportunity (i.e. when starting the cli)

### Steps to reproduce

```console
Run `everest run <config file>`
```

- [x] Python 3.11
- [x] Python 3.12
- [ ] macOS
- [ ] RHEL8
- [ ] RHEL9
- [x] Local queue
- [ ] LSF queue
- [ ] Slurm queue
- [ ] OpenPBS queue

verveerpj:
**Issue**
Resolves #13124

**Approach**
Suppress `ConfigWarning` while validating the configuration when starting via the experiment server

      --exec 'just rapid-tests'`)

## When applicable

### DIFF
--- a/src/ert/dark_storage/endpoints/experiment_server.py
+++ b/src/ert/dark_storage/endpoints/experiment_server.py
@@ -7,6 +7,7 @@ import queue
 import time
 import traceback
 import uuid
+import warnings
 from base64 import b64decode
 from queue import SimpleQueue
 from typing import Annotated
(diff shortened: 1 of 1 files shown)


src/ert/dark_storage/endpoints/experiment_server.py
```diff
@@ -192,7 +193,9 @@ async def start_experiment(
     _check_user(credentials)
     if shared_data.status.status == ExperimentState.pending:
         request_data = await request.json()
-        config = EverestConfig.with_plugins(request_data)
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", category=ConfigWarning)
```
berland, line 199: Should there be a comment here to address why we are filtering out this in this place?

verveerpj: Good point, will do.

PR #13146, opened 2026-03-16
Use monotonic() instead of time() for measuring duration

### LINKED ISSUE #13111: time.time() is used for measuring duration
scheduler.py, job.py and lsf_driver.py (among others) are using `time.time()` to measure duration.

`time.time()` can go backwards in the case of DST changes, and NTP adjustments, and `time.monotonic()` should be used instead.

This is involved in submit_sleep for example, and might create havoc especially around DST changes.

eqbech:
**Issue**
Resolves #13111

**Approach**
Switch to using `monotonic()` when measuring duration as this clock only moves forward and is not affected by system clock updates.

### DIFF
--- a/src/ert/scheduler/job.py
+++ b/src/ert/scheduler/job.py
@@ -82,8 +82,8 @@ class Job:
         self._scheduler: Scheduler = scheduler
         self._message: str = ""
         self._requested_max_submit: int | None = None
-        self._start_time: float | None = None
-        self._end_time: float | None = None
+        self._start_time_monotonic: float | None = None
+        self._end_time_monotonic: float | None = None
         self._remaining_file_verification_time = self.DEFAULT_FILE_VERIFICATION_TIMEOUT
         self._previous_file_verification_time = self._remaining_file_verification_time
         self._started_killing_by_evaluator: bool = False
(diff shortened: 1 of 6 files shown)


src/ert/scheduler/job.py
```diff
@@ -160,7 +160,7 @@ async def _submit_and_run_once(self, sem: asyncio.BoundedSemaphore) -> None:
 
             await self._send(JobState.PENDING)
             await self.started.wait()
-            self._start_time = time.time()
+            self._start_time = time.monotonic()
```
berland: Suggest to rename this to property to `_start_time_monotonic` to avoid anyone thinking that this float can be used as a timestamp.

src/ert/scheduler/job.py
```diff
@@ -438,7 +438,7 @@ async def _send(self, state: JobState) -> None:
                 await self._handle_aborted()
             case JobState.COMPLETED:
                 event = RealizationSuccess(real=str(self.iens))
-                self._end_time = time.time()
+                self._end_time = time.monotonic()
```
berland: same rename here.

src/ert/scheduler/scheduler.py
```diff
@@ -48,15 +48,13 @@ class SubmitSleeper:
 
     def __init__(self, submit_sleep: float) -> None:
         self._submit_sleep = submit_sleep
-        self._last_started = (
-            time.time() - submit_sleep
-        )  # Allow the first to start immediately
+        self._last_started = time.monotonic() - submit_sleep
```
berland: `_last_started_monotonic`

src/ert/scheduler/scheduler.py
```diff
@@ -48,15 +48,13 @@ class SubmitSleeper:
...
 
     async def sleep_until_we_can_submit(self) -> None:
-        now = time.time()
+        now = time.monotonic()
```
berland: we need some better naming here, or a comment to the function that all time values are monotonic.

